PR #16, opened 2025-12-14
Custom sleep screen support

treetrum:
### Changes

This PR adds the ability for the user to supply a custom `sleep.bmp` image at the root of the SD card that will be shown instead of the default sleep screen if present.

### Remaining work:

This PR is currently work in progress. Have opened as draft for visibility to prevent duplicate efforts by the community. The following is my high level plan of attack for getting this PR across the line.

TODO:

- [X] Support 1bit
- [X] Support 8bit
- [x] Support 24bit
- [X] Support 32bit (with alpha-channel ignored)
- [X] Test lots of images. Use https://www.readme.club/wallpapers as a good source of test images
- [ ] Update docs

### Follow up PRs:

- [ ] Handle incorrectly sized images better
- [ ] Greyscale rendering
- [ ] Allow for mutliple boot images (maybe inside a ./sleep directory) that get randomly chosen

### DIFF
--- a/lib/BmpReader/BmpReader.cpp
+++ b/lib/BmpReader/BmpReader.cpp
@@ -50,9 +50,9 @@ const char* BmpReader::errorToString(BmpReaderError err) {
     case BmpReaderError::BadPlanes:
       return "BadPlanes (!= 1)";
     case BmpReaderError::UnsupportedBpp:
-      return "UnsupportedBpp (expected 24)";
+      return "UnsupportedBpp (expected 24, 32 or 1)";
     case BmpReaderError::UnsupportedCompression:
-      return "UnsupportedCompression (expected BI_RGB)";
+      return "UnsupportedCompression (expected BI_RGB or BI_BITFIELDS for 32bpp)";
     case BmpReaderError::BadDimensions:
       return "BadDimensions";
     case BmpReaderError::SeekPixelDataFailed:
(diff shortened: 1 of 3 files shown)


lib/BmpReader/BmpReader.cpp
```diff
@@ -0,0 +1,210 @@
...
+  if (is8Bit) {
+    bytesPerPixel = 1u;
+  } else if (is32Bit) {
+    bytesPerPixel = 4u;
+  } else if (is24Bit) {
+    bytesPerPixel = 3u;
+  }
+  const uint32_t srcBytesPerRow =
+      is1Bit ? ((uint32_t)srcW + 7u) / 8u : (uint32_t)srcW * bytesPerPixel;  // bpp==1 ignores bytesPerPixel
+  const uint32_t srcRowStride = (srcBytesPerRow + 3u) & ~3u;
```
daveallie, line 156: A slightly easier way to get bytes per row/stride (from https://en.wikipedia.org/wiki/BMP_file_format#Pixel_storage) is to just go: `(srcW * bytesPerPixel + 31) / 32 * 4` and rely on the integer rounding, I find the bit-math / masking of anding with `0xFFFFFFFFB` kind of unnecessarily complex.

lib/BmpReader/BmpReader.cpp
```diff
@@ -0,0 +1,210 @@
...
+  if (srcH <= 0) return BmpReaderError::BadDimensions;
+
+  // Output dimensions after 90° CCW rotation
+  out.width = (int)srcH;
+  out.height = (int)srcW;
+
+  const size_t outBytesPerRow = (size_t)(out.width + 7) / 8;
+  out.len = outBytesPerRow * (size_t)out.height;
+
+  out.data = (uint8_t*)malloc(out.len);
```
daveallie, line 123: This is a bit borderline event for 1-bit images the size of the screen (800x480) - 48kB of space.
When we start to look at grayscale (2-bit), or downscaling larger images (1000x500) this is going to be problematic.

e.g. 1000x500 at 2-bit = 125kB

125kB is likely more than we can afford here. We might need to re-architect this to not use an internal buffer and instead ensure we're always writing straight to the existing screen buffer.

treetrum:
Hey @daveallie thanks for the feedback! Your comments make total sense 🙂 

I definitely haven't abandoned this - just had some personal stuff pop up and haven't had the chance to continue working on this. 

Not sure I'll be able to get back onto this now until probably early next week - so feel free to take over this one if you'd rather move this work along before then. I have rebased this onto the latest master so there's no more merge-conflicts.

Current state of this one is that I've been able to get the images showing for a wide range of test images from the community site linked in the description. Biggest omission for now (aside from your feedback) is more gracefully handling images that aren't the exact screen size. Though I imagine we could handle that in follow-ups?

lib/BmpReader/BmpReader.cpp
treetrum: Yep, makes sense! I'll admit that I'm very outside of my comfort zone with this low-level of programming 😅 So not surprised that there's better ways to do things.

daveallie:
Work continued in #57

PR #18, opened 2025-12-14
Settings Screen and first 2 settings

jonasdiemer:
Hi,

consider this a bit of a PoC (because it's mostly vibe coded).

Settings are stored in a global singleton that de/serializes to `sd/.crosspoint/settings.bin`.

The settings screen can be reached from the main screen (file browser) by going BACK. You leave it also by going BACK. OK_BTN  toggles the currently selected setting. I've opted for a `>` to mark the setting as it's more intuitive with only 2 settings (later, we could use inversion of colors like in the file browser).

There are two bool settings currently:
1. White sleep screen (avoids the inverting of the screen)
2. Extra paragraph spacing (to avoid spacing between "normal" paragraphs)

For the latter, I modified the epub parser in the lib to expose this behavior. It's also integrated into the cache so the file correctly re-indexes.

### DIFF
--- a/src/CrossPointSettings.h
+++ b/src/CrossPointSettings.h
@@ -1,4 +1,5 @@
 #pragma once
+#include <cstdint>
 #include <iosfwd>
 
 class CrossPointSettings {
@@ -15,10 +16,10 @@ class CrossPointSettings {
   CrossPointSettings& operator=(const CrossPointSettings&) = delete;
 
   // Sleep screen settings
-  bool whiteSleepScreen = false;
+  uint8_t whiteSleepScreen = 0;
 
   // Text rendering settings
-  bool extraParagraphSpacing = true;
+  uint8_t extraParagraphSpacing = 1;
 
   ~CrossPointSettings() = default;
 


jonasdiemer:
This addresses #13

lib/Epub/Epub/Section.cpp
daveallie: Can you please bump this to 5 as we've changed the contents of the file

src/screens/SleepScreen.h
```diff
@@ -1,8 +1,11 @@
 #pragma once
 #include "Screen.h"
 
+class CrossPointSettings;
+
 class SleepScreen final : public Screen {
  public:
-  explicit SleepScreen(GfxRenderer& renderer, InputManager& inputManager) : Screen(renderer, inputManager) {}
+  explicit SleepScreen(GfxRenderer& renderer, InputManager& inputManager)
+      : Screen(renderer, inputManager) {}
```
daveallie: Can revert these changes, looks like they're no longer needed since using `SETTINGS`

src/screens/FileSelectionScreen.cpp
```diff
@@ -91,11 +91,16 @@ void FileSelectionScreen::handleInput() {
...
-    updateRequired = true;
+  } else if (inputManager.wasPressed(InputManager::BTN_BACK)) {
+    if (basepath != "/") {
+      basepath = basepath.substr(0, basepath.rfind('/'));
+      if (basepath.empty()) basepath = "/";
+      loadFiles();
+      updateRequired = true;
+    } else {
+      // At root level, go to settings
+      onSettingsOpen();
```
daveallie, line 102: This is fine for now, but I'd love to follow up in a future change with a landing page similar to XTOS with a settings option and a read option.

src/screens/SettingsScreen.h
```diff
@@ -0,0 +1,42 @@
...
+#include <vector>
+
+#include "Screen.h"
+
+class CrossPointSettings;
+
+// Structure to hold setting information
+struct SettingInfo {
+  const char* name;       // Display name of the setting
+  bool CrossPointSettings::* valuePtr;  // Pointer to member in CrossPointSettings
```
daveallie: Might be good to consider treating this as a `uint8_t` from day 1, `0` and `1` can be used for boolean checking, but it also opens up setting enums in settings too, for example: left, right, justified rendering.

src/main.cpp
```diff
@@ -131,6 +133,7 @@ void waitForPowerRelease() {
 // Enter deep sleep mode
 void enterDeepSleep() {
   exitScreen();
+  SETTINGS.saveToFile();
```
daveallie: This seems unnecessary, we currently save settings on change and on exiting the setting screen, no need to do it on shutdown as well

src/CrossPointSettings.cpp
```diff
@@ -0,0 +1,51 @@
...
+constexpr char SETTINGS_FILE[] = "/sd/.crosspoint/settings.bin";
+
+bool CrossPointSettings::saveToFile() const {
+  // Make sure the directory exists
+  SD.mkdir("/.crosspoint");
+  
+  std::ofstream outputFile(SETTINGS_FILE);
+  serialization::writePod(outputFile, SETTINGS_FILE_VERSION);
+  serialization::writePod(outputFile, whiteSleepScreen);
+  serialization::writePod(outputFile, extraParagraphSpacing);
```
daveallie, line 25: To make settings expandable in the future without losing existing settings when adding new ones, I'd like to suggest a change to this file format.

```
0x00-0x01 - <VERSION> (1 for now)
0x01-0x02 - <Setting count> (2 settings for now)
for each setting (address offset by header above):
  0x00-0x01 - key length in bytes (K)
  0x01-(0x01+K) - key name
  (0x01+K)-(0x02+K) - setting value
```

Then when loading settings from file we could still load old setting by key name without having to throw away the file each time.

What do you think?

src/screens/SettingsScreen.cpp
```diff
@@ -0,0 +1,142 @@
...
+    
+    // Draw setting name and value
+    renderer.drawText(UI_FONT_ID, 20, settingY, settingsList[i].name);
+    bool value = SETTINGS.*(settingsList[i].valuePtr);
+    renderer.drawText(UI_FONT_ID, pageWidth - 80, settingY, value ? "ON" : "OFF");
+  }
+
+  // Draw help text
+  renderer.drawText(SMALL_FONT_ID, 20, pageHeight - 40,
+                   "Press OK to toggle, BACK to save & exit");
```
daveallie: nit: This text is 40px from the bottom of the screen, but the instruction text on the file selection screen is 30px from the bottom, they should match to avoid the text jumping when swapping between pages.

daveallie (changes requested):
Looking good, left a few comments, let me know if you have followup questions, probably biggest change is around the settings file format.

Also, I may setup a small github action to validate it, but could you please run `bin/clang-format-fix` to standardise the file formatting.

lib/Epub/Epub/Section.cpp
jonasdiemer: Sure. 

I am wondering what are the semantics of FILE_VERSION (same for e.g. settings and state). I would assume it could be good to make the code "a bit backwards compatible" by allowing us to add content to the file at the end without bumping the version. The deserialization would have to handle the case that there is not enough data (and then populate the missing pieces by default values).

src/main.cpp
jonasdiemer: Correct, this was from early testing.

src/CrossPointSettings.cpp
jonasdiemer: More or less what I thought about - see above comment on section file format. I would avoid keys and just go by position, adding the rule to not add new settings in between (without bumping the version).

Later, when we use NVS, I think key-value comes out of the box.

lib/Epub/Epub/Section.cpp
daveallie: I think for the added complexity, understanding defaults everywhere there's a cached file (specifically when the content is rebuildable) is not worth it. In this case we'd need to couple `Section` with the understanding of the default setting for the `extraParagraphSpacing` parameter (but more specifically not the default, but the behaviour prior to the setting existing, as those could be different things).

The forced rebuild of the cache costs users a little of time, but saves a huge backwards compatibility piece of work, which imo is worth the trade.

For places where we can't rebuild data (for example settings parameters), we should ensure that settings are retained so that's a bit of a different story and in those cases we should make sure the file format can support addition and removal of settings between versions without data loss.

src/CrossPointSettings.cpp
daveallie: My only thought is that we'd lose data if we removed a setting in the future, but I guess that's pretty fair given we're probably going to change the storage mechanism in the future, happy to keep your original implementation

daveallie:
Also, if the reason you wanted a white sleep screen is because of white streaking on the black one, that should be fixed here: https://github.com/daveallie/crosspoint-reader/pull/21, if it's just aesthetic preference, then ignore me!

jonasdiemer:
> Also, if the reason you wanted a white sleep screen is because of white streaking on the black one, that should be fixed here: #21, if it's just aesthetic preference, then ignore me!

No, I want it because I think it looks nicer on a white device ;)
This option is more of a PoC, we will remove it once we have custom sleep screens.

src/CrossPointSettings.cpp
jonasdiemer: Will rework to not loose data if we add new settings at the end with minimal invasiveness.

jonasdiemer:
I've reworked the PR, hopefully caught all topics.

PR #25, opened 2025-12-15
TOC location fix

osteotek:
## Summary

* Rely on media-type="application/x-dtbncx+xml" to find TOC instead of hardcoded values

## Additional Context

* Most of my epubs don't have id==ncx for toc file location. I think this media-type is EPUB standard


osteotek:
Found some bugs, converted to WIP

PR #27, opened 2025-12-15
Add Cyrillic range to fonts

osteotek:
## Summary

* Adds glyphs from the Cyrillic range to default fonts


daveallie:
Looks good @osteotek, could you run the format script `bin/clang-format-fix` to get this green and then we can merge this

daveallie (approved):
Small bump in flash usage, but worth it for readability

Before
```
RAM:   [====      ]  35.9% (used 117624 bytes from 327680 bytes)
Flash: [=         ]  14.3% (used 939820 bytes from 6553600 bytes)
```

After
```
RAM:   [====      ]  35.9% (used 117624 bytes from 327680 bytes)
Flash: [==        ]  16.6% (used 1085260 bytes from 6553600 bytes)
```

PR #29, opened 2025-12-16
Support uploading files via Wifi

daveallie:
## Summary

- Added UploadFileScreen which starts UploadServer (ESPAsyncWebServer under the hood)
  - Starts a wifi access point with a captive webpage to allow for uploads
- Currently inaccessible (I've been testing by just swapping it out with the settings screen in main), will need to restructure the home screen to allow for this and settings to appear

## Additional Context

- There's some UI work to do still, showing some instructions on the device
- [BLOCKING] Visiting the UploadFileScreen and then leaving it does not free back all the memory it allocates, I suspect there's something `ESPAsyncWebServer` or `WiFi` which creates static items
  - It's sucking up like 40kB, so it can't really be merged as it, one way would be to restart the device when leaving that screen as a workaround, but I'd love to understand what is soaking up that heap space
- Additionally, reserved memory for the firmware is getting kind of high with the change 😩

Addresses: https://github.com/daveallie/crosspoint-reader/discussions/15

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -21,8 +21,10 @@
 #include "screens/EpubReaderScreen.h"
 #include "screens/FileSelectionScreen.h"
 #include "screens/FullScreenMessageScreen.h"
+#include "screens/HomeScreen.h"
 #include "screens/SettingsScreen.h"
 #include "screens/SleepScreen.h"
+#include "screens/UploadFileScreen.h"
 
 #define SPI_FQ 40000000
 // Display SPI pins (custom pins for XteinkX4, not hardware SPI defaults)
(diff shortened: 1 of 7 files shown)


daveallie:
This may be useful: https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-reference/network/esp_wifi.html#_CPPv415esp_wifi_deinitv

daveallie:
Pausing work here, likely going to get these changes in instead: https://github.com/daveallie/crosspoint-reader/discussions/15#discussioncomment-15270831

daveallie:
Closing in favour of https://github.com/daveallie/crosspoint-reader/pull/41

PR #31, opened 2025-12-16
Wrap-around navigation in Settings.

jonasdiemer:
Let me know if you prefer larger PRs with various changes combined.


daveallie:
Small changes like this are preferred as they're easier to parse and don't block small improvements on debates, thanks!

daveallie (approved):

PR #34, opened 2025-12-16
Use single buffer mode for EInkDisplay

daveallie:
## Summary

* Frees up 48kB of statically allocated RAM in exchange for 48kB just when grayscale rendering is needed

## Additional Context

* Upstream changes: https://github.com/open-x4-epaper/community-sdk/pull/7


daveallie:
Heard from CidVonHighwind that he was experiencing some troubles with allocating 48k of contiguous memory in the past, really want this to work, so going to merge it in and test it for a bit

PR #37, opened 2025-12-16
Adds auto-resume last book setting

osteotek:
## Summary

* setting to disable auto resume last book after waking up

### DIFF
--- a/src/screens/SettingsScreen.cpp
+++ b/src/screens/SettingsScreen.cpp
@@ -10,7 +10,7 @@
 const SettingInfo SettingsScreen::settingsList[SettingsScreen::settingsCount] = {
     {"White Sleep Screen", &CrossPointSettings::whiteSleepScreen},
     {"Extra Paragraph Spacing", &CrossPointSettings::extraParagraphSpacing},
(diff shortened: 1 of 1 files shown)


daveallie:
Hi @osteotek, I would love to know a little more about what's motivating this change? 

Happy to debate it, but my personal believe is that as an ereader device, it should be helping you get straight into the book. 

If it's a question of retaining the last screen before sleep, I can understand that and perhaps we should just clear `CrossPointState`'s `openEpubPath` when leaving a book. But I don't think there's ever a strong case for turning off the device when you're in a book, and then turning it back on to the home screen.

osteotek:
@daveallie That was my preferred setting for debugging various epubs, but I agree it probably does not make sense for end user. I'll close it

jlaunay:
I thought it was a good option; if you have several books you're reading at the same time, you don't necessarily want to automatically open the last one, but have the choice.

daveallie:
The one extra button press you save to go back from your last open book to look at the file list again doesn't really ever seem worth turning on the setting? Are you really changing books *more often* than staying in the same book each time you turn on the device?

Issue #1, opened 2025-12-04
Add git clone instruction to README


drbourbon:
Just a minor suggestion. For people not used to git submodules, an explicit README instruction on how to properly git clone this would probably ease the first build process:

```
git clone --recursive https://github.com/daveallie/crosspoint-reader
cd crosspoint-reader
```

Also, for those who installed PlatformIO IDE but still want to build/upload from command line, the following one-shot copy&paste would also probably help:

```
git clone --recursive https://github.com/daveallie/crosspoint-reader
cd crosspoint-reader
$HOME/.platformio/penv/bin/pio run --target upload
```

(PS: very promising and well structured work, thanks!)

daveallie:
Added a small instruction for cloning the repo, thanks!

PR #41, opened 2025-12-17
Add connect to Wifi and File Manager Webserver

olearycrew:
## Summary

- **What is the goal of this PR?**  
  Implements wireless EPUB file management via a built-in web server, enabling users to upload, browse, organize, and delete EPUB files from any device on the same WiFi network without needing a computer cable connection.

  - **New Web Server** ([`CrossPointWebServer.cpp`](src/CrossPointWebServer.cpp), [`CrossPointWebServer.h`](src/CrossPointWebServer.h)):
    - HTTP server on port 80 with a responsive HTML/CSS interface
    - Home page showing device status (version, IP, free memory)
    - File Manager with folder navigation and breadcrumb support
    - EPUB file upload with progress tracking
    - Folder creation and file/folder deletion
    - XSS protection via HTML escaping
    - Hidden system folders (`.` prefixed, "System Volume Information", "XTCache")
  
  - **WiFi Screen** ([`WifiScreen.cpp`](src/screens/WifiScreen.cpp), [`WifiScreen.h`](src/screens/WifiScreen.h)):
    - Network scanning with signal strength indicators
    - Visual indicators for encrypted (`*`) and saved (`+`) networks
    - State machine managing: scanning, network selection, password entry, connecting, save/forget prompts
    - 15-second connection timeout handling
    - Integration with web server (starts on connect, stops on exit)
  
  - **WiFi Credential Storage** ([`WifiCredentialStore.cpp`](src/WifiCredentialStore.cpp), [`WifiCredentialStore.h`](src/WifiCredentialStore.h)):
    - Persistent storage in `/sd/.crosspoint/wifi.bin`
    - XOR obfuscation for stored passwords (basic protection against casual reading)
    - Up to 8 saved networks with add/remove/update operations
  
  - **On-Screen Keyboard** ([`OnScreenKeyboard.cpp`](src/screens/OnScreenKeyboard.cpp), [`OnScreenKeyboard.h`](src/screens/OnScreenKeyboard.h)):
    - Reusable QWERTY keyboard component with shift support
    - Special keys: Shift, Space, Backspace, Done
    - Support for password masking mode
  
  - **Settings Screen Integration** ([`SettingsScreen.h`](src/screens/SettingsScreen.h)):
    - Added WiFi action to navigate to the new WiFi screen
  
  - **Documentation** ([`docs/webserver.md`](docs/webserver.md)):
    - Comprehensive user guide covering WiFi setup, web interface usage, file management, troubleshooting, and security notes
    - See this for more screenshots!
    - Working "displays the right way in GitHub" on my repo: https://github.com/olearycrew/crosspoint-reader/blob/feature/connect-to-wifi/docs/webserver.md

**Video demo**

https://github.com/user-attachments/assets/283e32dc-2d9f-4ae2-848e-01f41166a731

## Additional Context

- **Security considerations**: The web server has no authentication—anyone on the same WiFi network can access files. This is documented as a limitation, recommending use only on trusted private networks. Password obfuscation in the credential store is XOR-based, not cryptographically secure.

- **Memory implications**: The web server and WiFi stack consume significant memory. The implementation properly cleans up (stops server, disconnects WiFi, sets `WIFI_OFF` mode) when exiting the WiFi screen to free resources.

- **Async operations**: Network scanning and connection use async patterns with FreeRTOS tasks to prevent blocking the UI. The display task handles rendering on a dedicated thread with mutex protection.

- **Browser compatibility**: The web interface uses standard HTML5/CSS3/JavaScript and is tested to work with all modern browsers on desktop and mobile.

### DIFF
--- a/src/activities/settings/SettingsActivity.cpp
+++ b/src/activities/settings/SettingsActivity.cpp
@@ -9,7 +9,7 @@
 
 const SettingInfo SettingsActivity::settingsList[settingsCount] = {
     {"White Sleep Screen", SettingType::TOGGLE, &CrossPointSettings::whiteSleepScreen},
-    {"Extra Paragraph Spacing", SettingType::TOGGLE, &CrossPointSettings::extraParagraphSpacing}},
+    {"Extra Paragraph Spacing", SettingType::TOGGLE, &CrossPointSettings::extraParagraphSpacing},
     {"Short Power Button Click", SettingType::TOGGLE, &CrossPointSettings::shortPwrBtn}};
 
 void SettingsActivity::taskTrampoline(void* param) {


daveallie:
Amazing work here @olearycrew, I haven't taken a full dive into it yet, I plan to take a look in a couple of hours, but I do have some surface level feedback which I'd love to get down:
1. I'll add it in a small PR in the next hour or so, but I'd love to add this a more global home screen instead of to the settings screen, I'll extract the work from https://github.com/daveallie/crosspoint-reader/pull/29/changes/c8cdf5cd5a9dd8317848aee53584d3474ef27572 into a separate PR.
   - Edit: Added here: https://github.com/daveallie/crosspoint-reader/pull/42
3. I don't love having the raw HTML in a long string in the CPP file, it makes it a little difficult to jump around. I'd prefer if you could include the prebuild script from my PR for HTML management, allowing us to split out the HTML files. It makes it a lot easier to edit the HTML in isolation (feel free to update the script if needed).
   - https://github.com/daveallie/crosspoint-reader/pull/29/changes#diff-66db77df7fce72feeec89b6163244b0c29c911150d98d6f1e84189cd60333222
   - https://github.com/daveallie/crosspoint-reader/pull/29/changes#diff-4446afd728a4f34cbcddc306a9cb6be845d1a61c216076a295683bcc9c106724R34-R35
   - https://github.com/daveallie/crosspoint-reader/pull/29/changes#diff-bc37d034bad564583790a46f19d807abfe519c5671395fd494d8cce506c42947R6
   - https://github.com/daveallie/crosspoint-reader/pull/29/changes#diff-cb44a29d4252b9f5bae91eeeaa1370a301337a664d3a350b9890229bcfa2c9c4
   - https://github.com/daveallie/crosspoint-reader/pull/29/changes#diff-6d24910682f37779076e666b5866c52ca45301fa4d3cc5b5f510123ba4863b8aR5-R6

daveallie:
There's a similar problem to what I was experiencing on my branch:

**Fresh boot, navigate to settings page**
```
[MEM] Free: 170564 bytes, Total: 237356 bytes, Min Free: 166568 bytes
```

**Running web-server once connected to WiFi (no clients)**
```
[MEM] Free: 120288 bytes, Total: 234444 bytes, Min Free: 117888 bytes
```

**Exit web-server, back on settings page**
```
[MEM] Free: 153700 bytes, Total: 235660 bytes, Min Free: 117664 bytes
```

Seemingly, there's 17kB of RAM which has been allocated but not freed up, switching to a single buffer for the display has granted a bit more headroom, but it's concerning that it's difficult to understand what is sucking that up

daveallie:
Perhaps this is a little too keen, but I'd love for the webpage to have a little more support for mobile screens. My workflow out in the wild will probably be connect to my phone hotspot to upload content to it.

The button dropdown end up off the left hand side of the screen and the table overflows on the right. Constraining the table and adding an overflow property should allow for horizontal scrolling on narrow viewports, I haven't actually looked at the button CSS yet.

Additionally on small viewports, we could tighten up those margins to free up a bit more horizontal space

CleanShot 2025-12-17 at 22 02 54

---

More of an overarching question:
I appreciate the work into the wifi AP connection code as it'll likely be beneficial in the future for other network related features, but did you consider using the softAP mode of the device to save needing an access point / typing in a wifi password on the device (which can be tedious).

daveallie:
Additionally, one more bit of more wide-scale feedback, I'd prefer to avoid globally allocating the server. Instead, we should have a network function / screen, which gets swapped to like any other screen, which can chose to render one of a few sub-screens (like the wifi connecting / connected screen), allocates and instantiates/deletes the server when needed, and relies on the main file's loop calling out to it to process connections. Feels like there's a slippery slope where we just start allocating everything in main and bloat out the loop, whereas I think delegated down to screens will keep the code cleaner.

On that note, I can do it separately, but `Screen::handleInput` should probably just be `Screen::loop` to be clearer. That way main loop handles any global responsibility and `Screen::loop` gets delegated down into to handle local responsibilities. Will mean a more stack like functional structure with pushing a popping onto a stack where the root level screen are setup in main, and sub screens can make their own choices.

daveallie:
FYI - re the above comment, I'm going to make the following change to the structure of the codebase, which should mean that the file manager webserver can become its own activity: https://github.com/daveallie/crosspoint-reader/pull/44

olearycrew:
> More of an overarching question:
> I appreciate the work into the wifi AP connection code as it'll likely be beneficial in the future for other network related features, but did you consider using the softAP mode of the device to save needing an access point / typing in a wifi password on the device (which can be tedious).

I did but I wanted to be able to do it without changing the network of my computer (and be able to get to it from a non-wifi computer in the future). I'd be happy to look at that though after we get this one sorted

olearycrew:
> FYI - re the above comment, I'm going to make the following change to the structure of the codebase, which should mean that the file manager webserver can become its own activity: #44

I've merged master and refactored to match this new pattern

olearycrew:
> I don't love having the raw HTML in a long string in the CPP file, it makes it a little difficult to jump around. I'd prefer if you could include the prebuild script from my PR for HTML management, allowing us to split out the HTML files. It makes it a lot easier to edit the HTML in isolation (feel free to update the script if needed).
> * [#29 (changes)](https://github.com/daveallie/crosspoint-reader/pull/29/changes#diff-66db77df7fce72feeec89b6163244b0c29c911150d98d6f1e84189cd60333222)
> * [#29 (changes)](https://github.com/daveallie/crosspoint-reader/pull/29/changes#diff-4446afd728a4f34cbcddc306a9cb6be845d1a61c216076a295683bcc9c106724R34-R35)
> * [#29 (changes)](https://github.com/daveallie/crosspoint-reader/pull/29/changes#diff-bc37d034bad564583790a46f19d807abfe519c5671395fd494d8cce506c42947R6)
> * [#29 (changes)](https://github.com/daveallie/crosspoint-reader/pull/29/changes#diff-cb44a29d4252b9f5bae91eeeaa1370a301337a664d3a350b9890229bcfa2c9c4)
> * [#29 (changes)](https://github.com/daveallie/crosspoint-reader/pull/29/changes#diff-6d24910682f37779076e666b5866c52ca45301fa4d3cc5b5f510123ba4863b8aR5-R6)

Okay I just refactored this to more closely match your pattern of pre-building most of the HTML. I do agree it is much cleaner. Sorry - tbh this is my first ever C++ webserver 😄

olearycrew:
CleanShot 2025-12-17 at 20 13 22@2x

New mobile view

olearycrew:
> There's a similar problem to what I was experiencing on my branch:
> 
> **Fresh boot, navigate to settings page**
> 
> ```
> [MEM] Free: 170564 bytes, Total: 237356 bytes, Min Free: 166568 bytes
> ```
> 
> **Running web-server once connected to WiFi (no clients)**
> 
> ```
> [MEM] Free: 120288 bytes, Total: 234444 bytes, Min Free: 117888 bytes
> ```
> 
> **Exit web-server, back on settings page**
> 
> ```
> [MEM] Free: 153700 bytes, Total: 235660 bytes, Min Free: 117664 bytes
> ```
> 
> Seemingly, there's 17kB of RAM which has been allocated but not freed up, switching to a single buffer for the display has granted a bit more headroom, but it's concerning that it's difficult to understand what is sucking that up

So I think I found and fixed this issue. This is the summary from my AI agent when we were done debugging it together:

Great news! The diagnostic logs reveal the memory leak is much smaller than initially reported.

## Memory Leak Analysis:

**Actual WebServer Memory Leak: ~584 bytes** (not 17kB!)

### Breakdown:
- **WebServer allocates**: 1,612 bytes total
  - 304 bytes for WebServer object
  - 672 bytes for lambda route handlers  
  - 636 bytes for server internals
  
- **WebServer cleanup recovers**: 1,028 bytes
  - Leaves ~584 bytes leaked

### The "17kB Leak" Explained:

The original PR report was measuring total memory including WiFi, which uses ~32kB. The WiFi memory IS properly freed (151,080 - 118,988 = 32,092 bytes recovered).

### Why server->stop() Appears to Consume Memory:

Interestingly, `server->stop()` actually *loses* 2,080 bytes (120,040 → 117,960), then `delete server` recovers 1,028 bytes. This suggests the WebServer library's stop() method may allocate cleanup buffers or have a bug.

### Memory Behavior Over Time:

After 90 seconds, memory stabilizes at 153,440 bytes, which is actually 2,208 bytes BETTER than immediately after exit (155,648 bytes). This indicates delayed garbage collection or heap defragmentation.

## Verdict:

✅ **The ~584 byte leak is acceptable** for an embedded system and likely unavoidable due to the WebServer library's lambda capture implementation.

✅ **The crash is fixed** - the mutex deletion bug in WifiScreen::onExit() has been resolved.

✅ **WiFi memory is properly freed** - the 32kB WiFi allocation is correctly released.

The diagnostic logs can be kept for future debugging or removed if desired.

olearycrew:
@daveallie I *think* I've addressed all the asks from your review

daveallie:
I'm seeing uploads fail about 1/2 times before it succeeds, something we can dive further into if others are seeing the same issues.

Also saw this crash when entering wifi mode for a bit, and then leaving it:
```
[40016] [MEM] Free: 120212 bytes, Total: 234396 bytes, Min Free: 91720 bytes
[48426] [WEB] handleClient active, server running on port 80
[50016] [MEM] Free: 120212 bytes, Total: 234396 bytes, Min Free: 91720 bytes
[53494] [WIFI] [MEM] Free heap at onExit start: 120212 bytes
[53494] [WIFI] [MEM] Free heap after scanDelete: 120212 bytes
[53494] [WEB] [MEM] Free heap before stop: 120212 bytes
[53495] [WEB] [MEM] Free heap after server->stop(): 120532 bytes
[53495] [WEB] Web server stopped
[53495] [WEB] [MEM] Free heap after delete server: 121560 bytes
[53495] [WEB] [MEM] Free heap final: 121560 bytes
[53495] [WIFI] [MEM] Free heap after webserver stop: 121560 bytes
[53500] [WIFI] [MEM] Free heap after WiFi disconnect: 143428 bytes
Guru Meditation Error: Core  0 panic'ed (Load access fault). Exception was unhandled.

Core  0 register dump:
MEPC    : 0x40043d16  RA      : 0x4208ab26  SP      : 0x3fcb90d0  GP      : 0x3fc8ee00
TP      : 0x3fc2bda8  T0      : 0x00000000  T1      : 0x42037a32  T2      : 0x00000001
S0/FP   : 0x3fcdf86c  S1      : 0x3fca0720  A0      : 0x00000000  A1      : 0x3fcc2bea
A2      : 0x3fcb90fc  A3      : 0x3fcaca78  A4      : 0x3fca4794  A5      : 0x00000000
A6      : 0x3fcb9278  A7      : 0x00000001  S2      : 0x3fce0000  S3      : 0x3fce0000
S4      : 0x00000000  S5      : 0x0000002a  S6      : 0x3fcc2bea  S7      : 0x00000000
S8      : 0x3c172e04  S9      : 0x3fcba2a4  S10     : 0x3fca46dc  S11     : 0x00000000
  #0  0x3c172e04 in ?? at /home/runner/work/esp32-arduino-lib-builder/esp32-arduino-lib-builder/esp-idf/components/lwip/lwip/src/netif/ethernet.c:64

T3      : 0x00000000  T4      : 0x0000000a  T5      : 0x00000000  T6      : 0x00000000
MSTATUS : 0x00001881  MTVEC   : 0x40380001  MCAUSE  : 0x00000005  MTVAL   : 0x00000018
MHARTID : 0x00000000

Stack memory:
3fcb90d0: 0x3fcb9278 0x3c172dfc 0x3fcba2a4 0xffffffff 0x3fce0000 0x3fca0720 0x00000000 0x4208ab10
  #0  0x3c172dfc in ?? at /home/runner/work/esp32-arduino-lib-builder/esp32-arduino-lib-builder/esp-idf/components/lwip/lwip/src/netif/ethernet.c:63
  #1  0x4208ab10 in ieee80211_output_do at ??:?

3fcb90f0: 0x0000000e 0x00000000 0x0000002c 0x00000000 0x0000000a 0x3fc9ed18 0x0000002c 0x3fcba1c4
  #0  0x3fc9ed18 in ?? at /home/runner/work/esp32-arduino-lib-builder/esp32-arduino-lib-builder/esp-idf/components/lwip/lwip/src/core/ipv4/etharp.c:106

3fcb9110: 0x3fcb9278 0x3c172dfc 0x3fcba2a4 0x00000006 0x3fcba1c0 0x3fcc2bd0 0x3fcc2bf8 0x4204c436
  #0  0x3c172dfc in ?? at /home/runner/work/esp32-arduino-lib-builder/esp32-arduino-lib-builder/esp-idf/components/lwip/lwip/src/netif/ethernet.c:63
  #1  0x4204c436 in low_level_output at /home/runner/work/esp32-arduino-lib-builder/esp32-arduino-lib-builder/esp-idf/components/lwip/port/esp32/netif/wlanif.c:137

```

src/WifiCredentialStore.cpp
```diff
@@ -0,0 +1,158 @@
...
+// File format version
+constexpr uint8_t WIFI_FILE_VERSION = 1;
+
+// WiFi credentials file path
+constexpr char WIFI_FILE[] = "/sd/.crosspoint/wifi.bin";
+
+// Obfuscation key - "CrossPoint" in ASCII
+// This is NOT cryptographic security, just prevents casual file reading
+constexpr uint8_t OBFUSCATION_KEY[] = {0x43, 0x72, 0x6F, 0x73, 0x73, 0x50, 0x6F, 0x69, 0x6E, 0x74};
+constexpr size_t KEY_LENGTH = sizeof(OBFUSCATION_KEY);
```
daveallie, line 22: nit: if these are local static variables, we can wrap them in an anonymous namespace to avoid them leaking from this file:

```cpp
namespace {
constexpr ...
...
}
```

src/CrossPointWebServer.cpp
```diff
@@ -0,0 +1,706 @@
...
+
+#include <algorithm>
+
+#include "config.h"
+#include "html/FilesPageFooterHtml.generated.h"
+#include "html/FilesPageHeaderHtml.generated.h"
+#include "html/HomePageHtml.generated.h"
+
+// Global instance
+CrossPointWebServer crossPointWebServer;
```
daveallie: Instead of instantiating this as a global instance and listening to events as part of the main loop, can we remove this extern instance, move the `CrossPointWebServer` files/logic into `activities/network/server` and have it be instantiated by `CrossPointWebServerActivity`

src/activities/network/server/CrossPointWebServer.cpp
```diff
@@ -0,0 +1,706 @@
...
+#include "html/FilesPageHeaderHtml.generated.h"
+#include "html/HomePageHtml.generated.h"
+
+// Global instance
+CrossPointWebServer crossPointWebServer;
+
+// Folders/files to hide from the web interface file browser
+// Note: Items starting with "." are automatically hidden
+static const char* HIDDEN_ITEMS[] = {"System Volume Information", "XTCache"};
+static const size_t HIDDEN_ITEMS_COUNT = sizeof(HIDDEN_ITEMS) / sizeof(HIDDEN_ITEMS[0]);
```
daveallie: nit, also wrap these in a anonymous namespace

src/activities/util/KeyboardEntryActivity.h
daveallie, line 1: Given the reusablity and utility of this, I think it should:
1. Be renamed to `KeybaordEntryActivity`
2. Subclass `Activity`
3. Be moved into `activities/util/KeyboardEntryActivity.h`

That way other activities, can start this activity from within themselves and exit out of it taking out the input.

src/activities/network/WifiScreen.cpp
```diff
@@ -0,0 +1,700 @@
...
+  }
+
+  // Small delay to ensure task is fully deleted before cleaning up mutex
+  vTaskDelay(10 / portTICK_PERIOD_MS);
+
+  // Now safe to delete the mutex
+  if (renderingMutex) {
+    vSemaphoreDelete(renderingMutex);
+    renderingMutex = nullptr;
+  }
```
daveallie: This is dangerous, as if the task has the mutex when it's deleted it can result in a hang or crash, what I've found works on other screens is to first acquire the semaphore before deleting the task, like this:

```cpp
xSemaphoreTake(renderingMutex, portMAX_DELAY);
if (displayTaskHandle) {
  vTaskDelete(displayTaskHandle);
  displayTaskHandle = nullptr;
}
vSemaphoreDelete(renderingMutex);
renderingMutex = nullptr;
```

src/activities/network/WifiSelectionActivity.cpp
daveallie, line 1: This is handling a few responsibilities:
- Starting / stopping wifi
- Listing / connecting to APs
- Controlling the `CrossPointWebServer`

IMO we should split this one up a bit into:
- `CrossPointWebServerActivity` which is the entry point, handles turning the wifi on / off
  - When it's entered, it immediately turns on the wifi, and launches `WifiSelectionActivity`
  - When `WifiSelectionActivity` calls `onComplete` (parameter for the activity), `CrossPointWebServerActivity` can check if you're connected to the wifi and initialize / launch the `CrossPointWebServer`, handling requests as part of the `loop` function of `CrossPointWebServerActivity`
  - When this activity is exited, it'll clean up the server, shut down the wifi, etc
- `WifiSelectionActivity` is responsible for scanning Wifi APs and then exiting once connected or cancelled
  - It'll enter scanning mode, list wifi, allow selection, launch `KeyboardEntryActivity` if needed, save the password back down, etc

There's probably even scope to split out the Wifi utility code a bit, but for now it can stay within the `WifiSelectionActivity`.

This should reduce the number of responsibilities this class has and better fit them into the "Activity" system.

src/main.cpp
```diff
@@ -222,7 +245,34 @@ void loop() {
...
+  }
+
+  unsigned long loopDuration = millis() - loopStartTime;
+  if (loopDuration > maxLoopDuration) {
+    maxLoopDuration = loopDuration;
+    if (maxLoopDuration > 50) {
+      Serial.printf("[%lu] [LOOP] New max loop duration: %lu ms (activity: %lu ms)\n", millis(), maxLoopDuration,
+                    activityDuration);
+    }
+  }
```
daveallie: As mentioned above all this logic can be encapsulated into `CrossPointWebServerActivity`

src/main.cpp
```diff
@@ -140,9 +143,16 @@ void onGoToReader(const std::string& initialEpubPath) {
...
+
+void onGoToWifi() {
+  exitActivity();
+  enterNewActivity(new WifiScreen(renderer, inputManager, onGoToSettings));
+}
+
 void onGoToSettings() {
   exitActivity();
-  enterNewActivity(new SettingsActivity(renderer, inputManager, onGoHome));
+  enterNewActivity(new SettingsActivity(renderer, inputManager, onGoHome, onGoToWifi));
```
daveallie: I'd love for the `CrossPointWebServerActivity` to be launchable from the home screen, rather than the settings screen, we should also update the label from "Wifi" to something a little more clear, perhaps "File transfer"?

olearycrew:
> Stack memory:
> 3fcb90d0: 0x3fcb9278 0x3c172dfc 0x3fcba2a4 0xffffffff 0x3fce0000 0x3fca0720 0x00000000 0x4208ab10
>   #0  0x3c172dfc in ?? at /home/runner/work/esp32-arduino-lib-builder/esp32-arduino-lib-builder/esp-idf/components/lwip/lwip/src/netif/ethernet.c:63
>   #1  0x4208ab10 in ieee80211_output_do at ??:?
> 
> 3fcb90f0: 0x0000000e 0x00000000 0x0000002c 0x00000000 0x0000000a 0x3fc9ed18 0x0000002c 0x3fcba1c4
>   #0  0x3fc9ed18 in ?? at /home/runner/work/esp32-arduino-lib-builder/esp32-arduino-lib-builder/esp-idf/components/lwip/lwip/src/core/ipv4/etharp.c:106
> 
> 3fcb9110: 0x3fcb9278 0x3c172dfc 0x3fcba2a4 0x00000006 0x3fcba1c0 0x3fcc2bd0 0x3fcc2bf8 0x4204c436
>   #0  0x3c172dfc in ?? at /home/runner/work/esp32-arduino-lib-builder/esp32-arduino-lib-builder/esp-idf/components/lwip/lwip/src/netif/ethernet.c:63
>   #1  0x4204c436 in low_level_output at /home/runner/work/esp32-arduino-lib-builder/esp32-arduino-lib-builder/esp-idf/components/lwip/port/esp32/netif/wlanif.c:137

Figured out this crash:

The stack trace reveals it's a timing issue in the ESP32's LWIP network stack:

```
ethernet.c:63 → etharp.c:106 → wlanif.c:137 → ieee80211_output_do
```

This shows the WiFi hardware trying to transmit packets (likely TCP responses from the web server) after the WiFi interface has been torn down by [WiFi.disconnect()](vscode-webview://1cskrj94bnpk2fui0mobuq3udm3shok4r4ed6dv8vsjg3vomecre/src/activities/network/WifiScreen.cpp:62) and [WiFi.mode(WIFI_OFF)](vscode-webview://1cskrj94bnpk2fui0mobuq3udm3shok4r4ed6dv8vsjg3vomecre/src/activities/network/WifiScreen.cpp:63).

Why This Happens

1. Web server receives a request and queues a response
1. Response gets buffered in LWIP's TCP/IP stack
1. The code code calls [crossPointWebServer.stop()](vscode-webview://1cskrj94bnpk2fui0mobuq3udm3shok4r4ed6dv8vsjg3vomecre/src/activities/network/WifiScreen.cpp:58)
1. Then we immediately call [WiFi.disconnect()](vscode-webview://1cskrj94bnpk2fui0mobuq3udm3shok4r4ed6dv8vsjg3vomecre/src/activities/network/WifiScreen.cpp:62)
1. WiFi interface is destroyed while LWIP still has pending packets
1. LWIP tries to send those packets → Load Access Fault (accessing freed WiFi interface memory)

src/activities/util/KeyboardEntryActivity.h
olearycrew: Done

src/activities/network/WifiScreen.cpp
olearycrew: Done

src/activities/network/WifiSelectionActivity.cpp
olearycrew: Okay I think I've addressed this like you said.

src/main.cpp
olearycrew: Moved to the new activity

src/main.cpp
olearycrew: Changed the label and added it to the new home screen

src/CrossPointWebServer.cpp
olearycrew: Done

src/WifiCredentialStore.cpp
olearycrew: Done

src/activities/network/server/CrossPointWebServer.cpp
olearycrew: Done

olearycrew:
@daveallie okay take another look. I *think* I've made the structural changes you were talking about

src/main.cpp
```diff
@@ -222,7 +236,29 @@ void loop() {
...
+    }
+  }
+
+  lastLoopTime = loopStartTime;
+
+  // Add delay at the end of the loop to prevent tight spinning
+  // When webserver is running, use yield() instead of delay for faster response
+  // When webserver is not running, use longer delay to save power
+  if (webServerActivity && webServerActivity->isWebServerRunning()) {
+    yield();  // Give FreeRTOS a chance to run tasks, but return immediately
```
daveallie: One final thing, but otherwise looks great, I'd like not to expose this underlying implementation detail so specifically in `main.cpp`. Instead, let's create a new opportunity for Activity to set its own behaviour.

If we define the following in `Activity`:
```cpp
virtual bool skipLoopDelay() { return false; }
```

It gives each activity the opportunity to define the global behaviour a bit more generically without needing the caller to understand the internal mechanisms. Obviously for the `CrossPointWebServerActivity` this would be overridden as `bool skipLoopDelay() const { return webServer && webServer->isRunning(); }`, same as `isWebServerRunning`'s current implementation.

Then here we can just check `currentActivity && currentActivity->skipLoopDelay()` and remove all tracking of `webServerActivity`

olearycrew: Just pushed this fix

olearycrew:
My only question remaining is - if we want to do the AP mode next, will it be just under "file transfer" and add an interstitial screen that says "connect to wifi" or "Activate AP" or something like that (need better naming) 

And then we'd either go down this same path or start a new path for the AP mode?

src/main.cpp
olearycrew: @daveallie take another look :)

src/activities/settings/SettingsActivity.cpp
```diff
@@ -8,9 +8,9 @@
 // Define the static settings list
 
 const SettingInfo SettingsActivity::settingsList[settingsCount] = {
-    {"White Sleep Screen", &CrossPointSettings::whiteSleepScreen},
-    {"Extra Paragraph Spacing", &CrossPointSettings::extraParagraphSpacing},
-    {"Short Power Button Click", &CrossPointSettings::shortPwrBtn}};
+    {"White Sleep Screen", SettingType::TOGGLE, &CrossPointSettings::whiteSleepScreen},
+    {"Extra Paragraph Spacing", SettingType::TOGGLE, &CrossPointSettings::extraParagraphSpacing}};
```
daveallie:
```suggestion
    {"Extra Paragraph Spacing", SettingType::TOGGLE, &CrossPointSettings::extraParagraphSpacing}},
```

daveallie:
> My only question remaining is - if we want to do the AP mode next, will it be just under "file transfer" and add an interstitial screen that says "connect to wifi" or "Activate AP" or something like that (need better naming)

Yeah I think that's reasonable, I also think that AP mode isn't a super high priority once we get all these changes in, but it may be a nice QOL thing. @meta-boy (from https://github.com/meta-boy/hojo) has mentioned that he may look into integrating with CrossPoint, so that may be another driver for it.

src/activities/settings/SettingsActivity.cpp
```diff
@@ -8,9 +8,9 @@
 // Define the static settings list
 
 const SettingInfo SettingsActivity::settingsList[settingsCount] = {
-    {"White Sleep Screen", &CrossPointSettings::whiteSleepScreen},
-    {"Extra Paragraph Spacing", &CrossPointSettings::extraParagraphSpacing},
-    {"Short Power Button Click", &CrossPointSettings::shortPwrBtn}};
+    {"White Sleep Screen", SettingType::TOGGLE, &CrossPointSettings::whiteSleepScreen},
+    {"Extra Paragraph Spacing", SettingType::TOGGLE, &CrossPointSettings::extraParagraphSpacing}},
```
daveallie: Sorry missed the bracket in the suggestion before

```suggestion
    {"Extra Paragraph Spacing", SettingType::TOGGLE, &CrossPointSettings::extraParagraphSpacing},
```

daveallie (approved):
Huge effort, thanks so much @olearycrew!

olearycrew:
Thanks for this project!

I'm excited to contribute - but maybe a smaller one next time 😜

PR #49, opened 2025-12-17
Feature/sleep screen

jonasdiemer:
A first attempt (mostly vibe coded) at bmp-based sleep screen.

- BMP are parsed 1bpp
- Parsing of 24bpp is prepared, but currently fails
  - 24bpp currently are mapped to black and white (we still need a drawing function for grayscale images)
- Code is likely quite sub-optimal, but sufficiently fast - 60ms for 1bpp


jonasdiemer:
Replaced by #52

PR #52, opened 2025-12-17
Added BMP sleep screen using /sleep.bmp (1bpp) from SD card.

jonasdiemer:
## Summary

Adresses #14. 
Supports 1bpp BMP only.


daveallie:
I have a slightly different implementation which I'd like to throw up which builds on @treetrum's work and is a bit more versatile: supports 1, 2, 8, 24, 32 bpp BMPs and also scales down too large images on the sleep screen. Will throw it up would love to get your eyes on it.

daveallie:
See https://github.com/daveallie/crosspoint-reader/pull/57

PR #38, opened 2025-12-16
Calculate the progress in the book by file sizes of each chapter.

jonasdiemer:
## Summary

Addresses #35.

Maybe it could be wise to do some caching of the spine sizes (but performance isn't too bad).

PR #54, opened 2025-12-18
Caching of spine item sizes for faster book loading (saves 1-4 seconds).

jonasdiemer:
As discussed in https://github.com/daveallie/crosspoint-reader/pull/38#issuecomment-3665142427, #38


daveallie:
Looks good, just needs a format

PR #56, opened 2025-12-18
Option to short-press power button.

jonasdiemer:
Adresses #53 

Please check if we still need the code to "Give the user up to 1000ms to start holding the power button, and must hold for SETTINGS.getPowerButtonDuration()" - the power button should be pressed already when waking up...

Also, decided to return before the delay to wait more to make the behavior more immediate.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -159,7 +159,12 @@ void setup() {
   SPI.begin(EPD_SCLK, SD_SPI_MISO, EPD_MOSI, EPD_CS);
 
   // SD Card Initialization
-  SD.begin(SD_SPI_CS, SPI, SPI_FQ);
+  if (!SD.begin(SD_SPI_CS, SPI, SPI_FQ)) {
+    Serial.printf("[%lu] [   ] SD card initialization failed\n", millis());
+    exitActivity();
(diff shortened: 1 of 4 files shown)


src/main.cpp
```diff
@@ -90,12 +86,16 @@ void verifyWakeupLongPress() {
     inputManager.update();
   }
```
daveallie: You were asking about it in your PR body, but this code is useful in validating that the source of the power cycle is the power switch and not something else like the reset button or some other wake from deep sleep event.

src/main.cpp
```diff
@@ -156,7 +156,7 @@ void setup() {
   Serial.printf("[%lu] [   ] Starting CrossPoint version " CROSSPOINT_VERSION "\n", millis());
 
   inputManager.begin();
-  verifyWakeupLongPress();
+  // verifyWakeupLongPress();
```
daveallie: Did you mean to comment this line out?

jonasdiemer: No, that was for testing - needs work (because with it un-commented, the wake up is too slow).

jonasdiemer:
Updated now. Had to reorder the boot sequence to load settings before checking the duration of the pwr button press.

daveallie:
just a conflict to deal with but looks good

PR #57, opened 2025-12-18
Custom sleep screen support with BMP reading

daveallie:
## Summary

* Builds on top of https://github.com/daveallie/crosspoint-reader/pull/16 - adresses https://github.com/daveallie/crosspoint-reader/discussions/14
* This PR adds the ability for the user to supply a custom `sleep.bmp` image at the root of the SD card that will be shown instead of the default sleep screen if present.
* Supports:
  * Different BPPs:
    * 1bit
    * 2bit
    * 8bit
    * 24bit
    * 32bit (with alpha-channel ignored)
  * Grayscale rendering

### DIFF
--- a/lib/GfxRenderer/Bitmap.cpp
+++ b/lib/GfxRenderer/Bitmap.cpp
@@ -99,18 +99,15 @@ BmpReaderError Bitmap::parseHeaders() {
 
   if (width <= 0 || height <= 0) return BmpReaderError::BadDimensions;
 
-  // Palette for 8-bit indexed images
-  for (int i = 0; i < 256; i++) paletteLum[i] = static_cast<uint8_t>(i);  // default grayscale ramp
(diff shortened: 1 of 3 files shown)


jonasdiemer:
Tested, working nicely with 1bpp and 24bpp. I couldn't test 2bpp because ImageMagick wouldn't produce a 2bpp file. The 8bpp file ImageMagic produced yielded: `[17819] Error parsing sleep.bmp: UnsupportedCompression (expected BI_RGB or BI_BITFIELDS for 32bpp)[17820]   Image drawn to frame buffer` (there seems to also be a missing line break).

GIMP also is no better, so I don't really have a way to generate 2bpp.

daveallie:
I generated a 2bpp image as part of https://github.com/daveallie/crosspoint-reader/pull/23, renders well

PR #70, opened 2025-12-19
Fix: restores cyrillic glyphs to Pixel Arial font

osteotek:
## Summary

* adds cyrillic glyphs to pixel arial font, used as Small font in UI

## Additional Context

* with recent changes pixel arial font lost cyrillic glyphs


daveallie:
Rerunning the `fontconvert.py` script on your branch reverts `pixelarial14.h` back to the state on master. Was there a change to the generator script you meant to include here?

osteotek:
@daveallie Hmm, I’ll check. Perhaps we’re using different fonts as sources.

jlaunay:
This also fixes the problem of Latin characters missing, like `é è` etc.

daveallie:
> @daveallie Hmm, I’ll check. Perhaps we’re using different fonts as sources.

I didn't even think of that 🤦‍♂️. If you have it on hand and it's available for use without a license, then I think we should commit it to the repo. We should do the same for the other fonts too, will need to swap out Bookerly for [Literata](https://fonts.google.com/specimen/Literata) as part of doing this.

daveallie:
Going to merge this in to restore the functionality, but would love to standardize the font generation with committed font files

osteotek:
@daveallie thanks, I'll try to find a small font with a suitable license. Pixel Arial only has latin characters

PR #71, opened 2025-12-19
Randomly load Sleep Screen from /sleep/*bmp (if exists).

jonasdiemer:
Load a random sleep screen. 

Only works for ".bmp" (not ".BMP"), but I think that's OK (we do the same for EPUBs).

### DIFF
--- a/src/activities/boot_sleep/SleepActivity.cpp
+++ b/src/activities/boot_sleep/SleepActivity.cpp
@@ -79,11 +79,10 @@ void SleepActivity::onEnter() {
 void SleepActivity::renderPopup(const char* message) const {
   const int textWidth = renderer.getTextWidth(READER_FONT_ID, message);
   constexpr int margin = 20;
-  // Round all coordinates to 8 pixel boundaries
-  const int x = ((GfxRenderer::getScreenWidth() - textWidth - margin * 2) / 2 + 7) / 8 * 8;
+  const int x = (GfxRenderer::getScreenWidth() - textWidth - margin * 2) / 2;
   constexpr int y = 117;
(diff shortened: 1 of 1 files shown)


src/activities/boot_sleep/SleepActivity.cpp
```diff
@@ -22,6 +76,21 @@ void SleepActivity::onEnter() {
...
 }
 
+void SleepActivity::renderPopup(const char* message) const {
+  const int textWidth = renderer.getTextWidth(READER_FONT_ID, message);
+  constexpr int margin = 20;
+  // Round all coordinates to 8 pixel boundaries
+  const int x = ((GfxRenderer::getScreenWidth() - textWidth - margin * 2) / 2 + 7) / 8 * 8;
+  constexpr int y = 117;
+  const int w = (textWidth + margin * 2 + 7) / 8 * 8;
+  const int h = (renderer.getLineHeight(READER_FONT_ID) + margin * 2 + 7) / 8 * 8;
```
daveallie: We don't actually need to round the coords to 8 pixels anymore, it was only necessary for windowed updates but as you've shown, we can just use the using display buffer without issue, feel free to remove the `+ 7) / 8 * 8)` on these line. Happy to merge as is if you want as well, I'm going to touch up the other one shortly.

daveallie (approved):

src/activities/boot_sleep/SleepActivity.cpp
jonasdiemer: Tx, updated.

daveallie (approved):

PR #77, opened 2025-12-20
Rotation Support

ChandhokTannay:
•  What is the goal of this PR?  
  Implement a horizontal EPUB reading mode so books can be read in landscape orientation (both 90° and 270°), while keeping the rest of the UI in portrait.

•  What changes are included?
◦  Rendering / Display
▪  Added an orientation model to GfxRenderer (Portrait, LandscapeNormal, LandscapeFlipped) and made:
▪  drawPixel, drawImage, displayWindow map logical coordinates differently depending on orientation.
▪  getScreenWidth() / getScreenHeight() return orientation‑aware logical dimensions (480×800 in portrait, 800×480 in landscape).
◦  Settings / Configuration
▪  Extended CrossPointSettings with:
▪  landscapeReading (toggle for portrait vs. landscape EPUB reading).
▪  landscapeFlipped (toggle to flip landscape 180° so both horizontal holding directions are supported).
▪  Updated settings serialization/deserialization to persist these fields while remaining backward‑compatible with existing settings files.
▪  Updated SettingsActivity to expose two new toggles:
▪  “Landscape Reading”
▪  “Flip Landscape (swap top/bottom)”
◦  EPUB Reader
▪  In EpubReaderActivity:
▪  On onEnter, set GfxRenderer orientation based on the new settings (Portrait, LandscapeNormal, or LandscapeFlipped).
▪  On onExit, reset orientation back to Portrait so Home, WiFi, Settings, etc. continue to render as before.
▪  Adjusted renderStatusBar to position the status bar and battery indicator relative to GfxRenderer::getScreenHeight() instead of hard‑coded Y coordinates, so it stays correctly at the bottom in both portrait and landscape.
◦  EPUB Caching / Layout
▪  Extended Section cache metadata (section.bin) to include the logical screenWidth and screenHeight used when pages were generated; bumped SECTION_FILE_VERSION.
▪  Updated loadCacheMetadata to compare:
▪  font/margins/line compression/extraParagraphSpacing and screen dimensions; mismatches now invalidate and clear the cache.
▪  Updated persistPageDataToSD and all call sites in EpubReaderActivity to pass the current GfxRenderer::getScreenWidth() / getScreenHeight() so portrait and landscape caches are kept separate and correctly sized.

Additional Context

•  Cache behavior / migration
◦  Existing section.bin files (old SECTION_FILE_VERSION) will be detected as incompatible and their caches cleared and rebuilt once per chapter when first opened after this change.
◦  Within a given orientation, caches will be reused as before. Switching orientation (portrait ↔ landscape) will cause a one‑time re‑index of each chapter in the new orientation.
•  Scope and risks
◦  Orientation changes are scoped to the EPUB reader; the Home screen, Settings, WiFi selection, sleep screens, and web server UI continue to assume portrait orientation.
◦  The renderer’s orientation is a static/global setting; if future code uses GfxRenderer outside the reader while a reader instance is active, it should be aware that orientation is no longer implicitly fixed.
◦  All drawing primitives now go through orientation‑aware coordinate transforms; any code that previously relied on edge‑case behavior or out‑of‑bounds writes might surface as logged “Outside range” warnings instead.
•  Testing suggestions / areas to focus on
◦  Verify in hardware:
▪  Portrait mode still renders correctly (boot, home, settings, WiFi, reader).
▪  Landscape reading in both directions:
▪  Landscape Reading = ON, Flip Landscape = OFF.
▪  Landscape Reading = ON, Flip Landscape = ON.
▪  Status bar (page X/Y, % progress, battery icon) is fully visible and aligned at the bottom in all three combinations.
◦  Open the same book:
▪  In portrait first, then switch to landscape and reopen it.
▪  Confirm that:
▪  Old portrait caches are rebuilt once for landscape (you should see the “Indexing…” page).
▪  Progress save/restore still works (resume opens to the correct page in the current orientation).
◦  Ensure grayscale rendering (the secondary pass in EpubReaderActivity::renderContents) still looks correct in both orientations.

### DIFF
--- a/lib/GfxRenderer/GfxRenderer.h
+++ b/lib/GfxRenderer/GfxRenderer.h
@@ -26,11 +26,9 @@ class GfxRenderer {
   static_assert(BW_BUFFER_CHUNK_SIZE * BW_BUFFER_NUM_CHUNKS == EInkDisplay::BUFFER_SIZE,
                 "BW buffer chunking does not line up with display buffer size");
 
-  // Orientation used for all rendering operations
-  Orientation orientation;
-
   EInkDisplay& einkDisplay;
   RenderMode renderMode;
+  Orientation orientation;
   uint8_t* bwBufferChunks[BW_BUFFER_NUM_CHUNKS] = {nullptr};
   std::map<int, EpdFontFamily> fontMap;
   void renderChar(const EpdFontFamily& fontFamily, uint32_t cp, int* x, const int* y, bool pixelState,
@@ -39,7 +37,7 @@ class GfxRenderer {
   void rotateCoordinates(int x, int y, int* rotatedX, int* rotatedY) const;
 
  public:
-  explicit GfxRenderer(EInkDisplay& einkDisplay) : einkDisplay(einkDisplay), renderMode(BW) {}
+  explicit GfxRenderer(EInkDisplay& einkDisplay) : einkDisplay(einkDisplay), renderMode(BW), orientation(Portrait) {}
   ~GfxRenderer() = default;
 
   static constexpr int VIEWABLE_MARGIN_TOP = 9;


lib/Epub/Epub/Section.cpp
```diff
@@ -69,17 +72,21 @@ bool Section::loadCacheMetadata(const int fontId, const float lineCompression, c
...
+    int fileScreenWidth, fileScreenHeight;
     serialization::readPod(inputFile, fileFontId);
     serialization::readPod(inputFile, fileLineCompression);
     serialization::readPod(inputFile, fileMarginTop);
     serialization::readPod(inputFile, fileMarginRight);
     serialization::readPod(inputFile, fileMarginBottom);
     serialization::readPod(inputFile, fileMarginLeft);
     serialization::readPod(inputFile, fileExtraParagraphSpacing);
+    serialization::readPod(inputFile, fileScreenWidth);
+    serialization::readPod(inputFile, fileScreenHeight);
```
daveallie: I think we could reasonably just cache on usable width and usable height, and then apply the top and left margin as part of rendering pages, this would help lower the count of cache busting variables we have here

lib/GfxRenderer/GfxRenderer.h
```diff
@@ -12,12 +12,22 @@ class GfxRenderer {
...
+  };
+
  private:
   static constexpr size_t BW_BUFFER_CHUNK_SIZE = 8000;  // 8KB chunks to allow for non-contiguous memory
   static constexpr size_t BW_BUFFER_NUM_CHUNKS = EInkDisplay::BUFFER_SIZE / BW_BUFFER_CHUNK_SIZE;
   static_assert(BW_BUFFER_CHUNK_SIZE * BW_BUFFER_NUM_CHUNKS == EInkDisplay::BUFFER_SIZE,
                 "BW buffer chunking does not line up with display buffer size");
 
+  // Global orientation used for all rendering operations
+  static Orientation orientation;
```
daveallie: I don't think this needs be a static variable, we pass the same instance around the whole firmware by reference, leaving it as an instance variable would be preferrable

lib/GfxRenderer/GfxRenderer.cpp
```diff
@@ -204,22 +225,55 @@ void GfxRenderer::displayBuffer(const EInkDisplay::RefreshMode refreshMode) cons
...
+    case Orientation::Portrait:
+      // 480px wide in portrait logical coordinates
+      return EInkDisplay::DISPLAY_HEIGHT;
+    case Orientation::LandscapeNormal:
+    case Orientation::LandscapeFlipped:
+      // 800px wide in landscape logical coordinates
+      return EInkDisplay::DISPLAY_WIDTH;
+  }
+  return EInkDisplay::DISPLAY_HEIGHT;
 }
```
daveallie, line 254: these functions should also move to instance functions to allow accessing the `orientation` instance variable

lib/GfxRenderer/GfxRenderer.h
```diff
@@ -12,12 +12,22 @@ class GfxRenderer {
  public:
   enum RenderMode { BW, GRAYSCALE_LSB, GRAYSCALE_MSB };
 
+  // Logical screen orientation from the perspective of callers
+  enum class Orientation {
+    Portrait,         // 480x800 logical coordinates (current default)
+    LandscapeNormal,  // 800x480 logical coordinates, native panel orientation
+    LandscapeFlipped  // 800x480 logical coordinates, rotated 180° (swap top/bottom)
+  };
```
daveallie: Can we please add `PortraitInverted`. Also, not super in love with `LandscapeNormal` and `LandscapeFlipped` it's not going to be obvious what "normal" is to most callers. I think we could use clearer names like `LandscapeClockwise` (or `LandscapeCW`) for when you rotate the device clockwise 90 degrees from the default portrait orientation, and `LandscaleCounterClockwise` (`LandscapeCCW`) for when you rotate the device counter-clockwise 90 degrees

src/activities/settings/SettingsActivity.cpp
```diff
@@ -1,29 +1,25 @@
...
-    {"Extra Paragraph Spacing", SettingType::TOGGLE, &CrossPointSettings::extraParagraphSpacing, {}},
-    {"Short Power Button Click", SettingType::TOGGLE, &CrossPointSettings::shortPwrBtn, {}}};
-}  // namespace
+
+const SettingInfo SettingsActivity::settingsList[settingsCount] = {
+    {"White Sleep Screen", SettingType::TOGGLE, &CrossPointSettings::whiteSleepScreen},
+    {"Extra Paragraph Spacing", SettingType::TOGGLE, &CrossPointSettings::extraParagraphSpacing},
+    {"Short Power Button Click", SettingType::TOGGLE, &CrossPointSettings::shortPwrBtn},
+    {"Landscape Reading", SettingType::TOGGLE, &CrossPointSettings::landscapeReading},
+    {"Flip Landscape (swap top/bottom)", SettingType::TOGGLE, &CrossPointSettings::landscapeFlipped}};
```
daveallie: These can be combined into an enum setting, there's an example of this on the `master` branch with the updated sleep screen setting.

daveallie:
Hi @ChandhokTannay I've cleaned up the things I mentioned in the PR review, will get it merged

PR #86, opened 2025-12-21
Paginate book list and avoid out of bounds rendering

daveallie:
## Summary

* Paginate book list
* Avoid out of bounds rendering of long book titles, truncate with ellipsis instead

## Additional Context

* Should partially help with https://github.com/daveallie/crosspoint-reader/issues/75 as it was previously rendering a lot of content off screen, will need to test with a large directory


daveallie:
Tested with ~250 items, took about 45 seconds to open the directory. Same directory takes 88 seconds on `master`. All that time save is just from preventing renders outside of the screen boundary.

PR #89, opened 2025-12-21
Book cover sleep screen

daveallie:
## Summary

* Fix issue with 2-bit bmp rendering
* Add support generate book cover BMP from JPG and use as sleep screen

## Additional Context

* It does not support other image formats beyond JPG at this point
* Something is cooked with my JpegToBmpConverter logic, it generates weird interlaced looking images for some JPGs

| Book 1 | Book 2|
| --- | --- |
| IMG_5653 | IMG_5652 |

### DIFF
--- a/USER_GUIDE.md
+++ b/USER_GUIDE.md
@@ -59,7 +59,11 @@ See the [webserver docs](./docs/webserver.md) for more information on how to con
 ### 3.5 Settings
 
 The Settings screen allows you to configure the device's behavior. There are a few settings you can adjust:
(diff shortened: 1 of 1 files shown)


daveallie:
Going to merge this one even if there are some issues with the JPG -> BMP converter, keen to offer some partial support for this feature as users can set their own preferences in settings if it's not working.

jlaunay:
I tried with about 25 books, but I couldn't find any that worked; each time I get a weird interlaced image.

daveallie:
Thanks @jlaunay I'll track that in #118

Issue #99, opened 2025-12-22
Unable to go to home screen or put device to sleep


dangson:
### Affected Version

0.8.1

### Describe the Bug

I just updated to 0.8.1 and now I'm unable to back out of a book or put device to sleep

### Steps to Reproduce

Currently reading Brandon Sanderson's Oathbringer which is a pretty long book in case that matters. The EPUB is 3.7 MB.

1. While reading a book, press the back button. Nothing happens.
2. Now all buttons except power become unresponsive. 
3. Holding power shows the "booting" screen but the device never goes to sleep. It just goes back to the book and buttons work again.

Other than this issue, this firmware is great! I almost chucked the device into a drawer since text rendering on the original firmware is so terrible, but I love reading on it with this firmware. Thank you! Happy to provide any additional info to help debug.

### Expected Behavior

Pressing back should take me back to the menu.

Holding power button should put the device to sleep.

### Relevant Log Output/Screenshots

```shell
https://share.cleanshot.com/6hmT87XJ
```

daveallie:
Thanks for sharing this @dangson, it's a very similar problem as to what is now being experienced over on https://github.com/daveallie/crosspoint-reader/issues/60 (funnily enough also another Stormlight Archive book). I believe I have a fix for this (hopefully should be available in a release within the next day or two).

What's actually happening here is a crash of the firmware, so while it looks like it may still be on and functioning, it has completely gone back to the off state. e-paper just makes it hard to communicate that 😅 (just in case you're worried about the battery draining or something).

dangson:
@daveallie thanks for the quick reply! These huge Stormlight Archive books are good stress tests of e-reader firmware 😂 

I'll be sure to give the next version a try. For now downgrading to 0.7.0 works.

daveallie:
Should be fixed in 0.9.0, going to close this preemptively, but if it's still happening, then please re-open this issue.

dangson:
Confirming this is fixed on 0.9.0. Thanks!

Issue #10, opened 2025-12-13
Large TOC causes crash


daveallie:
As an example, the German Luther Bible 1912 EPUB has 1259 `<navPoint />` elements in the TOC file. We're currently attempting to load them all as `EpubTocEntry` records into memory, which is currently causing a crash:

```
[8866] [EBP] Loading ePub: /bible_luther1912.epub
[8933] [ZIP] Decompressed 176 bytes into 254 bytes
[8934] [EBP] Found content.opf at: OEBPS/content.opf
[9033] [ZIP] Decompressed 1872 bytes into 15803 bytes
[9035] [EBP] Successfully parsed content.opf

abort() was called at PC 0x4203a269 on core 0
  #0  0x4203a269 in __cxa_call_unexpected at /builds/idf/crosstool-NG/.build/HOST-x86_64-apple-darwin12/riscv32-esp-elf/src/gcc/libstdc++-v3/libsupc++/eh_personality.cc:775

Core  0 register dump:
MEPC    : 0x4038236a  RA      : 0x40387172  SP      : 0x3fcb4d10  GP      : 0x3fc8e000
  #0  0x4038236a in panic_abort at /home/runner/work/esp32-arduino-lib-builder/esp32-arduino-lib-builder/esp-idf/components/esp_system/panic.c:408

TP      : 0x3fc617a0  T0      : 0x37363534  T1      : 0x7271706f  T2      : 0x33323130
S0/FP   : 0x3fcb4d3c  S1      : 0x3fcb4d3c  A0      : 0x3fcb4d48  A1      : 0x3fcb4d2a
A2      : 0x00000000  A3      : 0x3fcb4d75  A4      : 0x00000001  A5      : 0x3fcaa000
A6      : 0x7a797877  A7      : 0x76757473  S2      : 0x3fcd09d0  S3      : 0x3fcb54f0
S4      : 0x00004c00  S5      : 0x3fcd55d0  S6      : 0x3fcd55d0  S7      : 0x3fcb1a04
S8      : 0x00000000  S9      : 0x3fcb2388  S10     : 0x3fcc77d8  S11     : 0x00000004
T3      : 0x6e6d6c6b  T4      : 0x6a696867  T5      : 0x66656463  T6      : 0x62613938
MSTATUS : 0x00001801  MTVEC   : 0x40380001  MCAUSE  : 0x00000007  MTVAL   : 0x00000000
MHARTID : 0x00000000

Stack memory:
[...]

PR #104, opened 2025-12-22
New book.bin spine and table of contents cache

daveallie:
## Summary

* Use single unified cache file for book spine, table of contents, and core metadata (title, author, cover image)
* Use new temp item store file in OPF parsing to store items to be rescaned when parsing spine
  * This avoids us holding these items in memory
* Use new toc.bin.tmp and spine.bin.tmp to build out partial toc / spine data as part of parsing content.opf and the NCX file
  * These files are re-read multiple times to ultimately build book.bin

## Additional Context

* Spec for file format included below as an image
* This should help with:
  * #10 
  * #60 
  * #99

### DIFF
--- a/lib/Epub/Epub/BookMetadataCache.cpp
+++ b/lib/Epub/Epub/BookMetadataCache.cpp
@@ -260,6 +260,10 @@ bool BookMetadataCache::load() {
   serialization::readPod(bookFile, spineCount);
   serialization::readPod(bookFile, tocCount);
 
+  serialization::readString(bookFile, coreMetadata.title);
+  serialization::readString(bookFile, coreMetadata.author);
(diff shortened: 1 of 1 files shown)


daveallie:
Going to rebuild the parsing to end up with something like this:
image

The current state we hold in memory when parsing content.opf isn't great. Due to the sequential nature of expat's parsing, we'll have to hold a temp file with items to map back to the itemrefs

image

It'll be easiest to build out temporary spine and toc bin files before combining them
image

Joining the TOC and Spine items by href is a bit of a pain, to avoid loading them all into memory, we can use an N^2 approach (for each spine as we go to place it in book.bin, scan every TOC file to find a match (and vice versa)).

PR #108, opened 2025-12-22
Grid Browser with file thumbnails

CaptainFrito:
IMG_7326 Medium
IMG_7327 Medium

- UI design by Discord user gan_shu
- Drop in replacement for the existing FileSelectionActivity, with the same features and key bindings
- Currently only renders files with extension .thumb.bmp - meant as a proof of concept to decide if we want to pursue this further considering the limitations

### DIFF
--- a/src/activities/home/GridBrowserActivity.cpp
+++ b/src/activities/home/GridBrowserActivity.cpp
@@ -18,8 +18,8 @@ constexpr int TILE_PADDING = 5;
 constexpr int THUMB_W = 90;
 constexpr int THUMB_H = 120;
 constexpr int TILE_TEXT_H = 60;
-constexpr int gridLeftOffset = 37;
-constexpr int gridTopOffset = 125;
+constexpr int GRID_OFFSET_LEFT = 37;
+constexpr int GRID_OFFSET_TOP = 125;
 }  // namespace
 
 inline int min(const int a, const int b) { return a < b ? a : b; }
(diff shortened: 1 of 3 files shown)


CaptainFrito:
PR updated and ready for review/merge

IMG_7398 Large

- Works just like FileSelectionActivity, with the same features and key bindings. "UI Theme" setting to switch between the two
- Lazily extracts cover art from epubs while browsing, converts them to dithered 1-bit BMP of size 90x120px
- New GfxRenderer methods for drawing greyscale rects (using dithering), rounded corners, centered multiline text box

src/activities/home/GridBrowserActivity.cpp
```diff
@@ -0,0 +1,309 @@
...
+    update(false);
+    renderer.displayBuffer();
+  }
+} 
+
+void GridBrowserActivity::drawSelectionRectangle(int tileIndex, bool black) const {
+  renderer.drawRoundedRect(GRID_OFFSET_LEFT + tileIndex % 3 * TILE_W, GRID_OFFSET_TOP + tileIndex / 3 * TILE_H, TILE_W, TILE_H, 2, 5, black);
+}
+
+void GridBrowserActivity::update(bool render) const {
```
daveallie, line 302: nit: This parameter is unused

src/activities/home/GridBrowserActivity.cpp
```diff
@@ -0,0 +1,309 @@
...
+  // Redraw only changed tiles
+  // renderer.clearScreen();
+  if (previousSelectorIndex >= 0) {
+    drawSelectionRectangle(previousSelectorIndex, false);
+  }
+  drawSelectionRectangle(selectorIndex, true);
```
daveallie, line 308: Double tapping a button to go from selection `n` to `n+2` can mean that by the time this runs `previousSelectorIndex` can be `n+1` and `selectorIndex` can be `n+2`. This results in an issue where the item at `n` still has a selector indicator around it and it's not cleared until you go back over it. To fix this, you can wrap any code in loop which adjusts variables used by the rendering loop e.g. `selectorIndex` and `previousSelectorIndex` in the `renderingMutex`.

For this same reason, navigating too quickly into a folder after pressing left and right result in the screen not updating correctly.

src/activities/home/GridBrowserActivity.cpp
```diff
@@ -0,0 +1,309 @@
...
+        File bmpFile = SD.open(file.thumbPath.c_str());
+        if (bmpFile) {
+          Bitmap bitmap(bmpFile);
+          if (bitmap.parseHeaders() == BmpReaderError::Ok) {
+            constexpr int thumbOffsetX = (TILE_W - THUMB_W) / 2;
+            constexpr int thumbOffsetY = (TILE_H - TILE_TEXT_H - THUMB_H) / 2;
+            renderer.drawBitmap(bitmap, tileX + thumbOffsetX, tileY + thumbOffsetY, THUMB_W, THUMB_H);
+          }
+        }
+      }
```
daveallie, line 288: Would be nice to have a fallback icon for books where the cover can't be generated

lib/JpegToBmpConverter/JpegToBmpConverter.cpp
```diff
@@ -316,80 +316,91 @@ inline void write32Signed(Print& out, const int32_t value) {
...
-  write32(bmpOut, 2835);  // yPixelsPerMeter (72 DPI)
-  write32(bmpOut, 256);   // colorsUsed
-  write32(bmpOut, 256);   // colorsImportant
+int getBytesPerRow(int width, int bitsPerPixel) {
+  if (bitsPerPixel == 8) {
+    return (width + 3) / 4 * 4;  // 8 bits per pixel, padded
+  } else if (bitsPerPixel == 2) {
+    return (width * 2 + 31) / 32 * 4;  // 2 bits per pixel, round up
+  }
+  return (width + 31) / 32 * 4;  // 1 bit per pixel, round up
```
daveallie, line 332: nit: can this be simplified to `return (width * bitsPerPixel + 31) / 32 * 4;`?

daveallie (changes requested):
Overall some comments on the code, but probably more critically is that when entering a directory, I don't seem to get any of the laziness you were mentioning in the PR body, the whole device seems to lock up as the thumbnails are rendered. This won't be mergable if we can't fix this / do thumbnail generation at another point, waiting for 9 books to generate thumbnails is a bit slow

icannotttt:
I have a suggest. If it's difficult to render  3\*3 images, how about a 2\*2 one? We could use it to display the 4 most recently read books, which should work much better? Maybe we don't need to render all book.

stiivgiaco:
will this be the only option on how to display the library or will it be based on user preference; ex list or grid?

daveallie:
> will this be the only option on how to display the library or will it be based on user preference; ex list or grid?

The current implementation adds a setting to toggle between the existing setup (list) and this one (grid).

CaptainFrito:
Closing this PR now, see https://github.com/crosspoint-reader/crosspoint-reader/pull/528 for a more up to date implementation of UI themes. The thumbnail grid can be revisited later on.

PR #111, opened 2025-12-23
Add setting to enable status bar display options

1991AcuraLegend:
Add setting toggle that allows status bar display options in EpubReader. 

Supported options would be as follows: 

- FULL: display as is today
- PROGRESS: display progress bar only
- BATTERY: display battery only
- NONE: hide status bar

### DIFF
--- a/src/activities/reader/EpubReaderActivity.cpp
+++ b/src/activities/reader/EpubReaderActivity.cpp
@@ -342,16 +342,15 @@ void EpubReaderActivity::renderContents(std::unique_ptr<Page> page) {
 void EpubReaderActivity::renderStatusBar() const {
   // determine visible status bar elements
   const bool showProgress = SETTINGS.statusBar == CrossPointSettings::STATUS_BAR_MODE::FULL;
-  const bool showBattery = SETTINGS.statusBar == CrossPointSettings::STATUS_BAR_MODE::NO_PROGRESS || SETTINGS.statusBar == CrossPointSettings::STATUS_BAR_MODE::FULL;
-  const bool showChapterTitle = SETTINGS.statusBar == CrossPointSettings::STATUS_BAR_MODE::NO_PROGRESS || SETTINGS.statusBar == CrossPointSettings::STATUS_BAR_MODE::FULL;
+  const bool showBattery = SETTINGS.statusBar == CrossPointSettings::STATUS_BAR_MODE::NO_PROGRESS ||
+                           SETTINGS.statusBar == CrossPointSettings::STATUS_BAR_MODE::FULL;
+  const bool showChapterTitle = SETTINGS.statusBar == CrossPointSettings::STATUS_BAR_MODE::NO_PROGRESS ||
+                                SETTINGS.statusBar == CrossPointSettings::STATUS_BAR_MODE::FULL;
 
   // height variable shared by all elements
   constexpr auto textY = 776;
-
-  // Left aligned battery icon and percentage
-  const uint16_t percentage = battery.readPercentage();
-  const auto percentageText = std::to_string(percentage) + "%";
-  const auto percentageTextWidth = renderer.getTextWidth(SMALL_FONT_ID, percentageText.c_str());
+  int percentageTextWidth = 0;
+  int progressTextWidth = 0;
 
   if (showProgress) {
     // Calculate progress in book
@@ -360,13 +359,17 @@ void EpubReaderActivity::renderStatusBar() const {
 
     // Right aligned text for progress counter
     const std::string progress = std::to_string(section->currentPage + 1) + "/" + std::to_string(section->pageCount) +
-                                "  " + std::to_string(bookProgress) + "%";
-    const auto progressTextWidth = renderer.getTextWidth(SMALL_FONT_ID, progress.c_str());
+                                 "  " + std::to_string(bookProgress) + "%";
+    progressTextWidth = renderer.getTextWidth(SMALL_FONT_ID, progress.c_str());
     renderer.drawText(SMALL_FONT_ID, GfxRenderer::getScreenWidth() - marginRight - progressTextWidth, textY,
                       progress.c_str());
   }
-  
+
   if (showBattery) {
+    // Left aligned battery icon and percentage
+    const uint16_t percentage = battery.readPercentage();
+    const auto percentageText = std::to_string(percentage) + "%";
+    percentageTextWidth = renderer.getTextWidth(SMALL_FONT_ID, percentageText.c_str());
     renderer.drawText(SMALL_FONT_ID, 20 + marginLeft, textY, percentageText.c_str());
 
     // 1 column on left, 2 columns on right, 5 columns of battery body
@@ -393,9 +396,7 @@ void EpubReaderActivity::renderStatusBar() const {
       filledWidth = batteryWidth - 5;  // Ensure we don't overflow
     }
     renderer.fillRect(x + 1, y + 1, filledWidth, batteryHeight - 2);
-
   }
-
 
   if (showChapterTitle) {
     // Centered chatper title text
--- a/src/activities/settings/SettingsActivity.cpp
+++ b/src/activities/settings/SettingsActivity.cpp
@@ -9,7 +9,7 @@
 
 // Define the static settings list
 namespace {
-constexpr int settingsCount = 4;
+constexpr int settingsCount = 5;
 const SettingInfo settingsList[settingsCount] = {
     // Should match with SLEEP_SCREEN_MODE
     {"Sleep Screen", SettingType::ENUM, &CrossPointSettings::sleepScreen, {"Dark", "Light", "Custom", "Cover"}},


src/activities/reader/EpubReaderActivity.cpp
```diff
@@ -334,69 +332,77 @@ void EpubReaderActivity::renderContents(std::unique_ptr<Page> page) {
...
-  renderer.drawLine(x + batteryWidth - 4, y, x + batteryWidth - 4, y + batteryHeight - 1);
-  renderer.drawLine(x + batteryWidth - 3, y + 2, x + batteryWidth - 1, y + 2);
-  renderer.drawLine(x + batteryWidth - 3, y + batteryHeight - 3, x + batteryWidth - 1, y + batteryHeight - 3);
-  renderer.drawLine(x + batteryWidth - 1, y + 2, x + batteryWidth - 1, y + batteryHeight - 3);
-
-  // The +1 is to round up, so that we always fill at least one pixel
-  int filledWidth = percentage * (batteryWidth - 5) / 100 + 1;
-  if (filledWidth > batteryWidth - 5) {
-    filledWidth = batteryWidth - 5;  // Ensure we don't overflow
+  if (Settings.statusBar && CrossPointSettings::STATUS_BAR_MODE::PROGRESS) {
```
daveallie: I don't think these are conditional checks are working the way you expect, currently it's evaluating , it should be:

```suggestion
  if (SETTINGS.statusBar == CrossPointSettings::STATUS_BAR_MODE::PROGRESS) {
```

Similar to the ones below.


Also, I believe this block should be running when in full mode as well.

It might make a bit more sense to assign a few bools above this logic to help make parsing a bit better:

```cpp
const bool showProgress = SETTINGS.statusBar == CrossPointSettings::STATUS_BAR_MODE::PROGRESS || SETTINGS.statusBar == CrossPointSettings::STATUS_BAR_MODE::FULL;
const bool showBattery = SETTINGS.statusBar != CrossPointSettings::STATUS_BAR_MODE::NONE;
const bool showChapterTitle = SETTINGS.statusBar != CrossPointSettings::STATUS_BAR_MODE::NONE;
```

src/activities/reader/EpubReaderActivity.cpp
```diff
@@ -334,69 +332,77 @@ void EpubReaderActivity::renderContents(std::unique_ptr<Page> page) {
...
-    titleWidth = renderer.getTextWidth(SMALL_FONT_ID, title.c_str());
-    while (titleWidth > availableTextWidth && title.length() > 11) {
-      title.replace(title.length() - 8, 8, "...");
-      titleWidth = renderer.getTextWidth(SMALL_FONT_ID, title.c_str());
+  
+  if (Settings.statusBar && CrossPointSettings::STATUS_BAR_MODE::BATTERY) {
+    // Left aligned battery icon and percentage
+    const uint16_t percentage = battery.readPercentage();
+    const auto percentageText = std::to_string(percentage) + "%";
+    const auto percentageTextWidth = renderer.getTextWidth(SMALL_FONT_ID, percentageText.c_str());
```
daveallie: We'll need to move the instantiation out of the `if` block so it can be used in the full block.

src/CrossPointSettings.cpp
```diff
@@ -1,31 +1,30 @@
...
-  File outputFile;
-  if (!FsHelpers::openFileForWrite("CPS", SETTINGS_FILE, outputFile)) {
-    return false;
-  }
-
+  std::ofstream outputFile(SETTINGS_FILE);
   serialization::writePod(outputFile, SETTINGS_FILE_VERSION);
   serialization::writePod(outputFile, SETTINGS_COUNT);
   serialization::writePod(outputFile, sleepScreen);
+  serialization::writePod(outputFile, statusBar);
```
daveallie: Regardless of where this setting is in the UI, it should go at the end of the file, please move this down a few lines. All new settings should get appended to avoid breaking people's existing settings

src/CrossPointSettings.cpp
```diff
@@ -56,6 +57,8 @@ bool CrossPointSettings::loadFromFile() {
   do {
     serialization::readPod(inputFile, sleepScreen);
     if (++settingsRead >= fileSettingsCount) break;
+    serialization::readPod(inputFile, statusBar);
+    if (++settingsRead >= fileSettingsCount) break;
```
daveallie: Similar, move to be the last setting in the file

src/CrossPointSettings.h
```diff
@@ -18,8 +18,13 @@ class CrossPointSettings {
   // Should match with SettingsActivity text
   enum SLEEP_SCREEN_MODE { DARK = 0, LIGHT = 1, CUSTOM = 2, COVER = 3 };
 
+  // Status bar display type enum
+  enum STATUS_BAR_MODE { NONE = 0, BATTERY = 1, PROGRESS = 2, FULL = 3 };
```
daveallie: imo (v happy to hear yours), we should change to be:
- `NONE` - Blank
- `NO_PROGRESS` - Show the battery and the title
- `FULL` - Show everything

There doesn't seem to be a lot of value in showing the progress / battery without the chapter title?

src/activities/reader/EpubReaderActivity.cpp
1991AcuraLegend: The goal was for it to act as a bitflag. that space would be saved by having it just be one uint8 (1 byte) instead of 3 bools (3 bytes). Plus it makes it easier to add additional modes in the future (just expand the enum vs adding more settings)

src/CrossPointSettings.h
1991AcuraLegend: Since the bitflag affords more options, how about

-  `NONE` - blank, for people who want just the text
- `NO_PROGRESS` - show battery and title, for people who don't want progress to spoil how close to the end of the book they are
- `BATTERY` - show only battery, for people who just want to know if they need to charge
- `FULL` - show everything

Originally I was just thinking of it in terms of what my Kindle showed, but this probably makes more sense.

src/activities/reader/EpubReaderActivity.cpp
1991AcuraLegend: will fix this on next commit after the mode thread below is resolved

src/CrossPointSettings.cpp
1991AcuraLegend: will fix this on next commit after the mode thread below is resolved

src/CrossPointSettings.cpp
1991AcuraLegend: will fix this on next commit after the mode thread below is resolved

src/activities/reader/EpubReaderActivity.cpp
daveallie: It would still be an enum (there would only be a single `SETTINGS.statusBar` which would be stored as a `uint8_t`), but the value can be 0-255, I'm suggesting that the enum values are set based on the actual options available to the user:
- `NONE` = `0` = Show nothing
- `NO_PROGRESS` = `1` = Show the battery and the title
- `FULL` = `2`  = Show the battery, title, and progress

I messed up my original comment, we should definitely be using `NO_PROGRESS` instead of `PROGRESS`. But regardless, the logic for the display is means that battery and title would have to show up in `NO_PROGRESS` and `FULL` mode, while progress only shows up in `FULL` mode. The original comment was more about creating a few new local variables to make it clear to any future travelers the translation between the `SETTINGS.statusBar` value and the UI components which are shown, it would actually read:

```cpp
const bool showProgress = SETTINGS.statusBar == CrossPointSettings::STATUS_BAR_MODE::FULL;
const bool showBattery = SETTINGS.statusBar == CrossPointSettings::STATUS_BAR_MODE::NO_PROGRESS || SETTINGS.statusBar == CrossPointSettings::STATUS_BAR_MODE::FULL;
const bool showChapterTitle = SETTINGS.statusBar == CrossPointSettings::STATUS_BAR_MODE::NO_PROGRESS || SETTINGS.statusBar == CrossPointSettings::STATUS_BAR_MODE::FULL;
```

then in the lines below we can just go:

```cpp
if (showProgress) {
   // ... show progress logic
}

if (showBattery) {
  // ... show battery logic
}

if (showChapterTitle) {
  // ... show chapter title logic
}
```

It just separates the display logic and business logic a bit making it a bit easier to parse.

src/CrossPointSettings.h
daveallie: I would love to keep options low for most settings as the complexity of understanding the software grows exponentially with additional options/values. I really don't think `BATTERY` alone would be a valuable setting. Especially if we just add the battery to the home screens. Thoughts on removing it?

src/activities/reader/EpubReaderActivity.cpp
1991AcuraLegend: That works for me, readability for future maintainers is good. So tying it all together with the bottom thread, I'll update it to be this: 

- `NONE` = 0 = Show nothing
- `NO_PROGRESS` = 1 = Show the battery and the title
- `FULL` = 2 = Show the battery, title, and progress

daveallie (approved):

PR #113, opened 2025-12-23
Network details QR code

jonasdiemer:
Using QRCode library from pio to generate the QR code.

Done:
- Display QR code for URL in network mode
- minor fixes of layout
- Display QR for URL in AP mode
- Display QR for AP in AP mode

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,12 +12,6 @@ jobs:
       - uses: actions/checkout@v6
         with:
           submodules: recursive
-      - uses: actions/cache@v5
-        with:
-          path: |
-            ~/.cache/pip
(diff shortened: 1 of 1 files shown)


platformio.ini
```diff
@@ -40,6 +40,7 @@ lib_deps =
   InputManager=symlink://open-x4-sdk/libs/hardware/InputManager
   EInkDisplay=symlink://open-x4-sdk/libs/display/EInkDisplay
   ArduinoJson @ 7.4.2
+  QRCode
```
gebeto: I think it makes sense to lock this dependency version

gebeto: I think it makes sense to lock this dependency version

daveallie (approved):

PR #115, opened 2025-12-23
Feature / Enable CORS for Web API

gebeto:
## Summary

Enable CORS to be able to use API from any web service. It is going to be used by:
https://github.com/gebeto/crosspoint-web-manager

## Additional Context

Currently this Web UI is in development, and there are just some dummy data, but you can use it with your existing device right now, just use link like this, and put your device IP there. But first you need to use this firmware with enabled CORS, otherwise it is not going to work:

https://gebeto.github.io/crosspoint-web-manager/?api_url=http://{YOUR_DEVICE_IP}


gebeto:
No need for CORS, because there are another way

PR #116, opened 2025-12-23
Improve EPUB cover image quality with pre-scaling and Atkinson dithering

eunchurn:
## Summary

Replace simple threshold-based grayscale quantization with ordered dithering using a 4x4 Bayer matrix. This eliminates color banding artifacts and produces smoother gradients on e-ink display.

- Add 4x4 Bayer dithering matrix for 16-level threshold patterns
- Modify `grayscaleTo2Bit()` function to accept pixel coordinates and apply position-based dithering
- Replace simple `grayscale >> 6` threshold with ordered dithering algorithm that produces smoother gradients

## Additional Context

* Bayer matrix approach: The 4x4 Bayer matrix creates a repeating pattern that distributes quantization error spatially, effectively simulating 16 levels of gray using only 4 actual color levels (black, dark gray, light gray, white).

* Cache invalidation: Existing cached `cover.bmp` files will need to be deleted to see the improved rendering, as the converter only runs when the cache is missing.

### DIFF
--- a/src/activities/reader/ReaderActivity.cpp
+++ b/src/activities/reader/ReaderActivity.cpp
@@ -15,10 +15,6 @@ std::unique_ptr<Epub> ReaderActivity::loadEpub(const std::string& path) {
 
   auto epub = std::unique_ptr<Epub>(new Epub(path, "/.crosspoint"));
   if (epub->load()) {
-    // Pre-generate cover BMP for sleep screen (so it's cached when entering sleep)
-    if (!epub->generateCoverBmp()) {
-      Serial.printf("[%lu] [RDR] Cover BMP generation skipped or failed\n", millis());
-    }
     return epub;
   }
 


mcmphd:
Hello!  I was wondering if you looked at applying a blue noise mask for dithering before settling on Bayer 4x4?  This paper describes the technique: https://www.hajim.rochester.edu/ece/sites/parker/assets/pdf/44%20-%20Digital%20halftoning%20technique%20using%20a%20blue-noise%20mask.pdf

daveallie:
Hey @eunchurn thanks for this, keen to check it out in a few days once I'm back from holidays.

What are your thoughts on applying this to 8/24/32 bpp bmp rendering in `lib/GfxRenderer/Bitmap.cpp` too? Seems like 4x4 dithering is going to be an improvement for all rendering

daveallie:
It looks drastically better at full res

| Before | After |
| --- | --- |
| cover1 | covera1 |
| cover2 | covera2 |
| cover3 | covera3 |
| cover4 | covera4 |
| cover5 | covera5 |

However on the device when it's downrezed, we skip every nth column and every mth row in order to resize the image to the screen size, as a result of the even nature of the dithering, we end up with vertical and horizontal lines appearing and weird artifacting around the changes of colours.

IMG_5678

Seems like noise is going to be necessary here to make resized dithering work

Note: You may want the changes in https://github.com/daveallie/crosspoint-reader/pull/120 (merged into master) to see the same thing, the JPG processing in master is broken

CaptainFrito:
How about resizing the images when creating the cover BMPs? We could then dither the resized image, that way the device will always render it in its native size.

eunchurn:
After investigating the downsampling artifacts issue, I found the root cause: the device renders images at a smaller size than the source, which destroys any dithering pattern applied to the full-resolution image.

The fix is to **resize the image to the exact display dimensions (480×800) BEFORE applying dithering**. This ensures:

- The dithering pattern is applied at the final display resolution
- No post-dithering downsampling occurs
- The dithering pattern remains intact on screen

### Implementation Details

- Uses fixed-point (16.16) math for sub-pixel accurate scaling
- Area averaging for smooth downsampling (accumulates multiple source pixels per output pixel)
- Maintains aspect ratio while fitting within 480×800
## Switched from Bayer to Atkinson Dithering

While Bayer ordered dithering works well, I found **Atkinson dithering** produces cleaner results for e-ink displays:

- Distributes only 75% of the quantization error (6/8 instead of 100%)
- Less error buildup = fewer artifacts
- Produces a lighter, cleaner look compared to Floyd-Steinberg
- Originally developed for early Macintosh displays (similar low-grayscale environment)

The Bayer and clustered-dot approaches are still available as compile-time options if needed.
## Additional Image Processing Enhancements

Added configurable brightness/contrast/gamma adjustments to improve visibility on e-ink:

```cpp
constexpr bool USE_BRIGHTNESS = true;
constexpr int BRIGHTNESS_BOOST = 10;      // Offset (0-50)
constexpr bool GAMMA_CORRECTION = true;   // Brightens midtones
constexpr float CONTRAST_FACTOR = 1.15f;  // Enhances separation
```

Processing order: **Contrast -> Brightness -> Gamma -> Dithering**

This helps compensate for e-ink displays that tend to appear darker than LCD screens.

## Compile-Time Configuration Options

All processing options are easily configurable at the top of `JpegToBmpConverter.cpp`:

```cpp
// Output format
constexpr bool USE_8BIT_OUTPUT = false;  // false = 2-bit (4 levels)

// Dithering method (choose one)
constexpr bool USE_ATKINSON = true;        // Recommended
constexpr bool USE_FLOYD_STEINBERG = false;
constexpr bool USE_NOISE_DITHERING = false;

// Pre-scaling (critical for quality)
constexpr bool USE_PRESCALE = true;
constexpr int TARGET_MAX_WIDTH = 480;
constexpr int TARGET_MAX_HEIGHT = 800;
```

This allows easy A/B testing of different algorithms.

src/activities/reader/ReaderActivity.cpp
```diff
@@ -15,6 +15,10 @@ std::unique_ptr<Epub> ReaderActivity::loadEpub(const std::string& path) {
 
   auto epub = std::unique_ptr<Epub>(new Epub(path, "/.crosspoint"));
   if (epub->load()) {
+    // Pre-generate cover BMP for sleep screen (so it's cached when entering sleep)
+    if (!epub->generateCoverBmp()) {
+      Serial.printf("[%lu] [RDR] Cover BMP generation skipped or failed\n", millis());
+    }
```
daveallie: Not sure if I'm fully on board with this one line here, if there's a crash in `generateCoverBmp` it's going to prevent people from opening the reader. I don't think it's unreasonable to pay the one time cost of generation the first time the device goes to sleep.

eunchurn: Good point. I've removed this code.

There was an issue where some books with certain image sizes caused very long delays when entering sleep mode. With the improved algorithm now, entering sleep mode shouldn't take too long, so it makes sense to remove the pre-generation and let it happen on-demand when actually entering sleep.

daveallie (approved):

PR #128, opened 2025-12-25
Add retry logic and progress bar for chapter indexing

eunchurn:
## Summary

Improve reliability and user experience during chapter indexing by adding retry logic for SD card operations and a visual progress bar.

- **Retry logic**: Add 3 retry attempts with 50ms delay for ZIP to SD card streaming to handle timing issues after display refresh
- **Progress bar**: Display a visual progress bar (0-100%) during chapter indexing based on file read progress, updating every 10% to balance responsiveness with e-ink display limitations

## Additional Context

* **Problem observed**: When navigating quickly through books with many chapters (before chapter titles finish rendering), the "Indexing..." screen would appear frozen. Checking the serial log revealed the operation had silently failed, but the UI showed no indication of this. Users would likely assume the device had crashed. Pressing the next button again would resume operation, but this behavior was confusing and unexpected.

* **Solution**:
- Retry logic handles transient SD card timing failures automatically, so users don't need to manually retry
- Progress bar provides visual feedback so users know indexing is actively working (not frozen)

* **Why timing issues occur**: After display refresh operations, there can be timing conflicts when immediately starting SD card write operations. This is more likely to happen when rapidly navigating through chapters.

* **Progress bar design**: Updates every 10% to avoid excessive e-ink refreshes while still providing meaningful feedback during long indexing operations (especially for large chapters with CJK characters).

* **Performance**: Minimal overhead - progress calculation is simple byte counting, and display updates use `FAST_REFRESH` mode.

### DIFF
--- a/lib/Epub/Epub/Section.cpp
+++ b/lib/Epub/Epub/Section.cpp
@@ -115,12 +115,15 @@ bool Section::clearCache() const {
 
 bool Section::persistPageDataToSD(const int fontId, const float lineCompression, const int marginTop,
                                   const int marginRight, const int marginBottom, const int marginLeft,
-                                  const bool extraParagraphSpacing, const std::function<void(int)>& progressFn) {
+                                  const bool extraParagraphSpacing, const std::function<void()>& progressSetupFn,
+                                  const std::function<void(int)>& progressFn) {
+  constexpr size_t MIN_SIZE_FOR_PROGRESS = 50 * 1024;  // 50KB
   const auto localPath = epub->getSpineItem(spineIndex).href;
   const auto tmpHtmlPath = epub->getCachePath() + "/.tmp_" + std::to_string(spineIndex) + ".html";
 
   // Retry logic for SD card timing issues
   bool success = false;
+  size_t fileSize = 0;
   for (int attempt = 0; attempt < 3 && !success; attempt++) {
     if (attempt > 0) {
       Serial.printf("[%lu] [SCT] Retrying stream (attempt %d)...\n", millis(), attempt + 1);
@@ -137,6 +140,7 @@ bool Section::persistPageDataToSD(const int fontId, const float lineCompression,
       continue;
     }
     success = epub->readItemContentsToStream(localPath, tmpHtml, 1024);
+    fileSize = tmpHtml.size();
     tmpHtml.close();
 
     // If streaming failed, remove the incomplete file immediately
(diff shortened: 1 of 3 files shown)


lib/Epub/Epub/parsers/ChapterHtmlSlimParser.cpp
```diff
@@ -249,6 +254,16 @@ bool ChapterHtmlSlimParser::parseAndBuildPages() {
...
     }
 
+    // Update progress (call every 10% change to avoid too frequent updates)
+    bytesRead += len;
+    if (progressFn && totalSize > 0) {
+      const int progress = static_cast<int>((bytesRead * 100) / totalSize);
+      if (progress != lastProgress && progress % 10 == 0) {
+        lastProgress = progress;
+        progressFn(progress);
+      }
```
daveallie: progress has to fall on a 10% exactly to get an update, instead it'd be better to do something like: `lastProgress / 10 != progress / 10`

daveallie (commented):
Overall not opposed to a progress bar here for larger chapters, but for smaller ones the rendering time is taking up more than half the time to generate, which isn't really a great experience. We can't access the SD card and the display at the same time due to SPI contention, so perhaps we should only show the progress bar chapters larger than a certain size?

eunchurn:
> Overall not opposed to a progress bar here for larger chapters, but for smaller ones the rendering time is taking up more than half the time to generate, which isn't really a great experience. We can't access the SD card and the display at the same time due to SPI contention, so perhaps we should only show the progress bar chapters larger than a certain size?

I've added a minimum file size threshold (MIN_SIZE_FOR_PROGRESS = 50KB) - smaller chapters won't show the progress bar since the rendering overhead isn't worth the visual feedback.

Also fixed the progress boundary check as you suggested: lastProgress / 10 != progress / 10

Commit: f41e749

eunchurn:
## Additional Commits

### Bug Fixes
- **fix(epub): prevent blank pages from corrupted index files** - Clean up incomplete temp files, add data validation in TextBlock deserialization
- **fix(parser): handle oversized words in line break algorithm** - Fix cascade failure when word exceeds page width

These fixes address issues discovered while testing the progress bar feature.

lib/Epub/Epub/blocks/TextBlock.cpp
```diff
@@ -46,6 +53,13 @@ std::unique_ptr<TextBlock> TextBlock::deserialize(File& file) {
 
   // words
   serialization::readPod(file, wc);
+
+  // Sanity check: prevent allocation of unreasonably large lists (max 10000 words per block)
+  if (wc > 10000) {
+    Serial.printf("[%lu] [TXB] Deserialization failed: word count %u exceeds maximum\n", millis(), wc);
+    return nullptr;
+  }
```
daveallie, line 61: Fair sanity check, this shouldn't be possible with https://github.com/daveallie/crosspoint-reader/blob/2771579007656f78fec756987679abb98cc1ded8/lib/Epub/Epub/parsers/ChapterHtmlSlimParser.cpp#L148-L157, but happy to leave in.

lib/Epub/Epub/ParsedText.cpp
```diff
@@ -106,6 +106,18 @@ std::vector<size_t> ParsedText::computeLineBreaks(const int pageWidth, const int
...
+    // This prevents cascade failure where one oversized word breaks all preceding words
+    if (dp[i] == MAX_COST) {
+      ans[i] = i;  // Just this word on its own line
+      // Inherit cost from next word to allow subsequent words to find valid configurations
+      if (i + 1 < static_cast<int>(totalWordCount)) {
+        dp[i] = dp[i + 1];
+      } else {
+        dp[i] = 0;
+      }
+    }
```
daveallie, line 120: Great find, this completely addresses https://github.com/daveallie/crosspoint-reader/issues/126

There was a long word in that paragraph overflowing the page

Issue #126, opened 2025-12-25
large blank section in book


fancyboy96:
### Affected Version

v0.9.0

### Describe the Bug

Large blank section of text in the chapter that does not appear when opening the same .epub on a different device.

### Steps to Reproduce

1. Load "House of Suns.epub" onto xteink4 device with Crosspoint
2. Navigate to 'chapter 2', page 19/72
3. There should be a blank section that appears until page 65/72

### Expected Behavior

Expect that the chapter should appear unbroken without blank spaces

### Relevant Log Output/Screenshots

```shell

```

daveallie:
Can confirm that I can replicate the same behaviour, that's odd, will look deeper into this, thanks

daveallie:
Partially fixed with #132, still needs some attention.

daveallie:
Fixed by a small fix in https://github.com/daveallie/crosspoint-reader/pull/128

fancyboy96:
thank you, looking forward to seeing the fix in prod!

daveallie:
Yep, should be fixed in the latest release: 0.10.0

PR #132, opened 2025-12-26
fix(parser): remove MAX_LINES limit that truncates long chapters

eunchurn:
## Summary

* **What is the goal of this PR?** Fixes a bug where text disappears after approximately 25 pages in long chapters during EPUB indexing.

    - Removed the `MAX_LINES = 1000` hard limit in `ParsedText::computeLineBreaks()`
    - Added safer infinite loop prevention by checking if `nextBreakIndex <= currentWordIndex` and forcing advancement by one word when stuck

## Additional Context

* **Root cause:** The `MAX_LINES = 1000` limit was introduced to prevent infinite loops, but it truncates content in long chapters. For example, a 93KB chapter that generates ~242 pages (~9,680 lines) gets cut off at ~1000 lines, causing blank pages after page 25-27.

* **Solution approach:** Instead of a hard line limit, I now detect when the line break algorithm gets stuck (when `nextBreakIndex` doesn't advance) and force progress by moving one word at a time. This preserves the infinite loop protection while allowing all content to be rendered.

* **Testing:** Verified with a Korean EPUB containing a 93KB chapter - all 242 pages now render correctly without text disappearing.

### DIFF
--- a/lib/Epub/Epub/ParsedText.cpp
+++ b/lib/Epub/Epub/ParsedText.cpp
@@ -111,16 +111,17 @@ std::vector<size_t> ParsedText::computeLineBreaks(const int pageWidth, const int
   // Stores the index of the word that starts the next line (last_word_index + 1)
   std::vector<size_t> lineBreakIndices;
   size_t currentWordIndex = 0;
-  constexpr size_t MAX_LINES = 1000;
 
   while (currentWordIndex < totalWordCount) {
-    if (lineBreakIndices.size() >= MAX_LINES) {
-      break;
+    size_t nextBreakIndex = ans[currentWordIndex] + 1;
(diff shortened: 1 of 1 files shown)


daveallie (approved):
Thanks for this fix @eunchurn, this hasn't been something which has been experienced much in english books as the 1000 line limit applies to each text block within the chapter which often ends before that limit.

I think there is some other bug in the line break logic as in https://github.com/daveallie/crosspoint-reader/issues/126 the blank pages issue is happening but there are plenty of paragraph breaks. With this fix it's now showing one word per line, which is better than a massive blank space.

Keen to get this merged, but there's also a little further work which needs to be looked into to fix #126

PR #133, opened 2025-12-26
Support swapping the functionality of the front buttons

dangson:
## Summary

Adds a setting to swap the front buttons. The default functionality are: Back/Confirm/Left/Right. When this setting is enabled they become: Left/Right/Back/Confirm. This makes it more comfortable to use when holding in your right hand since your thumb can more easily rest on the next button. The original firmware has a similar setting.

- Add the new setting.
- Create a mapper to dynamically switch the buttons based on the setting.
- Use mapper on the various activity screens.
- Update the button hints to reflect the swapped buttons.

## Additional Context

Full disclosure: I used Codex CLI to put this PR together, but did review it to make sure it makes sense.

Also tested on my device:
https://share.cleanshot.com/k76891NY

### DIFF
--- a/src/activities/Activity.h
+++ b/src/activities/Activity.h
@@ -5,18 +5,19 @@
 #include <string>
 #include <utility>
 
-class InputManager;
+#include "../MappedInputManager.h"
+
 class GfxRenderer;
 
 class Activity {
  protected:
   std::string name;
   GfxRenderer& renderer;
-  InputManager& inputManager;
+  MappedInputManager& mappedInput;
 
  public:
-  explicit Activity(std::string name, GfxRenderer& renderer, InputManager& inputManager)
-      : name(std::move(name)), renderer(renderer), inputManager(inputManager) {}
+  explicit Activity(std::string name, GfxRenderer& renderer, MappedInputManager& mappedInput)
+      : name(std::move(name)), renderer(renderer), mappedInput(mappedInput) {}
   virtual ~Activity() = default;
   virtual void onEnter() { Serial.printf("[%lu] [ACT] Entering activity: %s\n", millis(), name.c_str()); }
   virtual void onExit() { Serial.printf("[%lu] [ACT] Exiting activity: %s\n", millis(), name.c_str()); }
--- a/src/activities/settings/SettingsActivity.cpp
+++ b/src/activities/settings/SettingsActivity.cpp
@@ -9,7 +9,7 @@
 
 // Define the static settings list
 namespace {
-constexpr int settingsCount = 6;
+constexpr int settingsCount = 7;
 const SettingInfo settingsList[settingsCount] = {
     // Should match with SLEEP_SCREEN_MODE
     {"Sleep Screen", SettingType::ENUM, &CrossPointSettings::sleepScreen, {"Dark", "Light", "Custom", "Cover"}},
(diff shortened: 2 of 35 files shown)


daveallie (changes requested):
To avoid having to swap between `inputManager` and `frontButtonMapper` depending on use case, let's upgrade the button mapper to make it able to handle all buttons. There have been asks to also swap the volume button functions when reading as well, which I think could be handled by this.

Imo the best way to do this is to map the current buttons from InputManager to a slightly larger set of virtual buttons. It'll be the same set of 7 hardware buttons plus two new ones: page forward, and page backwards. The reason for this is that we don't want to swap the up/down function, but we do want the page forward / page backwards to be adjustable.

If you could amend `FrontButtonMapper` to include this new functionality, and rename it to something like `MappedInputManager` or something then we should be able to cover a few requests with this one.

dangson:
@daveallie I updated the PR based on your suggestion. Since this PR is already kinda large, I didn't add the setting to swap the side buttons, but will follow up with another PR to do so once this is merged.

src/activities/settings/SettingsActivity.cpp
```diff
@@ -20,6 +20,7 @@ const SettingInfo settingsList[settingsCount] = {
      SettingType::ENUM,
      &CrossPointSettings::orientation,
      {"Portrait", "Landscape CW", "Inverted", "Landscape CCW"}},
+    {"Swap Front Buttons", SettingType::TOGGLE, &CrossPointSettings::swapFrontButtons, {}},
```
daveallie: "Swap" is a but unclear, even if were not doing the side buttons, there are a few different configurations which we may not want to add in this PR, but starting as an enum (even just the two values) will make this easier to expand.

Something like, "Front Button Layout", with the two string values shown to users as "Bck, Cnfrm, Lft, Rght", and "Lft, Rght, Bck, Cnfrm" so that it's more clear what the setting is actually affecting

src/activities/settings/SettingsActivity.cpp
```diff
@@ -174,9 +177,14 @@ void SettingsActivity::render() const {
...
-  renderer.drawText(SMALL_FONT_ID, pageWidth - 20 - renderer.getTextWidth(SMALL_FONT_ID, CROSSPOINT_VERSION),
-                    pageHeight - 30, CROSSPOINT_VERSION);
+  const auto labels = mappedInput.mapLabels("« Save", "Toggle", "", "");
+  renderer.drawButtonHints(UI_FONT_ID, labels.btn1, labels.btn2, labels.btn3, labels.btn4);
+  if (SETTINGS.swapFrontButtons) {
+    renderer.drawText(SMALL_FONT_ID, 20, pageHeight - 30, CROSSPOINT_VERSION);
+  } else {
+    renderer.drawText(SMALL_FONT_ID, pageWidth - 20 - renderer.getTextWidth(SMALL_FONT_ID, CROSSPOINT_VERSION),
+                      pageHeight - 30, CROSSPOINT_VERSION);
+  }
```
daveallie: Lets just move this text above the button hints to avoid having to move it in future complex cases

src/activities/Activity.h
```diff
@@ -13,10 +15,11 @@ class Activity {
   std::string name;
   GfxRenderer& renderer;
   InputManager& inputManager;
+  mutable MappedInputManager mappedInput;
```
daveallie: Can we completely replace `InputManager` with `MappedInputManager`? Constructing it in the top of main.cpp and passing it down in place of input manager would be my preference.

daveallie (approved):
Thanks!

PR #135, opened 2025-12-27
Add XTC/XTCH ebook format support

eunchurn:
## Summary

Add support for XTC (XTeink X4 native) ebook format, which contains pre-rendered 480x800 1-bit bitmap pages optimized for e-ink displays.

- New `lib/Xtc/` library with XtcParser for reading XTC files
- XtcReaderActivity for displaying XTC pages on e-ink display
- XTC file detection in FileSelectionActivity
- Cover BMP generation from first XTC page
- Correct XTG page header structure (22 bytes) and bit polarity handling

## Additional Context

- XTC files contain pre-rendered bitmap pages with embedded status bar (page numbers, progress %)
- XTG page header: 22 bytes (magic + dimensions + reserved fields + bitmap size)
- Bit polarity: 0 = black, 1 = white
- No runtime text rendering needed - pages display directly on e-ink
- Faster page display compared to EPUB since no parsing/rendering required
- Memory efficient: loads one page at a time (48KB per page)
- Tested with XTC files generated from https://x4converter.rho.sh/
- Verified correct page alignment and color rendering
- Please report any issues if you test with XTC files from other sources.

### DIFF
--- a/src/activities/reader/XtcReaderActivity.cpp
+++ b/src/activities/reader/XtcReaderActivity.cpp
@@ -239,8 +239,10 @@ void XtcReaderActivity::renderPage() {
     // Display BW with conditional refresh based on pagesUntilFullRefresh
     if (pagesUntilFullRefresh <= 1) {
       renderer.displayBuffer(EInkDisplay::HALF_REFRESH);
+      pagesUntilFullRefresh = pagesPerRefresh;
     } else {
       renderer.displayBuffer();
+      pagesUntilFullRefresh--;
     }
 
     // Pass 2: LSB buffer - mark DARK gray only (XTH value 1)
@@ -283,9 +285,6 @@ void XtcReaderActivity::renderPage() {
 
     // Cleanup grayscale buffers with current frame buffer
     renderer.cleanupGrayscaleWithFrameBuffer();
-
-    // Decrement refresh counter
-    pagesUntilFullRefresh--;
 
     free(pageBuffer);
 


eunchurn:
## Update: XTCH (2-bit Grayscale) Support Added

Added support for XTCH format, which uses XTH pages with 2-bit grayscale (4 levels).

### Changes in this update

- **XTCH container format** (`.xtch` extension)
- **XTH page format** with unique storage layout:
  - Two bit planes stored sequentially (Bit1, then Bit2)
  - Column-major order, scanned right to left
  - 8 vertical pixels per byte (MSB = topmost)
  - Grayscale mapping: 0=White, 1=Dark Grey, 2=Light Grey, 3=Black
- Updated file browser to show `.xtch` files
- Cover BMP generation for XTCH files

### Format Comparison

| Format | Extension | Page Type | Bit Depth | Storage |
|--------|-----------|-----------|-----------|---------|
| XTC | `.xtc` | XTG | 1-bit | Row-major |
| XTCH | `.xtch` | XTH | 2-bit | Column-major, two bit-planes |

### Note

The XTH format uses a non-standard storage layout optimized for the Xteink e-paper display refresh pattern. The two-bit-plane column-major format matches the device's native LUT behavior.

src/activities/reader/XtcReaderActivity.cpp
```diff
@@ -0,0 +1,292 @@
...
+        const uint8_t bit1 = (plane1[byteOffset] >> bitInByte) & 1;
+        const uint8_t bit2 = (plane2[byteOffset] >> bitInByte) & 1;
+        const uint8_t pixelValue = (bit1 << 1) | bit2;
+
+        // Threshold: 0=White, 1,2,3=Dark (for best text contrast)
+        const bool isBlack = (pixelValue >= 1);
+
+        if (isBlack) {
+          renderer.drawPixel(x, y, true);
+        }
```
daveallie: Looks like it's an intentional leave, but did you think about rendering the grays as well?

eunchurn: Yes, grayscale rendering is implemented but hits a memory constraint:

- 2-bit XTCH `pageBuffer`: **96KB** (two bit planes)
- `storeBwBuffer()` for grayscale: **48KB** (needed for 3-pass rendering: BW → LSB → MSB)
- **Total: 144KB** needed, but only **~122KB** available

Added fallback(testing): when `storeBwBuffer()` fails, renders as BW instead. For true grayscale support, would need to stream page data 3 times instead of loading into memory once (slower but lower peak memory). But in practice, most cases will likely fall back to BW.

Any other solutions you can think of?

daveallie: yeah that's the tough part of rendering grayscale for the other parts of the app. What's the rendering time like for a 3 pass read of the page? If it's not horrendous, we should try to render the grays at the cost of some additional rendering time. If users are making a choice to use XTCH over XTC we should honour it.

eunchurn: Removed the need for `storeBwBuffer()` by:
1. Display BW first with `displayBuffer()`
2. Render LSB/MSB passes for grayscale
3. Call `displayGrayBuffer()`
4. Re-render BW to framebuffer (replaces `restoreBwBuffer`)

Peak memory reduced from 144KB to **96KB** (pageBuffer only).
Also optimized `PageInfo` struct (24→16 bytes) saving ~15KB for large books.

Page data is read once from SD, then 4 rendering passes use the in-memory buffer.

eunchurn: Due to limited XTCH sample files, it's difficult to verify if the grayscale rendering matches the converter's intended output. Would appreciate feedback on whether the results look correct.

lib/Xtc/Xtc/XtcParser.cpp
```diff
@@ -0,0 +1,317 @@
...
+    close();
+  }
+
+  // Open file
+  m_file = SD.open(filepath, FILE_READ);
+  if (!m_file) {
+    Serial.printf("[%lu] [XTC] Failed to open file: %s\n", millis(), filepath);
+    m_lastError = XtcError::FILE_NOT_FOUND;
+    return m_lastError;
+  }
```
daveallie, line 38: I'd like to keep using `FsHelpers` everywhere we open files, just for consistency. This can be updated to:

```cpp
if (!FsHelpers::openFileForRead("XTC", filepath, m_file)) {
  m_lastError = XtcError::FILE_NOT_FOUND;
  return m_lastError;
}
```

src/activities/reader/FileSelectionActivity.cpp
```diff
@@ -40,8 +40,19 @@ void FileSelectionActivity::loadFiles() {
...
+      std::string ext4 = filename.substr(filename.length() - 4);
+      if (ext4 == ".xtc") {
+        files.emplace_back(filename);
+      } else if (filename.length() >= 5) {
+        std::string ext5 = filename.substr(filename.length() - 5);
+        if (ext5 == ".xtch") {
+          files.emplace_back(filename);
+        }
+      }
     }
```
daveallie: nit: this could do with a little cleanup:

```cpp
if (file.isDirectory()) {
  files.emplace_back(filename + "/");
} else {
  std::string ext4 = filename.length() >= 4 ? filename.substr(filename.length() - 4) : "";
  std::string ext5 = filename.length() >= 5 ? filename.substr(filename.length() - 5) : "";
  if (ext5 == ".epub" || ext5 == ".xtch" || ext4 == ".xtc") {
    files.emplace_back(filename);
  }
}
```

Allocates a potentially useless string, but makes it a lot more readable

src/activities/reader/XtcReaderActivity.cpp
```diff
@@ -0,0 +1,354 @@
...
+    for (uint16_t y = 0; y < pageHeight; y++) {
+      for (uint16_t x = 0; x < pageWidth; x++) {
+        if (getPixelValue(x, y) >= 1) {
+          renderer.drawPixel(x, y, true);
+        }
+      }
+    }
+
+    // Reset refresh counter (grayscale display is a full refresh)
+    pagesUntilFullRefresh = pagesPerRefresh;
```
daveallie: There are a couple of bugs here.

1. We're not actually rendering any grayscale at all:
   - The buffer is initialized to 0x00 everywhere for the gray passes which is correct, but based on the LUT we're using, we need to be setting the pixel values to a `1` bit, and the way to do that is `renderer.drawPixel(x, y, false);`.
   - So when you're doing your LSB/MSB passes, it should be `false` as the third parameter.
   - A `0` bit in our display LUT means that the pixel should be untouched, so all grays are currently staying black
2. We're not correctly resetting the screen state between gray passes:
   - We call `restoreBwBuffer()` at the end of the grayscale rendering in the epub reader, which does a little more than just reset the in-memory buffer, it also calls `einkDisplay.cleanupGrayscaleBuffers` which set's the display's RED RAM buffer back to the same as it was when the initial black pass was done
   - To fix this, we may need to make a new function in `GfxRenderer` called something like `restoreBwBufferWithMainBuffer` which just calls `einkDisplay.cleanupGrayscaleBuffers(einkDisplay.getFrameBuffer());`.
   - This should be done after your pass 4 block.
 3. We don't actually need a `HALF_REFRESH` every page turn, if we fix the above two issues, we can get fast, low ghosting, gray pages and rely on the `pagesPerRefresh` value decrementing like usual
    - So you can conditionally do the `HALF_REFRESH` in the BW pass and replace the full reset of `pagesUntilFullRefresh` with a decrement

src/activities/reader/XtcReaderActivity.cpp
daveallie: I've left some review comments, I think with those grayscale rendering changes suggested it should be working.

lib/Xtc/Xtc/XtcParser.cpp
eunchurn: Done. Updated to use `FsHelpers::openFileForRead` for consistency.

src/activities/reader/FileSelectionActivity.cpp
eunchurn: Applied as suggested.

src/activities/reader/XtcReaderActivity.cpp
eunchurn: Thanks for the detailed explanation! Fixed all three issues:
- Changed LSB/MSB passes to use `drawPixel(x, y, false)`
- Added `cleanupGrayscaleWithFrameBuffer()` after Pass 4
- Made HALF_REFRESH conditional and switched to decrement

PR #156, opened 2025-12-29
Feature: go to text/start reference in epub guide section at first start

jonasdiemer:
This parses the guide section in the content.opf for text/start references and jumps to this on first open of the book.

Currently, this behavior will be repeated in case the reader manually jumps to Chapter 0 and then re-opens the book. IMO, this is an acceptable edge case (for which I couldn't see a good fix other than to drag a "first open" boolean around).

### DIFF
--- a/lib/Epub/Epub/BookMetadataCache.cpp
+++ b/lib/Epub/Epub/BookMetadataCache.cpp
@@ -9,7 +9,7 @@
 #include "FsHelpers.h"
 
 namespace {
-constexpr uint8_t BOOK_CACHE_VERSION = 2;
+constexpr uint8_t BOOK_CACHE_VERSION = 3;
 constexpr char bookBinFile[] = "/book.bin";
 constexpr char tmpSpineBinFile[] = "/spine.bin.tmp";
 constexpr char tmpTocBinFile[] = "/toc.bin.tmp";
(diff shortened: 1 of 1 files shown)


daveallie:
This has the potential to skip things like forewards and dedications / other preamble. If people want to skip over it, I think the chapter select screen should be sufficient. I'm not really keen to just auto-skip to chapter 1.

Spec: https://idpf.org/epub/20/spec/OPF_2.0.1_draft.htm#Section2.6

jonasdiemer:
Correct, it would skip the "front matter", including the cover and publisher images (which we don't render anyways). 

I am used to this behavior from kindle (maybe other readers) and thought it a good solution to skip over those "unnamed" and garbage chapters. 

I agree it could be confusing and/or not what ppl want. 

How about a popup with a question, incl a list of chapters that would be skipped?

or we could breselect the `text` chapter in the chapter select screen on first open?

pablohc:
@jonasdiemer @daveallie  When loading the book for the first time, it could be implemented so that holding down the “Right” button takes you directly to Chapter 1. Those who do not want to skip it will be able to move forward page by page.

daveallie:
After a bit more digging, it looks like the field of ereaders out there generally honor the guide and jump to the start.

Given that's where the crowd is, happy to merge this one, will just need a small conflict resolution.

jonasdiemer:
OK, I (hopefully) resolved the conflicts.

Anyhow, I think it might even be a good default behavior to show the Chapter selection activity on first open (with the `text` chapter pre-selected if it exists in the `guide`)... @daveallie, what do you think?

daveallie:
oops, sorry I think I made some more conflicts 😬 as you were resolving those ones (looks like you sorted it)

I think we should just throw them into the content, even just testing with apple books, it drops you right into the prelude from the book I was testing with.

lib/Epub/Epub/BookMetadataCache.cpp
daveallie: I ended up updating this from 1 -> 2 in a different PR, given the format changed below, can you please bump this to 3, I think you also had it as 1 -> 2, but might have gotten lost in the merge.

jonasdiemer: OK, should be done.

daveallie (approved):

PR #161, opened 2025-12-30
Split XTC file version into major minor bytes

daveallie:
## Summary

* Split XTC file version into major minor bytes
  * Continue to support both 1.0 and 0.1

## Additional Context

* See https://github.com/daveallie/crosspoint-reader/issues/146#issuecomment-3698223951 for more detail

FYI @treetrum @eunchurn


treetrum:
Very interesting — this makes more sense! Thanks for the tag.

PR #168, opened 2025-12-30
Feature: Show img alt text

jonasdiemer:
Let's start small by showing the ALT text of IMG. This is rudimentary, but avoids those otherwise completely blank chapters.

I feel we will need this even when we can render images if that rendering takes >1s - I would then prefer rendering optional and showing the ALT text first.

### DIFF
--- a/lib/Epub/Epub/parsers/ChapterHtmlSlimParser.cpp
+++ b/lib/Epub/Epub/parsers/ChapterHtmlSlimParser.cpp
@@ -70,14 +70,13 @@ void XMLCALL ChapterHtmlSlimParser::startElement(void* userData, const XML_Char*
     if (atts != nullptr) {
       for (int i = 0; atts[i]; i += 2) {
         if (strcmp(atts[i], "alt") == 0) {
-          alt = "Image[" + std::string(atts[i + 1]) + "]";
+          alt = "[Image: " + std::string(atts[i + 1]) + "]";
         }
       }
       Serial.printf("[%lu] [EHP] Image alt: %s\n", millis(), alt.c_str());
 
       self->startNewTextBlock(TextBlock::CENTER_ALIGN);
       self->italicUntilDepth = min(self->italicUntilDepth, self->depth);
-      self->boldUntilDepth = min(self->boldUntilDepth, self->depth);
       self->depth += 1;
       self->characterData(userData, alt.c_str(), alt.length());
 


jonasdiemer:
> Weirdly, one image (with alt text "image") shows without italics, but a longer multi-word title shows with correctly
> 
> I haven't dug into it yet to understand why, v odd

Not seeing this issue in my ebooks - both 1-word and longer alt-texts work well. Maybe something to do with what came before the `img`?

daveallie:
For reference the HTML is:

First:

```html
<?xml version='1.0' encoding='utf-8'?>
<html xmlns="http://www.w3.org/1999/xhtml" xmlns:epub="http://www.idpf.org/2007/ops" lang="en" xml:lang="en">
  <head>
    <title>Wind and Truth</title>
    <meta http-equiv="Content-Type" content="text/html; charset=utf-8"/>
  <link rel="stylesheet" type="text/css" href="../stylesheet.css"/>
<link rel="stylesheet" type="text/css" href="../page_styles.css"/>
</head>
  <body id="4OIQ0-3bf1eb34c54644a8b1aee1dd68e1b09d" class="calibre">
<div class="image"><span epub:type="pagebreak" role="doc-pagebreak" id="pg_1" aria-label="pg_1"></span><a href="part0002.html#book5"><img class="width" src="../images/00002.jpeg" alt="image"/></a></div>
</body></html>
```

Second:

```html
<?xml version='1.0' encoding='utf-8'?>
<html xmlns="http://www.w3.org/1999/xhtml" xmlns:epub="http://www.idpf.org/2007/ops" lang="en" xml:lang="en">
  <head>
    <title>Wind and Truth</title>
    <meta http-equiv="Content-Type" content="text/html; charset=utf-8"/>
  <link rel="stylesheet" type="text/css" href="../stylesheet.css"/>
<link rel="stylesheet" type="text/css" href="../page_styles.css"/>
</head>
  <body id="5N3C0-3bf1eb34c54644a8b1aee1dd68e1b09d" class="calibre">
<div class="image"><span epub:type="pagebreak" role="doc-pagebreak" id="pg_2" aria-label="pg_2"></span><span epub:type="pagebreak" role="doc-pagebreak" id="pg_3" aria-label="pg_3"></span><img id="pg0016" class="width1" src="../images/00003.jpeg" alt="A Map of Roshar, showing the Endless Ocean and the Reshi Isle in the North; Alethkar in the East; Azir, the Frostlands and Theylenah in the South; Iri, Shinovar and Aimia in the West."/></div>
</body></html>
```

First one is in an `a` tag and follows one page break, the second is after 2 page breaks and isn't in an `a` tag, that's the only difference I can see.

efenner:
I would love to see this get merged even if some of the text isn't italicized!

daveallie:
Will merge this one, but will also look into the italic-less representation separately

PR #169, opened 2025-12-30
Show book title instead of "Select Chapter".

jonasdiemer:
I think this is nicer ;)


lib/GfxRenderer/GfxRenderer.cpp
```diff
@@ -238,6 +238,17 @@ void GfxRenderer::displayBuffer(const EInkDisplay::RefreshMode refreshMode) cons
   einkDisplay.displayBuffer(refreshMode);
 }
 
+std::string GfxRenderer::truncatedText(const int fontId, const char* text, const int maxWidth,
```
daveallie, line 241: Nice add, this will likely get a bunch of use

PR #178, opened 2025-12-30
Fix file browser navigation for non-ASCII folder names

eunchurn:
## Summary

Fix file browser failing to navigate into subdirectories with non-ASCII (Korean/Unicode) folder names.

- Enable UTF-8 long file names in SdFat (`USE_UTF8_LONG_NAMES=1`)
- Add directory validation before iterating files
- Add `rewindDirectory()` call for stability

### DIFF
--- a/src/activities/reader/FileSelectionActivity.cpp
+++ b/src/activities/reader/FileSelectionActivity.cpp
@@ -30,7 +30,15 @@ void FileSelectionActivity::taskTrampoline(void* param) {
 void FileSelectionActivity::loadFiles() {
   files.clear();
   selectorIndex = 0;
+
   auto root = SdMan.open(basepath.c_str());
+  if (!root || !root.isDirectory()) {
+    if (root) root.close();
+    return;
+  }
+
+  root.rewindDirectory();
+
   char name[128];
   for (auto file = root.openNextFile(); file; file = root.openNextFile()) {
     file.getName(name, sizeof(name));
--- a/platformio.ini
+++ b/platformio.ini
@@ -26,6 +26,8 @@ build_flags =
   -DXML_GE=0
   -DXML_CONTEXT_BYTES=1024
   -std=c++2a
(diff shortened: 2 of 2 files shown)


src/activities/reader/FileSelectionActivity.cpp
```diff
@@ -30,7 +30,15 @@ void FileSelectionActivity::taskTrampoline(void* param) {
 void FileSelectionActivity::loadFiles() {
   files.clear();
   selectorIndex = 0;
+
   auto root = SdMan.open(basepath.c_str());
+  if (!root || !root.isDirectory()) {
+    if (root) root.close();
```
gebeto, line 36: this check looks strange
 - if outer "if" will not pass, then inner "if" will be not checked
- if outer "if" will pass, then inner "if" not pass

This just doesn't make sense

gebeto: also `!root.isDirectory` will panic if `root` will be null, so need to do something like that:
```c
if (!root || (root && !root.isDirectory()) {
```

or am I missing something?

daveallie: It looks fine to me:

`root` set and is directory
```cpp
if (false || false) {
  ... ignored ...
}
```

`root` set and is not directory
```cpp
if (false || true) {
  if (true) root.close();
  return;
}
```

`root` not set
```cpp
if (true || ...this is not checked as the if early returns...) {
  if (false) root.close();
  return;
}
```

To get past the `||` condition in the `if`, `root` must be set, so `root.isDirectory()` should not crash

gebeto: Ahhh, I see, thanks for explanation!

Damn, it was confusing 🙈

eunchurn: @daveallie Thanks for explaining it instead. That’s exactly right!

PR #181, opened 2025-12-30
Show "Entering Sleep" on black, so it's quicker to notice (in book).

jonasdiemer:
Black popup is easier to notice at higher contrast.

### DIFF
--- a/src/activities/boot_sleep/SleepActivity.cpp
+++ b/src/activities/boot_sleep/SleepActivity.cpp
@@ -40,7 +40,7 @@ void SleepActivity::onEnter() {
 }
 
 void SleepActivity::renderPopup(const char* message) const {
-  const int textWidth = renderer.getTextWidth(UI_12_FONT_ID, message, BOLD);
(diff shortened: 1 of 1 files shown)


src/activities/boot_sleep/SleepActivity.cpp
```diff
@@ -49,8 +49,8 @@ void SleepActivity::renderPopup(const char* message) const {
   const int w = textWidth + margin * 2;
   const int h = renderer.getLineHeight(UI_12_FONT_ID) + margin * 2;
   // renderer.clearScreen();
-  renderer.fillRect(x + 5, y + 5, w - 10, h - 10, false);
-  renderer.drawText(UI_12_FONT_ID, x + margin, y + margin, message);
+  renderer.fillRect(x + 5, y + 5, w - 10, h - 10, true);
+  renderer.drawText(UI_12_FONT_ID, x + margin, y + margin, message, false);
   renderer.drawRect(x + 5, y + 5, w - 10, h - 10);
```
daveallie: This `drawRect` is now no longer needed right? It'll be drawing a black square on the already black pixels from `fillRect`

daveallie (approved):

PR #183, opened 2025-12-30
Update button hints on OTA update screen and update user guide to reflect current settings

dangson:
## Summary

- Update the button hints on the OTA update screen to reflect the current front button layout.
- Update user guide to reflect current available settings. A lot has been added recently!


daveallie:
Thanks for updating the user guide, I've been meaning to jump back through the docs and see what was out of date.

PR #196, opened 2025-12-31
fix: eliminate e-ink ghosting on menu navigation and beautify UI

juicecultus:
## Problem
E-ink display showed significant ghosting when navigating between menu screens (file browser, settings, home, chapter selection). The custom LUT that clears ghosting (HALF_REFRESH) was not being applied consistently when entering or re-entering menu activities.

## Root Cause
1. **HALF_REFRESH only on initial creation**: The refresh flag was only set when an activity was first created, not when re-entering via back navigation.
2. **Race conditions**: The display task could consume the refresh flag before the directory-change render had a chance to use it.
3. **Empty directory bypass**: The early return for empty directories skipped the HALF_REFRESH logic entirely.

### FileSelectionActivity (most complex case)
- Replaced boolean flag with `lastRenderedPath` string comparison to detect actual directory changes race-free.
- Use `HALF_REFRESH` only when the path differs (directory changed), not on navigation within the same directory.
- Protected all `updateRequired` writes with a mutex to ensure synchronization between input and display tasks.
- Fixed the empty directory case to also apply HALF_REFRESH logic.

### Other Menu Activities (Home, Settings, EpubChapterSelection, XtcChapterSelection)
- Added `isFirstRender` flag reset in `onEnter()` method to ensure `HALF_REFRESH` is applied every time the activity is entered.

## UI Improvements
- **FileSelectionActivity**: 
  - Header shows "Browse" at root or the current path in subfolders.
  - Directories now show as `> FolderName` (removed trailing slashes).
  - Files are indented for better visual alignment.
  - Improved empty state messaging.
- **XtcReaderChapterSelectionActivity**: Refactored layout for consistency, added text truncation and button hints.
- **EpubReaderChapterSelectionActivity**: Added text truncation for long chapter titles and button hints.

## Testing
- Navigate Browse -> /folder -> back -> /folder: HALF_REFRESH triggers on each directory change.
- Navigate to empty folders: HALF_REFRESH triggers correctly.
- Navigate within same directory (up/down): uses FAST_REFRESH (no unnecessary slow refreshes).
- Enter/exit Settings, Home, Chapter Selection screens: HALF_REFRESH on each entry.

### DIFF
--- a/src/activities/reader/FileSelectionActivity.cpp
+++ b/src/activities/reader/FileSelectionActivity.cpp
@@ -220,7 +220,16 @@ void FileSelectionActivity::render() const {
   const auto pageStartIndex = selectorIndex / PAGE_ITEMS * PAGE_ITEMS;
   renderer.fillRect(0, listStartY + (selectorIndex % PAGE_ITEMS) * rowHeight - 2, pageWidth - 1, rowHeight);
   for (int i = pageStartIndex; i < files.size() && i < pageStartIndex + PAGE_ITEMS; i++) {
-    auto item = renderer.truncatedText(UI_10_FONT_ID, files[i].c_str(), pageWidth - horizontalMargin * 2 - 8);
+    const auto& filename = files[i];
+    const bool isDir = !filename.empty() && filename.back() == '/';
+    // Format: folders show as "> FolderName", files show as "  FileName"
+    std::string displayName;
+    if (isDir) {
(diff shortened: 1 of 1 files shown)


daveallie:
Pending conflict resolution before I give this one a review

daveallie:
Going to close this as I believe it's abandoned.

Issue #143, opened 2025-12-28
EPUB 3 support missing


abhollan:
### Affected Version

v0.10.0

### Describe the Bug

Every time I open an epub file it says “loading” and then “failed to load epub,” before returning to the file screen. This happens with any epub file I try to load. I can load .xtc files without any issue. I had this issue on 0.9.0 as well. I have tried reformatting my SD card, a new SD card, and multiple epub files. 

### Steps to Reproduce

1. Go to file browser
2. select any EPUB file 

### Expected Behavior

EPUB opens properly 

### Relevant Log Output/Screenshots

```shell

```

daveallie:
Would you mind providing a few details about the epub? Size/chapter count/book title?

I have some improvements coming in #140 which should help with memory load when opening books, which is likely the cause of the issues here.

abhollan:
I just got one to work: On the Calculation of Volume by Solvej Balle. It’s a very small file, 972 kb, that I loaded via WiFi. Incidentally, the WiFi loading (both on my home network and making a hotspot) barely work — they both crash the device after about 60 seconds, so I can’t load a longer book via WiFi. 

The ones that didn’t work:
- Gardens of the Moon by Steven Erickson, 2.2 MB, 24 chapters
- Mythos by Stephen Fry, 5.6 MB, 323 chapters
- There was Nothing We Could Do by Steven Hyden, 714 kb, 7 chapters

daveallie:
Thanks for raising this, I have seen some similar issues with the Wifi file transfer, will work improving that once I get through a few more internal function PRs

daveallie:
Hey @abhollan, can you confirm this is still an issue in [v0.11.0](https://github.com/daveallie/crosspoint-reader/releases/tag/0.11.0)? If it is could I get you to provide some logs from the device, easiest way if you not comfortable around a terminal is to open https://www.serialmonitor.org/, plug your device in, turn it on, and then press connect down the bottom left of https://www.serialmonitor.org/. Once you're seeing content there try to open one of those failing EPUBs and paste the logs into a comment in this issue. I'll be able to help more from then.

abhollan:
Still happens on 0.11.0. I tried reformatting my SD to exFAT but that didn't help.
I couldn't get the serialmonitor site to show anything at all when I connect my device. I am happy to use Terminal if you tell me how to do it, or I can also try on a different computer when I get to work.

oscarsantillana:
[...]

PR #197, opened 2025-12-31
Add EPUB 3 nav.xhtml TOC support

bigbag:
## Summary

  * **What is the goal of this PR?** Add EPUB 3 support by implementing native navigation document (nav.xhtml) parsing with NCX fallback, addressing issue Fixes: #143.

    - New `TocNavParser` for parsing EPUB 3 HTML5 navigation documents (`<nav epub:type="toc">`)
    - Detection of nav documents via `properties="nav"` attribute in OPF manifest
    - Fallback logic: try EPUB 3 nav first, fall back to NCX (EPUB 2) if unavailable
    - Graceful degradation: books without any TOC now load with a warning instead of failing

  ## Additional Context

  * The implementation follows the existing streaming XML parser pattern using Expat to minimize RAM usage on the ESP32-C3
  * EPUB 3 books that include both nav.xhtml and toc.ncx will prefer the nav document (per EPUB 3 spec recommendation)
  * No breaking changes - existing EPUB 2 books continue to work as before
  * Tested on examples from https://idpf.github.io/epub3-samples/30/samples.html

### DIFF
--- a/lib/Epub/Epub/parsers/TocNavParser.h
+++ b/lib/Epub/Epub/parsers/TocNavParser.h
@@ -1,0 +1,47 @@
+#pragma once
+#include <Print.h>
+#include <expat.h>
+
+#include <string>
+
+class BookMetadataCache;
+
+// Parser for EPUB 3 nav.xhtml navigation documents
+// Parses HTML5 nav elements with epub:type="toc" to extract table of contents
+class TocNavParser final : public Print {
+  enum ParserState {
+    START,
+    IN_HTML,
+    IN_BODY,
+    IN_NAV_TOC,  // Inside <nav epub:type="toc">
+    IN_OL,       // Inside <ol>
+    IN_LI,       // Inside <li>
+    IN_ANCHOR,   // Inside <a>
+  };
+
+  const std::string& baseContentPath;
+  size_t remainingSize;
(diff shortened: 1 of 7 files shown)


lib/Epub/Epub/parsers/TocNavParser.h
```diff
@@ -0,0 +1,49 @@
...
+  const std::string& baseContentPath;
+  size_t remainingSize;
+  XML_Parser parser = nullptr;
+  ParserState state = START;
+  BookMetadataCache* cache;
+
+  // Track nesting depth for <ol> elements to determine TOC depth
+  uint8_t olDepth = 0;
+  // Track if we're in the correct nav (epub:type="toc")
+  bool inTocNav = false;
```
daveallie: I don't think this variable is needed, whenever it's set to true, we also set the state to `IN_NAV_TOC`. We then only update the state to `IN_OL` / `IN_LI` / `IN_ANCHOR` if that variable is true, which could just be replaced with a check to `state == IN_NAV_TOC`. If we say that the parser can only go to `IN_OL` if it's coming from `IN_NAV_TOC` then we don't need to additionally track this.

bigbag: Yes, you are right. Fixed.

Issue #179, opened 2025-12-30
Book Covers of current book doesn't load on home screen


cjrezz:
### Affected Version

0.11.0

### Describe the Bug

I've tried 4 different books. 3 of them are unmodified epubs, 1 of them I put throught the calibre workflow for the x4 and resized the book cover for the x4 and none of them properly show the book cover on the home screen of the book I'm currently reading. It just resorts to the default phot of a black book with a white bookmark in the top right corner and the book title in the middle with 'continue reading' written at the bottom.

### Steps to Reproduce

Use ereader normally

### Expected Behavior

Show the book cover of the book you're currently reading

### Relevant Log Output/Screenshots

```shell

```

lukestein:
I don't think the currently implemented home screen is "supposed" to show cover images (so this behavior isn't really a bug?), but I agree that it would be awesome if it did!

Perhaps ideally we'd get covers (or at least the blank book image) for several of the most recently read titles.

cjrezz:
I believe I read somewhere in the release notes that it would try and use the book cover but if it couldn't it would revert to the book title?

--Sent from my mobile device--
________________________________

daveallie:
It's just a plain placeholder at all times at this point, once https://github.com/daveallie/crosspoint-reader/pull/108 gets in, we should be able to swap it over to using a better image. @cjrezz you might referring to the details around the other metadata, we try to populate the title and author from the book metadata to show on the placeholder, but fallback to the file name if the book cache is invalid

daveallie:
Resolved in #200

PR #200, opened 2026-01-01
Add cover image display in *Continue Reading* card with framebuffer caching

eunchurn:
## Summary

* **What is the goal of this PR?** (e.g., Fixes a bug in the user authentication module, 

Display the book cover image in the **"Continue Reading"** card on the home screen, with fast navigation using framebuffer caching.

- Display book cover image in the "Continue Reading" card on home screen
- Load cover from cached BMP (same as sleep screen cover)
- Add framebuffer store/restore functions (`copyStoredBwBuffer`, `freeStoredBwBuffer`) for fast navigation after initial render
- Fix `drawBitmap` scaling bug: apply scale to offset only, not to base coordinates
- Add white text boxes behind title/author/continue reading label for readability on cover
- Support both EPUB and XTC file cover images
- Increase HomeActivity task stack size from 2048 to 4096 for cover image rendering

## Additional Context

- Performance: First render loads cover from SD card (~800ms), subsequent navigation uses cached framebuffer (~instant)
- Memory: Framebuffer cache uses ~48KB (6 chunks × 8KB) while on home screen, freed on exit
- Fallback: If cover image is not available, falls back to standard text-only display
- The `drawBitmap` fix corrects a bug where screenY = (y + offset) scale was incorrectly scaling the base coordinates. Now correctly uses screenY = y + (offset scale)

### DIFF
--- a/src/activities/home/HomeActivity.cpp
+++ b/src/activities/home/HomeActivity.cpp
@@ -9,6 +9,7 @@
 #include <cstring>
 #include <vector>
 
+#include "Battery.h"
 #include "CrossPointSettings.h"
 #include "CrossPointState.h"
 #include "MappedInputManager.h"
@@ -562,8 +563,13 @@ void HomeActivity::render() {
   const auto labels = mappedInput.mapLabels("", "Confirm", "Up", "Down");
   renderer.drawButtonHints(UI_10_FONT_ID, labels.btn1, labels.btn2, labels.btn3, labels.btn4);
 
-  const auto batteryX = pageWidth - 25 - renderer.getTextWidth(SMALL_FONT_ID, "100 %");
-  ScreenComponents::drawBattery(renderer, batteryX, 10);
+  const bool showBatteryPercentage =
+      SETTINGS.hideBatteryPercentage != CrossPointSettings::HIDE_BATTERY_PERCENTAGE::HIDE_ALWAYS;
+  // get percentage so we can align text properly
+  const uint16_t percentage = battery.readPercentage();
+  const auto percentageText = showBatteryPercentage ? std::to_string(percentage) + "%" : "";
+  const auto batteryX = pageWidth - 25 - renderer.getTextWidth(SMALL_FONT_ID, percentageText.c_str());
+  ScreenComponents::drawBattery(renderer, batteryX, 10, showBatteryPercentage);
 
   renderer.displayBuffer();
 }
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,9 +1,18 @@
 ## Summary
 
-* **What is the goal of this PR?** (e.g., Fixes a bug in the user authentication module, Implements the new feature for 
-  file uploading.)
+* **What is the goal of this PR?** (e.g., Implements the new feature for file uploading.)
 * **What changes are included?**
 
 ## Additional Context
 
-* Add any other information that might be helpful for the reviewer (e.g., performance implications, potential risks, specific areas to focus on).
+* Add any other information that might be helpful for the reviewer (e.g., performance implications, potential risks, 
+  specific areas to focus on).
+
+---
+
+### AI Usage
+
+While CrossPoint doesn't have restrictions on AI tools in contributing, please be transparent about their usage as it 
+helps set the right context for reviewers.
+
+Did you use AI tools to help write this code? _**< YES | PARTIALLY | NO >**_
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,11 +7,11 @@ name: CI
 jobs:
   build:
     runs-on: ubuntu-latest
-
     steps:
       - uses: actions/checkout@v6
         with:
           submodules: recursive
+
       - uses: actions/setup-python@v6
         with:
           python-version: '3.14'
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -7,17 +7,18 @@ on:
 jobs:
   build-release:
     runs-on: ubuntu-latest
-
     steps:
       - uses: actions/checkout@v6
         with:
           submodules: recursive
+
       - uses: actions/cache@v5
         with:
           path: |
             ~/.cache/pip
             ~/.platformio/.cache
           key: ${{ runner.os }}-pio
+
       - uses: actions/setup-python@v6
         with:
           python-version: '3.14'
--- a/USER_GUIDE.md
+++ b/USER_GUIDE.md
@@ -67,6 +67,7 @@ The Settings screen allows you to configure the device's behavior. There are a f
   - "Light" - The same default sleep screen, on a white background
   - "Custom" - Custom images from the SD card, see [Sleep Screen](#36-sleep-screen) below for more information
   - "Cover" - The book cover image (Note: this is experimental and may not work as expected)
+  - "Blank" - A blank screen
 - **Status Bar**: Configure the status bar displayed while reading:
   - "None" - No status bar
   - "No Progress" - Show status bar without reading progress
--- a/lib/Epub/Epub.cpp
+++ b/lib/Epub/Epub.cpp
@@ -167,7 +167,10 @@ bool Epub::parseTocNavFile() const {
   }
   const auto navSize = tempNavFile.size();
 
-  TocNavParser navParser(contentBasePath, navSize, bookMetadataCache.get());
+  // Note: We can't use `contentBasePath` here as the nav file may be in a different folder to the content.opf
+  // and the HTMLX nav file will have hrefs relative to itself
+  const std::string navContentBasePath = tocNavItem.substr(0, tocNavItem.find_last_of('/') + 1);
+  TocNavParser navParser(navContentBasePath, navSize, bookMetadataCache.get());
 
   if (!navParser.setup()) {
     Serial.printf("[%lu] [EBP] Could not setup toc nav parser\n", millis());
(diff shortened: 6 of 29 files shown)


lukestein:
This PR hopefully closes #179

daveallie:
In order to keep the home screen rendering quick, I think we should be generating a 1-bit bmp to avoid having to do gray passes on the home screen. There's some existing implementation in this PR: https://github.com/daveallie/crosspoint-reader/pull/108

As it currently stands, this PR prepares a 2-bit bmp but renders all grays as black, so images are overly dark. Additionally, there seems to be an initialization issue as when I first land on the home screen, there's a bookmark icon on the cover, but after navigating down and up, it switched back to the expected border.

eunchurn:
> In order to keep the home screen rendering quick, I think we should be generating a 1-bit bmp to avoid having to do gray passes on the home screen. There's some existing implementation in this PR: #108
> 
> As it currently stands, this PR prepares a 2-bit bmp but renders all grays as black, so images are overly dark. Additionally, there seems to be an initialization issue as when I first land on the home screen, there's a bookmark icon on the cover, but after navigating down and up, it switched back to the expected border.

I've addressed all the issues:


1. 1-bit BMP with proper dithering

Changed from 2-bit to true 1-bit BMP format for thumbnails
Implemented Atkinson dithering (error diffusion) instead of simple thresholding
The algorithm: resize to target size (240x400) first, then apply Atkinson dithering
This produces much better quality than the previous approach where grays were rendered as black

2. Bookmark ribbon fix

Replaced the rect + triangle "carve out" approach with a proper polygon (pentagon shape)
Added `fillPolygon()` method to `GfxRenderer` using scanline fill algorithm
The bookmark now renders correctly as a ribbon with a V-notch at the bottom

3. Initialization issue fixed

Selection state (border + inverted bookmark) is now shown on first render, not just after navigation
The bookmark inverts properly when the Continue Reading card is selected

4. Additional improvements

Fixed 1-bit BMP palette lookup in `Bitmap::readRow()`
Added optimized `drawBitmap1Bit()` rendering path
XTC thumbnails also updated to use 1-bit format with noise dithering

The home screen now renders without gray passes while maintaining good image quality through Atkinson error diffusion dithering.

src/activities/home/HomeActivity.cpp
```diff
@@ -144,34 +211,126 @@ void HomeActivity::render() const {
...
+          coverBufferStored = renderer.storeBwBuffer();
+          coverRendered = true;
+
+          // First render: if selected, draw selection indicators now
+          if (bookSelected) {
+            renderer.drawRect(bookX + 1, bookY + 1, bookWidth - 2, bookHeight - 2);
+            renderer.drawRect(bookX + 2, bookY + 2, bookWidth - 4, bookHeight - 4);
+            // Invert bookmark to black
+            renderer.fillPolygon(xPoints, yPoints, 5, true);
+          }
```
daveallie: Sorry maybe the feedback was bad, but we want to be doing the opposite here, if we've shown a cover, we can omit the bookmark icon as it's not adding anything and just covering the art. If there's no cover, then we should show the bookmark as an additional visual flair.

Currently we show the bookmark icon when there is a cover, and show nothing when there's no cover.

src/activities/home/HomeActivity.cpp
```diff
@@ -144,34 +211,126 @@ void HomeActivity::render() const {
...
+              bookmarkY + bookmarkHeight,               // bottom-right
+              bookmarkY + bookmarkHeight - notchDepth,  // center notch point
+              bookmarkY + bookmarkHeight                // bottom-left
+          };
+
+          // Draw bookmark ribbon (white normally, will be inverted if selected)
+          renderer.fillPolygon(xPoints, yPoints, 5, false);
+
+          // Store the buffer with cover image AND bookmark for fast navigation
+          coverBufferStored = renderer.storeBwBuffer();
```
daveallie: I think this is a bit of a bastardisation of the `storeBwBuffer` function. That function was designed to temporarily hold the BW buffer for gray renders, we're using it to store a full screen frame atm.

If we'd like to store a full screen frame, then we should make it `HomeActivity`'s responsibility to store that frame. When constructing `HomeActivity` we should be building our own buffer and cloning the frame buffer into it, something like:

```cpp
// During initialization / entering activity:
auto baseFrame = static_cast<uint8_t *>(malloc(renderer.getBufferSize()));

// Instead of the this storeBwBuffer call
memcpy(baseFrame, renderer.getFrameBuffer(), renderer.getBufferSize());

// When restoring it
memcpy(renderer.getFrameBuffer(), baseFrame, renderer.getBufferSize());

// When exiting
free(baseFrame);
```

Alternatively, we could just copy a subregion of the frame buffer where the cover is (which is only 1/4 of the screen) and then display that back onto the screen when we need it.

lib/GfxRenderer/GfxRenderer.cpp
```diff
@@ -435,6 +577,38 @@ void GfxRenderer::restoreBwBuffer() {
...
+
+  return true;
+}
+
+/**
+ * Free the stored BW buffer chunks manually.
+ * Use this when you no longer need the stored buffer.
+ */
+void GfxRenderer::freeStoredBwBuffer() { freeBwBufferChunks(); }
+
```
daveallie: Given some of the feedback in the other files, I think we should remove this functionality as it opens up a little too much room for abuse for my liking.

lib/GfxRenderer/GfxRenderer.cpp
```diff
@@ -221,6 +228,141 @@ void GfxRenderer::drawBitmap(const Bitmap& bitmap, const int x, const int y, con
...
+      }
+      // White pixels (val == 3) are not drawn (leave background)
+    }
+  }
+
+  free(outputRow);
+  free(rowBytes);
+}
+
+void GfxRenderer::fillPolygon(const int* xPoints, const int* yPoints, int numPoints, bool state) const {
```
daveallie, line 318: Nice add, thanks!

src/activities/home/HomeActivity.cpp
```diff
@@ -1,14 +1,51 @@
...
+  str.resize(pos);
+  return pos;
+}
+
+// Truncate string by removing N UTF-8 characters from the end
+void utf8TruncateChars(std::string& str, size_t numChars) {
+  for (size_t i = 0; i < numChars && !str.empty(); ++i) {
+    utf8RemoveLastChar(str);
+  }
+}
```
daveallie, line 40: For now these can stay here, but will relocate them once StringUtils get merged in from the Calibre Syncing PR

src/activities/home/HomeActivity.cpp
eunchurn: I've refactored this to use HomeActivity's own buffer management. Added three private methods: `storeCoverBuffer()`, `restoreCoverBuffer()`, and `freeCoverBuffer()` that handle a dedicated `uint8_t* coverBuffer` member. This keeps the buffer responsibility within HomeActivity and leaves `storeBwBuffer`/`restoreBwBuffer` for their intended grayscale rendering purpose.

lib/GfxRenderer/GfxRenderer.cpp
eunchurn: Agreed. I've removed both `copyStoredBwBuffer()` and `freeStoredBwBuffer()` from GfxRenderer. The original `storeBwBuffer()` and `restoreBwBuffer()` are kept since they're still used by `EpubReaderActivity` for grayscale rendering as intended.

src/activities/home/HomeActivity.cpp
eunchurn: Done! I've inverted the logic as requested. Now the bookmark ribbon is only shown when there's no cover image (as visual decoration), and hidden when a cover is displayed so it doesn't obscure the artwork. The bookmark also properly inverts when selected in the no-cover case.

daveallie:
Apologies, I've created more merge conflicts with the merge of https://github.com/daveallie/crosspoint-reader/pull/302, if you could resolve them, then I can get this in.

itsthisjustin:
Fwiw I'd recommend that if the cover image is available that we hide the book name and author since that's usually on the book cover as well

lukestein:
> Fwiw I'd recommend that if the cover image is available that we hide the book name and author since that's usually on the book cover as well

Respectfully disagree. This info can be small or hard to read in the thumbnail depending on the book.

I do think we could use a smaller font in the overlay.

And the overlay should invert colors (white text on black background) when selected, consistent with the other buttons on the home screen.

itsthisjustin:
> > Fwiw I'd recommend that if the cover image is available that we hide the book name and author since that's usually on the book cover as well
> 
> Respectfully disagree. This info can be small or hard to read in the thumbnail depending on the book.
> 
> I do think we could use a smaller font in the overlay.
> 
> And the overlay should invert colors (white text on black background) when selected, consistent with the other buttons on the home screen.

Well, what does Kindle or others do? I've never seen a reader UI that has both the cover and the book title stacked on top of each other haha. To me, it looks like a bug or a modal with information i need to dismiss. I'm fine with it being there in addition to the cover but why have the cover if you're gonna cover it with a box? defeats the purpose of having a nice beautiful book cover at all

pablohc:
> Fwiw I'd recommend that if the cover image is available that we hide the book name and author since that's usually on the book cover as well

I second this. You're already supposed to know what book you're reading; you recognize the art, and there wouldn't be an issue with slightly less legible text in the art. Adding the current book's title label ruins the experience.

If book covers were shown in a grid instead of a list of titles, displaying the name below the cover would make sense, especially when you haven’t opened the book yet and the text is hard to read.
But overlaying the title on the cover adds nothing, neither in the current "Continue Reading" view nor in a future cover grid (if it ever happens).

@eunchurn Is it possible to modify this?

I've also been thinking that if the cover image is going to be scaled to fit the width of the container, vertical alignment might not be necessary. It could be aligned to the bottom, which would avoid the white boxes that currently appear above and below the covers.
I apologize if what I'm saying isn't clear; I could create a mockup to better illustrate the idea, in case the modification is feasible.

daveallie:
I am hesitant to completely remove the title when the book cover is being shown, perhaps we can show it below the cover to avoid occluding it, but I have a few books where the "cover" is just the Penguin Books logo or something equally silly.

Perhaps that's just a library management issue though

PR #203, opened 2026-01-01
Prevent device sleep during WiFi file transfer and OTA updates

### LINKED ISSUE #199: Device can fall asleep during wifi file transfer
### Affected Version

v0.11.2

### Describe the Bug

Joining a network and accessing the interface in the browser works great, but as there are no keypresses on the device, it falls asleep after 10 minutes. Is it possible to increase or remove the timeout when in this state?

### Steps to Reproduce

1. Go to file transfer
2. Press on "Join a Network"
3. Join a network
4. Access the web interface
5. After 10 minutes of no button presses on the device, it falls asleep, severing the webserver

### Expected Behavior

Ideally the device doesn't sleep, or perhaps stays awake for 30 minutes or something instead

```shell

```

bigbag:
## Summary

  * **What is the goal of this PR?** Fixes #199 - Device falls asleep during WiFi file transfer after 10 minutes of inactivity, disconnecting the web server.

    - Add `preventAutoSleep()` virtual method to `Activity` base class
    - Modify main loop to reset inactivity timer when `preventAutoSleep()` returns true
    - Override `preventAutoSleep()` in `CrossPointWebServerActivity` (returns true when web server running)
    - Override `preventAutoSleep()` in `OtaUpdateActivity` (returns true during update check/download)

  ## Additional Context

  * The existing `skipLoopDelay()` method controls loop timing (yield vs delay) for HTTP responsiveness. The new `preventAutoSleep()` method is semantically separate - it explicitly signals that an activity should keep the device awake.
  * `CrossPointWebServerActivity` uses both methods: `skipLoopDelay()` for responsive HTTP handling, `preventAutoSleep()` for staying awake.
  * `OtaUpdateActivity` only needs `preventAutoSleep()` since the OTA library handles HTTP internally.

### DIFF
--- a/src/activities/Activity.h
+++ b/src/activities/Activity.h
@@ -22,4 +22,5 @@ class Activity {
   virtual void onExit() { Serial.printf("[%lu] [ACT] Exiting activity: %s\n", millis(), name.c_str()); }
   virtual void loop() {}
   virtual bool skipLoopDelay() { return false; }
+  virtual bool preventAutoSleep() { return false; }
 };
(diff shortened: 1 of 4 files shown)


daveallie:
Thanks for this!

PR #216, opened 2026-01-02
Refactor semantic version comparison for OTA updates

embedded4ever:
## Summary

* **What is the goal of this PR?** (e.g., Fixes a bug in the user authentication module, Implements the new feature for 
  file uploading.)

This PR refactors the semantic version comparison logic used during OTA update checks.

Memory stats before : 
RAM:   [===       ]  30.8% (used 101068 bytes from 327680 bytes)
Flash: [========= ]  85.7% (used **5617830** bytes from 6553600 bytes)

Memory stats before : 
RAM:   [===       ]  30.8% (used 101068 bytes from 327680 bytes)
Flash: [========= ]  85.7% (used **5616870** bytes from 6553600 bytes)

Replaced std::string::substr() and std::stoi() based parsing with a lightweight, heap-free approach.
Version parsing is now done in a single pass without creating temporary std::string objects.
Behavior remains identical: versions are still compared as MAJOR.MINOR.PATCH.

## Additional Context

`std::string::substr() ` creates a new string and performs heap allocation

### DIFF
--- a/src/network/OtaUpdater.cpp
+++ b/src/network/OtaUpdater.cpp
@@ -3,7 +3,6 @@
 #include <ArduinoJson.h>
 #include <HTTPClient.h>
 #include <Update.h>
-#include <WiFiClientSecure.h>
 
 namespace {
 constexpr char latestReleaseUrl[] = "https://api.github.com/repos/daveallie/crosspoint-reader/releases/latest";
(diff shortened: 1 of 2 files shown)


daveallie (approved):
Thanks for this optimisation @embedded4ever, my C++-foo isn't super strong so I appreciate the improvements

PR #219, opened 2026-01-03
Calibre Web Epub Downloading + Calibre Wireless Device Syncing

itsthisjustin:
## Summary

Adds support for browsing and downloading books from a Calibre-web server via OPDS.                         
                                                                                                              
  How it works                                                                                                
                                                                                                              
  1. Configure server URL in Settings → Calibre Web URL (e.g., https://myserver.com:port I use Cloudflare tunnel to make my server accessible anywhere fwiw)                  
  2. "Calibre Library" will now show on the the home screen                                                              
  3. Browse the catalog - navigate through categories like "By Newest", "By Author", "By Series", etc.        
  4. Download books - select a book and press Confirm to download the EPUB to your device                     
                                                                                                              
  Navigation                                                                                                  
                                                                                                              
  - Up/Down - Move through entries                                                                            
  - Confirm - Open folder or download book                                                                    
  - Back - Go to parent catalog, or exit to home if at root                                                   
  - Navigation entries show with > prefix, books show title and author                                        
  - Button hints update dynamically ("Open" for folders, "Download" for books)                                
                                                                                                              
  Technical details                                                                                           
                                                                                                              
  - Fetches OPDS catalog from {server_url}/opds                                                               
  - Parses both navigation feeds (catalog links) and acquisition feeds (downloadable books)                   
  - Maintains navigation history stack for back navigation                                                    
  - Handles absolute paths in OPDS links correctly (e.g., /books/opds/navcatalog/...)                         
  - Downloads EPUBs directly to the SD card root                                                              
                                                                                                              
  Note                                                                                                        
                                                                                                              
  The server URL should be typed to include https:// if the server requires it - HTTP→HTTPS redirects may cause SSL errors on ESP32.                                        

## Additional Context

* I also changed the home titles to use uppercase for each word and added a setting to change the size of the side margins

PR #198, opened 2026-01-01
Feature/smaller font

yingirene:
## Summary

* **What is the goal of this PR?** (e.g., Fixes a bug in the user authentication module, Implements the new feature for 
  file uploading.)

Added a smaller font size (X Small) for all 3 font families. I think I saw that Bookerly was replaced with Aleo, but those changes didn't show up when I cloned the branch and I'm not sure how to get those changes. Happy to update the PR if someone can help me with that.

* **What changes are included?**

## Additional Context

* Add any other information that might be helpful for the reviewer (e.g., performance implications, potential risks, specific areas to focus on).

PR #223, opened 2026-01-03
add settings for reader screen margin

fischer-hub:
## Summary

* This PR adds a setting to control the top left and right margins of the reader screen in 4 sizes (5, 10, 20, 40 pt?) and defaults to `SMALL` which is equivalent to the fixed margin of 5 that was already in use before.

## Additional Context

### DIFF
--- a/src/activities/settings/SettingsActivity.cpp
+++ b/src/activities/settings/SettingsActivity.cpp
@@ -27,7 +27,7 @@ const SettingInfo settingsList[settingsCount] = {
                       {"Bookerly", "Noto Sans", "Open Dyslexic"}),
     SettingInfo::Enum("Reader Font Size", &CrossPointSettings::fontSize, {"Small", "Medium", "Large", "X Large"}),
     SettingInfo::Enum("Reader Line Spacing", &CrossPointSettings::lineSpacing, {"Tight", "Normal", "Wide"}),
-    SettingInfo::Value("Reader Screen Margin", &CrossPointSettings::screenMargin, { 5, 40, 5 }),
+    SettingInfo::Value("Reader Screen Margin", &CrossPointSettings::screenMargin, {5, 40, 5}),
     SettingInfo::Enum("Reader Paragraph Alignment", &CrossPointSettings::paragraphAlignment,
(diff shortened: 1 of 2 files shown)


lukestein:
Noting this may interact with #198

src/activities/settings/SettingsActivity.cpp
```diff
@@ -34,6 +34,7 @@ const SettingInfo settingsList[settingsCount] = {
      {"Bookerly", "Noto Sans", "Open Dyslexic"}},
     {"Reader Font Size", SettingType::ENUM, &CrossPointSettings::fontSize, {"Small", "Medium", "Large", "X Large"}},
     {"Reader Line Spacing", SettingType::ENUM, &CrossPointSettings::lineSpacing, {"Tight", "Normal", "Wide"}},
+    {"Reader Screen Margin", SettingType::ENUM, &CrossPointSettings::screenMargin, {"5", "10", "20", "40"}},
```
daveallie: Yeah it might be better in this case to have the actual values of `5`, `10`, `20`, `40` be stored in the settings object instead of the enum options. It will mean it won't clash with changes to other existing enums. This is one of the first settings (besides the sleep timeout and refresh frequency) where the value is the setting.

Can you please create a new `SettingType` called `VALUE` and support setting values here. It may require some additional tweaks to `SettingInfo` to setup a few static constructors to make creating the `settingsList` a bit easier with all the different types.

itsthisjustin:
FYI I also added this in #219 haha

fischer-hub:
> FYI I also added this in #219 haha

Where? I don't see any changes related to the reading screen margin in your PR, maybe Im missing something?

itsthisjustin:
> > FYI I also added this in #219 haha
> 
> Where? I don't see any changes related to the reading screen margin in your PR, maybe Im missing something?

Dave asked me to remove them for a cleaner PR

daveallie:
Thanks for this, given the continued conflicts from this and other's PR as there are a lot of settings changes coming through and the structural changes to setting setup, I've stepped into resolve the conflict and make a few small tweaks to get it merged.

PR #225, opened 2026-01-03
Feature/cover crop mode

jonasdiemer:
Added a setting to select `fit` or `crop` for cover image on sleep screen.

Might add a `expand` feature in the future that does not crop but rather fills the blank space with a mirror of the image.

### DIFF
--- a/src/CrossPointSettings.cpp
+++ b/src/CrossPointSettings.cpp
@@ -12,7 +12,7 @@ CrossPointSettings CrossPointSettings::instance;
 namespace {
 constexpr uint8_t SETTINGS_FILE_VERSION = 1;
 // Increment this when adding new persisted settings fields
-constexpr uint8_t SETTINGS_COUNT = 14;
+constexpr uint8_t SETTINGS_COUNT = 15;
 constexpr char SETTINGS_FILE[] = "/.crosspoint/settings.bin";
 }  // namespace
 
@@ -40,6 +40,7 @@ bool CrossPointSettings::saveToFile() const {
   serialization::writePod(outputFile, paragraphAlignment);
   serialization::writePod(outputFile, sleepTimeout);
   serialization::writePod(outputFile, refreshFrequency);
+  serialization::writePod(outputFile, screenMargin);
   serialization::writePod(outputFile, sleepScreenCoverMode);
   outputFile.close();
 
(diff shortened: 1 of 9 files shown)


itsthisjustin:
Thank you! I was going to add this next myself cause it was bugging me

lib/GfxRenderer/Bitmap.cpp
```diff
@@ -257,6 +257,7 @@ BmpReaderError Bitmap::parseHeaders() {
 }
 
 // packed 2bpp output, 0 = black, 1 = dark gray, 2 = light gray, 3 = white
+// TODO: This seems to only work sequentially, rowY is mostly ignored
```
daveallie: Yeah rowY can probably be removed, it's used for quantization, but giving we're tracking `lastRowY` this can probably be simplified a bit by just tracking current row Y instead

jonasdiemer: Yup, lost some time here ("why is rowY not working"). Refactored it (I think - the case for non-linear was never triggered I hope).

daveallie (approved):
Happy to merge this as is, but worth noting that swapping this setting does not clear the existing covers and can introduce some strange dithering artefacts.

lukestein:
> Happy to merge this as is, but worth noting that swapping this setting does not clear the existing covers and can introduce some strange dithering artefacts.

Sorry to comment on a closed PR @daveallie but I just installed a dev build including this and wanted to note that the UX of this setting feels broken since the existing cover cache isn't cleared… even when the user changes the setting, it _feels_ broken since the existing book cover doesn't update.

I'm happy to update the docs to try to explain, but would suggest that before a new release with this setting, there would ideally be a way to clear the cover cache.

jonasdiemer:
Thanks, @lukestein - see #330

lukestein:
> Thanks, @lukestein - see #330 

Awesome @jonasdiemer! Thanks!! 🙏

itsthisjustin:
Fwiw this still isn't really doing anything for me. I think I'd be expecting a "Fill" option? Fit and Crop do the same thing for the book I'm reading. File here if you want to test: https://files.sofriendly.com/l5LMQxhE

jonasdiemer:
> Fwiw this still isn't really doing anything for me. I think I'd be expecting a "Fill" option? Fit and Crop do the same thing for the book I'm reading. File here if you want to test: https://files.sofriendly.com/l5LMQxhE

@itsthisjustin this is an issue with cover images that are smaller than the screen resolution - we currently don't scale up. Should be added/changed though IMO (I may look into it).

lukestein:
> an issue with cover images that are smaller than the screen resolution - we currently don't scale up. Should be added/changed though IMO (I may look into it).

Just updated my PR on the user guide #337 to clarify that these modes only downscale, not upscale

PR #237, opened 2026-01-04
Rename items, Delete and Remove multiple items at once via Web UI

JakeLDev:
## Summary

- More improvements to user experience when managing files and directories via the web UI

- Add checkboxes next to each item to perform actions on multiple selected items at once.
  - Currently supports deleting selected items, and moving selected items.
- Add button to rename items, works on both files, and directories
- New endpoint to fetch directory list, used to populate the "move item" dialog dropdown list

## Additional Context
- Also bumped the reference to open-x4-sdk to include the rename function introduced by [gebeto](https://github.com/gebeto)

Images of different screens and dialogs in use:
Selection checkboxes:
image

Delete multiple items (with scroll)
image

Delete progress:
image

Delete with failure:
image

Rename file (epub) (works with directory too but looks same minus extension)
image

Move items, directory dropdown:
image

Moving progress: 
image

### DIFF
--- a/src/network/CrossPointWebServer.cpp
+++ b/src/network/CrossPointWebServer.cpp
@@ -72,6 +72,7 @@ void CrossPointWebServer::begin() {
 
   server->on("/api/status", HTTP_GET, [this] { handleStatus(); });
   server->on("/api/files", HTTP_GET, [this] { handleFileListData(); });
+  server->on("/api/folders", HTTP_GET, [this] { handleFolderList(); });
 
   // Upload endpoint with special handling for multipart form data
   server->on("/upload", HTTP_POST, [this] { handleUploadPost(); }, [this] { handleUpload(); });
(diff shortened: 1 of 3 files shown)


JakeLDev:
Getting semi-frequent crashes with errors that appear to be memory related. Not sure if the html has gotten too large or something? Will keep investigating

pablohc:
@JakeLDev Is it possible to add a top bar that shows the total progress when dealing with multiple files, or a single bar that shows progress based on N items?

10 items copied = each book copied represents 10% of the total progress bar.

open-x4-sdk
gebeto, line 1: sdmanager was already updated and "rename" method was added: https://github.com/open-x4-epaper/community-sdk/pull/11

gebeto: need to update the reference

JakeLDev: Yep I believe that's what I've done isn't it? (Haven't worked with submodules before so might've done it wrong)

JakeLDev:
@pablohc That would be possible. I do like the current way of showing the progress of the current file, and I wonder if having a second bar would be confusing for users, it already does show the progress via showing something like "8/10". 
Not to mention that both moving and deleting are very quick operations, so it flashes by pretty quick, would be more significant for uploading multiple files I reckon.
I'll have a think about it 😁 👍🏻

pablohc:
Hi @JakeLDev, I understand the point... maybe moving the current progress "8/10" as the content of the bar itself will help to visualize it better without having to add an additional bar.

open-x4-sdk
gebeto: Ah, yeah looks like it is fine, ignore my comments

JakeLDev:
Been quite busy, if anyone wants to pick this up and finish it off (or re-do it), be my guest :). Otherwise I'll get around to finishing it eventually

PR #240, opened 2026-01-04
Add TXT file reader support

eunchurn:
## Summary

Add support for reading plain text (.txt) files, enabling users to browse, read, and track progress in TXT documents alongside existing EPUB and XTC formats.

- New Txt library for loading and parsing plain text files
- New TxtReaderActivity with streaming page rendering using 8KB chunks to handle large files without memory issues on ESP32-C3
- Page index caching system (index.bin) for instant re-open after sleep or app restart
- Progress bar UI during initial file indexing (matching EPUB style)
- Word wrapping with proper UTF-8 support
- Cover image support for TXT files:
  - Primary: image with same filename as TXT (e.g., book.jpg for book.txt)
  - Fallback: cover.bmp/jpg/jpeg in the same folder
  - JPG to BMP conversion using existing converter
  - Sleep screen cover mode now works with TXT files
- File browser now shows .txt files

## Additional Context

* Add any other information that might be helpful for the reviewer

* Memory constraints: The streaming approach was necessary because ESP32-C3 only has 320KB RAM. A 700KB TXT file cannot be loaded entirely into memory, so we read 8KB chunks and build a page offset index instead.

* Cache invalidation: The page index cache automatically invalidates when file size, viewport width, or lines per page changes (e.g., font size or orientation change).

* Performance: First open requires indexing (with progress bar), subsequent opens load from cache instantly.

* Cover image format: PNG is detected but not supported for conversion (no PNG decoder available). Only BMP and JPG/JPEG work.

### DIFF
--- a/src/activities/reader/TxtReaderActivity.cpp
+++ b/src/activities/reader/TxtReaderActivity.cpp
@@ -110,6 +110,8 @@ void TxtReaderActivity::loop() {
   const bool prevReleased = mappedInput.wasReleased(MappedInputManager::Button::PageBack) ||
                             mappedInput.wasReleased(MappedInputManager::Button::Left);
   const bool nextReleased = mappedInput.wasReleased(MappedInputManager::Button::PageForward) ||
+                            (SETTINGS.shortPwrBtn == CrossPointSettings::SHORT_PWRBTN::PAGE_TURN &&
+                             mappedInput.wasReleased(MappedInputManager::Button::Power)) ||
                             mappedInput.wasReleased(MappedInputManager::Button::Right);
 
   if (!prevReleased && !nextReleased) {
@@ -227,9 +229,9 @@ void TxtReaderActivity::buildPageIndex() {
       pageOffsets.push_back(offset);
     }
 
-    // Update progress bar every 2%
+    // Update progress bar every 10% (matching EpubReaderActivity logic)
     int progressPercent = (offset * 100) / fileSize;
-    if (progressPercent != lastProgressPercent && progressPercent % 2 == 0) {
+    if (lastProgressPercent / 10 != progressPercent / 10) {
       lastProgressPercent = progressPercent;
 
       // Fill progress bar
@@ -272,7 +274,6 @@ bool TxtReaderActivity::loadPageAtOffset(size_t offset, std::vector<std::string>
 
   // Parse lines from buffer
   size_t pos = 0;
-  size_t bytesConsumed = 0;
 
   while (pos < chunkSize && static_cast<int>(outLines.size()) < linesPerPage) {
     // Find end of line
@@ -289,13 +290,18 @@ bool TxtReaderActivity::loadPageAtOffset(size_t offset, std::vector<std::string>
       break;
     }
 
-    // Extract line (without newline)
-    std::string line(reinterpret_cast<char*>(buffer + pos), lineEnd - pos);
+    // Calculate the actual length of line content in the buffer (excluding newline)
+    size_t lineContentLen = lineEnd - pos;
 
-    // Remove carriage return if present
-    if (!line.empty() && line.back() == '\r') {
-      line.pop_back();
-    }
+    // Check for carriage return
+    bool hasCR = (lineContentLen > 0 && buffer[pos + lineContentLen - 1] == '\r');
+    size_t displayLen = hasCR ? lineContentLen - 1 : lineContentLen;
+
+    // Extract line content for display (without CR/LF)
+    std::string line(reinterpret_cast<char*>(buffer + pos), displayLen);
+
+    // Track position within this source line (in bytes from pos)
+    size_t lineBytePos = 0;
 
     // Word wrap if needed
     while (!line.empty() && static_cast<int>(outLines.size()) < linesPerPage) {
@@ -303,6 +309,8 @@ bool TxtReaderActivity::loadPageAtOffset(size_t offset, std::vector<std::string>
 
       if (lineWidth <= viewportWidth) {
         outLines.push_back(line);
+        lineBytePos = displayLen;  // Consumed entire display content
+        line.clear();
         break;
       }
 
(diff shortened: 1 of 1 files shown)


src/activities/reader/TxtReaderActivity.cpp
```diff
@@ -0,0 +1,626 @@
...
+}
+
+bool TxtReaderActivity::loadPageIndexCache() {
+  // Cache file format:
+  // - 4 bytes: magic "TXTI"
+  // - 4 bytes: file size (to validate cache)
+  // - 4 bytes: viewport width
+  // - 4 bytes: lines per page
+  // - 4 bytes: total pages count
+  // - N * 4 bytes: page offsets (size_t stored as uint32_t)
```
daveallie: Different fonts may have different character widths, even if the lines per page stays the same, the number of pages could be larger. This cache file should be busted when the font ID changes.

src/activities/reader/TxtReaderActivity.cpp
```diff
@@ -0,0 +1,626 @@
...
+
+  // Write page offsets
+  for (size_t offset : pageOffsets) {
+    uint8_t offsetData[4];
+    offsetData[0] = offset & 0xFF;
+    offsetData[1] = (offset >> 8) & 0xFF;
+    offsetData[2] = (offset >> 16) & 0xFF;
+    offsetData[3] = (offset >> 24) & 0xFF;
+    f.write(offsetData, 4);
+  }
```
daveallie: would love to use the serialization module for this to keep it consistent, and then same with the reading back out

src/activities/reader/TxtReaderActivity.cpp
```diff
@@ -0,0 +1,626 @@
...
+#include "CrossPointSettings.h"
+#include "CrossPointState.h"
+#include "MappedInputManager.h"
+#include "ScreenComponents.h"
+#include "fontIds.h"
+
+namespace {
+constexpr unsigned long goHomeMs = 1000;
+constexpr int topPadding = 10;
+constexpr int horizontalPadding = 15;
```
daveallie: This is now a setting and show be read from there, and also bust the cache

src/activities/reader/TxtReaderActivity.cpp
```diff
@@ -0,0 +1,626 @@
...
+
+  const int lineHeight = renderer.getLineHeight(SETTINGS.getReaderFontId());
+
+  int y = orientedMarginTop;
+  for (const auto& line : currentPageLines) {
+    if (!line.empty()) {
+      renderer.drawText(SETTINGS.getReaderFontId(), orientedMarginLeft, y, line.c_str());
+    }
+    y += lineHeight;
+  }
```
daveallie: This doesn't respect text alignment settings, or deal with the grayscale pass. The former can possibly be added in a follow up, but the latter should be included here.

src/activities/reader/TxtReaderActivity.cpp
eunchurn: I've added `cachedFontId` to the cache file header and validation logic. When the font ID changes, the cache will now be invalidated and rebuilt. The cache version has been bumped to 2 to ensure old caches are also invalidated.

src/activities/reader/TxtReaderActivity.cpp
eunchurn: Done. I've replaced all the manual byte manipulation with `serialization::readPod/writePod` calls from the Serialization module. This keeps it consistent with how other parts of the codebase handle binary data.

src/activities/reader/TxtReaderActivity.cpp
eunchurn: I've added a `screenMargin` setting to `CrossPointSettings` and replaced the hardcoded `topPadding/horizontalPadding` constants with `SETTINGS.getScreenMargin()`. The margin value is also stored in the cache header, so changing the margin setting will invalidate the cache and trigger a rebuild.

src/activities/reader/TxtReaderActivity.cpp
eunchurn: Both are now implemented. For text alignment, I've added support for `LEFT_ALIGN`, `CENTER_ALIGN`, `RIGHT_ALIGN`, and `JUSTIFIED` (though for plain text, justified falls back to left-aligned since true justification would require word spacing adjustments). For grayscale rendering, I've added the same two-pass approach used in `EpubReaderActivity` - storing the BW buffer, rendering in `GRAYSCALE_LSB` and `GRAYSCALE_MSB` modes, displaying the gray buffer, then restoring the BW buffer.

src/activities/reader/TxtReaderActivity.cpp
```diff
@@ -0,0 +1,683 @@
...
+    }
+
+    offset = nextOffset;
+    if (offset < fileSize) {
+      pageOffsets.push_back(offset);
+    }
+
+    // Update progress bar every 2%
+    int progressPercent = (offset * 100) / fileSize;
+    if (progressPercent != lastProgressPercent && progressPercent % 2 == 0) {
```
daveallie: This has a similar issue as the progress bar had in the epub parser, it'll only trigger if the percentage falls on a even number, if it goes from 1% -> 3% -> 5% etc, it'll never trigger, would be good to match the logic from the epub one where we check `lastProgress / 10 != progress / 10` to determine rendering.

src/activities/reader/TxtReaderActivity.cpp
```diff
@@ -0,0 +1,683 @@
...
+  if (pagesUntilFullRefresh <= 1) {
+    renderer.displayBuffer(EInkDisplay::HALF_REFRESH);
+    pagesUntilFullRefresh = SETTINGS.getRefreshFrequency();
+  } else {
+    renderer.displayBuffer();
+    pagesUntilFullRefresh--;
+  }
+
+  // Save BW buffer for restoration after grayscale pass
+  renderer.storeBwBuffer();
```
daveallie: This is a bug in the other implementation too, but the store and restore of the BW buffer only need to happen if doing gray passes

src/activities/reader/TxtReaderActivity.cpp
```diff
@@ -0,0 +1,683 @@
...
+  // Short press BACK goes to file selection
+  if (mappedInput.wasReleased(MappedInputManager::Button::Back) && mappedInput.getHeldTime() < goHomeMs) {
+    onGoBack();
+    return;
+  }
+
+  const bool prevReleased = mappedInput.wasReleased(MappedInputManager::Button::PageBack) ||
+                            mappedInput.wasReleased(MappedInputManager::Button::Left);
+  const bool nextReleased = mappedInput.wasReleased(MappedInputManager::Button::PageForward) ||
+                            mappedInput.wasReleased(MappedInputManager::Button::Right);
```
daveallie, line 115: nit: we should respect the new power button next setting

src/activities/reader/TxtReaderActivity.cpp
```diff
@@ -0,0 +1,683 @@
...
+    return false;
+  }
+
+  int32_t margin;
+  serialization::readPod(f, margin);
+  if (margin != cachedScreenMargin) {
+    Serial.printf("[%lu] [TRS] Cache screen margin mismatch, rebuilding\n", millis());
+    f.close();
+    return false;
+  }
```
daveallie, line 645: This will only change if `viewportWidth` and `linesPerPage` are changing. It probably doesn't need to be included. Doesn't cost much to leave in though.

daveallie:
I haven't dipped deeply into the functionality yet, but [The Letter.txt](https://github.com/user-attachments/files/24561065/The.Letter.txt) seems not to work, there are repeated sections and the end of the file can't be reached.

PR #241, opened 2026-01-04
Allow disabling anti-aliasing via a setting

khromov:
## Summary

Fixes https://github.com/daveallie/crosspoint-reader/issues/233

## Additional Context

By disabling the Text Anti-Aliasing in the settings, you can get faster black-and-white page turns that work exactly like the stock firmware, instead of having an additional flash after rendering (see issue linked above for example).

Tested on the X4 and confirmed working.


khromov:
@daveallie I just fixed the merge conflicts in this branch, please let me know if you would like me to do it again in the future.

PR #248, opened 2026-01-04
Add superscripts/subscripts to fonts and add Nix devshell

cor:
I was reading an epub that contained superscripts that weren't being rendered, so I made this PR.

This adds:
1. Superscripts and subscripts to the built-in fonts
2. A reproducible Nix development shell

I generated apps for all of the scripts because I wasn't sure which ones are relevant in the flow.

I flashed this firmware myself and it was working well.

NOTE: This does _not_ introduce fully reproducible builds, as wrapping PlatformIO in Nix is outside of my skillset. Instead it introduces a reproducible environment in which platformIO can be ran on Nix machines, which _should_ be reasonably reproducible for now.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -127,19 +127,9 @@
               buildScript
               pythonEnv
 
-              # C/C++ language server
               pkgs.clang-tools
-
-              # Python language server
               pkgs.python3Packages.python-lsp-server
-
-              # Nix language server
               pkgs.nixd
-
-              # Git for submodules
-              pkgs.git
-
-              # Useful utilities
               pkgs.esptool
             ];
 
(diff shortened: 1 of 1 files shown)


patrolek:
Most of my books contain superscripts, is it possible to know, when this futere will be added to Crosspoint? I'm not a Guthub guy, so if this not a place for such a comment please forgive me and delete it : ))

daveallie:
Cherry-picked the relevant commit into https://github.com/crosspoint-reader/crosspoint-reader/pull/463, and updated the fonts there. Will close this one out as the only other thing in it is the Nix devshell which I don't want to merge.

PR #252, opened 2026-01-05
fix: resolve early boot heap corruption via local static entities

Wferr:
Was working on a addition ([TRMNL](https://github.com/usetrmnl/byos_hanami/blob/main/doc/api.adoc) sleep mode that will download/refresh background images) and was running into problems/crashes unexpectedly, Let gemini guide me through digging into/fixing and got down to this, Not sure if this was the true cause, But it at least makes the [heap_caps_check_integrity_all](https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-reference/system/heap_debug.html#:~:text=It%20is%20also,corrupt%20heap%20structures.) check pass. And my feature seems to be working as expected now. 

I looked for the "entities_ww2.epub" to test opening to see if anything is broken but couldnt find it on the internet. A .epub file I have worked fine.

AI-GEND Desc:

## Fix: Boot Heap Corruption

This PR resolves a heap corruption crash during early boot on ESP32.

### Changes
- **Local Static Entities**: Moved the HTML entity lookup map to a local static. This prevents heavy heap allocations during global initialization before the memory manager is stable.
- **Heap Check UI**: Added an integrity check to setup(). If corruption is detected, it now initializes the display and shows a "Heap corruption" error.

### Result
Fixes the boot crash and adds robust error reporting to the screen.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -258,7 +258,8 @@ void setup() {
     Serial.println("!!! HEAP CORRUPTION DETECTED AT BOOT !!!");
     setupDisplayAndFonts();
     exitActivity();
-    enterNewActivity(new FullScreenMessageActivity(renderer, mappedInputManager, "Heap corruption", EpdFontFamily::BOLD));
+    enterNewActivity(
+        new FullScreenMessageActivity(renderer, mappedInputManager, "Heap corruption", EpdFontFamily::BOLD));
     return;
   }
 


daveallie:
Hey @Wferr, I'll take a look at these changes as they stand, but as mentioned in https://github.com/daveallie/crosspoint-reader/discussions/244#discussioncomment-15414236, I will be unlikely to accept a TRMNL style sleep screen with remote data fetching as you suggested you're working on in the body of this PR.

Not to discourage you from working on cool things, I just didn't want you to put a bunch of work into it and assume it would make it into this repo. The beauty of open source however, is that you're free to choose your own adventure.

daveallie (commented):
I don't love moving the unordered map to a static local variable. However I do think we can fix this another way by changing to map to a vector and moving it out of the heap into the flash. I'll put a change to get that sorted.

daveallie:
Turns out when expat was added, this whole module because useless, going to remove it for now here: https://github.com/daveallie/crosspoint-reader/pull/274. That should effectively make this PR moot

PR #253, opened 2026-01-05
Move battery status on home screen to top left

jonasdiemer:
So it doesn't look so lost on a row on its own.

Also sligthly (1px) moved symbol in on reader view.

### DIFF
--- a/src/activities/home/HomeActivity.cpp
+++ b/src/activities/home/HomeActivity.cpp
@@ -316,7 +316,8 @@ void HomeActivity::render() const {
   const auto labels = mappedInput.mapLabels("", "Confirm", "Up", "Down");
   renderer.drawButtonHints(UI_10_FONT_ID, labels.btn1, labels.btn2, labels.btn3, labels.btn4);
 
-  ScreenComponents::drawBattery(renderer, 20, 26);
+  const auto batteryX = pageWidth - 25 - renderer.getTextWidth(SMALL_FONT_ID, "100 %");
+  ScreenComponents::drawBattery(renderer, batteryX, 10);
 
   renderer.displayBuffer();
 }


pablohc:
Hi @jonasdiemer,

standard buttons
inverted buttons

Please note that in the standard view the "Back/Home" button does not have a label and there is a lot of space for the battery icon and percentage to be visible, but by mistake I tried inverting the buttons, where it replaced "Back" with "Up" and the battery icon and the button label overlapped. I don't use that button arrangement but anyone who does will be bothered.

jonasdiemer:
Oh, you're right, thanks for pointing this out, @pablohc. 

Maybe we want to move it to the title bar then... (recent commit)

pablohc:
I would like to make a design proposal, but I don't know how to obtain the interface currently. Is there any guide to reproduce it? Maybe I'll sketch it in some vector app.

daveallie:
The UI is ✨hand crafted✨ so there's no existing design doc or artifact. If you'd like a visual tool aide to help propose some changes, probably best way is a basic mock up in the likes of Excalidraw or Figma

jlaunay:
> Oh, you're right, thanks for pointing this out, @pablohc.
> 
> Maybe we want to move it to the title bar then... (recent commit)

Yes, title bar (top right) would be great 👍

pablohc:
I have tried to differentiate the labels of the buttons from the menus, to have a little more space for the battery.
I will try to make another proposal for the different screens in the next few days.

X4 UI - Home Menu

daveallie (approved):

daveallie:
Happy to merge this as is, or did you want to cook a bit more and leverage some of @pablohc's ideas @jonasdiemer?

I think the updated button options look good, and the battery indicator can definitely just hide away there (but also happy with the top of the screen too)

jonasdiemer:
The screens shown by @pablohc are mocks, this PR has the battery at top left, but I agree with @jlaunay that top right would make more sense. Will update it to move it top right, but would leave further tuning to future work.

jonasdiemer:
Done. May want to make it truly right-aligned in the future.

jlaunay:
> The screens shown by @pablohc are mocks, this PR has the battery at top left, but I agree with @jlaunay that top right would make more sense. Will update it to move it top right, but would leave further tuning to future work.

This way we keep controls (bottom) separate from informations (top) 👌

PR #254, opened 2026-01-05
Glyph compression

daveallie:
## Summary

- Glyph compression using zlib conditionally if the compressed data is at least smaller than 90% of the original
  - It is possible for small glyphs to get bigger when using DEFLATE, and this seems to work best on the larger fonts
- Inflate the compressed glyphs at render time

## Additional details

- Reduces flash size from 5,624,290 (85.8%) to 4,730,684 (72.2%), a 15.8% reduction
- For bookerly 14pt, this seems to incur a cost of about 300ms per full page render of text
  - Went from about 700ms to do both BW and gray pass to ~1000ms
  - I believe I can lower this further by using a small LRU cache in the font to hold often used glyph data (perhaps we also only ever use the cache for compressed glyphs too)

### DIFF
--- a/lib/GfxRenderer/GfxRenderer.cpp
+++ b/lib/GfxRenderer/GfxRenderer.cpp
@@ -4,24 +4,22 @@
 #include <miniz.h>
 
 namespace {
+tinfl_decompressor* inflator = nullptr;
+
 bool inflateOneShot(const uint8_t* inputBuf, const size_t deflatedSize, uint8_t* outputBuf, const size_t inflatedSize) {
-  // Setup inflator
-  const auto inflator = static_cast<tinfl_decompressor*>(malloc(sizeof(tinfl_decompressor)));
   if (!inflator) {
-    Serial.printf("[%lu] [ZIP] Failed to allocate memory for inflator\n", millis());
+    Serial.printf("[%lu] [GFX] Inflator not initialized\n", millis());
     return false;
   }
-  memset(inflator, 0, sizeof(tinfl_decompressor));
-  tinfl_init(inflator);
 
   size_t inBytes = deflatedSize;
   size_t outBytes = inflatedSize;
+  tinfl_init(inflator);
   const tinfl_status status = tinfl_decompress(inflator, inputBuf, &inBytes, nullptr, outputBuf, &outBytes,
                                                TINFL_FLAG_USING_NON_WRAPPING_OUTPUT_BUF);
-  free(inflator);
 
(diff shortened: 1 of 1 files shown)


jonasdiemer:
30% slowdown (in everyday page rendering) for 15% space reduction doesn't sound like a good trade off. We should rather move fonts to SPIFFS.

Instead of caching, maybe consider only compressing rare glyphs?

DJPoulter:
Is there any way to have font files on the SD card and then put into flash when they are selected in the settings? This would allow us to have more fonts/sizes.

khromov:
Agree with Jonas, page change responsivity is the main "feature" of an e-reader, it trumps more fonts for sure. If you feel this is a "must-have", perhaps it's possible to consider creating a "light" version of Crosspoint with limited font choices for people who prioritize responsiveness?

jonasdiemer:
> Is there any way to have font files on the SD card and then put into flash when they are selected in the settings? This would allow us to have more fonts/sizes.

We will do that after 1.0 when we are ready to ditch the stock os (and hence can overwrite the SPIFFS partition of the embedded flag with font cache data loaded from SD card)

daveallie:
Yeah agree that the current tradeoff is not yet worth merging this, unless that time cost gets quite low this isn't a valid path. I mostly put it together to experiment with the idea.

"rare glyphs" is a bit of an interesting one as it's likely language dependent, for example for my books all Cyrillic glyphs are "rare", and for others, other sets will be less common. 

SPIFFS is a much more reasonable alternatively (or even just expanding the main app partitions further), before doing so I'd like to explore what the impact is to the stock firmware if the SPIFFS partition is missing their expected data.

The TLDR is that this PR won't get merged in as it stands.

KraXen72:
what about zstd (z-standard) compression? from a quick search, it seems to outpreform deflate in decompression speed.

daveallie:
I believe the zstd dictionary is significantly larger (resulting in needing 128MB of RAM to even just set it up). There might be some dictionaryless one-shot decompression like we're doing with deflate, but I haven't looked into it. 

Also, given the small volume of bytes we're decompressing at a time (but doing it many many times), the throughput speed of the compression algorithm isn't the bottle here (afaict) but instead all the setup/teardown, which would likely be similar for zstd.

carpentiers:
How about Tamp?

daveallie:
Planning to reapproach this with something like Group5 using bb_paper, will have a chat to bitbank2 about it.

PR #260, opened 2026-01-06
feat: remember parent folder index in menu when ascending folders

### LINKED ISSUE #259: File browser should remember what folder you were in when going up a level
If I'm looking in directory `a/b/c`, and I ascend the tree to `a/b`, directory `./c` should be the highlighted option.

justinluque:
## Summary

Adds feature to file selection activity for better user navigation when ascending folders. The activity now remembers the index of the parent folder instead of always resetting to the first element.

I don't have any means of testing this, so if someone could test it that'd be great

Resolves #259


justinluque:
Still needs someone to check that the feature behaves correctly

PR #261, opened 2026-01-06
Refactor auto-sleep settings to use minutes instead of predefined tim…

lordbink:
…eouts

Enhances the auto-sleep timeout feature with expanded customization options, allowing users to choose from more granular timeout intervals or disable auto-sleep entirely.

- Refactored auto-sleep settings from predefined enum constants to a flexible minutes-based system
- Expanded auto-sleep timeout options from 5 choices to 8 choices: 2 min, 5 min, 10 min, 15 min, 20 min, 30 min, 60 min, and Never
- Renamed setting from "Time to Sleep" to "Auto Sleep Timeout" for clarity
- Changed internal variable from sleepTimeout to autoSleepMinutes for better semantic meaning
- Moved timeout calculation logic from CrossPointSettings.cpp to an inline method in CrossPointSettings.h for better performance
- Updated main loop to properly handle the "Never" option (timeout = 0)
- Updated README.md to document the new feature

### Additional Context
- **Backward compatibility:** The settings file format remains compatible. Existing settings will map to the closest new timeout value when loaded.
- **Default behavior:** Default timeout remains at 5 minutes (same as before), ensuring no change in user experience for existing devices.
- **Performance:** Moving the timeout calculation to an inline method eliminates function call overhead in the main loop.
- **User benefit:** The "Never" option is particularly useful for users who want manual control over sleep, and the 2-minute option is helpful for maximizing battery life.
- **Testing:** Code compiles successfully. Settings are persisted correctly across reboots.


src/activities/settings/SettingsActivity.cpp
```diff
@@ -13,29 +13,42 @@ namespace {
...
+    {"Auto Sleep Timeout",
+     SettingType::ENUM,
+     &CrossPointSettings::autoSleepMinutes,
+     {"2 min", "5 min", "10 min", "15 min", "20 min", "30 min", "60 min", "Never"}},
+    {"Refresh Frequency",
+     SettingType::ENUM,
+     &CrossPointSettings::refreshFrequency,
+     {"1 page", "5 pages", "10 pages", "15 pages", "30 pages"}},
+    {"Check for updates", SettingType::ACTION, nullptr, {}},
+};
```
daveallie, line 54: Looks like this got reverted in the merge, can you please adjust this back to the SettingInfo::* initialisers

src/main.cpp
```diff
@@ -316,16 +316,16 @@ void loop() {
...
     lastActivityTime = millis();  // Reset inactivity timer
   }
 
-  const unsigned long sleepTimeoutMs = SETTINGS.getSleepTimeoutMs();
-  if (millis() - lastActivityTime >= sleepTimeoutMs) {
-    Serial.printf("[%lu] [SLP] Auto-sleep triggered after %lu ms of inactivity\n", millis(), sleepTimeoutMs);
+  // Check auto-sleep timeout (if enabled - 0 means never sleep)
+  const unsigned long autoSleepTimeout = SETTINGS.getAutoSleepTimeoutMs();
+  if (autoSleepTimeout > 0 && millis() - lastActivityTime >= autoSleepTimeout) {
+    Serial.printf("[%lu] [SLP] Auto-sleep triggered after %lu ms of inactivity\n", millis(), autoSleepTimeout);
```
daveallie, line 336: Prevent auto-sleep should still be respected here

daveallie:
Both pieces of feedback are still relevant @lordbink, I'll wait until those are addressed before reviewing again.

daveallie:
Been about a month since the last activity here @lordbink, going to close this one out, feel free to reopen it when you come back to it.

PR #266, opened 2026-01-06
New Sleep Screens

carpentiers:
## New Sleep Screens

* **Implements new sleep screens** 
  - "Inverted Cover" - Book cover with inverted colors
  - "Light Page" - Last page read (like a book left open)
  - "Dark Page" - Last page read (like a book left open)
 
Inverted/dark colors make it easy to spot that the device is on sleep .
Light and dark page screens allow the user to continue reading even when the device is on sleep.

### DIFF
--- a/src/activities/boot_sleep/SleepActivity.cpp
+++ b/src/activities/boot_sleep/SleepActivity.cpp
@@ -45,11 +45,11 @@ void SleepActivity::onEnter() {
   }
 
   if (SETTINGS.sleepScreen == CrossPointSettings::SLEEP_SCREEN_MODE::LIGHT_PAGE) {
-    return renderCoverSleepScreen();
+    return renderPageSleepScreen();
   }
 
   if (SETTINGS.sleepScreen == CrossPointSettings::SLEEP_SCREEN_MODE::DARK_PAGE) {
-    return renderCoverSleepScreen();
+    return renderPageSleepScreen();
   }
 
   renderDefaultSleepScreen();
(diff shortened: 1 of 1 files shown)


carpentiers:
"Light Page" and "Dark Page" are highly likely not well implemented as they just call onContinueReading() then put the device to sleep.

They surely need a better way to render the last page read.
It would be best if it didn't display the bottom bar, just the page text, with maybe a small sized message saying the device is on sleep.

daveallie (commented):
I really don't love using something like `onContinueReading()` to fire back up the reader activity before sleep.

On the whole I guess I don't fully understand the value of the page sleep screens, if you want to read, the device should be unlocked, if it's sleeping, there should be a very obvious visual indicator to signal that. Leaving the user with a screen which looks like the book screen, but the device in a state where no buttons (except the power button) will be response is confusing.

Additionally inverted book covers also seem a little perplexing. Could you share some examples of when this would be useful?

carpentiers:
Good questions. 

The "Page" sleep screens allow users who need to leave a page open for reference for a long time to do it with lower energy consumption and without losing their page because of the device suddenly entering sleep mode. The "Dark Page" screen is just to make it obvious the device is on sleep, however it could be removed if "Light Page" makes it clear. As pointed out in my comment, I don't like my implementation either and think it should have its own function which could show the last page read but also not display the usual bottom bar and add a non intrusive "on sleep" message.

The "Inverted Cover" sleep screen isn't really useful. It is just for aesthetic purpose or the users to know the device is on sleep and not actively looking at a book cover. It becomes more useful if a browsing cover books function gets added someday.

Note: I guess if "Dark Page" and "Inverted Cover" are not useful and just for aesthetic purpose, we could just have a dark mode instead.

daveallie:
I haven't run power consumption benchmarks, but my preference is to encourage people to adjust the auto sleep setting if they're reviewing reference material rather than repurposing the sleep screen. While sleeping should barely even sip power, idling without page turns should be reasonable low on power consumption as well. Majority of power usage will be driving the display changes (there is some baseline power to the display rail while the device is on also), which if someone is just looking at one page, won't be a big issue.

Going to close this one as the complexity to render book pages from the sleep screen is a higher than its benefit. Thank you for the thought and contribution though.

PR #269, opened 2026-01-06
Update User Guide to reflect release 0.12.0

lukestein:
## Summary

* Update the User Guide per current firmware release.
* Make some formatting improvements including for readability and consistency.

### DIFF
--- a/USER_GUIDE.md
+++ b/USER_GUIDE.md
@@ -7,10 +7,10 @@ Welcome to the **CrossPoint** firmware. This guide outlines the hardware control
 The device utilises the standard buttons on the Xtink X4 (in the same layout as the manufacturer firmware, by default):
 
 ### Button Layout
-| Location        | Buttons                                    |
-|-----------------|--------------------------------------------|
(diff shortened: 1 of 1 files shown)


pablohc:
Hi @lukestein, please add Reset button to "Button Layout" in User Guide. Some users don't know it exists or how to use it.

lukestein:
> Hi @lukestein, please add Reset button to "Button Layout" in User Guide. Some users don't know it exists or how to use it.

Good tip @pablohc and I'm happy to update the PR. But I'm not exactly sure what information about how to use the reset button might be useful.

I think I've only ever used it as a quick press, followed by a long hold on the power button to restart?

pablohc:
> Good tip @pablohc and I'm happy to update the PR. But I'm not exactly sure what information about how to use the reset button might be useful.
> 
> I think I've only ever used it as a quick press, followed by a long hold on the power button to restart?

I was only referring to the button table that does not mention the "Reset":

image

Regarding the behavior, it is just as you describe:
Press and release the "Reset" button, and then hold the "Power" button for a few seconds.

daveallie (approved):

PR #275, opened 2026-01-07
Support up to 500 character file names

### LINKED ISSUE #265: New epubs added to SD card do not show up in Browse Files menu
### Affected Version

0.12.0

### Describe the Bug

I added four files in one go using the personal hotspot feature and the first two appeared on in the browse files list and the other two did not. I repeatedly tried to add the other two and got "success", but I can't see the files in the browse files list. I pulled the SD card out and put it in my computer and I can see the files on my computer. I also added more epubs to the SD card from my computer and then put it back into my X4 and I still cannot see any of the added files besides those first two.

Am I doing something wrong? I took a look at the user guide and there's no mention of needing to manually refresh the file list or reset the device after adding files.

### Steps to Reproduce

1. add four files at once through personal hotspot
2. see that only two are added and two are missing
3. pull out sd card and put it in computer and see all four files
4. add more files to the sd card
5. put the sd card back in the x4 and see that none of the new files are in the browse files list

### Expected Behavior

I expected to see the files i uploaded to the device in the browse files list

Image

Image

```shell
Please note that the files "The Upcycled Self" and "Wayfarers 04" appear on the SD card but not in the browse files list. This screenshot was taken before I attempted to add files directly to the SD card and they also didn't appear on the X4.
```

daveallie:
## Summary

- Support up to 500 character file names

## Additional Context

- Fixes #265

### DIFF
--- a/src/activities/reader/FileSelectionActivity.cpp
+++ b/src/activities/reader/FileSelectionActivity.cpp
@@ -123,6 +123,7 @@ void FileSelectionActivity::loop() {
     if (files[selectorIndex].back() == '/') {
       basepath += files[selectorIndex].substr(0, files[selectorIndex].length() - 1);
       loadFiles();
+      selectorIndex = 0;
       updateRequired = true;
     } else {
       onSelect(basepath + files[selectorIndex]);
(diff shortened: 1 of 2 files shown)


swwilshub:
With escaping for special characters can these exceed 512 bytes for 

```
char output[512];  // Stack buffer - UNSAFE!
constexpr size_t outputSize = sizeof(output);
```

daveallie:
Could you explain more about this issue @swwilshub? "Character" probably want's the right word to use in the title, it should really be "byte", we're allocating a 500 byte array for the title and then calling `file.getName` on the file, from the docs of `getName`:
```cpp
  /**
   * Get a file's name followed by a zero byte.
   *
   * \param[out] name An array of characters for the file's name.
   * \param[in] len The size of the array in bytes. The array
   *             must be at least 13 bytes long.  The file's name will be
   *             truncated if the file's name is too long.
   * \return The length of the returned string.
   */
```

So for files with names longer than 500 bytes, we'd just get a truncated name as before. These changes don't fully solve the issue, but extending it to 500 bytes seems like an acceptable limit to me atm.

PR #286, opened 2026-01-08
Add page turn on power button press

fischer-hub:
## Summary

* This PR adds a setting to (additionally) map the forward page turn onto the powerbutton when in `EPUBReaderActivity` and powerbutton short press is not mapped to sleep mode. I find the powerbutton to be exactly where my thumb is while reading so it is very convenient to map the forwardpage turn to that. Maybe Im not alone with this ^^

## Additional Context

### DIFF
--- a/src/activities/reader/EpubReaderActivity.cpp
+++ b/src/activities/reader/EpubReaderActivity.cpp
@@ -151,11 +151,10 @@ void EpubReaderActivity::loop() {
 
   const bool prevReleased = mappedInput.wasReleased(MappedInputManager::Button::PageBack) ||
                             mappedInput.wasReleased(MappedInputManager::Button::Left);
-  const bool nextReleased =
-      mappedInput.wasReleased(MappedInputManager::Button::PageForward) ||
-      (SETTINGS.shortPwrBtn == CrossPointSettings::SHORT_PWRBTN::PAGE_TURN &&
-       mappedInput.wasReleased(MappedInputManager::Button::Power) && mappedInput.getHeldTime() < goHomeMs) ||
-      mappedInput.wasReleased(MappedInputManager::Button::Right);
+  const bool nextReleased = mappedInput.wasReleased(MappedInputManager::Button::PageForward) ||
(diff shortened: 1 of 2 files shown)


lukestein:
I think this is a great feature idea and I will use it!

One small implementation suggestion in terms of the settings: Rather than have two settings (`Short Power Button Click` and `Power Button Page Turn`), I would suggest combining into a single setting (`Short Power Button Click`) with the following three options:

1. **Power off** (equivalent of current `ON` setting)
2. **Ignore** (equivalent of current `OFF` setting)
3. **Next page** (new setting)

fischer-hub:
> One small implementation suggestion in terms of the settings: Rather than have two settings (`Short Power Button Click` and `Power Button Page Turn`), I would suggest combining into a single setting (`Short Power Button Click`) with the following three options:

I had the same idea today and for some reason ... forgot about it again, thx for reminding me haha!

src/activities/reader/EpubReaderActivity.cpp
```diff
@@ -151,8 +151,11 @@ void EpubReaderActivity::loop() {
 
   const bool prevReleased = mappedInput.wasReleased(MappedInputManager::Button::PageBack) ||
                             mappedInput.wasReleased(MappedInputManager::Button::Left);
-  const bool nextReleased = mappedInput.wasReleased(MappedInputManager::Button::PageForward) ||
-                            mappedInput.wasReleased(MappedInputManager::Button::Right);
+  const bool nextReleased =
+      mappedInput.wasReleased(MappedInputManager::Button::PageForward) ||
+      (SETTINGS.shortPwrBtn == CrossPointSettings::SHORT_PWRBTN::PAGE_TURN &&
+       mappedInput.wasReleased(MappedInputManager::Button::Power) && mappedInput.getHeldTime() < goHomeMs) ||
```
daveallie: I don't think we need the duration check. The main `loop` in `main.cpp` should pick up the power button press and send the device to sleep before this code runs so anything that reaches this point should be a valid non-sleep power button press.

daveallie (approved):

Issue #190, opened 2025-12-31
Suggestion: Show button labels at bottom of Chapter Selection (TOC) screen


lukestein:
Most screens show button labels at the bottom using the current mapping per settings.

Chapter Selection (Table of Contents) Screen is currently missing this.

I suggest it would be useful, thematically consistent, and hopefully require minimal resources since code exists in other places.

![image](https://github.com/user-attachments/assets/c33f1391-adbb-4912-93fd-f2d47b68df6f)

lukestein:
Closed by #294

daveallie:
Available in [v0.13.0](https://github.com/daveallie/crosspoint-reader/releases/tag/0.13.0)

Sunspark-007:
The button labels don't change according to orientation.. e.g. Landscape CCW.

PR #294, opened 2026-01-09
Add navigation hints to ChapterSelectionActivities

selunders:
## Summary

Add navigation hints to Chapter Select - #190 

### Before

Mi 11X_20260108_214114_lmc_8 4

### After
Mi 11X_20260108_213803_lmc_8 4


maeveynot:
This doesn't look right in landscape orientation. Opening an issue to track...

PR #300, opened 2026-01-09
Ensure new custom sleep image every time

duh102:
When picking a random sleep image from a set of custom images, compare the randomly chosen index against a cached value in settings. If the value matches, use the next image (rolling over if it's the last image). Cache the chosen image index to settings in either case.

## Summary

Implements a tweak on the custom sleep image feature that ensures that the user gets a new image every time the device goes to sleep.
This change adds a new setting (perhaps there's a better place to cache this?) that stores the most recently used file index. During picking the random image index, we compare this against the random index and choose the next one (modulo the number of image files) if it matches, ensuring we get a new image.

## Additional Context

As mentioned, I used settings to cache this value since it is a persisted store, perhaps that's overkill. Open to suggestions on if there's a better way.

### DIFF
--- a/src/CrossPointState.cpp
+++ b/src/CrossPointState.cpp
@@ -5,7 +5,7 @@
 #include <Serialization.h>
 
 namespace {
-constexpr uint8_t STATE_FILE_VERSION = 1;
+constexpr uint8_t STATE_FILE_VERSION = 2;
 constexpr char STATE_FILE[] = "/.crosspoint/state.bin";
 }  // namespace
 
@@ -19,6 +19,7 @@ bool CrossPointState::saveToFile() const {
 
   serialization::writePod(outputFile, STATE_FILE_VERSION);
   serialization::writeString(outputFile, openEpubPath);
+  serialization::writePod(outputFile, lastSleepImage);
   outputFile.close();
   return true;
 }
@@ -31,13 +32,18 @@ bool CrossPointState::loadFromFile() {
 
   uint8_t version;
   serialization::readPod(inputFile, version);
-  if (version != STATE_FILE_VERSION) {
+  if (version > STATE_FILE_VERSION) {
     Serial.printf("[%lu] [CPS] Deserialization failed: Unknown version %u\n", millis(), version);
     inputFile.close();
     return false;
   }
 
   serialization::readString(inputFile, openEpubPath);
+  if (version >= 2) {
+    serialization::readPod(inputFile, lastSleepImage);
+  } else {
+    lastSleepImage = 0;
+  }
 
   inputFile.close();
   return true;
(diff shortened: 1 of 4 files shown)


src/activities/boot_sleep/SleepActivity.cpp
```diff
@@ -81,7 +81,13 @@ void SleepActivity::renderCustomSleepScreen() const {
     if (numFiles > 0) {
       // Generate a random number between 1 and numFiles
       const auto randomFileIndex = random(numFiles);
-      const auto filename = "/sleep/" + files[randomFileIndex];
+      // If we re-generated the same index as the last sleep screen,
+      //   use the next one (modulo the number of files we have)
+      const auto realFileIndex =
+          randomFileIndex == SETTINGS.lastUsedSleep ? randomFileIndex + 1 % numFiles : randomFileIndex;
+      SETTINGS.lastUsedSleep = realFileIndex;
+      SETTINGS.saveToFile();
```
daveallie: Would prefer this existed in `CrossPointState` instead of the settings.

src/activities/boot_sleep/SleepActivity.cpp
```diff
@@ -81,7 +81,13 @@ void SleepActivity::renderCustomSleepScreen() const {
     if (numFiles > 0) {
       // Generate a random number between 1 and numFiles
       const auto randomFileIndex = random(numFiles);
-      const auto filename = "/sleep/" + files[randomFileIndex];
+      // If we re-generated the same index as the last sleep screen,
+      //   use the next one (modulo the number of files we have)
+      const auto realFileIndex =
+          randomFileIndex == SETTINGS.lastUsedSleep ? randomFileIndex + 1 % numFiles : randomFileIndex;
```
daveallie: A slightly better way to do this would be:

```cpp
// Generate a random number between 1 and numFiles
auto randomFileIndex = random(numFiles);
// If we re-generated the same index as the last sleep screen,
//   find a new one
while (numFiles > 1 && randomFileIndex == SETTINGS.lastUsedSleep) {
  randomFileIndex = random(numFiles);
}
```

no need for both `randomFileIndex` and `realFileIndex`

src/activities/boot_sleep/SleepActivity.cpp
duh102: How do you prefer to resolve serialization issues? I noticed in Settings that it's basically based on the length of elements present in the file, but so far the State seems only to have the open epub. Should I bump the version and do it based on that, or follow a similar pattern?

src/CrossPointState.cpp
```diff
@@ -31,13 +32,18 @@ bool CrossPointState::loadFromFile() {
...
     inputFile.close();
     return false;
   }
 
   serialization::readString(inputFile, openEpubPath);
+  if (version >= 2) {
+    serialization::readPod(inputFile, lastSleepImage);
+  } else {
+    lastSleepImage = 0;
+  }
```
duh102, line 46: This is probably more safety than needed, depending on how you want to handle serialization differences

daveallie (approved):

PR #302, opened 2026-01-09
Fix BMP rendering gamma/brightness

jonasdiemer:
1.  Refactor Bitmap.cpp/h to expose the options for FloydSteinberg and brightness/gamma correction at runtime
2. Fine-tune the thresholds for Floyd Steiberg and simple quantization to better match the display's colors

Turns out that 2 is enough to make the images render properly, so the brightness boost and gamma adjustment doesn't seem necessary currently (at least for my test image).

### DIFF
--- a/lib/GfxRenderer/Bitmap.cpp
+++ b/lib/GfxRenderer/Bitmap.cpp
@@ -14,6 +14,9 @@ constexpr bool USE_ATKINSON = true;  // Use Atkinson dithering instead of Floyd-
 Bitmap::~Bitmap() {
   delete[] errorCurRow;
   delete[] errorNextRow;
+
+  delete atkinsonDitherer;
+  delete fsDitherer;
 }
 
 uint16_t Bitmap::readLE16(FsFile& f) {
@@ -168,7 +171,7 @@ BmpReaderError Bitmap::readNextRow(uint8_t* data, uint8_t* rowBuffer) const {
     if (atkinsonDitherer) {
       color = atkinsonDitherer->processPixel(adjustPixel(lum), currentX);
     } else if (fsDitherer) {
-      color = fsDitherer->processPixel(adjustPixel(lum), currentX, fsDitherer->isReverseRow());
+      color = fsDitherer->processPixel(adjustPixel(lum), currentX);
     } else {
       if (bpp > 2) {
         // Simple quantization or noise dithering
--- a/lib/GfxRenderer/BitmapHelpers.h
+++ b/lib/GfxRenderer/BitmapHelpers.h
@@ -134,7 +134,7 @@ class FloydSteinbergDitherer {
 
   // Process a single pixel and return quantized 2-bit value
   // x is the logical x position (0 to width-1), direction handled internally
-  uint8_t processPixel(int gray, int x, bool reverseDirection) {
+  uint8_t processPixel(int gray, int x) {
     // Add accumulated error to this pixel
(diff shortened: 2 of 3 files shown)


jonasdiemer:
Due to the updated quantization, I could  disable the brightness adjustments in the JPEG converter (which resulted in a brightened up generated BMP file to compensate the unfitting quantization values).

If BMPs are not regenerated from JPEG, they will be too bright.

@daveallie is there a way to invalidate the cover BMPs during update? Maybe we (also) need the "clear cache" functionality in the settings or file browser.

jonasdiemer:
Maybe we should also move to Atkinson's Dithering for BMP. We should then factor out the ditherers (and also the brightness and gamma adjustments) from `JpegToBmpConverter.cpp` into a dedicated file to be used in both BMP and JPEG processing.

jonasdiemer:
Moved the dithering and color adjustment to one place, switched to Atkinsons by default (though they look quite similar on the test image). Ready for review / merge from my side.

daveallie:
Thanks for all this @jonasdiemer, will give it a peek now

lib/GfxRenderer/Bitmap.cpp
```diff
@@ -282,12 +165,18 @@ BmpReaderError Bitmap::readNextRow(uint8_t* data, uint8_t* rowBuffer) const {
   // Helper lambda to pack 2bpp color into the output stream
   auto packPixel = [&](const uint8_t lum) {
     uint8_t color;
-    if (useFS) {
-      // Floyd-Steinberg error diffusion
-      color = quantizeFloydSteinberg(lum, currentX, width, errorCurRow, errorNextRow, false);
+    if (atkinsonDitherer) {
+      color = atkinsonDitherer->processPixel(adjustPixel(lum), currentX);
+    } else if (fsDitherer) {
+      color = fsDitherer->processPixel(adjustPixel(lum), currentX, fsDitherer->isReverseRow());
```
daveallie: Can `inReverseRow()` just be called internally from `processPixel`?

lib/GfxRenderer/Bitmap.cpp
```diff
@@ -244,13 +137,14 @@ BmpReaderError Bitmap::parseHeaders() {
...
-    errorCurRow = new int16_t[width + 2]();  // +2 for boundary handling
-    errorNextRow = new int16_t[width + 2]();
-    prevRowY = -1;
+  // Create ditherer if enabled (only for 2-bit output)
+  // Use OUTPUT dimensions for dithering (after prescaling)
+  if (bpp > 2 && dithering) {
+    if (USE_ATKINSON) {
+      atkinsonDitherer = new AtkinsonDitherer(width);
+    } else {
+      fsDitherer = new FloydSteinbergDitherer(width);
```
daveallie, line 149: Is there a memory leak here in that these ditherers aren't deleted?

jonasdiemer: Thanks, will add to DTOR.

daveallie (approved):

jonasdiemer:
> > If BMPs are not regenerated from JPEG, they will be too bright.
> 
> This is no longer true because of the other changes you made right? Looks like the cached BMP from the JPEG cover will still be used and the dithering is still happening at the JPEG -> BMP step and not the BMP -> screen step.

Yes, I think it should not be a problem (but I have not tested it again at the end).

PR #304, opened 2026-01-09
Adds option to add remaining time to complete chapter based on reader WPM

MatthijsMars:
## Summary

Adds the estimated amount of minutes to complete a chapter to the status bar. This estimate is based on the readers speed (words per minute) which can be set in the settings menu.

## Additional Context

Currently WPM can be set in a range from 150-300, with increments of 5. Not sure if this is a sensible limit but seems like a good trade-off between not pressing the toggle button too often and providing a good range. We might need a value setter that can be adjusted up or down. 

Additionally, currently the chapter name is not in the centre of the status bar (without it it would be roughly centred). It could be centred for more symmetry.

## Screenshots
Screenshot of reading experience:
time_left_in_chapter

Screenshot of settings:
settings


daveallie:
Seems like https://github.com/daveallie/crosspoint-reader/pull/309 would supersede the need for the setting wouldn't it? I don't mind duration showing, but manually entering it seems like a bit of a pain.

MatthijsMars:
@daveallie Yes we could combine them into one. Would have to think about how to get an accurate estimate of the WPM. I'll have a look at it some time this week and propose something.

znelson:
Closing due to inactivity

Issue #217, opened 2026-01-02
EPUB Rendering: long hyphenated words are cut off


aryascripts:
### Affected Version

0.11.2

### Describe the Bug

When a word is really long and hyphenated, it gets cut off instead of being on the next line when rendering in epub format.

Details:
- Version: 0.11.2
- Format: EPUB
- Font size: Medium
- Render line spacing: Normal

### Steps to Reproduce

1. Have an epub with long hyphenated words
2. Go to the page

### Expected Behavior

The hyphenated word should be cut off at the end of the line, and render on the next one (continued)

### Relevant Log Output/Screenshots

![Image](https://github.com/user-attachments/assets/eb2efeaa-2d53-45c0-97c7-2f2ba6d2586d)

lukestein:
Can confirm. More generally, it seems like the reader will not linebreak after a hyphen (and perhaps even a dash – — ?) as it should.

daveallie:
Hyphenation support is something which is still in the works, see:
- https://github.com/daveallie/crosspoint-reader/discussions/17
- https://github.com/daveallie/crosspoint-reader/pull/47

It should be addressed by v1.0 as tracked by https://github.com/daveallie/crosspoint-reader/discussions/186

lukestein:
Appears to be closed by #305 🙌

PR #305, opened 2026-01-09
feat: dict based Hyphenation

osteotek:
## Summary

* Adds (optional) Hyphenation for English, French, German, Russian languages

## Additional Context

* Included hyphenation dictionaries add approximately 280kb to the flash usage (German alone takes 200kb)
* Trie encoded dictionaries are adopted from hypher project (https://github.com/typst/hypher)
* Soft hyphens (and other explicit hyphens) take precedence over dict-based hyphenation. Overall, the hyphenation rules are quite aggressive, as I believe it makes more sense on our smaller screen.

### DIFF
--- a/src/CrossPointSettings.cpp
+++ b/src/CrossPointSettings.cpp
@@ -14,7 +14,7 @@ CrossPointSettings CrossPointSettings::instance;
 namespace {
 constexpr uint8_t SETTINGS_FILE_VERSION = 1;
 // Increment this when adding new persisted settings fields
-constexpr uint8_t SETTINGS_COUNT = 19;
+constexpr uint8_t SETTINGS_COUNT = 20;
 constexpr char SETTINGS_FILE[] = "/.crosspoint/settings.bin";
 }  // namespace
 
@@ -111,6 +111,7 @@ bool CrossPointSettings::loadFromFile() {
       strncpy(opdsServerUrl, urlStr.c_str(), sizeof(opdsServerUrl) - 1);
       opdsServerUrl[sizeof(opdsServerUrl) - 1] = '\0';
     }
+    if (++settingsRead >= fileSettingsCount) break;
     serialization::readPod(inputFile, textAntiAliasing);
     if (++settingsRead >= fileSettingsCount) break;
     serialization::readPod(inputFile, hideBatteryPercentage);
--- a/src/activities/settings/SettingsActivity.cpp
+++ b/src/activities/settings/SettingsActivity.cpp
@@ -14,7 +14,7 @@
 
 // Define the static settings list
 namespace {
-constexpr int settingsCount = 21;
+constexpr int settingsCount = 22;
 const SettingInfo settingsList[settingsCount] = {
     // Should match with SLEEP_SCREEN_MODE
     SettingInfo::Enum("Sleep Screen", &CrossPointSettings::sleepScreen, {"Dark", "Light", "Custom", "Cover", "None"}),


osteotek:
@daveallie I apologize for this monster of a PR, but I think it’s ready for review. Here are a few points:

- I couldn’t figure out how to combine the existing DP-based text layout approach with hyphenated words that can mutate the line. So, when hyphenation is enabled, the text layout uses a simple greedy approach. Looks ok in most cases, in my opinion

- Hyphenation dictionaries are copied from the [Hypher](https://github.com/typst/hypher) project. A script for converting them to header files is included. Since both projects are under the MIT license, I hope this is acceptable.

- The previous PR (https://github.com/daveallie/crosspoint-reader/pull/47) is kept in a separate branch in case we want to explore simpler heuristics-based hyphenation. However, I couldn’t achieve satisfactory results while working on it.

jonasdiemer:
Nice. Looks good at first glance. Will be testing this PR on my "production unit" (i.e. the main and only X4 that I use for reading ;) ).

jlaunay:
I have been testing with French books since today, with and without soft hyphens.
So far, nothing else to say except: good job, this is the best I have seen so far, it seems to work really well.

lukestein:
I'm testing on my x4 (with English-language books) and it's looking great so far, including successfully dealing with the issues in #217 (i.e., allowing breaking after visible hyphens and dashes). Thanks @osteotek!!)

jonasdiemer:
Yes, looking really good. All hyphenations I encountered so far made sense.

Areas for optimization:
1. Consider limiting the amount of hyphenations per page / per n lines. E.g. I had situations where 7 out of 12 lines were hyphenated. I haven't checked the code, but if possible increase penalty for each hyphenation based on the distance since the last hyphenation.
2. Consider avoiding leaving "short" parts (e.g. avoid <3 letters, such as ab-breviation)

jonasdiemer:
Noticed that when going back to a FW that does not do hyphenation, chapters cached with hyphenation support are broken. I think the `BOOK_CACHE_VERSION` may need a bump?

osteotek:
@jonasdiemer Yes, it seems that BOOK_CACHE_VERSION needs to be updated. I’ll investigate ways to reduce the number of hyphenations. Increasing the minimum suffix/prefix length to 3 should also help with that. Thanks for testing!

lib/Epub/Epub/Section.h
```diff
@@ -29,11 +29,11 @@ class Section {
...
   ~Section() = default;
   bool loadSectionFile(int fontId, float lineCompression, bool extraParagraphSpacing, uint8_t paragraphAlignment,
-                       uint16_t viewportWidth, uint16_t viewportHeight);
+                       uint16_t viewportWidth, uint16_t viewportHeight, bool hyphenationEnabled);
   bool clearCache() const;
   bool createSectionFile(int fontId, float lineCompression, bool extraParagraphSpacing, uint8_t paragraphAlignment,
                          uint16_t viewportWidth, uint16_t viewportHeight,
                          const std::function<void()>& progressSetupFn = nullptr,
-                         const std::function<void(int)>& progressFn = nullptr);
+                         const std::function<void(int)>& progressFn = nullptr, bool hyphenationEnabled = false);
```
daveallie: nit: Let's make this required and move it before the optional `std::function` arguments

lib/Epub/Epub/hyphenation/LanguageRegistry.cpp
```diff
@@ -0,0 +1,41 @@
...
+#include "generated/hyph-en.trie.h"
+#include "generated/hyph-fr.trie.h"
+#include "generated/hyph-ru.trie.h"
+
+namespace {
+
+LanguageHyphenator englishHyphenator(en_us_patterns, isLatinLetter, toLowerLatin);
+LanguageHyphenator frenchHyphenator(fr_patterns, isLatinLetter, toLowerLatin);
+LanguageHyphenator germanHyphenator(de_patterns, isLatinLetter, toLowerLatin);
+LanguageHyphenator russianHyphenator(ru_ru_patterns, isCyrillicLetter, toLowerCyrillic, 2, 2);
```
daveallie: nit: This min max is already defaulted to 2, 2 right?

lib/Epub/Epub/hyphenation/Hyphenator.cpp
```diff
@@ -0,0 +1,130 @@
...
+    primary.push_back(c);
+  }
+  if (primary.empty()) return nullptr;
+
+  return getLanguageHyphenatorForPrimaryTag(primary);
+}
+
+// Cached hyphenator instance for the current preferred language.
+const LanguageHyphenator*& cachedHyphenator() {
+  static const LanguageHyphenator* hyphenator = nullptr;
```
daveallie: Out of curiosity (mostly my ignorance around advanced C++ usage), why initialize this here, instead of as a static value in `Hyphenator.h`? `cachedHyphenator()` is only called from within this class so wouldn't a private static variable mean clearer accessing? If it's for delayed allocation of the `LanguageHyphenator` pointer that doesn't feel super useful as pointers are quite small?

lib/Epub/Epub/hyphenation/Hyphenator.cpp
```diff
@@ -0,0 +1,130 @@
...
+size_t byteOffsetForIndex(const std::vector<CodepointInfo>& cps, const size_t index) {
+  return (index < cps.size()) ? cps[index].byteOffset : (cps.empty() ? 0 : cps.back().byteOffset);
+}
+
+// Builds a vector of break information from explicit hyphen markers in the given codepoints.
+std::vector<Hyphenator::BreakInfo> buildExplicitBreakInfos(const std::vector<CodepointInfo>& cps) {
+  std::vector<Hyphenator::BreakInfo> breaks;
+
+  // Scan every codepoint looking for explicit/soft hyphen markers that are surrounded by letters.
+  for (size_t i = 1; i + 1 < cps.size(); ++i) {
```
daveallie, line 39: I realise it's just honoring the placement of the soft hypens, but I wonder if we should respect `hyphenator->minPrefix()` and `hyphenator->maxPrefix()` here?

lib/Epub/Epub/hyphenation/Hyphenator.cpp
```diff
@@ -0,0 +1,130 @@
...
+  }
+
+  if (breaks.empty()) {
+    return breaks;
+  }
+
+  // Sort by byte offset so we can deduplicate sequential markers in-place.
+  std::sort(breaks.begin(), breaks.end(), [](const Hyphenator::BreakInfo& lhs, const Hyphenator::BreakInfo& rhs) {
+    return lhs.byteOffset < rhs.byteOffset;
+  });
```
daveallie: Is this even necessary? `cps` should be pre-sorted by `byteOffset` as `collectCodepoints` (which builds `cps`) is scanning through each word's characters sequentially.

lib/Epub/Epub/hyphenation/Hyphenator.cpp
```diff
@@ -0,0 +1,130 @@
...
+  size_t writePos = 0;
+  for (size_t readPos = 1; readPos < breaks.size(); ++readPos) {
+    if (breaks[readPos].byteOffset == breaks[writePos].byteOffset) {
+      // Merge: explicit hyphen wins over soft hyphen at same offset.
+      breaks[writePos].requiresInsertedHyphen =
+          breaks[writePos].requiresInsertedHyphen || breaks[readPos].requiresInsertedHyphen;
+    } else {
+      breaks[++writePos] = breaks[readPos];
+    }
+  }
```
daveallie: not really sure how this is possible if currently there's a need for explicit bytes (soft or hard hyphens) to mark a break, how is it possible for two breaks to be created at the same byte offset?

lib/Epub/Epub/hyphenation/Hyphenator.cpp
```diff
@@ -0,0 +1,130 @@
...
+  trimTrailingFootnoteReference(cps);
+  const auto* hyphenator = cachedHyphenator();
+  const size_t minPrefix = hyphenator ? hyphenator->minPrefix() : LiangWordConfig::kDefaultMinPrefix;
+  const size_t minSuffix = hyphenator ? hyphenator->minSuffix() : LiangWordConfig::kDefaultMinSuffix;
+
+  // Explicit hyphen markers (soft or hard) take precedence over heuristic breaks.
+  auto explicitBreakInfos = buildExplicitBreakInfos(cps);
+  if (!explicitBreakInfos.empty()) {
+    return explicitBreakInfos;
+  }
```
daveallie: I this this is fine, but just calling out that if a long word has a single hard or soft hyphen, it will only ever be broken at that point. Alternative would be to recursively build additional legal break points with lower priority for the word parts, but honestly overkill here.

lib/Epub/Epub/ParsedText.cpp
```diff
@@ -25,10 +64,19 @@ void ParsedText::layoutAndExtractLines(const GfxRenderer& renderer, const int fo
...
-  const auto wordWidths = calculateWordWidths(renderer, fontId);
-  const auto lineBreakIndices = computeLineBreaks(pageWidth, spaceWidth, wordWidths);
+  auto wordWidths = calculateWordWidths(renderer, fontId);
+  std::vector<size_t> lineBreakIndices;
+  if (hyphenationEnabled) {
+    // Use greedy layout that can split words mid-loop when a hyphenated prefix fits.
+    lineBreakIndices = computeHyphenatedLineBreaks(renderer, fontId, pageWidth, spaceWidth, wordWidths);
+  } else {
+    lineBreakIndices = computeLineBreaks(pageWidth, spaceWidth, wordWidths);
+  }
```
daveallie, line 79: If there isn't a supported language (and imo even when the hyphenator is disabled) we should be supporting fallback hyphenation on the first word of the line if it exceeds the available space.

lib/Epub/Epub/hyphenation/HyphenationCommon.cpp
```diff
@@ -0,0 +1,146 @@
...
+
+uint32_t toLowerLatin(const uint32_t cp) { return toLowerLatinImpl(cp); }
+
+uint32_t toLowerCyrillic(const uint32_t cp) { return toLowerCyrillicImpl(cp); }
+
+bool isLatinLetter(const uint32_t cp) { return (cp >= 'A' && cp <= 'Z') || (cp >= 'a' && cp <= 'z'); }
+
+bool isCyrillicLetter(const uint32_t cp) { return (cp >= 0x0400 && cp <= 0x052F); }
+
+bool isAlphabetic(const uint32_t cp) { return isLatinLetter(cp) || isCyrillicLetter(cp); }
```
daveallie, line 70: I'm not sure what the impact is as I'm a mono-lingual simpleton, but what about the other Latin-1 characters https://en.wikipedia.org/wiki/ISO/IEC_8859-1 with all the accents. I'm not fully across if their codepoints are included in the French trie set, or if it makes sense to consider those codepoints like the standard Latin base one (e.g. does it make sense to break at an `a` and a `á` equally within a word?

lib/Epub/Epub/ParsedText.cpp
```diff
@@ -140,6 +182,136 @@ std::vector<size_t> ParsedText::computeLineBreaks(const int pageWidth, const int
...
+    const bool needsHyphen = info.requiresInsertedHyphen;
+    const int prefixWidth = measureWordWidth(renderer, fontId, word.substr(0, offset), style, needsHyphen);
+    if (prefixWidth > availableWidth || prefixWidth <= chosenWidth) {
+      continue;  // Skip if too wide or not an improvement
+    }
+
+    chosenWidth = prefixWidth;
+    chosenOffset = offset;
+    chosenNeedsHyphen = needsHyphen;
+  }
```
daveallie, line 303: This, combined with `allowFallbackBreaks` for the first word of the line, will mean that the first word will always ignore the trie data and just break at the last possible character for the line. It would be great to have it prefer the language hyphenator supplied breaks and if none are suitable (no breaks or all result in too long of a word), then fallback to the fallback breaks.

lib/Epub/Epub/hyphenation/HyphenationCommon.cpp
osteotek: Good catch! I’ve updated the Latin character detection to include specific Latin-1 characters. This has significantly improved the hyphenation evaluation for German and French.

lib/Epub/Epub/hyphenation/LanguageRegistry.cpp
osteotek: changed to default values

lib/Epub/Epub/hyphenation/Hyphenator.cpp
osteotek: I primarily did that to keep it private, but I’ve moved the initialization to the header file anyway.

lib/Epub/Epub/Section.h
osteotek: refactored

lib/Epub/Epub/hyphenation/Hyphenator.cpp
osteotek: Here I assume that the text was processed by external hyphenation engine (like "Hyphen This" plugin for Calibre), and it added soft hyphens according to personal user preferences, which might be different from minPrefix/minSuffix we set by default in the firmware. That's why scanning the whole word makes sense, in my opinion.

lib/Epub/Epub/hyphenation/Hyphenator.cpp
osteotek: This was for cases when text contains two different markers in a row (a soft hyphen followed by a literal '-'). But going through code now not sure how I thought it should affect logic. Let's remove it

lib/Epub/Epub/hyphenation/Hyphenator.cpp
osteotek: Russian language hyphenation rules prohibit breaking words with hyphens at all. I included this rule-breaking logic here because, in my opinion, it makes sense on our smaller screen (and to avoid having words with two hyphens in them). English, on the other hand, doesn’t seem to have such a rule (but I couldn’t find any properly formulated rules for end-of-line hyphenation in English, likely because it’s less commonly used due to shorter words)

lib/Epub/Epub/ParsedText.cpp
osteotek: agreed, added logic to use fallback breaks only if language hyphenator returns no results

lib/Epub/Epub/hyphenation/Hyphenator.cpp
osteotek: simplified logic here

lib/Epub/Epub/ParsedText.cpp
osteotek: added logic to break overfit first words on the line even if hyphenation is disabled

osteotek:
@jonasdiemer I've increased min suffix/prefix length for English to 3, would you be able to test? We can implement proper text layout algorithm that penalize hyphenations (like Knuth-Plass algorithm), but it can be done in another PR, I think

lukestein:
I've been testing (with English-language epubs) and all continues to look good to me.

Even flashing back and forth to the release without hyphenation seems OK despite my not clearing cache in `.crosspoint`, which was an issue before some of your more recent updates.

Continued thanks, @osteotek! For me this is the most important feature to get in for actual quality of reading experience.

jlaunay:
@osteotek Same here, I tested it with French like @lukestein  did with English and everything seems to work perfectly. In my opinion, it's an essential feature for proper epub rendering especially on a small screen. Thanks for your work.

lukestein:
CI seems to want the PR title renamed from "feature: " to "feat: "

```
Available types:
 - feat: A new feature
 - fix: A bug fix
 - docs: Documentation only changes
 - style: Changes that do not affect the meaning of the code (white-space, formatting, missing semi-colons, etc)
 - refactor: A code change that neither fixes a bug nor adds a feature
 - perf: A code change that improves performance
 - test: Adding missing tests or correcting existing tests
 - build: Changes that affect the build system or external dependencies (example scopes: gulp, broccoli, npm)
 - ci: Changes to our CI configuration files and scripts (example scopes: Travis, Circle, BrowserStack, SauceLabs)
 - chore: Other changes that don't modify src or test files
 - revert: Reverts a previous commit
```

lib/Epub/Epub/hyphenation/Hyphenator.cpp
osteotek: Actually, in my testing of dictionary-based language hyphenator, words with hard hyphens do not produce break points at all. Therefore, there’s no point in parsing such words with the language hyphenator.

lukestein:
Sorry to be repeatedly thanking @osteotek, but much appreciate that you keep merging the master commits into your branch. Makes using/testing your feature easy.

fischer-hub:
> @osteotek Same here, I tested it with French like @lukestein did with English and everything seems to work perfectly. In my opinion, it's an essential feature for proper epub rendering especially on a small screen. Thanks for your work.

German also looked good to me while testing over the last days, great work really!

lukestein:
@osteotek, I don't know if this is a reversion or something specific about the epub I'm reading now, but with the latest commits there may be a problem with failure to wrap after dashes:

image

osteotek:
@lukestein yes, there was a change in recent commits on how we define explicit hyphen. We need to decide if we want to break on such dashes. Technically rules of hyphenation in English do not allow that, but it might make sense on such a small screen

lukestein:
> @lukestein yes, there was a change in recent commits on how we define explicit hyphen. We need to decide if we want to break on such dashes. Technically rules of hyphenation in English do not allow that, but it might make sense on such a small screen

Continue to appreciate your work on this!

My screenshot here shows em dashes (—) and my understanding is that **all** major English style guides suggest allowing line breaks after an em (Chicago, APA, MLA, AP, Oxford, and NY Times; some of these also allow breaking before an em but I wouldn't recommend it).

I also think that especially on small screens, this is really important.


I'm less opinionated about en dashes. Style guide generally discourage breaking after these.


For explicit hyphens, I just know your earlier solution fixed #217 and I want to make sure the final code still does.


Once again, many thanks.

lukestein:
@osteotek, with caveats that I don't entirely understand the code, I chatted a bit with AI about it (and my understanding of typography), where it eventually suggested this to me. You're the expert, but just in case useful:

> Looking at the C++ diff in your image, it appears the developer removed several Unicode codepoints from the `isExplicitHyphen` function. By removing them, the engine no longer recognizes these characters as valid "hard" break points, which explains why your em dash wrapping recently broke.
> 
> To restore the behavior you want—where the engine breaks lines after these characters—you should selectively add specific cases back.
> 
> Based on the typographic standards we discussed, you should restore the following:
> 
> * **`case 0x2014:` (Em Dash)**: This is the most critical. Restoring this will allow the engine to break a line immediately following an em dash, preventing the "word—word" strings from being treated as a single unbreakable block.
> * **`case 0x2013:` (En Dash)**: While style guides suggest avoiding breaks in simple ranges, an en dash *is* an explicit hyphenation point in complex compound adjectives (e.g., "pre–World War II"). Including it provides the engine with a necessary fallback for long strings.
> * **`case 0x2E3A:` and `case 0x2E3B:` (Two-em and Three-em Dashes)**: These are used in bibliography work or to indicate missing text. If they aren't in the explicit hyphen list, the engine will likely overflow the line or create massive gaps because these characters are quite wide.
> * **`case 0x2012:` (Figure Dash)**: This is a dash the width of a digit. Like the en dash, it is often used in phone numbers or numerical data where a break point might be needed on very narrow screens.

osteotek:
@lukestein agreed, restored breaking after en/em dashes. Should work the same as before

lukestein:
> @lukestein agreed, restored breaking after en/em dashes. Should work the same as before

Looking great with this commit 🙌 Will keep testing (i.e., reading books)

IMG_2894

lukestein:
Hi @osteotek.

**Personally, I think from a user experience perspective the current PR could/should be merged as-is** (pending @daveallie's review ofc) and then there is room for potential refinements later. Thank you!

However, one strong suggestion based on continued testing is to **please add the ellipsis character** (`…`, `U+2026`) to the `isExplicitHyphen` list.

---

Other likely good candidates for`isExplicitHyphen` (although these haven't actually come up in my reading) informed by conversations with LLM:

Character | Unicode | Usage | Why include it?
-- | -- | -- | --
Solidus (Slash) | U+002F | / | Essential for breaking URLs and "and/or" constructs.
Backslash | U+005C | \ | Critical for technical text, file paths, and coding documentation.
Underscore | U+005F | _ | Prevents "runaway" line lengths in usernames or code snippets.
Middle Dot | U+00B7 | · | Acts as a semantic separator in dictionaries or stylistic lists.
Ellipsis | U+2026 | … | Prevents justification failure when dialogue lacks following spaces.
Midline Horizontal Ellipsis | U+22EF | ⋯ | Useful for mathematical sequences and technical notation.

```
    case 0x002F:  // Solidus (slash)
    case 0x005C:  // Backslash
    case 0x005F:  // Underscore
    case 0x00B7:  // Middle dot
    case 0x2026:  // Ellipsis
    case 0x22EF:  // Midline horizontal ellipsis
```

Issue #321, opened 2026-01-11
Pin actions in github workflows to specific versions


andrewb1269:
Pin actions to github workflows to specific versions to prevent upstream attacks.

If an upstream action is compromised and we use the latest tag, we will consume the compromised action.

For example:
`uses: actions/checkout@v6`
We use action `v6.1.0`
A malicious individual compromises the action, adds a new commit, and publishes `v6.2.0`. Our workflow will automatically consume the latest `v6.2.0` which may compromise sensitive information from our repo.

Solution:
Pin actions to specific commit hashes rather than using the latest tag:
`uses: actions/checkout@1a2b3c # v6.1.0`

daveallie:
CI currently only uses GitHub published actions from the Actions repo: https://github.com/actions

While there is always a non-zero chance of a supply chain attack, the risk and surface here are extremely low. This is an open source repo without secrets processing no secure information, there is no sensitive data to be stolen. The only risk I can see is compilation of a compromised firmware through the avenue of a convoluted multi-step supply chain attack, which is highly unlikely.

I don't intend to be testing every release of new minor and patch versions of actions, so pinning to specific versions is likely to result in a bigger risk where a older compromised version of the action is in use and the manual action to upgrade to the newer version isn't performed.

I will enable dependabot on this repo which will alert for sensitive actions being used. But will not be pinning specific hashes for the GitHub controlled actions.

Issue #319, opened 2026-01-11
Cleanup Github Workflow Spacing


andrewb1269:
Clean up some of the github workflow file spacing. Minor changes only.

daveallie:
Merged that original PR, just as an FYI, fixes don't need a corresponding issue, this one could have had all the required context in the PR.

PR #323, opened 2026-01-11
chore: pin actions in workflows to commit hashes

andrewb1269:
**Description**:

Pin all actions used in `.github/workflows` to specific commit hashes.

**Related Issue(s)**:

Implements #319 if that has not already been resolved.
Implements #321


daveallie:
See https://github.com/daveallie/crosspoint-reader/issues/321#issuecomment-3737916487

PR #330, opened 2026-01-12
Generate different .bmp for cropped covers so settings have effect.

jonasdiemer:
Addresses https://github.com/daveallie/crosspoint-reader/pull/225#issuecomment-3735150337


daveallie:
Thanks for addressing this

PR #341, opened 2026-01-12
feat: Add setting to toggle long-press chapter skip

armandocerna:
## Summary

Adds a new "Long-press Chapter Skip" toggle in Settings to control whether holding the side buttons skips chapters.

I kept accidentally triggering chapter skips while reading, which caused me to lose my place in the middle of long chapters.

## Additional Context

  specific areas to focus on).

---

### AI Usage

Did you use AI tools to help write this code? _**PARTIALLY **_


lukestein:
Thank you! This has been a regular annoyance for me. I thought maybe I'd created an issue but haven't found it. Given we have the Chapter Selector (and lack any undo/back or similar) I'll definitely turn these long-press-chapter buttons off.

mvclgit:
Thank you for doing this! It was killing me!

PR #351, opened 2026-01-13
Feature: Extend Cover image

jonasdiemer:
## Summary

This feature adds an option "extend" to the Cover mode, which takes 20 pixels from the corners of the cover image and repeats them outwards in accordeon style. This avoids cropping and still fills the whole screen. Works best when the image "almost" fits.

---

### AI Usage

Did you use AI tools to help write this code? _** PARTIALLY**_


pablohc:
Hi @jonasdiemer related to the sleep screen, I opened this issue: https://github.com/daveallie/crosspoint-reader/issues/348.
Is there any difference in image processing depending on the image adjust mode?

daveallie:
Obviously every book cover is different, but the usable covers for this kind of upscaling is extremely small, as an example:

| Fit/Crop | Extend |
| --- | --- |
| IMG_5813 | IMG_5812 |

If we plan to upscale to fit/crop then we should actually be upscaling instead of repeating parts of the image. Even a very basic repeat every nth pixel, like how we downscale by skipping every nth pixel would be good.

jonasdiemer:
> Obviously every book cover is different, but the usable covers for this kind of upscaling is extremely small, ...
> If we plan to upscale to fit/crop then we should actually be upscaling instead of repeating parts of the image. Even a very basic repeat every nth pixel, like how we downscale by skipping every nth pixel would be good.

I fully agree, we'd need separate upscaling for when the image is too small in both directions. 

I found the extension working very well when the cover is slightly too large/small in one dimension as a good alternative to cropping (which has the risk of removing content) or fitting (which leaves blank space)...

Eg here, "fit" leaves space:
IMG_1737

whereas expand works nicely (agreed, in this example crop would have worked, too):

IMG_1738

daveallie:
Before this is merged I'd love to solve the case raised here: https://github.com/crosspoint-reader/crosspoint-reader/pull/351#issuecomment-3749683185

It might just as simple as **always upscaling** to at least the screen size (based on fit/fill), and then applying extend after that upscale.

jonasdiemer:
> Before this is merged I'd love to solve the case raised here: [#351 (comment)](https://github.com/crosspoint-reader/crosspoint-reader/pull/351#issuecomment-3749683185)
> 
> It might just as simple as **always upscaling** to at least the screen size (based on fit/fill), and then applying extend after that upscale.

Yes, we should probably always upscale. If anyone wants to take a stab at this, don't wait for me 😉

daveallie:
Moving this back into draft (mostly for reviewer benefit when scanning open PRs), won't be merged without addressing comments

znelson:

PR #352, opened 2026-01-13
Reclaim space if we don't show battery Percentage

jonasdiemer:
## Summary

Give space to the chapter title if we don't show battery percentage.

---

### AI Usage

### DIFF
--- a/src/activities/reader/EpubReaderActivity.cpp
+++ b/src/activities/reader/EpubReaderActivity.cpp
@@ -479,7 +479,8 @@ void EpubReaderActivity::renderStatusBar(const int orientedMarginRight, const in
       }
     }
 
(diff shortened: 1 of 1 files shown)


pablohc:
Hi @jonasdiemer related to chapter title, I opened this issue https://github.com/daveallie/crosspoint-reader/issues/354

Is it possible to align it with respect to the center of the screen?

jonasdiemer:
It's possible to center, but then we give up space in case left and right "clutter" (battery + progress) are using up different amounts of space.

daveallie:
There's likely a middle ground here, calculating the max width we can display the text while centered. If it's less than that, then render it in the center, otherwise use the existing centering logic. Would just require knowing the max of left and right title margins.

Thoughts @jonasdiemer?

andrewb1269 (approved):

daveallie (dismissed):
Adjusted the logic to support https://github.com/crosspoint-reader/crosspoint-reader/pull/352#issuecomment-3749516349

PR #363, opened 2026-01-13
fix: draw button hints correctly if orientation is not portrait

maeveynot:
~~Quick~~ fix for https://github.com/daveallie/crosspoint-reader/issues/362

(this just applies to the chapter selection menu:)

~~If the orientation is portrait, hints as we know them make sense to draw. If the orientation is inverted, we'd have to change the order of the labels (along with everything's position), and if it's one of the landscape choices, we'd have to render the text and buttons vertically. All those other cases will be more complicated.~~

~~Punt on this for now by only rendering if portrait.~~

Update: this now draws the hints at the physical button position no matter what the orientation is, by temporarily changing orientation to portrait.

### DIFF
--- a/lib/GfxRenderer/GfxRenderer.cpp
+++ b/lib/GfxRenderer/GfxRenderer.cpp
@@ -484,6 +484,7 @@ void GfxRenderer::drawButtonHints(const int fontId, const char* btn1, const char
     // Only draw if the label is non-empty
     if (labels[i] != nullptr && labels[i][0] != '\0') {
       const int x = buttonPositions[i];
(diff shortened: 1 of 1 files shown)


maeveynot:
Okay, pushing a new version that does that. Since we modify the orientation inside `drawButtonHints` it isn't `const` anymore; is that a problem?

daveallie:
> Okay, pushing a new version that does that. Since we modify the orientation inside `drawButtonHints` it isn't `const` anymore; is that a problem?

Not an issue

daveallie (approved):

PR #366, opened 2026-01-14
fix: render U+FFFD replacement character instead of ?

maeveynot:
The current behavior of rendering `?` for an unknown Unicode character can be hard to distinguish from a typo. Use the standard Unicode "replacement character" instead, that's what it's designed for:

https://en.wikipedia.org/wiki/Specials_(Unicode_block)

I'm making this PR as a draft because I'm not sure I did everything that was needed to change the character set covered by the fonts. Running that script is in its own commit. If this is proper, I'll rebase/squash into one commit and un-draft.


maeveynot:
An example of this showing up in a user question is https://www.reddit.com/r/xteinkereader/comments/1qbmn2p/x4_crosspoint_0120_question_marks_mid_text/

Re: their question, It looks like we can just uncomment Greek, yeah? But that's a separate PR.

daveallie:
Your script updating looks good, no concerns there, happy to approve and merge this once you're happy.

maeveynot:
OK, ready to go! Thanks!

PR #384, opened 2026-01-14
feat: add SD card format option

khromov:
## Summary

The PR adds a new screen that allows the user to format the inserted SD card via the Settings page. This is helpful if the user has no access to a computer, or would like to use the device without connecting to one (ie. by uploading books directly via the hotspot feature).

Requires the changeset in https://github.com/open-x4-epaper/community-sdk/pull/14

PXL_20260114_230408087(1)

## Additional Context

  specific areas to focus on).

---

### AI Usage

Did you use AI tools to help write this code? _**PARTIALLY**_ - partially written with AI, manually reviewed and validated on-device.


danwilldev:
Would it be possible for the user to pass the cluster size via crosspoint settings for https://github.com/open-x4-epaper/community-sdk/pull/14#issuecomment-3756470523

andrewb1269:
> Would it be possible for the user to pass the cluster size via crosspoint settings for https://github.com/open-x4-epaper/community-sdk/pull/14#issuecomment-3756470523

I don't think we should allow that. It will clutter the interface and generally folks using the device to format the SD card will not care about the cluster size. I think keeping it simple makes more sense and defaulting to 32K with exFat is the best path forward. If someone wants to format a specific cluster size they can use a computer to do so.

The other concern with cluster size space usage may not really matter in the grand scheme of things. Generally Epubs are small and large SD cards are cheap so saving every possible KB of space is not necessary with options of a cluster size.

daveallie (commented):
Pending resolution and merge of https://github.com/open-x4-epaper/community-sdk/pull/14, moving back to draft. Once that's resolved, feel free to move back to ready for review.

Jason2866:
Why at all?  EVERY byte added code not really needed is waste. It is not a CPU it is a tiny small MCU with very low resources. SD Cards can be formatted elsewhere

znelson:

PR #390, opened 2026-01-15
fix: Invert colors on home screen cover overlay when recent book is selected

### LINKED ISSUE #388: UI: Text overlay for recent book on home screen should have inverted colors when selected.
#200 is great and now we have covers on the home screen!

It should be clearer when the book is selected—this currently only shows up as a thicker black border around the cover.

When the recent book is selected, I suggest the overlay text should invert colors (white text on black background), consistent with the other buttons on the home screen.

## Unselected

image

## Selected

image

lukestein:
## Summary

* Fixes #388 

## Additional Context

* Tested on my own device
  * See images at #388 for what home screen looked like before.
  * With this PR the home screen shows the following (selected and unselected recent book × cover image rendered or not)

Picsew_20260115153419

---

### AI Usage

### DIFF
--- a/src/activities/home/HomeActivity.cpp
+++ b/src/activities/home/HomeActivity.cpp
@@ -327,8 +327,7 @@ void HomeActivity::render() {
   if (hasContinueReading) {
     // Invert text colors based on selection state:
     // - With cover: selected = white text on black box, unselected = black text on white box
-    // - Without cover: selected = white text on black card, unselected = black text
-    const bool invertText = coverRendered ? bookSelected : !bookSelected;
+    // - Without cover: selected = white text on black card, unselected = black text on white card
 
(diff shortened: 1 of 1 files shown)


lukestein:
Just noting that @itsthisjustin has made the [reasonable but controversial? suggestion](https://github.com/crosspoint-reader/crosspoint-reader/pull/200#issuecomment-3755448894) that there should be no text overlays on the rendered book cover at all.

daveallie:
As per https://github.com/crosspoint-reader/crosspoint-reader/pull/200#issuecomment-3767961417, I am not keen to remove the title just yet, however I do think we can move it to be 2 lines under the book and push some other content around. Happy to approve this while we figure out that other discussion.

PR #392, opened 2026-01-16
feat: hold confirm/select for caps

### LINKED ISSUE #374: Keyboard: hold Select for caps
It'd be nice to not have to move down to the corner to hit the caps button every time.

justinluque:
## Summary

Implements a quick caps feature in the keyboard entry activity. Previously you would have to navigate the cursor to the shift key each time they need to change case. With this change, users can hold confirm/select for a short period and release it to toggle caps. Resolves #374.

## Additional Context

This is a draft pull request because I still have no way to test the firmware myself (waiting for the device to ship.) I'm not completely sure this commit works out of the box since I have no way of trying it out myself.

### AI Usage

I didn't use any AI tools in making this change.

### DIFF
--- a/src/activities/util/KeyboardEntryActivity.cpp
+++ b/src/activities/util/KeyboardEntryActivity.cpp
@@ -226,13 +226,12 @@ void KeyboardEntryActivity::loop() {
   }
 
   // Selection
-  if (mappedInput.isPressed(MappedInputManager::Button::Confirm) && mappedInput.getHeldTime() >= capsMs) {
-    onCapsHeld();
-    return;
-  }
-
-  if (mappedInput.wasPressed(MappedInputManager::Button::Confirm) && mappedInput.getHeldTime() < capsMs) {
-    handleKeyPress();
+  if (mappedInput.wasPressed(MappedInputManager::Button::Confirm)) {
+    if (mappedInput.getHeldTime() >= capsMs) {
+      shiftActive = !shiftActive;
+    } else {
+      handleKeyPress();
+    }
     updateRequired = true;
   }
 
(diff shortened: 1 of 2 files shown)


src/activities/util/KeyboardEntryActivity.cpp
```diff
@@ -221,7 +226,12 @@ void KeyboardEntryActivity::loop() {
...
-  if (mappedInput.wasPressed(MappedInputManager::Button::Confirm)) {
+  if (mappedInput.isPressed(MappedInputManager::Button::Confirm) && mappedInput.getHeldTime() >= capsMs) {
+    onCapsHeld();
+    return;
+  }
+
+  if (mappedInput.wasPressed(MappedInputManager::Button::Confirm) && mappedInput.getHeldTime() < capsMs) {
     handleKeyPress();
     updateRequired = true;
   }
```
daveallie: Given that `onCapsHeld()` immediately just waits for you to release it (even before updating the UI), then we should either simplify this to:

```cpp
  if (mappedInput.wasPressed(MappedInputManager::Button::Confirm)) {
    if (mappedInput.getHeldTime() >= capsMs) {
      shiftActive = !shiftActive;
    } else {
      handleKeyPress();
    }
    updateRequired = true;
  }
```

or if you want the UI to update before you release the button, then `onCapsHeld` is going to have to make sure `updateRequired` is set to `true` before the busy wait happens so the other task will update the screen.

daveallie (changes requested):

src/activities/util/KeyboardEntryActivity.cpp
justinluque: Fair enough, that makes more sense.

osteotek:
We have Shift Lock option now (https://github.com/crosspoint-reader/crosspoint-reader/pull/513) with the same effect, this PR can be closed

Issue #281, opened 2026-01-07
Feature request/Bug report: Some way to delete cached metadata


lukestein:
There currently appears to be no way to re-index an updated epub file other than manually deleting the associated metadata folder from `.crosspoint` on the SD card.

I can't figure out a way to do this over the network, so it requires physically removing the SD card.

As noted in #279, this can come up after e.g., adding or correcting a TOC and then updating the epub file on the device.

As noted in #225, this is the only way to get the cover options (fit/crop) to apply to a book that's previously been opened).

I wonder if it would be possible to either

1. Add a way to manually reset a book's metadata
2. Have the web uploader delete metadata when an epub is deleted or replaced
3. Automatically generate new metadata if an epub changes (using a file update time? using a hash check, which may already be included in draft code for koreader sync?)

lukestein:
Closed by #393?

PR #393, opened 2026-01-16
feat: invalidate cache on web uploads and opds downloads and add Clear Cache action

lgarbarini:
## Summary

When uploading or downloading an updated ebook from SD/WebUI/OPDS with same the filename the `.crosspoint` cache is not cleared. This can lead to issues with the Table of Contents and hangs when switching between chapters. 

I encountered this issue in two places:
- When I need to do further ePub cleaning using Calibre after I load an ePub and find that some of its formatting should be cleaned up. When I reprocess the same book and want to place it back in the same location I need a way to invalidate the cache.
- When syncing RSS feed generated epubs. I generate news ePubs with filenames like `news-outlet.epub` and so every day when I fetch new news the crosspoint cache needs to be cleared to load that file. 

This change offers the following features:
- On web uploads, if the file already exists, the cache for that file is cleared
- On OPDS downloads, if the file already exists, the cache for that file is cleared
- There's now an action for `Clear Cache` in the Settings page which can clear the cache for all books

Addresses https://github.com/crosspoint-reader/crosspoint-reader/issues/281

---

### AI Usage

Did you use AI tools to help write this code? PARTIALLY


itsthisjustin:
Would love some more details on the why behind this

lgarbarini:
> Would love some more details on the why behind this

Updated description

lukestein:
Another place this is useful is that after installing dev builds of the firmware I sometimes need to rebuild the cache. Getting tired of sd card eject and `rm`…

Issue #400, opened 2026-01-16
Keep book text on lock screen?


nickmorgan19457:
Is there a way to keep  book text on the lock screen so I can start reading before it’s turned back on?

Great work, by the way. It’s a very pleasant reading experience.

efenner:
I would be interested in this as a mode. Maybe you could have one behavior when manually putting the device to sleep, and another when the device goes to sleep from idle?

It would be a great way to put some quick reference items on, especially now that we have TXT file support. Something like a grocery list or workout plan where you want it to be glanceable and not time out.

And maybe indicate in the status bar at the bottom that the device is sleeping

andrewb1269:
I think this is a good idea and seems straightforward enough. I left a comment on the linked PR.

PR #410, opened 2026-01-17
feat: current screen as lock screen

yingirene:
## Summary

- This PR aims to solve the issue in #400 . 
- Added a new sleep screen setting for "Last Screen" that just shows what was last on the screen with a box at the bottom of the screen that says "Sleeping...".
- Added a new sleep screen setting and functionality for keeping the last screen on the device when in sleep mode.

## Additional Context

  specific areas to focus on).
Tested with both manually putting device the sleep and entering sleep after an idle period.

---

### AI Usage

Used AI to understand the existing code and decide how to implement the solution.


src/activities/boot_sleep/SleepActivity.cpp
```diff
@@ -268,6 +272,26 @@ void SleepActivity::renderCoverSleepScreen() const {
...
+  const int h = renderer.getLineHeight(UI_12_FONT_ID) + margin * 2;
+  const int y = renderer.getScreenHeight() - h - margin;
+
+  renderer.fillRect(x - 5, y - 5, w + 10, h + 10, true);
+  renderer.fillRect(x + 5, y + 5, w - 10, h - 10, false);
+  renderer.drawText(UI_12_FONT_ID, x + margin, y + margin, "SLEEPING...", true, EpdFontFamily::BOLD);
+
+  // Make sleep screen dark unless light is selected in settings
+  if (SETTINGS.sleepScreen == CrossPointSettings::SLEEP_SCREEN_MODE::DARK) {
+    // Draw dark sleep image
```
andrewb1269: Do you expect to have this if-check actually do something? You're checking to see `SLEEP_SCREEN_MODE::DARK` but never performing an action on it.

yingirene: No, I was thinking it would be good to have a dark mode version, but it's probably not necessary at this point. Updated!

andrewb1269 (dismissed):

src/activities/boot_sleep/SleepActivity.cpp
andrewb1269: Awesome! I approved but I don't have any real power here. Just showing that someone else looked at the code and thinks it looks reasonable.

Did you test this on device by chance? If not we should probably have someone test it

andrewb1269:
Please see [this](https://github.com/crosspoint-reader/crosspoint-reader/pull/410#discussion_r2703000974) comment before merging.

src/activities/boot_sleep/SleepActivity.cpp
yingirene: Yes, I tested with both manually putting device the sleep and entering sleep after an idle period.

src/CrossPointSettings.h
```diff
@@ -16,7 +16,7 @@ class CrossPointSettings {
   CrossPointSettings& operator=(const CrossPointSettings&) = delete;
 
   // Should match with SettingsActivity text
-  enum SLEEP_SCREEN_MODE { DARK = 0, LIGHT = 1, CUSTOM = 2, COVER = 3, BLANK = 4 };
+  enum SLEEP_SCREEN_MODE { DARK = 0, LIGHT = 1, CUSTOM = 2, COVER = 3, LAST_SCREEN = 4, BLANK = 5 };
```
daveallie: There's an issue here we should probably solve (either now or in the future), but due to the mapping change, users who had their  sleep screen mode setting set to "none", will be automatically swapped to "last screen".

I don't have a good way around this, except by making the enum setting type take a set of ordered options.

To ensure we're retaining user settings, let's move this new option to the end of the list as option 5 (even though it feels a little out of order).

daveallie (changes requested):
Appreciate the first pass @andrewb1269, I realise that it may feel a little powerless with the current setup but your review signal is still very useful in speeding up reviews!

src/CrossPointSettings.h
yingirene: Fixed!

Eloren1:
Is there any sleeping indication? An icon or some graphic. How does it look?

ngxson:
Merging some discussions from https://github.com/crosspoint-reader/crosspoint-reader/pull/495 to here, I think we should implement this feature as a dedicated setting: "Show sleep screen":
- "Always" --> always show sleep screen on power button press, or on timeout
- "Except timeout" --> show sleep screen on power button press, but keep last screen on timeout
- "Never" --> always keep last screen on timeout

yingirene:
> Is there any sleeping indication? An icon or some graphic. How does it look?

Here's how it should look:
last screen test

Eloren1:
> Here's how it should look:

What if I read in landscape? We discussed maybe a moon icon then in the bottom left corner, that looks good in all orientations

yingirene:
> > Here's how it should look:
> 
> What if I read in landscape? We discussed maybe a moon icon then in the bottom left corner, that looks good in all orientations

I can look into that.

osteotek:
What about instead of the sleep screen options based on timeout we add a transparent sleep screen option (in addition to light, dark, cover, custom) with the same functionality? Do we really need different options if the device goes to sleep on timeout or on the power button?

Eloren1:
> instead of the sleep screen options based on timeout we add a transparent sleep screen option (in addition to light, dark, cover, custom)

It is already done

osteotek:
If I understand this PR correctly, it can lead to inconsistency in settings: if Sleep Screen Mode is set to Cover, but Show Sleep Screen set to Never, then cover is actually never shown. Sounds confusing to me that one setting has priority over the other. It can probably be solved by better wording

ngxson:
> Do we really need different options if the device goes to sleep on timeout or on the power button?

At least in my use case, it can be useful when I want to display a memo note, I don't want the sleep screen to be turned on. I think it's a valid generic use case (?)

Ofc I can change the settings whenever I want it to be either a reader or a memo, but that's not very convenient

Re. one setting take precedence over the other, we also have the sleep screen cover mode which requires sleep to set to cover. I think the mode "never" can be simply change to "Disable", and maybe in the future we can allow show/hide setting entries on the fly

yingirene:
Updated to use a moon icon. Drawn by yours truly.

moon_test

yingirene:
> Re. one setting take precedence over the other, we also have the sleep screen cover mode which requires sleep to set to cover. I think the mode "never" can be simply change to "Disable", and maybe in the future we can allow show/hide setting entries on the fly

I think "Disable" may be more confusing that "Never', but I do think the wording can use some tweaking. Maybe this is something we can change later after getting user feedback?

osteotek (dismissed):

src/CrossPointSettings.cpp
```diff
@@ -89,6 +90,8 @@ bool CrossPointSettings::loadFromFile() {
   do {
     readAndValidate(inputFile, sleepScreen, SLEEP_SCREEN_MODE_COUNT);
     if (++settingsRead >= fileSettingsCount) break;
+    serialization::readPod(inputFile, showSleepScreen);
+    if (++settingsRead >= fileSettingsCount) break;
```
daveallie: New settings must appear at the end of the file, otherwise when people update, they'll be pulling in values for the wrong settings

daveallie (changes requested):
Small change needed, but otherwise looks good

osteotek:
Tested, works as intended. One thing I thought: would be great if boot screen didn't show up after waking up when current screen is used as last screen, would work much cleaner this way.

yingirene:
> Tested, works as intended. One thing I thought: would be great if boot screen didn't show up after waking up when current screen is used as last screen, would work much cleaner this way.

Changed it so the boot screen doesn't show when the reader is the last open activity. I feel like it still makes sense to have the boot screen when the other activities are open because the screen would change to the reader.

osteotek:
> > Tested, works as intended. One thing I thought: would be great if boot screen didn't show up after waking up when current screen is used as last screen, would work much cleaner this way.
> 
> Changed it so the boot screen doesn't show when the reader is the last open activity. I feel like it still makes sense to have the boot screen when the other activities are open because the screen would change to the reader.

thank you! Apologies, with the flurry of merges to master there are quite a few conflicts to resolve. Otherwise, this is ready for merging, I believe

lukestein (dismissed):
No code review, but tested on device and seems to work exactly as expected
- Triggered with power button (set to "Never" show sleep screen) or reached via sleep timer
- epub, text, and xtc

Thanks!🎉

yingirene:
> > > Tested, works as intended. One thing I thought: would be great if boot screen didn't show up after waking up when current screen is used as last screen, would work much cleaner this way.
> > 
> > 
> > Changed it so the boot screen doesn't show when the reader is the last open activity. I feel like it still makes sense to have the boot screen when the other activities are open because the screen would change to the reader.
> 
> thank you! Apologies, with the flurry of merges to master there are quite a few conflicts to resolve. Otherwise, this is ready for merging, I believe

No worries! Thank you for the feedback and review.

FetchFast:
> changed boot screen

I just want to highlight that there are actually 2 things being discussed here to think about in planning: 

1) Don't need to wait past the boot screen to start reading 

and 

2) Use the device in some way during sleep screen. 

(1) is a general behaviour improvement for everyone IMHO, whereas
(2) is a feature addition

IMHO, 
(1) boot screen is what I'd prioritise
(2) sleep screen could open a door to other features like having quick reference to a flight ticket QRcode or calendar

osteotek:
@yingirene would it be possible to fix merge conflicts? This is valuable feature, I believe

Eloren1:
No updates? 😥

Eloren1:
@projectcato check out this one?

znelson:
@yingirene I'll check back in a week or two. I know there's interest here so if you're still around please merge with main and we can bring this in.

itsthisjustin:
> @yingirene I'll check back in a week or two. I know there's interest here so if you're still around please merge with main and we can bring this in.

I just happened upon this by sorting by oldest. If the community is ok with it, I'm considering part of the maintainers tasks being that we help contributors by solving merge conflicts and moving these along. I would also put forth the idea of tagging prs as stale if there's been no updates for x amount of time so we can start closing branches out

itsthisjustin:
I also am wondering if this conflicts with #1994 and we should just merge that one instead as you could simply have a fully transparent image there which would solve the same purpose as this I believe

Eloren1:
Don't these image overlays add a lot more complication? Almost overwhelmed by looking at that. It looks like something from CrossPet and other more advanced firmwares. Currently I'm actualizing this pull request. The device will seamlessly go to sleep (moon icon in the bottom left), and will wake up seamlessly (loading icon in the bottom left). Minimal and simple.

Eloren1:
image


20260519_121633 (Small)


v1.3.0 WIP 😇

Eloren1:
@itsthisjustin I invite you to see this #2064

itsthisjustin:
> Don't these image overlays add a lot more complication? Almost overwhelmed by looking at that. It looks like something from CrossPet and other more advanced firmwares. Currently I'm actualizing this pull request. The device will seamlessly go to sleep (moon icon in the bottom left), and will wake up seamlessly (loading icon in the bottom left). Minimal and simple.

While I agree, one can simply drop a fully transparent png in there to solve that issue. I do like your latest updates here so we should move this to a discussion and get some community feedback regarding best way forward

itsthisjustin:
Closed in favor of the above fresh PR which adds an indicator to show sleep status which is good ux

Issue #182, opened 2025-12-30
Style change introduces spurious spaces


jonasdiemer:
### Affected Version

0.11.1-dev

### Describe the Bug

It looks like (from the code in ChapterHtmlSlimParser.cpp) the styling is currently limited to be on word boundaries.

For some cases, this introduces extra spacing (i.e. word gaps between characters that should be together), e.g. this
```html
Because that’s <em>what we do</em>.
```
turns into
```text
Because that’s what we do .
```

These spaces are obviously exaggerated by justified text alignment.

I am seeing we are already aware of the situation (code comment states "text styling needs to be overhauled to fix it") - creating this issue to discuss a solution.

### Steps to Reproduce

1. Open EPUB with inline italic (or bold) formatting (e.g. "...</em>.")

### Expected Behavior

No space after end of style if source material contains no space

### Relevant Log Output/Screenshots

```shell

```

jonasdiemer:
I wonder if a solution could be to introduce a special character at the end of words who should be "merged" with their successor after styling was applied. This could avoid major rewrite of the layout code (I assume).

daveallie:
yeah I've been mulling over this one for a bit, it's been around since the beginning.

A sentinel to prevent a space is a great idea I hadn't thought of. All inline HTML elements suffer from this problem, e.g. `this en<span>orm</span>ous word` would show as `this en orm ous word`, where a true HTML renderer should show `this enormous word`

We'd need to make the layouter aware that linebreaks shouldn't happen between these "words" as well. The other thought I had was shifting away from the current `wordStyles` approach and have a vector which represented the indices and new style each time the style changes. It would mean that the parser could be updated to output these style change points but still maintain words interrupted by an inline block as one word.

jonasdiemer:
Yeas, keeping explicitly track of style changes ids likely the cleaner approach. We'd need to consider that characters might be added (or removed) during by layouter, e.g. due to (soft) hyphenation. 

 Might be goid to check how this is soccer in microreader (https://github.com/CidVonHighwind/microreader).

Also, this should be done together with (or after?) implementing hyphenation. At least the soft hyphenation (see https://github.com/daveallie/crosspoint-reader/discussions/17#discussioncomment-15378475) should be rather quick to implement and a good foundation for full hyphenation support (which could be bolted on later).

carsonwick:
Commenting to add interest in this bug and also provide an image that demonstrates the behavior and also how it can push punctuation to the next line.

[...]

PR #411, opened 2026-01-17
feat: Add CSS parsing and CSS support in EPUBs

jdk2pq:
## Summary

- Adds basic CSS parsing to EPUBs and determine the CSS rules when rendering to the screen so that text is styled correctly. Currently supports bold, underline, italics, margin, padding, and text alignment

## Additional Context

- My main reason for wanting this is that the book I'm currently reading, Carl's Doomsday Scenario (2nd in the Dungeon Crawler Carl series), relies _a lot_ on styled text for telling parts of the story. When text is bolded, it's supposed to be a message that's rendered "on-screen" in the story. When characters are "chatting" with each other, the text is bolded and their names are underlined. Plus, normal emphasis is provided with italicizing words here and there. So, this greatly improves my experience reading this book on the Xteink, and I figured it was useful enough for others too.
- For transparency: I'm a software engineer, but I'm mostly frontend and TypeScript/JavaScript. It's been _years_ since I did any C/C++, so I would not be surprised if I'm doing something dumb along the way in this code. Please don't hesitate to ask for changes if something looks off. I heavily relied on Claude Code for help, and I had a lot of inspiration from how [microreader](https://github.com/CidVonHighwind/microreader) achieves their CSS parsing and styling. I did give this as good of a code review as I could and went through everything, and _it works on my machine_ 😄 

### Before
IMG_6271
IMG_6272

### After
IMG_6268
IMG_6269

---

### AI Usage

Did you use AI tools to help write this code? **YES**, Claude Code

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,13 +102,18 @@ After flashing the new features, it’s recommended to capture detailed logs fro
 First, make sure all required Python packages are installed:
 
 ```python
-python3 -m pip install serial colorama matplotlib
+python3 -m pip install pyserial colorama matplotlib
 ```
 after that run the script:
 ```sh
+# For Linux
+# This was tested on Debian and should work on most Linux systems.
 python3 scripts/debugging_monitor.py
+
+# For macOS
+python3 scripts/debugging_monitor.py /dev/cu.usbmodem2101
 ```
-This was tested on Debian and should work on most Linux systems. Minor adjustments may be required for Windows or macOS.
+Minor adjustments may be required for Windows.
 
 ## Internals
 
--- a/src/activities/network/WifiSelectionActivity.cpp
+++ b/src/activities/network/WifiSelectionActivity.cpp
@@ -520,7 +520,7 @@ void WifiSelectionActivity::renderNetworkList() const {
     const auto height = renderer.getLineHeight(UI_10_FONT_ID);
     const auto top = (pageHeight - height) / 2;
     renderer.drawCenteredText(UI_10_FONT_ID, top, "No networks found");
-    renderer.drawCenteredText(SMALL_FONT_ID, top + height + 10, "Press OK to scan again");
+    renderer.drawCenteredText(SMALL_FONT_ID, top + height + 10, "Press Connect to scan again");
   } else {
     // Calculate how many networks we can display
     constexpr int startY = 60;
@@ -546,8 +546,8 @@ void WifiSelectionActivity::renderNetworkList() const {
 
       // Draw network name (truncate if too long)
       std::string displayName = network.ssid;
-      if (displayName.length() > 16) {
-        displayName.replace(13, displayName.length() - 13, "...");
+      if (displayName.length() > 33) {
+        displayName.replace(30, displayName.length() - 30, "...");
       }
       renderer.drawText(UI_10_FONT_ID, 20, networkY, displayName.c_str());
 


roceb:
Can you fix the conflicts on this PR.

jdk2pq:
> Can you fix the conflicts on this PR.

Was in the process of working on that when I got this notification! It should be fixed up now

ttil:
@jdk2pq Thank you for this PR. The handling of the paragraph indentation seems to be off.
`text-indent: 0;` results in an indentation with EmSpace character via the fallback logic. But something like `text-indent: 1.2em;` is not indented at all.
While I have a fix for this, there are different ways to address this. So I'll let you decide how you want to handle it in the PR.

jdk2pq:
> @jdk2pq Thank you for this PR. The handling of the paragraph indentation seems to be off. `text-indent: 0;` results in an indentation with EmSpace character via the fallback logic. But something like `text-indent: 1.2em;` is not indented at all. While I have a fix for this, there are different ways to address this. So I'll let you decide how you want to handle it in the PR.

Thanks for testing this out @ttil ! I just pushed up a commit that should fix that up. There's some additional fixes and improvements around `margin` and `padding` as well. 

I had Claude help me out with generating a test EPUB for all the supported CSS, and it's rendering correctly on my device as far as I can tell. Here's a repo with the code and file: https://github.com/jdk2pq/css-test-epub in case anyone else wants to test it out!

jdk2pq:
Additionally, here's another good test EPUB from the MobileRead wiki https://wiki.mobileread.com/wiki/EPub_Reader_Test

Their test EPUB tests out far more CSS properties than are currently supported by this PR, but the text renders correctly and the CSS properties that are supported also work as expected on my device.

lib/Epub/Epub/css/CssParser.h
```diff
@@ -0,0 +1,99 @@
...
+  // Internal parsing helpers
+  void processRuleBlock(const std::string& selectorGroup, const std::string& declarations);
+  static CssStyle parseDeclarations(const std::string& declBlock);
+
+  // Individual property value parsers
+  static TextAlign interpretAlignment(const std::string& val);
+  static CssFontStyle interpretFontStyle(const std::string& val);
+  static CssFontWeight interpretFontWeight(const std::string& val);
+  static CssTextDecoration interpretDecoration(const std::string& val);
+  static float interpretLength(const std::string& val, float emSize = 16.0f);
```
ttil: Consider increasing this size. To my eye, 32 looks closer to what I see in other EPUB viewers. But I also recognize that this might be a matter of preference.

ttil (commented):
I've picked up the latest changes and it looks great so far on my books! Thank you so much for driving this. Would love to see this merged.

lib/Epub/Epub/css/CssParser.h
jdk2pq: Thanks @ttil! I updated this to use the current font selection to determine the line height, and then we'll use that as the default `emSize` instead of assigning it to 16 by default. That should help with dynamically scaling the length if people choose a smaller/larger text size as well.

ttil: Checked it out and it looks great.

lib/GfxRenderer/GfxRenderer.cpp
```diff
@@ -462,6 +462,20 @@ int GfxRenderer::getSpaceWidth(const int fontId) const {
...
+    return 0;
+  }
+
+  uint32_t cp;
+  int width = 0;
+  while ((cp = utf8NextCodepoint(reinterpret_cast<const uint8_t**>(&text)))) {
+    width += fontMap.at(fontId).getGlyph(cp, EpdFontFamily::REGULAR)->advanceX;
+  }
+  return width;
+}
```
daveallie: This should be renamed to something like `getTextAdvanceX` or similar, it's not clear what it's currently doing. It's similar but a bit different to `getTextWidth` as it counts cursor movement instead of width.

lib/Epub/Epub/blocks/TextBlock.h
```diff
@@ -22,15 +23,30 @@ class TextBlock final : public Block {
   std::list<std::string> words;
   std::list<uint16_t> wordXpos;
   std::list<EpdFontFamily::Style> wordStyles;
+  std::list<bool> wordUnderlines;  // Track underline per word
   Style style;
+  BlockStyle blockStyle;
```
daveallie, line 18: We have moved from two sets of styles to 4, however we still have two kinds of style types:
1. Per-word styles 
   - Applied to every word
   - Currently `wordStyles` (regular, bold, italic, bold italic)
2. TextBlock styles
   - Applied to the whole text block (usually any block element)
   - Just currently contains the text alignment

One notable limitation is if any non-block HTML element is rendered with a style only available in `Style` or more relevantly `BlockStyle`, then we won't be able to apply it. I don't think this is a big problem, as we'd need to parse and understand the CSS `display` attribute which can be future expansion. So for now assuming that only block elements can contain styles for margin/padding/alignment is fine.

Aside from the limitation (which can be ignored for now), I think we should combine these 4 styles into 2 items, we can do that by:
1. Remapping `EpdFontFamily::Style` to be more of a bitflag setup, so `1 = Bold`, `2 = Italic`, `4 = Underline`
   - Combining these into `Bold | Underline` (5) or `Bold | Italic | Underline` (7) as examples will mean we can merge these styles without a whole lot of changes.
2. Merging BlockStyle and Style into a single struct.
   - Even simpler, merging will simplify logic for parsing content around.

lib/Epub/Epub.cpp
```diff
@@ -213,6 +265,8 @@ bool Epub::load(const bool buildIfMissing) {
 
   // Try to load existing cache first
   if (bookMetadataCache->load()) {
+    // Parse CSS files from loaded cache
+    parseCssFiles();
```
daveallie: I think there is value in caching these consolidated CSS rules as a file in the EPUB crosspoint cache directory. At the moment this will run and parse every CSS file every time the epub is loaded even if just for loading the cover for the sleep screen (or thumbnail). Would be saving and loading `rulesBySelector_` (afaict), and I think a separate file should possibly be used for this.

lib/Epub/Epub/ParsedText.cpp
```diff
@@ -334,55 +379,91 @@ void ParsedText::extractLine(const size_t breakIndex, const int pageWidth, const
...
-    lineWordWidthSum += wordWidths[i];
+  size_t actualGapCount = 0;
+  auto countWordIt = words.begin();
+
+  for (size_t wordIdx = 0; wordIdx < lineWordCount; wordIdx++) {
+    lineWordWidthSum += wordWidths[lastBreakAt + wordIdx];
+    // Count gaps: each word after the first creates a gap, unless it's attaching punctuation
+    if (wordIdx > 0 && !isAttachingPunctuationWord(*countWordIt)) {
+      actualGapCount++;
+    }
```
daveallie: Can you please split this one out for some testing, I believe it'll address some of https://github.com/crosspoint-reader/crosspoint-reader/issues/182. However there are likely some wider changes which could be done to avoid adding spacing after inline tags.

lib/Epub/Epub/BookMetadataCache.cpp
```diff
@@ -132,6 +137,11 @@ bool BookMetadataCache::buildBookBin(const std::string& epubPath, const BookMeta
   serialization::writeString(bookFile, metadata.language);
   serialization::writeString(bookFile, metadata.coverItemHref);
   serialization::writeString(bookFile, metadata.textReferenceHref);
+  // CSS files
+  serialization::writePod(bookFile, static_cast<uint16_t>(metadata.cssFiles.size()));
+  for (const auto& css : metadata.cssFiles) {
+    serialization::writeString(bookFile, css);
+  }
```
daveallie: If we serialize the parsed CSS into our own file, I don't think we'd need to write back the CSS files to the metadata file right

lib/Epub/Epub/css/CssParser.cpp
```diff
@@ -0,0 +1,527 @@
...
+namespace {
+
+// Buffer size for reading CSS files
+constexpr size_t READ_BUFFER_SIZE = 512;
+
+// Maximum CSS file size we'll process (prevent memory issues)
+constexpr size_t MAX_CSS_SIZE = 64 * 1024;
+
+// Check if character is CSS whitespace
+bool isCssWhitespace(const char c) { return c == ' ' || c == '\t' || c == '\n' || c == '\r' || c == '\f'; }
```
daveallie, line 17: TIL: `\f` is form feed

lib/Epub/Epub/css/CssParser.cpp
```diff
@@ -0,0 +1,527 @@
...
+
+// Read entire file into string (with size limit)
+std::string readFileContent(FsFile& file) {
```
daveallie, line 20: Can be done in a future PR but as CSS files are linear, then end of style blocks or rule lines should be enough to partially process the content, meaning we could process it in chunks without loading the whole file.

lib/Epub/Epub/css/CssParser.cpp
```diff
@@ -0,0 +1,527 @@
...
+  const size_t bodyStart = bracePos + 1;
+  size_t bodyEnd = bodyStart;
+
+  while (bodyEnd < css.size() && depth > 0) {
+    if (css[bodyEnd] == '{')
+      ++depth;
+    else if (css[bodyEnd] == '}')
+      --depth;
+    ++bodyEnd;
+  }
```
daveallie, line 131: This could be a little sketchy if there are brace characters in string literals in the CSS values, a similar thing exists in your `parseDeclarations` implementation. I think this is pretty ok for now as it's relatively defensively coded, but may need a bit of a refactor where we better tokenize rules as part of the parsing.

lib/Epub/Epub/css/CssParser.cpp
```diff
@@ -0,0 +1,527 @@
...
+      const auto values = splitWhitespace(propValue);
+      if (!values.empty()) {
+        const CssLength top = interpretLength(values[0]);
+        const CssLength right = values.size() >= 2 ? interpretLength(values[1]) : top;
+        const CssLength bottom = values.size() >= 3 ? interpretLength(values[2]) : top;
+        const CssLength left = values.size() >= 4 ? interpretLength(values[3]) : right;
+        style.marginTop = top;
+        style.marginRight = right;
+        style.marginBottom = bottom;
+        style.marginLeft = left;
```
daveallie: Is there a reason for intermediary values here? Could we just assign directly to style?

lib/Epub/Epub/css/CssParser.cpp
```diff
@@ -0,0 +1,527 @@
...
+      const auto values = splitWhitespace(propValue);
+      if (!values.empty()) {
+        const CssLength top = interpretLength(values[0]);
+        const CssLength right = values.size() >= 2 ? interpretLength(values[1]) : top;
+        const CssLength bottom = values.size() >= 3 ? interpretLength(values[2]) : top;
+        const CssLength left = values.size() >= 4 ? interpretLength(values[3]) : right;
+        style.paddingTop = top;
+        style.paddingRight = right;
+        style.paddingBottom = bottom;
+        style.paddingLeft = left;
```
daveallie: Similar to above, can we collapse the `style.padding*` assignment?

lib/Epub/Epub/BookMetadataCache.cpp
jdk2pq: Yep, I think you're right and that makes sense to me. I'll revert this logic, and we'll just go with serializing the parsed CSS 👍

jdk2pq:
> Awesome work @jdk2pq! Left some comments inline, but overall looking really promising!

@daveallie Thank you for the thorough review! I've addressed everything from the review, and hopefully simplified things along the way too, and I'll plan to put up another PR for the spurious spacing issue later this week.

jdk2pq:
Coming back to this today, I felt like my singular giant commit addressing the feedback was a bit challenging to read through all at once, so I broke it up into smaller commits and force-pushed over top of my changes from yesterday. Exact same content and changes from yesterday for anyone who was partially through reviewing, but now it's hopefully a bit easier to read through.

daveallie:
Thanks for your contribution here @jdk2pq, I've sent you an invite to add you to the maintainers team if you're interested in helping out with directing the firmware and managing PRs and issues.

PR #413, opened 2026-01-18
feat: Add custom font selection from SD card

eunchurn:
## Summary

* **What is the goal of this PR?** Add custom font selection feature that allows users to load and use custom fonts from SD card for EPUB/TXT reading.
  - New `FontSelectionActivity` for browsing and selecting fonts from `/.crosspoint/fonts/` directory
  - `SdFont` and `SdFontFamily` classes for loading `.epdfont` files from SD card with LRU caching
  - `UnifiedFontFamily` abstraction to handle both flash and SD fonts transparently in `GfxRenderer`
  - Dynamic font reloading without device reboot
  - Reader cache invalidation when font changes to ensure proper layout
  - Binary search optimization for TXT word wrap (improves indexing performance with SD fonts)
  - Page bounds clamping when font change causes different page counts

## Additional Context

* The custom fonts use the `.epdfont` binary format which supports:
  - 2-bit antialiasing for smooth text rendering on e-ink displays
  - On-demand glyph loading with LRU cache (memory-optimized for ESP32-C3's ~320KB RAM)
  - Hash-based font ID generation for proper cache invalidation when switching fonts
* Memory impact: GlyphMetadataCache (128 entries, ~3KB) + GlyphBitmapCache (32KB shared)
* Font converter tool is available in a separate repository for generating `.epdfont` files from TTF/OTF
* I've deployed the **epdfont converter** at <https://epdfont.clev.app>.

This PR started as a great alternative to the excellent XTEink X4 firmware provided by CrossPoint.
However, one notable limitation was that **a feature available in the store firmware — custom font support — was missing here.**

For multilingual users, this meant modifying the source code directly.
To address this gap, I implemented **custom font support along with a font converter**, allowing users to generate and use their own font files without touching the code.

I tested this extensively within the Korean community using various EPUB and TXT files, and fixed several minor bugs during the process.
If you test fonts using this PR and encounter any issues, please leave a comment here.

The font converter is available at:
<https://epdfont.clev.app>
I built and published this tool myself, and any feedback after testing would be greatly appreciated.

For more advanced or fine-grained font generation, the recommended approach is to create and use a custom Python script.

A reference implementation is available here:
<https://github.com/eunchurn/x4-epdfont-converter/blob/main/reference/ttf_to_epdfont.py>

---

### AI Usage

### DIFF
--- a/src/activities/reader/TxtReaderActivity.cpp
+++ b/src/activities/reader/TxtReaderActivity.cpp
@@ -32,7 +32,7 @@ size_t findUtf8Boundary(const std::string& str, size_t pos) {
 
 // Binary search to find max characters that fit in width
 // Returns the position (byte offset) where to break the string
-size_t findBreakPosition(GfxRenderer& renderer, int fontId, const std::string& line, int maxWidth) {
+size_t findBreakPosition(const GfxRenderer& renderer, int fontId, const std::string& line, int maxWidth) {
   if (line.empty()) return 0;
 
   // First check if the whole line fits


kylejw1:
I'm not a maintainer just commenting as a curious user.

It seems that bold/italics aren't loaded unless I'm mistaken.

Also I think /fonts would be a better font directory as the dot directory tends to be hidden in operating system dialogs.

Functionally this seems to work excellent and performs quite well

eunchurn:
> I'm not a maintainer just commenting as a curious user.
> 
> It seems that bold/italics aren't loaded unless I'm mistaken.
> 
> Also I think /fonts would be a better font directory as the dot directory tends to be hidden in operating system dialogs.
> 
> Functionally this seems to work excellent and performs quite well

Thanks for the feedback!

I’m not a maintainer — just commenting and contributing as a user — but I’ll take another look at the font style loading issue (bold / italics / bold-italics) and follow up with an additional commit. It’s possible I missed something there.

If you happen to know which part of the code might be related, I’d really appreciate a quick review or pointer, and I’ll be happy to reflect it.

Also, I agree about the font directory. In addition to .crosspoint/fonts, I’ll update the logic to also scan a root-level fonts directory and include any .epdfont files found there, since dot-prefixed directories can be inconvenient in OS file dialogs.

Thanks again — glad to hear it’s working well and performing nicely.

src/main.cpp
```diff
@@ -131,6 +132,67 @@ EpdFont ui12RegularFont(&ubuntu_12_regular);
...
+  if (!SdMan.exists(fontPath)) {
+    Serial.printf("[%lu] [FNT] Custom font file not found: %s\n", millis(), fontPath);
+    // Clear invalid font path
+    SETTINGS.customFontPath[0] = '\0';
+    SETTINGS.saveToFile();
+    return false;
+  }
+
+  // Create SdFontFamily for the custom font
+  SdFontFamily* font = new SdFontFamily(fontPath);
```
kylejw1, line 158: I see the font being created here but with only one font path which is why I believe it is not loading bold/italic.  Perhaps I'm reading wrong and the font is initialized fully elsewhere.

I also am not totally sure how the font styles should be created.  What I attempted was to name them with a suffix like "Bold.epdfont" "Italic.epdfont" but that was purely a guess

eunchurn: This is already implemented in
https://github.com/ruby-builds/crosspoint-reader/tree/main

The current PR will be closed, and a new PR with Ruby(@ruby-builds)’s implementation will be opened instead.

Regarding font styles: at the moment, the reader initializes the font from a single .epdfont path, so bold/italic are not loaded as separate variants. Simply naming files with suffixes like Bold.epdfont or Italic.epdfont won’t work unless the font-loading logic explicitly maps and initializes those styles. That part will be handled in the upcoming Ruby PR.

eunchurn:
Thanks for the work on this PR — the approach and implementation make sense, and it’s really nice work.

That said, the same functionality is already implemented here:
https://github.com/ruby-builds/crosspoint-reader/tree/main

To avoid duplication and keep the implementation consistent, I’m going to close this PR.
A new PR based on Ruby(@ruby-builds)’s implementation will be opened shortly.

Thanks again for the contribution!

PR #331, opened 2026-01-12
feat: adding categories to settings screen

DJPoulter:
## Summary

* **What is the goal of this PR?** (e.g., Fixes a bug in the user authentication module, Implements the new feature for 
  file uploading.)

As we get more settings, I think it makes sense to do categories for them. This just allows users to find the settings easier and navigate to them. 

* **What changes are included?**

## Additional Context

* Add any other information that might be helpful for the reviewer (e.g., performance implications, potential risks, specific areas to focus on).

PR #417, opened 2026-01-18
feat: Add settings menu to epub reader

GenesiaW:
## Summary

  - Implements a menu in EpubReaderActivity class to allow quick access to settings while reading
  - Menu can be extended in the future if footnotes/bookmarks features are added

 1. Changed behaviour of long press in EpubReaderActivity to show a menu instead of all chapters directly
 2. Chapter Selection is moved into the menu instead
 3. Added EpubReadMenuActivity class

photo_2026-01-19_02-12-19

## Additional Context

  specific areas to focus on).

1. When settings is accessed via the quick access menu, page number is set to 0 and epub section will be reset. This helps to prevent cases of out of bound due to reader setting changes
2. Point 1 can be improved in the future once a method to store relative progress is implemented
3. It is my first time writing C++, mainly a Java/Python dev. Not sure if I made any mistakes

---

### AI Usage

### DIFF
--- a/src/activities/reader/EpubReaderActivity.cpp
+++ b/src/activities/reader/EpubReaderActivity.cpp
@@ -7,6 +7,7 @@
 
 #include "CrossPointSettings.h"
 #include "CrossPointState.h"
+#include "EpubReaderChapterSelectionActivity.h"
 #include "EpubReaderMenuActivity.h"
 #include "MappedInputManager.h"
 #include "ScreenComponents.h"
@@ -114,8 +115,8 @@ void EpubReaderActivity::loop() {
     return;
   }
 
-  // Enter chapter selection activity
-  if (mappedInput.wasReleased(MappedInputManager::Button::Confirm)) {
+  // Enter shortcut menu activity
+  if (mappedInput.wasReleased(MappedInputManager::Button::Confirm) && mappedInput.getHeldTime() >= skipChapterMs) {
     // Don't start activity transition while rendering
     xSemaphoreTake(renderingMutex, portMAX_DELAY);
     exitActivity();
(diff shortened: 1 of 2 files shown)


simsudo:
I think that with normal use, you need to open the list of chapters more often than the settings menu. An extra step will complicate the UX

eunchurn:
> I think that with normal use, you need to open the list of chapters more often than the settings menu. An extra step will complicate the UX

I see the point, and I agree that chapter navigation is something readers will likely use more often than global settings.

That said, I think this feature itself is quite promising. Rather than using the menu mainly as a shortcut to the device- or app-wide settings, it would be especially valuable if it focused on **reader-specific actions that make sense in the reading context**.

For example, having quick access to things like:
- Go To Page
- Bookmarks
- Other reader-only actions (e.g. re-indexing or navigation-related tools)

would feel very natural here. These are functions users often want while reading, and exposing them directly from the reader screen would be a strong UX improvement compared to jumping into global settings.

In that sense, keeping the chapter list as the default entry point, while placing these reader-scoped actions in the menu, sounds like a good balance between usability and extensibility.

GenesiaW:
> I think that with normal use, you need to open the list of chapters more often than the settings menu. An extra step will complicate the UX

My initial idea was to do something similar to the official firmware, where the behaviour was:
- Short press: menu
- Long press: add a bookmark

However as mentioned, I think the extra step does complicate the UX, hence I will be making these changes instead:
- Short press; chapters list
- Long press: menu

I believe the best scenario would be to have a setting which allows user to choose the behaviour of button press, but that can probably be done in the future after clean up of the settings

> That said, I think this feature itself is quite promising. Rather than using the menu mainly as a shortcut to the device- or app-wide settings, it would be especially valuable if it focused on **reader-specific actions that make sense in the reading context**.

Indeed, reader-specific actions/settings would be better than device-wide settings, I think that can be improved upon if #331 or other version of settings clean up is merged

daveallie:
> I think that with normal use, you need to open the list of chapters more often than the settings menu. An extra step will complicate the UX

I'm not so worried about this, changing chapters can afford the additional keypress imo.

> My initial idea was to do something similar to the official firmware, where the behaviour was:
> 
> - Short press: menu
> - Long press: add a bookmark
>
> However as mentioned, I think the extra step does complicate the UX, hence I will be making these changes instead:
>
> - Short press; chapters list
> - Long press: menu

To me, the former makes a lot more sense than the latter. Short press should be the user's first port of call, and we should reserve long-press for things which skip deeper into menu, fast-track actions, etc. Your current implementation inverts that thinking and expects users to use the long press to get to core functionality, which I'd like to avoid. I think the two presses to get to chapter select is low enough friction that that users should be able to deal with it without much pain. Especially as this is how the OFW handles it.

Please go back towards your initial idea, putting the chapter select under the menu only, and considering the long press for bookmark or other functionality in the future.

Just as an FYI, similar work is being done here: https://github.com/crosspoint-reader/crosspoint-reader/pull/433

GenesiaW:
Thanks for the code review! I will be closing this PR to prevent merge conflict, as the other PR has the similar functionality

PR #433, opened 2026-01-19
feat: Add reading menu and delete cache function

jonasdiemer:
## Summary

* Adds a menu in the Epub reader
* The Chapter selection is moved there to pos 1 (so it can be reached by double tapping the confirm button)
* A Go Home is there, too
* Most significantly, a function "Delete Book Cache" is added. This returns to main (to avoid directly rebuilding cached items, eg. if this is used to debug/develop other areas - and it's also easier ;))

Probably, the Sync function could now be moved from the Chapter selection to this menu, too.

---

### AI Usage

### DIFF
--- a/src/activities/reader/EpubReaderActivity.cpp
+++ b/src/activities/reader/EpubReaderActivity.cpp
@@ -283,6 +283,7 @@ void EpubReaderActivity::onReaderMenuConfirm(EpubReaderMenuActivity::MenuAction 
         // We use the current variables that track our position
         uint16_t backupSpine = currentSpineIndex;
         uint16_t backupPage = section->currentPage;
+        uint16_t backupPageCount = section->pageCount;
 
         section.reset();
         // 3. WIPE: Clear the cache directory
@@ -291,7 +292,7 @@ void EpubReaderActivity::onReaderMenuConfirm(EpubReaderMenuActivity::MenuAction 
         // 4. RESTORE: Re-setup the directory and rewrite the progress file
         epub->setupCacheDir();
 
-        saveProgress(backupSpine, backupPage);
+        saveProgress(backupSpine, backupPage, backupPageCount);
       }
       exitActivity();
       updateRequired = true;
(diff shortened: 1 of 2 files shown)


src/activities/reader/EpubReaderActivity.cpp
```diff
@@ -121,30 +122,88 @@ void EpubReaderActivity::loop() {
...
+              // 2. Trigger the reader's "Go Home" callback
+              if (onGoHome) {
+                onGoHome();
+              }
+
+              break;
+            }
+            case EpubReaderMenuActivity::MenuAction::DELETE_CACHE: {
+              xSemaphoreTake(renderingMutex, portMAX_DELAY);
+              if (epub) {
```
lgarbarini: Looks like this is quite similar to my PR (https://github.com/crosspoint-reader/crosspoint-reader/pull/393).

I don't really care too much which gets merged as I just want an easy way to clear cache :smiley: 

That said, I'll advocate for my use case on here as well where I think I'd appreciate an option not to preserve reading progress as part of the purpose of cache clearing is for when I change the underlying ePub content and so preserving reading position will drop me in a random location in the book.

jonasdiemer: Yeah, two similar but different use cases, I would also like to keep both. I think the reading menu makes sense anyways (even just to have the sync option there and not on the TOC).

lukestein:
This PR fixes #475 (and presumably also fixes #477… THANK YOU!), but also does other important stuff with broader UX implications.

I hope @jonasdiemer won't mind that I've split his TOC/kosync bug fix code from here into a separate PR.

src/activities/reader/EpubReaderActivity.cpp
```diff
@@ -121,30 +122,88 @@ void EpubReaderActivity::loop() {
...
-          if (currentSpineIndex != newSpineIndex) {
-            currentSpineIndex = newSpineIndex;
-            nextPageNumber = 0;
-            section.reset();
+        [this](EpubReaderMenuActivity::MenuAction action) {  // On Select
+          switch (action) {
+            case EpubReaderMenuActivity::MenuAction::SELECT_CHAPTER: {
+              // Calculate values BEFORE we start destroying things
+              const int currentP = section ? section->currentPage : 0;
+              const int totalP = section ? section->pageCount : 0;
```
daveallie: Following examples in other activities, I would prefer to flatten this out a bit, exposing the actions from the menu as private functions with `EpubReaderActivity` and then passing them into `EpubReaderMenuActivity` to be called. This would help a lot with readability, the nesting of callbacks within callbacks is pretty hard to follow.

jonasdiemer: Thanks for taking the time to review and suggest this improvement. Fixed in 1cbe081.

jonasdiemer:
Good catch, @daveallie - unified this (incl. the position of the Chapter selection title, which was slightly off). Maybe we should have global constants for title position, menu offsets, fonts, ...

src/activities/reader/EpubReaderActivity.cpp
```diff
@@ -419,9 +482,11 @@ void EpubReaderActivity::renderScreen() {
     data[5] = (section->pageCount >> 8) & 0xFF;
```
daveallie: `section` gets reset on line 287, so this crashes

jonasdiemer: Sorry this slipped through (during rebase). Pushed a quick fix, but wondering if it would be better to re-initialize section.

pablohc:
Hi @jonasdiemer, in version 0.15 I was working on what I call "Magic Key." The BACK button on the HOME screen functions as a wildcard, allowing you to define these types of functions to simplify debugging tasks or any frequent function with just hold Back button. I've been working with other functions and I need to double-check that everything is working correctly in 0.16 before making a PR, but its usefulness will depend on whether there are already integrated functions in the system to be able to.

jonasdiemer:
> Hi @jonasdiemer, in version 0.15 I was working on what I call "Magic Key." The BACK button on the HOME screen functions as a wildcard, allowing you to define these types of functions to simplify debugging tasks or any frequent function with just hold Back button. I've been working with other functions and I need to double-check that everything is working correctly in 0.16 before making a PR, but its usefulness will depend on whether there are already integrated functions in the system to be able to.

Hi @pablohc. Not sure if this is a question... I would probably prefer explicit "debug" functions in menus rather than a magic key (if you were asking for my opinion ;) ).

PR #447, opened 2026-01-20
feat: keep power button duration in RTC_DATA_ATTR

osteotek:
## Summary

* keep power button duration required to wakeup in RTC_DATA_ATTR to avoid reading from flash on every button press

---

### AI Usage

Did you use AI tools to help write this code? _**< PARTIALLY>**_

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -43,6 +43,8 @@ MappedInputManager mappedInputManager(inputManager);
 GfxRenderer renderer(einkDisplay);
 Activity* currentActivity;
 
+RTC_DATA_ATTR uint16_t rtcPowerButtonDurationMs = 400;
+
 // Fonts
 EpdFont bookerly14RegularFont(&bookerly_14_regular);
 EpdFont bookerly14BoldFont(&bookerly_14_bold);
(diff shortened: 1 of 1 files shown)


osteotek:
This only works when the device is connected to USB 😭 . Apparently, it disconnects the battery when going into deep sleep, and RTC attributes are probably cleared.

DavidOrtmann:
> This only works when the device is connected to USB 😭 . Apparently, it disconnects the battery when going into deep sleep, and RTC_DATA is probably cleared.

I think Larry found a solution for that: https://github.com/bitbank2/bb_epaper/commit/2404ab610081e8fe91029ef0dad9c4284424f785

osteotek:
> > This only works when the device is connected to USB 😭 . Apparently, it disconnects the battery when going into deep sleep, and RTC_DATA is probably cleared.
> 
> I think Larry found a solution for that: [bitbank2/bb_epaper@2404ab6](https://github.com/bitbank2/bb_epaper/commit/2404ab610081e8fe91029ef0dad9c4284424f785)

I wonder how much it affects battery life in sleep mode.

ngxson:
Light sleep does affect battery life significantly as it's ~20 times more power consuming than deep sleep.

Just wondering though, the partition table does have spiffs, why can't we use it to store or cache some settings?

PR #460, opened 2026-01-20
feat: add new configuration for front buttons, more usable on landscape ccw

borisfaure:
When reading on Landscape Counter ClockWise mode, the left/right button appear inverted: the upper button (left) goes down and the lower button (right) goes up.

Discussion: #449

## Summary

Add a new configuration for the front buttons: Back, Confirm, Right, Left

---

### AI Usage

### DIFF
--- a/src/MappedInputManager.cpp
+++ b/src/MappedInputManager.cpp
@@ -14,6 +14,9 @@ decltype(InputManager::BTN_BACK) MappedInputManager::mapButton(const Button butt
         case CrossPointSettings::LEFT_BACK_CONFIRM_RIGHT:
           return InputManager::BTN_CONFIRM;
         case CrossPointSettings::BACK_CONFIRM_LEFT_RIGHT:
+          /* fall through */
+        case CrossPointSettings::BACK_CONFIRM_RIGHT_LEFT:
+          /* fall through */
         default:
           return InputManager::BTN_BACK;
       }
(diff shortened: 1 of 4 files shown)


src/MappedInputManager.cpp
```diff
@@ -5,12 +5,13 @@
...
   const auto frontLayout = static_cast<CrossPointSettings::FRONT_BUTTON_LAYOUT>(SETTINGS.frontButtonLayout);
   const auto sideLayout = static_cast<CrossPointSettings::SIDE_BUTTON_LAYOUT>(SETTINGS.sideButtonLayout);
+  const bool is_landscape_ccw = SETTINGS.orientation == CrossPointSettings::ORIENTATION::LANDSCAPE_CCW;
 
   switch (button) {
     case Button::Back:
       switch (frontLayout) {
         case CrossPointSettings::LEFT_RIGHT_BACK_CONFIRM:
-          return InputManager::BTN_LEFT;
+          return is_landscape_ccw ? InputManager::BTN_RIGHT : InputManager::BTN_LEFT;
```
JakeLDev: Could this be configurable in the settings instead? I use CCW too, but am now used to pressing "up" to go to the next page when holding the device with the buttons on the right. Would rather this be an option than have it be forcibly changed.

maeveynot: Agreed. I have my Kobo Libra 2 page turn buttons set to "inverted" and do not want to deal with different button orderings on each device.

osteotek:
This can be done as another option for front button layout here - https://github.com/crosspoint-reader/crosspoint-reader/blob/master/src/CrossPointSettings.h#L35

borisfaure:
> This can be done as another option for front button layout here - https://github.com/crosspoint-reader/crosspoint-reader/blob/master/src/CrossPointSettings.h#L35

You're right. I'll do it that way tonight.

src/MappedInputManager.cpp
```diff
@@ -24,24 +27,35 @@ decltype(InputManager::BTN_BACK) MappedInputManager::mapButton(const Button butt
...
           return InputManager::BTN_LEFT;
       }
     case Button::Right:
       switch (frontLayout) {
         case CrossPointSettings::LEFT_RIGHT_BACK_CONFIRM:
           return InputManager::BTN_CONFIRM;
+        case CrossPointSettings::BACK_CONFIRM_RIGHT_LEFT:
+          return InputManager::BTN_LEFT;
         case CrossPointSettings::BACK_CONFIRM_LEFT_RIGHT:
+          /* fall through */
```
JakeLDev, line 56: Not sure all these comments add much, but otherwise looks good to me 😁👍🏻

PR #476, opened 2026-01-21
feat: Add support to B&W filters to image covers

el:
## Summary

Implementation of a new feature in Display options as Image Filter
Black & White and Inverted Black & White options are added.

## Additional Context

Here are some examples:

| None | Contrast | Inverted |
| --- | --- | --- |
| image | image | image |
| image | image | image |

  specific areas to focus on).

I have also tried adding Color inversion, but could not see much difference with that. It might be because my implementation was wrong.

---

### AI Usage


lukestein:
I like this. Thanks! (Pull request title should start with "`feat:`" to pass ci tests btw)

Eloren1:
```
  SettingInfo::Enum("Sleep Screen Cover Filter", &CrossPointSettings::sleepScreenCoverFilter,
                      {"No filter", "B&W", "Inverted B&W"}),
```

Not sure if such long names are fine, also it's already B&W... Maybe something like "None / Contrast / Inverted".

el:
> ```
>   SettingInfo::Enum("Sleep Screen Cover Filter", &CrossPointSettings::sleepScreenCoverFilter,
>                       {"No filter", "B&W", "Inverted B&W"}),
> ```
> 
> Not sure if such long names are fine, maybe something like "None / Mono / Inverted". But then... It's already B&W or mono - confusing.

Seems to be fine on the screen?

image

Eloren1:
Caps Locked B&W is a bit distracting, and it's already B&W as I mentioned, calling it Contrast would be very nice. "None" - based on other settings naming. "Inverted" balances "Contrast". Like they all match together. Small nitpicks though, overall great result

PR #483, opened 2026-01-22
feat: add bluetooth file transfers

aleksejspopovs:
## Summary

This PR adds support for a very simple [custom BLE protocol](https://github.com/aleksejspopovs/longform/blob/main/app/docs/BLE%20Protocol.md) for receiving files from a phone. It is intended to be used with an Android app I'm also working on called [Longform](https://github.com/aleksejspopovs/longform) that can send arbitrary files over to the X4 but can also use the Accessibility API to scrape text from any app on your phone, so that you can quickly send an article from any app to your reader.

Here's a 35s demo:

https://github.com/user-attachments/assets/e4f54e89-335b-47e7-888b-db0e7e43e245

I figure this is not quite ready to merge yet given that Longform is not yet available on the Play Store, but getting there might still take me a bit because there's still a bit of polishing to be done and probably some work to be compliant with the store's requirements for apps using Accessibility APIs. So I wanted to share this branch here and see if there's any interest in this in the first place, or any feedback (is there a more standard BLE protocol I should be using here? did I fundamentally misunderstand anything about the architecture of Crosspoint?).

## Additional Context

* the ESP32-C3 only supports BLE, not Bluetooth Classic, so we cannot implement the Object Push Profile, the usual Bluetooth file transfer protocol
* framework-arduinoespressif32 comes with a Bluedroid-based BLE stack included, but it uses a lot of Flash/RAM and has a weird callback implementation that makes it impossible to actually use BLE Write With Response as intended. That's why I pull in the lighter (and supposedly more performant) h2zero/NimBLE-Arduino.

---

### AI Usage

The code in this PR was written by hand with some AI autocomplete. (The Android app, which is not part of this PR, is 100% vibe-coded.)

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -48,6 +48,7 @@ lib_deps =
   bblanchon/ArduinoJson @ 7.4.2
   ricmoo/QRCode @ 0.0.1
   links2004/WebSockets @ 2.7.3
+  h2zero/NimBLE-Arduino @ 2.3.7
 
 [env:default]
 extends = base
--- a/src/activities/bluetooth/BluetoothActivity.cpp
+++ b/src/activities/bluetooth/BluetoothActivity.cpp
@@ -260,7 +260,7 @@ void BluetoothActivity::onConnected(bool isConnected) {
   intoState(isConnected ? STATE_CONNECTED : STATE_WAITING);
 }
 
-void BluetoothActivity::onRequest(lfbt_message* msg, size_t msg_len) {
+void BluetoothActivity::onRequest(const lfbt_message* msg, size_t msg_len) {
   if (state == STATE_ERROR) {
     // ignore further messages in error state
     return;
@@ -312,7 +312,7 @@ void BluetoothActivity::onRequest(lfbt_message* msg, size_t msg_len) {
         .txnId = txnId,
         .body = {.serverResponse = {.status = 0}}
       };
-      pResponseChar->setValue((uint8_t*)&response, 8 + sizeof(lfbt_msg_server_response));
+      pResponseChar->setValue(reinterpret_cast<uint8_t*>(&response), 8 + sizeof(lfbt_msg_server_response));
       pResponseChar->indicate();
 
       updateRequired = true;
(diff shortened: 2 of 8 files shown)


Uri-Tauber:
What the transfer rate you were able to achieve? I think BLE is very slow - maybe 20kbps. is it worth it when we can use Wifi for much faster and stable communication?

ngxson:
I think another thing worth consider is that the BLE stack of Arduino / esp-idf is quite large and can take up a significant amount of flash, especially when allowing BLE+Wifi to coexist.

aleksejspopovs:
> What the transfer rate you were able to achieve? I think BLE is very slow - maybe 20kbps.

I have not measured it, and the protocol I implemented is far from optimized for maximum transfer rate. Espressif [quotes](https://docs.espressif.com/projects/esp-faq/en/latest/software-framework/bt/ble.html#what-is-the-throughput-of-esp32-bluetooth-le) 700 Kbps in good conditions, fwiw. In the video you can see that transferring a short article took around 2s—far less time than it took Crosspoint to index the file.

> is it worth it when we can use Wifi for much faster and stable communication?

My experience with WiFi on the X4 has been far from stable. But the main reason I reached for BLE here is the quick and hassle-free connection: the use case I have in mind is on-the-go transmission of short articles, not syncing an entire collection, so not messing with my phone's WiFi connection/not wasting time associating and re-associating with networks seemed more worthwhile to me than shaving some time off an already acceptably fast transfer.

> I think another thing worth consider is that the BLE stack of Arduino / esp-idf is quite large

I did mention on the PR that this is why I am not using Espressif's stack, which indeed is prohibitively large. With NimBLE, this PR adds 6416 bytes to RAM use (+1.9%) and 202K to flash (+3.1%).

aleksejspopovs:
Setting this to "ready for review" as I haven't received any further feedback. I've been running this for the past few days and it's been working well. I'm now preparing a submission to the Play Store.

daveallie (changes requested):
Thanks for the submission @aleksejspopovs! In the interest of ensuring that future travelers can improve upon this (and so that Longform doesn't have to be the only implementation), can you please include some documentation about the protocol (a link back to https://github.com/aleksejspopovs/longform/blob/main/app/docs/BLE%20Protocol.md would be enough).

As you mention, Longform isn't available yet, and while I'll look over the rest of the code and am comfortable adding this functionality to CrossPoint, I'd like to wait until Longform (or another implementation) is available before this is merged.

aleksejspopovs:
Makes sense, and thanks for taking a look! I just finished the Play Store submission, Google says it will be reviewed "in under seven days, or longer" :upside_down_face: I'll let you know when that's done.

In the next few days I'll add a link to the protocol docs and rebase&retest on top of 0.16.0.

tweitzel:
Is it feasible to make this use the standard bluetooth object exchange push which has been around since 1997?

aleksejspopovs:
@tweitzel from the PR description:

> the ESP32-C3 only supports BLE, not Bluetooth Classic, so we cannot implement the Object Push Profile, the usual Bluetooth file transfer protocol

Jason2866:
Will there be an iOS App too?

Jason2866:
Interested what is free heap with and without the BLE part compiled in and used and when it is not there?

znelson:
Closing due to inactivity. There has been a lot of movement in the BLE X3/X4 space since this PR. Please open a discussion if you want to bring this forward.

PR #484, opened 2026-01-22
UX improvment to Forget Network page

### LINKED ISSUE #427: "Forget Network" screen would benefit from UX improvement
I often wind up on the "Forget Network" screen because my wifi is weak and the x4 has trouble connecting to it.

It is very easy to select "remove saved password" → "yes" by mistake, since it's the (dangerous) default. And there are no button tips.

I suggest that rather than being Yes/No the choices be more like "Back/Try Again" and "Forget Network" with the former as default. And button tips based on current button settings.

image

olearycrew:
## Summary

On the Forget Network page

* Update the default option to be DON'T forget the network
* Make the options clearer ("Cancel" and "Forget network")
* Unify the button hints to match the rest of the UI

## Additional Context

Closes #427

---

### AI Usage

Did you use AI tools to help write this code? PARTIALLY

### DIFF
--- a/src/activities/network/WifiSelectionActivity.cpp
+++ b/src/activities/network/WifiSelectionActivity.cpp
@@ -677,7 +677,7 @@ void WifiSelectionActivity::renderConnectionFailed() const {
 
   renderer.drawCenteredText(UI_12_FONT_ID, top - 20, "Connection Failed", true, EpdFontFamily::BOLD);
   renderer.drawCenteredText(UI_10_FONT_ID, top + 20, connectionError.c_str());
-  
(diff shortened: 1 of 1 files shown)


lukestein:
Thanks for this @olearycrew. Tested a bit and functionality seems to work perfectly.

Small suggestion that the text is slightly duplicative and confusing to have two questions ("Forget Network?" and "Remove saved password?").
IMG_2972

Maybe change header to something like "**Connection Failed**" and the small text below the network name to "Forget network and remove saved password?"
IMG_2974

Issue #220, opened 2026-01-03
Page number should update after font size change (to maintain approximately same place in chapter, and avoid out-of-bounds)


lukestein:
**Affected Version**

0.11.2

**Describe the Bug**

After changing font size, reader mode tries to stay on the same numbered page rather than the same location in text (or approximating based on the fraction of the way through the current chapter).

This results in moving to a different part of the text, and perhaps an "out of bounds" error if the font was made smaller and reader was previously near end of a chapter.

**Expected Behavior**

After changing font size, the reader re-indexes the current chapter. Should jump to approximately the same part of the text rather than keeping the literal page number (within chapter) the same.

**Relevant Log Output/Screenshots**

_Before:_

![image](https://github.com/user-attachments/assets/eda48059-31ef-4bc5-a5d2-1d789175922f)

_After shrinking font size (note page number 4/3):_

![image](https://github.com/user-attachments/assets/a9ee3f09-01e5-4ab6-9fbe-ae9a81140938)

lukestein:
I'm just noting that koreader sync code from @itsthisjustin [may do a percentage progress calculation](https://github.com/daveallie/crosspoint-reader/pull/232#issuecomment-3711656395) that could possibly be useful here?

daveallie:
Following https://github.com/crosspoint-reader/crosspoint-reader/pull/486, it's a lot more accurate, going to close this one

PR #486, opened 2026-01-22
fix: goes to relative position when reader settings are changed

GenesiaW:
## Summary

* Aims to fix Issue #220 

- Increased size of `progress.bin` such that total page count of current section can be stored
- Comparison of total page count is done to determine if reader settings were changed
- New position/page number is calculated using percentage calculated from read progress

## Additional Context

  specific areas to focus on).

---

### AI Usage


lukestein:
[Testing on device](https://github.com/crosspoint-reader/crosspoint-reader/discussions/546) and seems to work great with epubs. I've tried both rotation and font size changes. Thank you!

**Does not appear to do anything for text files.**

PR #489, opened 2026-01-22
feat: Add Power Button Hold Duration setting

Eloren1:
upd in a new PR

### DIFF
--- a/src/CrossPointSettings.h
+++ b/src/CrossPointSettings.h
@@ -56,7 +56,7 @@ class CrossPointSettings {
   enum SHORT_PWRBTN { IGNORE = 0, SLEEP = 1, PAGE_TURN = 2 };
 
   // Power button hold duration options
-  enum POWER_BTN_HOLD_DURATION { PB_FAST = 0, PB_NORMAL = 1, PB_SLOW = 2, PB_XSLOW = 3 };
+  enum POWER_BTN_HOLD_DURATION { PB_FAST = 0, PB_NORMAL = 1, PB_SLOW = 2 };
 
   // Hide battery percentage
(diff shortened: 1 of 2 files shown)


phillipspc:
I'm just a passerby with no say whatsoever in how this repo is handled/what direction it's taken. But with that caveat out of the way:
I would caution against reaching for a new setting too quickly. While giving people the flexibility to choose whatever option suits them best sounds nice, you can quickly end up with way too many settings for too many things this way. Each setting represents a slightly different branch of logic that must then be accounted for in all future changes. This can make a codebase very difficult to maintain or make changes to over time.

If the hold duration feels too short as-is, and there's consensus on that, I think the better solution would be to simply change the duration, rather than introducing a new setting for it.

Just my 2 cents!

Eloren1:
image

Sooo cluttered right? How about removing Short Power Button Click option? There are already 4 buttons for page changing so let's delete it. I can say this literally about any setting. But they are here and we use them.

DavidOrtmann:
I think we should all stay civil here and act in good faith, which is what @phillipspc did here, imo.

Eloren1:
Look, I apologize for my tone earlier. I let my passion for this topic and I regret how I presented it. This shouldn't have been a place for discussion in the first place.

phillipspc:
@Eloren1 I never saw what I assume was some now-deleted spicy comment 😄 But I apologize if my feedback came off the wrong way. Wasn't trying to get this PR closed or anything. Merely just weighing in on it with an opinion that comes through my own experiences writing and maintaining software. I didn't see a linked issue/discussion to comment on, so commenting here was the only way of doing that.
But again, I'm just a passerby with no authority here, so by all means, feel free to take or leave my advice.

PR #474, opened 2026-01-21
feat: Better skip chapters (click+hold skip and display pop-up)

Eloren1:
## Summary

- [x] Change from 0.7s to 2s for long-press value
- [x] Show "Skipping" pop-up until skip refresh / indexing
- [x] Automatically skip the chapter instead of waiting for unpress
- [x] Don't change pages during first unpress after chapter skip

## Additional

Tested in many scenarios, skipping to unindexed and indexed chapters, in epub and xtc. Holding for 2 seconds is surprisingly long enough. And now you don't need to count 1..2..3.. — just hold the button and it will skip automatically. Way better than before with 0.7 seconds and done, no indication and no feedback. 

---

### AI Usage

Did you use AI tools to help write this code? _**PARTIALLY**_

Closes #464, #389

PR #493, opened 2026-01-22
Feature/In-Reader Rotation Shortcut 

barbarhan:
Summary
What is the goal of this PR? 

Implement a shortcut to rotate the screen orientation directly from the reader view.

What changes are included?
Added a long-press (1s+) action to the CONFIRM button in EpubReaderActivity which cycles through the 4 screen orientations (Portrait → Landscape CW → Inverted → Landscape CCW).
Ensured that the current page number is preserved when the screen rotates and the view re-renders, preventing the reader from resetting to the start of the chapter.
The standard short press on CONFIRM for chapter selection is preserved.

Additional Context
This feature improves the reading experience by allowing users to quickly adapt the device orientation (e.g. for one-handed reading) without needing to exit the book and navigate through the settings menu.

AI Usage

### DIFF
--- a/src/activities/reader/EpubReaderActivity.cpp
+++ b/src/activities/reader/EpubReaderActivity.cpp
@@ -512,6 +512,11 @@ void EpubReaderActivity::rotateScreen() {
   // We don't want to change orientation mid-render, so grab the semaphore
   xSemaphoreTake(renderingMutex, portMAX_DELAY);
 
+  // If a section is loaded, preserve the current page number
+  if (section) {
+    nextPageNumber = section->currentPage;
+  }
+
   //
   // Cycle to next orientation
   //


src/activities/reader/EpubReaderActivity.cpp
```diff
@@ -17,6 +17,7 @@ namespace {
 // pagesPerRefresh now comes from SETTINGS.getRefreshFrequency()
 constexpr unsigned long skipChapterMs = 700;
 constexpr unsigned long goHomeMs = 1000;
+constexpr unsigned long rotateScreenMs = 1000;
```
Eloren1, line 20: Can connect to this Long-press parameter later: https://github.com/crosspoint-reader/crosspoint-reader/pull/474

src/activities/reader/EpubReaderActivity.cpp
```diff
@@ -116,8 +117,14 @@ void EpubReaderActivity::loop() {
     return;
   }
 
+  // Long press CONFIRM (1s+) rotates the screen
+  if (mappedInput.wasReleased(MappedInputManager::Button::Confirm) && mappedInput.getHeldTime() >= rotateScreenMs) {
```
Eloren1, line 121: Most of the functions are activated during hold: Sleep mode, Long-press home

Not during unpress state. Such functions don't require unpressing and feel more responsive

src/activities/reader/EpubReaderActivity.cpp
```diff
@@ -500,3 +507,48 @@ void EpubReaderActivity::renderStatusBar(const int orientedMarginRight, const in
...
+  }
+
+  //
+  // Cycle to next orientation
+  //
+  uint8_t newOrientation = (SETTINGS.orientation + 1) % 4;
+  SETTINGS.orientation = newOrientation;
+  SETTINGS.saveToFile();
+
+  //
```
Eloren1, line 527: 3-lined comments?

osteotek:
Instead of overloading confirm button I believe we should implement a quick settings menu for the reader screen, where we can change this and other settings (similar to https://github.com/crosspoint-reader/crosspoint-reader/pull/433) . But it can be done later

whyte-j:
Really love the idea of making the reader settings more accessible from the reader itself!

I'd be concerned with setting the page rotate as the long-press option, though. There are no tool-tips once you enter the reader, so if you were to accidentally rotate, and didn't understand what you had done to achieve that, it would be difficult to discover the rotate function and resolve it (and might even get stuck if the user isn't aware that you can rotate from the settings menu). I think it would be better to hold until the quick-menu, as in that case you also don't have to wait for the page to render as you cycle through the options.

daveallie:
+1 to the other sentiments here, this shouldn't exist on the confirm button long-press as discoverability will be low and confusion high (if accidentally activated).

osteotek:
superseded by https://github.com/crosspoint-reader/crosspoint-reader/pull/685

PR #495, opened 2026-01-22
feat: allow last content to display in sleep mode

ngxson:
## Summary

Allow the current content to continue to display in sleep mode. This is actually a feature mirrored from KOReader.

Not quite sure if this should be called "Off", "Disabled" or "Last"

A small "Sleeping" message will be shown at the bottom of the screen:

image

---

### AI Usage

### DIFF
--- a/USER_GUIDE.md
+++ b/USER_GUIDE.md
@@ -90,6 +90,7 @@ The Settings screen allows you to configure the device's behavior. There are a f
   - "Custom" - Custom images from the SD card; see [Sleep Screen](#36-sleep-screen) below for more information
   - "Cover" - The book cover image (Note: this is experimental and may not work as expected)
   - "None" - A blank screen
+  - "Last" - Preserve the last content displayed on the screen
 - **Sleep Screen Cover Mode**: How to display the book cover when "Cover" sleep screen is selected:
   - "Fit" (default) - Scale the image down to fit centered on the screen, padding with white borders as necessary
   - "Crop" - Scale the image down and crop as necessary to try to to fill the screen (Note: this is experimental and may not work as expected)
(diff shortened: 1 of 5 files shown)


ngxson:
Hmm, on second thought, maybe a better UX if we can only trigger this on timeout activity.

At least, IIRC the stock firmware display a small "moon" icon at the bottom-right corner.

Eloren1:
What if I have Display — Status Bar — None?

Eloren1:
> Hmm, on second thought, maybe a better UX if we can only trigger this on timeout activity.

Keeping both would be nice. Some people might need this every time, and I would only use it for timeouts.

ngxson:
> What if I have Display — Status Bar — None?

This currently draw on top of the status bar, so the "sleeping" should still show under this case.

> Keeping both would be nice. Some people might need this every time, and I would only use it for timeouts.

Hmm yeah maybe better to add a new setting entry called "Show sleep screen": "Always" - "Except timeout" - "Never"

osteotek:
Same functionality in another PR - https://github.com/crosspoint-reader/crosspoint-reader/pull/410

Eloren1:
Oh and also what if I use horizontal mode? In that case an icon is better. But the text variant is also nice and bold 🤔

ngxson:
@osteotek ahah this PR is almost the same as the other 😂 I'll close mine and will ask if the suggested behavior "Always" - "Except timeout" - "Never" can be implemented on that PR

@Eloren1 hmm yeah good question. I think a smaller "sleeping" text that placed in the corner should make more sense

Eloren1:
Okay let's check what happens to that PR but yours is more promising tbh

PR #498, opened 2026-01-22
feat: treat .md files as .txt

osteotek:
## Summary

* Quick fix for markdown reading - open them as txt files

### DIFF
--- a/src/activities/reader/ReaderActivity.cpp
+++ b/src/activities/reader/ReaderActivity.cpp
@@ -22,9 +22,8 @@ bool ReaderActivity::isXtcFile(const std::string& path) {
 }
 
 bool ReaderActivity::isTxtFile(const std::string& path) {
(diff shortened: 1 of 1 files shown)


lukestein:
Sorry since I'm only a very amateur reader of C++, but do we also need to do something with the file extension checks in `Txt.cpp` and `SleepActivity.cpp`?

osteotek:
> Sorry since I'm only a very amateur reader of C++, but do we also need to do something with the file extension checks in `Txt.cpp` and `SleepActivity.cpp`?

There are enough safeguards in those checks that it works for .md files without modifying.

lukestein:
[Testing on device](https://github.com/crosspoint-reader/crosspoint-reader/discussions/546) and seems to work great. 🎉 

I do note that text and markdown files do not seem to show up in the recent files tab.

PR #499, opened 2026-01-22
feat: Apply orientation to UI too, move setting to Display

maeveynot:
## Summary

As a landscape-orientation reader, I would prefer to not have to physically rotate my device to use some (but not all) other parts of the UI. This PR moves the orientation setting from the Reader category to the Device category, and applies orientation to all activities, not just the reader.

Rather than setting the orientation in the reader activity and setting it back, we change the orientation only:
* At boot
* When changing the setting (takes effect immediately)
* Back to portrait when going to sleep, since all existing wallpaper is portrait and nobody wants to change that
* Similarly, into portrait for reading XTCs, since those are all logically portrait (if they were generated as landscape, they're still 480x800 with the text sideways)

We introduce some code into the base Activity class to handle both this and the concept of "margins", which in many cases were kind of hard-coded into other activities with the assumption of portrait. With a place for some code there I then refactored menus a bit by collecting duplicated code for the calculation of number of menu items on screen. This work could be continued further to actually drawing the menu items, but I'm hoping for feedback first.

Finer-grained breakdown of changes:

* Adjust the size of the button hints to better match the physical buttons. I came up with the new width by eyeballing it. Text still fits. This was needed because otherwise the buttons would overlap with the battery display in the home activity.
* Move Activity's onEnter/onExit out of the .h file. While the renderer orientation only needs to be changed in the circumstances above, the margins are member variables of each activity instance, so they need to be initialized in onEnter based on the current orientation.
* Take the CONTENT_START_Y and LINE_HEIGHT constants from the my-library activity and set them in Activity so that all menu-drawing activities can use them instead of literal `60` and `30`. (also, instead of 60 it's now the top margin + 45.)
* The book tile label is one case where we were centering things relative to the screen, but now we have to center them relative to the tile, since in landscape there is more margin on one side.
* In the web server activity, a width of `480` was hardcoded; use the screen width instead so that the QR code is still centered in landscape.
* In the wifi selection screen, the MAC address etc labels would run over the network names list as they were, so draw them right-aligned instead (adding some space on the right to not overlap any indicators).
* In settings, draw the version number in the corner of the header instead of next to the button hints, because for landscape there are labels right next to the button hints.
* Move all keyboard entry activity keyboards down to 50 pixels, like wifi password entry (centering would be nice, but I we don't know the height when calling the constructor). Previously this was inconsistent between wifi and OPDS/KOreader settings.

Visual problems that still exist at this point:
* In the my-library activity, Landscape CCW, the button hints cover up the scroll indicator. I don't really need the scroll indicator, but maybe just not drawing it or moving the numbers to the corner would be preferable. Hoping for feedback.

Random other notes:
* I initially prototyped this work off of 0.14, where all settings were still in one screen, so I had to adapt the settings activity to scroll between pages like other menus. With the settings split into categories, this is not currently an issue, but, more settings will be added in the future and landscape orientation will run out of space before portrait does, so at some point that should be revisited.

## Additional Context

There may be unpolished spots here but since this will affect and conflict with other menu work (EDIT: there are now PRs: #528, #548) I want to get it up here and under discussion (and hopefully merged) as soon as possible, with improvements to follow.

I don't use every feature of CrossPoint, so I may have missed testing some things. I'm prioritizing attention to usability in Portrait and Landscape CCW orientations.

If we want to go in this direction, I would like to follow up with opening a discussion about remapping buttons in the new my-library activity, because I really strongly don't like the current ones. For me the outer side of the button should always be "next page or menu item" and the inner side of the button should always be "previous page or menu item" and now they switch between tabs and I have to use the side button to scroll. I am not changing this new behavior here since discussion of what mapping settings to add is needed.

### AI Usage

No

### DIFF
--- a/lib/GfxRenderer/GfxRenderer.cpp
+++ b/lib/GfxRenderer/GfxRenderer.cpp
@@ -494,11 +494,11 @@ void GfxRenderer::drawButtonHints(const int fontId, const char* btn1, const char
   setOrientation(Orientation::Portrait);
 
   const int pageHeight = getScreenHeight();
-  constexpr int buttonWidth = 106;
+  constexpr int buttonWidth = 94;
   constexpr int buttonHeight = 40;
   constexpr int buttonY = 40;     // Distance from bottom
   constexpr int textYOffset = 7;  // Distance from top of button to text baseline
-  constexpr int buttonPositions[] = {25, 130, 245, 350};
+  constexpr int buttonPositions[] = {50, 143, 247, 340};
   const char* labels[] = {btn1, btn2, btn3, btn4};
 
   for (int i = 0; i < 4; i++) {
@@ -516,7 +516,10 @@ void GfxRenderer::drawButtonHints(const int fontId, const char* btn1, const char
   setOrientation(orig_orientation);
 }
 
-void GfxRenderer::drawSideButtonHints(const int fontId, const char* topBtn, const char* bottomBtn) const {
+void GfxRenderer::drawSideButtonHints(const int fontId, const char* topBtn, const char* bottomBtn) {
+  const Orientation orig_orientation = getOrientation();
+  setOrientation(Orientation::Portrait);
+
   const int screenWidth = getScreenWidth();
   constexpr int buttonWidth = 40;   // Width on screen (height when rotated)
   constexpr int buttonHeight = 80;  // Height on screen (width when rotated)
@@ -565,6 +568,8 @@ void GfxRenderer::drawSideButtonHints(const int fontId, const char* topBtn, cons
       drawTextRotated90CW(fontId, textX, textY, labels[i]);
     }
   }
+
+  setOrientation(orig_orientation);
 }
 
 int GfxRenderer::getTextHeight(const int fontId) const {
--- a/lib/GfxRenderer/GfxRenderer.h
+++ b/lib/GfxRenderer/GfxRenderer.h
@@ -85,7 +85,7 @@ class GfxRenderer {
 
   // UI Components
   void drawButtonHints(int fontId, const char* btn1, const char* btn2, const char* btn3, const char* btn4);
-  void drawSideButtonHints(int fontId, const char* topBtn, const char* bottomBtn) const;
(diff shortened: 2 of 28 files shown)


maeveynot:
I removed the visual change where the continue-reading book tile expanded to fill available space, and kept it at half the screen size as before--we just limit the height if that's needed. By doing this we can eliminate another max height check later on.

I'm absolutely open to better dimension or a side-by-side cover and text continue-reading tile, but that can be decided later.

maeveynot:
I originally included some stuff about image orientation in the boot/sleep activity here, but I broke that out into https://github.com/crosspoint-reader/crosspoint-reader/pull/557 .

~~This branch is one commit on top of that PR's branch; the other PR should be merged first to get its own squash commit. I think this PR should squash fine after that but if needs to be rebased please feel free to do that.~~

~~If reviewing changes here before merging #557, you will see that branch included in the diff.~~

maeveynot:
Conflicts resolved and rebased on `master`. Thanks for merging #557!

maeveynot:
I have updated this to essentially undo #637 since we provide space here for the hints to be drawn. @daveallie could you let me know if it's worth continuing to do this or if you'd like to move in the direction of themeing systems instead? It seems like UI rotation will be a lot more complicated there since both the theme engine and the themes themselves would need to support it.

CaptainFrito:
Hi @maeveynot, I wrote the UI theme PR https://github.com/crosspoint-reader/crosspoint-reader/pull/528, I really dig the work you've done here to make the UI more "responsive". With the introduction of themes it will actually be easier to do this, since we're migrating to a components architecture rather than rendering hardcoded primitives on every screen. Every component is drawn using a Rect{x, y, w, h} frame parameter and layout constants are centralized in each theme. We could work together on implementing your changes into this system if you'd like.

maeveynot:
@CaptainFrito thanks! I'll take a look at what needs to be done or not now that #528 is merged, and see what I can reimplement... maybe this PR as is should be abandoned.

znelson:

PR #500, opened 2026-01-22
feat: (wip) add qemu for development

ngxson:
## Summary

> [!IMPORTANT]  
> This PR is currently on pause, see https://github.com/crosspoint-reader/crosspoint-reader/pull/500#issue-3844145247

Emulation can benefit the development of functionalities that are not directly related to hardware.

Still WIP, just posting for discussion. My plan is:
1. Add a HAL for buttons, display and sdcard
2. Write an implementation of emulated buttons, display and sdcard that forward calls to Serial output, then have a python server to forward it to a websocket connection
3. A web ui can be added to interact with the emulated machine via the websocket connection

For example: `Display::displayBuffer()` will simply turn the buffer into base64, then print it to Serial output. The python script then forward the info back into web ui.

The more complicated part will be to emulate the sdcard. Not sure yet how to do, probably emulate inside python context? (This is left open for discussion)

For now, it can already boot to home screen, but watchdog timeout is triggered randomly:

image

Note: The docker code is adapted from https://github.com/mluis/qemu-esp32

---

### AI Usage

Did you use AI tools to help write this code? **Yes, but only for python/html code, here are some prompts that I used:**
- "write stub (unimplemented) from this header"
- write python file to spawn the process and forward stdio to websocket + parent stdio
- write minimal html, no style, to receive the websocket data and output to `<div id="output">`
- decode b64 which is 1-bit bw data to canvas, rotate clockwise 90d

**All C++ code is written by hand**

### DIFF
--- a/lib/hal/EmulationUtils.h
+++ b/lib/hal/EmulationUtils.h
@@ -18,12 +18,18 @@ static const char* CMD_FS_STAT  = "FS_STAT";  // arg0: path -- return file size 
 static const char* CMD_FS_WRITE = "FS_WRITE"; // arg0: path, arg1: base64-encoded data, arg2: offset, arg3: is inplace (0/1) -- return int64_t bytes written
 static const char* CMD_FS_MKDIR = "FS_MKDIR"; // arg0: path -- return int64_t: 0=success
 static const char* CMD_FS_RM    = "FS_RM";    // arg0: path -- return int64_t: 0=success
+static const char* CMD_BUTTON   = "BUTTON";   // arg0: action ("read") -- return int64_t: HalInput state bitmask
 
 static std::string base64_encode(const char* buf, unsigned int bufLen);
 static std::vector<uint8_t> base64_decode(const char* encoded_string, unsigned int in_len);
 
+static SemaphoreHandle_t sendMutex = xSemaphoreCreateMutex();
+
 static void sendCmd(const char* cmd, const char* arg0 = nullptr, const char* arg1 = nullptr, const char* arg2 = nullptr, const char* arg3 = nullptr) {
-  Serial.printf("[%lu] [EMU] Sending command: %s\n", millis(), cmd);
+  xSemaphoreTake(sendMutex, portMAX_DELAY);
+  if (cmd != CMD_BUTTON) {
+    Serial.printf("[%lu] [EMU] Sending command: %s\n", millis(), cmd);
+  }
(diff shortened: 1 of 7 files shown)


ngxson:
Ok so my version has been able to boot into home screen. However, watchdog is triggered randomly for no reason (not even at a fixed interval, just pure random)

I suppose that esp32c3 qemu is not that stable. A simple hello world may sometimes trigger wdt timeout.

At this point, I'm not confident if this is a global issue or isolated to my host machine (docker on a macbook M3). If possible, could you give this a try on your machine? @daveallie @osteotek 

(edit: you need to comment out all calls to digitalRead and analogRead, they will trigger panic immediately)

An alternative solution could be to just run this infrastructure on a real standalone esp32c3 which is significantly cheaper than buying a new xteink.

ngxson:
Last version should "just work" without further modifications to the source code. Some small things should be polished like: debouce button, add mutex to the emulation serial

But at this point, I'm quite convinced that I should get rid of qemu, it has too many problems:
- No way to emulate GPIO or deepsleep
- No wifi or bluetooth support (ref: [docs](https://github.com/espressif/esp-toolchain-docs/blob/main/qemu/esp32c3/README.md))
- No good way to emulate exact RAM and CPU config
- Very unstable

Instead, I'll replace it with an option that make it flashable to a cheap esp32c3 board (4MB flash, 3.5MB usable with huge_app partition layout). With `OMIT_FONTS`, the compiled fw is ~2.5MB:
- Display, sdcard and buttons will be forwarded to the emulated implementation in this PR
- Power button will be mapped to BOOT button on the board

image

The 16MB flash version is a bit harder to find, but still cheap:

image

ngxson:
FWIW, I'm implementing it on my fork [here](https://github.com/ngxson/pluspoint-reader) because I think it will be too complicated upstream to crosspoint.

My forked version requires an ESP32C3 board as mentioned in the last message, but it does work very stable. Wifi can also work exactly as on the real device. On top of that, FS proxy make debugging things much easier.

I'm keeping this PR up for a bit more for anyone who wants to play with it

Image

daveallie:
@ngxson given these changes aren't going ahead, are you ok for this to be closed out, just trying to do a sweep of all the PRs.

ngxson:
Yeah actually I have a habit of keeping prototype PR as draft for visibility. But it's fine to close this

PR #501, opened 2026-01-22
fix: Chapter Selection UI bugs when koreader sync is enabled, and clarify default kosync URL

### LINKED ISSUE #428: Suggestion: Include default kosync URL (sync.koreader.rocks)
I've been trying the KOReader sync with thanks to @itsthisjustin and seems to be working great. Thanks!

A small suggestion: The settings could default to the default kosync server (`https://sync.koreader.rocks`) and then allow users to enter a custom alternative, rather than requiring entering the default.

### LINKED ISSUE #475: Bug: ">> Sync Progress" label broken in TOC when kosync turned on
### Affected Version

0.15.0

### Describe the Bug

The koreader sync ">> Sync Progress" message overwrites a Table of Contents entry at the top, and clicking a TOC entry goes to the wrong chapter (off by one).

Image

Image

I have tried clearing cache using the new menu option. And this is showing up for me in several different books.

### Steps to Reproduce

1. Go to table of contents by clicking "confirm" button in reader mode
2. Scroll to the top or bottom

### Expected Behavior

TOC links work correctly.

">> Sync Progress" shows clearly at top and bottom

```shell

```

### LINKED ISSUE #477: Different chapter loads from the one selected
### Affected Version

0.15.0

### Describe the Bug

I've been troubleshooting some Koreader sync issues (perhaps related to this, but probably not) and I noticed that when I select a chapter, it loads the chapter before it. So if I select "Chapter 4" it will actually open "Chapter 3. I've tried with two different books. Same issue. 

In case this is somehow related to the KOreader sync, I can push progress from the X4 but when I try to pull that progress from my Kobo or Readest, they both tell me there is no new progress to sync. I ensured that I'm using the exact same file on all devices. 

### Steps to Reproduce

1. Load book
2. Open chapter menu
3. Select chapter
4. Wrong chapter opens

### Expected Behavior

Selected chapter should open.

```shell

```

lukestein:
## Summary

* Fixes #475
* Fixes #477
* Closes #428

## Additional Context

* Updates to `src/activities/reader/EpubReaderChapterSelectionActivity.cpp` are copied verbatim from #433 (thanks to @jonasdiemer)
* Update to `src/activities/settings/KOReaderSettingsActivity.cpp` per discussion with @itsthisjustin at #428

Tested on my device with several books and koreader sync turned on and off.

---

### AI Usage

### DIFF
--- a/src/activities/reader/EpubReaderChapterSelectionActivity.cpp
+++ b/src/activities/reader/EpubReaderChapterSelectionActivity.cpp
@@ -188,8 +188,6 @@ void EpubReaderChapterSelectionActivity::renderScreen() {
   const auto pageStartIndex = selectorIndex / pageItems * pageItems;
   renderer.fillRect(0, 60 + (selectorIndex % pageItems) * 30 - 2, pageWidth - 1, 30);
 
-  // for (int itemIndex = pageStartIndex; itemIndex < totalItems && itemIndex < pageStartIndex + pageItems; itemIndex++)
-  // {
   for (int i = 0; i < pageItems; i++) {
     int itemIndex = pageStartIndex + i;
     if (itemIndex >= totalItems) break;
@@ -206,7 +204,6 @@ void EpubReaderChapterSelectionActivity::renderScreen() {
       const std::string chapterName =
           renderer.truncatedText(UI_10_FONT_ID, item.title.c_str(), pageWidth - 40 - indentSize);
 
-      // FIX: Use displayY here instead of recalculating based on tocIndex
       renderer.drawText(UI_10_FONT_ID, indentSize, displayY, chapterName.c_str(), !isSelected);
     }
   }


src/activities/reader/EpubReaderChapterSelectionActivity.cpp
```diff
@@ -188,22 +188,26 @@ void EpubReaderChapterSelectionActivity::renderScreen() {
...
       const int tocIndex = tocIndexFromItemIndex(itemIndex);
       auto item = epub->getTocItem(tocIndex);
+
       const int indentSize = 20 + (item.level - 1) * 15;
       const std::string chapterName =
           renderer.truncatedText(UI_10_FONT_ID, item.title.c_str(), pageWidth - 40 - indentSize);
-      renderer.drawText(UI_10_FONT_ID, indentSize, 60 + (tocIndex % pageItems) * 30, chapterName.c_str(),
-                        tocIndex != selectorIndex);
+
+      // FIX: Use displayY here instead of recalculating based on tocIndex
```
osteotek: I am not sure if this comment will make sense after the merge, I'd remove it

src/activities/reader/EpubReaderChapterSelectionActivity.cpp
```diff
@@ -188,22 +188,26 @@ void EpubReaderChapterSelectionActivity::renderScreen() {
   const auto pageStartIndex = selectorIndex / pageItems * pageItems;
   renderer.fillRect(0, 60 + (selectorIndex % pageItems) * 30 - 2, pageWidth - 1, 30);
 
-  for (int itemIndex = pageStartIndex; itemIndex < totalItems && itemIndex < pageStartIndex + pageItems; itemIndex++) {
-    const int displayY = 60 + (itemIndex % pageItems) * 30;
+  // for (int itemIndex = pageStartIndex; itemIndex < totalItems && itemIndex < pageStartIndex + pageItems; itemIndex++)
+  // {
```
osteotek: better to remove commented code if we don't plan to reuse it

osteotek (changes requested):

lukestein:
I've been **testing on device** successfully for a few days now

Bapf:
Just testet the PR. Everything looks fine.

One thought though: If the menu is split in multiple points now, is the sync option still right in the chapter menu, or should it be an own point on the top level.

lukestein:
> One thought though: If the menu is split in multiple points now, is the sync option still right in the chapter menu, or should it be an own point on the top level.

I also think the current UI (display the sync option at the top and bottom of the TOC, which may be multiple pages long…) is *not* ideal.

Would request/suggest to @daveallie that we try to get this bug fix merged ASAP, and then let the UI decisions be integrated with all the other UI discussions going on elsewhere.

itsthisjustin:
Open to suggestions on placement. I figured in long books no one would want to scroll all the way up if the end was much closer

itsthisjustin:
Open to suggestions on placement. I figured in long books no one would want to scroll all the way up if the end was much closer

lukestein:
> Open to suggestions on placement. I figured in long books no one would want to scroll all the way up if the end was much closer

I know there are some PRs and discussion about having some kind of "reader menu" separate from the TOC, which if it happens would presumably be the right place for this to go.

But given we currently only have the TOC, the problem with the current approach (imho) is that if you're at chapter 50 of a 100-chapter book you have to go multiple pages up or down to get to the sync. It feels like it should somehow be reachable on each "page" of the TOC.

itsthisjustin:
Agree but you can also press and hold to skip to the end

PR #506, opened 2026-01-23
feat: Add Power Button Hold Duration setting

Eloren1:
## Summary

- [x] Add Power Button Hold Duration setting ("0.5s", "1s", "2s")
- [x] Set Power Button Hold Duration to 0.5 seconds by default

## Additional Context

Previous hold duration was hardcoded to 400ms, it's a bit too fast and doesn't feel like "long-press". 

Now it's configurable. I still keep 0.5s as a default value, with an ability to select higher values in settings. 

In fact, the difference is noticeable and therefore this is a good customization to avoid accidental clicks for some.

image

---

### AI Usage

### DIFF
--- a/src/CrossPointSettings.h
+++ b/src/CrossPointSettings.h
@@ -56,7 +56,7 @@ class CrossPointSettings {
   enum SHORT_PWRBTN { IGNORE = 0, SLEEP = 1, PAGE_TURN = 2 };
 
   // Power button hold duration options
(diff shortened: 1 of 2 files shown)


daveallie:
I'd prefer if we just tweaked it to a reasonable static value, I get that people want to customise their experience, but this feels like a spot where we can just make up our minds and keep the user on the rails.

Issue #504, opened 2026-01-23
Support for showing book progress without decimal places


treetrum:
Prior to 0.15 book's progress was shown as an integer, however now it shows with 1 decimal of precision. I.e. `21.2%` instead of `21%`.

I personally preferred this and would love a way to have the old functionality. Understand that this is just a preference though, so feels like a good time to add a setting.

~I may have time over the coming days to implement this — but thought I'd gather opinions first.~ Couldn't help myself. Pr raised: https://github.com/crosspoint-reader/crosspoint-reader/pull/507

PR #507, opened 2026-01-23
fix: remove decimal places from progress %

treetrum:
## Summary

Addresses https://github.com/crosspoint-reader/crosspoint-reader/issues/504

- Reverts book progress % to showing as an integer instead of with a decimal place
  - This was changed to 1 decimal point of precision in https://github.com/crosspoint-reader/crosspoint-reader/pull/232 from what I can tell
  - As this wasn't the primary intention of that PR, I'm assuming it was left in accidentally

IMO having a decimal place of precision is too much for something as vague as book completion percent. This de-clutters the status bar and prevents extra updates as you change pages.

---

### AI Usage

YES

### DIFF
--- a/src/CrossPointSettings.cpp
+++ b/src/CrossPointSettings.cpp
@@ -14,7 +14,7 @@ CrossPointSettings CrossPointSettings::instance;
 namespace {
 constexpr uint8_t SETTINGS_FILE_VERSION = 1;
 // Increment this when adding new persisted settings fields
-constexpr uint8_t SETTINGS_COUNT = 21;
+constexpr uint8_t SETTINGS_COUNT = 20;
 constexpr char SETTINGS_FILE[] = "/.crosspoint/settings.bin";
 }  // namespace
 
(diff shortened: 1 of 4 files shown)


osteotek:
I'm not sure if we should create setting for every possible customization option 🙂 it is already getting bit out of hand. How about in this PR just revert to integer progress and let the maintainer(s) decide

Eloren1:
Then we should just leave the integer

treetrum:
Yeah, that's a fair point. Agreed that not everything needs to be a setting.

My opinion is that you probably don't need decimal level precision on book progress, so happy to make this PR just revert to pre 0.15.0 behaviour. 

I'll update.

Eloren1:
It's a fix not feat

lukestein:
[Testing on device](https://github.com/crosspoint-reader/crosspoint-reader/discussions/546) and seems to work for both epub and text files. 🎉

This issue was slightly annoying me. Thanks for fixing!

Issue #176, opened 2025-12-30
OTA update freezes


cjrezz:
### Affected Version

0.11.0

### Describe the Bug

When trying to update OTA from 0.10 to 0.11, the update freezes at 0%. 

### Steps to Reproduce

Try to update OTA

### Expected Behavior

Should update successfully

![Image](https://github.com/user-attachments/assets/031ffd13-743d-4ede-938c-de85cda68ad0)

### Relevant Log Output/Screenshots

```shell

```

paulandrewbowers-cloud:
On version 0.10.0 upgrading to version 0.11.0 appears to freeze at 98% left for 15 mins, but actually completed when reset device. Reflashed again on usb to be sure. FYI the upgrade took 30mins over OTA
-------- Original message --------From: cjrezz ***@***.***> Date: 30/12/2025  16:08  (GMT+00:00) To: daveallie/crosspoint-reader ***@***.***> Cc: Subscribed ***@***.***> Subject: [daveallie/crosspoint-reader] OTA update freezes (Issue #176) cjrezz created an issue (daveallie/crosspoint-reader#176)
Affected Version
0.11.0
Describe the Bug
When trying to update OTA from 0.10 to 0.11, the update freezes at 0%.
Steps to Reproduce
Try to update OTA
Expected Behavior
Should update successfully
20251230_080350.jpg (view on web)
Relevant Log Output/Screenshots

cjrezz:
I tried again, letting it run for 30 minutes, but the update still didn't move past 0% for me.

lukestein:
Not sure if this is the same or separate issue, but I am getting "Update failed" when attempting OTA update from 0.11.0 to 0.11.1.

Trying to update using the [Flash Tools](https://xteink.dve.al/) site also gives an error
<img width="704" height="274" alt="Image" src="https://github.com/user-attachments/assets/da83d841-2235-4254-8700-dbdd9299fc20" />

cjrezz:
> Not sure if this is the same or separate issue, but I am getting "Update failed" when attempting OTA update from 0.11.0 to 0.11.1.
> 
> Trying to update using the [Flash Tools](https://xteink.dve.al/) site also gives an error 

I just came back to report the same issue.

cjrezz:
Also, the OTA update, despite not moving past 0% seems to have done something to the other boot partition as I can't swap partitions any longer to go back to the stock firmware.

lukestein:
[...]

PR #509, opened 2026-01-23
refactor: Re-work for OTA feature

embedded4ever:
## Summary

Finally, I have received my device and got to chance to work on OTA. 
https://github.com/crosspoint-reader/crosspoint-reader/issues/176

Existing OTA functionality is very buggy, many of times (I would say 8 out of 10) are end up with fail for me. When the time that it works it is very slow and take ages. For others looks like end up with crash or different issues. 

To be honest, I'm not familiar with Arduino APIs of OTA process, but looks like not good as much esp-idf itself. I always found Arduino APIs very bulky for esp32. Wrappers and wrappers.

## Additional Context
Right now, OTA takes ~ 3min 10sec (of course depends on size of .bin file). Can be tested with playing version info inside from `platform.ini` file.

```
[crosspoint]
version = 0.14.0
```
---

### AI Usage

### DIFF
--- a/src/network/OtaUpdater.cpp
+++ b/src/network/OtaUpdater.cpp
@@ -1,38 +1,123 @@
 #include "OtaUpdater.h"
 
 #include <ArduinoJson.h>
-#include <HTTPClient.h>
-#include <Update.h>
+
+#include "esp_http_client.h"
+#include "esp_https_ota.h"
+#include "esp_wifi.h"
 
 namespace {
 constexpr char latestReleaseUrl[] = "https://api.github.com/repos/crosspoint-reader/crosspoint-reader/releases/latest";
+
+/* This is buffer and size holder to keep upcoming data from latestReleaseUrl */
+char* local_buf;
+int output_len;
+
+/*
+ * When esp_crt_bundle.h included, it is pointing wrong header file
+ * which is something under WifiClientSecure because of our framework based on arduno platform.
+ * To manage this obstacle, don't include anything, just extern and it will point correct one.
+ */
+extern "C" {
+extern esp_err_t esp_crt_bundle_attach(void* conf);
 }
 
+esp_err_t http_client_set_header_cb(esp_http_client_handle_t http_client) {
+  return esp_http_client_set_header(http_client, "User-Agent", "CrossPoint-ESP32-" CROSSPOINT_VERSION);
+}
+
+esp_err_t event_handler(esp_http_client_event_t* event) {
+  /* We do interested in only HTTP_EVENT_ON_DATA event only */
+  if (event->event_id != HTTP_EVENT_ON_DATA) return ESP_OK;
+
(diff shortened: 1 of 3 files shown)


osteotek:
Nice work! I couldn’t quite understand from your description. Does it actually improve OTA updates? They’ve been borked since their release.

embedded4ever:
Sure, actually I believe there are some different issues with existing Arduino APIs, I tried to track by enabling log_d and log_w to see what logs are printing when even update checking ends up with this failing settings inside 

`<your-path>/.platformio/packages/framework-arduinoespressif32/cores/esp32/esp32-hal-log.h`



This is what I got most of time : 

```
[ 96653][D][HTTPClient.cpp:1463] writeToStreamDataBlock(): bytesWritten 6400 and size 9222 mismatch!.
[ 96662][W][HTTPClient.cpp:1483] returnError(): error(-10): Stream write error
```

Looks like connection is closing in somehow. 

Another issue that I faced with existing one, request ends up successfully but when I check the content of payload, looks like half of the content is there and form of json is not valid and return EMPTY INPUT by json library. 


I used esp-idf apis over time and I would say more healthier than Arduino one, even error handling-wise.

src/network/OtaUpdater.cpp
```diff
@@ -1,38 +1,123 @@
...
+ * When esp_crt_bundle.h included, it is pointing wrong header file
+ * which is something under WifiClientSecure because of our framework based on arduno platform.
+ * To manage this obstacle, don't include anything, just extern and it will point correct one.
+ */
+extern "C" {
+extern esp_err_t esp_crt_bundle_attach(void* conf);
+}
+
+esp_err_t http_client_set_header_cb(esp_http_client_handle_t http_client) {
+  return esp_http_client_set_header(http_client, "User-Agent", "ESP32-CrossPoint-v1.0");
```
daveallie: Can we please continue to use the crosspoint version in the UA string

src/network/OtaUpdater.cpp
```diff
@@ -1,38 +1,123 @@
...
+
+  esp_http_client_handle_t client_handle = esp_http_client_init(&client_config);
+  if (!client_handle) {
+    Serial.printf("[%lu] [OTA] HTTP Client Handle Failed\n", millis());
+    return INTERNAL_UPDATE_ERROR;
+  }
 
-  http.begin(*client, latestReleaseUrl);
-  http.addHeader("User-Agent", "CrossPoint-ESP32-" CROSSPOINT_VERSION);
+  esp_err = esp_http_client_set_header(client_handle, "User-Agent", "ESP32-CrossPoint-v1.0");
```
daveallie: Same as above, can we continue to use the firmware version in the UA

src/network/OtaUpdater.cpp
embedded4ever: You are right, will be corrected with my following patch.

PR #522, opened 2026-01-23
feat: add HalDisplay and HalGPIO

ngxson:
## Summary

Extracted some changes from https://github.com/crosspoint-reader/crosspoint-reader/pull/500 to make reviewing easier

This PR adds HAL (Hardware Abstraction Layer) for display and GPIO components, making it easier to write a stub or an emulated implementation of the hardware.

SD card HAL will be added via another PR, because it's a bit more tricky.

---

### AI Usage

### DIFF
--- a/src/MappedInputManager.cpp
+++ b/src/MappedInputManager.cpp
@@ -64,9 +64,7 @@ bool MappedInputManager::mapButton(const Button button, bool (HalGPIO::*fn)(uint
 
 bool MappedInputManager::wasPressed(const Button button) const { return mapButton(button, &HalGPIO::wasPressed); }
 
-bool MappedInputManager::wasReleased(const Button button) const {
-  return mapButton(button, &HalGPIO::wasReleased);
-}
+bool MappedInputManager::wasReleased(const Button button) const { return mapButton(button, &HalGPIO::wasReleased); }
 
 bool MappedInputManager::isPressed(const Button button) const { return mapButton(button, &HalGPIO::isPressed); }
 


daveallie:
Is the idea here that a test-friendly non-X4 specific subclass of `HalDisplay` is made with relevant stubs to be used in local testing?

ngxson:
I'm not 100% what do you mean by "local testing", but yes the overall idea is that `HalDisplay` can be hardware-agnostic to facilitate testings.

I've experimented with an implementation that instead of sending frame buffer to the real display (aka via SPI), it sends to USB UART. I can then capture the frame buffer on my computer and display it. Here is the code if you want to have a look (quite hacky, but works well for now): https://github.com/ngxson/pluspoint-reader/blob/9558dc1251863bd9a951675e0550adc522bb793c/ngxson/hal/EInkDisplay.cpp#L76-L79

daveallie:
Yep, makes sense, I also saw you're working on some QEMU, all together, this'll be a good win for local emulation/testing. I've introduced a bunch of merge conflicts, but we can get this merged once they're resolved.

daveallie:
Ahh I see it's been benched

ngxson:
Yeah I opened this PR also for discussion. Since QEMU is far from being stable, the only option left here is to use a standalone esp32c3 for testing.

I can resolve these conflicts a bit later. But just want to discuss if the overall direction here is ok, and if there are any other ideas that can be useful for testings in general.

ngxson:
One idea that I've also been thinking about is to compile the whole firmware as an executable that can run directly on a PC without emulation, but that may require far more changes.

Currently, many functionalities depend on Arduino library, like wifi for example. Can be quite tricky to implement HAL for those

daveallie:
Yeah, I assume there'd need to be a lot of overridden linking and an outrageous amount of stubbing to get that to work. Regardless, I think this HAL implementation will help with working towards a simple unit test harness.

ngxson:
I resolved the merge conflicts & tested it. Most of the changes are just renaming though

PR #528, opened 2026-01-24
feat: UI themes, Lyra

CaptainFrito:

- Visual UI overhaul
- UI theme selection

- Added a setting "UI Theme": Classic, Lyra
- The classic theme is the current Crosspoint theme
- The Lyra theme implements these mockups: https://www.figma.com/design/UhxoV4DgUnfrDQgMPPTXog/Lyra-Theme?node-id=2003-7596&t=4CSOZqf0n9uQMxDt-0 by Discord users yagofarias, ruby and gan_shu
- New functions in GFXRenderer to render rounded rectangles, greyscale fills (using dithering) and thick lines
- Basic UI components are factored into BaseTheme methods which can be overridden by each additional theme. Methods that are not overridden will fallback to BaseTheme behavior. This means any new features/components in CrossPoint only need to be developed for the "Classic" BaseTheme.
- Additional themes can easily be developed by the community using this foundation

IMG_7649 Medium
IMG_7746 Medium
IMG_7651 Medium

## Additional Context

- Only the Home, Library and main Settings screens have been implemented so far, this will be extended to the transfer screens and chapter selection screen later on, but we need to get the ball rolling somehow :)
- Loading extra covers on the home screen in the Lyra theme takes a little more time (about 2 seconds), I added a loading bar popup (reusing the Indexing progress bar from the reader view, factored into a neat UI component) but the popup adds ~400ms to the loading time.
- ~~Home screen thumbnails will need to be generated separately for each theme, because they are displayed in different sizes. Because we're using dithering, displaying a thumb with the wrong size causes the picture to look janky or dark as it does on the screenshots above. No worries this will be fixed in a future PR.~~ Thumbs are now generated with a size parameter
- UI Icons will need to be implemented in a future PR.

---

### AI Usage

This is not a vibe coded PR. Copilot was used for autocompletion to save time but I reviewed, understood and edited all generated code.

### DIFF
--- a/lib/GfxRenderer/GfxRenderer.h
+++ b/lib/GfxRenderer/GfxRenderer.h
@@ -95,6 +95,7 @@ class GfxRenderer {
   void drawText(int fontId, int x, int y, const char* text, bool black = true,
                 EpdFontFamily::Style style = EpdFontFamily::REGULAR) const;
   int getSpaceWidth(int fontId) const;
+  int getTextAdvanceX(int fontId, const char* text) const;
   int getFontAscenderSize(int fontId) const;
   int getLineHeight(int fontId) const;
   std::string truncatedText(int fontId, const char* text, int maxWidth,
--- a/src/activities/home/HomeActivity.cpp
+++ b/src/activities/home/HomeActivity.cpp
@@ -70,7 +70,8 @@ void HomeActivity::loadRecentBooks(int maxBooks, int coverHeight) {
         // If epub, try to load the metadata for title/author and cover
         if (StringUtils::checkFileExtension(lastBookFileName, ".epub")) {
           Epub epub(book.path, "/.crosspoint");
-          epub.load(false);
+          // Skip loading css since we only need metadata here
+          epub.load(false, true);
 
           // Try to generate thumbnail image for Continue Reading card
           if (!showingLoading) {
--- a/README.md
+++ b/README.md
@@ -102,13 +102,18 @@ After flashing the new features, it’s recommended to capture detailed logs fro
 First, make sure all required Python packages are installed:
 
 ```python
-python3 -m pip install serial colorama matplotlib
+python3 -m pip install pyserial colorama matplotlib
 ```
 after that run the script:
 ```sh
+# For Linux
+# This was tested on Debian and should work on most Linux systems.
 python3 scripts/debugging_monitor.py
+
+# For macOS
+python3 scripts/debugging_monitor.py /dev/cu.usbmodem2101
 ```
-This was tested on Debian and should work on most Linux systems. Minor adjustments may be required for Windows or macOS.
+Minor adjustments may be required for Windows.
 
 ## Internals
 
--- a/docs/webserver.md
+++ b/docs/webserver.md
@@ -153,7 +153,7 @@ Click **File Manager** to access file management features.
 
 1. Click the **+ Add** button in the top-right corner
 2. Select **New Folder** from the dropdown menu
-3. Enter a folder name (letters, numbers, underscores, and hyphens only)
+3. Enter a folder name (must not contain characters \" * : < > ? / \\ | and must not be . or ..)
 4. Click **Create Folder**
 
 This is useful for organizing your ebooks by genre, author, or series.
--- a/lib/EpdFont/EpdFontFamily.cpp
+++ b/lib/EpdFont/EpdFontFamily.cpp
@@ -1,22 +1,18 @@
 #include "EpdFontFamily.h"
 
 const EpdFont* EpdFontFamily::getFont(const Style style) const {
-  if (style == BOLD && bold) {
+  // Extract font style bits (ignore UNDERLINE bit for font selection)
+  const bool hasBold = (style & BOLD) != 0;
+  const bool hasItalic = (style & ITALIC) != 0;
+
+  if (hasBold && hasItalic) {
+    if (boldItalic) return boldItalic;
+    if (bold) return bold;
+    if (italic) return italic;
+  } else if (hasBold && bold) {
     return bold;
-  }
-  if (style == ITALIC && italic) {
+  } else if (hasItalic && italic) {
     return italic;
-  }
-  if (style == BOLD_ITALIC) {
-    if (boldItalic) {
-      return boldItalic;
-    }
-    if (bold) {
-      return bold;
-    }
-    if (italic) {
-      return italic;
-    }
   }
 
   return regular;
--- a/lib/EpdFont/EpdFontFamily.h
+++ b/lib/EpdFont/EpdFontFamily.h
@@ -3,7 +3,7 @@
 
 class EpdFontFamily {
  public:
-  enum Style : uint8_t { REGULAR = 0, BOLD = 1, ITALIC = 2, BOLD_ITALIC = 3 };
+  enum Style : uint8_t { REGULAR = 0, BOLD = 1, ITALIC = 2, BOLD_ITALIC = 3, UNDERLINE = 4 };
 
   explicit EpdFontFamily(const EpdFont* regular, const EpdFont* bold = nullptr, const EpdFont* italic = nullptr,
                          const EpdFont* boldItalic = nullptr)
--- a/lib/Epub/Epub.cpp
+++ b/lib/Epub/Epub.cpp
@@ -86,6 +86,10 @@ bool Epub::parseContentOpf(BookMetadataCache::BookMetadata& bookMetadata) {
     tocNavItem = opfParser.tocNavPath;
   }
 
+  if (!opfParser.cssFiles.empty()) {
+    cssFiles = opfParser.cssFiles;
+  }
+
   Serial.printf("[%lu] [EBP] Successfully parsed content.opf\n", millis());
   return true;
 }
@@ -204,15 +208,91 @@ bool Epub::parseTocNavFile() const {
   return true;
 }
 
+std::string Epub::getCssRulesCache() const { return cachePath + "/css_rules.cache"; }
+
+bool Epub::loadCssRulesFromCache() const {
+  FsFile cssCacheFile;
+  if (SdMan.openFileForRead("EBP", getCssRulesCache(), cssCacheFile)) {
+    if (cssParser->loadFromCache(cssCacheFile)) {
+      cssCacheFile.close();
+      Serial.printf("[%lu] [EBP] Loaded CSS rules from cache\n", millis());
+      return true;
+    }
+    cssCacheFile.close();
+    Serial.printf("[%lu] [EBP] CSS cache invalid, reparsing\n", millis());
+  }
+  return false;
+}
+
+void Epub::parseCssFiles() const {
+  if (cssFiles.empty()) {
+    Serial.printf("[%lu] [EBP] No CSS files to parse, but CssParser created for inline styles\n", millis());
+  }
+
+  // Try to load from CSS cache first
+  if (!loadCssRulesFromCache()) {
+    // Cache miss - parse CSS files
+    for (const auto& cssPath : cssFiles) {
+      Serial.printf("[%lu] [EBP] Parsing CSS file: %s\n", millis(), cssPath.c_str());
+
+      // Extract CSS file to temp location
+      const auto tmpCssPath = getCachePath() + "/.tmp.css";
+      FsFile tempCssFile;
+      if (!SdMan.openFileForWrite("EBP", tmpCssPath, tempCssFile)) {
+        Serial.printf("[%lu] [EBP] Could not create temp CSS file\n", millis());
+        continue;
+      }
+      if (!readItemContentsToStream(cssPath, tempCssFile, 1024)) {
+        Serial.printf("[%lu] [EBP] Could not read CSS file: %s\n", millis(), cssPath.c_str());
+        tempCssFile.close();
+        SdMan.remove(tmpCssPath.c_str());
+        continue;
+      }
+      tempCssFile.close();
+
+      // Parse the CSS file
+      if (!SdMan.openFileForRead("EBP", tmpCssPath, tempCssFile)) {
+        Serial.printf("[%lu] [EBP] Could not open temp CSS file for reading\n", millis());
+        SdMan.remove(tmpCssPath.c_str());
+        continue;
+      }
+      cssParser->loadFromStream(tempCssFile);
+      tempCssFile.close();
+      SdMan.remove(tmpCssPath.c_str());
+    }
+
+    // Save to cache for next time
+    FsFile cssCacheFile;
+    if (SdMan.openFileForWrite("EBP", getCssRulesCache(), cssCacheFile)) {
+      cssParser->saveToCache(cssCacheFile);
+      cssCacheFile.close();
+    }
+
+    Serial.printf("[%lu] [EBP] Loaded %zu CSS style rules from %zu files\n", millis(), cssParser->ruleCount(),
+                  cssFiles.size());
+  }
+}
+
 // load in the meta data for the epub file
-bool Epub::load(const bool buildIfMissing) {
+bool Epub::load(const bool buildIfMissing, const bool skipLoadingCss) {
   Serial.printf("[%lu] [EBP] Loading ePub: %s\n", millis(), filepath.c_str());
 
   // Initialize spine/TOC cache
   bookMetadataCache.reset(new BookMetadataCache(cachePath));
+  // Always create CssParser - needed for inline style parsing even without CSS files
+  cssParser.reset(new CssParser());
 
   // Try to load existing cache first
   if (bookMetadataCache->load()) {
+    if (!skipLoadingCss && !loadCssRulesFromCache()) {
+      Serial.printf("[%lu] [EBP] Warning: CSS rules cache not found, attempting to parse CSS files\n", millis());
+      // to get CSS file list
+      if (!parseContentOpf(bookMetadataCache->coreMetadata)) {
+        Serial.printf("[%lu] [EBP] Could not parse content.opf from cached bookMetadata for CSS files\n", millis());
+        // continue anyway - book will work without CSS and we'll still load any inline style CSS
+      }
+      parseCssFiles();
+    }
     Serial.printf("[%lu] [EBP] Loaded ePub: %s\n", millis(), filepath.c_str());
     return true;
   }
@@ -307,6 +387,11 @@ bool Epub::load(const bool buildIfMissing) {
   if (!bookMetadataCache->load()) {
     Serial.printf("[%lu] [EBP] Failed to reload cache after writing\n", millis());
     return false;
+  }
+
+  if (!skipLoadingCss) {
+    // Parse CSS files after cache reload
+    parseCssFiles();
   }
 
   Serial.printf("[%lu] [EBP] Loaded ePub: %s\n", millis(), filepath.c_str());
--- a/lib/Epub/Epub.h
+++ b/lib/Epub/Epub.h
@@ -8,6 +8,7 @@
 #include <vector>
 
 #include "Epub/BookMetadataCache.h"
+#include "Epub/css/CssParser.h"
 
 class ZipFile;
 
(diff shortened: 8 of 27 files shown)


el:
Amazing work, 

I tried this out and I have a few feedback:
- Everytime you go back to the home screen there are two separate "loading" screens before the book covers pop up
- Book covers should be cropped imo, otherwise we get a white part below or on the right
- When a book does not have a book cover, it appears blank. There should be a default book cover
- Bottom bar buttons do not align with the physical buttons I think they should be a bit wider to match the actual ones
- Nit: Right buttons should be a few pixels higher to match the physical ones.
- I find it a bit counterintuitive to have Left-Right (< / >) buttons on the right and up down buttons on the bottom because of the positions of the buttons. I think they should be switched


Overall really liked the theme and how quick it is to browse around. I think all of my feedback can be taken care of very easily even as a follow-up PR.

DavidOrtmann:
Great job with the Themes. Agreeing with @el's observations here, don't want to repeat everything, but things I found most noticable:

- Switching left-right, up-down is not intuitive yet for switching categories and settings, respectively.
- The active status of the category (Display, Reader) is too faint, I find it barely noticeable (only when switching). I think the selected category should also be filled with gray or be bolded.
- Browse files/ File Transfer in the main menu look a bit too minimalistic and lost, especially with very few books (I only have two), so there is a lot of white space underneath. It feels like this part of the menu should be separated more with some kind of border or fill

IMG_20260125_214132

Apart from this, I think this looks really professional, and I love the introduction of Themes.

CaptainFrito:
Thanks a lot for the detailed feedback guys!

The two separate "Loading" popups you get every time you return to the home screen, is that the plain popup and then the popup with the progress bar? The way it works is we show the plain popup when we're loading existing thumbnails (which should be most of the time) and the progress bar one when we have to generate the thumbnails (which should be the first time only, or whenever a new book is added to the list). I didn't test the case where a book doesn't have a cover, it might try to generate it every time and show you the progress bar, is that what's happening? We need a special case for that then.

On the subject of button bindings, we had initially set the side buttons for Up and Down and the front buttons for Left and Right, which would seem more intuitive. But while using it that way all of us found it almost unusable. Switching tabs will happen a lot less than scrolling the list so in the end it makes sense that it's bound to the less prominent side buttons. We could make this a setting too if you'd like.

I will leave the designers respond to the visual feedback. I can update the PR today with corrections. Thanks team!

CaptainFrito:
I updated the PR with a fix for the crashes some users were experiencing while testing, and an empty thumbnail is now created when the book has no cover, so that you don't get the loading popup repeatedly on the home screen. This is now stable and awaiting a response from the design team for the rest of the feedback.

Uri-Tauber:
> Switching tabs will happen a lot less than scrolling the list so in the end it makes sense that it's bound to the less prominent side buttons. We could make this a setting too if you'd like.

I agree with @el and @DavidOrtmann that this feels a bit unintuitive. Additionally, I tend to use the side buttons more easily, maybe because I'm left-handed. Adding a setting to control this would be a good idea.

DavidOrtmann:
Another thought I've had: Maybe we just don't separate category and setting navigation from each other? We could do all these with the same button pair:
Select category > Select > Select setting
Go to start or end of list to change the category.
(I would prefer a button logic most people "get" instead of too many granular settings.)

CaptainFrito:
> Another thought I've had: Maybe we just don't separate category and setting navigation from each other? We could do all these with the same button pair: Select category > Select > Select setting Go to start or end of list to change the category. (I would prefer a button logic most people "get" instead of too many granular settings.)

You mean like in the 0.15 release where there's a first settings menu with the 4 categories, and then another screen for each of them?

DavidOrtmann:
> You mean like in the 0.15 release where there's a first settings menu with the 4 categories, and then another screen for each of them?

Not like an extra page, no. The page layout would stay the same as in the current dev version. This currently has Settings Categories and the settings themselves on one screen. Now imagine you can reach the categories and the settings with the same button. It would feel like this:

(Symbol ↓ means "clicking down button"; bold text is a category)

**Display** ↓ **Reader** [click select = enters Reader settings] ↓ **Controls** ↓ **System** ↓ Font Family ↓ Font Size

The above would be the path to change the font size. It would only need the up/down button and the select button, no need for left/right.

CaptainFrito:
> > You mean like in the 0.15 release where there's a first settings menu with the 4 categories, and then another screen for each of them?
> 
> Not like an extra page, no. The page layout would stay the same as in the current dev version. This currently has Settings Categories and the settings themselves on one screen. Now imagine you can reach the categories and the settings with the same button. It would feel like this:
> 
> (Symbol ↓ means "clicking down button"; bold text is a category)
> 
> **Display** ↓ **Reader** [click select = enters Reader settings] ↓ **Controls** ↓ **System** ↓ Font Family ↓ Font Size
> 
> The above would be the path to change the font size. It would only need the up/down button and the select button, no need for left/right.

Dude this is actually a step in the right direction! To expand on that we could even treat the tab bar as just another line, meaning you can highlight the tab bar when you land on it with the up/down arrows, and when it's highlighted you can press enter to change tabs. The selection cursor would be the same as any other lines, light grey background and selected tab on black background.

Thanks man this sounds like it would be way more intuitive than what we have now, but since we're not using the side buttons at all we could probably keep them as a shortcut to change tabs still.

daveallie:
Really neat work @CaptainFrito! Agree with a lot of the feedback already here, but counter to a few of the others, I actually prefer the new tab/item navigation button layout for basically the reason you mentioned, the side buttons feel like a "higher-order" control and them controlling the sections makes more intuitive sense to me, that being said, I'm sure I'd adjust to either option.

el:
Not sure if it is related to my device, but the Calibre URL screen is broken for this branch.

Also, instead of hiding the button completely, can we hide partially like the Figma mocks?

SCR-20260127-pezw

CaptainFrito:
> Not sure if it is related to my device, but the Calibre URL screen is broken for this branch.
> 
> Also, instead of hiding the button completely, can we hide partially like the Figma mocks?
> 

Thanks @el it is now fixed, and I added the little inactive button indicator which looks pretty cool!

I worked on the controls for the tab bar, and I think we're now at a point where it's a lot more usable:
IMG_7798 Medium
IMG_7799 Medium

So when we land on a screen with tabs, they are in a selected state first, which lets you change tabs with the confirm button. You can move up and down normally to select the settings, and the tab bar is considered the first item in the list, so when you move your selection to the top they get selected. It's really nice overall to be able to control everything with just up/down/confirm. 
So now we can decide what to do with the side buttons, they can either be also linked to up/down, just like the left/right buttons, for people who like to use the side buttons for scrolling, or we can link them to tab left/right for a quick shortcut when you're down the list and want to change tab. 

The selected tab was also made more visible.

CaptainFrito:
In order to simplify interactions on the Library screen, the design team proposed we split the My Library screen into Browse Files and Recents. Both are accessible from the home screen which also gives the main menu a bit more substance. I have updated the PR with this change.

This will leave extra room on either screen for extra features that are included in the Lyra mockups:
- In the File browser, icons and a toolbar with the current path
- In the Recents screen, a grid of book cover thumbnails

Most of the other feedback has been addressed, except for:
- Different cover aspect ratios cause a white bar under the cover when selected
- No default cover for books without one

These things could be added in a separate PR to avoid having this one drag on too long and accumulating conflicts with master.

src/activities/home/RecentBooksActivity.cpp
```diff
@@ -0,0 +1,148 @@
...
+  const int contentTop = metrics.topPadding + metrics.headerHeight + metrics.verticalSpacing;
+  const int contentHeight = pageHeight - contentTop - metrics.buttonHintsHeight - metrics.verticalSpacing * 2;
+
+  // Recent tab
+  if (recentBooks.empty()) {
+    renderer.drawText(UI_10_FONT_ID, metrics.contentSidePadding, contentTop + 20, "No recent books");
+  } else {
+    UITheme::drawList(
+        renderer, Rect{0, contentTop, pageWidth, contentHeight}, recentBooks.size(), selectorIndex,
+        [this](int index) { return recentBooks[index].title; }, false, nullptr, false, nullptr);
```
el: Why are we only showing title here and not author?

src/activities/home/HomeActivity.cpp
```diff
@@ -171,21 +207,24 @@ void HomeActivity::loop() {
   if (mappedInput.wasReleased(MappedInputManager::Button::Confirm)) {
     // Calculate dynamic indices based on which options are available
     int idx = 0;
-    const int continueIdx = hasContinueReading ? idx++ : -1;
+    int menuSelectedIndex = selectorIndex - static_cast<int>(recentBooks.size());
```
el, line 203: You need to also check if opds is enabled

src/activities/settings/SettingsActivity.cpp
el, line 1: You can no longer switch between categories using the side buttons

CaptainFrito: Yeah as someone pointed out in the discussion they like to use side buttons to scroll up and down, so right now we have both front and side buttons for up/down selection to please everybody, but you can long press side buttons to change tabs, as a kind of power user shortcut.

src/activities/home/RecentBooksActivity.cpp
CaptainFrito: We will eventually, the recent books screen is going to evolve quite a bit. Let’s leave that for a future PR though.

src/activities/settings/SettingsActivity.cpp
CaptainFrito: Hard to make a decision that will please everybody on this front :)

src/activities/home/HomeActivity.cpp
CaptainFrito: Fixed! Thanks

CaptainFrito:
Little update to the home page covers, they are now cropped horizontally to fit the whole selection rectangle and avoid having a white band at the bottom. What do you guys think?

IMG_7813 Large

This PR is now pretty stable and caught up to master, we could probably merge it soon to avoid having to fix a billion conflicts again please @daveallie :)

basvdploeg:
> Little update to the home page covers, they are now cropped horizontally to fit the whole selection rectangle and avoid having a white band at the bottom. What do you guys think?
> 
> IMG_7807 2 Large
> 
> This PR is now pretty stable and caught up to master, we could probably merge it soon to avoid having to fix a billion conflicts again please @daveallie :)

What about keeping the right aspect ratio (all the same width) and then align them to the bottom?

Like how Apple does it in their Books app:

IMG_0057

CaptainFrito:
> > Little update to the home page covers, they are now cropped horizontally to fit the whole selection rectangle and avoid having a white band at the bottom. What do you guys think?
> > IMG_7807 2 Large
> > This PR is now pretty stable and caught up to master, we could probably merge it soon to avoid having to fix a billion conflicts again please @daveallie :)
> 
> What about keeping the right aspect ratio (all the same width) and then align them to the bottom?
> 
> Like how Apple does it in their Books app:
> 
> IMG_0057

Good thinking man, this is more or less what we had before cropping (see screenshots above) except aligned at the top not the bottom. To be honest it looks a lot less cool than the Apple Books app when the pictures are small, greyscale, and without a drop shadow effect. I'll let the design team weigh in on this.

src/activities/home/RecentBooksActivity.cpp
daveallie: Given the recents screen on master includes this information (and is currently merged as "My Library" on master), I'd like to keep the author listing in recents functionality. I understand it may be supplanted by newer UI in the future, but until then we should retain existing functionality to avoid blocking releases.

lib/GfxRenderer/GfxRenderer.h
```diff
@@ -7,6 +7,16 @@
...
 
+// Color representation: uint8_t mapped to 4x4 Bayer matrix dithering levels
+// 0 = transparent, 1-16 = gray levels (white to black)
+using Color = uint8_t;
+
+constexpr Color COLOR_CLEAR = 0x00;
+constexpr Color COLOR_WHITE = 0x01;
+constexpr Color COLOR_LIGHT_GRAY = 0x05;
+constexpr Color COLOR_DARK_GRAY = 0x0A;
+constexpr Color COLOR_BLACK = 0x10;
```
daveallie: nit: Would an enum be better here?

src/RecentBooksStore.h
```diff
@@ -10,6 +10,11 @@ struct RecentBook {
   bool operator==(const RecentBook& other) const { return path == other.path; }
 };
 
+struct RecentBookWithCover {
+  RecentBook book;
+  std::string coverBmpPath;
```
daveallie: Should this just be merged into `RecentBook`? Afaict whenever we use `RecentBookWithCover` we check if `coverBmpPath` isn't empty. Would it make sense to persist this content make loading the cover faster and avoiding loading the epub file each time the HomeActivity is first loaded

src/components/themes/BaseTheme.cpp
```diff
@@ -0,0 +1,628 @@
...
+  if (showBatteryPercentage) {
+    const uint16_t percentage = battery.readPercentage();
+    const auto percentageText = std::to_string(percentage) + "%";
+    batteryX -= renderer.getTextWidth(SMALL_FONT_ID, percentageText.c_str());
+  }
+  drawBattery(renderer, Rect{batteryX, rect.y + 5, BaseMetrics::values.batteryWidth, BaseMetrics::values.batteryHeight},
+              showBatteryPercentage);
+
+  if (title) {
+    renderer.drawCenteredText(UI_12_FONT_ID, rect.y + 5, title, true, EpdFontFamily::BOLD);
```
daveallie: This has no truncation and will render content out of bounds (like folder name in the explorer)

src/components/themes/lyra/LyraTheme.cpp
```diff
@@ -0,0 +1,361 @@
...
+    const auto percentageText = std::to_string(percentage) + "%";
+    batteryX -= renderer.getTextWidth(SMALL_FONT_ID, percentageText.c_str());
+  }
+  drawBattery(renderer,
+              Rect{batteryX, rect.y + 10, LyraMetrics::values.batteryWidth, LyraMetrics::values.batteryHeight},
+              showBatteryPercentage);
+
+  if (title) {
+    renderer.drawText(UI_12_FONT_ID, rect.x + LyraMetrics::values.contentSidePadding,
+                      rect.y + LyraMetrics::values.batteryBarHeight + 3, title, true, EpdFontFamily::BOLD);
```
daveallie: As below, no truncation here, so can draw text out of bounds

src/activities/settings/SettingsActivity.cpp
daveallie: I think this is a reasonable change and imo feels a little nicer, swapping tabs by toggling when the tab row is selected is nice

src/components/UITheme.h
```diff
@@ -0,0 +1,95 @@
...
+  static void drawTabBar(const GfxRenderer& renderer, Rect rect, const std::vector<TabInfo>& tabs, bool selected);
+  static void drawRecentBookCover(GfxRenderer& renderer, Rect rect, const std::vector<RecentBookWithCover>& recentBooks,
+                                  const int selectorIndex, bool& coverRendered, bool& coverBufferStored,
+                                  bool& bufferRestored, std::function<bool()> storeCoverBuffer);
+  static void drawButtonMenu(GfxRenderer& renderer, Rect rect, int buttonCount, int selectedIndex,
+                             const std::function<std::string(int index)>& buttonLabel, bool hasIcon,
+                             const std::function<std::string(int index)>& rowIcon);
+  static Rect drawPopup(const GfxRenderer& renderer, const char* message);
+  static void fillPopupProgress(const GfxRenderer& renderer, const Rect& layout, const int progress);
+  static void drawBookProgressBar(const GfxRenderer& renderer, const size_t bookProgress);
```
daveallie: Following other implementations, let's make UITheme a regular class with a constructor taking `CrossPointSettings::UI_THEME`, create a static instance of the class which will be used everywhere, then all these `static` methods can become instance methods.

src/CrossPointSettings.h
```diff
@@ -137,6 +140,8 @@ class CrossPointSettings {
   uint8_t hideBatteryPercentage = HIDE_NEVER;
   // Long-press chapter skip on side buttons
   uint8_t longPressChapterSkip = 1;
+  // UI Theme
+  uint8_t uiTheme = CLASSIC;
```
daveallie: If we're confident this is just straight up better (which IMO I think it is), and that it's going to benefit the majority of the user base switching over, we should just make Lyra the default. Thoughts?

daveallie (changes requested):
Few bits inline.

Probably the biggest thing is the perceived additional lag when rendering the home screen (just with the initial loading screen popping up). I don't think it's bad enough to block, but would love to find any way to reduce the time.

src/RecentBooksStore.h
CaptainFrito: One reason for this is that the coverBmpPath is actually the path for the thumbnail on the home screen, which depends on the theme. Each theme has a different size for the thumbs. So if we persist this, we'd need to reset it whenever the theme is changed. Or we could persist a generic path and each theme can add their size suffix to it.

src/CrossPointSettings.h
el: I think instead of making it default, we can relase it like this, once we are confident it is feature complete, we can delete Classic.

daveallie: I don't think it's a regression in terms of functionality, just that it mirrors the base theme in some screens (where there are designs and future work to be done). Making it the default with an opt-out feels like the best way to get feedback and onboard new users onto this sooner.

If there are any functional gaps I think it's definitely reasonable to avoid updating the default, will do a full click through of features before we merge, but I believe at worst it's just inconsistent between updated and not-updated screens. @CaptainFrito can confirm.

CaptainFrito:
> Few bits inline.
> 
> Probably the biggest thing is the perceived additional lag when rendering the home screen (just with the initial loading screen popping up). I don't think it's bad enough to block, but would love to find any way to reduce the time.

Managed to gain a few hundred ms thanks to your suggestion to persist the thumbnail bmp path in the RecentBook struct, now we don't need to load the epub/xtc when entering the Home screen, we just load the bmps from that saved path. It's still not instant though.

The lag perception is also emphasized by the fact that we're rendering the rest of the screen elements while still loading the images. Previously the home screen was rendered in one go, but it would take a little while to load the image which would simply delay the entire screen render. Do we want to go that route instead?

I still need to make the static instance change in UITheme and then we're good :)

Uri-Tauber:
@CaptainFrito Awesome work! Personally, I prefer a fast, responsive UI over thumbnails. Would it be possible to add a setting to disable thumbnails on the home screen?

Maybe it could show simple book icons with the names of recent books instead. That would allow instant loading, while still giving users the option to enable thumbnails if they prefer.

CaptainFrito:
> @CaptainFrito Awesome work! Personally, I prefer a fast, responsive UI over thumbnails. Would it be possible to add a setting to disable thumbnails on the home screen?
> 
> Maybe it could show simple book icons with the names of recent books instead. That would allow instant loading, while still giving users the option to enable thumbnails if they prefer.

Oh yeah good idea! Let's save this for another PR though so we don't pile features into this one

CaptainFrito:
PR updated with UITheme as a singleton

src/RecentBooksStore.cpp
```diff
@@ -92,11 +94,12 @@ bool RecentBooksStore::loadFromFile() {
     recentBooks.reserve(count);
 
     for (uint8_t i = 0; i < count; i++) {
-      std::string path, title, author;
+      std::string path, title, author, coverBmpPath;
       serialization::readString(inputFile, path);
       serialization::readString(inputFile, title);
       serialization::readString(inputFile, author);
-      recentBooks.push_back({path, title, author});
+      serialization::readString(inputFile, coverBmpPath);
```
daveallie, line 142: This line will crash when reading existing recent book files. We'll need to bump `RECENT_BOOKS_FILE_VERSION` and then handle the version as 2 / 3.

daveallie (changes requested):
Small change needed for recents

src/RecentBooksStore.cpp
CaptainFrito: Fixed! Thanks

CaptainFrito:
I've improved the upgrade path for recent books coming from versions 1 and 2. Now it will reload the books to get the title, author and cover.

lepislepis:
Since some people only read one book at a time, would it be possible to display just a single featured book on the homepage?

daveallie:
Single featured book can come as a followup PR, keen to get this one merged to close out this monster

Issue #502, opened 2026-01-22
KoSync progress does not sync between Crosspoint-reader and KOReader (Kindle)


oscarsantillana:
### Affected Version

0.15

### Describe the Bug

 KoSync progress does not sync between Crosspoint-reader and KOReader (Kindle) even when both devices use the exact same EPUB file.
The issue appears to be that both clients generate **different document IDs** (`partial_md5_checksum`) from identical files, causing KoSync to treat them as different books.

* **Device A**: Crosspoint-reader

  * Version 0.15

* **Device B**: Kindle + KOReader

  * KOReader version (latest)

### Server

* Autocaliweb 

### Steps to Reproduce

1. Download the same EPUB book from OPDS on both devices
2. Verify both EPUB files are byte-identical (`sha256sum`)
3. Open the book on both devices
4. Enable KoSync on both with binary matching. (tried filename too without luck)
5. Observe that progress sync creates two separate records instead of syncing

### Expected Behavior

Same id is generated so progress can be sync

### Relevant Log Output/Screenshots

```shell
Device A KoSync document ID: be268332cfbeaf6c10722fc5a6f4c2e9
Device B KoSync document ID: f5781ce39df5ff34f5c34548abf41e4f
```

getthething:
I'm having the same issue between crosspoint and my kobo running KOReader. Though I haven't looked at logs. The first time I tried syncing, I pushed from my X4 and pulled down to my Kobo and it went to the correct chapter, but not the exact spot in the chapter. All syncing attempts after that are not working at all.

carsonhicks:
I'm seeing the same issue but with Android for my Koreader device and Booklore as my sync server.

Filename matching isn't really a good option since some non-ascii characters seem to get removed from the filename when downloading an epub file to Crosspoint.

Bapf:
Same here. Have a working setup between KOReader on Kobo, Kindle and Android with binary matching via sync.koreader.rocks, but one crosspoint no progress is found for my books.
Looking forward to the fix from yesterday. :)

edit: Just compiled a version with the change. Works flawlessly now.

PR #529, opened 2026-01-24
fix: Fix KOReader document md5 calculation for binary matching progress sync

carsonhicks:
## Summary

Resolve [KoSync progress does not sync between Crosspoint-reader and KOReader (Kindle)](https://github.com/crosspoint-reader/crosspoint-reader/issues/502)

KOReaderDocumentId::getOffset() - Update the value for the md5 offset calculation to match KOReader.

## Additional Context

I've tested this with a couple of my ebooks and binary matching with KOReader sync seems to be working fine now for both pushing and pulling progress.

---

### AI Usage


itsthisjustin:
Thanks for doing this! I wasn't 100% sure why that specific piece wasn't working

PR #480, opened 2026-01-22
feat: Add new front button layout - Back, Left, Right, Confirm

ishuru:
## Description

This PR adds a fourth front button layout option: **Back, Left, Right, Confirm**.

This arrangement places the navigation buttons (Left/Right) in the center of the button strip, with Back on the far left and Confirm on the far right. This provides an ergonomic alternative for users who prefer having the most-used navigation buttons (Left/Right for page turning) in the center position.

## Changes

- Added `BACK_LEFT_RIGHT_CONFIRM` enum value (value `3`) to `FRONT_BUTTON_LAYOUT`
- Updated `MappedInputManager::mapButton()` to handle the new layout
- Updated `MappedInputManager::mapLabels()` for correct button hint display
- Added new option to the Settings UI
- Updated USER_GUIDE.md documentation

## Button Layout Options (after this PR)

| Option | Physical Layout |
|--------|-----------------|
| Default | Back, Confirm, Left, Right |
| Option 2 | Left, Right, Back, Confirm |
| Option 3 | Left, Back, Confirm, Right |
| **New** | Back, Left, Right, Confirm |

## Testing

- [x] Successfully compiled with PlatformIO
- [x] Tested on physical device

PR #535, opened 2026-01-24
feat: Add new front button layout - Back, Left, Right, Confirm

### LINKED ISSUE #652: Button Layout -> Back, Left, Right, Confirm
When I started reading in horizontal mode I realized this is a convenient layout: It allows for the next and prev buttons to be positioned directly in the center under the thumb. And the back and confirm buttons, separated on different sides, also feel convenient. The device is balanced by its weight and requires no movement while reading.

Image

Also, other conventional variations are already implemented: 
1. Left Right Back Ok
2. Back Ok Left Right
3. Left Back Ok Right

The only one missing is this one: Back, Left, Right, Confirm

Eloren1:
Thanks to @ishuru and this PR #480. I just helped with fixing auto-formatting issues, so that only real changes are displayed in diff. 

This layout started making sense when I tried reading in horizontal mode: It allows for the `next `and `prev `buttons to be positioned directly in the center under the thumb. And the `back `and `confirm `buttons, separated on different sides, also feel convenient. The device is balanced by its weight and requires no movement while reading. 

540147705-58237dd9-51c9-4339-b914-e5d37094b767

## Summary

This PR adds a fourth front button layout option: Back, Left, Right, Confirm. It's essentially a final combination of these buttons.

- Added BACK_LEFT_RIGHT_CONFIRM enum value (value 3) to FRONT_BUTTON_LAYOUT
- Updated MappedInputManager::mapButton() to handle the new layout
- Updated MappedInputManager::mapLabels() for correct button hint display
- Added new option to the Settings UI
- Updated USER_GUIDE.md documentation
- [x] Tested on physical devices

---

### AI Usage

Did you use AI tools to help write this code? _**---**_

Closes #652

### DIFF
--- a/src/MappedInputManager.cpp
+++ b/src/MappedInputManager.cpp
@@ -18,12 +18,13 @@ struct SideLayoutMap {
 };
 
 // Order matches CrossPointSettings::FRONT_BUTTON_LAYOUT.
+// [back, confirm, left, right] are actioned by these physical buttons:
 constexpr FrontLayoutMap kFrontLayouts[] = {
     {HalGPIO::BTN_BACK, HalGPIO::BTN_CONFIRM, HalGPIO::BTN_LEFT, HalGPIO::BTN_RIGHT},
     {HalGPIO::BTN_LEFT, HalGPIO::BTN_RIGHT, HalGPIO::BTN_BACK, HalGPIO::BTN_CONFIRM},
     {HalGPIO::BTN_CONFIRM, HalGPIO::BTN_LEFT, HalGPIO::BTN_BACK, HalGPIO::BTN_RIGHT},
     {HalGPIO::BTN_BACK, HalGPIO::BTN_CONFIRM, HalGPIO::BTN_RIGHT, HalGPIO::BTN_LEFT},
-    {HalGPIO::BTN_BACK, HalGPIO::BTN_LEFT, HalGPIO::BTN_RIGHT, HalGPIO::BTN_CONFIRM},
+    {HalGPIO::BTN_BACK, HalGPIO::BTN_RIGHT, HalGPIO::BTN_CONFIRM, HalGPIO::BTN_LEFT},
 };
 
 // Order matches CrossPointSettings::SIDE_BUTTON_LAYOUT.


nguyennvc:
IMG20260126180958

Hi. Could you also add this combination for inverted users like me? Or even better, could you make it so users can assign buttons in any order they want, similar to the stock firmware?

Eloren1:
@osteotek hmm, I updated the PR for the 0.16.0, checked everything and somehow it doesn't work properly.

In menus: "Back Left Right Open" displayed correctly

But while reading front bottom buttons don't work at all. Side buttons work as Back and Open. Can you please check why it's happening?

src/activities/settings/SettingsActivity.cpp
```diff
@@ -40,9 +40,9 @@ const SettingInfo readerSettings[readerSettingsCount] = {
 
 constexpr int controlsSettingsCount = 4;
```
osteotek, line 41: probably need to update this count

src/MappedInputManager.cpp
osteotek, line 26: also need to add new option here

osteotek (changes requested):

src/activities/settings/SettingsActivity.cpp
Eloren1: Nope, it's related to the total amount of parameters in this tab. The next tip worked, thanks!

ishuru:
I think we can abstract the button layout by grouping as orientation, I will try it out.
> IMG20260126180958
> 
> Hi. Could you also add this combination for inverted users like me? Or even better, could you make it so users can assign buttons in any order they want, similar to the stock firmware?

Eloren1:
Now it works but incorrectly. In menus and while reading it's actually: BACK + RIGHT + CONFIRM + LEFT (non existent layout). But the legends on screen look correct. And the code itself looks correct... Not ready for merging. Unknown remapping issue?

ishuru (approved):

aaroncunliffe:
Hi @Eloren1, your changes are close from what I can see.

The layout definition you have [here](https://github.com/crosspoint-reader/crosspoint-reader/pull/535/changes#diff-1b8ccb9439d5a6fec0c5db8fa94c1ebbfb4cbbff8899bd5430b2459b034220feR26) is not left to right button order. It's actually mapping for the actions [back][confirm][left][right], you plug in what button you want to perform that action in order.

You can see this in the existing definition on master for `LEFT_BACK_CONFIRM_RIGHT`, which maps to `{BTN_CONFIRM, BTN_LEFT, BTN_BACK, BTN_RIGHT}`.  This shows it's not just a left to right list.

For your layout it should be:
`{HalGPIO::BTN_BACK, HalGPIO::BTN_RIGHT, HalGPIO::BTN_CONFIRM, HalGPIO::BTN_LEFT}` 

That means:
`[back]` is actioned by the button `BTN_BACK`
`[confirm]`  is actioned by the button `BTN_RIGHT`
`[left]` is actioned by the button `BTN_CONFIRM`
`[right]`  is actioned by the button `BTN_LEFT`

Eloren1:
Thank you, I get it now. Although I need to stare at it for a few minutes to figure out which function corresponds to which button. Fixed!

osteotek:
@Eloren1 Would this work for your case? https://github.com/crosspoint-reader/crosspoint-reader/pull/664

daveallie:
Closing in favour of #664

PR #552, opened 2026-01-26
fix: correct home cover dimensions

pablohc:
## Summary

  Fixes the cover image dimensions for the home screen (Continue Reading card).

  - Modifies image generation to use 400px height with proportional width
  - Optimizes rendering by removing unnecessary centering calculations
  - Updates Epub and Xtc classes to use the new fixed dimension system

## Additional Context

**Performance:** Images use 1-bit format for faster rendering, but loading times still need optimization.

---

### AI Usage


pablohc:
Replaced by: #683

PR #564, opened 2026-01-27
fix: missing front layout in mapLabels()

osteotek:
## Summary

* adds missing front layout to mapLabels function


daveallie:
Ahh sorry, I missed that in a merge

PR #566, opened 2026-01-27
Draft: Add new sleep screen mode, Cover else Custom

Marbelymarble:
## Summary

- Add another sleep screen mode that will either display the current book cover if available and a custom sleep screen if a book cover is not available. 

## Additional Context

- As this is simply a combination of previously established sleep screen modes, not many changes have been made. 
- A possible consideration may be compiled size as we are approaching the limit of the onboard flash. This PR adds about 1.8% to the flash usage or ~200k bytes. Lots of code reused, can (and maybe will) be optimised. 

---

### AI Usage

NO

### DIFF
--- a/src/CrossPointSettings.h
+++ b/src/CrossPointSettings.h
@@ -15,7 +15,7 @@ class CrossPointSettings {
   CrossPointSettings(const CrossPointSettings&) = delete;
   CrossPointSettings& operator=(const CrossPointSettings&) = delete;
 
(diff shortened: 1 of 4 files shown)


Marbelymarble:
After reading a (now deleted?) comment, i agree that it is worth it to somehow integrate this into renderCoverSleepScreen(). Will pedal back to this later. (Excuse my sloppy handling of github PRs, most of my work has so far been handled off-line)

daveallie:
Ended up being added in https://github.com/crosspoint-reader/crosspoint-reader/pull/582

PR #572, opened 2026-01-27
fix: do not wake up on usb connected

ngxson:
## Summary

Fix: https://github.com/crosspoint-reader/crosspoint-reader/pull/482#issuecomment-3806291939 (cc @osteotek)

Tested cases:
- sleep / wake up when usb not connected
- sleep / wake up when usb connected
- sleep when on battery ; wake up via power button when usb connected
- sleep when usb connected ; wake up via power button when on battery

---

### AI Usage

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -205,6 +205,15 @@ void waitForPowerRelease() {
   }
 }
 
+// TDO: move this to HAL in the future
+void hwDeepSleepStart() {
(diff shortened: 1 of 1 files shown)


osteotek:
I think I have another solution for this, let me try one thing first..

osteotek:
@ngxson tried to make it more readable, though ultimately same logic - https://github.com/crosspoint-reader/crosspoint-reader/pull/576

PR #510, opened 2026-01-23
feat: rework CrossPointSettings

borisfaure:
## Summary

Multiple goals are achieved with this change:

- make descriptions of settings close to their definitions: the labels are in the same file as the configuration declarations
- settings validation on loading with reset to default in case of bad value
- (de)serialization is based on a single list of descriptors: single point of truth
- possibly to have default values for the String configuration type
- more `constexpr` to reduce RAM usage
- less hardcoded values
- more maintainable

## RAM Usage:
 From
```
RAM:   [===       ]  32.5% (used 106516 bytes from 327680 bytes)
Flash: [========= ]  94.9% (used 6217240 bytes from 6553600 bytes)
```
 To
```
RAM:   [===       ]  32.3% (used 105844 bytes from 327680 bytes)
Flash: [========= ]  94.8% (used 6213636 bytes from 6553600 bytes)
```

## Boot config validation:
```
[1186] [SD] SD card detected
[1192] [CPS] Invalid value for Status Bar, resetting to default [1192] [CPS] Settings loaded from file
```

---

### AI Usage

While CrossPoint doesn't have restrictions on AI tools in contributing, please be transparent about their usage as it 
helps set the right context for reviewers.

Did you use AI tools to help write this code? _**PARTIALLY**_

PR #576, opened 2026-01-27
fix: don't wake up after USB connect

osteotek:
## Summary

* fixes problem that if short power button press is enabled, connecting device to usb leads to waking up

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -316,6 +316,7 @@ void setup() {
   Serial.printf("[%lu] [   ] Starting CrossPoint version " CROSSPOINT_VERSION "\n", millis());
 
   setupDisplayAndFonts();
+
   exitActivity();
   enterNewActivity(new BootActivity(renderer, mappedInputManager));
 


lib/hal/HalGPIO.cpp
```diff
@@ -44,12 +45,20 @@ bool HalGPIO::isUsbConnected() const {
...
+  const bool usbConnected = isUsbConnected();
   const auto wakeupCause = esp_sleep_get_wakeup_cause();
   const auto resetReason = esp_reset_reason();
-  if (isUsbConnected()) {
-    return wakeupCause == ESP_SLEEP_WAKEUP_GPIO;
-  } else {
-    return (wakeupCause == ESP_SLEEP_WAKEUP_UNDEFINED) && (resetReason == ESP_RST_POWERON);
+
+  if ((wakeupCause == ESP_SLEEP_WAKEUP_UNDEFINED && resetReason == ESP_RST_POWERON && !usbConnected) ||
+      (wakeupCause == ESP_SLEEP_WAKEUP_GPIO && resetReason == ESP_RST_DEEPSLEEP && usbConnected)) {
```
ngxson, line 54: I think the original `if (isUsbConnected) .. else ..` looks more clean. Logically say, `(A & B) || (!A & C)` is equivalent to `if (A) then B else C`

daveallie (approved):
Thanks!

osteotek:
@daveallie I think there is a bug in this logic preventing device to wakeup, can we revert it

daveallie:
yep

osteotek:
@daveallie sorry for early panic, seems like bootloop was caused by my cache containing settings from other branch. I've tested this changes again (after merging master), works as intended. Opened second PR - https://github.com/crosspoint-reader/crosspoint-reader/pull/644

borisfaure:
PR #510 might help on bad configs

PR #587, opened 2026-01-28
feat: holding back button while booting, boots to home screen as a mean of escaping boot loop

GenesiaW:
## Summary

  - Allows back button to be held to escape boot loops when reader activity crashes and device attempts to boot to previous state
  - Reduces the need of removing SD card and access to another device to delete the `/.crosspoint/state.bin`

  - Back button can be held while booting to boot to home screen
  - Update of User Guide section to include this feature 

## Additional Context

  specific areas to focus on).

---

### AI Usage

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -294,10 +294,22 @@ void setup() {
   SETTINGS.loadFromFile();
   KOREADER_STORE.loadFromFile();
 
-  if (gpio.isWakeupByPowerButton()) {
-    // For normal wakeups, verify power button press duration
-    Serial.printf("[%lu] [   ] Verifying power button press duration\n", millis());
-    verifyPowerButtonDuration();
+  switch (gpio.getWakeupReason()) {
+    case HalGPIO::WakeupReason::PowerButton:
+      // For normal wakeups, verify power button press duration
+      Serial.printf("[%lu] [   ] Verifying power button press duration\n", millis());
+      verifyPowerButtonDuration();
+      break;
+    case HalGPIO::WakeupReason::AfterUSBPower:
+      // If USB power caused a cold boot, go back to sleep
+      Serial.printf("[%lu] [   ] Wakeup reason: After USB Power\n", millis());
+      gpio.startDeepSleep();
+      break;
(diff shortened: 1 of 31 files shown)


osteotek:
That’s much needed, thanks. Do you think we can automatically escape the boot loop? Perhaps we could add a flag alongside the book cache.

GenesiaW:
> That’s much needed, thanks. Do you think we can automatically escape the boot loop? Perhaps we could add a flag alongside the book cache.

That's a great idea, that will definitely help improve the UX if any boot loop were to occur. I will take a look into that and see what can be done

osteotek:
Yes, perhaps add a flag after starting the processing of an EPUB file and clear it after it finishes. Upon boot, if such a flag is present, we know that the previous load had failed and don’t start again.

GenesiaW:
Hi @osteotek,

I have added the changes as requested. Instead of using a flag, I chose to use a counter to track the number of attempts as an app wide state. The attempts counter will be reset on successful exit of any form of reader (epub, txt, xtc) activity which implies that it was able to exit safely.

A decision that I made was to allow the reader have up to 3 attempts to load the book. I'm not sure if this retry mechanism is required

pablohc:
@GenesiaW Often, wrong configurations loaded in settings.ini generate that loop. I suppose it could also be added to the deletion by holding down the button when starting the device.

src/main.cpp
```diff
@@ -311,13 +311,16 @@ void setup() {
   APP_STATE.loadFromFile();
   RECENT_BOOKS.loadFromFile();
 
-  if (APP_STATE.openEpubPath.empty()) {
+  // Boot to home screen directly when back button is held or when reader activity crashes 3 times
+  if (APP_STATE.openEpubPath.empty() || mappedInputManager.isPressed(MappedInputManager::Button::Back) ||
+      APP_STATE.readerActivityLoadCount > 2) {
```
osteotek: let's make it > 0
```suggestion
      APP_STATE.readerActivityLoadCount > 0) {
```

daveallie (approved):

PR #455, opened 2026-01-20
feat: Custom Font Implementation

ruby-builds:
## Summary

This document details the modifications and enhancements made in this fork/branch compared to the upstream master branch (`https://github.com/crosspoint-reader/crosspoint-reader`).

## 1. Custom Font Support Framework

A complete system for loading and rendering custom fonts from the SD card was implemented.

### New Core Components
*   **`lib/EpdFontLoader/EpdFontLoader.{cpp,h}`**: A new library responsible for discovering, validating, and ensuring custom fonts are loaded into the renderer. It includes safety fallbacks to the default "Bookerly" font if a custom font fails to load.
*   **`src/managers/FontManager.{cpp,h}`**: A singleton manager that scans the `/fonts` directory on the SD card, parses `.epdfont` headers, and creates `EpdFontFamily` instances for valid fonts.
    *   **Update**: Now supports both **V0** (Legacy Python script) and **V1** (Web Converter) font formats.
    *   **Update**: Implements flexible filename parsing (supports `Family_Style_Size.epdfont`, `Family-Style-Size.epdfont`, and single-file variants).
*   **`lib/EpdFont/CustomEpdFont.{cpp,h}`**: A new font implementation that reads glyph data directly from the binary `.epdfont` files on the SD card, implementing an LRU cache for bitmaps to optimize RAM usage.
    *   **Update**: Handles dynamic glyph stride (13 bytes for V0, 16 bytes for V1) and header sizes (48 bytes for V0, 32 bytes for V1).
*   **`src/activities/settings/FontSelectionActivity.{cpp,h}`**: A new UI screen allowing the user to select from available custom fonts found on the SD card.
*   **[lib/EpdFont/EpdFontStyles.h](cci:7://file:///home/ruby/Documents/Dev/CrossPoint/crosspoint-reader/lib/EpdFont/EpdFontStyles.h:0:0-0:0)**: Added to define styles (Regular, Bold, Italic, BoldItalic) for better font family management.

### Tooling Updates
*   **[lib/EpdFont/scripts/fontconvert.py](cci:7://file:///home/ruby/Documents/Dev/CrossPoint/crosspoint-reader/lib/EpdFont/scripts/fontconvert.py:0:0-0:0)**: Rewritten to generate binary `.epdfont` files (V0 format) with correct offset calculations.
*   **Web Converter Support**: Added compatibility with fonts generated by `epdfont.clev.app` (V1 format).

## 2. EPUB Rendering & Parsing Improvements

[...]

PR #589, opened 2026-01-28
feat: Overhaul font format into CrossPoint font

daveallie:
## Summary

* This is an overhaul of the font format to use a similar initial structure, but leverage the Group5 compression from Larry Bank's bb_epaper https://github.com/bitbank2/bb_epaper/blob/main/src/Group5.h
  * We can get about 1.7:1 -> 2:1 compression/reduction of flash usage using this new layout
* It also inlines all the style variants into a single byte stream, so building a `.cpf` binary file to use at load time leveraging some ideas from Larry's implementation and ideas in https://github.com/crosspoint-reader/crosspoint-reader/pull/455

## Additional Context

---

### AI Usage


cbix:
@daveallie keep in mind Group5 uses GPL-3.0-or-later (restrictive) while CrossPoint is MIT (permissive), so the resulting binary (if it contains the compiled Group5 lib) could only be distributed under GPL as well. IANAL but you might consider one of these options
1. making everything GPL3
2. keep MIT but add separate build targets with/without Group5 so people can build an MIT-compatible firmware
3. [clean-room](https://en.wikipedia.org/wiki/Clean-room_design) reimplementation of the format under MIT (feels a bit excessive tho)
4. it seems Group5 changed its license [here](https://github.com/bitbank2/bb_epaper/releases/tag/2.0.2), can't tell if the previous license is more or less permissive and would allow use of that older version under MIT
5. talk to the Group5 author, maybe they can make this library MIT-compatible

Someone with more experience in software licenses should add more context here but I'm just trying to raise the awareness of this potential issue before it's too late :) Also, thanks for all your work, I couldn't be happier with my new e-reader ❤️

daveallie:
Thanks for your concern @cbix, I did actually talk with @bitbank2 (Group5's author) and confirm that it was ok to include inside of CrossPoint. Perhaps there is a better way to reference this fact through a custom license file.

iandchasse:
Hi, just checking in on this concept. I looked into using my own version of custom font implementation on my own hardware, which is S3-based and has 2MB of PSRAM. I was able to store .epdfont files on SD similar to #455, but I was also able to take advantage of stb_truetype.h and generate these files directly on device from .ttf/.otf files rather than pre-converted ones from the Python script. I first lazily allocated it to the PSRAM, but I tried to challenge myself to implement chunking/streaming to only use SRAM, but that was much less stable and crashed with some larger sized fonts. 

IMG_8626
IMG_8627
IMG_8629


The above images are from my version that uses PSRAM, not the buggy SRAM-exclusive implementation, but only attaching to propose the general concept. 

I think it has some potential, but I’d like to see what happens with this PR first to see if it’s feasible to implement on top of the eventual custom font engine that’s chosen, whether it’s this or #455

znelson:
Closing due to inactivity. Dave, we're discussing fonts in #1693.

Issue #570, opened 2026-01-27
Status bar "Progress Bar" not consistently aligned with screen edge


lukestein:
### Affected Version

#438

### Describe the Bug

@alexfaria introduced a great new “Full w/ Progress Bar” status bar option in #438 .

Depending on screen orientation (and perhaps specific hardware assembly tolerances), the bar can either be too small (hard to see) or not quite at the screen edge, with a white border visible. I do know my screen will draw black or white all the way out further than the bezels, so I wonder if there's a way to make the progress bar bleed further outside the margin.

~~It would be great if anyone who can test this PR build could also see how it does on their device and report [per discussion here](https://github.com/crosspoint-reader/crosspoint-reader/pull/438#issuecomment-3806275609).~~

Any **feedback on your device** at #618 can help figure out how best to move forward.

### Steps to Reproduce

1. Install including #438 
2. Set Full w/ Progress Bar status bar option
3. Go to reader mode in various orientations

### Expected Behavior

Looks perfect

### Relevant Log Output/Screenshots

![Image](https://github.com/user-attachments/assets/0df5ba03-6901-4255-948b-269aed2faf1c)

![Image](https://github.com/user-attachments/assets/4909fe8f-5a44-4a31-8ab2-775d9454b76e)

![Image](https://github.com/user-attachments/assets/41b750e9-4c2a-41a0-847b-646eafeed7bd)

![Image](https://github.com/user-attachments/assets/5234f0be-826b-4c16-8f1e-306306618fc9)

DavidOrtmann:
I would say it's working perfectly on my device on all orientations:

![Image](https://github.com/user-attachments/assets/f56396f3-ebeb-4188-b5b9-5ebc3a613751)
![Image](https://github.com/user-attachments/assets/66b45ee8-c97d-42a2-972d-20213e96b141)
![Image](https://github.com/user-attachments/assets/cbd875f4-224d-4439-aa90-3f3bb5249997)
![Image](https://github.com/user-attachments/assets/a0a3f704-9c0d-4bfd-b04c-c884b26a6b0f)
![Image](https://github.com/user-attachments/assets/58638edb-2430-46be-a7e7-a327756b7367)

lukestein:
> I would say it's working perfectly on my device on all orientations:

Yup, yours looks perfect. We'll see if we can get some broader feedback to figure out if there's a code improvement in order.

Eloren1:
[...]

PR #602, opened 2026-01-28
fix: Adjust progress bar height based on orientation

Eloren1:
## Summary

BOOK_PROGRESS_BAR_HEIGHT now returns:

- [x] 4 in portrait vertical mode (like before)
- [x] 2 in landscape horizontal mode (thinner)

## Additional Context

In portrait mode, the existing `height = 4` renders correctly.

In landscape mode, the progress bar appear thicker than intended. It becomes bolder than the font and is visually distracting.

Reducing the height to 2 in landscape restores visual balance and matches the intended thin appearance just like in vertical orientation.

## Comparison

Look at the same random word and notice how the progress bar feels too thick out of sight. In vertical orientation it doesn't look that thick. Note that in real life the difference is even more visible.

| Horizontal w/ height 4 | Horizontal w/ height 2 |
|----------|----------|
| IMG20260129001330 | IMG20260129001404 |

---

### AI Usage

### DIFF
--- a/src/ScreenComponents.cpp
+++ b/src/ScreenComponents.cpp
@@ -59,10 +59,13 @@ int ScreenComponents::bookProgressBarHeight() {
   const bool isVertical = (orientation == CrossPointSettings::ORIENTATION::PORTRAIT ||
                            orientation == CrossPointSettings::ORIENTATION::INVERTED);
 
+  constexpr int bookProgressBarVerticalHeight = 4;
+  constexpr int bookProgressBarHorizontalHeight = 2;
+
   if (isVertical) {
-    return ScreenComponents::BOOK_PROGRESS_BAR_VERTICAL_HEIGHT;
+    return bookProgressBarVerticalHeight;
   } else {
-    return ScreenComponents::BOOK_PROGRESS_BAR_HORIZONTAL_HEIGHT;
+    return bookProgressBarHorizontalHeight;
   }
 }
 
--- a/src/ScreenComponents.h
+++ b/src/ScreenComponents.h
@@ -13,8 +13,6 @@ struct TabInfo {
 
 class ScreenComponents {
  public:
-  static const int BOOK_PROGRESS_BAR_VERTICAL_HEIGHT = 4;
-  static const int BOOK_PROGRESS_BAR_HORIZONTAL_HEIGHT = 2;
   static int bookProgressBarHeight();
 
   static void drawBattery(const GfxRenderer& renderer, int left, int top, bool showPercentage = true);


src/ScreenComponents.h
```diff
@@ -13,7 +13,7 @@ struct TabInfo {
 
 class ScreenComponents {
  public:
-  static const int BOOK_PROGRESS_BAR_HEIGHT = 4;
+  static int bookProgressBarHeight();
```
osteotek, line 16: maybe just make second constant? like 
```
static const int BOOK_PROGRESS_BAR_VERTICAL_HEIGHT = 4;
static const int BOOK_PROGRESS_BAR_HORIZONTAL_HEIGHT = 2;
```

lukestein:
Would love to see some more feedback on the current implementation and on this PR per #570 to see what looks best for different people's devices

Eloren1:
It's not related to that hardware / margins problem, it looks good to me. All I did is to make horizontal bar thickness look like vertical one that is thin already.

And yep it looks the same in Landscape CCW and Landscape CW, so it's not like the display is shifted to one side

osteotek (changes requested):

src/ScreenComponents.h
Eloren1: There are 5+ usages of it, so using ternary operator every time is not an option

src/ScreenComponents.cpp
```diff
@@ -48,9 +49,21 @@ void ScreenComponents::drawBookProgressBar(const GfxRenderer& renderer, const si
...
+  renderer.fillRect(vieweableMarginLeft, progressBarY, barWidth, bookProgressBarHeight(), true);
+}
+
+int ScreenComponents::bookProgressBarHeight() {
+  const auto orientation = SETTINGS.orientation;
+  const bool isVertical = (orientation == CrossPointSettings::ORIENTATION::PORTRAIT ||
+                           orientation == CrossPointSettings::ORIENTATION::INVERTED);
+
+  if (isVertical) {
+    return 4;
```
jeffcardillo: Would it make sense to keep the constants and return them here instead of the integers?

Eloren1: These values are only meaningful to this function and nothing else, so no

osteotek: these should be constants, sorry. Add them in place of previous constant

Eloren1:
image
 
@osteotek  These ones for example, they are used only once and only here. Just like with my function. Moving them somewhere outside the method is a complication. So I did it similar to the example in the screenshot.

Issue #561, opened 2026-01-27
Potential mitigation for screen fading issue


allgoewer:
### Affected Version

0.15.0

### Describe the Bug

I attached a video of the sceen-fading issue which has been discussed on https://www.reddit.com/r/xteinkereader/ previously.

[Fading issue (Youtube)](https://www.youtube.com/watch?v=5SsKPg7kYW0)

I made some observations:

* It only happens in bright sunlight, so it seems to be UV-related
* It _does not_ happen if I cover the space center above the lower-screen buttons with my thumb
* It seems not to be happening after entering sleep
* _Not seen in the Video_: If covering parts of the screen from direct sunlight, that part won't fade.

Some assumptions:

* It might be related to the screen being powered?
* Most likely a hardware issue, but maybe mitigatable in software?
* Maybe the issue could be partially mitigated in software be powering down the screen after updating?
  * An option to force the screen to be powered-down (for the beach dwellers) might be an option, should powering down / up the screen take too long for day-to-day use.
* Maybe putting some UV masking tape above the lower-screen buttons might help
 

### Steps to Reproduce

1. Force a screen refresh in bright sunlight
2. Observe the screen fading to white

### Expected Behavior

Screen should not fade to white

### Relevant Log Output/Screenshots

```shell

```

allgoewer:
Got an idea what's going on: https://www.reddit.com/r/xteinkereader/comments/1qo90uw/comment/o1zv2sr/



I have a theory what's going on here:

1. According to this repo (https://github.com/open-x4-epaper/community-sdk/blob/c39f253a7dabbc193a8d7d310fb8777dca0ab8f1/libs/display/EInkDisplay/doc/SSD1677_GUIDE.md) which contains the driver crosspoint is using, the X4 is using the SSD1677 EPD driver chip.

1. According to this datasheet (https://cursedhardware.github.io/epd-driver-ic/SSD1677.pdf) Chapter 3, Ordering Information, the SSD1677 is packaged as "Gold Bump Die", this means that the driver IC is most likely a bare IC with no resin-packaging around it, which will make it succeptible to UV radiation.

1. Look at the images of the PCB and the screen in this teardown: https://min.news/en/digital/3d78670f66f36741698fd086320d6488.html . No mention of a driver IC, I bet the IC is part of the screen assembly, probably on top of the assembly, right underneath the plastic housing of the X4.

1. That would explain why shielding the IC from sun exposure with my thumb mitigates the issue.

[...]

PR #603, opened 2026-01-28
feat: Implement fix for sunlight fading issue

allgoewer:
## Summary

The goal of this PR is to deliver a fix for or at least mitigate the impact of the issue described in #561 

This PR includes a new option "Sunlight Fading Fix" under "Settings -> Display".

When set to ON, we will disable the displays analog supply voltage after every update and turn it back on before the next update.  

## Additional Context

* Until now, I was only able to do limited testing because of limited sunlight at my location, but the fix seems to be working. I'll also attach a pre-built binary based on 0.16.0 (current master) with the fix applied to the linked ticket, as building this fix is a bit annoying because the submodule open-x4-sdk also needs an update.
* [PR in open-x4-sdk](https://github.com/open-x4-epaper/community-sdk/pull/15) needs to be merged first, we also need to add another commit to this here PR, updating this dependency.
* I decided to hide this behind a default-OFF option. While I'm not really concerned that this fix might potentially damage the display, someone more knowledgeable on E-Ink technology could maybe have a look at this.
* There's a binary attached in the linked issue, if someone has the required sunlight to test this in-depth.

---

### AI Usage


allgoewer:
I still got no feedback from anyone regarding the effectiveness of this PR. I was able to do only limited testing due to weather conditions but it seemed to do the trick.

osteotek:
as discussed in https://github.com/crosspoint-reader/crosspoint-reader/issues/561#issuecomment-3840339596 seems to be working.

osteotek (dismissed):

daveallie (approved):

cbix:
Nice fix, I got the black version anyway but would this theoretically save battery as well?

allgoewer:
In theory, yes. But I didn't measure.

insunaa:
doesn't actually seem to work for me :( but setting the refresh interval to every page-turn keeps the text readable enough

allgoewer:
@insunaa Would you mind providing a short clip?

Are you sure there is no difference between enabled / disabled fix?
With the fix disabled, you should see the screen completely fading to white over the period of a minute or so, with the fix enabled the fading should stop as soon as the screen-update finishes.
There will still be fading during the screen update, even with the fix enabled. That is expected.

pablohc:
> doesn't actually seem to work for me :( but setting the refresh interval to every page-turn keeps the text readable enough

Are you sure it's because of the sun? I perceive degradation of the screen, and I always have the two X4s (black and white) in my room without any exposure to the sun. Somebody has made a post in reddit comparing 0.16 vs 1.1.0 and the old version of the firmware showed more contrast in the black screen than in the newer FW.

insunaa:
@pablohc some people on reddit speculated it was caused by infrared light heating the ink particles and changing their magnetism for a short bit

@allgoewer I'll make a video as soon as there's sunlight in my area again. For me the fading isn't just the dark pixels turning light, it's also the light pixels splotchily turning dark and after-images of the last ~5-10 refreshes appearing. this happens much more strongly if I am actively flipping pages, but it also happens to the book cover when the device is in deep sleep (as in: put it to sleep while not affected by sunlight and then several minutes later exposing it to direct sunlight)

it also doesn't seem to be affected by UV-A (365nm), I have a UV-A flashlight and neither shining it at the e-ink screen nor shining it between the buttons or into the USB-C port causes any noticeable changes to the display content (regardless of whether the fix is enabled or not)

allgoewer:
Yeah, that sounds like a completely different issue.
Nonetheless, really wanna see the vid :-)

Joseph-DiGiovanni:
Other than the unknown risk of damage to the display, is there any downside to using this option? I don't see any obvious difference in my testing and I assume nobody has reported a damaged display caused by this.

If not I'd like to suggest making it the default behavior and removing the toggle in settings. I don't see an obvious reason it would cause damage and as previously stated it potentially uses less power as well.

Issue #559, opened 2026-01-27
Feature Request: File Transfer feature to allow moving file location


lyndahu:
Just started using CrossPoint because of the KoSync capabilities, and have been really enjoying it so far. One thing I miss from the OFW are the capabilities of the wireless transferring feature. I like not having to plug the SD card in to get books on to the device, but the transfer web page is missing the ability to move file locations.

On the OFW you can rename the path for an individual file and move where it is stored (for example, I would move a book I finished from `Fantasy` to `Finished` folder). I don't see anything comparable in the CrossPoint web portal.

Understand it might not be possible since it is not the official firmware.. but curious if other people want this ability to also be available in CrossPoint and if it is possible! Thank you!

ngude:
i frequently find myself trying to move files as well. i have since just started ejecting the sd card and using a reader to manage and it's just as simple

Tritlo:
I'm looking into this.

PR #619, opened 2026-01-29
feat: add burger menu for file renaming and relocation

LZando:
## What is the goal of this PR?

Introduce a burger (overflow) menu that allows users to rename files and change their location directly from the file item actions.

image

---

## UI / Frontend

- Added a burger (overflow) menu to file items for quick access to actions.
- Implemented modals and related logic for file and folder renaming and moving.
- Relaxed filename validation rules to support spaces  
  _(e.g., `Philosophy of sd.epub`)_.

---

## Backend / Logic

- **BookCacheManager**
  - Introduced a new utility that automatically migrates the cache folder  
    (metadata, reading progress, pre-rendered chapters) when a book’s path changes.
- **Recursive migration**
  - Moving a folder now triggers a recursive cache update for all contained books.
- **State persistence**
  - Updated **RecentBooksStore** to automatically refresh file paths in the *Recent Books* list.
  - Updated **CrossPointState** to keep the last opened book in sync after path changes.

---

## AI Usage

While CrossPoint does not restrict the use of AI tools for contributions, transparency is encouraged to provide proper context for reviewers.

**Did you use AI tools to help write this code?**  
**Answer:** PARTIALLY


lukestein:
Closes #661 and #663

lukestein:
Closes #559

src/util/BookCacheManager.cpp
```diff
@@ -0,0 +1,103 @@
...
+  if (oldPath == newPath) return true;
+
+  // Update Recent Books list
+  RECENT_BOOKS.updatePath(oldPath.c_str(), newPath.c_str());
+
+  // Update last opened book state if matches
+  if (CrossPointState::getInstance().openEpubPath == oldPath.c_str()) {
+    CrossPointState::getInstance().openEpubPath = newPath.c_str();
+    CrossPointState::getInstance().saveToFile();
+  }
```
daveallie: For consistency, we should use `APP_STATE` instead of `CrossPointState::getInstance()` here.

daveallie (changes requested):
Small nit, fixed up a small bug when testing. Will fix this up to get it merged into this release

daveallie (dismissed):

daveallie:
After reading both implementations, going to merge https://github.com/crosspoint-reader/crosspoint-reader/pull/630.

If you'd like to update this to focus on the index migration, that can be done, otherwise, we can consider this solved and close it.

Tritlo:
Great @LZando should we work together on adding updated cache management on rename and move?

LZando:
> Great @LZando should we work together on adding updated cache management on rename and move?

I would love to work together, even though my availability for the next three weeks will be limited due to work, is it a problem?

Tritlo:
no worries, just ping me when you're ready! Though I don't think I can contribute much, you are more familiar with the cache stuff.

LZando:
> no worries, just ping me when you're ready! Though I don't think I can contribute much, you are more familiar with the cache stuff.

I will be back with some architecture ideas

LZando:
> no worries, just ping me when you're ready! Though I don't think I can contribute much, you are more familiar with the cache stuff.

I’m back with two approaches.

1) I was thinking that the file hash cannot be based on any information that the user can change inside the file. 
So the values we can rely on are:
- File size
- File hash

I tried to run a small script using a **sparse hash**, which uses the file size + the beginning and the end of the file. (in my test 4k of the head and 8kn of the tail where there is the ZIP Central Directory, which is unique to every specific compilation of a file.)

Pros of this approach:
- Independence: With a sparse hash, we don’t care where the file is located or what its name is; the file’s properties are what determine the hash.

Cons of this approach:
- Slower: The previous method was obviously faster because it only required hashing the path, so the SH in my test result slower 0.01ms (legacy) vs 55ms (Sparse), is possible to bring the 55ms to 20ms extracting less information but with an higher risk of collisions (10^-16 vs 10^-9, still hard it can happen).

Shouldn’t be a big deal on a enik device, but obv is important to pay attention to performance on ESP32.

2) The other approach is closer to the solution I proposed in my PR: using a dictionary to store the book attributes. However, this could be problematic in the future because we would always need to keep the dictionary and the book data in sync.

Probably the first one is better because easier, what do you think?

Tritlo:
Good points. 

For such a low powered device, I don't think we should worry about collisions. For 10^-9, you'd need ~100 million samples to even have 10% chance of a collision. Given that there have only been ~130 million books written EVER, I think that should be fine.

daveallie:
Will move this one into draft while you two continue to collaborate, just so that that myself and other reviewers it's clear that this doesn't need additional review eyes.

LZando:
> Good points.
> 
> For such a low powered device, I don't think we should worry about collisions. For 10^-9, you'd need ~100 million samples to even have 10% chance of a collision. Given that there have only been ~130 million books written EVER, I think that should be fine.

Great! Give me two weeks and I will try to scratch the first version

znelson:

PR #630, opened 2026-01-31
feat: rename and move in file manager

### LINKED ISSUE #559: Feature Request: File Transfer feature to allow moving file location

Tritlo:
## Summary

This adds renaming and moving files to the File Manager

New `/move` and `/rename` endpoints, and corresponding modals and icons added. Uses the `file.rename()` function, after sanity checking.

## Additional Context

  specific areas to focus on).

Fixes #559, #661, #663. Only touches the File Manager, so low risk of affecting other systems.

Simpler than #619, at the cost of not migrating the cache of renamed books.

image
image
image

---

### AI Usage

I used Codex for the implementation itself, and then carefully reviewed the code myself. As this is a simple change and only to the webserver, it is low risk.

### DIFF
--- a/src/network/CrossPointWebServer.cpp
+++ b/src/network/CrossPointWebServer.cpp
@@ -44,6 +44,36 @@ void clearEpubCacheIfNeeded(const String& filePath) {
     Serial.printf("[%lu] [WEB] Cleared epub cache for: %s\n", millis(), filePath.c_str());
   }
 }
+
+String normalizeWebPath(const String& inputPath) {
+  if (inputPath.isEmpty() || inputPath == "/") {
+    return "/";
+  }
+  std::string normalized = FsHelpers::normalisePath(inputPath.c_str());
+  String result = normalized.c_str();
+  if (result.isEmpty()) {
+    return "/";
+  }
+  if (!result.startsWith("/")) {
+    result = "/" + result;
+  }
+  if (result.length() > 1 && result.endsWith("/")) {
+    result = result.substring(0, result.length() - 1);
+  }
+  return result;
+}
+
+bool isProtectedItemName(const String& name) {
+  if (name.startsWith(".")) {
+    return true;
+  }
+  for (size_t i = 0; i < HIDDEN_ITEMS_COUNT; i++) {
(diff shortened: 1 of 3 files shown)


LZando:
Hi @Tritlo !

I have a suggest, implementing the same [feature](https://github.com/crosspoint-reader/crosspoint-reader/pull/619) I faced one problem that you should consider.

Book metadata such as the reading index is saved using a hash of the file path. So if you change the file name or move the file to a different path, you’ll lose that data.

Tritlo:
I mention that in the PR. I think it's better to just rebuild the index than to migrate, it just adds extra maintenance and complexity.

LZando:
> I mention that in the PR. I think it's better to just rebuild the index than to migrate, it just adds extra maintenance and complexity.

True, I missed it!

LZando:
> I mention that in the PR. I think it's better to just rebuild the index than to migrate, it just adds extra maintenance and complexity.

Would you like to collaborate on building a new index solution?
That way, we can propose the design of a more maintainable solution that incorporates the changes from our PR.

Tritlo:
that could work. Let's wait and hear what @daveallie has to say and go from there.

lukestein:
Tested on device. Moving and renaming files both seem to work exactly as expected, and old versions are removed from Recent files view (which I regard as a feature rather than a bug, especially since there's no other obvious way to get something off the Recents list.

Thank you! 🎉

lukestein:
Closes #661 and #663

Unintendedsideeffects (approved):

Befbo:
Was it planned to be able to rename folders too? Or only files?

PR #633, opened 2026-01-31
feat: Double shift to turn on caps mode

gokulp01:
## Summary

While typing in the passwords for wifi etc. I realized that it is very cumbersome to keep hititng shift for consecutive capital letter characters. This PR adds a common feature that is available on all modern devices i.e., double shift to turn on caps. 

## Additional Context

The changes are minimal, the hardcoded value is the double-tap window (currently it is 500ms, happy to change it as seen fit). 

https://github.com/user-attachments/assets/35d34acc-f9ea-40f0-98b7-4d9b6d2e9a9e

---

### AI Usage

Partially. I used codex to understand the codebase. Additionally, I verified any tab completion changes before pushing.

### DIFF
--- a/src/activities/util/KeyboardEntryActivity.cpp
+++ b/src/activities/util/KeyboardEntryActivity.cpp
@@ -92,8 +92,26 @@ void KeyboardEntryActivity::handleKeyPress() {
   // Handle special row (bottom row with shift, space, backspace, done)
   if (selectedRow == SPECIAL_ROW) {
     if (selectedCol >= SHIFT_COL && selectedCol < SPACE_COL) {
-      // Shift toggle
+      // Shift toggle (double-tap enables caps lock)
+      const unsigned long now = millis();
+      const bool isDoubleTap = (lastShiftTapMs != 0) && ((now - lastShiftTapMs) <= SHIFT_DOUBLE_TAP_MS);
+
+      if (capsLockActive) {
+        capsLockActive = false;
+        shiftActive = false;
+        lastShiftTapMs = 0;
+        return;
+      }
+
+      if (isDoubleTap) {
+        capsLockActive = true;
+        shiftActive = true;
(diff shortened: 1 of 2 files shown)


el (approved):

lukestein:
Tested on device and seems to work great. Thank you! 🎉

whyte-j:
Hey! I've got a slightly different implementation in PR https://github.com/crosspoint-reader/crosspoint-reader/pull/513, which cycles the three states rather than uses a double tap window. I can close it if we prefer this method

lukestein:
> Hey! I've got a slightly different implementation in PR #513, which cycles the three states rather than uses a double tap window. I can close it if we prefer this method

Ah thanks @whyte-j and sorry I hadn't seen that. I have no strong preference across implementations, noting that the UI in #513 is perhaps more discoverable than this one. Would love community input including (but not only) from @crosspoint-reader/firmware-maintainers.

whyte-j:
> > Hey! I've got a slightly different implementation in PR #513, which cycles the three states rather than uses a double tap window. I can close it if we prefer this method
> 
> 
> 
> Ah thanks @whyte-j and sorry I hadn't seen that. I have no strong preference across implementations, noting that the UI in #513 is perhaps more discoverable than this one. Would love community input including (but not only) from @crosspoint-reader/firmware-maintainers.

No worries! I also have no strong preference - the double tap behaviour is common enough that it should be in some people's mental models, although they may not expect it in this input scheme. I suppose the benefit of cycle here is that you would still discover it if you were expecting double tap, whereas the inverse may not be true.

Definitely think we should get some other voices as I would imagine I'm somewhat biased 😉

daveallie:
The cycle approach is more inline with my personally mental model and marginally more discoverable as you mentioned @whyte-j, going to progress #513 and close this one. Appreciate both of your efforts here! Thanks!

Issue #383, opened 2026-01-14
ToC/chapter navigation don't work if epub chapters arent aligned with spines


efenner:
### Affected Version

v.0.13.1

### Describe the Bug

When reading Moby Dick as downloaded from Project Gutenberg (EPUB, No Images), the whole table of contents appears correct when rendered in the reader and in the TOC Navigation interface reached by holding the select button, but the chapter progress indicator and navigation are both behaving incorrectly.

1. Long-pressing the next and previous page buttons to go to the next or previous chapter often jumps multiple chapters. I believe the behavior is consistent. Chapter 1 -> Chapter 4 -> Chapter 10 -> Chapter 17 (which actually jumps 3 pages from the end of chapter 16) -> Chapter 22 -> Chapter 29 -> Chapter 33 ...
2. The chapter progress indicator seems to show the pages remaining to the next chapter that would be jumped to rather than the actual next chapter. In the image below, you can see that the header for Chapter 2 is on screen, but it says that I am on page 21 of 91 in Chapter 1.
3. When trying to navigate through the TOC navigation interface, selecting a chapter other than those that are jumped to navigates to the nearest chapter that would be jumped to with a number lower than the selection (i.e. selecting chapter 9 jumps to chapter 4)  

Inspection of the EPUB shows that these chapter jumps seem to correspond with how the HTML files within the EPUB are separated, so Crosspoint seems to be assuming that there is one chapter per file. The chapters all have appropriate anchor links.

It would be great if this could be handled better on-device so that fewer EPUBs require conversion, especially since Project Gutenberg is such a reliable and recommended source for EPUBs. A Calibre conversion does seem to re-divide the file and fix this issue, but it is nice when that can be avoided.

Any chance a fix is possible?

### Steps to Reproduce

1. Download the [No Images EPUB of Moby Dick from Project Gutenberg](https://www.gutenberg.org/ebooks/2701.epub.noimages)
2. Upload to device (without any optimization)
3. Try to navigate the chapters as described

### Expected Behavior

Choosing a chapter in the TOC navigator should jump to the selected chapter, navigating to the next chapter using the buttons should navigate to the next chapter, and the progress should accurately show the current chapter and pages until the next chapter.

### Relevant Log Output/Screenshots

![Image](https://github.com/user-attachments/assets/535bf520-10ec-4ed6-92cc-9b5ade8b06a2)

andrewb1269:
[...]

PR #636, opened 2026-01-31
feat: Add Chapter Progress Bar status bar option

lukestein:
## Summary
This pull request introduces a new "Chapter Progress Bar" mode to the status bar, allowing users to track their progress within the current chapter in addition to the existing book-level progress options. It also unifies and increases the progress bar height for better visibility, and updates the settings UI to support the new mode.

Closes #636

**Status Bar/Progress Bar Enhancements:**

* Added a new `CHAPTER_PROGRESS_BAR` mode to `CrossPointSettings::STATUS_BAR_MODE`, and updated the settings UI to allow users to select this mode. [[1]](diffhunk://#diff-3af36372bb6233a83387a68091b5e0651c23585c7c0a95669ed893268ca709a8R34) [[2]](diffhunk://#diff-c55df9ec3ade843be000ba463cb75aa3df27dc34620a56c248fc4cc4e917b34bL22-R23)
* Implemented `drawChapterProgressBar` in `ScreenComponents` and integrated it into both EPUB and TXT reader activities, so the chapter progress bar is displayed when the new mode is selected. [[1]](diffhunk://#diff-be271778a942f7fab0d920acd73442512346ff811a4625c011275a7ca6be3a3eL51-R64) [[2]](diffhunk://#diff-dd410cab3a363d78172706d2ad6591f327e9b5b05f314db405db31a667af03faL16-R20) [[3]](diffhunk://#diff-82798dedbe135495e619d4aa27a4bef560c70c7663cf43148b67a26ddde45682R518-R525) [[4]](diffhunk://#diff-471ba9d9eb65b1a8451d41246db2aa695a42ea4ae4762163adfda4c20fec0950R563-R567)
* Updated logic in EPUB and TXT reader activities to show the correct progress bar, progress text, and battery indicator based on the selected status bar mode, including the new chapter progress bar mode. [[1]](diffhunk://#diff-82798dedbe135495e619d4aa27a4bef560c70c7663cf43148b67a26ddde45682R470-R481) [[2]](diffhunk://#diff-82798dedbe135495e619d4aa27a4bef560c70c7663cf43148b67a26ddde45682L490-R503) [[3]](diffhunk://#diff-471ba9d9eb65b1a8451d41246db2aa695a42ea4ae4762163adfda4c20fec0950R522-R533) [[4]](diffhunk://#diff-471ba9d9eb65b1a8451d41246db2aa695a42ea4ae4762163adfda4c20fec0950L539-R548)

**UI/Visual Tweaks:**

* Increased the progress bar height from 4 to 6 pixels for improved visibility, and refactored code to use the new constant. [[1]](diffhunk://#diff-dd410cab3a363d78172706d2ad6591f327e9b5b05f314db405db31a667af03faL16-R20) [[2]](diffhunk://#diff-82798dedbe135495e619d4aa27a4bef560c70c7663cf43148b67a26ddde45682L295-R295) [[3]](diffhunk://#diff-471ba9d9eb65b1a8451d41246db2aa695a42ea4ae4762163adfda4c20fec0950L177-R177)

These changes collectively provide users with more granular progress tracking options and a clearer visual indicator for reading progress.

## Additional Context

---

### AI Usage


el:
Can you share some screenshots to show how it looks

lukestein:
Bar shows position in current chapter (or in full txt file, since they don't have chapters)

Percentage display is progress in full book/file

Centered is chapter title (or text file name)

image

image

cc: @el

osteotek:
I haven't checked the code in detail, but would it work if instead of adding drawChapterProgressBar() function just reuse existing one and send chapter progress number to it? Functions seems to be identical

lukestein:
> I haven't checked the code in detail, but would it work if instead of adding drawChapterProgressBar() function just reuse existing one and send chapter progress number to it? Functions seems to be identical

Thanks @osteotek. Refactored and then tested on device with both epub and text files in both (book) progress and chapter progress settings and seems to work as expected (i.e., identical to before refactor).

I can imagine the value of the two progress bars looking different—for example, many readers offer tick marks on the book progress bar to show where the chapter breaks are—but in the absence of any plan to differentiate the two I think a single function as you suggested is def cleaner.

src/activities/reader/EpubReaderActivity.cpp
```diff
@@ -351,9 +351,10 @@ void EpubReaderActivity::renderScreen() {
   if (SETTINGS.statusBar != CrossPointSettings::STATUS_BAR_MODE::NONE) {
     // Add additional margin for status bar if progress bar is shown
     const bool showProgressBar = SETTINGS.statusBar == CrossPointSettings::STATUS_BAR_MODE::FULL_WITH_PROGRESS_BAR ||
-                                 SETTINGS.statusBar == CrossPointSettings::STATUS_BAR_MODE::ONLY_PROGRESS_BAR;
+                                 SETTINGS.statusBar == CrossPointSettings::STATUS_BAR_MODE::ONLY_PROGRESS_BAR ||
```
osteotek: maybe rename ONLY_PROGRESS_BAR to BOOK_PROGRESS_BAR for clarity

src/activities/reader/EpubReaderActivity.cpp
```diff
@@ -532,9 +539,16 @@ void EpubReaderActivity::renderStatusBar(const int orientedMarginRight, const in
                       progressStr);
   }
 
-  if (showProgressBar) {
+  if (showBookProgressBar) {
```
osteotek: maybe move book progress calculation here

src/activities/reader/EpubReaderActivity.cpp
lukestein: I wasn't sure if renaming old (existing) options would screw up migration of old settings. Or is the only thing saved the enum code so altering the name is ok?

osteotek: no, it shouldn't matter, enum names are not used in compiled code

efenner:
Looks like this would address idea https://github.com/crosspoint-reader/crosspoint-reader/discussions/593

lukestein:
> Looks like this would address idea #593

Thank you! I had thought there was an issue and somehow couldn't find it.

aaroncunliffe:
Works great on my device




However 2 small observations:
- There is documentation for this setting that seems to be missing a few options now [USER_GUIDE.md](https://github.com/crosspoint-reader/crosspoint-reader/blob/master/USER_GUIDE.md?plain=1#L112) 
- The other options that involve progress bar are `Full /w Progress Bar` and `Progress Bar`. Full meaning it includes the battery percentage, chapter name, and book percentage - like this new option does. So does it fit better as `Full /w Chapter Progress Bar`?

lukestein:
> * There is documentation for this setting that seems to be missing a few options now [USER_GUIDE.md](https://github.com/crosspoint-reader/crosspoint-reader/blob/master/USER_GUIDE.md?plain=1#L112)
> * The other options that involve progress bar are `Full /w Progress Bar` and `Progress Bar`. Full meaning it includes the battery percentage, chapter name, and book percentage - like this new option does. So does it fit better as `Full /w Chapter Progress Bar`?

Thanks @aaroncunliffe!

I re-named the options (and added to the User Guide in this PR) as follows, but I'm not attached to the names:

> **Status Bar**: Configure the status bar displayed while reading:
> - "None" - No status bar
> - "No Progress" - Show status bar without reading progress
> - "Full w/ Percentage" - Show status bar with book progress (as percentage)
> - "Full w/ Book Bar" - Show status bar with book progress (as bar)
> - "Book Bar Only" - Show book progress (as bar)
> - "Full w/ Chapter Bar" - Show status bar with chapter progress (as bar)

osteotek (approved):

osteotek (approved):

aaroncunliffe:
The latest changes work great on device, I'll definitely be using this as my default once it's in, nice change @lukestein.

One behaviour note: on some books it looks like the bar tracks spine item progress rather than full chapter progress. With an EPUB of The Martian, I can finish a subchapter and the bar resets while still in that same parent chapter.

I actually prefer this behaviour since the full chapter title still shows, but flagging it in case it’s unexpected.

lukestein:
> One behaviour note: on some books it looks like the bar tracks spine item progress rather than full chapter progress. With an EPUB of The Martian, I can finish a subchapter and the bar resets while still in that same parent chapter.
> 
> I actually prefer this behaviour since the full chapter title still shows, but flagging it in case it’s unexpected.

Thanks @aaroncunliffe. I just tried to use the existing values (without really considering a spine/chapter distinction) used to calculate the chapter page display in the old, existing "full" status bar:

```
    if (showProgressPercentage) {
      snprintf(progressStr, sizeof(progressStr), "%d/%d  %.0f%%", section->currentPage + 1, section->pageCount,
               bookProgress);
    } else if (showBookPercentage) {
      snprintf(progressStr, sizeof(progressStr), "%.0f%%", bookProgress);
    } else {
      snprintf(progressStr, sizeof(progressStr), "%d/%d", section->currentPage + 1, section->pageCount);
    }
```

so I did

```
    const float chapterProgress =
        (section->pageCount > 0) ? (static_cast<float>(section->currentPage + 1) / section->pageCount) * 100 : 0;
```

I'm happy to take a suggestion (from anyone) if there's a better way to do this.

aaroncunliffe:
I'm not familiar enough with the underlying EPUB construction just yet to suggest a fix.

From your snippet above I've realised `Full w/ Percentage` which falls into the else statement also tracks this the same way, so I don't think it's too much of a problem right now. It might only be a small percentage of books structured this way.

I'd suggest leaving it as it is right now, but I am *slightly* biased as I prefer it this way 😄

lukestein:
> I'm not familiar enough with the underlying EPUB construction just yet to suggest a fix.

@aaroncunliffe, I suspect we may have encountered what's behind #383 …

lukestein:
Testing on device (successfully) with both default and Lyra UI

PR #637, opened 2026-01-31
fix: Hide button hints in landscape CW mode

nscheung:
## Summary

* This change hides the button hints from overlapping chapter titles when in landscape CW mode.

Before
Before
After
After

## Additional Context

* I initially considered implementing an offset fix, but with potential UI changes on the horizon, hiding the button hints appears to be the simplest solution for now. 

---

### AI Usage

Did you use AI tools to help write this code? < Partially >


daveallie (approved):

daveallie:
Agree that worth just hiding for now, can re-evaluate with newer UI

PR #645, opened 2026-02-01
feat: Option for word spacing

jonasdiemer:
## Summary

I noticed words sometimes get too close for my liking, so I've added an option for word spacing.

It defaults to 100%, so behavior should be as before. 

Though I am not sure there is another issue lurking with the narrow spacing... @osteotek, maybe you have an idea (you should be quite familiar with ParseText.cpp).

I also fixed an issue in `CategorySettingsActivity.cpp` where an `int8_t` was used instead of `uint8_t`, causing an overflow issue with my particular setting (which goes >127).

---

### AI Usage

### DIFF
--- a/src/activities/settings/SettingsActivity.cpp
+++ b/src/activities/settings/SettingsActivity.cpp
@@ -29,7 +29,7 @@ const SettingInfo readerSettings[readerSettingsCount] = {
     SettingInfo::Enum("Font Family", &CrossPointSettings::fontFamily, {"Bookerly", "Noto Sans", "Open Dyslexic"}),
     SettingInfo::Enum("Font Size", &CrossPointSettings::fontSize, {"Small", "Medium", "Large", "X Large"}),
     SettingInfo::Enum("Line Spacing", &CrossPointSettings::lineSpacing, {"Tight", "Normal", "Wide"}),
-    SettingInfo::Value("Word Spacing %", &CrossPointSettings::wordSpacing, {100, 250, 25}),
+    SettingInfo::Value("Word Spacing %", &CrossPointSettings::wordSpacing, {75, 250, 25}),
     SettingInfo::Value("Screen Margin", &CrossPointSettings::screenMargin, {5, 40, 5}),
     SettingInfo::Enum("Paragraph Alignment", &CrossPointSettings::paragraphAlignment,
                       {"Justify", "Left", "Center", "Right"}),
--- a/src/activities/settings/SettingsActivity.cpp
+++ b/src/activities/settings/SettingsActivity.cpp
@@ -24,11 +24,12 @@ const SettingInfo displaySettings[displaySettingsCount] = {
     SettingInfo::Enum("Refresh Frequency", &CrossPointSettings::refreshFrequency,
                       {"1 page", "5 pages", "10 pages", "15 pages", "30 pages"})};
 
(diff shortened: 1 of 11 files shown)


src/activities/settings/SettingsActivity.cpp
```diff
@@ -24,11 +24,12 @@ const SettingInfo displaySettings[displaySettingsCount] = {
     SettingInfo::Enum("Refresh Frequency", &CrossPointSettings::refreshFrequency,
                       {"1 page", "5 pages", "10 pages", "15 pages", "30 pages"})};
 
-constexpr int readerSettingsCount = 9;
+constexpr int readerSettingsCount = 10;
 const SettingInfo readerSettings[readerSettingsCount] = {
     SettingInfo::Enum("Font Family", &CrossPointSettings::fontFamily, {"Bookerly", "Noto Sans", "Open Dyslexic"}),
     SettingInfo::Enum("Font Size", &CrossPointSettings::fontSize, {"Small", "Medium", "Large", "X Large"}),
     SettingInfo::Enum("Line Spacing", &CrossPointSettings::lineSpacing, {"Tight", "Normal", "Wide"}),
+    SettingInfo::Value("Word Spacing %", &CrossPointSettings::wordSpacing, {75, 250, 25}),
```
osteotek, line 32: Doesn't seem to have default option of 100?

jonasdiemer: The `ValueRange` takes min, max, step, so default of `100` is included (`min + 1*step`).

osteotek: ah, thanks, I missed that

osteotek:
@jonasdiemer spaceWidth on line 71 is the only variable that controls spacing between words - it's the one you change with new option. Both computeLineBreaks() and computeHyphenatedLineBreaks() build lines using spaceWidth between words. That should guarantee that spacing variable on line 350 used for justified text `(spacing = spareSpace / (lineWordCount - 1))` is always `>= spaceWidth`. For last lines or non‑justified styles, spacing stays at spaceWidth.

jonasdiemer:
> @jonasdiemer spaceWidth on line 71 is the only variable that controls spacing between words - it's the one you change with new option. Both computeLineBreaks() and computeHyphenatedLineBreaks() build lines using spaceWidth between words. That should guarantee that spacing variable on line 350 used for justified text `(spacing = spareSpace / (lineWordCount - 1))` is always `>= spaceWidth`. For last lines or non‑justified styles, spacing stays at spaceWidth.

Thanks for checking, @osteotek. I was thinking we might be "off by 1" somewhere (either with space or maybe character width is 1px too wide?), but couldn't spot it either.

jonasdiemer:
I think I can confirm that there is no "other issue". Below is a screenshot from Kindle app, overlaid with a photo from CrossPoint (left aligned). All characters line up perfectly (considering that the photo likely has some perspective error):

image

osteotek (approved):

znelson:
@jonasdiemer Is this something we can still get in?

Flavioccastro:
This feature was implemented?

Uri-Tauber:
@Flavioccastro Nope.
If @jonasdiemer will solve conflicts, we can get this in soon.

serialx:
I've updated this PR in #3528 and it is merged. Thanks for your contribution!

PR #651, opened 2026-02-01
feat:  button layout support - right, left, confirm, back

### LINKED ISSUE #620: Button Layout -> Right, Left, Confirm, Back
Just flashed crosspoint today, looks and feels so buttery smooth compared to stock!

I got pretty used to using the stock firmware in inverted mode (upside down, since that's how it's mounted to my phone), and I notice when trying to configure the front button layout there is no "Right, Left, Confirm, Back" combination. I know it sounds confusing, but when using it inverted this feels like the natural key combo because the far right key is then right, then left next to it, and then confirm/back also feels the right way around.

I imagine this wouldn't be a huge change, so happy to submit a PR!

aaroncunliffe:
## Summary

Implements new front button layouts discussed in issue: https://github.com/crosspoint-reader/crosspoint-reader/issues/620

resolves #620 

New Layout:
Right, Left, Confirm, Back

Works best when Inverted reader mode is on.

## Additional Context
Reformatted enum at [SettingsActivity.cpp#L43](https://github.com/aaroncunliffe/crosspoint-reader/blob/feature/additional-front-button-layout/src/activities/settings/SettingsActivity.cpp#L43), but made sure to run clang-format

---

### AI Usage

Did you use AI tools to help write this code? _< No >_


benkaiser:
Thanks @aaroncunliffe for making the change!

lukestein:
Tested on device and seems to work as expected (although changing this setting remains confusing since toggle button jumps around with each switch)

aaroncunliffe:
> although changing this setting remains confusing since toggle button jumps around with each switch

Looks like this is a common sentiment, this issue was also raised earlier with some good discussion ongoing - https://github.com/crosspoint-reader/crosspoint-reader/issues/654

osteotek:
@aaroncunliffe would this work for your case? https://github.com/crosspoint-reader/crosspoint-reader/pull/664

aaroncunliffe:
> @aaroncunliffe would this work for your case? #664

@osteotek It wasn't for my needs, this was raised in #620, I was just looking for a nice easy first issue to jump into the codebase. I'm happy to close if #664 was the direction we were wanting to go down, it would close a few open issues/PRs

aaroncunliffe:
Closing in favour of #664

PR #317, opened 2026-01-11
feature: adding a default wifi option

DJPoulter:
## Summary

* **What is the goal of this PR?** (e.g., Fixes a bug in the user authentication module, Implements the new feature for 
  file uploading.)

Calibre library currently uses the first wifi in your wifi file. This allows you to select which WiFi to use. This also allows future features that will be able to use WiFi like clock and syncing. 

* **What changes are included?**

## Additional Context

* Add any other information that might be helpful for the reviewer (e.g., performance implications, potential risks, specific areas to focus on).

PR #655, opened 2026-02-01
feat: Save and use the last connected wifi

el:
## Summary

Saves the last connected wifi and uses that first when using Koreader Sync, OPDS browser or File Transfer. this way, the user do not have to keep selecting the wifi again and again.

- Save the last connected wifi
- Create a new settings item under System as "Network" that shows the network list
- Add the option to Forget a network 
- This makes using these features faster since most of the time we connect to the same WiFi and bypassing the network search and select is a huge plus.

IMG_6463

---

### AI Usage

Did you use AI tools to help write this code? _**  YES

### DIFF
--- a/src/activities/reader/EpubReaderActivity.cpp
+++ b/src/activities/reader/EpubReaderActivity.cpp
@@ -204,15 +204,15 @@ void EpubReaderActivity::loop() {
     xSemaphoreGive(renderingMutex);
   }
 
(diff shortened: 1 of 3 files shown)


DJPoulter:
This is similar to #317 , although I have left that one at the moment until it gets reviewed or traction.

el:
Found some bugs

PR #656, opened 2026-02-01
feat: add HalStorage

ngxson:
## Summary

Continue my changes to introduce the HAL infrastructure from https://github.com/crosspoint-reader/crosspoint-reader/pull/522

This PR touches quite a lot of files, but most of them are just name changing. It should not have any impacts to the end behavior.

## Additional Context

My plan is to firstly add this small shim layer, which sounds useless at first, but then I'll implement an emulated driver which can be helpful for testing and for development.

Currently, on my fork, I'm using a FS driver that allow "mounting" a local directory from my computer to the device, much like the `-v` mount option on docker. This allows me to quickly reset `.crosspoint` directory if anything goes wrong. I plan to upstream this feature when this PR get merged.

---

### AI Usage

### DIFF
--- a/lib/Epub/Epub/ParsedText.cpp
+++ b/lib/Epub/Epub/ParsedText.cpp
@@ -19,38 +19,6 @@ namespace {
 constexpr char SOFT_HYPHEN_UTF8[] = "\xC2\xAD";
 constexpr size_t SOFT_HYPHEN_BYTES = 2;
 
-// Known attaching punctuation (including UTF-8 sequences)
-const std::vector<std::string> punctuation = {
-    ".",
-    ",",
-    "!",
-    "?",
-    ";",
-    ":",
-    "\"",
-    "'",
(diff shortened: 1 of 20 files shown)


daveallie (commented):
Keen, may need to do another sweep with the newer changes. Should we be using `Storage` in the `lib` directory instead of `SdMan`?

ngxson:
> Keen, may need to do another sweep with the newer changes.

Done, rebased to latest `master`

> Should we be using `Storage` in the `lib` directory instead of `SdMan`?

Yes, I just pushed a small hack to forbidden the usage of `SdMan`: https://github.com/crosspoint-reader/crosspoint-reader/pull/656/commits/8354d0cfb8c669ec7f91d159c757bc2841c92629

Not sure it it's the best way though, I'm open for discussion

ngxson:
@daveallie Could you help merging this in a short time frame to avoid having more conflicts? Thanks!

PR #640, opened 2026-02-01
docs: Add small SCOPE.md and GOVERNANCE.md documents

daveallie:
## Summary

* Additional documentation for CrossPoint in the way of a SCOPE.md and GOVERNANCE.md document.
* These are intentionally pretty lightweight for now

---

### AI Usage

While CrossPoint doesn't have restrictions on AI tools in contributing, please be transparent about their usage as it 
helps set the right context for reviewers.

Did you use AI tools to help write this code? Partially

PR #659, opened 2026-02-02
feat: Add inverse display setting

obra:
Adds an "Inverse Display" toggle under Display settings that renders white text/UI on a black background across the entire UI — home screen, settings, file browser, readers, and all other activities.

## Approach

Rather than threading `!inverse` through every draw call and activity, this implements auto-inversion at the GfxRenderer level:

- **`drawPixel()`** flips pixel state when `inverseDisplay && renderMode == BW`
- **`clearScreen()`** flips the fill color under the same condition
- **`setInverseDisplay()`** on GfxRenderer triggers a full refresh when the value changes, so the transition between modes is clean

This means all activities get inverse display for free with zero per-activity changes. The setting is synced from `SETTINGS.inverseDisplay` every loop iteration in `main.cpp`, and also immediately on toggle in `SettingsActivity` to avoid a race with the render task.

### Grayscale anti-aliasing

The `renderMode == BW` guard ensures grayscale buffer setup (`clearScreen(0x00)` and `drawPixel` calls during `GRAYSCALE_LSB`/`GRAYSCALE_MSB` passes) is unaffected by auto-inversion. Reader activities reorder `setRenderMode()` before `clearScreen(0x00)` so the mode is already non-BW when the buffer is cleared.

For the grayscale LUT row selection in `renderChar`, the conditions now use `!inverseDisplay` directly (instead of relying on the caller's `pixelState` parameter):

- **GRAYSCALE_MSB**: `(bmpVal == 1 || (!inverseDisplay && bmpVal == 2))` — in inverse mode, only bmpVal=1 gets MSB marked
- **GRAYSCALE_LSB**: `bmpVal == (inverseDisplay ? 2 : 1)` — swaps which gray level gets LSB marked

This reassigns LUT rows so fringe pixels use waveforms that push FROM the physically-white starting state TOWARD black, producing visible gray transitions on the inverted background. (In normal mode, fringe pixels start black and get pulled toward white — the inverse needs the opposite direction.)

### Why not hardware inversion?

The SSD1677's command 0x21 bit 3 can invert BW source output, but this breaks differential fast refresh: the BW RAM gets inverted while RED RAM (previous frame) doesn't, so every pixel gets the wrong LUT transition applied, causing severe ghosting. Software-level inversion avoids this entirely.

## Files changed

- **GfxRenderer.h/.cpp** — `inverseDisplay` member, `setInverseDisplay()`, auto-inversion in `drawPixel`/`clearScreen`/`displayBuffer`, grayscale conditions in `renderChar`/`drawTextRotated90CW`
- **CrossPointSettings.h/.cpp** — `inverseDisplay` field with serialization
- **SettingsList.h** — Toggle in Display category
- **EpubReaderActivity.cpp / TxtReaderActivity.cpp** — Reorder `setRenderMode()` before `clearScreen(0x00)` in grayscale passes
- **XtcReaderActivity.cpp** — Add `setRenderMode()` calls around grayscale passes for consistency
- **SettingsActivity.cpp** — Sync `renderer.setInverseDisplay()` on toggle
- **main.cpp** — Sync `renderer.setInverseDisplay()` in main loop

---

### AI Usage

Designed and tested by me on a physical device. Implementation by Claude Opus 4.6 using Claude Code.

### DIFF
--- a/lib/GfxRenderer/GfxRenderer.cpp
+++ b/lib/GfxRenderer/GfxRenderer.cpp
@@ -49,6 +49,10 @@ static inline void rotateCoordinates(const GfxRenderer::Orientation orientation,
 // IMPORTANT: This function is in critical rendering path and is called for every pixel. Please keep it as simple and
 // efficient as possible.
 void GfxRenderer::drawPixel(const int x, const int y, const bool state) const {
+  // In BW mode with inverse display, flip the pixel state so all drawing
+  // appears inverted without callers needing to know about it.
+  const bool effectiveState = (inverseDisplay && renderMode == BW) ? !state : state;
+
   int phyX = 0;
   int phyY = 0;
 
@@ -65,7 +69,7 @@ void GfxRenderer::drawPixel(const int x, const int y, const bool state) const {
   const uint16_t byteIndex = phyY * HalDisplay::DISPLAY_WIDTH_BYTES + (phyX / 8);
   const uint8_t bitPosition = 7 - (phyX % 8);  // MSB first
 
-  if (state) {
+  if (effectiveState) {
     frameBuffer[byteIndex] &= ~(1 << bitPosition);  // Clear bit
   } else {
     frameBuffer[byteIndex] |= 1 << bitPosition;  // Set bit
@@ -644,7 +648,8 @@ static unsigned long start_ms = 0;
 
 void GfxRenderer::clearScreen(const uint8_t color) const {
   start_ms = millis();
-  display.clearScreen(color);
+  const uint8_t effectiveColor = (inverseDisplay && renderMode == BW) ? ~color : color;
+  display.clearScreen(effectiveColor);
 }
 
 void GfxRenderer::invertScreen() const {
@@ -656,7 +661,12 @@ void GfxRenderer::invertScreen() const {
 void GfxRenderer::displayBuffer(const HalDisplay::RefreshMode refreshMode) const {
   auto elapsed = millis() - start_ms;
   Serial.printf("[%lu] [GFX] Time = %lu ms from clearScreen to displayBuffer\n", millis(), elapsed);
-  display.displayBuffer(refreshMode, fadingFix);
+  auto effectiveMode = refreshMode;
+  if (fullRefreshPending) {
+    effectiveMode = HalDisplay::FULL_REFRESH;
+    fullRefreshPending = false;
+  }
+  display.displayBuffer(effectiveMode, fadingFix);
 }
 
 std::string GfxRenderer::truncatedText(const int fontId, const char* text, const int maxWidth,
@@ -815,11 +825,11 @@ void GfxRenderer::drawTextRotated90CW(const int fontId, const int x, const int y
             const uint8_t bit_index = (3 - pixelPosition % 4) * 2;
             const uint8_t bmpVal = 3 - (byte >> bit_index) & 0x3;
 
-            if (renderMode == BW && bmpVal < 3) {
+            if (renderMode == BW && (inverseDisplay ? bmpVal == 0 : bmpVal < 3)) {
               drawPixel(screenX, screenY, black);
             } else if (renderMode == GRAYSCALE_MSB && (bmpVal == 1 || bmpVal == 2)) {
               drawPixel(screenX, screenY, false);
-            } else if (renderMode == GRAYSCALE_LSB && bmpVal == 1) {
+            } else if (renderMode == GRAYSCALE_LSB && bmpVal == (inverseDisplay ? 2 : 1)) {
               drawPixel(screenX, screenY, false);
             }
           } else {
@@ -982,15 +992,11 @@ void GfxRenderer::renderChar(const EpdFontFamily& fontFamily, const uint32_t cp,
           // 0 -> black, 1 -> dark grey, 2 -> light grey, 3 -> white
           const uint8_t bmpVal = 3 - (byte >> bit_index) & 0x3;
 
-          if (renderMode == BW && bmpVal < 3) {
-            // Black (also paints over the grays in BW mode)
+          if (renderMode == BW && (inverseDisplay ? bmpVal == 0 : bmpVal < 3)) {
             drawPixel(screenX, screenY, pixelState);
           } else if (renderMode == GRAYSCALE_MSB && (bmpVal == 1 || bmpVal == 2)) {
-            // Light gray (also mark the MSB if it's going to be a dark gray too)
-            // We have to flag pixels in reverse for the gray buffers, as 0 leave alone, 1 update
             drawPixel(screenX, screenY, false);
-          } else if (renderMode == GRAYSCALE_LSB && bmpVal == 1) {
-            // Dark gray
+          } else if (renderMode == GRAYSCALE_LSB && bmpVal == (inverseDisplay ? 2 : 1)) {
             drawPixel(screenX, screenY, false);
           }
         } else {
--- a/lib/GfxRenderer/GfxRenderer.h
+++ b/lib/GfxRenderer/GfxRenderer.h
@@ -33,6 +33,8 @@ class GfxRenderer {
   RenderMode renderMode;
   Orientation orientation;
   bool fadingFix;
+  bool inverseDisplay;
+  mutable bool fullRefreshPending = false;
   uint8_t* frameBuffer = nullptr;
   uint8_t* bwBufferChunks[BW_BUFFER_NUM_CHUNKS] = {nullptr};
   std::map<int, EpdFontFamily> fontMap;
@@ -46,7 +48,7 @@ class GfxRenderer {
 
  public:
   explicit GfxRenderer(HalDisplay& halDisplay)
-      : display(halDisplay), renderMode(BW), orientation(Portrait), fadingFix(false) {}
+      : display(halDisplay), renderMode(BW), orientation(Portrait), fadingFix(false), inverseDisplay(false) {}
   ~GfxRenderer() { freeBwBufferChunks(); }
 
   static constexpr int VIEWABLE_MARGIN_TOP = 9;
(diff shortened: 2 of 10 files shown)


nkocher:
Hey, been running this on my build and really liking it! Very nice work.

One thing I ran into — with inverse display on, the status bar elements still render black (battery icon, percentage text, and the progress bar at the bottom). renderStatusBar() doesn't get the inverse flag, so the ScreenComponents draw calls (drawBattery, drawBookProgressBar, etc.) use their default black = true and always draw black regardless of the setting.

I fixed it by threading the inverse parameter through renderStatusBar() into the ScreenComponents methods, then passing !inverse to all the draw calls — same pattern you use for the content text:

     // Before (always black)
     renderer.drawText(SMALL_FONT_ID, left + 20, top, percentageText.c_str());
     renderer.fillRect(x + 2, y + 2, filledWidth, batteryHeight - 4);

     // After (respects inverse)
     renderer.drawText(SMALL_FONT_ID, left + 20, top, percentageText.c_str(), !inverse);
     renderer.fillRect(x + 2, y + 2, filledWidth, batteryHeight - 4, !inverse);

I've got the full fix ready — I can open a PR against your branch if you'd like, or feel free to incorporate it directly. The changes touch ScreenComponents.cpp/.h, EpubReaderActivity.cpp/.h, and TxtReaderActivity.cpp/.h.

 (AI disclosure: Claude Code + superpowers :))

obra:
If it sounds like the maintainers want this, I'm happy to get your stuff merged in. Also, it makes me happy every time I see a Superpowers user in the wild :)

lukestein:
> If it sounds like the maintainers want this, I'm happy to get your stuff merged in.

I'm just one maintainer, but I think getting a good dark mode into crosspoint is important in our aim to **facilitate focused reading** (#640).

Some updates to status bar hopefully coming soon #636

obra:
@nkocher - I'm having my claude update based on your report.

@lukestein I'm very happy to revise after #636  merges if that's landing first.

obra:
@lukestein updated!

lib/GfxRenderer/GfxRenderer.cpp
daveallie: This check would need to be adjusted. Currently there's not going to be any anti-aliasing happening. It's pretty hard to follow, but the pixels need to be pushed to black, before they can be pulled back towards white to produce a lighter or darker grey, that's just the way the LUTs are currently configured.

When not inverted, we color all the `bmpVal < 3` pixels (black and both greys) black before coming back for the other passes. When inverted now, these pixels are made **white** instead on the first pass so pulling them further white during the grayscale pass is having no effect.

daveallie (changes requested):

obra:
@daveallie My buddy Claude is proposing a solution involving hacking up the SDK to add a single method to do whole-display inversion just before refresh. If this feels plausible and reasonable to you, I'm happy to build it out and test it locally before submitting, but I wanted to check in since it requires an SDK change.  The upside, is that it looks to be a much simpler change that reduces code complexity and gets the whole UI inverted.  (We might need to play games to do something smarter with images down the road, but I'm not sure.)

⏺ Here's the synthesis from all four investigations:     

  We need to touch the SDK, but it's a 3-line change and the path is clear.                                       
   
  Key findings:                                                                                                   
                                                                                                                
  1. Dave Allie maintains both the SDK and crosspoint-reader. The SDK is at open-x4-epaper/community-sdk, has zero
   local modifications, and Dave authored most commits in both repos. He's also the reviewer requesting changes on
   our PR. Proposing an SDK change is natural here.                                                             
  2. There's no way around the SDK change. sendCommand/sendData are private, and even if accessible,    
  refreshDisplay() immediately overwrites command 0x21. Every display path (displayBuffer, displayGrayBuffer,
  displayWindow) converges at refreshDisplay().
  3. The SDK change follows an existing pattern. The class already manages state booleans: isScreenOn,
  customLutActive, inGrayscaleMode. The change is:
    - Add member: bool invertDisplay = false;
    - Add setter: void setInvertDisplay(bool enabled);
    - In refreshDisplay(): if (invertDisplay) ctrl1 |= 0x08;
  4. Hardware inversion is confirmed in the family. Espressif's official SSD1681 BSP defines
  SSD1681_PARAM_COLOR_BW_INVERSE_BIT (1<<3) for command 0x21. The SSD1677 shares the same command set.

  Net result: With this one SDK change, we can rip out almost all the software-level inverse threading from the
  PR. The reader code just calls display.setInvertDisplay(SETTINGS.inverseDisplay) and renders everything
  normally. No !inverse parameter passing, no grayscale LSB swaps, no status bar changes, no theme changes.

daveallie:
The display driver is actually the SSD1677, not the SSD1681, but it does infact seem possible to have the driver itself invert the display RAM ([datasheet here](https://cursedhardware.github.io/epd-driver-ic/SSD1677.pdf)):
CleanShot 2026-02-10 at 19 41 16

However there's still the issue of how the LUTs are setup, even with the inverted reading of the RAM, the application logic would still need to adjust how it treats the pixels which would become grey. In the current setup, they are painted black and then in a second display update pulled closer to white. We don't have LUTs setup the push the pixels closer to black, so with the inverted logic you'd end up painting the grey pixels white instead of black in the first pass and then pushing them further white (or more likely just skipping over them like is currently happening).

It'll still need to be application logic for now unfortunately. I am keen to test out Larry's (bitbank2) update LUTs which should allow us to simplify the display loop and do a single display update cycle for blacks and greys at once, but it's still unproven.

obra:
@daveallie Completely rewrote this based on your feedback. Here's what changed:

**Hardware inversion (0x21 bit 3):** Tried it locally — confirmed it causes severe ghosting because RED RAM retains the non-inverted previous frame, so differential fast refresh computes wrong LUT transitions for every pixel. Ruled out.

**New approach — auto-inversion in GfxRenderer:** Instead of threading `!inverse` through every draw call (which was fragile and missed most of the UI), `drawPixel()` and `clearScreen()` now auto-invert when `inverseDisplay && renderMode == BW`. This gives the entire UI inverse display for free — home screen, settings, file browser, everything — with zero per-activity changes.

**Grayscale fix (your review comment):** The grayscale conditions in `renderChar` now use `!inverseDisplay` directly instead of the caller's `pixelState`. In inverse mode, `GRAYSCALE_MSB` only marks bmpVal=1 (not bmpVal=2), and `GRAYSCALE_LSB` marks bmpVal=2 (not bmpVal=1). This reassigns LUT rows so fringe pixels get waveforms that push from white toward black — the correct direction for an inverted background. Tested on device, anti-aliasing looks correct.

**Diff is much simpler now:** 10 files, +47/-16 lines. No changes to Page, TextBlock, BaseTheme, or LyraTheme. The Page/TextBlock/Theme layers don't need a `bool black` parameter at all since GfxRenderer handles everything.

obra:
Real human replying here - Yep. I tested out Claude's version based on the sdk-level whole display reversing and it was kind of a disaster. (There was more data out there about this functionality in the '81 than the '77. I was impressed that it made the connection, but the whole thing was crawling with bugs, mostly related to refresh.)

This new implementation is app logic, but more centralized, so we get full-UI inverse/dark mode for 'free' with less needing to thread things together. For debugging I almost want to compile in a much, much larger font to make anti-aliasing problems obvious without magnification.

I'm not 100% convinced that anti-aliasing is behaving right, so I'm converting this over to a draft for the moment.

obra:
Okay, I've tested this on actual hardware and zoomed in in dark mode, and it sure looks like anti-aliasing is behaving correctly for me. I'd love yet if somebody else would take it for a test drive and see if it seems sane.

lib/GfxRenderer/GfxRenderer.cpp
```diff
@@ -815,11 +825,11 @@ void GfxRenderer::drawTextRotated90CW(const int fontId, const int x, const int y
             const uint8_t bit_index = (3 - pixelPosition % 4) * 2;
             const uint8_t bmpVal = 3 - (byte >> bit_index) & 0x3;
 
-            if (renderMode == BW && bmpVal < 3) {
+            if (renderMode == BW && (inverseDisplay ? bmpVal == 0 : bmpVal < 3)) {
               drawPixel(screenX, screenY, black);
```
daveallie, line 829: There is another small consideration here. When text anti-aliasing is off, then in the non-inverted experience, the grays are all left as solid blacks (the font color), but now the grays are still left as blacks (background color), making the lettering look quite thin.

It's going to be a pain to drill the fact that anti-aliasing is off down into here. I have half a mind to just declare that inverse + anti-aliasing just won't be thick and leave it at that. Give it a spin on your device and lmk what you think. To me, it doesn't look ideal, but I also just leave AA on the whole time.

obra: I'm traveling for the next couple of days, but I'm happy to test it out then. I haven't done deep into the architecture enough to know whether this is an unreasonable question: is there not some global place to hang a property like antialiasing off of so it wouldn't need to be drilled? or is there a good reason that you want antialiasing to be an option in the readers, but not in the rest of the UI?

daveallie: AA currently makes rendering slower as we need to:
1. Render screen to local buffer (blacks, light greys, dark greys)
2. Update BW RAM
3. Trigger Black only render
4. Render screen to local buffer (dark greys only)
5. Update BW RAM
6. Render screen to local buffer (light greys only)
7. Update RED RAM
8. Update LUT
9. Trigger greyscale render
10. Revert RED RAM to the correct value

Everything 4 down is AA only code. It just simplifies interface code and speeds up the experience by skipping the greys. It is available from `SETTINGS` and perhaps we could relax it but to date, I've tried to keep lib code away from depending on the src code directly. Which in this case would encourage drilling from the src -> lib entry point which would just be as part of the `page->render` call I believe.

I would love to just leave it always enabled in the reader, but some users find the experience of the two pass rendering bad.

obra: that makes a ton of sense. Is there a place that you want a doc patch to capture that knowledge for future devs?

daveallie: Might make sense somewhere like this: https://github.com/crosspoint-reader/crosspoint-reader/pull/894/changes#diff-ea8ccb967c4933aadf651d159b3ecd9f176bdb4463954932fc474e5e8bcfd9c2 or in a new `rendering.md` file, there's a fair amount of tribal knowledge around that piece which could do with some capturing.

znelson:
Closing due to inactivity. I think there are newer "dark mode" PRs in the pipeline.

obra:
Apologies for dropping the ball on this one. If I can be helpful, let me know.

znelson:
@obra You didn't at all, please don't take it that way! 

The project is going through a phase of growth, and some things didn't make it early on. If you're still engaged and interested, we're happy to have you.

PR #660, opened 2026-02-02
feat: Add f-ligature support (ff, fi, fl, ffi, ffl)

obra:
Add ligature codepoints (U+FB00–FB06) to font conversion so all fonts include ligature glyphs. Implement utf8NextCodepointWithLigatures() that does lookahead substitution for f-ligature sequences. Use it in text rendering and measurement so ligatures render correctly and word-wrapping accounts for the ligature glyph widths.

## Summary

This is a first pass implementation of the most common ligatures, ff, fi, fl, ffi, ffl for the readers. 

## Additional Context

The feature takes a real amount of flash, (primarily because of the fact that it adds glyphs to all fonts at all sizes) and I 100% understand if this is not a direction that Crosspoint wants to go. But it sure looks pretty when I'm reading books.

---

### AI Usage

This was implemented by Claude Opus 4.5 using Claude Code 2.1.25. I manually tested it on an X4.


osteotek:
that's awesome, I've been wanting to add ligatures for a long time. I see that it results in 95.9% of Flash used 🤔 A bit too close to the maximum for my liking. I've seen device become unstable at around 98% mark. Size improvements changes should be merged into the master branch soon, maybe we can add this PR then, and add other ligatures as well

obra:
@osteotek That makes a ton of sense. Quick question on the size improvement front. I noticed that the build is currently without LTO. Enabling LTO absolutely shaves a couple percent off flash usage. Is it known to break crosspoint?

osteotek:
@obra I'm not sure if anyone tried that. If you can test it that would be great

obra:
> @obra I'm not sure if anyone tried that. If you can test it that would be great

Oh! I did. I'll submit a PR. It will want more folks than just me testing it, though.

osteotek:
Separate PR with ligatures support has been merged

PR #670, opened 2026-02-02
fix: Increase network SSID display length

aaroncunliffe:
## Rationale 

I have 2 wifi access points with almost identical names, just one has `_EXT` at the end of it. With the current display limit of 13 characters before adding ellipsis, I can't tell which is which.

Before device screenshot with masked SSIDs:

## Summary

Adjusted displayed length from 13 characters to 30 in the Wifi selection screen - I've left some space for potential proportional font changes in the future

After image with masked SSIDs:

---

### AI Usage


lukestein:
I don't have an easy way to set up a 30+ character SSID, but I'm running this and all still seems to work fine with my (short) SSIDs.

PR #671, opened 2026-02-02
add unique hostname support

reukiodo:
## Summary

* Implements dynamic hostname (based on last 6 of MAC address) to allow multiple XTeink on the same network
* src/activities/network/CalibreConnectActivity.cpp
* src/activities/network/CrossPointWebServerActivity.cpp

## Additional Context

  specific areas to focus on).
Slightly more memory used to make the hostname string

---

### AI Usage

Did you use AI tools to help write this code? PARTIALLY to help understand the code base and where to make all the necessary changes to ensure nothing was left out.

### DIFF
--- a/USER_GUIDE.md
+++ b/USER_GUIDE.md
@@ -230,6 +230,7 @@ Accessible by pressing **Confirm** while inside a book.
 Please note that this firmware is currently in active development. The following features are **not yet supported** but are planned for future updates:
 
 * **Images:** Embedded images in e-books will not render.
+* **Cover Images:** Large cover images embedded into EPUB require several seconds (~10s for ~2000 pixel tall image) to convert for sleep screen and home screen thumbnail. Consider optimizing the EPUB with e.g. https://github.com/bigbag/epub-to-xtc-converter to speed this up.
 
 ---
 
@@ -242,3 +243,5 @@ pio device monitor
 ```
 
 If the device is stuck in a bootloop, press and release the Reset button. Then, press and hold on to the configured Back button and the Power Button to boot to the Home Screen.
+
+There can be issues with broken cache or config. In this case, delete the `.crosspoint` directory on your SD card (or consider deleting only `settings.bin`, `state.bin`, or `epub_*` cache directories in the `.crosspoint/` folder).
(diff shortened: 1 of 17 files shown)


osteotek:
Like this addition. But for AP mode do we need unique hostname? Correct me if I am wrong but it would matter only if we connect one x4 to another one running in AP mode, seems unlikely

lukestein:
Tested on device and seems to work as expected. And I can see how this will be useful.

My only concern is that there are third-party workflows (e.g., using curl, Apple Shortcuts) that expect to be able to reach the device at `crosspoint.local`. With this PR, we need to use `XTeinkX4-XXXXXX.local` .

I'm not sure how to weigh the additional hassle we create for those workflows requiring users to enter their device MAC address versus the ability to connect multiple x4 devices to the network simultaneously.

Of course, using device specific hostnames could be made optional (via a setting), but I'm also mindful of setting proliferation. Not sure what I think the best path forward is, so will look for further community input.

reukiodo:
@lukestein, since you have tested this - could you tell me what the XXXXXX of your X4 are? I just got a second X4 in, and it seems to be the same EABF8C as my first X4! I was operating under the assumption that all MAC should be unique, according to the standards.

At least for me, this does not provide a solution to having both of my X4 on the same network. 🤔

lukestein:
> @lukestein, since you have tested this - could you tell me what the XXXXXX of your X4 are? I just got a second X4 in, and it seems to be the same EABF8C as my first X4! I was operating under the assumption that all MAC should be unique, according to the standards.

@reukiodo:

Mine is at `https://xteinkx4-04a6b0.local`.

My device MAC address is `b0:a6:04:48:70:14` so it seems like your code is using the *first* three blocks of the MAC address (in reverse order), which is the Organizationally Unique Identifier (OUI) and therefore **likely the same** on different x4s (although yours and mine look to have different OUIs)

See #381 for code that may be helpful for getting the *last* three (what you want for uniqueness).

reukiodo:
I've made the changes and now verified they are the LAST 3, but how do I update that here in this pull request?

daveallie:
What's the behaviour when multiple devices are on the network? Just that one of them is reachable via crosspoint.local and the other must be navigated to via IP address.

I have a similar concern to @lukestein in that this would mean that any tool relying on the device being accessible at crosspoint.local would now need user configuration. Can you talk through a little more about the usecase of having multiple X4 devices on the network? Is it unreasonable to ask users to connect via IP address if they have multiple devices on the network?

I don't think we should accept this as is, as for most users they'll have a single device, and the problem described in this PR doesn't exist. Instead those users may now face new issues with an unknown (but thankfully deterministic) mDNS registration on the network.

reukiodo:
If multiple CrossPoint devices are on the same network (in file transfer), the epub will go to a random one. This is really only an issue if the user is attempting to transfer on a public network (like a school campus or airport wifi), as at home it is highly unlikely to be an issue as you state above unless housemates are both trying to transfer books at the same time ... resulting in maybe your book ending up on someone else's X4 ...

daveallie:
That is only an issue if you're using the mDNS crosspoint.local address right. In the interest of not breaking other integrations, I'd like to leave it as is, and encourage people to use the IP address of the device if they're in environments where there may be multiple X4 devices on the same network.

PR #679, opened 2026-02-03
feat(extensions): add partition-based app support

DChells:
## Summary

This PR adds **extension/app support** to CrossPoint by allowing standalone apps to live on SD and be launched via the existing OTA partition mechanism.

It also improves the WiFi sideload workflow so you can upload an app package from the browser without pulling the SD card.

Addresses discussion #257:
- https://github.com/crosspoint-reader/crosspoint-reader/discussions/257

Related context:
- https://github.com/crosspoint-reader/crosspoint-reader/discussions/575

## SD layout

Apps live on SD under:

```
/.crosspoint/apps/<appId>/
  app.json   # manifest (id/name/version/author/minFirmware...)
  app.bin    # ESP32-C3 firmware image
```

CrossPoint discovers apps by scanning `/.crosspoint/apps/*/app.json` and exposes them in **Home → Apps** (can be disabled).

## Install / launch behavior

Launching an app sets the boot partition to the app image and reboots.

To avoid unnecessary flashing, CrossPoint hashes `app.bin` and only flashes when the binary changed:
- unchanged: switch boot partition + reboot (no flash)
- changed/new: flash to the other OTA slot + reboot

Safety checks during flashing:
- SD ready
- Battery >= 20%
- Magic byte validation (`0xE9`)
- File size < target partition size
- Target partition is not the currently-running partition

## WiFi sideload (File Transfer → Apps)

The File Transfer web UI includes an **Apps** page:
- Upload `app.bin` (requires App ID; metadata optional)
- Upload `app.zip` (recommended): must contain `app.bin` + `app.json` (zero fields)

When Apps are disabled (Settings → System → Enable Apps):
- Home hides the Apps menu
- Web Apps page + app upload endpoints are blocked/hidden

## OTA Update Interaction

This reuses the existing two-slot OTA model. A firmware OTA update may overwrite the currently-installed app slot.

This does not brick anything: the app remains on SD and can be reinstalled/launched again.

## How to try it

1) Build and upload CrossPoint:
```bash
.venv/bin/pio run -e default -t upload
```

2) Put an app on SD (recommended):
- `app.zip` containing:
  - `app.bin`
  - `app.json` (must include: id/name/version)

3) Upload over WiFi:
- Device: Home → File Transfer
- Browser: open the URL shown on device → Apps → upload

4) Install on device:
- Home → Apps → select app → Install

## Notes

- There is no embedded Hello World app in this repo; use the community sample firmware repo as a starting point:
  https://github.com/open-x4-epaper/sample-firmware
  Note: that sample firmware is not tailored for CrossPoint's OTA slot switching. If you flash it as-is, you'll likely be stuck booting it until you add a return path (e.g. switch boot partition back to CrossPoint and reboot).

_AI-assisted development. Changes tested on hardware._

### DIFF
--- a/src/activities/apps/AppsActivity.cpp
+++ b/src/activities/apps/AppsActivity.cpp
@@ -52,7 +52,9 @@ void AppsActivity::loop() {
     }
   } else if (mappedInput_.wasReleased(MappedInputManager::Button::Back)) {
     if (exitCallback_) {
-      exitCallback_();
+      auto callback = exitCallback_;
+      callback();
+      return;
     }
   }
 


Uri-Tauber:
I'm afraid this feature is out of scope per SCOPE.md. see: https://github.com/crosspoint-reader/crosspoint-reader/pull/640

DChells:
> I'm afraid this feature is out of scope per SCOPE.md. see: #640

I do not think this is out of scope because this is implementing what @daveallie mentioned in my discussion #575 where he said "I'd love to find a solution for the community to boot more than just CrossPoint on their devices which would be a good place for games and puzzles like this, but I don't think they belong directly in CrossPoint". This is an attempt  to create an extension/app mechanism which was also discussed in #257 standalone firmware images installed via the existing OTA slot model. So out of scope features, which users still might want, out of crosspoint firmware itself while still utilizing crosspoint and more specifically the existing sdk. Although This probably needs to set terminology whether to call it apps/extensions/plugins as that might cause confusion since this isn't to add actual app in the firmware but to allow people to boot more than crosspoint on their device easily from their device instead of having to be more technical.

ngxson:
FWIW, I think the approach has a major problem is that apps must be flashed into flash memory before executing. This reduces the overall lifespan of the NAND memory, especially when user can have a large number of apps.

Loading the program into RAM makes more sense as it aligns with how modern OS works, but that will need ELF loader which does not exist on ESP32C3 platform.

Btw, I'm working on a javascript-based implementation of custom apps via [my fork](https://github.com/ngxson/pluspoint-reader) if you want to give it a try. Apps will be loaded to RAM, so will be limited in term of size. But it can also be compiled to byte code format to be more space-efficient. I'm working on optimizing assets loading, so big assets like sprites, database, etc can be loaded dynamically from SD card.

DChells:
> FWIW, I think the approach has a major problem is that apps must be flashed into flash memory before executing. This reduces the overall lifespan of the NAND memory, especially when user can have a large number of apps.
> 
> Loading the program into RAM makes more sense as it aligns with how modern OS works, but that will need ELF loader which does not exist on ESP32C3 platform.
> 
> Btw, I'm working on a javascript-based implementation of custom apps via [my fork](https://github.com/ngxson/pluspoint-reader) if you want to give it a try. Apps will be loaded to RAM, so will be limited in term of size. But it can also be compiled to byte code format to be more space-efficient. I'm working on optimizing assets loading, so big assets like sprites, database, etc can be loaded dynamically from SD card.

@ngxson Yeah you make a good point I didn't think about initially. After looking into it though, my Xteink X4 (got it from AliExpress not the official store) doesn't actually use NAND flash—it uses NOR flash specifically the PUYA-PY25Q128HA. According to the datasheet it supports 100k erase cycles so it shouldn't be a major issue, but you're right it's still not ideal to be constantly flashing.

What I ended up doing is adding a check so if you already flashed an app, it just switches the OTA partition instead of reflashing. Should cut down on the flash wear assuming users mainly stick to the same app day-to-day and not constantly switching between apps.

I did check out your fork with the mquickjs implementation I thought it was a very cool apprach. Given the RAM limitations and other hardware constraints with that method though, I wanted to try this native binary approach first.

I verified the flash chip with:
`esptool --chip esp32c3 flash_id`

Output: 
Chip is ESP32-C3 (QFN32) (revision v0.4)
Features: WiFi, BLE
Manufacturer: 85
Device: 2018
Detected flash size: 16MB

Manufacturer 0x85 and device 0x2018 maps to the PUYA-PY25Q128HA according to the datasheet and flashrom support:
- https://www.puyasemi.com/download_path/%E6%95%B0%E6%8D%AE%E6%89%8B%E5%86%8C/Flash%20%E8%8A%AF%E7%89%87/PY25Q128HA_Datasheet_V2.3.pdf

- https://github.com/gyuwon-yonsei-cpss/flashrom-add-puya-py25q128ha-support

DChells:
One note on the "what if an app boot-loops" question (and the X4 reset button): with the current partition table we only have two OTA slots (app0/app1). Reset just reboots whatever slot is currently set as boot, so if an extension is crash-looping it will keep coming back.

The robust way to get an automatic return-to-CrossPoint is to adopt ESP-IDF OTA rollback: mark the newly-booted slot as pending verify, and only keep it if the app calls esp_ota_mark_app_valid_cancel_rollback() after it successfully starts. If it crashes/reboots before that (or a watchdog triggers), the bootloader can rollback to the previous slot automatically.

That’s a bigger change (bootloader/sdkconfig + defining "app is healthy" semantics), so I didn’t try to fold it into this PR. If the extensions flow here lands cleanly, we can follow up with a separate PR to add rollback/safe-mode behavior.

Uri-Tauber:
I'm closing this PR for now, as it seems to be a dead end. #1012 might be the direction to continue exploring this idea.

Issue #348, opened 2026-01-13
FIT mode in "Sleep Screen Cover Mode" shows a grid-like grain effect


pablohc:
### Affected Version

v0.13.1

### Describe the Bug

When switching modes in Sleep Screen Cover Mode, the cover image quality is not consistent:

CROP mode: The image displays with excellent quality, without distortions.
FIT mode: The image shows a grainy, grid-like effect that significantly reduces visual quality.

This behavior suggests that the adjustment algorithm in FIT mode introduces artifacts into the image.

### Steps to Reproduce

1. Choose CROP mode in Sleep Screen Cover Mode
2. Open a book
3. Put the device to sleep
4. Turn the device back on
5. Choose FIT mode in Sleep Screen Cover Mode
6. Put the device to sleep again
7. Observe the difference in image quality.

### Expected Behavior

The image should maintain similar quality in both modes, without grid effects or loss of sharpness.

### Screenshots
Perfect quality CROP Mode: ![Image](https://github.com/user-attachments/assets/fa6eba12-a901-40c4-9a35-8a309aee1190)

Grainy, grid-like quality FIT Mode: ![Image](https://github.com/user-attachments/assets/d0f2154a-2f46-4b2a-b6fa-59e763a8e162)

```shell

```

jonasdiemer:
Thanks for submitting this issue.

This is likely a result of the way the processing works:
1. JPEGs are converted to BMP - during this step, we generate a slightly larger picture so it can be cropped.
2. In case of "fit", this image is later scaled down, leading tho the above artifacts.

Solution: Generate different version for "fit" and "crop" during JPEG->BMP conversion.

pablohc:
> Thanks for submitting this issue.
> 
> This is likely a result of the way the processing works:
> 
> 1. JPEGs are converted to BMP - during this step, we generate a slightly larger picture so it can be cropped.
> 2. In case of "fit", this image is later scaled down, leading tho the above artifacts.
> 
> Solution: Generate different version for "fit" and "crop" during JPEG->BMP conversion.

Thank you very much for the detailed clarification @jonasdiemer

jonasdiemer:
Looking into fixing it.

jonasdiemer:
Should be fixed in #358

PR #683, opened 2026-02-03
fix: dynamic continue reading card sizing based on cover image

pablohc:
Original Issue: #348

## Summary

- Adapt card width to cover image aspect ratio when available
- Maintain fixed 400px height for consistent layout
- Use actual image dimensions instead of hardcoded 240px width for EPUBs
- Add fallback to half-screen size when no cover image exists
- Parameter correction to avoid rendering artifacts

- HomeActivity.cpp: Dynamic logic to adapt the card width according to the cover's aspect ratio
- Epub.cpp: Calculation of dynamic dimensions based on the actual dimensions of the JPG image
- JpegToBmpConverter.cpp: Parameter correction to avoid rendering artifacts based on fix #662

## Additional Context

**Continue Reading Card**

v0.16.0 artifacts / fixed size card:
IMG20260203122332

Fix #662 without artifacts / fixed size card:
IMG20260203120651

This PR without artifacts / dinamic card width:
IMG20260203120959

---

### AI Usage

### DIFF
--- a/lib/Epub/Epub.cpp
+++ b/lib/Epub/Epub.cpp
@@ -468,10 +468,47 @@ bool Epub::generateThumbBmp() const {
       coverJpg.close();
       return false;
     }
-    // Use smaller target size for Continue Reading card (half of screen: 240x400)
-    // Generate 1-bit BMP for fast home screen rendering (no gray passes needed)
-    constexpr int THUMB_TARGET_WIDTH = 240;
-    constexpr int THUMB_TARGET_HEIGHT = 400;
+
+    // Generate 1-bit BMP maintaining aspect ratio with fixed height of 400px
+    // This ensures the card width adapts to the actual cover image aspect ratio
+    const int THUMB_TARGET_HEIGHT = 400;
+
+    // Read a small portion of the JPG file to determine its dimensions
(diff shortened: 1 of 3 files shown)


lukestein:
Seems to work great for me. Thank you! 🎉

Not pictured here, but also seems to work as expected for me with XTC files, text files, and EPUBs without a cover.

<details>
<summary> Old vs. New images (example 1)</summary>

IMG_3185

IMG_3188
</details>

<details>
<summary> Old vs. New images (example 2)</summary>

IMG_3186

IMG_3187
</details>

pablohc:
> Seems to work great for me. Thank you! 🎉
> 
> Not pictured here, but also seems to work as expected for me with XTC files, text files, and EPUBs without a cover.
> 

Thanks @lukestein!! ;)

The "Continue Reading" card has the same aspect ratio as the screen, half the width and height, so the covers of XTC files scale without any noticeable difference from before. ePubs without covers use the same card size. 
If the XTC format saves the cover dimensions somewhere, we could obtain the original proportions and create a new PR where we crop the cover and avoid white borders on all four sides.

pablohc:
Conflicts solved in PR #990

PR #537, opened 2026-01-25
feat: add charging indicator to battery icon (#516)

didacta:
Summary

  - What is the goal of this PR? 
  Adds a visual charging indicator (lightning bolt) to the battery icon when the
  device is plugged in via USB.
  - What changes are included?
    - Add isCharging() function to Battery.h that detects USB connection via GPIO20
    - Extend drawBattery() to accept a charging parameter
    - Draw white lightning bolt inside the battery fill area when charging
    - Ensure bolt visibility at low battery by enforcing minimum fill width
    - Update all call sites (HomeActivity, EpubReaderActivity, TxtReaderActivity)

  Additional Context

image preview: 
photo_2026-01-24_20-16-42

  - The charging state is detected by reading GPIO20 (U0RXD), which reads HIGH when USB is connected - same pin
  already used for serial detection in main.cpp
  - The lightning bolt is drawn in white (inverted) for visibility against the black battery fill
  - At very low battery levels (<~70%), the fill is slightly extended when charging to ensure the bolt remains
  fully visible
  - E-ink refresh behavior means the icon updates on next screen redraw (page turn, menu interaction), not
  immediately on plug/unplug - this is expected

  Closes #516

  ---
  AI Usage

  Did you use AI tools to help write this code? YES

  ---

Issue #516, opened 2026-01-23
Battery Charging Icon


FiftyFive74:
### Affected Version

0.15.0

### Describe the Bug

Not really a "bug" and it's a very minor detail but the battery icon doesn't change when device is charging (e.g. lightning bolt icon over the battery).

### Steps to Reproduce

See battery icon at the top right corner on main home screen or in the reading screen at the bottom left corner when device is plugged in and charging.

### Expected Behavior

Lightning bolt or other graphic indicating that the battery is charging.

### Relevant Log Output/Screenshots

```shell

```

didacta:
I started working on this, will let you know once I open a PR
edit: see [PR](https://github.com/crosspoint-reader/crosspoint-reader/pull/537)

FiftyFive74:
> I started working on this, will let you know once I open a PR edit: see [PR](https://github.com/crosspoint-reader/crosspoint-reader/pull/537)

Thank you

steka:
I think this one can be closed, see merged PR #1427.

znelson:
Agree, this is done now

PR #684, opened 2026-02-03
feat: Add battery percentage smoothing and charging icon

fischer-hub:
## Summary

Building on the changes in #537 (draw battery charging icon while charging device), this PR adds smoothing to the battery percentage displayed, based on its recent values. This should result in more expected battery behaviour (not so sudden drops in percentage when the temperature changes temporarily, no flickering between two values when on the edge of two percentage values.

* changes implemented in #537, as well as the requested changes in the review (move charging check into `drawBattery()`)
* a ringbuffer handling the last 10 battery percentage readouts (updated on `drawBattery()` call) as well as logic to display the average battery percentage of the ringbuffer. This also keeps the battery percentage from increasing unexpectedly while the device is not charged

## Additional Context

* Testing is a bit difficult because it relies on environmental factors such as temperature, so additional testing by other users would be great.

---

### AI Usage

### DIFF
--- a/src/components/themes/lyra/LyraTheme.h
+++ b/src/components/themes/lyra/LyraTheme.h
@@ -34,7 +34,7 @@ constexpr ThemeMetrics values = {.batteryWidth = 16,
 
 class LyraTheme : public BaseTheme {
  public:
-  LyraTheme(HalGPIO* gpio) : BaseTheme(gpio) {}
+  explicit LyraTheme(HalGPIO* gpio) : BaseTheme(gpio) {}
   // Component drawing methods
   //   void drawProgressBar(const GfxRenderer& renderer, Rect rect, size_t current, size_t total) override;
   void drawBattery(const GfxRenderer& renderer, Rect rect, bool showPercentage = true) const override;


lukestein:
Continuing testing, but it looks like maybe when plugged into the charger, the battery always shows 100% (not just in icon, but in battery percentage text)?

fischer-hub:
@lukestein Thanks for testing! While charging the `.readPercentage()` Arduino function always returns 99% /100%, so that was already like this before this PR. Not sure if there is an easy way to fix it and correctly read out the battery charge while charging.

fischer-hub:
Adjusted to the new Theme architecture, not sure how the lyratheme wants to handle the charging battery icon so I skipped that part and just implemented the buffer changes here. The base theme uses the charging icon from #516 still.

I also feel like now that there are theme-dependent functions to draw the battery, this change should maybe not go into that function to avoid duplicated code. Maybe someone has some input on that

src/components/themes/BaseTheme.cpp
```diff
@@ -19,10 +19,25 @@ constexpr int homeMenuMargin = 20;
...
+  uint8_t percentage = battery.readPercentage();
+
+  if (charging) {
+    // If charging reinitialize buffer with current percentage and display this value
+    batteryBuffer.init(percentage);
+  } else {
+    // Else update buffer with new percentage and return smoothed validated percentage to display
+    batteryBuffer.update(percentage);
+    percentage = batteryBuffer.evaluate();
+  }
```
daveallie: Just thinking out loud, I wonder if we should encapsulate this logic into the buffer's `update` function to have it read from the battery store the internal value and return the averaged one. I can't see any need to update the buffer but not have immediately previously called `readPercentage` each time.

ngxson: Yeah, this logic should be moved to `HalGPIO`. I already moved the battery percent logic there, just forgot to replace it in the code

fischer-hub: My intuition here was that we anyway have to detect whether the device is charging in `drawBattery()` to draw the correct battery icon, so we can handle the buffer behavior on charge here too. But we can also move it and check for charging twice, I don't mind.

ngxson: Not quite sure if I understand your intention correctly, but what I imagine is that:
- Charging state is already provided by `HalGPIO::isUsbConnected()`
- Raw battery state is already provided by `HalGPIO::getBatteryPercentage()`
- You only need to implement the smoothed version of it, maybe `getBatteryPercentage(bool isSmooth)`, and the ring buffer can be managed by HalGPIO

lib/hal/HalGPIO.h
```diff
@@ -26,6 +28,8 @@ class HalGPIO {
  public:
   HalGPIO() = default;
 
+  static BatteryPercentageRingBuffer batteryBuffer;
```
daveallie, line 31: Why is this static?

src/components/UITheme.cpp
```diff
@@ -11,8 +11,11 @@
 
 UITheme UITheme::instance;
 
-UITheme::UITheme() {
+UITheme::UITheme() : gpio(nullptr), currentMetrics(nullptr), currentTheme(nullptr) {}
```
daveallie, line 14: We have significantly changed the empty constructor's logic, if we don't want it/expect it to be called, then we should not declare it. Previously, it was setting the theme, but but it does nothing.

src/main.cpp
```diff
@@ -305,7 +305,8 @@ void setup() {
 
   SETTINGS.loadFromFile();
   KOREADER_STORE.loadFromFile();
-  UITheme::getInstance().reload();
+  UITheme uiTheme(gpio);
+  uiTheme.getInstance().reload();
```
daveallie, line 309: This is really odd and I'm confused why it would even work. We have a singleton instance of `UITheme` accessible via `UITheme::getInstance()`. You've constructed a new `UITheme` with an assigned gpio and then called `getInstance()` on it returning the other, static instance before calling reload (and ultimately discarding the newly constructed `uiTheme`).

Your code is equivalent to:
```cpp
UITheme uiTheme(gpio);
UITheme::getInstance().reload();
```

Where that first line is setting gpio as an instance variable to that new local instance.

fischer-hub:
> This has all gotten fairly stinky in the refactor. We should probably be setting the UITheme's static instance's HalGPIO as part of the boot sequence instead of during initialisation.

Then I might not be the right person to go on with this implementation, I dont think Im capable of adding the newly requested changes.

PR #694, opened 2026-02-05
fix: prevent spurious spaces before attaching punctuation

jdk2pq:
Fixes issue #182

## Summary

When inline styles change mid-paragraph, words like periods, commas, and quotes could end up as separate tokens. The justified text algorithm was treating these as regular words, adding space before them.

Now tracks which words are "attaching punctuation" (., , ! ? ; : " ' and smart quotes) and excludes them from gap counting. These punctuation marks attach directly to the preceding word without spacing.

## Additional Context

This is split out from code in #411 to address this comment https://github.com/crosspoint-reader/crosspoint-reader/pull/411#discussion_r2751166631

---

### AI Usage

Did you use AI tools to help write this code? _**YES**_, Claude Code


lukestein:
Testing this on device, and the extraneous spaces all seem to be gone. Thank you!🎉 

Before and after (in a spot with terrible justified spacing either way, but highlights the fix):

IMG_3201

IMG_3202

PR #704, opened 2026-02-05
feat: Add "Short Power Button Click" option to rotate reader orientation

GavinHigham:
## Summary

I wanted a quicker way to rotate the display to best fit the content that I'm reading.
Added an additional option to Settings->Controls->Short Power Button Click, "Reader Rotate". When this option is chosen, short pressing the power button while reading will rotate the display.

IMG_3078

## Additional Context

  specific areas to focus on).
Also removed the `applyReaderOrientation` helper function across several readers. By using static_assert we can make casting between the Gfx and Settings orientation enums safe, so the helper function amounts to a cast which can just be done inline.

---

### AI Usage

**Did you use AI tools to help write this code?**
Partially, yes. I asked GPT 5.2 Codex to identify the areas that would need to be changed for this change, let it make the change, reviewed its changes, tested on device, and made an additional change myself to avoid overflowing text in the settings menu. Then when rebasing I went over everything again and rewrote most of it (and removed some bits that in retrospect didn't make any sense).

### DIFF
--- a/src/CrossPointSettings.h
+++ b/src/CrossPointSettings.h
@@ -114,7 +114,7 @@ class CrossPointSettings {
   };
 
   // Short power button press actions
-  enum SHORT_PWRBTN { IGNORE = 0, SLEEP = 1, PAGE_TURN = 2, SHORT_PWRBTN_COUNT };
+  enum SHORT_PWRBTN { IGNORE = 0, SLEEP = 1, PAGE_TURN = 2, ROTATE_ORIENTATION = 3, SHORT_PWRBTN_COUNT };
 
   // Hide battery percentage
   enum HIDE_BATTERY_PERCENTAGE { HIDE_NEVER = 0, HIDE_READER = 1, HIDE_ALWAYS = 2, HIDE_BATTERY_PERCENTAGE_COUNT };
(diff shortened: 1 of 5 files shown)


GavinHigham:
Sorry, I cannot seem to find an option to rename the title. It should be prepended with "feat: "

Eloren1:
I can't rename mine as well now

GavinHigham:
I've rebased onto the most recent changes and went over all the code again to improve a few things.

finn-mcswiggan:
Is there a way to also automatically flip the prev/next page buttons when the quick rotation is applied? I.e in landscape cw flip the side buttons, inverted flip both buttons, landscape ccw flip the bottom button

GavinHigham:
That's a good idea - it should probably be a different PR adding a new setting or two, working with the global reader orientation control, not just this quick rotate. Related to #449 and #460.

GavinHigham:
> Is there a way to also automatically flip the prev/next page buttons when the quick rotation is applied? I.e in landscape cw flip the side buttons, inverted flip both buttons, landscape ccw flip the bottom button

I've addressed this in #966!

itsthisjustin:
@znelson @vjapolitzer @osteotek Is this something we want to eventually merge? Stuff like this should probably go into a discussion for community voting

itsthisjustin:
I'm going to close this for now as it's stale and has critical coderabbit suggestions but it IS a good idea and should follow the same patterns our existing short button settings provide

PR #714, opened 2026-02-05
feat: page text export via Confirm long-press

tnaftali:
## Summary

Adds a lightweight page text export feature that saves the current page's rendered text to a `.txt` file on the SD card. Triggered by holding the Confirm button for ~1 second during reading.

- **Long-press Confirm (1s)** exports the current page text to `/.crosspoint/exports/<book-title>.txt`
- **Short-press Confirm** still opens the reader menu (existing behavior preserved)
- Status bar briefly shows "Page saved" / "Save failed", clears on next page turn
- Each export appends to a per-book file with chapter name, page number, and book progress metadata
- Files are plain UTF-8 text, easy to read on any device after pulling off the SD card

Related discussion: #481

## Motivation

The Xteink X4 has no touchscreen, making traditional text highlighting impractical. This provides a simple alternative for saving passages for later reference — hold a button, and the full page text is captured with context.

## Implementation

- `TextBlock::getPlainText()` / `Page::getPlainText()` — text extraction from the existing page data structures
- `PageExporter` — new utility class handling FAT32 filename sanitization, directory creation, and append-mode file writing
- `EpubReaderActivity` — Confirm button split into long-press (export) and short-press (menu), with `exportTriggered` guard to prevent repeated exports while held
- Status bar override mechanism for temporary feedback, cleared on page turn

## Export file format

```
== The Great Gatsby — F. Scott Fitzgerald ==

--- Chapter 3: The End of Something | Page 42 | 18% ---
In my younger and more vulnerable years my father
gave me some advice that I've been turning over in
my mind ever since...

--- Chapter 5: The Party | Page 87 | 38% ---
The lights grow brighter as the earth lurches away
from the sun...
```

## Test plan

- [x] Build compiles cleanly (RAM: 31.2%, Flash: 95.7%)
- [x] Flashed and tested on Xteink X4 hardware
- [x] Short-press Confirm still opens reader menu
- [x] Hold Confirm ~1s shows "Page saved" in status bar
- [x] Status bar reverts on next page turn
- [x] Export file created at `/.crosspoint/exports/`
- [x] Multiple captures append to same file
- [ ] Books with no title fall back to hash filename

### Actual files generated on my device while testing the functionality
- [A_months_retrospective_from_the_developer__rxteinkereader.txt](https://github.com/user-attachments/files/25113944/A_months_retrospective_from_the_developer__rxteinkereader.txt)
- [Red_Rising.txt](https://github.com/user-attachments/files/25113945/Red_Rising.txt)

### DIFF
--- a/src/activities/reader/EpubReaderActivity.cpp
+++ b/src/activities/reader/EpubReaderActivity.cpp
@@ -936,7 +936,6 @@ void EpubReaderActivity::renderContents(std::unique_ptr<Page> page, const int or
 
   // restore the bw data
   renderer.restoreBwBuffer();
-
 }
 
 void EpubReaderActivity::renderStatusBar(const int orientedMarginRight, const int orientedMarginBottom,


kurtisgrant:
I tested this out and it needs some changes but it seems like a good start. (Didn't review code, just functionality)

My generated .txt file excerpt:
```
== A Tale of Two Cities — Charles Dickens ==

--- II: The Grindstone | Page 27 | 69% ---
have—as I believe you
have—make yourself known
to these devils, and get taken
to La Force. It may be too...
```

Major:
- I thought it wasn't working initially because it is putting the export dir in .crosspoint, thus making it functionally invisible both within the reader's file browser and on the computer when browsing the SD card. Since crosspoint can process .txt files, we should be able to open and view these exports on the reader itself as well as on desktop obviously.  So the **exports dir should be in root, not .crosspoint** (May be worth renaming exports to "my saves" or "saved passages" or something as well, since many users may opt to only interface with these files exclusively within the reader's file browser which would not really feel like an "export")

Minor:
- This feature should to be added into the docs, since there is nothing indicating the feature is available from the UI itself. 
- I don't think we should include the page number for each excerpt as it isn't a true page number and changes depending on the current font size setting. 
- With the current formatting, the percentage almost looks like it is a percentage of chapter completion, while it is actually book completion. This being said, percentages may be the only thing we can use to indicate location in the absence of stable page numbers so it might be worth trying to find a way to explicitly state the book percentage AND chapter completion percentage. Something like:
--- II: The Grindstone | 69% of title | 78% of chapter ---
(Although I admit doesn't look as nice, maybe we can think of some alternative)
- It would be really nice to be able to capture additional pages of text in one "save" by pressing the next page button while holding the initial confirm button, after the save has been initiated. Especially since with the small page size, the chance of the passage you like being chopped in half by a page division is quite high. (This is just my preference though and isn't needed to make the feature complete, imo)

lukestein:
Some nice suggestions from @kurtisgrant. I think in particular making the clippings files readable on device would be a real improvement.

I am testing on device and the UI (long press confirm; success indicated in status bar area) is very slick. Part of me wonders whether this belongs in the reader menu instead of (or perhaps in addition to?) access via long-press. Happy to get some more community input.

iandchasse:
Maybe capture the page after it too in case a quote cuts off? Or maybe just a setting for X pages before and Y pages after?

I could see this as a good basic option for “storing” passages linked to a book % that can be jumped back to later. Maybe a separate menu can go through the passages and let the user manually delete the portions of text that aren’t needed after the fact?

drbourbon:
Nice feature, thanks! (given the small page size of the device and the lack of a touch screen this looks like a reasonable alternative to proper **text highlighting**)

A note on the format: what about adopting a `markdown` syntax? 

> This would open up the possibility for better parsing and on device rendering when we'll have actual markdown support in crosspoint (not to mention the possibility to implement highlights marge & sync between devices by implementing something like https://github.com/gitalexcampos/highlightsync.koplugin)

Example:

```markdown
---
title: "The Great Gatsby"
author: "F. Scott Fitzgerald"
---

# The Great Gatsby — F. Scott Fitzgerald 

## Chapter 3: The End of Something | Page 42 | 18% 
In my younger and more vulnerable years my father
gave me some advice that I've been turning over in
my mind ever since...

## Chapter 5: The Party | Page 87 | 38% 
The lights grow brighter as the earth lurches away
from the sun...
```

tnaftali:
Hi everyone, thanks for the feedback! I'll see what I can do about the suggestions and keep you posted.

tnaftali:
## v2: Multi-page passage capture with markdown export

  This replaces the single-page text export with a full capture state machine and addresses all the review feedback.

  ### Changes

  **File format & location**
  - Moved from `/.crosspoint/exports/` to `/Saved Passages/` (visible in file browser and desktop)
  - Changed from `.txt` to `.md` with YAML frontmatter
  - Dropped page number (unstable across font sizes), added book % and chapter %
  - Format: `## Chapter Name | 69% of book | 78% of chapter`
  - `---` separator between passages for readability

  **Multi-page capture mode**
  - Long-press Confirm toggles capture on/off (IDLE → CAPTURING → saved)
  - Forward page turns append to buffer, each page captured automatically
  - Backward turn saves the passage then navigates back
  - Back button discards capture without saving
  - Section boundary crossings handled via deferred capture after render

  **Reader menu integration**
  - Added "Save Passage" as 2nd menu item
  - Shows "Stop Capture" when already capturing (saves the passage)
  - Capture state persists across menu visits

  **Status bar**
  - `*` indicator prepended to chapter title during capture
  - Fallback standalone `*` for status bar modes without chapter title

  **Safety**
  - Books inside `/Saved Passages/` blocked from capture ("Cannot capture here")

  **Docs**
  - Added "Saving Passages" section to USER_GUIDE.md covering quick save, multi-page save, all stop methods, and file location

  ### Testing

  Built, flashed, and tested on device:
  - Single-page capture (hold → hold)
  - Multi-page capture across pages and chapter boundaries
  - Backward-stops-capture behavior
  - Menu start/stop capture toggle
  - Back-discards-capture behavior
  - File format verified on SD card

File sample:
- [Bullet_Train.md](https://github.com/user-attachments/files/25123514/Bullet_Train.md)

tnaftali:
### Small update
 The passages files are now named the same as the books (but .md).

osteotek:
The text export feature is much needed and seems like a great addition. However, I believe the long-press confirm button should be reserved for bookmarks, similar to the stock firmware. Perhaps this PR could be combined with the bookmarks feature, where you can bookmark a page first and later export it if needed.

tnaftali:
@osteotek thank you for the feedback! Maybe we can keep the menu operated option only for now? What do you think?

kludgescarcity:
Maybe someone can take on the task in the new PR to add Bookmark functionality, include a Bookmark option in the Reader Menu, and create a small pop-up menu that appears after a long press Confirm. This menu should have two options: "Bookmark" and "Save." (Sync as third option perhaps?) I don't have any experience with C++, unfortunately.

osteotek:
> Maybe someone can take on the task in the new PR to add Bookmark functionality, include a Bookmark option in the Reader Menu, and create a small pop-up menu that appears after a long press Confirm. This menu should have two options: "Bookmark" and "Save." (Sync as third option perhaps?) I don't have any experience with C++, unfortunately.

Come to think of it, this is already pretty much bookmark functionality, isn't it? Just needs a little reorganizing and UI. We should just combine two actions, Bookmark should also do export, there is no harm in it

tnaftali:
### Update: Bookmark & Save

Based on feedback from @osteotek and @kludgescarcity, long-press Confirm now shows a **popup menu** with two options instead of directly starting a capture:

- **Bookmark** — records the current page location (chapter, book %, chapter %) to the book's `.md` file
- **Save Passage** — starts multi-page capture (same as before)

**How it works:**

1. Long-press **Confirm** (~1s) — a small popup appears
2. Use **Left**/**Right** to select an option
3. Press **Confirm** to execute, or **Back** to dismiss

The reader menu item has been renamed from "Save Passage" to **"Bookmark & Save"** and opens the same popup.

While capturing, long-press Confirm still stops the capture directly (no popup).

**Bookmark format in the `.md` file:**

```
III: The Shadow | 72% of book | 2% of chapter

Bookmarked
```

Let me know what you think!

osteotek:
But is there a way in the ui to review bookmarks and jump to their position? That's the main point of the bookmark

DavidOrtmann:
> But is there a way in the ui to review bookmarks and jump to their position? That's the main point of the bookmark

Maybe having it as a second tab in the chapter selection screen would make sense?

tnaftali:
Closed in favour of #734

PR #715, opened 2026-02-05
feat: add host-mounted SD card support, mirror display to host

ngxson:
## Summary

This PR is built upon the recently added HAL infrastructure. It aims to help developers to test and debug changes more quickly.

The "[emulator](https://en.wikipedia.org/wiki/Emulator)" in this context is not a VM (not QEMU, see https://github.com/crosspoint-reader/crosspoint-reader/pull/500 for why), but more like the `-v` mounting option on Docker. Currently, it allows to:
- Use `.sdcard` directory on host machine (or PC) as a virtual SD card
- Forward frame buffer to be displayed on host

It requires https://github.com/crosspoint-reader/crosspoint-reader/pull/656 to work (hence why you see a lot of changes here)

https://github.com/user-attachments/assets/81c865be-2839-4c1e-bdd5-9bd0c69b2b4e

To try this out:
1. Switch to the `emulated` pio environment
2. Compile and upload the firmware
3. Run `python scripts/dev_server.py`
4. You may need to press the Reset button on the device
5. Once the emulator is up, go to http://127.0.0.1:8090

Known bugs:
- Not yet tested on windows / linux - only tested on mac
- File operations are extremely slow compared to a real SD card
- Sometime, both log and data are using the same `Serial` object. If 2 tasks (aka 2 threads) trying to use `Serial.print` at the same time, the emulator can crash. We should implement a proper logger in near future.

TODO:
- [ ] Wait for https://github.com/crosspoint-reader/crosspoint-reader/pull/656 to be merged
- [ ] Write more docs
- [ ] Test it on linux
- [ ] Make it work on windows?

---

### AI Usage

Did you use AI tools to help write this code? --> **The python script and web UI are vibe-coded. C++ is hand-written.**


osteotek:
Can this be used for automated testing, I wonder

ngxson:
I think this PR lays the foundation of communication between host <--> device, which may enable the development of auto or semi-auto testing tools in the future.

I imagine having a script that automatically compiles the firmware, flash it to the device, prepare a fake `.sdcard` with some epub/txt/etc inside, then tell the device to open these. The script can then validate the log, the frame buffer and/or the output cache file written to `.crosspoint`

itsthisjustin:
Due to how stale this is, the opportunity for a mistake during merge conflicts seems very high. I think it's a great idea but I'd recommend a fresh PR without trying to merge with upstream so I'm closing this for now. It's a solid idea but 400+ commits behind and 20+ conflicted files is very risky.

PR #700, opened 2026-02-05
chore(release): 1.0.0 Release Candidate

daveallie:
## Current RC Download

[1.0.0-rc+9e04eec](https://github.com/crosspoint-reader/crosspoint-reader/actions/runs/21805762722/artifacts/5424767542)

## CrossPoint Release Candidate

This is a CrossPoint Release Candidate. Given the speed of changes in this repo, and the volume of people using it, running a little pre-release testing is becoming necessary. This set of changes will make up the public release, any functional issues found in testing this branch over the next few days will be merged back in before the release is cut.

### Is this for me?

Are you feeling a little adventurous? Do you feel comfortable running into a few last minute bugs? Then this might be for you. As mentioned below, this will only typically be relevant for a few days before these changes become generally available.

### What do I need to do?

If you're willing to test out this build, just run it on your device over the next few days and report back any functional issues you find with the release. Keep in mind the feedback for this RC should not be feature requests or recommendations, they should only be bugs which would block this release from going to the public.

### Where do I find the build?

The easiest way is to look at the most recent completed "Compile Release Candidate" action which should have an artifact on it containing the firmware.bin file. You can directly flash this file using https://xteink.dve.al.

### When's it going to be generally available?

I'll typically leave a few days between creating a release candidate, and then cutting a public release in order to gather feedback.

---

## What's Changed

There are a fair number of large changes in this release, including a new default UI theme, Lyra, CSS parsing within EPUBs, and full front button remapping support.

More thorough release notes will be written for the full release, as an RC tester take a scan of the items below and keep them in mind as they are the most likely areas to encounter issues.

### Features
* feat: add HalDisplay and HalGPIO by @ngxson in https://github.com/crosspoint-reader/crosspoint-reader/pull/522
* feat: Add reading menu and delete cache function by @jonasdiemer in https://github.com/crosspoint-reader/crosspoint-reader/pull/433
* feat(ui): change popup logic by @osteotek in https://github.com/crosspoint-reader/crosspoint-reader/pull/442
* feat: Debugging monitor script by @Uri-Tauber in https://github.com/crosspoint-reader/crosspoint-reader/pull/555
* feat: Add CSS parsing and CSS support in EPUBs by @jdk2pq in https://github.com/crosspoint-reader/crosspoint-reader/pull/411
* feat: UI themes, Lyra by @CaptainFrito in https://github.com/crosspoint-reader/crosspoint-reader/pull/528
[...]

PR #600, opened 2026-01-28
feat: unify navigation handling with system-wide continuous navigation

istiak-tridip:
This PR unifies navigation handling & adds system-wide support for continuous navigation.

## Summary
Holding down a navigation button now continuously advances through items until the button is released. This removes the need for repeated press-and-release actions and makes navigation faster and smoother, especially in long menus or documents.

When page-based navigation is available, it will navigate through pages. If not, it will progress through menu items or similar list-based UI elements.

Additionally, this PR fixes inconsistencies in wrap-around behavior and navigation index calculations.

Places where the navigation system was updated:
- Home Page
- Settings Pages
- My Library Page
- WiFi Selection Page
- OPDS Browser Page
- Keyboard
- File Transfer Page
- XTC Chapter Selector Page
- EPUB Chapter Selector Page

I’ve tested this on the device as much as possible and tried to match the existing behavior. Please let me know if I missed anything. Thanks 🙏

crosspoint

---

Following the request from @osteotek and @daveallie for system-wide support, the old PR (#379) has been closed in favor of this consolidated, system-wide implementation.

---

### AI Usage

Did you use AI tools to help write this code? _**PARTIALLY**_

PR #716, opened 2026-02-06
fix: Scrolling page items calculation

CaptainFrito:
## Summary

Fix for the page skip issue detected https://github.com/crosspoint-reader/crosspoint-reader/pull/700#issuecomment-3856374323 by user @whyte-j 

Skipping down on the last page now skips to the last item, and up on the first page to the first item, rather than wrapping around the list in a weird way.

## Additional Context

The calculation was outdated after several changes were added afterwards

---

### AI Usage

### DIFF
--- a/src/activities/home/HomeActivity.cpp
+++ b/src/activities/home/HomeActivity.cpp
@@ -268,7 +268,6 @@ void HomeActivity::render() {
     menuItems.insert(menuItems.begin() + 2, "OPDS Browser");
   }
 
-  Serial.printf("Home menu items count: %d\n", static_cast<int>(menuItems.size()));
   GUI.drawButtonMenu(
       renderer,
       Rect{0, metrics.homeTopPadding + metrics.homeCoverTileHeight + metrics.verticalSpacing, pageWidth,
--- a/src/components/UITheme.h
+++ b/src/components/UITheme.h
@@ -6,9 +6,6 @@
 #include "CrossPointSettings.h"
 #include "components/themes/BaseTheme.h"
 
-inline int min(const int a, const int b) { return a < b ? a : b; }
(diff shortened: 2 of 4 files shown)


lukestein (approved):
No code review, but tested on device and seems to fix the bugs with the file selectors (on both original and Lyra UIs).

Thank you!

This PR may get merged first as a bug fix, but since there may still be some weirdness in other page scrolls, I hope we may eventually get something like #600

src/activities/home/HomeActivity.cpp
```diff
@@ -272,6 +272,7 @@ void HomeActivity::render() {
     menuItems.insert(menuItems.begin() + 2, "OPDS Browser");
   }
 
+  Serial.printf("Home menu items count: %d\n", static_cast<int>(menuItems.size()));
```
daveallie: nit: Can we remove or follow the standard logging format?

src/components/UITheme.h
```diff
@@ -6,6 +6,9 @@
 #include "CrossPointSettings.h"
 #include "components/themes/BaseTheme.h"
 
+inline int min(const int a, const int b) { return a < b ? a : b; }
+inline int max(const int a, const int b) { return a > b ? a : b; }
```
daveallie: Remove these and just use `std::min` and `std::max` instead.

CaptainFrito: Ah thanks! that was the missing piece

CaptainFrito:
@lukestein this is ready for the next RC

lukestein (approved):
Re-tested on device (as usual, no code review from me) and continues to fix the bugs with the file selectors (on both original and Lyra UIs).

I'll leave this one for @daveallie et al to merge.

PR #721, opened 2026-02-06
fix: Lag before displaying covers on home screen

CaptainFrito:
## Summary

Reduce/fix the lag on the home screen before recent book covers are rendered

## Additional Context

We were previously rendering the screen in two steps, delaying the recent book covers render to avoid a lag before the screen loads. 
In this PR, we are now doing that only if at least one book doesn't have the cover thumbnail generated yet. If all thumbs are already generated, we load and display them right away, with no lag.

---

### AI Usage

Did you use AI tools to help write this code? _**NO **_


Uri-Tauber:
I've tested it and the lag is gone.
Great job, @CaptainFrito!

Eloren1:
Is the book selectable instantly without needing to wait for loading?

lukestein:
Fantastic UX improvement. Thanks @CaptainFrito!!

PR #723, opened 2026-02-06
feat: Auto Page Turn for Epub Reader

GenesiaW:
## Summary

    - Implements auto page turn feature for epub reader in the reader submenu

    - added auto page turn feature in epub reader in the submenu
    - currently there are 5 settings, `OFF, 1, 3, 6, 12` pages per minute

## Additional Context

  specific areas to focus on).
    - when auto turn is enabled, space reserved for chapter title will be used to indicate auto page turn being active
    - Back and Confirm button is used to disable it

---

### AI Usage

### DIFF
--- a/src/activities/reader/EpubReaderActivity.cpp
+++ b/src/activities/reader/EpubReaderActivity.cpp
@@ -599,8 +599,10 @@ void EpubReaderActivity::toggleAutoPageTurn(const uint8_t selectedPageTurnOption
     automaticPageTurnActive = true;
   }
 
-  // resets cached section so that space is reserved for auto page turn indicator when progress bar is None
-  if (SETTINGS.statusBar == CrossPointSettings::STATUS_BAR_MODE::NONE) {
+  const uint8_t statusBarHeight = UITheme::getInstance().getStatusBarHeight();
+  // resets cached section so that space is reserved for auto page turn indicator when progress bar only or None
+  if (statusBarHeight == 0 || statusBarHeight == UITheme::getInstance().getProgressBarHeight()) {
+    updateLastPageTurnTime = true;  // To only update lastPageTurnTime after rendering is complete
     // Preserve current reading position so we can restore after reflow.
     RenderLock lock(*this);
     if (section) {
@@ -644,8 +646,17 @@ void EpubReaderActivity::render(Activity::RenderLock&& lock) {
   orientedMarginTop += SETTINGS.screenMargin;
   orientedMarginLeft += SETTINGS.screenMargin;
   orientedMarginRight += SETTINGS.screenMargin;
-  orientedMarginBottom +=
-      std::max(SETTINGS.screenMargin, static_cast<uint8_t>(UITheme::getInstance().getStatusBarHeight()));
+  const uint8_t statusBarHeight = UITheme::getInstance().getStatusBarHeight();
+
+  // reserves space for automatic page turn indicator when no status bar or progress bar only
+  if (automaticPageTurnActive &&
+      (statusBarHeight == 0 || statusBarHeight == UITheme::getInstance().getProgressBarHeight())) {
+    orientedMarginBottom +=
+        std::max(SETTINGS.screenMargin,
+                 static_cast<uint8_t>(statusBarHeight + UITheme::getInstance().getMetrics().statusBarVerticalMargin));
+  } else {
+    orientedMarginBottom += std::max(SETTINGS.screenMargin, statusBarHeight);
+  }
 
   if (!section) {
     const auto filepath = epub->getSpineItem(currentSpineIndex).href;
@@ -831,9 +842,17 @@ void EpubReaderActivity::renderStatusBar() const {
   const float bookProgress = epub->calculateProgress(currentSpineIndex, sectionChapterProg) * 100;
 
   std::string title;
+  int textYOffset = 0;
 
   if (automaticPageTurnActive) {
     title = tr(STR_AUTO_TURN_ENABLED) + std::to_string(60 * 1000 / pageTurnDuration);
+
+    // calculates textYOffset when rendering title in status bar
+    const uint8_t statusBarHeight = UITheme::getInstance().getStatusBarHeight();
+    // offsets text if no status bar or progress bar only
+    if (statusBarHeight == 0 || statusBarHeight == UITheme::getInstance().getProgressBarHeight()) {
+      textYOffset = UITheme::getInstance().getMetrics().statusBarVerticalMargin;
+    }
   } else if (SETTINGS.statusBarTitle == CrossPointSettings::STATUS_BAR_TITLE::CHAPTER_TITLE) {
     const int tocIndex = epub->getTocIndexForSpineIndex(currentSpineIndex);
     if (tocIndex == -1) {
@@ -844,9 +863,7 @@ void EpubReaderActivity::renderStatusBar() const {
     }
   } else if (SETTINGS.statusBarTitle == CrossPointSettings::STATUS_BAR_TITLE::BOOK_TITLE) {
     title = epub->getTitle();
-  } else {
-    title = "";
   }
 
-  GUI.drawStatusBar(renderer, bookProgress, currentPage, pageCount, title);
+  GUI.drawStatusBar(renderer, bookProgress, currentPage, pageCount, title, 0, textYOffset);
 }
--- a/src/activities/reader/TxtReaderActivity.cpp
+++ b/src/activities/reader/TxtReaderActivity.cpp
@@ -436,7 +436,10 @@ void TxtReaderActivity::renderPage() {
 
 void TxtReaderActivity::renderStatusBar() const {
   const float progress = totalPages > 0 ? (currentPage + 1) * 100.0f / totalPages : 0;
-  std::string title = txt->getTitle();
+  std::string title;
+  if (SETTINGS.statusBarTitle != CrossPointSettings::STATUS_BAR_TITLE::HIDE_TITLE) {
+    title = txt->getTitle();
+  }
 
   GUI.drawStatusBar(renderer, progress, currentPage + 1, totalPages, title);
 }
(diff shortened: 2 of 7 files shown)


src/activities/reader/EpubReaderActivity.cpp
```diff
@@ -776,6 +832,31 @@ void EpubReaderActivity::renderStatusBar(const int orientedMarginRight, const in
                     showBatteryPercentage);
   }
 
+  // uses chapter name space to indicate auto page turn is enabled
```
lukestein: I am slightly concerned about baking pixel-level assumptions about the status bar layout into this feature's implementation.

If we are "allowing" features to use status bar space (and this is not the only current PR trying to), I suspect we want some abstraction. Also, this will work better with current requests for fuller status bar customization.

GenesiaW: > If we are "allowing" features to use status bar space (and this is not the only current PR trying to), I suspect we want some abstraction. Also, this will work better with current requests for fuller status bar customization.

I copied the code from the display of chapter name. I can probably clean up my code and reduce code duplication by changing the initialization of text in center of the status bar

GenesiaW:
> I like this but worried about ad hoc status bar overloading

Yes, I have been thinking of ways that could be used to indicate that auto page turn is enabled and it feels like the status bar is one of the least intrusive location to indicate it. 

Would love to get additional ideas/feedback on it can be better implemented

osteotek:
> > I like this but worried about ad hoc status bar overloading
> 
> Yes, I have been thinking of ways that could be used to indicate that auto page turn is enabled and it feels like the status bar is one of the least intrusive location to indicate it.
> 
> Would love to get additional ideas/feedback on it can be better implemented

What if status bar is disabled?

GenesiaW:
> What if status bar is disabled?
image

Here's a image of it enabled with a reader margin of 20.  However, at reader margin of 5, the text will be cut off at the bottom, I will have to edit the code to reserve some space if auto page turn is enabled. Great catch!

osteotek (dismissed):

osteotek:
Seems like a nice addition to feature parity with the stock firmware.

jdk2pq:
Thanks for the contribution! IMO, this is a great feature to add and agreed with @osteotek that it brings us more in line with the stock firmware.

One problem I'm noticing is that the text is _slightly_ more pixelated/has more jagged edges when this is enabled. I see a distinct difference in text rendering between having this off and turned on. For me, it's more noticeable in styled text, like italics and bolded text, but I also see it in normal text. I'm using "Noto Sans" for my font if that's helpful for seeing what I'm seeing it.

Also, I did merge `master` into this while testing locally, so there may be some recent code change from `master` that causes this. It's worth pulling in the latest changes and testing this out some before we merge.

jdk2pq:
> Thanks for the contribution! IMO, this is a great feature to add and agreed with @osteotek that it brings us more in line with the stock firmware.
> 
> One problem I'm noticing is that the text is _slightly_ more pixelated/has more jagged edges when this is enabled. I see a distinct difference in text rendering between having this off and turned on. For me, it's more noticeable in styled text, like italics and bolded text, but I also see it in normal text. I'm using "Noto Sans" for my font if that's helpful for seeing what I'm seeing it.
> 
> Also, I did merge `master` into this while testing locally, so there may be some recent code change from `master` that causes this. It's worth pulling in the latest changes and testing this out some before we merge.

Actually, false alarm, ignore this! After testing this out more, I don't think this has anything to do with this PR.

However, interestingly enough, if I enable auto turn, then press the back button to disable it, the text looks _better_ to me (I believe it's the anti-aliasing) but _only for that page_ that I was on when I turn off auto turn. As soon as I change the page, it goes back to the slightly more pixelated/jagged rendering. I wonder if there's some second-pass on the anti-aliasing we're doing after disabling auto-turn that makes text look briefly better. 😅 I'm gonna dig into this more...

jdk2pq (dismissed):

jdk2pq:
Wow, this is both awesome and so funny to me. @GenesiaW I think you inadvertently uncovered a way of making anti-aliased text look even smoother on the e-reader! I even tested in 0.16.0 and 0.15.0 to see if maybe we had this before and some other code changes along the way reverted it, but I don't think that's the case. This seems new 😄 

From what I can gather (and to be clear I am NOT an expert in this at all, just trying to read the code as best I can and piece together what I'm seeing), toggling auto turn off does _indeed_ run a _second_ pass in calling the `renderScreen()` function in the `displayTaskLoop`. That's because `renderScreen` is called once while auto turn is enabled, then pressing the back/confirm buttons disables it and sets `updateRequired = true`. The `displayTaskLoop` then calls `renderScreen()` _again_ on the same page that's already on screen.

This _second_ `renderScreen` makes the text look much smoother and less pixelated/jagged. I've tried taking photos with my phone, and it's unfortunately very hard to show the difference in the photos, but let me see if I can get another camera to show the difference.

To force this behavior, I just simply duplicated the `renderScreen()` call in `displayTaskLoop`, and yep, it's a _very noticeable_ improvement.

Screenshot 2026-02-14 at 9 47 03 AM

I doubt this is the best way to get this improvement, but as a simple test case, it's worth trying it out! @crosspoint-reader/firmware-maintainers Can anyone else test this and verify this behavior (and that I'm not seeing things/don't need to go to an eye doctor 😄)?

jdk2pq:
And finally, here's the best visual comparison I can show

Single-pass:
single-pass

Double-pass:
double-pass

I notice it most in capital M's and lowercase g's on this page (look for "Mongo" in the text above). 

On the left edge of the capital M, in the "single-pass" image, I can see what appears to me as slight textual bleeding, as if there's a single dark line and then a slightly lighter, grayer, shorter line to the right of that line to create the left edge of the M. 

In the "double-pass" image, that left edge of the capital M is a single, perfectly straight line, no second line next to it.

Lowercase g's also have a smoother curve in the circle at the top of the character and the curve on the lower end.

ngxson:
I can confirm that I get the same behavior.

Indeed, you can archive the same behavior just by duplicating this code inside the render function:

```cpp
    // Render and copy to MSB buffer
    renderer.clearScreen(0x00);
    renderer.setRenderMode(GfxRenderer::GRAYSCALE_MSB);
    page->render(renderer, SETTINGS.getReaderFontId(), orientedMarginLeft, orientedMarginTop);
    renderer.copyGrayscaleMsbBuffers();

    // display grayscale part
    renderer.displayGrayBuffer();
    renderer.setRenderMode(GfxRenderer::BW);
```

Note that it doesn't work if I only duplicate this line: `renderer.displayGrayBuffer();`

I suspect that's because `copyGrayscaleMsbBuffers` actually send out RAM data to the display, that triggers an actual refresh on the next `displayGrayBuffer`. This applies the voltage, moves the particles a bit. Just my theory though, take as a grain of salt.

GenesiaW:
Set to draft while cleaning up merge conflict

src/activities/reader/EpubReaderActivity.cpp
```diff
@@ -163,6 +164,36 @@ void EpubReaderActivity::loop() {
...
+      } else {
+        RenderLock lock(*this);
+        nextPageNumber = 0;
+        currentSpineIndex++;
+        section.reset();
+      }
+      requestUpdate();
+      return;
+    }
+  }
```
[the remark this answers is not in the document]
GenesiaW: Decided to not reset section cache for StatusBar::None as it would appear more seamless when the auto turn is disabled

lib/I18n/translations/czech.yaml
```diff
@@ -315,3 +315,9 @@ STR_UPLOAD: "Nahrát"
 STR_BOOK_S_STYLE: "Styl knihy"
 STR_EMBEDDED_STYLE: "Vložený styl"
 STR_OPDS_SERVER_URL: "URL serveru OPDS"
+STR_AUTO_TURN_ENABLED: "Auto Turn Enabled: "
+STR_AUTO_TURN_PAGES_PER_MIN: "Auto Turn (Pages Per Minute)"
```
daveallie: Instead of including English translations, please remove these lines, they'll show up as warnings during compilation so we can tell what's in need of translation, and will fallback to English anyway.

src/activities/reader/EpubReaderMenuActivity.h
```diff
@@ -51,12 +52,15 @@ class EpubReaderMenuActivity final : public ActivityWithSubactivity {
   ButtonNavigator buttonNavigator;
   std::string title = "Reader Menu";
   uint8_t pendingOrientation = 0;
+  uint8_t selectedPageTurnOption = 0;
   const std::vector<StrId> orientationLabels = {StrId::STR_PORTRAIT, StrId::STR_LANDSCAPE_CW, StrId::STR_INVERTED,
                                                 StrId::STR_LANDSCAPE_CCW};
+  const std::vector<StrId> pageTurnLabels = {StrId::STR_STATE_OFF, StrId::STR_1, StrId::STR_3, StrId::STR_6,
+                                             StrId::STR_12};
```
daveallie: I don't think we should be translating numbers. Numeric values of other settings are not translated, and here we can follow suite. Let's remove the `STR_<NUM>` translations and drop in the direct number strings alongside `tr(StrId::STR_STATE_OFF)`

src/activities/reader/EpubReaderActivity.cpp
```diff
@@ -163,6 +164,36 @@ void EpubReaderActivity::loop() {
...
+        section->currentPage++;
+      } else {
+        RenderLock lock(*this);
+        nextPageNumber = 0;
+        currentSpineIndex++;
+        section.reset();
+      }
+      requestUpdate();
+      return;
+    }
```
daveallie, line 203: I don't think we should updating `lastPageTurnTime` until after the render completes, I managed to get myself stuck in autoturn.

For books with large chapters, the indexing and first page render can take longer than the autoturn duration, and it was immediately jumping to the next chapter as soon as the page loaded, triggering a new indexing, which was also longer than the duration, so I was just seeing the first page of each chapter fly by.

daveallie (changes requested):

src/activities/reader/EpubReaderActivity.cpp
GenesiaW: I have added a `updateLastPageTurnTime` flag which will be set to `false` on completion of render. At each loop iteration, `lastPageTurnTime` will be updated to `currentTime` as it waits for screen to finish rendering

GenesiaW:
Closed due to messy commit history. Opened #1219 as replacement

PR #726, opened 2026-02-06
feat(ux): invert BACK button behavior in reader activities

osteotek:
## Summary

* Inverts back button behaviour while reading - short press to go home, long press to open file browser

## Additional Context

* It seems counterintuitive that going into a book from home screen and pressing back doesn’t take you back to the home screen. With the recent books now displayed in the home view and a separate recents view, going directly to the file browser is less necessary.

### DIFF
--- a/src/activities/reader/EpubReaderActivity.cpp
+++ b/src/activities/reader/EpubReaderActivity.cpp
@@ -204,15 +204,15 @@ void EpubReaderActivity::loop() {
     xSemaphoreGive(renderingMutex);
   }
 
-  // Long press BACK (1s+) goes directly to home
+  // Long press BACK (1s+) goes to file selection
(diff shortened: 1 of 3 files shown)


andrewb1269 (approved):
Comment: the `if-checks` are now backwards. 

`getHeldTime() >= goHomeMs` --> goBack
`getHeldTime() < goHomeMs` --> goHome

I would expect `heldTime > goHome` time would result in going home instead of going back. This is a variable naming situation that I don't have a good solution for. I'm bringing this up as a discussion point and is non-blocking for merge.

This doesn't matter for functionality but I'm not sure if it will be confusing or not.

Eloren1:
Is it live in 1.0.0-rc+9e04eec? It doesn't work like that for me. Short click still opens file manager

daveallie:
Nah sorry, it'll make it into next release

Issue #719, opened 2026-02-06
docs: create a document for Github usernames and languages for translation


andrewb1269:
Create a new document to track Github usernames and the language the individual is comfortable with. Place this new document `translation-contacts.md` in the `docs` folder. 

The full volunteer discussion can be found [here](https://github.com/crosspoint-reader/crosspoint-reader/discussions/409).

Something like:

-----------------

## French
- [Spigaw](https://github.com/Spigaw)

## German
- [DavidOrtmann](https://github.com/DavidOrtmann)

## Italian
- [fragolinux](https://github.com/fragolinux)

## Russian
- [madebyKir](https://github.com/madebyKir)

## Spanish
- [yeyeto2788](https://github.com/yeyeto2788)
- [Skrzakk](https://github.com/Skrzakk)

## Swedish
- [dawiik](https://github.com/dawiik)

DavidOrtmann:
I can do German!

fragolinux:
let us know when a proper gettext setup (or the likes) is available, so we can start

madebyKir:
I have prepared a Russian translation for myself. And I made it possible to choose the language between EN/RU

<details>
  <summary>Click to see the image</summary>

  <p>
    <img src="https://github.com/user-attachments/assets/1e136162-6b8f-4267-9045-0858f4ff61b4" />
    <img src="https://github.com/user-attachments/assets/43ad2715-5a3b-447a-ba10-f2a5dafcc4ef" />
    <img src="https://github.com/user-attachments/assets/3e48d0f7-0f9b-45f0-9f6a-cedc2335e817" />
  </p>
</details>

yeyeto2788:
@madebyKir did you make a PR for it? Is there a way we can adapt or start translating those words that you've changed into other languages?

@andrewb1269 maybe in a separate branch we can start at least listing the lines that need to be translated?

andrewb1269:
PR opened in #792 that will close this issue

madebyKir:
> [@madebyKir](https://github.com/madebyKir) did you make a PR for it? Is there a way we can adapt or start translating those words that you've changed into other languages?
> 
> [@andrewb1269](https://github.com/andrewb1269) maybe in a separate branch we can start at least listing the lines that need to be translated?


I didn't do PR. I did everything locally. Through the transfer table. (i18n). If you send translations into the required languages in the tables below, I will be able to add them. But this applies for version 0.16.0, I didn't like version 1.0. 

[...]

PR #728, opened 2026-02-06
feat: User-Interface I18n System

Uri-Tauber:
## Summary

This PR introduces Internationalization (i18n) support, enabling users to switch the UI language dynamically.

- Core Logic: Added I18n class (`lib/I18n/I18n.h/cpp`) to manage language state and string retrieval.

- Data Structures:

  - `lib/I18n/I18nStrings.h/cpp`: Static string arrays for each supported language.
  - `lib/I18n/I18nKeys.h`: Enum definitions for type-safe string access.
  - `lib/I18n/translations.csv`: single source of truth. 

- Documentation: Added `docs/i18n.md` detailing the workflow for developers and translators.

- New Settings activity: `src/activities/settings/LanguageSelectActivity.h/cpp`

## Additional Context

This implementation (building on concepts from #505) prioritizes performance and memory efficiency.

The core approach is to store all localized strings for each language in dedicated arrays and access them via enums. This provides O(1) access with zero runtime overhead, and avoids the heap allocations, hashing, and collision handling required by `std::map` or `std::unordered_map`.

The main trade-off is that enums and string arrays must remain perfectly synchronized—any mismatch would result in incorrect strings being displayed in the UI.

To eliminate this risk, I added a Python script that automatically generates `I18nStrings.h/.cpp` and `I18nKeys.h` from a CSV file, which will serve as the single source of truth for all translations. The full design and workflow are documented in `docs/i18n.md`.

### Next Steps

- [x] Python script `generate_i18n.py` to auto-generate C++ files from CSV
- [x] Populate translations.csv with initial translations.

Currently available translations: English, Español, Français, Deutsch, Čeština, Português (Brasil), Русский, Svenska.
Thanks, community!

**Status:** EDIT: ready to be merged.

As a proof of concept, the SPANISH strings currently mirror the English ones, but are fully uppercased. 

---

### AI Usage

I used AI for the black work of replacing strings with I18n references across the project, and for generating the documentation. EDIT: also some help with merging changes from master.


Uri-Tauber:
Could someone with write access please add `feat:` to the PR title?

CaptainFrito (approved):
Very smart implementation 👍🏻

CaptainFrito:
Hit me up when we can start translating 🙂

brbla:
Thumbs up! Now I see we had a very similar idea. If you want to try mine solution, I have a [firmware.bin](https://github.com/brbla/crosspoint-reader/releases/tag/v1.0.0-i18n-preview) there for you to try and some pictures of my reader with it.

Uri-Tauber:
## I18n Engine Ready!
The Python automation is now working. This PR establishes the core infrastructure for multi-language support.

### How to Translate
No manual synchronization is needed. All changes happen in the spreadsheet:

Edit: Open lib/I18n/translations.csv in Excel or LibreOffice.

Translate: Add your text to the corresponding language column.

Sync: Run the script to update the code:

```
python3 scripts/gen_i18n.py lib/I18n/translations.csv lib/I18n/
```

Details are available in docs/i18n.md.

### Next Steps
I suggest we merge this PR now to establish the workflow (@CaptainFrito @daveallie). Afterwards, Translators can open individual PRs for their specific languages. Until this PR will be merged, feel free to attach your version of the CSV here, and I'll update the PR.

lib/I18n/I18nKeys.h
```diff
@@ -0,0 +1,362 @@
...
+extern const char* const LANGUAGE_NAMES[];
+
+// Character sets for each language (defined in I18nStrings.cpp)
+extern const char* const CHARACTER_SETS[];
+
+// String IDs
+enum class StrId : uint16_t {
+  CROSSPOINT,
+  BOOTING,
+  SLEEPING,
```
ngxson: nits, but I would prefer defining those as macro, something like:

```cpp
#define TEXT_CROSSPOINT i18n(0)
#define TEXT_BOOTING i18n(1)
#define TEXT_SLEEPING i18n(2)
```

so that the downstream code will look a bit more intuitive, like:

```cpp
drawCenteredText(TEXT_CROSSPOINT, ...);
```

src/activities/settings/SettingsActivity.h
```diff
@@ -15,10 +16,10 @@ class CrossPointSettings;
 enum class SettingType { TOGGLE, ENUM, ACTION, VALUE, STRING };
 
 struct SettingInfo {
-  const char* name;
+  StrId nameId;
```
ngxson, line 27: I think the `const char*` is still fine though (?) it's just a pointer to string so we never actually copy the string

lib/I18n/translations.csv
ngxson, line 1: IMO a csv may not be a very good idea. For example, if 2 translators open 2 PRs for the 2 different languages, each PR will still modify the same line, thus making diff harder to merge.

Maybe something like JSON or YAML can be a better choice?

```yaml
en:
  CROSSPOINT: CrossPoint
  BOOTING: Booting
```

Or:

```yaml
CROSSPOINT:
  en: Crosspoint
  fr: ...
```

Uri-Tauber: @ngxson Handling merge conflicts in the CSV is straightforward via a quick column copy-paste in LibreOffice or Excel. I chose CSV because it’s much more accessible for non-technical contributors compared to JSON or YAML, which usually require a specialized editors (like Dadroit) to avoid syntax errors.

src/activities/settings/SettingsActivity.h
Uri-Tauber: @ngxson I might be missing something, but wouldn't `const char*` prevent the setting names from being translated? With `StrId`, the name gets resolved to the current language at display time.
For example: looking at line 277 in SettingsActivity.cpp, we call `I18N.get(settings[index].nameId)` which resolves the name to the current language at display time. With a `const char*`, the string would be fixed at construction time and wouldn't update when the user changes their language setting.

lib/I18n/translations.csv
DavidOrtmann: Here's my first stab at German. (German is notorious for having long words, so I'm happy to check this on my device later and shrink strings further down.)

[translations_de2.csv](https://github.com/user-attachments/files/25193695/translations_de2.csv)

lib/I18n/I18nKeys.h
Uri-Tauber: I appreciate the suggestions! 
The current `i18n(KEY)` pattern matches how other modern i18n libraries work (Qt's `tr()` for example), so it's a well-established pattern. That said, I understand it's a matter of preference. 

If there's strong preference for the TEXT_ pattern from other maintainers, I'm happy to adjust.

lib/I18n/translations.csv
Uri-Tauber: Thanks @DavidOrtmann!
I will wait with updating the PR until you finished the shrinking.

DavidOrtmann: Is there a way to check the strings on-device yet?

ngxson: > Handling merge conflicts in the CSV is straightforward via a quick column copy-paste in LibreOffice or Excel.

What I mean was the `git merge` conflict style, I'm wondering how non-git program can handle these

For example, 2 translators edit the same line:

```
<<<<<<< HEAD
Open,Open
=======
Open,Ouvrir
>>>>>>> branch-a
```

> I chose CSV because it’s much more accessible for non-technical contributors compared to JSON or YAML, which usually require a specialized editors (like Dadroit) to avoid syntax errors.

I would say YAML is pretty straight-forward for non-tech people too. I imagine most translators here know what it is, given they already know how to use git.

src/activities/settings/SettingsActivity.h
ngxson: hmm right, sorry I didn't notice that `SettingInfo` is always statically allocated at compile time. It's fine to keep StrId here then

lib/I18n/I18nKeys.h
ngxson: I think there are 2 problems with the `i18n(KEY)` though, firstly it may conflict with other macro or variable naming. Something like `AUTH_FAILED` can likely introduce conflicts with another enum or`#define` having the same name

I know namespace enum like `i18n(String::LOADING)` can also be the solution, but still, it's more verbose than something like `T_LOADING` or `TEXT_LOADING`

Secondly, other frameworks uses shorter naming for the text function, like `t()`, `tr()`, `_e()` for the translation function to make DX a bit better. `i18n` is a bit too verbose I think.

lib/I18n/translations.csv
Uri-Tauber: @DavidOrtmann Which program did you use to edit the CSV?
LibreOffice wraps cells with commas in quotes, but the tool you used didn’t, so one cell is now being interpreted as several.

image


@ngxson This is a strong point in your approach favor.

DavidOrtmann: > @DavidOrtmann Which program did you use to edit the CSV? LibreOffice wraps cells with commas in quotes, but the tool you used didn’t, so one cell is now being interpreted as several.

I used Excel. Here's another export. I think it looks identical to [your own file](https://github.com/Uri-Tauber/crosspoint-reader/blob/reimplement-i18n/lib/I18n/translations.csv?plain=1), which does not have quotes.

[translations_de3.csv](https://github.com/user-attachments/files/25195557/translations_de3.csv)

Uri-Tauber: @DavidOrtmann nope. exactly the same. look, for example, at lines 68 and 302 with text editor. you would notice that each cell is warped in quotes in my version, and not in yours. 
If you could figure out how to export the csv with quotes, it would be useful.

lib/I18n/I18nKeys.h
Uri-Tauber: @ngxson You’re right about the second point. Renaming it to `tr()` would simplify things and make it more clear to developers, especially those who have worked with Qt.

Regarding the first point: Just to clarify, `i18n()` is actually already a macro 
(defined as `#define i18n(id) I18n::getInstance().get(StrId::id)` in I18n.h).

This means when we write:
```cpp
drawCenteredText(i18n(CROSSPOINT), ...);
```

The preprocessor expands it to:
```cpp
drawCenteredText(I18n::getInstance().get(StrId::CROSSPOINT), ...);
```

ngxson: I don't mean it's a problem with the macro, but the enum is the problem.

Generic enum naming like the AUTH_FAILED example I mentioned above can always conflict with any piece of code that declares another enum or macro having the same AUTH_FAILED value, especially C code where they cannot use scope.

At least it is a real problem that I experienced a real production code, so a common practice is to prefix the enum value no matter what.

Ref: https://www.google.com/search?q=enum%20conflicts%20with%20a%20previous%20declaration

lib/I18n/translations.csv
DavidOrtmann: Sure thing, will do.

lib/I18n/I18nKeys.h
ngxson: Something like this may break the logic:

#define AUTH_FAILED -1 // maybe defined inside a library header, like wifi.h for example
...
i18n(AUTH_FAILED) // now expanded to i18n(-1)

lib/I18n/translations.csv
Uri-Tauber: @DavidOrtmann I attempted to reproduce the issue by saving and exporting from Excel in various ways, but the quotes were always preserved (which is the correct behavior). Could you please walk me through the exact steps you took to generate the file?

Uri-Tauber: @ngxson would you be able to suggest a YAML structure that preserves the two main advantages of the current CSV approach:
 – adding a new string by adding a single row
 – adding a new language by adding a single column

From my understanding of YAML, it tends to favor one of these at the expense of the other. For example, if strings are grouped by language, adding a new string would require updating multiple locations in the file, which feels more error-prone. I’d be interested to hear if you see a clean way around this.

lib/I18n/I18nKeys.h
Uri-Tauber: @ngxson You're absolutely right - I tested it and the macro expansion does happen before the scope resolution.
image

I'm thinking of adding a prefix to the enum values. What prefix would you suggest?

Options:
- `STR_` (short, common pattern)
- `I18N_` (clear purpose)
- `TEXT_` (matches your original suggestion)

CaptainFrito:
Anybody working on translations? I'm doing le French

CaptainFrito:
[translations.csv](https://github.com/user-attachments/files/25212381/translations.csv)


Here is a first stab at the french version. It didn't export with quotes except for the strings that have a coma, is that okay?
I'm gonna checkout your branch and put it in the code now to see what it looks like. The settings are probably going to look horrible with all these long ass strings, but at least it's not German!

CaptainFrito:
IMG_8037 Medium

HA! Your move, Germans!

Uri-Tauber:
> [translations.csv](https://github.com/user-attachments/files/25212381/translations.csv)
> 
> Here is a first stab at the french version. It didn't export with quotes except for the strings that have a coma, is that okay? I'm gonna checkout your branch and put it in the code now to see what it looks like. The settings are probably going to look horrible with all these long ass strings, but at least it's not German!

@CaptainFrito Looks great.

lib/I18n/I18nKeys.h
ngxson: It think either `STR_` or `TEXT_` sounds good. I don't usually add number to variable name (just personal preference though)

Even better, since the discussion above moves `i18n()` to something shorter like `tr()`, maybe we can call the whole thing `TR_BACK` (which expands to something like `i18n(STR_ID_BACK)` ?)

lib/I18n/translations.csv
ngxson: > adding a new string would require updating multiple locations in the file, which feels more error-prone.

IMO this is expected and I can't think of how it can be error-prone. We can tell the translators to follow a specific order, for example always `en,fr,de,...`, then each of the string translation will be like:

```yaml
CROSSPOINT:
  en: Crosspoint
  fr: ...
  de: ...
```

It's true that adding a new language will add a new line everywhere in the file, but I bet adding a new language is a rare case.

The case where 2 or more translators try to edit the same string in different languages will happen much more often. The YAML approach set out to make `git merge` to work as automatically as possible in such case.

ariel-lindemann:
> Anybody working on translations? I'm doing le French

translation work is being discussed here #719

madebyKir:
> [translations.csv](https://github.com/user-attachments/files/25212381/translations.csv)
> 
> Here is a first stab at the french version. It didn't export with quotes except for the strings that have a coma, is that okay? I'm gonna checkout your branch and put it in the code now to see what it looks like. The settings are probably going to look horrible with all these long ass strings, but at least it's not German!

Do I understand correctly that to add the Russian language, I must add the Russian translation after the comma in each line? 

P.S. Sorry for the stupid question, I'm not a programmer, I'm just a person with a great desire to participate in the project

Uri-Tauber:
> Do I understand correctly that to add the Russian language, I must add the Russian translation after the comma in each line?
> 
> P.S. Sorry for the stupid question, I'm not a programmer, I'm just a person with a great desire to participate in the project

Not a stupid question at all.

You might find it easier to work with CSV files in Excel or LibreOffice. Essentially, yes—just add a comma and your translation at the end of each line. If any value contains a comma, wrap it in quotes so it isn’t interpreted as multiple cells.

In Excel or LibreOffice, this is usually handled automatically when you add a new column.

lib/I18n/translations.csv
DavidOrtmann: > @DavidOrtmann I attempted to reproduce the issue by saving and exporting from Excel in various ways, but the quotes were always preserved (which is the correct behavior). Could you please walk me through the exact steps you took to generate the file?

Can you check this file? I've learned that because of German number delimiters (`,` instead of `.`), German Excel defaults to `;` as a CSV delimiter, and you have to tinker with system settings to trick it (and it didn't even work for me then). I now used edit-csv.net.

[translations_de_new.csv](https://github.com/user-attachments/files/25222326/translations_de_new.csv)

Uri-Tauber:
* Switch the i18n() macro to shorter tr(), per @ngxson suggestion.
* Prefix all new translation keys with SRT_ to avoid clashes with existing keys, per @ngxson suggestion.
* Add German and French translations. Credit to @CaptainFrito and @DavidOrtmann!

french
german

Firmware file for testing:
[firmware.bin.zip](https://github.com/user-attachments/files/25227698/firmware.bin.zip)

lib/I18n/translations.csv
Uri-Tauber: @DavidOrtmann It works fine. I’ve uploaded the compiled firmware for you to test. Several strings need to be shortened.

lib/I18n/I18nKeys.h
Uri-Tauber: @ngxson I Updated the PR according to your suggestions.

lib/I18n/translations.csv
DavidOrtmann: Thank you for your work! I'm glad I aggressively shorted strings before uploading (I added a column that showed length difference), will now work through anything that is too long still.

DavidOrtmann: By the way, we might need a different string for "Inverted", because this is something different in German (and maybe other languages) if it relates to the device's orientation or colours of an image.

I went through all menus (I hope) and adjusted some lengths in this version of the translation:

[translations_de_1102.csv](https://github.com/user-attachments/files/25229465/translations_de_1102.csv)

Uri-Tauber:
Updated the German strings provided by @DavidOrtmann.
[firmware.bin.zip](https://github.com/user-attachments/files/25231179/firmware.bin.zip)

lib/I18n/I18nKeys.h
ngxson: Looks better now, just one more concern is that maybe writing `tr(STR_...)` potentially making the whole `tr(STR_` as a prefix? Just look a bit verbose to me, but not a very big problem.

My suggestion is:

```cpp
#define TR_BACK tr(STR_BACK)
#define TR_CONFIRM tr(STR_CONFIRM)
...

// then use it: drawText(TR_BACK, ...)
```

These macros can be autogenerated on the same `I18nKeys.h` header

brbla:
If it helps, I have prepared Czech text strings.
[cz-translations-updated-libreoffice.csv.zip](https://github.com/user-attachments/files/25233499/cz-translations-updated-libreoffice.csv.zip)

lib/I18n/I18nKeys.h
Uri-Tauber: I work a lot with Qt, so, in my eyes, the `tr(STR_CONFIRM)` style feels more intuitive than `TR_CONFIRM` macro. If that’s acceptable, I’d like to keep it as is.

Uri-Tauber:
Add Czech strings. Credit to @brbla 
[firmware.bin.zip](https://github.com/user-attachments/files/25234479/firmware.bin.zip)

yagofarias:
[cz-translations-ptbr.csv](https://github.com/user-attachments/files/25235687/cz-translations-ptbr.csv)
My contribution for Portuguese Brazil

madebyKir:
I have added a Russian translation, I would be grateful if you add it to the tests and in a future release.
Update line 285
[!translations_ru.csv](https://github.com/user-attachments/files/25239170/translations_ru.csv)

Uri-Tauber:
Add Portuguese-Brazil and Russian strings. Credit to @yagofarias and @madebyKir.

@madebyKir please note, your CSV have the same commas problem like @DavidOrtmann earlier. I fixed them by hand, but please use his solution if you want to update the strings again.

[firmware.bin.zip](https://github.com/user-attachments/files/25240885/firmware.bin.zip)

lib/I18n/I18nKeys.h
ngxson: I usually prefer brevity for variable names that use very frequently, but no strong opinion in this case. Maybe asking the team @crosspoint-reader/firmware-maintainers as this seems to be an important decision.

yagofarias:
[cz-translations-ptbr+ru.csv](https://github.com/user-attachments/files/25241590/cz-translations-ptbr%2Bru.csv)
@Uri-Tauber thanks for the implementation! i changed a couple of strings and updated the file. Also included russian from @madebyKir file.
Now i think portuguese is pretty much done

Uri-Tauber:
Add Spanish and some fixes for Portuguese-Brazil. Credit to @yagofarias and @yeyeto2788!
[firmware.bin (1).zip](https://github.com/user-attachments/files/25242181/firmware.bin.1.zip)

madebyKir:
> Add Spanish and some fixes for Portuguese-Brazil. Credit to @yagofarias and @yeyeto2788! [firmware.bin (1).zip](https://github.com/user-attachments/files/25242181/firmware.bin.1.zip)

I tested the Russian language in a classic theme, and tried to optimize the length of the words to avoid overlapping letters. 

<details>
  <summary>Update lines</summary>
96
150
164-166
223-226
228
258
261
</details>
<details>
  <summary>Click to see the image</summary>

  <p>
    
  </p>
</details>

[!cz-translations-ptbr+ru.csv](https://github.com/user-attachments/files/25244035/cz-translations-ptbr%2Bru.csv)

yeyeto2788:
> > Add Spanish and some fixes for Portuguese-Brazil. Credit to @yagofarias and @yeyeto2788! [firmware.bin (1).zip](https://github.com/user-attachments/files/25242181/firmware.bin.1.zip)
> 
> I tested the Russian language in a classic theme, and tried to optimize the length of the words to avoid overlapping letters.
> Update lines
> 96 150 164-166 223-226 228 258 261
> Click to see the image
> 
> 
> 
> [!cz-translations-ptbr+ru.csv](https://github.com/user-attachments/files/25244035/cz-translations-ptbr%2Bru.csv)

That seems to be happening in Spanish as well depending on the size of the text. I guess we might need to rework a bit the phrases on each language or come up with a different way of showing the content. Maybe truncating the value until it gets selected?

Even if it is not looking as polished as it should this is actually a very useful feature and worth the effort on adding it to the firmware.

@Uri-Tauber I also notice that there are some word that might be duplicated, not sure if we could potentialy reuse some of them on when rendering the text something like:

|Key|Value|Result|
|---|---|---|
|STR_HOME|Home|Home|
|STR_GO_HOME|Long press to go to %STR_HOME%|Long press to go to Home


Something similar that is done already (`STR_PAGE_OVERALL_FORMAT` > `Page %d, %.2f%% overall`) but not sure if possible to apply what I mentioned above. I know in this case is just formatting the type of value but just a suggestion. Maybe on a separate PR.

Uri-Tauber:
@yeyeto2788 Let's keep it for a seperate PR. 
I'm going to accept @ngxson suggestion and migrate from CSV to YAML.
Apart from that, I don't want any more big changes at this point. I’d like to get this merged quickly—Adding extra features would extend the time needed for review.

DavidOrtmann:
> @Uri-Tauber I also notice that there are some word that might be duplicated, not sure if we could potentialy reuse some of them on when rendering the text something like:
> Key 	Value 	Result
> STR_HOME 	Home 	Home
> STR_GO_HOME 	Long press to go to %STR_HOME% 	Long press to go to Home

Practically, I would respectfully argue against that, because depending on languages, the wording or even just declination might change the word from the original to something different. But maybe I'm missing a big benefit here.

yagofarias:
@Uri-Tauber just reviewed the last version of Portuguese (Brazil) and it's ok for me! Congrats on the initiative!

If anyone else that speaks Portuguese and wants to review as well fell free

dawiik:
Have no clue what I am doing really, Github is a bit out of my world (Only translated Min-browser here before) and I am more of a Weblatian. But...

I added swedish translation to this [translations.csv](https://github.com/user-attachments/files/25273267/translations.csv)

lib/I18n/translations.csv
Uri-Tauber: @ngxson I started migrating to YAML, and while doing that I had an idea that might address the merging issue you mentioned, while keeping the workflow simple.

What if we rotate the CSV by 90 degrees — so languages become rows and strings become columns? That way, adding a new language or a new string would still be straightforward.

It would also allow translators without programming experience to contribute in a familiar spreadsheet-style format, without feeling intimidated by YAML.

If two translators submit PRs for different languages, the merge should happen automatically since each language would be a separate row. The only time manual conflict resolution would be needed is when two people work on the same language, which should be much less common.

What do you think?

ngxson: > What if we rotate the CSV by 90 degrees — so languages become rows and strings become columns? That way, adding a new language or a new string would still be straightforward.

I imagine that won't be very readable on something like excel. Also, if 2 people change 2 different texts in the same language, git diff will still fail

> It would also allow translators without programming experience to contribute in a familiar spreadsheet-style format, without feeling intimidated by YAML.

Honestly I don't see how yaml can be difficult to understand. It looks almost like a todo list and I'm pretty sure that it's intuitive enough that translator can also ask AI for help if needed.

I suggest we do this way: instead of assuming which one is easier, let's try making 2 demo, one yaml, one csv, and ask translators for opinions ?

DavidOrtmann: I would like to add how TRMNL does its localisation efforts, and it's working very well, imo: https://github.com/usetrmnl/trmnl-i18n/blob/main/lib/trmnl/i18n/locales/web_ui/en.yml

Uri-Tauber:
I migrated the translation system from CSV to YAML in order to address the potential issues that @ngxson pointed out.

To keep the files manageable, I chose to use a separate file per language. I took inspiration from the project that @DavidOrtmann referenced.
In the future, we might also consider grouping strings into categories, which could help translators better understand the context of each string.

I updated the Python script to make the workflow as friendly as possible for translators, and I’m quite happy with how it turned out.

Currently:
* To add a new language, you simply create a new file in `lib/I18N/translations` and follow the existing syntax (the easiest way is to copy `english.yaml` and adapt it).
* To add a new string, you only need to add it to `english.yaml`. The script will automatically generate the corresponding entries for the other languages.

More details are available in the updated `docs/i18n.md`.

I believe the PR is ready for review.

Uri-Tauber:
Hi @dawiik,
I’m currently having trouble downloading your CSV from here. It seems to be an issue on my side — not related to anything you did. I’ll probably be able to resolve it by early next week.

If it’s not too much trouble, would you mind sending the CSV to my email in the meantime?

dawiik:
Oh I would not be surprised if the issue is on my side either, you know. Being somewhat of an Github-noob, every step I take is a bit of a challenge. 

But an email can be arranged. If I only could seem to find your emailadress somewhere. d.wiik at outlook com if you want to reach out.

> Hi @dawiik, I’m currently having trouble downloading your CSV from here. It seems to be an issue on my side — not related to anything you did. I’ll probably be able to resolve it by early next week.
> 
> If it’s not too much trouble, would you mind sending the CSV to my email in the meantime?

Uri-Tauber:
Add Swedish and some fixes for Russian. Credit to @madebyKir and @dawiik!

[firmware.bin.zip](https://github.com/user-attachments/files/25291127/firmware.bin.zip)

Please be aware of the updated translation system and begin working with the YAML files.
Thank you, everyone!

madebyKir:
> * To add a new language, you simply create a new file in lib/I18N/translations and follow the existing syntax (the easiest way is to copy english.yaml and adapt it).

Do I need to send a prepared russian.yaml file? Or is it unnecessary now? 

> * In the future, we might also consider grouping strings into categories, which could help translators better understand the context of each string. 

I also noticed that ~39-40 characters are placed in one line, the context would help to avoid layering the text.

Uri-Tauber:
@madebyKir not unless you want to update the strings again, since I already integrated your fixes. :)

Feel free to check `lib/I18n/translations/russia.yaml`

scripts/gen_i18n.py
ngxson, line 1: Optional, but you can also include this as an `extra_scripts` in `platformio.ini`, it will run every time the project is built

(Just make sure you provide the default value for `<translations_dir> <output_dir>`)

src/main.cpp
```diff
@@ -299,11 +300,13 @@ void setup() {
     LOG_ERR("MAIN", "SD card initialization failed");
     setupDisplayAndFonts();
     exitActivity();
-    enterNewActivity(new FullScreenMessageActivity(renderer, mappedInputManager, "SD card error", EpdFontFamily::BOLD));
+    enterNewActivity(
+        new FullScreenMessageActivity(renderer, mappedInputManager, tr(STR_SD_CARD_ERROR), EpdFontFamily::BOLD));
```
ngxson: Just note that i18n is not initialized at this point. Not a big problem, but it's probably an undefined behavior

fangguangyang:
Simplified Chinese translation (简体中文) @Uri-Tauber 

[chinese_simplified.yaml](https://github.com/user-attachments/files/25298407/chinese_simplified.yaml)

Uri-Tauber:
> Simplified Chinese translation (简体中文) @Uri-Tauber
> 
> [chinese_simplified.yaml](https://github.com/user-attachments/files/25298407/chinese_simplified.yaml)

Thank you for your effort!
Unfortunately, At the moment the bundled fonts only cover Latin and Cyrillic characters.
That's why I haven’t added Hebrew (my native language) yet.
Once one of the custom‑font PRs is merged, we can add your chinese_simplified.

src/main.cpp
Uri-Tauber: I added a call to `I18N.loadSettings(); `just before this line to resolve the UB.
In theory, I could simply revert to “SD card error” and get the same result, because when the SD card is missing `loadSettings()` defaults to English.
However, keeping the handling unified is preferable.

scripts/gen_i18n.py
Uri-Tauber: Adding the script `to platformio.ini` is irrelevant for most developers; but the slight delay can be annoying.
Including it in the CI workflow, however, would be valuable—though I agree it should be addressed in a separate PR.

madebyKir:
I've made the most of all the places I've seen. But I also found a lot of bugs that may not be related to the translation. 

[russia.yaml](https://github.com/user-attachments/files/25318179/russia.yaml)

<details>
  <summary>Bugs that I saw</summary>






There is also a problem with displaying the % charge, it jumps up and down. Shows 97% values, then 95%, then 97% words.

Can't be translated "Continue Reading" 


  <p></p>
</details>

p.s. Thanks a lot @Uri-Tauber, you're pumping my skill and forcing me to develop programming skills.

Uri-Tauber:
@madebyKir I updated the Russian strings you shared, and fixed the "Continue Reading" label.
Could you elaborate on the other bugs you found? I don’t read Cyrillic, so the photos aren’t very helpful.

[firmware.bin.zip](https://github.com/user-attachments/files/25318390/firmware.bin.zip)

madebyKir:
> Could you elaborate on the other bugs you found?

When switching between menu items, translations are not displayed correctly or are not fully displayed. 

<details>
  <summary>Example 1.</summary>

English: "Browse Files"
Russian (as it should be): "Обзор файлов"
Russian (bug in my image):  "Обзор ф"

<details>
  <summary>Pic.</summary>



  <p></p>
</details>

  <p></p>
</details>


<details>
  <summary>Example 2.</summary>

English: "File Transfer"
Russian (as it should be): "Передача файлов"
Russian (bug in my image):  "Передача ф?ds»ов"

<details>
  <summary>Pic.</summary>



  <p></p>
</details>

  <p></p>

</details>


<details>
  <summary>Correct rendering</summary>



  <p></p>
</details>

Uri-Tauber:
@madebyKir Interesting. I got the same behavior, but only on Russian. For all other languages the menu options stay the same. There is probably a difference between how latin and cyrillic characters are handled.
I will dig into this.

DavidOrtmann:
> @madebyKir Interesting. I got the same behavior, but only on Russian. For all other languages the menu options stay the same. There is probably a difference between how latin and cyrillic characters are handled. I will dig into this.

I saw the same when using German Umlaute (ü) for File Transfer. I thought it was because the string was too long, I've since shortened it, and it now doesn't have an Umlaut there.

Uri-Tauber:
@madebyKir Fixed!
Let me know if you find other bugs.

[firmware.bin.zip](https://github.com/user-attachments/files/25318715/firmware.bin.zip)


<details>
<summary>If anyone interested to know what was the problem</summary>

The line:
   `const char* label = buttonLabel(i).c_str();`
  `buttonLabel(i)` returns a `std::string` by value (a temporary). Calling `.c_str()` gives a pointer into that temporary, which is destroyed at the semicolon, leaving label as a dangling pointer. This is undefined behavior.
  Why it only affects Russian (and would affect other non-Latin languages): English strings are short enough (≤~22 bytes) to fit in the `std::string` Small String Optimization (SSO) buffer, so the data lives on the stack and tends to survive briefly after the temporary dies. Russian strings in UTF-8 are ~2x longer in bytes (e.g., "Передача файлов" = 29 bytes vs "File Transfer" = 13 bytes), exceeding the SSO threshold — the string is heap-allocated, freed immediately when the temporary dies, and the memory gets corrupted/reused, causing garbled or truncated display.
  Fix
  Store the `std::string` in a named local variable so it lives until the end of the loop iteration:
   `std::string labelStr = buttonLabel(i);`
   `const char* label = labelStr.c_str();`
Credit to Claude.

</details>

madebyKir:
Я проверил 

> @madebyKir Fixed! Let me know if you find other bugs.
> 
> [firmware.bin.zip](https://github.com/user-attachments/files/25318715/firmware.bin.zip)

I checked this issue, and I didn't see any problems with translation or incorrect display anymore. 

There is only a bug with the display of the battery charge level. 

It can show 99%-100%-99%, while the charge level icon can be seen shifting to the left. 

<details>
  <summary>% battery</summary>





  <p></p>
</details>

CaptainFrito:
> Я проверил
> 
> > @madebyKir Fixed! Let me know if you find other bugs.
> > [firmware.bin.zip](https://github.com/user-attachments/files/25318715/firmware.bin.zip)
> 
> I checked this issue, and I didn't see any problems with translation or incorrect display anymore.
> 
> There is only a bug with the display of the battery charge level.
> 
> It can show 99%-100%-99%, while the charge level icon can be seen shifting to the left.
> 

Hi! This bug is fixed in https://github.com/crosspoint-reader/crosspoint-reader/pull/732

src/main.cpp
ngxson: I would prefer keeping the original message "SD card error" because the `I18N.loadSettings` doesn't make sense in this case.

Also, we should do minimal of stuff in this code branch because it's usually used as a safe mode, for example in case the device stuck on bootloop, removing sdcard will stop it. See: https://github.com/crosspoint-reader/crosspoint-reader/issues/793

Uri-Tauber: @ngxson No problem, updated.

ngxson:
looking good! the change is quite big, so I would like to ask for @coderabbitai for an additional review

Uri-Tauber:
Fixed Format String Mismatches and Enum Naming suggested by @coderabbitai. 
The Thread Safety is irrelevant for ESP32C3, which have only one CPU.

ngxson:
I agree that thread safety is not a really big problem here because it's only changed in settings.

However, keep in mind that we can still have thread safety issue even with single core CPU. freeRTOS tasks can always be interrupted mid-execution and being swapped out.

A very easy-to-understand video on how freeRTOS manage multiple tasks: https://www.youtube.com/watch?v=95yUbClyf3E&list=PLEBQazB0HUyQ4hAPU1cJED6t3DU0h34bz&index=3

scripts/gen_i18n.py
daveallie: I do think that running this as part of build and ignoring the generated files is worth it, the generation script only takes about 50ms to run, we already depend on python for the HTML generation, it'll save committing the 14k lines of generated translations.

Uri-Tauber: @daveallie No problems, but I think we should commit `I18nKeys.h` and `I18nStrings.h` while adding `I18nStrings.cpp` (the 14k‑line file) to `.gitignore`. Otherwise VS Code throws a lot of “cannot open source file” errors.

Uri-Tauber:
Add `gen_i18n.py` to `platformio.ini` and `I18nStrings.cpp` to `.gitignore` per @daveallie suggestion.
Fixed some merge conflicts.

@ngxson Could you review `/src/activities/settings/LanguageSelectActivity.h/cpp` and verify that the rendering logic aligns with your refactoring?

src/activities/settings/LanguageSelectActivity.h
```diff
@@ -0,0 +1,33 @@
...
+
+#include "../ActivityWithSubactivity.h"
+#include "components/UITheme.h"
+
+class MappedInputManager;
+
+/**
+ * Activity for selecting UI language
+ */
+class LanguageSelectActivity final : public ActivityWithSubactivity {
```
ngxson:
```suggestion
class LanguageSelectActivity final : public Activity {
```

If I understand correctly, we never have a sub-activity here

daveallie (approved):
Looking good to me, but would love another set of eyes on it

src/activities/settings/LanguageSelectActivity.h
Uri-Tauber: Good catch! 
I copied some code from other settings activities, and it looks like it sneaked in.

Fixed.

lib/I18n/translations/russia.yaml
```diff
@@ -0,0 +1,317 @@
...
+STR_MAC_ADDRESS: "MAC-адрес:"
+STR_CHECKING_WIFI: "Проверка Wi-Fi..."
+STR_ENTER_WIFI_PASSWORD: "Введите пароль Wi-Fi"
+STR_ENTER_TEXT: "Введите текст"
+STR_TO_PREFIX: "к "
+STR_CALIBRE_DISCOVERING: "Поиск Calibre..."
+STR_CALIBRE_CONNECTING_TO: "Подключение к "
+STR_CALIBRE_CONNECTED_TO: "Подключено к "
+STR_CALIBRE_WAITING_COMMANDS: "Ожидание команд..."
+STR_CONNECTION_FAILED_RETRYING: "(Ошибка подключения"
```
mrtnvgr, line 70:
```suggestion
STR_CONNECTION_FAILED_RETRYING: "(Ошибка подключения)"
```

lib/I18n/translations/russia.yaml
```diff
@@ -0,0 +1,317 @@
...
+STR_FONT_SIZE: "Размер шрифта интерфейса"
+STR_LINE_SPACING: "Межстрочный интервал"
+STR_ASCII_LETTER_SPACING: "Интервал букв ASCII"
+STR_ASCII_DIGIT_SPACING: "Интервал цифр ASCII"
+STR_CJK_SPACING: "Интервал CJK"
+STR_COLOR_MODE: "Цветовой режим"
+STR_SCREEN_MARGIN: "Поля экрана"
+STR_PARA_ALIGNMENT: "Выравнивание абзаца"
+STR_HYPHENATION: "Перенос слов"
+STR_TIME_TO_SLEEP: "Сон Через"
```
mrtnvgr, line 112:
```suggestion
STR_TIME_TO_SLEEP: "Сон через"
```

daveallie:
@mrtnvgr would you mind opening a PR with these changes?

lib/I18n/translations/russia.yaml
mrtnvgr, line 1: This file should be called "russian.yaml".

lib/I18n/translations/russia.yaml
```diff
@@ -0,0 +1,317 @@
...
+STR_IGNORE: "Игнорировать"
+STR_SLEEP: "Сон"
+STR_PAGE_TURN: "Перелистывание"
+STR_PORTRAIT: "Портрет"
+STR_LANDSCAPE_CW: "Ландшафт (CW)"
+STR_INVERTED: "Инверсия"
+STR_LANDSCAPE_CCW: "Ландшафт (CCW)"
+STR_FRONT_LAYOUT_BCLR: "Наз"
+STR_FRONT_LAYOUT_LRBC: "Лев"
+STR_FRONT_LAYOUT_LBCR: "Лев"
```
mrtnvgr, line 169:
```suggestion
STR_FRONT_LAYOUT_BCLR: "Наз, Ок, Лев, Прав"
STR_FRONT_LAYOUT_LRBC: "Лев, Прав, Наз, Ок"
STR_FRONT_LAYOUT_LBCR: "Лев, Наз, Ок, Прав"
```

`Прав` can be changed to `Прв` if needed, i think it would be understandable for language speakers.

lib/I18n/translations/russia.yaml
```diff
@@ -0,0 +1,317 @@
...
+STR_PREV_NEXT: "Назад/Вперёд"
+STR_NEXT_PREV: "Вперёд/Назад"
+STR_BOOKERLY: "Bookerly"
+STR_NOTO_SANS: "Noto Sans"
+STR_OPEN_DYSLEXIC: "Open Dyslexic"
```
mrtnvgr, line 174: Font names should not be translated, maybe we can omit them here?

lib/I18n/translations/russia.yaml
```diff
@@ -0,0 +1,317 @@
...
+STR_CHECKING_UPDATE: "Проверка обновлений..."
+STR_NEW_UPDATE: "Доступно новое обновление!"
+STR_CURRENT_VERSION: "Текущая версия:"
+STR_NEW_VERSION: "Новая версия:"
+STR_UPDATING: "Обновление..."
+STR_NO_UPDATE: "Обновлений нет"
+STR_UPDATE_FAILED: "Ошибка обновления"
+STR_UPDATE_COMPLETE: "Обновление завершено"
+STR_POWER_ON_HINT: "Удерживайте кнопку питания для включения"
+STR_EXTERNAL_FONT: "Внешний шрифт"
```
mrtnvgr, line 206:
```suggestion
STR_EXTERNAL_FONT: "Сторонний шрифт"
```

`Пользовательский` is nice too.

lib/I18n/translations/russia.yaml
```diff
@@ -0,0 +1,317 @@
...
+STR_NOT_SET: "Не установлено"
+STR_DIR_LEFT: "Влево"
+STR_DIR_RIGHT: "Вправо"
+STR_DIR_UP: "Вверх"
+STR_DIR_DOWN: "Вниз"
+STR_CAPS_ON: "CAPS"
+STR_CAPS_OFF: "caps"
+STR_OK_BUTTON: "OK"
+STR_ON_MARKER: "[ВКЛ]"
+STR_SLEEP_COVER_FILTER: "Фильтр обложки сна"
```
mrtnvgr, line 249:
```suggestion
STR_SLEEP_COVER_FILTER: "Фильтр экрана сна"
```
Definition of `screen` here will fit better I think.

Also:

Image

mrtnvgr: As the filter applies only to the `Обложка` (cover), the original value can be right. Something worth considering nonetheless :)

mrtnvgr:
> @mrtnvgr would you mind opening a PR with these changes?

Yes, when I've started the review the PR was open 😆

PR #731, opened 2026-02-06
feat: wakeup target detection

osteotek:
## Summary

* If going to sleep was from the Reader view, wake up to the same book. Otherwise, wakeup to the Home view

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -333,7 +333,8 @@ void setup() {
   APP_STATE.loadFromFile();
   RECENT_BOOKS.loadFromFile();
 
-  // Boot to home screen directly when back button is held or when reader activity crashes 3 times
+  // Boot to home screen if no book is open, last sleep was not from reader, back button is held, or reader activity
+  // crashed (indicated by readerActivityLoadCount > 0)
   if (APP_STATE.openEpubPath.empty() || !APP_STATE.lastSleepFromReader ||
       mappedInputManager.isPressed(MappedInputManager::Button::Back) || APP_STATE.readerActivityLoadCount > 0) {
     onGoHome();


iandchasse:
Could this be extended to effectively implement deep sleep during reading? As in a flow of:

1. Open book
2. Deep sleep
3. Turn page 
4. Wake up (detect ADC input in ULP?)
5. Turn page (or exit/power depending on action)
6. Repeat from step 2

It would effectively save a lot of battery life for the use case of reading

osteotek:
> Could this be extended to effectively implement deep sleep during reading? As in a flow of:
> 
Yeah, I thought of something similar. But the only way to wake up from deepsleep is the power button, that's how the device is wired. And waking up can take up to a second, not really suitable for reading

src/main.cpp
```diff
@@ -331,9 +333,10 @@ void setup() {
   APP_STATE.loadFromFile();
   RECENT_BOOKS.loadFromFile();
 
-  // Boot to home screen directly when back button is held or when reader activity crashes 3 times
-  if (APP_STATE.openEpubPath.empty() || mappedInputManager.isPressed(MappedInputManager::Button::Back) ||
-      APP_STATE.readerActivityLoadCount > 0) {
+  // Boot to home screen if no book is open, last sleep was not from reader, back button is held, or reader activity
+  // crashed (indicated by readerActivityLoadCount > 0)
```
jdk2pq, line 337: Thanks for catching this and updating the comment to reflect the actual behavior!

src/main.cpp
```diff
@@ -194,6 +194,8 @@ void waitForPowerRelease() {
 
 // Enter deep sleep mode
 void enterDeepSleep() {
+  APP_STATE.lastSleepFromReader = currentActivity && currentActivity->isReaderActivity();
+  APP_STATE.saveToFile();
```
jdk2pq, line 198: minor, non-blocking: `saveToFile()` returns a bool that _could be_ checked and have some kind of Serial debug statement about not being able to save the APP_STATE. But that's not currently a pattern we follow anywhere else `saveToFile` or `loadFromFile` gets called, so I'm completely fine with leaving this as is too.

CaptainFrito (approved):

PR #733, opened 2026-02-06
feat: split status bar setting

whyte-j:
## Summary

This PR aims to reduce the complexity of the status bar by splitting the setting into 5:
- Chapter Page Count
- Book Progress %
- Progress Bar
- Chapter Title
- Battery Indicator

These are located within the new StausBarSettings activity, which also shows a preview of the bar the user has created

image

image

When updating from a previous version, the user's past settings are honoured.

## Additional Context

The PR aims to remove any duplication of status bar code where possible, and extracts the status bar rendering into a new component - StatusBar

It also adds a new (optional) padding option to the progress bar to allow the status bar to be shifted upwards - this is only intended for use in the settings.

---

### AI Usage

No - although did help to decode some C++ errors


el:
I liked the idea of a more customizable status bar, but the settings look to crowded this way. I think we need settings grouping first.

whyte-j:
> I liked the idea of a more customizable status bar, but the settings look to crowded this way. I think we need settings grouping first.

Absolutely agree - I think a separate menu activity (such as the one we use for KOReader Sync) would be the best route, with a preview of what the menu bar would look like. As a (_very_) rough example:
image

This PR so far does the very basics, updating the code to split the status bar settings, but I didn't want to pursue any UI changes without some feedback first - I do think we need those changes before we merge this PR, though 🙂

whyte-j:
image

image


New update! Settings have been moved to a separate menu, found in the Reader sub-menu.

In my view, the following tasks remain:
* I've noticed the names of the buttons are inconsistent - I'll update these to match settings.
* Battery % for status bar should be part of this menu, rather than where it is currently.
* Pulling the render behaviour from txt & epub and creating a utility to draw it instead. This can then be used to render an example status bar in the status bar menu. 
* Remove old status bar setting from page.

Eloren1:
Thank you, it looks nice with this kind of preview. Small things: 

Display Book Progress Bar missing, 

- Display Book Progress Bar: `OFF` / `Thin` (2px) / `Medium` (4px) (default) / `Thick` (6px)

This is a small setting for those who want it. It works as before by default.

---

You can also implement the setting called `Progress Bar Fill Edges`, useful for black devices. It is also a simple one line fix:

https://github.com/crosspoint-reader/crosspoint-reader/issues/570#issuecomment-3818073964

---

`Remap Front Buttons` panel has a button called `Remap Front Buttons` 

`Status Bar Settings` panel has a button called just `Status Bar` (inconsistent)

---

`Progress Bar` → `Book` naming feels off. In my initial Issue I used titles as close to the original as possible

And it should probably say `Display ___` like before. You used `Show` for the battery because it feels short, but same goes with the rest: it's just `Chapter Title` (inconsistent)

Other than that, this page looks really good

src/activities/reader/EpubReaderActivity.cpp
```diff
@@ -458,7 +459,7 @@ void EpubReaderActivity::onReaderMenuConfirm(EpubReaderMenuActivity::MenuAction
     case EpubReaderMenuActivity::MenuAction::SYNC: {
       if (KOREADER_STORE.hasCredentials()) {
         xSemaphoreTake(renderingMutex, portMAX_DELAY);
-        const int currentPage = section ? section->currentPage : 0;
+        const int currentPage = section ? section->currentPage + 1 : 0;
```
osteotek: what is the reason behind this change?

whyte-j: Ah, good spot - didn't mean to change it there. Thanks https://github.com/crosspoint-reader/crosspoint-reader/pull/733/commits/ec0e997ca1d8994952265b73c31e09848adc85da

whyte-j:
> Display Book Progress Bar: OFF / Thin (2px) / Medium (4px) (default) / Thick (6px)
> This is a small setting for those who want it. It works as before by default.

Currently working on this now - is this a highly sought after feature? I'm weary that we don't want too add too many small settings, as we could end up with hundreds of those! Happy for Dave/a collaborator to weigh in before the merge on that, though.

> You can also implement the setting called Progress Bar Fill Edges, useful for black devices. It is also a simple one line fix:

I don't know enough about this issue - looks like from the discussions the fill could be better as a fix rather than a new setting?

> Progress Bar → Book naming feels off. In my initial Issue I used titles as close to the original as possible
> And it should probably say Display ___ like before. You used Show for the battery because it feels short, but same goes with the rest: it's just Chapter Title (inconsistent)

They're intended to match (more or less) to the naming structure of the rest of the settings - the majority have the target as the name and the action as the option (think Paragraph Alignment -> Justify rather than Align Paragraph -> Justified). I do agree that it needs a little update (thanks!), so I've changed the options to verbs https://github.com/crosspoint-reader/crosspoint-reader/pull/733/commits/b423319f85c2ce6fdd370d33cb4b0cb887b735d6

> Remap Front Buttons panel has a button called Remap Front Buttons
> Status Bar Settings panel has a button called just Status Bar (inconsistent)

Good point - I've updated to "Customise Status Bar" as that feels like it would be suitable for both places https://github.com/crosspoint-reader/crosspoint-reader/pull/733/commits/65dc63f7f66cf6674823d5cdc8d5539e810b2f6b

Thanks for your feedback! 🙂

Eloren1:
Bar thickness is not a small setting, in horizontal mode on a white model the default line looks too thick #602 and distracting, not how it was intended (it looks normal in vertical mode). And border fill will look even thicker on a white model. 

The two new settings inside a specific subpage don't affect the basic experience at all, it works like before by default. And these settings actually improve the experience, as they provide a decent solution for black devices without ruining the visual appearance of white devices. This is literally taking everyone's interests into account, rather than trying to create a uniform experience for everyone.

whyte-j:
> Bar thickness is not a small setting, in horizontal mode on a white model the default line looks too thick #602 and distracting, not how it was intended (it looks normal in vertical mode). And border fill will look even thicker on a white model.
> 
> The two new settings inside a specific subpage don't affect the basic experience at all, it works like before by default. And these settings actually improve the experience, as they provide a decent solution for black devices without ruining the visual appearance of white devices. This is literally taking everyone's interests into account, rather than trying to create a uniform experience for everyone.

Ah - got you. In that case I agree, although that does make the preview rather redundant for this setting, given the use case. I think we may have to just accept that, though - unless anyone has suggestions?

Eloren1:
It's as if you're overthinking the settings without prioritizing actual reading experience. Navigating through this settings section will take up 1% of the user's time. But the other 99% of the time:
- Users of black e-readers will see big status bar margins or they won't use it at all
- Users of white e-readers in portrait orientation are fine 
- Users of white e-readers in landscape orientation see a thicker bar that is distracting

I'm fine with the case where, after reading in horizontal mode I see a thick bar and go to Settings → change the thickness.
Yes, the preview is only for the vertical one, but I know what I'm changing, and the preview wasn't here before anyway.

osteotek:
> It's as if you're overthinking the settings without prioritizing actual reading experience.


@Eloren1 you've been told again and again. Please stop with accusatory language. It's not welcome in this community

whyte-j:
> > It's as if you're overthinking the settings without prioritizing actual reading experience.
> 
> @Eloren1 you've been told again and again. Please stop with accusatory language. It's not welcome in this community

And for what it's worth - I was agreeing with you. I think we may have to just accept that the user will have to flick back and fourth a couple times if they want to check what it looks like for them

Eloren1:
> accusatory

You're giving meaning to words that weren't meant to be, interrupting the discussion.

It seems there's already a bias towards my words. Although it would be better not to forget [the "Human First" Rule.](https://github.com/crosspoint-reader/crosspoint-reader/blob/master/GOVERNANCE.md#1-the-human-first-rule)

Eloren1:
> flick back and fourth a couple times if they want to check what it looks like for them

Haha that's generally how it was before. But now at least 90% of the items will be available for preview in real time

the-iek:
Do you think it is possible to add an option for decimals for book progress? I think it was a bug back in Crosspoint 0.15 and got fixed but I really grew to like it.

osteotek:
@whyte-j Would it be possible to fix conflicts? Keen to merge this

whyte-j:
> @whyte-j Would it be possible to fix conflicts? Keen to merge this

Happy to hear it! Will try and get it sorted this evening 🙂 

As for the bar thickness @Eloren1, if we're eager to get this merged I can take a look at doing that in a follow up PR, as it won't need to use the legacy setting transformation. I got a couple of odd bugs that I haven't had the chance to squash just yet

whyte-j:
@osteotek all sorted!

src/components/StatusBar.h
```diff
@@ -0,0 +1,7 @@
+#include <GfxRenderer.h>
+
+namespace StatusBar {
+void renderStatusBar(GfxRenderer& renderer, const int marginRight, int marginBottom, const int marginLeft,
+                     const float sectionChapterProg, const float bookProgress, const int currentPage,
+                     const int pageCount, std::string title, const int paddingBottom = 0);
```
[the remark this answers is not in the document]
whyte-j: The title cannot be set to constant in this case as title is altered if the string is too long to be fully rendered - see:
`title = renderer.truncatedText(SMALL_FONT_ID, title.c_str(), availableTitleSpace);`

src/components/StatusBar.cpp
```diff
@@ -0,0 +1,84 @@
...
+
+#include "CrossPointSettings.h"
+#include "CrossPointState.h"
+#include "components/UITheme.h"
+#include "fontIds.h"
+
+namespace StatusBar {
+void renderStatusBar(GfxRenderer& renderer, const int marginRight, int marginBottom, const int marginLeft,
+                     const float sectionChapterProg, const float bookProgress, const int currentPage,
+                     const int pageCount, std::string title, int paddingBottom) {
```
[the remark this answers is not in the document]
whyte-j: As above - the title cannot be set to constant in this case as title is altered if the string is too long to be fully rendered - see:
`title = renderer.truncatedText(SMALL_FONT_ID, title.c_str(), availableTitleSpace);`

osteotek:
With only the progress bar visible, it seems like we can add one more line of text? That's with default font
image

whyte-j:
> With only the progress bar visible, it seems like we can add one more line of text? That's with default font image

Ah, good point! I've updated the margin calculations:
image

It looks like since the CSS updates I do get a variable distance between the status/progress bar and the text, but (anecdotally) that looks to be tied to gaps between paragraphs

osteotek (dismissed):

whyte-j:
We've got new merge conflicts! Taking a look now 🙂

src/activities/settings/StatusBarSettingsActivity.cpp
```diff
@@ -0,0 +1,178 @@
...
+  while (true) {
+    if (updateRequired) {
+      updateRequired = false;
+      xSemaphoreTake(renderingMutex, portMAX_DELAY);
+      render();
+      xSemaphoreGive(renderingMutex);
+    }
+    vTaskDelay(10 / portTICK_PERIOD_MS);
+  }
+}
```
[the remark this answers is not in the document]
whyte-j: Removed as part of update from https://github.com/crosspoint-reader/crosspoint-reader/pull/774

src/components/StatusBar.h
```diff
@@ -0,0 +1,8 @@
+#pragma once
+#include <GfxRenderer.h>
+
+namespace StatusBar {
+void renderStatusBar(GfxRenderer& renderer, const int marginRight, const int marginBottom, const int marginLeft,
+                     const float bookProgress, const int currentPage, const int pageCount, std::string title,
+                     const int paddingBottom = 0);
```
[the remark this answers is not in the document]
whyte-j: Added

whyte-j:
```
STR_SHOW: "Show"
STR_HIDE: "Hide"
STR_CUSTOMISE_STATUS_BAR: "Customise Status Bar"
STR_CHAPTER_PAGE_COUNT: "Chapter Page Count"
STR_BOOK_PROGRESS_PERCENTAGE: "Book Progress Percentage"
STR_PROGRESS_BAR: "Progress Bar"
STR_SHOW_BOOK_PROGRESS: "Show Book Progress"
STR_SHOW_CHAPTER_PROGRESS: "Show Chapter Progress"
STR_EXAMPLE_CHAPTER: "Chapter 21"
STR_PREVIEW: "Preview"
STR_CHAPTER_TITLE: "Chapter Title"
STR_BATTERY: "Battery"
```

Rebased - we do need translations for the strings above, is this compulsory for the PR, or are these done after?

lib/I18n/I18nKeys.h
```diff
@@ -268,6 +268,8 @@ enum class StrId : uint16_t {
   STR_RETRY,
   STR_YES,
   STR_NO,
+  STR_SHOW,
+  STR_HIDE,
```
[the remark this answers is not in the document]
whyte-j: Although not ideal, there is fallback to the english strings if they do not exist in the other languages, so there will not be any misalignment at runtime

src/activities/settings/StatusBarSettingsActivity.cpp
```diff
@@ -0,0 +1,145 @@
...
+    } else if (i == 2) {
+      status = I18N.get(progressBarNames[SETTINGS.statusBarProgressBar]);
+    } else if (i == 3) {
+      status = SETTINGS.statusBarChapterTitle ? translatedShow : translatedHide;
+    } else if (i == 4) {
+      status = SETTINGS.statusBarBattery ? translatedShow : translatedHide;
+    }
+    const auto width = renderer.getTextWidth(UI_10_FONT_ID, status);
+    renderer.drawText(UI_10_FONT_ID, pageWidth - 20 - width, settingY, status, !isSelected);
+  }
```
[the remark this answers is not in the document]
whyte-j: I've updated the mod to use the size of progressBarNames. Since that is the only way the variable is updated, it now cannot accidentally be set incorrectly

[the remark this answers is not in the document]
whyte-j: Ah, good point. I've added the clamping on enter, so that we don't run the code more often than we need to

src/components/StatusBar.cpp
```diff
@@ -0,0 +1,84 @@
...
+    titleWidth = renderer.getTextWidth(SMALL_FONT_ID, title.c_str());
+    if (titleWidth > availableTitleSpace) {
+      // Not enough space to center on the screen, center it within the remaining space instead
+      availableTitleSpace = rendererableScreenWidth - titleMarginLeft - titleMarginRight;
+      titleMarginLeftAdjusted = titleMarginLeft;
+    }
+    if (titleWidth > availableTitleSpace) {
+      title = renderer.truncatedText(SMALL_FONT_ID, title.c_str(), availableTitleSpace);
+      titleWidth = renderer.getTextWidth(SMALL_FONT_ID, title.c_str());
+    }
```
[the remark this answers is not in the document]
whyte-j: Although technically possible, this seems like overkill as our margin size options are preset in settings. We could potentially instead cap the margin width at the greatest option, like we did for the settings, but I'm not convinced we'd need to do that either

src/components/StatusBar.cpp
```diff
@@ -0,0 +1,84 @@
+
+#include <GfxRenderer.h>
+
+#include <string>
+
+#include "CrossPointSettings.h"
+#include "components/UITheme.h"
+#include "fontIds.h"
+
```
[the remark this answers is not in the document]
whyte-j: Added

whyte-j:
@coderabbitai review

src/components/StatusBar.cpp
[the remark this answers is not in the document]
whyte-j: Can you mark as resolved, please?

lib/I18n/I18nKeys.h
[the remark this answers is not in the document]
whyte-j: Can you mark as resolved, please?

whyte-j:
> **Actionable comments posted: 2**
> 
> > [!CAUTION]
> > Some comments are outside the diff and can’t be posted inline due to platform limitations.
> > 
> > 
> > 
> > <details>
> > <summary>⚠️ Outside diff range comments (1)</summary><blockquote>
> > 
> > <details>
> > <summary>lib/I18n/translations/english.yaml (1)</summary><blockquote>
> > 
> > `237-262`: _⚠️ Potential issue_ | _🟠 Major_
> > 
> > **Add missing translation keys to all non-English locale files.**
> > 
> > All 12 new strings (STR_SHOW, STR_HIDE, STR_BATTERY, STR_BOOK_PROGRESS_PERCENTAGE, STR_CHAPTER_PAGE_COUNT, STR_CHAPTER_TITLE, STR_CUSTOMISE_STATUS_BAR, STR_EXAMPLE_CHAPTER, STR_PREVIEW, STR_PROGRESS_BAR, STR_SHOW_BOOK_PROGRESS, STR_SHOW_CHAPTER_PROGRESS) are missing from czech.yaml, french.yaml, german.yaml, portuguese.yaml, russia.yaml, spanish.yaml, and swedish.yaml. This will cause runtime issues if these strings are displayed to users in non-English locales.
> > 
> > <details>
> > <summary>🤖 Prompt for AI Agents</summary>
> > 
> > ```
> > Verify each finding against the current code and only fix it if needed.
> > 
> > In `@lib/I18n/translations/english.yaml` around lines 237 - 262, The non-English
> > locale files are missing 12 translation keys introduced in english.yaml
> > (STR_SHOW, STR_HIDE, STR_BATTERY, STR_BOOK_PROGRESS_PERCENTAGE,
> > STR_CHAPTER_PAGE_COUNT, STR_CHAPTER_TITLE, STR_CUSTOMISE_STATUS_BAR,
> > STR_EXAMPLE_CHAPTER, STR_PREVIEW, STR_PROGRESS_BAR, STR_SHOW_BOOK_PROGRESS,
> > STR_SHOW_CHAPTER_PROGRESS); add these keys to czech.yaml, french.yaml,
> > german.yaml, portuguese.yaml, russia.yaml, spanish.yaml, and swedish.yaml with
> > appropriate translations (or temporary English fallbacks) so every locale
> > defines the same keys as english.yaml to avoid runtime key-missing errors.
> > Ensure key names exactly match the ones in english.yaml and maintain YAML
> > formatting consistent with the existing locale files.
> > ```
> > 
> > </details>
> > 
> > </blockquote></details>
> > 
> > </blockquote></details>
> 
> <details>
> <summary>🤖 Fix all issues with AI agents</summary>
> 
> ```
> Verify each finding against the current code and only fix it if needed.
> 
> 
> In `@lib/I18n/translations/english.yaml`:
> - Around line 237-262: The non-English locale files are missing 12 translation
> keys introduced in english.yaml (STR_SHOW, STR_HIDE, STR_BATTERY,
> STR_BOOK_PROGRESS_PERCENTAGE, STR_CHAPTER_PAGE_COUNT, STR_CHAPTER_TITLE,
> STR_CUSTOMISE_STATUS_BAR, STR_EXAMPLE_CHAPTER, STR_PREVIEW, STR_PROGRESS_BAR,
> STR_SHOW_BOOK_PROGRESS, STR_SHOW_CHAPTER_PROGRESS); add these keys to
> czech.yaml, french.yaml, german.yaml, portuguese.yaml, russia.yaml,
> spanish.yaml, and swedish.yaml with appropriate translations (or temporary
> English fallbacks) so every locale defines the same keys as english.yaml to
> avoid runtime key-missing errors. Ensure key names exactly match the ones in
> english.yaml and maintain YAML formatting consistent with the existing locale
> files.
> 
> In `@src/activities/settings/StatusBarSettingsActivity.cpp`:
> - Around line 97-107: The updateRequired flag is subject to a data race between
> StatusBarSettingsActivity::loop() and
> StatusBarSettingsActivity::displayTaskLoop(); fix by making updateRequired a
> thread-safe variable or protecting all accesses: change the member
> updateRequired to std::atomic<bool> (or alternatively wrap every read/write of
> updateRequired in xSemaphoreTake/xSemaphoreGive using renderingMutex), update
> the assignments in loop() (the writes at the spots you noted) and the read/write
> in displayTaskLoop() to use the atomic operations (or mutex-protected blocks) so
> all accesses are synchronized.
> - Around line 20-22: The file-scoped constant progressBarNames (and its size
> PROGRESS_BAR_ITEMS) is declared outside the anonymous namespace while menuNames
> is inside; move PROGRESS_BAR_ITEMS and progressBarNames into the same anonymous
> namespace as menuNames to give them internal linkage and avoid
> cross-translation-unit symbol collisions—update the declaration of
> PROGRESS_BAR_ITEMS and the definition of progressBarNames so they reside inside
> the anonymous namespace alongside menuNames.
> - Around line 42-50: onExit() currently takes renderingMutex and deletes
> displayTaskHandle directly which can leave the mutex/state inconsistent; instead
> add a volatile/atomic termination flag (e.g., stopRequested) that
> displayTaskLoop checks and exits cleanly, have onExit() set that flag, unblock
> the task if it might be waiting (release or give renderingMutex or use a task
> notification), wait for displayTaskLoop to exit (poll or use a notification/join
> mechanism until displayTaskHandle becomes nullptr or the task signals
> termination), and only then delete the task handle and
> vSemaphoreDelete(renderingMutex); update displayTaskLoop to check stopRequested
> and return cleanly so the mutex state remains consistent.
> 
> In `@src/components/StatusBar.cpp`:
> - Around line 1-6: The implementation file is missing its own header include;
> add `#include` "StatusBar.h" at the top of src/components/StatusBar.cpp so the
> definition of the StatusBar class and its declarations are validated against the
> implementation (this ensures any mismatches between the StatusBar interface and
> the cpp definitions are caught at compile time).
> 
> In `@src/components/StatusBar.h`:
> - Around line 1-7: The header StatusBar is not self-contained because
> renderStatusBar declares a parameter of type std::string but the header doesn't
> include <string>; add `#include` <string> to the top of the file so the
> declaration of StatusBar::renderStatusBar (and any future uses of std::string in
> this header) is valid regardless of transitive includes from GfxRenderer.h.
> ```
> 
> </details>
> 
> <details>
> <summary>🧹 Nitpick comments (3)</summary><blockquote>
> 
> <details>
> <summary>🤖 Fix all nitpicks with AI agents</summary>
> 
> ```
> Verify each finding against the current code and only fix it if needed.
> 
> 
> In `@src/activities/settings/StatusBarSettingsActivity.cpp`:
> - Around line 20-22: The file-scoped constant progressBarNames (and its size
> PROGRESS_BAR_ITEMS) is declared outside the anonymous namespace while menuNames
> is inside; move PROGRESS_BAR_ITEMS and progressBarNames into the same anonymous
> namespace as menuNames to give them internal linkage and avoid
> cross-translation-unit symbol collisions—update the declaration of
> PROGRESS_BAR_ITEMS and the definition of progressBarNames so they reside inside
> the anonymous namespace alongside menuNames.
> - Around line 42-50: onExit() currently takes renderingMutex and deletes
> displayTaskHandle directly which can leave the mutex/state inconsistent; instead
> add a volatile/atomic termination flag (e.g., stopRequested) that
> displayTaskLoop checks and exits cleanly, have onExit() set that flag, unblock
> the task if it might be waiting (release or give renderingMutex or use a task
> notification), wait for displayTaskLoop to exit (poll or use a notification/join
> mechanism until displayTaskHandle becomes nullptr or the task signals
> termination), and only then delete the task handle and
> vSemaphoreDelete(renderingMutex); update displayTaskLoop to check stopRequested
> and return cleanly so the mutex state remains consistent.
> 
> In `@src/components/StatusBar.cpp`:
> - Around line 1-6: The implementation file is missing its own header include;
> add `#include` "StatusBar.h" at the top of src/components/StatusBar.cpp so the
> definition of the StatusBar class and its declarations are validated against the
> implementation (this ensures any mismatches between the StatusBar interface and
> the cpp definitions are caught at compile time).
> ```
> 
> </details>
> 
> <details>
> <summary>src/components/StatusBar.cpp (1)</summary><blockquote>
> 
> `1-6`: **Missing include for `StatusBar.h` header.**
> 
> The implementation file should include its own header to ensure interface consistency and catch declaration/definition mismatches at compile time.
> 
> 
> <details>
> <summary>♻️ Add missing header include</summary>
> 
> ```diff
> +#include "StatusBar.h"
>  
>  `#include` <GfxRenderer.h>
>  
>  `#include` "CrossPointSettings.h"
>  `#include` "components/UITheme.h"
>  `#include` "fontIds.h"
> ```
> </details>
> 
> <details>
> <summary>🤖 Prompt for AI Agents</summary>
> 
> ```
> Verify each finding against the current code and only fix it if needed.
> 
> In `@src/components/StatusBar.cpp` around lines 1 - 6, The implementation file is
> missing its own header include; add `#include` "StatusBar.h" at the top of
> src/components/StatusBar.cpp so the definition of the StatusBar class and its
> declarations are validated against the implementation (this ensures any
> mismatches between the StatusBar interface and the cpp definitions are caught at
> compile time).
> ```
> 
> </details>
> 
> </blockquote></details>
> <details>
> <summary>src/activities/settings/StatusBarSettingsActivity.cpp (2)</summary><blockquote>
> 
> `20-22`: **Inconsistent namespace scoping for constants.**
> 
> `progressBarNames` is defined at file scope outside the anonymous namespace, while `menuNames` is inside it. This inconsistency could lead to linkage issues if another translation unit defines a symbol with the same name.
> 
> 
> <details>
> <summary>♻️ Move to anonymous namespace</summary>
> 
> ```diff
>  namespace {
>  constexpr int MENU_ITEMS = 5;
>  const StrId menuNames[MENU_ITEMS] = {StrId::STR_CHAPTER_PAGE_COUNT, StrId::STR_BOOK_PROGRESS_PERCENTAGE,
>                                       StrId::STR_PROGRESS_BAR, StrId::STR_CHAPTER_TITLE, StrId::STR_BATTERY};
> -}  // namespace
>  
> -constexpr int PROGRESS_BAR_ITEMS = 3;
> -const StrId progressBarNames[PROGRESS_BAR_ITEMS] = {StrId::STR_SHOW_BOOK_PROGRESS, StrId::STR_SHOW_CHAPTER_PROGRESS,
> +constexpr int PROGRESS_BAR_ITEMS = 3;
> +const StrId progressBarNames[PROGRESS_BAR_ITEMS] = {StrId::STR_SHOW_BOOK_PROGRESS, StrId::STR_SHOW_CHAPTER_PROGRESS,
>                                                      StrId::STR_HIDE};
> +}  // namespace
> ```
> </details>
> 
> <details>
> <summary>🤖 Prompt for AI Agents</summary>
> 
> ```
> Verify each finding against the current code and only fix it if needed.
> 
> In `@src/activities/settings/StatusBarSettingsActivity.cpp` around lines 20 - 22,
> The file-scoped constant progressBarNames (and its size PROGRESS_BAR_ITEMS) is
> declared outside the anonymous namespace while menuNames is inside; move
> PROGRESS_BAR_ITEMS and progressBarNames into the same anonymous namespace as
> menuNames to give them internal linkage and avoid cross-translation-unit symbol
> collisions—update the declaration of PROGRESS_BAR_ITEMS and the definition of
> progressBarNames so they reside inside the anonymous namespace alongside
> menuNames.
> ```
> 
> </details>
> 
> ---
> 
> `42-50`: **Task deletion while holding mutex could cause issues.**
> 
> In `onExit()`, the mutex is taken before deleting the task. If `displayTaskLoop` is currently waiting on the mutex when `vTaskDelete` is called, the mutex may be left in an inconsistent state. While FreeRTOS handles this by simply deleting the task, it's a subtle pattern.
> 
> The current approach works but could be made more explicit by setting a termination flag and allowing the task to exit gracefully.
> 
> <details>
> <summary>🤖 Prompt for AI Agents</summary>
> 
> ```
> Verify each finding against the current code and only fix it if needed.
> 
> In `@src/activities/settings/StatusBarSettingsActivity.cpp` around lines 42 - 50,
> onExit() currently takes renderingMutex and deletes displayTaskHandle directly
> which can leave the mutex/state inconsistent; instead add a volatile/atomic
> termination flag (e.g., stopRequested) that displayTaskLoop checks and exits
> cleanly, have onExit() set that flag, unblock the task if it might be waiting
> (release or give renderingMutex or use a task notification), wait for
> displayTaskLoop to exit (poll or use a notification/join mechanism until
> displayTaskHandle becomes nullptr or the task signals termination), and only
> then delete the task handle and vSemaphoreDelete(renderingMutex); update
> displayTaskLoop to check stopRequested and return cleanly so the mutex state
> remains consistent.
> ```
> 
> </details>
> 
> </blockquote></details>
> 
> </blockquote></details>
> 
> <details>
> <summary>📜 Review details</summary>
> 
> **Configuration used**: Organization UI
> 
> **Review profile**: CHILL
> 
> **Plan**: Pro
> 
> <details>
> <summary>📥 Commits</summary>
> 
> Reviewing files that changed from the base of the PR and between 97c33141bd02d225be68ddb9722fa85b3165cce7 and 20f3f46dd5d17652487122dd5819f71d5c68d4af.
> 
> </details>
> 
> <details>
> <summary>📒 Files selected for processing (16)</summary>
> 
> * `lib/I18n/I18n.cpp`
> * `lib/I18n/I18nKeys.h`
> * `lib/I18n/translations/english.yaml`
> * `src/CrossPointSettings.cpp`
> * `src/CrossPointSettings.h`
> * `src/SettingsList.h`
> * `src/activities/reader/EpubReaderActivity.cpp`
> * `src/activities/reader/TxtReaderActivity.cpp`
> * `src/activities/settings/SettingsActivity.cpp`
> * `src/activities/settings/SettingsActivity.h`
> * `src/activities/settings/StatusBarSettingsActivity.cpp`
> * `src/activities/settings/StatusBarSettingsActivity.h`
> * `src/components/StatusBar.cpp`
> * `src/components/StatusBar.h`
> * `src/components/themes/BaseTheme.cpp`
> * `src/components/themes/BaseTheme.h`
> 
> </details>
> 
> <details>
> <summary>🧰 Additional context used</summary>
> 
> <details>
> <summary>🧠 Learnings (1)</summary>
> 
> <details>
> <summary>📚 Learning: 2026-02-16T20:43:12.609Z</summary>
> 
> ```
> Learnt from: Levrk
> Repo: crosspoint-reader/crosspoint-reader PR: 909
> File: src/activities/reader/EpubReaderActivity.cpp:461-461
> Timestamp: 2026-02-16T20:43:12.609Z
> Learning: In src/activities/reader/EpubReaderActivity.cpp, when using ConfirmationActivity with enterNewActivity(), it's not necessary to call exitActivity() beforehand. The ConfirmationActivity operates as a modal dialog and cleanup is handled via the pendingSubactivityExit flag in the callback lambda. This differs from other activities like EpubReaderChapterSelectionActivity or KOReaderSyncActivity which do require exitActivity() before enterNewActivity().
> ```
> 
> **Applied to files:**
> - `src/activities/settings/StatusBarSettingsActivity.h`
> - `src/activities/settings/SettingsActivity.cpp`
> - `src/activities/settings/StatusBarSettingsActivity.cpp`
> - `src/activities/reader/EpubReaderActivity.cpp`
> - `src/activities/reader/TxtReaderActivity.cpp`
> 
> </details>
> 
> </details><details>
> <summary>🧬 Code graph analysis (7)</summary>
> 
> <details>
> <summary>src/components/themes/BaseTheme.cpp (1)</summary><blockquote>
> 
> <details>
> <summary>src/main.cpp (1)</summary>
> 
> * `renderer` (37-37)
> 
> </details>
> 
> </blockquote></details>
> <details>
> <summary>src/components/themes/BaseTheme.h (1)</summary><blockquote>
> 
> <details>
> <summary>src/components/themes/BaseTheme.cpp (2)</summary>
> 
> * `drawReadingProgressBar` (660-671)
> * `drawReadingProgressBar` (660-661)
> 
> </details>
> 
> </blockquote></details>
> <details>
> <summary>src/activities/settings/StatusBarSettingsActivity.h (1)</summary><blockquote>
> 
> <details>
> <summary>src/activities/settings/StatusBarSettingsActivity.cpp (14)</summary>
> 
> * `onEnter` (29-40)
> * `onEnter` (29-29)
> * `onExit` (42-50)
> * `onExit` (42-42)
> * `loop` (52-73)
> * `loop` (52-52)
> * `taskTrampoline` (24-27)
> * `taskTrampoline` (24-24)
> * `displayTaskLoop` (97-107)
> * `displayTaskLoop` (97-97)
> * `render` (109-178)
> * `render` (109-109)
> * `handleSelection` (75-95)
> * `handleSelection` (75-75)
> 
> </details>
> 
> </blockquote></details>
> <details>
> <summary>src/activities/settings/SettingsActivity.cpp (1)</summary><blockquote>
> 
> <details>
> <summary>src/activities/settings/SettingsActivity.h (1)</summary>
> 
> * `Action` (77-83)
> 
> </details>
> 
> </blockquote></details>
> <details>
> <summary>src/activities/settings/StatusBarSettingsActivity.cpp (2)</summary><blockquote>
> 
> <details>
> <summary>lib/I18n/I18n.cpp (2)</summary>
> 
> * `getInstance` (15-18)
> * `getInstance` (15-15)
> 
> </details>
> <details>
> <summary>src/CrossPointSettings.h (1)</summary>
> 
> * `getInstance` (187-187)
> 
> </details>
> 
> </blockquote></details>
> <details>
> <summary>src/activities/reader/EpubReaderActivity.cpp (2)</summary><blockquote>
> 
> <details>
> <summary>src/activities/reader/TxtReaderActivity.cpp (2)</summary>
> 
> * `renderStatusBar` (456-463)
> * `renderStatusBar` (456-457)
> 
> </details>
> <details>
> <summary>src/components/StatusBar.cpp (2)</summary>
> 
> * `renderStatusBar` (9-81)
> * `renderStatusBar` (9-11)
> 
> </details>
> 
> </blockquote></details>
> <details>
> <summary>src/activities/reader/TxtReaderActivity.cpp (1)</summary><blockquote>
> 
> <details>
> <summary>src/activities/reader/EpubReaderActivity.cpp (2)</summary>
> 
> * `renderStatusBar` (678-697)
> * `renderStatusBar` (678-679)
> 
> </details>
> 
> </blockquote></details>
> 
> </details>
> 
> </details>
> 
> <details>
> <summary>⏰ Context from checks skipped due to timeout of 90000ms. You can increase the timeout in your CodeRabbit configuration to a maximum of 15 minutes (900000ms). (1)</summary>
> 
> * GitHub Check: cppcheck
> 
> </details>
> 
> <details>
> <summary>🔇 Additional comments (19)</summary><blockquote>
> 
> <details>
> <summary>lib/I18n/I18nKeys.h (1)</summary><blockquote>
> 
> `271-272`: **LGTM — new StrId entries look consistent.**
> 
> The added string IDs align with the new status bar options and show/hide labels.  
> 
> 
> 
> Also applies to: 287-296
> 
> </blockquote></details>
> <details>
> <summary>src/components/themes/BaseTheme.cpp (1)</summary><blockquote>
> 
> `660-669`: **LGTM — padding applied consistently.**
> 
> Line 667’s Y-offset change cleanly integrates the new `additionalYPadding` parameter.
> 
> </blockquote></details>
> <details>
> <summary>lib/I18n/I18n.cpp (1)</summary><blockquote>
> 
> `89-96`: **LGTM — no functional change.**
> 
> </blockquote></details>
> <details>
> <summary>src/activities/settings/SettingsActivity.h (1)</summary><blockquote>
> 
> `16-19`: **LGTM — enum extension is straightforward.**
> 
> </blockquote></details>
> <details>
> <summary>src/activities/settings/SettingsActivity.cpp (1)</summary><blockquote>
> 
> `15-15`: **LGTM — Customise Status Bar action is wired correctly.**  
> 
> 
> 
> Also applies to: 55-55, 186-188
> 
> </blockquote></details>
> <details>
> <summary>src/components/themes/BaseTheme.h (1)</summary><blockquote>
> 
> `120-121`: No action needed. The codebase contains only one derived theme class (LyraTheme), which does not override `drawReadingProgressBar`. The base class implementation and all call sites already use the new signature correctly.
> 
> 
> 
> > Likely an incorrect or invalid review comment.
> 
> </blockquote></details>
> <details>
> <summary>src/CrossPointSettings.cpp (3)</summary><blockquote>
> 
> `78-126`: **LGTM! Legacy status bar migration logic is well-structured.**
> 
> The `applyLegacyStatusBarSettings` function correctly maps old `STATUS_BAR_MODE` enum values to the new granular settings fields. The switch cases properly cover all legacy modes, and the default case handles the `FULL` mode appropriately.
> 
> ---
> 
> `186-190`: **LGTM! New status bar fields serialization is correctly placed.**
> 
> The new fields are appended at the end of the serialization order, maintaining backward compatibility with older settings files as noted in the comment on line 191.
> 
> ---
> 
> `319-340`: **LGTM! Status bar deserialization and migration logic is correct.**
> 
> The `statusBarSettingsRead` flag properly tracks whether the new fields were present in the file, and the migration is correctly applied only when the new fields are absent. This ensures seamless upgrades from older versions.
> 
> </blockquote></details>
> <details>
> <summary>src/activities/reader/TxtReaderActivity.cpp (2)</summary><blockquote>
> 
> `146-150`: **LGTM! Status bar margin calculation is correctly implemented.**
> 
> The logic properly determines `showStatusBar` and `showProgressBar` based on the new granular settings, and adjusts the bottom margin accordingly. This mirrors the pattern in `EpubReaderActivity.cpp`.
> 
> ---
> 
> `456-462`: **LGTM! Status bar rendering correctly delegates to the new component.**
> 
> The method properly calculates progress as a percentage (0-100) and passes all required parameters to `StatusBar::renderStatusBar`. The title is obtained from `txt->getTitle()`, which aligns with the component's expectations.
> 
> </blockquote></details>
> <details>
> <summary>src/activities/reader/EpubReaderActivity.cpp (2)</summary><blockquote>
> 
> `523-527`: **LGTM! Status bar margin calculation matches the TxtReaderActivity implementation.**
> 
> The logic correctly determines visibility flags and adjusts margins consistently with the other reader activity.
> 
> ---
> 
> `678-696`: **No null pointer dereference risk at the stated call sites.**
> 
> The guard conditions on lines 583 (`if (section->pageCount == 0)`) and 591 (`if (section->currentPage < 0 || section->currentPage >= section->pageCount)`) already dereference `section` before calling `renderStatusBar` on lines 588 and 596. If `section` were null, the program would crash when evaluating these guards, not when calling `renderStatusBar`. The same applies to line 643: `renderContents` is only called after successful page loading when `section` is guaranteed to exist. Adding a null-check to `renderStatusBar` is defensive but unnecessary given the current call patterns.
> 
> 
> 
> > Likely an incorrect or invalid review comment.
> 
> </blockquote></details>
> <details>
> <summary>src/activities/settings/StatusBarSettingsActivity.h (1)</summary><blockquote>
> 
> `1-32`: **LGTM! Well-structured activity header following existing codebase patterns.**
> 
> The class correctly uses the FreeRTOS task-based pattern with mutex synchronization for thread-safe rendering. The `[[noreturn]]` attribute on `displayTaskLoop` is appropriate as it runs an infinite loop.
> 
> </blockquote></details>
> <details>
> <summary>src/components/StatusBar.cpp (1)</summary><blockquote>
> 
> `8-81`: **LGTM! Status bar rendering logic is well-implemented.**
> 
> The component correctly handles all visibility settings and properly calculates layout positions. The title centering logic appropriately falls back to available-space centering when the title is too wide, and truncates when necessary.
> 
> </blockquote></details>
> <details>
> <summary>src/CrossPointSettings.h (2)</summary><blockquote>
> 
> `38-48`: **LGTM! Status bar enums are correctly defined.**
> 
> The legacy `STATUS_BAR_MODE` is properly documented as legacy, and the new `STATUS_BAR_PROGRESS_BAR` enum provides clear options. The enum values match the usage in `StatusBar.cpp` and `StatusBarSettingsActivity.cpp`.
> 
> ---
> 
> `132-138`: **LGTM! New status bar settings fields have appropriate defaults.**
> 
> The default values (`statusBarProgressBar = HIDE`, others = 1) provide a reasonable initial configuration that shows chapter page count, book progress percentage, chapter title, and battery, matching the "FULL" legacy mode minus the progress bar.
> 
> </blockquote></details>
> <details>
> <summary>src/SettingsList.h (1)</summary><blockquote>
> 
> `116-128`: **LGTM! Status bar settings are correctly defined for web API.**
> 
> The settings entries properly map to the new `CrossPointSettings` fields with appropriate types (Toggle for binary settings, Enum for progress bar). The category `STR_CUSTOMISE_STATUS_BAR` ensures these are grouped correctly and excluded from the device UI (handled by `StatusBarSettingsActivity` instead).
> 
> </blockquote></details>
> <details>
> <summary>src/activities/settings/StatusBarSettingsActivity.cpp (1)</summary><blockquote>
> 
> `109-178`: **LGTM! Rendering logic is clean and well-organized.**
> 
> The UI layout with header, menu items, button hints, and status bar preview is correctly implemented. The preview uses hardcoded example values (75%, 8/32, "Chapter 21") which is appropriate for demonstrating the configuration.
> 
> </blockquote></details>
> 
> </blockquote></details>
> 
> </details>
> 
> 

@coderabbitai this has been resolved as per other comments - localisation is handled at build to fall back to English if keys do not exist in the yamls

daveallie:
> ```
> STR_SHOW: "Show"
> STR_HIDE: "Hide"
> STR_CUSTOMISE_STATUS_BAR: "Customise Status Bar"
> STR_CHAPTER_PAGE_COUNT: "Chapter Page Count"
> STR_BOOK_PROGRESS_PERCENTAGE: "Book Progress Percentage"
> STR_PROGRESS_BAR: "Progress Bar"
> STR_SHOW_BOOK_PROGRESS: "Show Book Progress"
> STR_SHOW_CHAPTER_PROGRESS: "Show Chapter Progress"
> STR_EXAMPLE_CHAPTER: "Chapter 21"
> STR_PREVIEW: "Preview"
> STR_CHAPTER_TITLE: "Chapter Title"
> STR_BATTERY: "Battery"
> ```
> 
> Rebased - we do need translations for the strings above, is this compulsory for the PR, or are these done after?

Ideally would love translations here, but can do it as part of the RC as well.

whyte-j:
> The sub-menu for setting up the status bar looks fine on classic theme, but still uses the classic style, even when using the Lyra theme.

Ah, I was waiting to see if https://github.com/crosspoint-reader/crosspoint-reader/pull/732 got merged first - will update 👍🏻

osteotek:
@whyte-j while you are working on this, can we add option to show book title in place of chapter title, some users prefer that

whyte-j:
> @whyte-j while you are working on this, can we add option to show book title in place of chapter title, some users prefer that

Will do - I'll take another look at the bar thickness as well when I've got a few hours tomorrow. In the meantime though, any suggestions for the `Progress Bar` options? `Show Book Progress` and `Show Chapter Progress` are too long and get truncated in the Lyra theme. I don't love `Show Book`, currently hovering around `Book Progress` (and their chapter counterparts), but the repetition of Progress feels a little clunky

lukestein:
> any suggestions for the Progress Bar options? Show Book Progress and Show Chapter Progress are too long and get truncated in the Lyra theme. I don't love Show Book, currently hovering around Book Progress (and their chapter counterparts), but the repetition of Progress feels a little clunky

Personally I think the setting called "Progress Bar" and the options `Book`/`Chapter`/`Off`.

whyte-j:
> The sub-menu for setting up the status bar looks fine on classic theme, but still uses the classic style, even when using the Lyra theme.

Updated now - https://github.com/crosspoint-reader/crosspoint-reader/pull/733/commits/9031b736345d75127ea4aa21b45dc1023d6ca31b

whyte-j:
> @whyte-j while you are working on this, can we add option to show book title in place of chapter title, some users prefer that

Added! https://github.com/crosspoint-reader/crosspoint-reader/pull/733/commits/950dcac3868c7e663540689d145dec4774e5dd5d

whyte-j:
@Eloren1 thickness has been added here! https://github.com/crosspoint-reader/crosspoint-reader/pull/733/commits/dc3d283651929fa3f35d4e99b9c56db359c9bf1a

src/activities/settings/StatusBarSettingsActivity.cpp
```diff
@@ -0,0 +1,179 @@
...
+    return;
+  }
+
+  if (mappedInput.wasPressed(MappedInputManager::Button::Confirm)) {
+    handleSelection();
+    requestUpdate();
+    return;
+  }
+
+  if (mappedInput.wasPressed(MappedInputManager::Button::Up) ||
```
osteotek: let's switch to ButtonNavigator, like in settings. See https://github.com/crosspoint-reader/crosspoint-reader/blob/master/src/activities/settings/SettingsActivity.cpp#L99

whyte-j: Updated in https://github.com/crosspoint-reader/crosspoint-reader/pull/733/commits/d904d0feea4f84b8723c5f0582fcb3a95b6a9d54

osteotek:
@whyte-j looks good to me, thank you, that's a nice feature

src/activities/reader/EpubReaderActivity.cpp
```diff
@@ -519,14 +520,14 @@ void EpubReaderActivity::render(Activity::RenderLock&& lock) {
...
-                            (showProgressBar ? (metrics.bookProgressBarHeight + progressBarMarginTop) : 0);
-  }
+  const bool showStatusBar = SETTINGS.statusBarChapterPageCount || SETTINGS.statusBarBookProgressPercentage ||
+                             SETTINGS.statusBarTitle != CrossPointSettings::STATUS_BAR_TITLE::HIDE_TITLE ||
+                             SETTINGS.statusBarBattery;
+  const bool showProgressBar =
+      SETTINGS.statusBarProgressBar != CrossPointSettings::STATUS_BAR_PROGRESS_BAR::HIDE_PROGRESS;
+  orientedMarginBottom +=
+      (showStatusBar ? (statusBarMargin - SETTINGS.screenMargin) : 0) +
+      (showProgressBar ? (((SETTINGS.statusBarProgressBarThickness + 1) * 2) + progressBarMarginTop) : 0);
```
znelson: I was investigating this logic last night when I worked on #1058, and I think there's a small issue here (though I might be wrong).

```
orientedMarginBottom += SETTINGS.screenMargin;
// ...
orientedMarginBottom +=
      (showStatusBar ? (statusBarMargin - SETTINGS.screenMargin) : 0) // ...
```

This line in particular is replacing the margin added for `SETTINGS.screenMargin` with `statusBarMargin`.

I think we want the logic here to calculate the size of the status bar + progress bar, and set the bottom margin to the max of that and `SETTINGS.screenMargin`.


```suggestion
  const bool showStatusBar = SETTINGS.statusBarChapterPageCount || SETTINGS.statusBarBookProgressPercentage ||
                             SETTINGS.statusBarTitle != CrossPointSettings::STATUS_BAR_TITLE::HIDE_TITLE ||
                             SETTINGS.statusBarBattery;
  const bool showProgressBar =
      SETTINGS.statusBarProgressBar != CrossPointSettings::STATUS_BAR_PROGRESS_BAR::HIDE_PROGRESS;
  const uint8_t bottomMargin =
      (showStatusBar ? statusBarMargin : 0) +
      (showProgressBar ? (((SETTINGS.statusBarProgressBarThickness + 1) * 2) + progressBarMarginTop) : 0);

  orientedMarginBottom += std::max(SETTINGS.screenMargin, bottomMargin);
```

src/activities/reader/TxtReaderActivity.cpp
```diff
@@ -139,17 +140,15 @@ void TxtReaderActivity::initializeReader() {
...
-                            (showProgressBar ? (metrics.bookProgressBarHeight + progressBarMarginTop) : 0);
-  }
+  const bool showStatusBar = SETTINGS.statusBarChapterPageCount || SETTINGS.statusBarBookProgressPercentage ||
+                             SETTINGS.statusBarTitle != CrossPointSettings::STATUS_BAR_TITLE::HIDE_TITLE ||
+                             SETTINGS.statusBarBattery;
+  const bool showProgressBar =
+      SETTINGS.statusBarProgressBar != CrossPointSettings::STATUS_BAR_PROGRESS_BAR::HIDE_PROGRESS;
+  orientedMarginBottom +=
+      (showStatusBar ? (statusBarMargin - cachedScreenMargin) : 0) +
+      (showProgressBar ? ((SETTINGS.statusBarProgressBarThickness + 1) * 2 + progressBarMarginTop) : 0);
```
znelson: One more thought, not critical: 

```
    (showProgressBar ? (((SETTINGS.statusBarProgressBarThickness + 1) * 2) + progressBarMarginTop) : 0);
```

I wonder if the calculation of the progress bar thickness ought to be centralized, could be in the Theme metrics or in a StatusBar function?

src/activities/reader/EpubReaderActivity.cpp
whyte-j: Good point! It does mean that the status bar is lifted higher as the margin increases, which I was going to update, but it's actually grown on me to the point where I've left it in - @osteotek @lukestein, thoughts? (this is is at the maximum 40px)

image

I've standardised the entire height calculation process into a single function as part of this change, so I've made a separate commit - https://github.com/crosspoint-reader/crosspoint-reader/pull/733/commits/961ce3aa73f0cacb997830a424b1bf2bba3fefb6

src/activities/reader/TxtReaderActivity.cpp
whyte-j: As above - https://github.com/crosspoint-reader/crosspoint-reader/pull/733/commits/961ce3aa73f0cacb997830a424b1bf2bba3fefb6

Thanks for your feedback! Really appreciate the help 🙂

src/activities/reader/EpubReaderActivity.cpp
osteotek: hmm, not sure about that. I think I'd prefer status bar position not to depend on book margin

znelson: I think I agree with @osteotek: I prefer the aesthetics of keeping the status bar consistently pinned to the bottom, while ensuring the text honors reader margin settings.

whyte-j: How about the bar width? The text was previously tied to the margin - I don't think this makes sense anymore if we pin it to the bottom

osteotek: Not sure I understand, but I'd prefer constant position at the bottom, independent of text margin. With maximum possible bar width

whyte-j: Maybe a little hard to see with my example, but what I meant was that the original implementation of the status bar respects the margins for the left and right sides, and shrinks to fit (e.g. the text and bar both respect the 40px margin in the image previously posted).

After some thinking, I think my confusion with how the status bar should handle margins is due to the way the way margins are currently handled by the reader activities. They currently use the reader margin both to reduce the horizontal sizing and the vertical sizing. What would we think about removing it for the vertical sizing? Unless it's to account for a reader which has a different visible region, I can't see a use case, and that should probably be handled by a setting activity anyway. Looking at my kindle (for example) the text margins are also only updated to reduce the horizontal sizing.

Happy to update as part of this PR if we agree, but will need to come back to it in a couple days 🙂

znelson: I see what you're saying about the current 1.0 behavior that the battery and percentage in the status bar respect the side reader margins. I'd argue that's a bug, and the status bar should span the full width regardless of reader margins. But a lot of this is subjective. 🙂 

In addition, the 1.0 behavior where we ignore the bottom reader margin when the status bar is present seems like a bug. I believe the bottom should maintain at least enough space for the status bar, and with larger margins keep symmetry with the top. 

Here's a photo of the 1.0 behavior with large margins and status bar. Looks off to my eye:

<br /><br />

> What would we think about removing it for the vertical sizing?

My vote is to honor the top/bottom/left/right margins on reader text for now, and keep the status bar text against the bottom edge and fully justified to the sides. 

Then we could discuss top/bottom margins as a separate change.

whyte-j: After seeing how it looks now, I totally agree - not sure what I was thinking 😄 https://github.com/crosspoint-reader/crosspoint-reader/pull/733/commits/309359aa3cf732fb7661a311fe7f8f4d4b7ae7f7

I do think we should still consider dropping those top/bottom margins, but happy to leave as-is for now
image

osteotek:
@whyte-j apologies, there is another conflict, would it be possible to fix it

whyte-j:
@osteotek up now 🙂

osteotek (dismissed):

src/components/themes/BaseTheme.h
```diff
@@ -131,7 +129,8 @@ class BaseTheme {
                               const std::function<UIIcon(int index)>& rowIcon) const;
   virtual Rect drawPopup(const GfxRenderer& renderer, const char* message) const;
   virtual void fillPopupProgress(const GfxRenderer& renderer, const Rect& layout, const int progress) const;
-  virtual void drawReadingProgressBar(const GfxRenderer& renderer, const size_t bookProgress) const;
+  virtual void drawReadingProgressBar(const GfxRenderer& renderer, const size_t bookProgress,
```
ngxson: API-design-wise, I'm quite confused between `BaseTheme::drawReadingProgressBar` versus `StatusBar::renderStatusBar`: does one call another, or I need to use `drawReadingProgressBar` separately?

IMO the `renderStatusBar` function should better be part of `BaseTheme` instead of being a dedicated `StatusBar` namespace.

src/components/StatusBar.h
```diff
@@ -0,0 +1,13 @@
+#pragma once
+#include <GfxRenderer.h>
+
+namespace StatusBar {
+const uint8_t statusBarMargin = 19;
+const uint8_t progressBarMarginTop = 1;
+const uint8_t horizontalMargin = 5;
+
+uint8_t getStatusBarHeight();
```
ngxson: this should be part of theme metrics

lib/I18n/translations/english.yaml
```diff
@@ -249,10 +251,21 @@ STR_OK_BUTTON: "OK"
 STR_ON_MARKER: "[ON]"
 STR_SLEEP_COVER_FILTER: "Sleep Screen Cover Filter"
 STR_FILTER_CONTRAST: "Contrast"
-STR_STATUS_BAR_FULL_PERCENT: "Full w/ Percentage"
```
el, line 252: I think we need to delete these on other languages as well. No need to add the new ones.

whyte-j: Updated 🙂

src/components/themes/BaseTheme.h
whyte-j: That's a very good point - I've updated to add the draw into BaseTheme and the height calculation into UITheme, as that seems to be the place where other calculations of the sort are being made

src/components/StatusBar.h
whyte-j: Also updated 🙂

osteotek (dismissed):

PR #735, opened 2026-02-06
fix: Change button hint labels to Prev/Next

maeveynot:
## Summary

After merging #528, if the theme is Lyra, the navigation buttons on the home screen move through a multi-column menu, so they don't always mean exactly "up" or "down". Fix this by renaming them Prev and Next, i.e., logical movement instead of physical movement.

## Additional Context

I feel like this maps better onto the concept of the buttons navigating to the previous or next page in the reader (since we think of turning pages as left to right, or right to left in RTL languages).

This doesn't change the up/down/left/right nature of hints in the keyboard entry activity, since that actually is directional.

### AI Usage

None

### DIFF
--- a/src/activities/home/MyLibraryActivity.cpp
+++ b/src/activities/home/MyLibraryActivity.cpp
@@ -205,7 +205,7 @@ void MyLibraryActivity::render() const {
   }
 
   // Help text
-  const auto labels = mappedInput.mapLabels("« Home", "Open", "Up", "Down");
+  const auto labels = mappedInput.mapLabels("« Home", "Open", "Prev", "Next");
   GUI.drawButtonHints(renderer, labels.btn1, labels.btn2, labels.btn3, labels.btn4);
 
   renderer.displayBuffer();
--- a/src/activities/home/RecentBooksActivity.cpp
+++ b/src/activities/home/RecentBooksActivity.cpp
@@ -142,7 +142,7 @@ void RecentBooksActivity::render() const {
   }
 
   // Help text
-  const auto labels = mappedInput.mapLabels("« Home", "Open", "Up", "Down");
+  const auto labels = mappedInput.mapLabels("« Home", "Open", "Prev", "Next");
   GUI.drawButtonHints(renderer, labels.btn1, labels.btn2, labels.btn3, labels.btn4);
 
   renderer.displayBuffer();
--- a/src/activities/reader/EpubReaderChapterSelectionActivity.cpp
+++ b/src/activities/reader/EpubReaderChapterSelectionActivity.cpp
@@ -168,7 +168,7 @@ void EpubReaderChapterSelectionActivity::renderScreen() {
     renderer.drawText(UI_10_FONT_ID, indentSize, displayY, chapterName.c_str(), !isSelected);
   }
 
-  const auto labels = mappedInput.mapLabels("« Back", "Select", "Up", "Down");
+  const auto labels = mappedInput.mapLabels("« Back", "Select", "Prev", "Next");
   GUI.drawButtonHints(renderer, labels.btn1, labels.btn2, labels.btn3, labels.btn4);
 
   renderer.displayBuffer();
--- a/src/activities/reader/EpubReaderMenuActivity.cpp
+++ b/src/activities/reader/EpubReaderMenuActivity.cpp
@@ -139,7 +139,7 @@ void EpubReaderMenuActivity::renderScreen() {
   }
 
   // Footer / Hints
-  const auto labels = mappedInput.mapLabels("« Back", "Select", "Up", "Down");
+  const auto labels = mappedInput.mapLabels("« Back", "Select", "Prev", "Next");
   GUI.drawButtonHints(renderer, labels.btn1, labels.btn2, labels.btn3, labels.btn4);
 
   renderer.displayBuffer();
--- a/src/activities/reader/XtcReaderChapterSelectionActivity.cpp
+++ b/src/activities/reader/XtcReaderChapterSelectionActivity.cpp
@@ -172,7 +172,7 @@ void XtcReaderChapterSelectionActivity::renderScreen() {
 
   // Skip button hints in landscape CW mode (they overlap content)
   if (renderer.getOrientation() != GfxRenderer::LandscapeClockwise) {
-    const auto labels = mappedInput.mapLabels("« Back", "Select", "Up", "Down");
+    const auto labels = mappedInput.mapLabels("« Back", "Select", "Prev", "Next");
     GUI.drawButtonHints(renderer, labels.btn1, labels.btn2, labels.btn3, labels.btn4);
   }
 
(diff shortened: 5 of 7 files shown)


lukestein:
I am open to this, but it feels to me like the `src/activities/home/HomeActivity.cpp` change is fairly unambiguous (given the two-dimensional layout in the Lyra Home Screen), but everywhere else I think the buttons still map to an `up`/`down` idiom in an on-screen menu, right?

Assuming so, I might suggest we only change the home screen tips but not the others. But I don't feel strongly.

src/activities/home/MyLibraryActivity.cpp
```diff
@@ -205,7 +205,7 @@ void MyLibraryActivity::render() const {
   }
 
   // Help text
-  const auto labels = mappedInput.mapLabels("« Home", "Open", "Up", "Down");
+  const auto labels = mappedInput.mapLabels("« Home", "Open", "Prev", "Next");
```
lukestein, line 208:
```suggestion
  const auto labels = mappedInput.mapLabels("« Back", "Open", "Up", "Down");
```

lukestein: In Library (Browse Files), if we are in a folder then the first button actually goes back to the previous folder rather than home.

I suggest retaining up/down since they match the on-screen logic.

src/activities/home/RecentBooksActivity.cpp
```diff
@@ -142,7 +142,7 @@ void RecentBooksActivity::render() const {
   }
 
   // Help text
-  const auto labels = mappedInput.mapLabels("« Home", "Open", "Up", "Down");
+  const auto labels = mappedInput.mapLabels("« Home", "Open", "Prev", "Next");
```
lukestein, line 145:
```suggestion
  const auto labels = mappedInput.mapLabels("« Home", "Open", "Up", "Down");
```

lukestein: I suggest retaining up/down since they match the on-screen logic.

src/activities/reader/EpubReaderChapterSelectionActivity.cpp
```diff
@@ -168,7 +168,7 @@ void EpubReaderChapterSelectionActivity::renderScreen() {
     renderer.drawText(UI_10_FONT_ID, indentSize, displayY, chapterName.c_str(), !isSelected);
   }
 
-  const auto labels = mappedInput.mapLabels("« Back", "Select", "Up", "Down");
+  const auto labels = mappedInput.mapLabels("« Back", "Select", "Prev", "Next");
```
lukestein, line 171:
```suggestion
  const auto labels = mappedInput.mapLabels("« Back", "Select", "Up", "Down");
```

lukestein: I suggest retaining up/down since they match the on-screen logic.

src/activities/reader/EpubReaderMenuActivity.cpp
```diff
@@ -139,7 +139,7 @@ void EpubReaderMenuActivity::renderScreen() {
   }
 
   // Footer / Hints
-  const auto labels = mappedInput.mapLabels("« Back", "Select", "Up", "Down");
+  const auto labels = mappedInput.mapLabels("« Back", "Select", "Prev", "Next");
```
lukestein, line 142:
```suggestion
  const auto labels = mappedInput.mapLabels("« Back", "Select", "Up", "Down");
```

lukestein: I suggest retaining up/down since they match the on-screen logic.

src/activities/reader/XtcReaderChapterSelectionActivity.cpp
```diff
@@ -172,7 +172,7 @@ void XtcReaderChapterSelectionActivity::renderScreen() {
 
   // Skip button hints in landscape CW mode (they overlap content)
   if (renderer.getOrientation() != GfxRenderer::LandscapeClockwise) {
-    const auto labels = mappedInput.mapLabels("« Back", "Select", "Up", "Down");
+    const auto labels = mappedInput.mapLabels("« Back", "Select", "Prev", "Next");
```
lukestein, line 175:
```suggestion
    const auto labels = mappedInput.mapLabels("« Back", "Select", "Up", "Down");
```

lukestein: I suggest retaining up/down since they match the on-screen logic.

src/activities/settings/SettingsActivity.cpp
```diff
@@ -311,7 +311,7 @@ void SettingsActivity::render() const {
                     metrics.versionTextY, CROSSPOINT_VERSION);
 
   // Draw help text
-  const auto labels = mappedInput.mapLabels("« Back", "Toggle", "Up", "Down");
+  const auto labels = mappedInput.mapLabels("« Back", "Toggle", "Prev", "Next");
```
lukestein, line 314:
```suggestion
  const auto labels = mappedInput.mapLabels("« Home", "Toggle", "Up", "Down");
```

lukestein: In Settings, I believe the first button always goes to Home.

I suggest retaining up/down since they match the on-screen logic. This also helps disambiguate from the **previous**/**next** *tabs* (which are reached by selecting the tab bar and then clicking "Toggle").

I remain kind of unhappy with the label "Toggle" since the button does a mix of toggle (on some items), moving to the next tab (on the top tab selector row), clicking into a submenu (as in KOReader Sync or OPDS Browser), and choosing an item (as in Check for updates)

lukestein: We should really have something better than "Toggle" here, but I don't know what.

lukestein (changes requested):
I've made suggestions inline (which I admit go slightly beyond the title of this PR, but @maeveynot was nice to get these all laid out next to each other for us).

I realize that perhaps our worst button hint label (in terms of matching functionality) is the "Toggle" label in settings, but I don't know what's better in terms of capturing all the things that button is used for. I left
```
const auto labels = mappedInput.mapLabels("« Home", "Toggle", "Up", "Down");
```
as a suggestion but the "Toggle" is really not quite right.

Eloren1:
@lukestein Are you saying that Up/Down should be kept in some settings?

Prev/Next makes sense on the KOReader Sync page as well, because you cycle through the options. Otherwise, when the last option is selected, would clicking Down take you back to the first one?

image

The same applies to the Quick Menu. In Landscape CCW, the controls are inverted; having Up/Down labels upside down is more confusing than simply using Prev/Next.

image

Ultimately, Prev/Next allows you to abstract away from the physical orientation (Left/Right/Up/Down) and focus on the order.

osteotek:
@maeveynot With recent changes to Lyra theme doesn't look like this PR is relevant anymore. Please reopen if you have any further thoughts.

PR #738, opened 2026-02-07
feat: Add percentage support to CSS properties

### LINKED ISSUE #730: bug: Paragraph tab indents are only 1 space long
### Affected Version

v1.0.0rc2

### Describe the Bug

 —

1. No CSS

Image

—

2. With CSS

Image

### Steps to Reproduce

1. Extra Paragraph Spacing is OFF
2. Embedded Style is ON

### Expected Behavior

Display normal new paragraph intents just like in KOReader or how it worked before

```shell

```

jdk2pq:
## Summary
- Closes #730

- Adds percentage-based value support to CSS properties that accept percentages (padding, margin, text-indent)
 
- Adds `Percent` as another CSS unit
- Passes the viewport width to `fromCssStyle` so that we can resolve percentage-based values
- Adds a fallback of using an emspace for text-indent if we have an unresolvable value for whatever reason

## Additional Context

- This was missed in my CSS support feature, and the fallback when we encounter a percentage value is to use px instead. This means 5% (which would be ~30px on the screen) turns into 5px. When percentages are used in `text-indent`, this fallback behavior makes the indent look like a single space character. Whoops! 😬 

My test EPUB has been updated [here](https://github.com/jdk2pq/css-test-epub) with percentage based CSS values at the end of the book.

---

### AI Usage

Did you use AI tools to help write this code? _**YES**_, Claude Code

### DIFF
--- a/lib/Epub/Epub/blocks/BlockStyle.h
+++ b/lib/Epub/Epub/blocks/BlockStyle.h
@@ -64,8 +64,8 @@ struct BlockStyle {
   // Create a BlockStyle from CSS style properties, resolving CssLength values to pixels
   // emSize is the current font line height, used for em/rem unit conversion
   // paragraphAlignment is the user's paragraphAlignment setting preference
-  static BlockStyle fromCssStyle(const CssStyle& cssStyle, const float emSize,
-                                 const CssTextAlign paragraphAlignment, const uint16_t viewportWidth = 0) {
+  static BlockStyle fromCssStyle(const CssStyle& cssStyle, const float emSize, const CssTextAlign paragraphAlignment,
+                                 const uint16_t viewportWidth = 0) {
     BlockStyle blockStyle;
     const float vw = viewportWidth;
     // Resolve all CssLength values to pixels using the current font's em size and viewport width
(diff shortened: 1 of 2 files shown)


Eloren1:
image

Can you also rename `Embedded Style` to `Book Embedded Style` so it shows the relation to the book file more clearly? And it matches with the `Book's Style` as well. Thank you

jdk2pq:
> image
> Can you also rename `Embedded Style` to `Book Embedded Style` so it shows the relation to the book file more clearly? And it matches with the `Book's Style` as well. Thank you

@Eloren1 I added a separate PR for this [here](https://github.com/crosspoint-reader/crosspoint-reader/pull/746), since I didn't want to do two things at once in this PR.

Eloren1:
I appreciate your work with feedback, you managed to convince a skeptical user to enable this feature :))

Eloren1 (approved):
After clearing the cache, it looks the way it was intended

Image

jdk2pq:
@Eloren1 Very glad to hear it! 😄 Thank you for the feedback!

lukestein (approved):
Tested on device and seems to work well. Recommend we get this in 1.0.0 **if** we're doing another RC (everything should get publicly tested at least briefly), or else it's for 1.0.1.

Eloren1:
@daveallie yes please, so that the new Embedded Style rolls out in the final 1.0.0 without bugs from the past

Eloren1:
🔥🔥

PR #751, opened 2026-02-07
fix: Account for nbsp; character as whitespace

### LINKED ISSUE #743: Single-character prepositions merge with following words in some EPUB files
### Affected Version

0.16.0

### Describe the Bug

In some EPUB files (tested only 3 files, issue appears in 1), single-character Polish prepositions (i, w, z) incorrectly merge with the following word, creating malformed text.

Examples
- "znami" instead of "z nami" (with us)
- "wdomu" instead of "w domu" (at home)
- "ifioletową" instead of "i fioletową" (and purple)
- "iwmiejscu" instead of "i w miejscu" (and in place)

Additional context
- **The same EPUB file displays correctly on other readers:**
  - FBReader (iPhone)
  - Calibre (macOS)

- The issue does NOT appear to be affected by:
  - Text justification settings
  - Hyphenation settings
- Multiple prepositions can be affected in sequence (e.g., "i w" → "iw")

### Steps to Reproduce

Open an affected Polish-language EPUB file and observe merged prepositions throughout the text.

### Expected Behavior

Single-character prepositions should be separated from following words by a space.
Examples
- "z nami" instead of "znami"  (with us)
- "w domu" instead of "wdomu" (at home)
- "i fioletową" instead of "ifioletową" (and purple)
- "i w miejscu" instead of "iwmiejscu" (and in place)

```shell

```

jdk2pq:
## Summary

- Fixed an issue with `nbsp;` not being counted as whitespace, which caused words to incorrectly join together

- Added separate handling for `nbsp;` characters similar to how we already handle BOM characters

## Additional Context

- Closes #743 
- Updated my test EPUB [here](https://github.com/jdk2pq/css-test-epub) with examples of this at the end of the book

---

### AI Usage

Did you use AI tools to help write this code? _**YES*_, Claude Code

### DIFF
--- a/lib/Epub/Epub/parsers/ChapterHtmlSlimParser.cpp
+++ b/lib/Epub/Epub/parsers/ChapterHtmlSlimParser.cpp
@@ -360,8 +360,6 @@ void XMLCALL ChapterHtmlSlimParser::characterData(void* userData, const XML_Char
     }
 
     // Treat Non-Breaking Space (U+00A0) = 0xC2 0xA0 as a word separator
-    // EPUBs use &nbsp; between short prepositions and following words (e.g. Polish "w domu")
-    // The nbsp should render as a visible space, not merge words together (issue #743)
     const XML_Char NBSP_BYTE_1 = static_cast<XML_Char>(0xC2);
     const XML_Char NBSP_BYTE_2 = static_cast<XML_Char>(0xA0);
     if (s[i] == NBSP_BYTE_1 && (i + 1 < len) && s[i + 1] == NBSP_BYTE_2) {


jdk2pq:
One thing worth considering before merging this is _if_ this is the behavior we want for `nbsp;` characters, because I definitely also see a case for using these characters as intended and not breaking lines on them in the line-breaking algorithm. It wouldn't be too difficult of an addition to add this behavior, but the reason I didn't do this initially is because we're already so limited in screen real estate that it felt like the better tradeoff to just treat `nbsp;` as regular ol' whitespace. Happy to consider other opinions on this though, and if there's consensus on the other solution, I can close this and put up a PR for that instead.

lukestein:
I'm also going to tag in @osteotek since he and I both typically have opinions about line breaking and hyphenation-related stuff ;)

My inclination is that a non-breaking space **absolutely should** be a non-breaking space. I know there are files that overuse them, but publishers are often (and should be) using them intentionally and it should be respected. And often when nbsp are used interword, the word on one side or the other is short so it doesn't screw up justification etc. too much, e.g.,
- `7:00~p.m.`
- `Marcus~Jr.`
- `C.~S. Lewis`
- `see p.~27`
- `JPY~4,000`
- `65°~C`
- or apparently single-letter prepositions in Polish as in OP's issue

osteotek:
I think this is not as simple as treating nbsp as whitespace. Nbsp should be treated as literally non breaking whitespace - space to delimit words, but should not break lines. We're not as limited in width space to break them, like in examples provided by @lukestein

jdk2pq:
@lukestein @osteotek Makes sense to me! I'll close this and open a separate PR for that shortly. Thanks for the feedback!

PR #759, opened 2026-02-08
feat: Take screenshots

el:
## Summary

* **What is the goal of this PR?** Implements a take-screenshot feature

- Quick press Power button and Down button at the same time to take a screenshot
- Screenshots are saved in `screenshots` folder

## Additional Context

- Currently it does not use the device orientation.

---

Example screenshots:

screenshot-6771.bmp
[screenshot-6771.bmp](https://github.com/user-attachments/files/25157071/screenshot-6771.bmp)

screenshot-14158.bmp
[screenshot-14158.bmp](https://github.com/user-attachments/files/25157073/screenshot-14158.bmp)

### AI Usage

Did you use AI tools to help write this code? _** YES

### DIFF
--- a/USER_GUIDE.md
+++ b/USER_GUIDE.md
@@ -41,7 +41,7 @@ Button layout can be customized in **[Settings](#35-settings)**.
 ### Taking a Screenshot
 When the Power Button and Volume Down button are pressed at the same time, it will take a screenshot and save it in the folder `screenshots/`.
 
-Alternatively, while reading a book, hold the **Confirm** button to open the reader menu and select **Take screenshot**.
+Alternatively, while reading a book, press the **Confirm** button to open the reader menu and select **Take screenshot**.
 
 ---
 


CaptainFrito:
Very cool idea, would it be a good idea to add a visual indication that a screenshot was taken?

el:
> Very cool idea, would it be a good idea to add a visual indication that a screenshot was taken?

Done

CaptainFrito:
Hey @el, sorry to nitpick but if you have the setting "Short power button click" set to "sleep", when trying to take screenshots the device goes to sleep if you're not careful to press both buttons exactly at the same time. Not sure that's a big deal, I still managed to take a few screenshots.

lib/GfxRenderer/BitmapHelpers.cpp
```diff
@@ -104,3 +106,48 @@ uint8_t quantize1bit(int gray, int x, int y) {
...
   return (gray >= adjustedThreshold) ? 1 : 0;
 }
+
+uint8_t* createBmpHeader(int width, int height, uint32_t* headerSize) {
+  *headerSize = sizeof(BmpHeader);
+  uint32_t rowSize = (width + 31) / 32 * 4;
+  uint32_t imageSize = rowSize * height;
+  uint32_t fileSize = sizeof(BmpHeader) + imageSize;
+
+  BmpHeader* bmpHeader = new BmpHeader();
```
ngxson: Nits, but it's recommended to initalize this using list initializer (the {} version), as it makes the compiler easier to optimize

lib/FsHelpers/FsHelpers.cpp
```diff
@@ -37,3 +41,74 @@ std::string FsHelpers::normalisePath(const std::string& path) {
...
+  }
+
+  FsFile file;
+  if (!SdMan.openFileForWrite("SCR", filename, file)) {
+    Serial.printf("[%lu] [SCR] Failed to open file for writing\n", millis());
+    return false;
+  }
+
+  uint32_t headerSize = 0;
+  uint8_t* bmpHeader = createBmpHeader(width, height, &headerSize);
```
ngxson: Stack-allocated is preferable here, I think something like this will make it looks cleaner:

BmpHeader header;
createBmpHeader(&header, width, height);

write(&header, sizeof(header));

lib/FsHelpers/FsHelpers.cpp
```diff
@@ -37,3 +41,74 @@ std::string FsHelpers::normalisePath(const std::string& path) {
...
+  }
+
+  const uint32_t rowSize = (width + 31) / 32 * 4;
+  const uint32_t fbRowSize = width / 8;
+  const uint32_t paddingSize = rowSize - fbRowSize;
+  uint8_t padding[4] = {0, 0, 0, 0};
+
+  for (int y = 0; y < height; y++) {
+    const uint8_t* fbRow = framebuffer + (height - 1 - y) * fbRowSize;
+    if (file.write(fbRow, fbRowSize) != fbRowSize) {
```
ngxson: An extra idea: you can rotate the image on-the-fly by allocating a row of pixels on the stack, then fill it.

The shorter edge of the screen is 480 pixels, with 1bpp that means 60 bytes to be allocated, we should always have enough memory for that.

lib/FsHelpers/FsHelpers.cpp
```diff
@@ -37,3 +41,74 @@ std::string FsHelpers::normalisePath(const std::string& path) {
...
+    Serial.printf("[%lu] [SCR] Failed to open file for writing\n", millis());
+    return false;
+  }
+
+  uint32_t headerSize = 0;
+  uint8_t* bmpHeader = createBmpHeader(width, height, &headerSize);
+  if (!bmpHeader) {
+    file.close();
+    SdMan.remove(filename);
+    return false;
```
embedded4ever: In case of code hits here, memory leak will be happen for `bmpHeader`, make sure you release the memory before return.

Eloren1:
Can you please add `Take screenshot` button to a quick reader menu as well? (hold ok to open)

src/main.cpp
```diff
@@ -273,6 +274,22 @@ void setupDisplayAndFonts() {
...
 }
 
+void takeScreenshot() {
+  const uint8_t* fb = display.getFrameBuffer();
+  if (!fb) {
+    Serial.println("[SCR] Framebuffer not available");
+    return;
+  }
+
+  String filename_str = "/screenshots/screenshot-" + String(millis()) + ".bmp";
```
bramschulting: At some point `millis()` wraps around. This results in the (incredibly small) chance of the same name being used twice, but more importantly the screenshots will potentially be out of order. Perhaps we can store the last 'screenshot number' as a `uint16` in `CrossPointState`? That would give room for 65,535 screenshots before this becomes a problem.

Eloren1: > That would give room for 65,535 screenshots before this becomes a problem.

wrapped `screenshot number` + millis = infinite amount of screenshots

el: I would have used date+time but x4 doesnt have an rtc :)

I think this is enough for now

el:
> Can you please add `Take screenshot` button to a quick reader menu as well? (hold ok to open)

I dont see how it would be useful, you mean take a screenshot after closing the menu?

If so, I think we can do that after this is merged.

lib/GfxRenderer/BitmapHelpers.cpp
el: done

lib/FsHelpers/FsHelpers.cpp
el: done

lib/FsHelpers/FsHelpers.cpp
```diff
@@ -1,8 +1,14 @@
 #include "FsHelpers.h"
 
+#include "Bitmap.h" // Required for BmpHeader struct definition
+#include <BitmapHelpers.h>
+#include <GfxRenderer.h>
+#include <SDCardManager.h>
```
ngxson: note that we migrated to `HalStorage.h` on master branch, you should replace this with the new header and replace `SdMan` with the `Storage macro`

lib/FsHelpers/FsHelpers.cpp
ngxson: would you mind if I submit a patch here to rotate the bitmap on-the-fly?

el: not at all, feel free to update

Eloren1:
> > Can you please add `Take screenshot` button to a quick reader menu as well? (hold ok to open)
> 
> I dont see how it would be useful, you mean take a screenshot after closing the menu?
> 
> If so, I think we can do that after this is merged.

Yes after closing. So it is possible to do one handed. And more noticeable

lib/FsHelpers/FsHelpers.cpp
```diff
@@ -81,23 +86,28 @@ bool FsHelpers::saveFramebufferAsBmp(const char* filename, const uint8_t* frameb
...
-  const uint32_t rowSize = (width + 31) / 32 * 4;
-  const uint32_t fbRowSize = width / 8;
-  const uint32_t paddingSize = rowSize - fbRowSize;
+  const uint32_t rowSizePadded = (phyWidth + 31) / 32 * 4;
   uint8_t padding[4] = {0, 0, 0, 0};
 
-  for (int y = 0; y < height; y++) {
-    const uint8_t* fbRow = framebuffer + (height - 1 - y) * fbRowSize;
-    if (file.write(fbRow, fbRowSize) != fbRowSize) {
+  // rotate the image 90d counter-clockwise on-the-fly while writing to save memory
```
el: @ngxson  Although most people use it in portrait, it would be better if we actually use `GfxRenderer::Orientation` and rotate accordingly

ngxson: I think it can be left as a future feature though, otherwise you can also extend my code to add the missing orientations.

Having an abstraction like `fillOutputRow(uint8_t* buffer, Orientation o)` can already be useful (`buffer` size matches either width or height depending on current orientation)

osteotek:
I’ve been testing this and found that setting the “Short power button click” to “sleep” actually conflicts with the combination for taking screenshots. And overall, I believe the current input manager isn’t very well-suited for reading multiple key combinations at once. Perhaps it would be better to add an action for taking screenshots to the reader menu?

el:
> I’ve been testing this and found that setting the “Short power button click” to “sleep” actually conflicts with the combination for taking screenshots. And overall, I believe the current input manager isn’t very well-suited for reading multiple key combinations at once. Perhaps it would be better to add an action for taking screenshots to the reader menu?

@Eloren1 has also suggested that change. I have implemented the Take Screenshot to the menu and also moved all logic to a new util.

I have also increased sleep duration slightly to avoid conflicts.

osteotek:
Thank you for the changes! I guess my question is, do we really need button combination for taking screenshots? Now that we have menu option, I am not sure it is that much required, considering we have to change existing logic to properly support it. @crosspoint-reader/firmware-maintainers what do you think

el:
> Thank you for the changes! I guess my question is, do we really need button combination for taking screenshots? Now that we have menu option, I am not sure it is that much required, considering we have to change existing logic to properly support it. @crosspoint-reader/firmware-maintainers what do you think

I think it is still usefull, I imagine this feature could be used when
- You want to share your home screen or recents
- Someone creates a PR that affects the UI, they can share its screenshot.

osteotek:
> > Thank you for the changes! I guess my question is, do we really need button combination for taking screenshots? Now that we have menu option, I am not sure it is that much required, considering we have to change existing logic to properly support it. @crosspoint-reader/firmware-maintainers what do you think
> 
> I think it is still usefull, I imagine this feature could be used when
> 
> * You want to share your home screen or recents
> * Someone creates a PR that affects the UI, they can share its screenshot.

Absolutely, I am not talking about screenshot feature in general, it is useful, thank you for this btw. I'm just not sure if I like button overloading. But let's wait for other maintainers opinion

ngxson:
Personally I think it's still be useful for development and bug reports in general. The implementation introduced in this PR is easy enough, so I think it's worth merging into master.

Re. problem with key combination, one idea is to add a new dialog that will be shown when holding power button, and we can move the "screenshot" to that dialog menu. Quite the same idea as on android (see screenshot below). Probably worth discussing this as an UI feature. 

image

osteotek:
@el apologies, I forgot to rereview your PR after changes, let me do that now

el:
> @el apologies, I forgot to rereview your PR after changes, let me do that now

np, there were some issues related to merge conflicts (I trusted antigravity and it kinda failed me ◡̈ )

Should be ready for review

osteotek:
@el if I press screenshot combination while page is still rendering I get crash with this error. Might be related to issue Copilot suggested

`assert failed: xQueueGenericSend queue.c:832 (pxQueue->pcHead != ((void *)0) || pxQueue->u.xSemaphore.xMutexHolder == ((void *)0) || pxQueue->u.xSemaphore.xMutexHolder == xTaskGetCurrentTaskHandle())`

src/main.cpp
```diff
@@ -404,6 +405,21 @@ void loop() {
...
+  static bool screenshotButtonsReleased = true;
+  if (gpio.isPressed(HalGPIO::BTN_POWER) && gpio.isPressed(HalGPIO::BTN_DOWN)) {
+    if (screenshotButtonsReleased) {
+      screenshotButtonsReleased = false;
+      // Take a screenshot by dumping the current framebuffer over serial.
+      logSerial.printf("SCREENSHOT_START:%d\n", HalDisplay::BUFFER_SIZE);
+      uint8_t* buf = display.getFrameBuffer();
+      logSerial.write(buf, HalDisplay::BUFFER_SIZE);
+      logSerial.printf("SCREENSHOT_END\n");
+    }
```
ngxson: Unless I missed something, this dumps the screenshot to serial, but does not save it to sdcard.

You should use a render lock to avoid render function accessing the framebuffer at the same

```suggestion
      if (currentActivity) {
        Activity::RenderLock lock(currentActivity);
        ScreenshotUtil::takeScreenshot(renderer);
      }
    }
```

src/activities/reader/EpubReaderActivity.cpp
```diff
@@ -431,6 +432,15 @@ void EpubReaderActivity::onReaderMenuConfirm(EpubReaderMenuActivity::MenuAction
...
     }
+    case EpubReaderMenuActivity::MenuAction::SCREENSHOT: {
+      {
+        RenderLock lock(*this);
+        pendingScreenshot = true;
+      }
+      exitActivity();
+      requestUpdate();
+      break;
+    }
```
ngxson, line 443: This is likely redundant I think, the button combination should already be enough

src/main.cpp
```diff
@@ -404,6 +405,21 @@ void loop() {
     powerManager.setPowerSaving(false);  // Restore normal CPU frequency on user activity
   }
 
+  static bool screenshotButtonsReleased = true;
+  if (gpio.isPressed(HalGPIO::BTN_POWER) && gpio.isPressed(HalGPIO::BTN_DOWN)) {
+    if (screenshotButtonsReleased) {
+      screenshotButtonsReleased = false;
```
ngxson, line 411: the next loop, this will be flipped back to `true`, so this logic doesn't serve any real purpose.

instead, you should debounce it by keeping track of last `millis()` and for example, only allow one screenshot per 2 seconds

src/main.cpp
osteotek: Correct, Copilot's suggestion is not valid here.

src/activities/reader/EpubReaderActivity.cpp
el: this is for the menu item, i didnt implement the menu item at first, but got multiple feedback asking for it

osteotek: Button combination is less discoverable, and conflicts with Going to sleep action sometimes

src/main.cpp
el: This prevents multiple fire of the takescreenshot function, regardless of how long it takes to take a screenshot. So untill you let go and press again, the loop wont call takescreenshot function. Let me know if i am mistaken.

ngxson: hmm ok then, it looks good

src/CrossPointSettings.h
```diff
@@ -181,7 +181,7 @@ class CrossPointSettings {
   static CrossPointSettings& getInstance() { return instance; }
 
   uint16_t getPowerButtonDuration() const {
-    return (shortPwrBtn == CrossPointSettings::SHORT_PWRBTN::SLEEP) ? 10 : 400;
+    return (shortPwrBtn == CrossPointSettings::SHORT_PWRBTN::SLEEP) ? 80 : 400;
```
osteotek: let's set it back to 10
```suggestion
    return (shortPwrBtn == CrossPointSettings::SHORT_PWRBTN::SLEEP) ? 10 : 400;
```

USER_GUIDE.md
```diff
@@ -38,6 +38,11 @@ The device utilises the standard buttons on the Xtink X4 (in the same layout as
 
 Button layout can be customized in **[Settings](#35-settings)**.
 
+### Taking a Screenshot
+When the Power Button and Volume Down button are pressed at the same time, it will take a screenshot and save it in the folder `screenshots/`.
+
+Alternatively, while reading a book, hold the **Confirm** button to open the reader menu and select **Take screenshot**.
```
osteotek:
```suggestion
Alternatively, while reading a book, press the **Confirm** button to open the reader menu and select **Take screenshot**.
```

el:
@daveallie fyi

PR #760, opened 2026-02-08
fix: Add EPUB 3 cover image detection

itsthisjustin:
I had an epub that just showed a blank cover and wouldnt work for the sleep screen either, turns out it was an epub3 and I guess we didn't support that. Super simple fix here

### DIFF
--- a/lib/Epub/Epub/parsers/ContentOpfParser.cpp
+++ b/lib/Epub/Epub/parsers/ContentOpfParser.cpp
@@ -232,6 +232,14 @@ void XMLCALL ContentOpfParser::startElement(void* userData, const XML_Char* name
         Serial.printf("[%lu] [COF] Found EPUB 3 nav document: %s\n", millis(), href.c_str());
       }
     }
+
+    // EPUB 3: Check for cover image (properties contains "cover-image")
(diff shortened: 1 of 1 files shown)


lib/Epub/Epub/parsers/ContentOpfParser.cpp
```diff
@@ -232,6 +232,14 @@ void XMLCALL ContentOpfParser::startElement(void* userData, const XML_Char* name
         Serial.printf("[%lu] [COF] Found EPUB 3 nav document: %s\n", millis(), href.c_str());
       }
     }
+
+    // EPUB 3: Check for cover image (properties contains "cover-image")
+    if (!properties.empty() && self->coverItemHref.empty()) {
+      if (properties == "cover-image" || properties.find("cover-image ") == 0 ||
```
el, line 238: Is the empty space after cover-image by mistake?

itsthisjustin: > Is the empty space after cover-image by mistake?

No apparently epub 3 uses space separated values for properties so this is an easy way to catch the cover image no matter where it shows in the list

PR #762, opened 2026-02-08
fix(epub): avoid crashes when CSS cache is missing

### LINKED ISSUE #761: Fix crash when EPUB CSS cache missing + avoid repeated cover thumb generation
## What happens

On ESP32-C3 devices, opening some EPUBs can crash (abort/reboot loop) when `/.crosspoint/epub_<hash>/css_rules.cache` is missing, because we attempt to parse CSS on-demand and hit an exception/OOM.

Separately, Home can periodically pause showing a Loading popup because it keeps retrying thumbnail generation when the EPUB cover JPEG is unsupported (progressive JPEG), so the thumb file never gets created.

## Repro

Crash:
1. Have an EPUB with cached spine/TOC metadata (e.g. `/.crosspoint/epub_<hash>/book.bin` exists), but delete/miss `/.crosspoint/epub_<hash>/css_rules.cache`
2. Open the book in Reader
3. Firmware logs:
   - `[EBP] File does not exist: .../css_rules.cache`
   - `[EBP] Warning: CSS rules cache not found, attempting to parse CSS files`
   - `[EBP] Parsing content.opf: ...`
   - then `abort() was called ...` and reboot

Home pause:
- Logs show repeated `[HOME] File does not exist: .../thumb_400.bmp` then thumb generation:
  - `[EBP] Generating thumb BMP from JPG cover image`
  - `[JPG] JPEG decode init failed with error code: 37`
  - `[EBP] Failed to generate thumb BMP ...`
  This repeats on subsequent Home renders causing a visible pause.

## Expected

- Reader should fail open (continue without CSS) instead of abort/reboot
- Home should not repeatedly retry thumbnail generation when the cover is unsupported

## Notes

- picojpeg error `37` corresponds to `PJPG_UNSUPPORTED_MODE` (progressive JPEG not supported).

DChells:
## Summary

Prevents EPUB reader reboot loops when `css_rules.cache` is missing, and stops Home from repeatedly retrying cover thumbnail generation for unsupported EPUB cover images.

Fixes #761.

## Problem

Two issues that show up most often on ESP32-C3 with large EPUBs:

1) CSS cache miss can crash the reader
- If `/.crosspoint/epub_<hash>/css_rules.cache` is missing, we try to parse CSS on-demand.
- On large books / tight heap this can throw/abort and the device ends up reboot-looping.

2) Home periodically pauses with a "Loading..." popup
- Home attempts to generate `thumb_<height>.bmp` for recent book covers.
- Some EPUB covers are progressive JPEGs, which picojpeg doesn’t support (error `37` / `PJPG_UNSUPPORTED_MODE`).
- Since the thumb never appears, Home keeps retrying on subsequent renders → periodic pauses.

## Solution

- CSS cache miss: fail open (continue without CSS) instead of crashing
  - Skip CSS parsing on cache-miss when the book is large or free heap is low
  - Guard best-effort CSS parsing so exceptions don’t crash the device
- Home thumbnails:
  - If JPEG decode fails, write a 0-byte thumb sentinel so Home stops retrying forever
  - Lyra theme: when cover is missing/invalid, render a simple placeholder with the book title in the cover box
- Metadata reads / recents migration: always skip CSS parsing (metadata only)

## Notes / tradeoffs

- On books that hit the “skip CSS parse” path, rendering may be less styled (we intentionally skip building CSS rules to prioritize stability over perfect styling).
- For progressive JPEG covers, we intentionally stop retrying; users can convert the cover to baseline JPEG if they want the thumb.

## Testing

On hardware (Xteink X4 / ESP32-C3):
- Delete `/.crosspoint/epub_<hash>/css_rules.cache` for a large EPUB, then open it → should not reboot; should log skipping CSS parse
- Return to Home repeatedly → should not keep retrying thumb generation for the same book/thumb size
- Verify Lyra Home shows title fallback when cover thumb is missing/invalid

## Files changed

- `lib/Epub/Epub.cpp`
- `src/RecentBooksStore.cpp`
- `src/components/themes/lyra/LyraTheme.cpp`

### DIFF
--- a/lib/Epub/Epub.cpp
+++ b/lib/Epub/Epub.cpp
@@ -287,11 +287,17 @@ bool Epub::load(const bool buildIfMissing, const bool skipLoadingCss) {
     if (!skipLoadingCss && !loadCssRulesFromCache()) {
       Serial.printf("[%lu] [EBP] Warning: CSS rules cache not found, attempting to parse CSS files\n", millis());
       // to get CSS file list
-      if (!parseContentOpf(bookMetadataCache->coreMetadata)) {
-        Serial.printf("[%lu] [EBP] Could not parse content.opf from cached bookMetadata for CSS files\n", millis());
-        // continue anyway - book will work without CSS and we'll still load any inline style CSS
+      try {
+        if (!parseContentOpf(bookMetadataCache->coreMetadata)) {
+          Serial.printf("[%lu] [EBP] Could not parse content.opf from cached bookMetadata for CSS files\n", millis());
+          // continue anyway - book will work without CSS and we'll still load any inline style CSS
+        }
+        parseCssFiles();
+      } catch (...) {
+        // On ESP32, some STL allocations may throw and abort if uncaught.
+        // CSS parsing is optional, so prefer failing open.
+        Serial.printf("[%lu] [EBP] Exception while parsing CSS files; continuing without CSS\n", millis());
       }
-      parseCssFiles();
     }
     Serial.printf("[%lu] [EBP] Loaded ePub: %s\n", millis(), filepath.c_str());
     return true;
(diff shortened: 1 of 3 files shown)


itsthisjustin:
Closing this and the related PR as it appears to no longer be an issue

PR #764, opened 2026-02-08
chore: Add unit test infrastructure and CI integration

BlindBat:
## Summary

* **What is the goal of this PR?** Add native C++ unit tests and performance benchmarks that run off-device, giving contributors fast feedback on correctness without needing hardware.

  - **Test stubs** (`test/stubs/`) — Minimal Arduino/hardware header stubs (`HardwareSerial.h`, `SdFat.h`, `Print.h`, `WString.h`, `SDCardManager.h`) that let library code compile on a regular desktop compiler via `-I` path ordering.
  - **Test harness** (`test/test_harness.h`) — Lightweight assertion macros (`ASSERT_TRUE`, `ASSERT_EQ`, `ASSERT_NEAR`, etc.) with pass/fail counting and summary reporting. No external framework dependencies.
  - **7 unit test suites** (`test/unit/`) covering the project's pure-logic libraries:

    | Suite | What's tested | Assertions |
    |---|---|---|
    | `Utf8Test` | `utf8NextCodepoint` (ASCII, 2/3/4-byte, invalid), `utf8RemoveLastChar`, `utf8TruncateChars` | 23 |
    | `UrlUtilsTest` | `isHttpsUrl`, `ensureProtocol`, `extractHost`, `buildUrl` | 20 |
    | `StringUtilsTest` | `sanitizeFilename` (invalid chars, length limits, fallback), `checkFileExtension` (case-insensitive) | 16 |
    | `SerializationTest` | `writePod`/`readPod`/`writeString`/`readString` roundtrips via `std::stringstream` | 8 |
    | `BlockStyleTest` | `fromCssStyle`, `getCombinedBlockStyle`, inset calculations | 23 |
    | `CssParserTest` | `parseInlineStyle` (all properties, shorthands), `CssLength::toPixels`, `applyOver`, `resolveStyle` cascade | 63 |
    | `OpdsParserTest` | Navigation/acquisition feeds, namespace prefixes, chunked writes, malformed XML, `clear()` | 31 |

  - **2 performance benchmarks** (`test/perf/`) — `HyphenationBench` (throughput over 30 English words) and `CssParserBench` (1/5/15-property complexity levels). Informational only, never fail CI.
  - **Shell scripts** — `test/run_unit_tests.sh` compiles and runs all test binaries, reports pass/fail summary. `test/run_perf_tests.sh` compiles and runs benchmarks. Both follow the pattern established by `test/run_hyphenation_eval.sh`.
  - **`CssParser::loadFromString()`** — New public method that parses CSS from a `std::string` instead of `FsFile`. Enables testing `resolveStyle()` cascade logic without SD card I/O.
  - **CI integration** — New `unit-tests` job in `.github/workflows/ci.yml` added to the `test-status` gate. Very lightweight (just `checkout` + `run_unit_tests.sh`, no PlatformIO install needed).

## Additional Context

- All 7 suites pass: **184 assertions, 0 failures**
- Firmware build is unaffected — `pio run` succeeds with identical RAM/Flash usage
- `clang-format` passes on all new files
- The `unit-tests` CI job is the fastest in the pipeline (~10-15s) since it only needs a C++ compiler, no PlatformIO or apt installs
- The test stubs and harness approach follows the existing `test/hyphenation_eval/` pattern — plain C++ with bash compilation, no external test frameworks

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -90,6 +90,16 @@ jobs:
           path: .pio/build/default/firmware.bin
           if-no-files-found: error
 
+  unit-tests:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+        with:
+          submodules: recursive
+
+      - name: Run unit tests
+        run: ./test/run_unit_tests.sh
+
   # This job is used as the PR required actions check, allows for changes to other steps in the future without breaking
   # PR requirements.
   test-status:
@@ -98,6 +108,7 @@ jobs:
       - build
       - clang-format
       - cppcheck
+      - unit-tests
     if: always()
     runs-on: ubuntu-latest
     steps:
--- a/lib/Epub/Epub/css/CssParser.cpp
+++ b/lib/Epub/Epub/css/CssParser.cpp
@@ -472,6 +472,20 @@ bool CssParser::loadFromStream(FsFile& source) {
   return true;
 }
 
+// Load and parse CSS from a string
+
+void CssParser::loadFromString(const std::string& content) {
+  if (content.empty()) return;
+
+  const std::string cleaned = stripComments(content);
+
+  size_t pos = 0;
+  std::string selector, body;
+  while (extractNextRule(cleaned, pos, selector, body)) {
+    processRuleBlock(selector, body);
+  }
+}
+
 // Style resolution
 
 CssStyle CssParser::resolveStyle(const std::string& tagName, const std::string& classAttr) const {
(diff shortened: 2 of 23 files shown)


daveallie (commented):
Following the merge of https://github.com/crosspoint-reader/crosspoint-reader/pull/656, I think we should be able to remove the `SdFat` and `SDCardManager` stubs and replace them with a `HalStorage` stub

ngxson:
For visibility, I left a comment expressing my concerns in https://github.com/crosspoint-reader/crosspoint-reader/pull/765#pullrequestreview-3773119362

I think it would be useful to track the memory usage in testing too.

For this PR, I think most of the stubs can be removed, but I would also like to see if it can be overlapped with https://github.com/crosspoint-reader/crosspoint-reader/pull/715 , where I implement an emulated version of `HalStorage`

PR #755, opened 2026-02-07
feat: Add FB2 format support

BlindBat:
Resolves #592

## Summary

- Add full reading support for FictionBook (`.fb2`) files — an XML-based ebook format popular in Russian-speaking countries
- New `lib/Fb2/` library with expat-based streaming parsers for metadata extraction, section content rendering, and base64 cover image extraction
- New `Fb2ReaderActivity` and `Fb2ReaderChapterSelectionActivity` for reading and chapter navigation
- Registered `.fb2` across all UI touchpoints: file browser, recent books, home screen covers, sleep screen, web server cache clearing

**Note:** Image support (inline images, ImageUtils, PngToBmpConverter) has been removed from this PR per reviewer feedback. Image rendering will be addressed in a separate follow-up PR.

## Details

**Library (`lib/Fb2/`)**
- `Fb2` — main class with binary metadata caching, section navigation, progress calculation
- `Fb2MetadataParser` — first-pass parser extracting title-info, author, language, cover ID, and section boundaries with file offsets
- `Fb2SectionParser` — content parser converting FB2 tags (`<p>`, `<title>`, `<emphasis>`, `<strong>`, `<poem>`, `<cite>`, `<epigraph>`, etc.) into Pages using the existing Epub rendering pipeline (ParsedText/TextBlock/Page)
- `Fb2CoverExtractor` — streams base64-encoded `<binary>` elements, decodes to JPEG, converts to BMP via JpegToBmpConverter
- `Fb2Section` — section cache manager with binary LUT for random page access

**Reader Activities**
- `Fb2ReaderActivity` — full reader with FreeRTOS display task, orientation support, progress save/restore, percent jump, status bar, text anti-aliasing. Reuses `EpubReaderMenuActivity` directly
- `Fb2ReaderChapterSelectionActivity` — TOC-based chapter selection from section titles

**Touchpoint Registration**
- `ReaderActivity` — FB2 file detection and dispatch
- `MyLibraryActivity` — `.fb2` in file browser filter
- `RecentBooksStore` — FB2 metadata extraction for recent books
- `HomeActivity` — FB2 cover thumbnail generation
- `SleepActivity` — FB2 cover for sleep screen
- `CrossPointWebServer` — FB2 cache clearing on upload/delete/rename

## AI Usage

YES — Code was generated with AI assistance (Claude). All code was reviewed and tested on device.

## Test plan

- [ ] Place `.fb2` files on SD card, verify they appear in Browse Files
[...]

PR #765, opened 2026-02-08
perf: Optimize CssParser and hyphenation allocation

BlindBat:
> **Stacked on #764** — This PR includes the unit test commit from #764 as its base. Please review and merge #764 first, then this PR's diff will show only the optimization changes.

## Summary

* **What is the goal of this PR?** Reduce CPU time and heap allocation churn in the two hottest library paths — CSS inline style parsing and hyphenation word breaking — without changing any user-facing behavior.

### CssParser — 2.3-3.8x faster

| Benchmark (desktop, -O2) | Before | After | Speedup |
|---|---|---|---|
| `parseInlineStyle` — 1 property | 237 ns | 62 ns | **3.8x** |
| `parseInlineStyle` — 5 properties | 979 ns | 390 ns | **2.5x** |
| `parseInlineStyle` — 15 properties | 3,244 ns | 1,386 ns | **2.3x** |

**What changed:**
- **Eliminated triple normalization** — Previously, `splitOnChar(';')` normalized the full string, then `parseDeclarations` normalized each half again, and every interpreter (`interpretAlignment`, `interpretLength`, etc.) called `normalized()` a third time. Now a single `normalizeInto()` helper fills reusable buffers once per property, and interpreters receive pre-normalized input.
- **Zero-allocation length parsing** — `interpretLength()` and `interpretSpacing()` now use `strtof()` directly on the input string and compare unit characters inline, instead of creating substring allocations for the numeric part and unit part separately.
- **Inline declaration scanning** — `parseDeclarations()` scans for `;` and `:` boundaries in a single pass rather than calling `splitOnChar` (which allocated a `vector<string>` of N normalized strings).
- **Inline selector scanning** — `processRuleBlock()` scans for `,` boundaries inline with `normalizeInto()` instead of `splitOnChar`.

### Hyphenation — ~17% faster

| Benchmark (desktop, -O2) | Before | After | Speedup |
|---|---|---|---|
| `breakOffsets` per-word avg | 570 ns | 489 ns | **1.17x** |

**What changed:**
- **Bulk erase** — `trimSurroundingPunctuationAndFootnote()` replaced an O(n²) erase-from-front loop (each `erase(begin())` shifts all remaining elements) with a single bulk `erase(begin, begin + count)`.
- **Static buffer reuse** — Since ESP32-C3 is single-threaded, `Hyphenator::breakOffsets()`, `liangBreakIndexes()`, and `collectCodepoints()` now reuse `static` local vectors that retain their allocated capacity across calls, eliminating ~8 heap allocations per word. Added a `collectCodepoints(word, vec&)` overload for this purpose.

### Correctness verification

Two new regression test suites specifically target every optimized code path:

| Suite | Assertions | What it verifies |
|---|---|---|
| `CssParserRegressionTest` | 502 | Whitespace normalization (tabs, newlines, multiple spaces), case insensitivity, trailing/double semicolons, missing values/colons, all 4 CSS unit types, negative/decimal/zero lengths, grouped selector comma scanning, 15-property stress test, 100x repeated-call leakage test |
| `HyphenationRegressionTest` | 122 | Golden byte-offset values for 17 English words, `collectCodepoints` overload equivalence, buffer reuse safety, bulk-erase edge cases (all-punctuation, single char, footnotes, empty), 500-call determinism check, interleaved short/long word static-buffer test, explicit/soft hyphen detection |

Combined with the 7 existing suites from #764: **9 suites, 808 total assertions, all passing.**

## Additional Context

- Firmware builds successfully with identical RAM/Flash footprint — these are pure algorithmic changes
- `clang-format` passes on all modified files
- The allocation-reuse optimization will have an even larger impact on the actual ESP32-C3 target, where `malloc`/`free` is significantly more expensive than on desktop
- The `static` local pattern is safe because the ESP32-C3 is single-core with no multi-threading

### DIFF
--- a/lib/Epub/Epub/blocks/TextBlock.cpp
+++ b/lib/Epub/Epub/blocks/TextBlock.cpp
@@ -11,16 +11,13 @@ void TextBlock::render(const GfxRenderer& renderer, const int fontId, const int 
     return;
   }
 
-  auto wordIt = words.begin();
-  auto wordStylesIt = wordStyles.begin();
-  auto wordXposIt = wordXpos.begin();
   for (size_t i = 0; i < words.size(); i++) {
-    const int wordX = *wordXposIt + x;
-    const EpdFontFamily::Style currentStyle = *wordStylesIt;
-    renderer.drawText(fontId, wordX, y, wordIt->c_str(), true, currentStyle);
+    const int wordX = wordXpos[i] + x;
+    const EpdFontFamily::Style currentStyle = wordStyles[i];
+    renderer.drawText(fontId, wordX, y, words[i].c_str(), true, currentStyle);
 
     if ((currentStyle & EpdFontFamily::UNDERLINE) != 0) {
-      const std::string& w = *wordIt;
+      const std::string& w = words[i];
       const int fullWordWidth = renderer.getTextWidth(fontId, w.c_str(), currentStyle);
       // y is the top of the text line; add ascender to reach baseline, then offset 2px below
       const int underlineY = y + renderer.getFontAscenderSize(fontId) + 2;
@@ -40,10 +37,6 @@ void TextBlock::render(const GfxRenderer& renderer, const int fontId, const int 
 
       renderer.drawLine(startX, underlineY, startX + underlineWidth, underlineY, true);
     }
-
-    std::advance(wordIt, 1);
-    std::advance(wordStylesIt, 1);
-    std::advance(wordXposIt, 1);
   }
 }
 
@@ -79,15 +72,15 @@ bool TextBlock::serialize(FsFile& file) const {
 
 std::unique_ptr<TextBlock> TextBlock::deserialize(FsFile& file) {
   uint16_t wc;
-  std::list<std::string> words;
-  std::list<uint16_t> wordXpos;
-  std::list<EpdFontFamily::Style> wordStyles;
+  std::vector<std::string> words;
+  std::vector<uint16_t> wordXpos;
+  std::vector<EpdFontFamily::Style> wordStyles;
   BlockStyle blockStyle;
 
   // Word count
   serialization::readPod(file, wc);
 
-  // Sanity check: prevent allocation of unreasonably large lists (max 10000 words per block)
+  // Sanity check: prevent allocation of unreasonably large vectors (max 10000 words per block)
   if (wc > 10000) {
     Serial.printf("[%lu] [TXB] Deserialization failed: word count %u exceeds maximum\n", millis(), wc);
     return nullptr;
--- a/test/stubs/Bitmap.h
+++ b/test/stubs/Bitmap.h
@@ -1,0 +1,43 @@
+#pragma once
+
+#include <SdFat.h>
+
+#include <cstdint>
+
+#include "BitmapHelpers.h"
+
+enum class BmpReaderError : uint8_t {
+  Ok = 0,
+  FileInvalid,
+  SeekStartFailed,
+  NotBMP,
+  DIBTooSmall,
+  BadPlanes,
+  UnsupportedBpp,
+  UnsupportedCompression,
+  BadDimensions,
+  ImageTooLarge,
+  PaletteTooLarge,
+  SeekPixelDataFailed,
+  BufferTooSmall,
+  OomRowBuffer,
+  ShortReadRow,
+};
+
+class Bitmap {
+ public:
+  static const char* errorToString(BmpReaderError) { return "stub"; }
+
+  explicit Bitmap(FsFile&, bool = false) {}
+  ~Bitmap() = default;
+  BmpReaderError parseHeaders() { return BmpReaderError::Ok; }
+  BmpReaderError readNextRow(uint8_t*, uint8_t*) const { return BmpReaderError::Ok; }
+  BmpReaderError rewindToData() const { return BmpReaderError::Ok; }
+  int getWidth() const { return 0; }
+  int getHeight() const { return 0; }
+  bool isTopDown() const { return false; }
+  bool hasGreyscale() const { return false; }
+  int getRowBytes() const { return 0; }
+  bool is1Bit() const { return true; }
+  uint16_t getBpp() const { return 1; }
+};
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -90,6 +90,16 @@ jobs:
           path: .pio/build/default/firmware.bin
           if-no-files-found: error
 
+  unit-tests:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+        with:
+          submodules: recursive
+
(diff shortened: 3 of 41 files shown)


osteotek:
Would it be possible to benchmark optimization on device?

BlindBat:
> Would it be possible to benchmark optimization on device?

Good idea! I'll flash the firmware and run before/after measurements on the X4 device. The desktop benchmarks give relative speedups, but absolute timings on the ESP32-C3 will be more meaningful — especially for the allocation-heavy paths where `malloc`/`free` overhead is much higher than on desktop.

I'll measure:
- **Text layout** (`layoutAndExtractLines`) — the list→vector and unique_ptr changes should show the biggest improvement here due to cache locality and elimination of atomic ref-counting
- **Page load** — the file handle reuse should eliminate ~2-3 SD card open/close cycles per page flip
- **Section build** — end-to-end section creation time for a representative chapter

Will update with device numbers.

osteotek:
Additional note: It’s tempting to optimize everything, especially since there are many areas where optimization is overdue. However, let’s ensure that it’s not bundled into a single large PR, as it significantly complicates the review process. Smaller, single-focused PRs will be reviewed and merged faster.

BlindBat:
## On-Device Benchmark Results (ESP32-C3, X4 hardware)

Tested by flashing master (baseline) and perf branch with identical timing instrumentation, then reading the same two EPUB books on the device and capturing serial output.

### Page Load (`loadPage`) — biggest win

| Metric | Master | Perf | Improvement |
|--------|--------|------|-------------|
| **First load** (file open + seek + deserialize) | 19.1ms avg | 15.8ms avg | **17%** |
| **Subsequent loads** (seek + deserialize, file already open) | 17.1ms avg | 5.8ms avg | **66%** |

Master opens/seeks/reads/closes the section file on every single page flip. The perf branch caches the file handle and LUT offset, paying the open cost once. Every subsequent page turn saves ~11ms of SD card latency.

### Section Build (`parseAndBuildPages`) — 15% faster

| Section | Pages | Master | Perf | Improvement |
|---------|-------|--------|------|-------------|
| part0001 (13 pages) | 13 | 738ms | 553ms | **25%** |
| part0009 (7 pages) | 7 | 207ms | 169ms | **18%** |
| part0003 (3 pages) | 3 | 75ms | 65ms | **13%** |
| sect-7p (7 pages) | 7 | 220ms | 191ms | **13%** |
| sect-4p (4 pages) | 4 | 118ms | 94ms | **20%** |
| Average across all matched sections | | | | **~15%** |

### Text Layout (`layoutAndExtractLines`) — 22% faster per-word

For paragraphs with 30+ words:

| Metric | Master | Perf | Improvement |
|--------|--------|------|-------------|
| Average ms/word | 0.129 | 0.101 | **22%** |

The gains come from `std::vector` iteration (cache-friendly) replacing `std::list` traversal (pointer-chasing), and `unique_ptr` eliminating atomic ref-count overhead from `shared_ptr`.

### Summary

| Optimization | What changed | Measured gain |
|-------------|-------------|---------------|
| File handle reuse | Open section file once, keep handle + cached LUT offset | **66% faster page loads** (subsequent) |
| `list` → `vector` | Words, positions, styles stored contiguously | **22% faster layout**, **15% faster builds** |
| `shared_ptr` → `unique_ptr` | TextBlock/PageLine/PageElement | Eliminates ~16 bytes + atomic ops per block |

The page load improvement is the most user-visible: it shaves ~11ms off every page turn, which on an e-ink device stacks directly onto the already-tight refresh budget.

BlindBat:
> Additional note: It’s tempting to optimize everything, especially since there are many areas where optimization is overdue. However, let’s ensure that it’s not bundled into a single large PR, as it significantly complicates the review process. Smaller, single-focused PRs will be reviewed and merged faster.

Understood — this PR is now scoped to just the Epub library performance changes (list→vector, shared_ptr→unique_ptr, file handle reuse). The test infrastructure lives in #764, and FB2 feature work is separate in #755.

All CI checks are passing and on-device benchmarks have been posted above. Ready for review!

jdk2pq:
@BlindBat It looks like we had some very similar ideas! 😁  I had been working on this today https://github.com/crosspoint-reader/crosspoint-reader/pull/779 I'll see if we can find a way to merge the two and get the best performance for the CSS parsing.

lib/Epub/Epub/blocks/TextBlock.cpp
```diff
@@ -11,16 +11,13 @@ void TextBlock::render(const GfxRenderer& renderer, const int fontId, const int
     return;
   }
 
-  auto wordIt = words.begin();
-  auto wordStylesIt = wordStyles.begin();
-  auto wordXposIt = wordXpos.begin();
   for (size_t i = 0; i < words.size(); i++) {
-    const int wordX = *wordXposIt + x;
-    const EpdFontFamily::Style currentStyle = *wordStylesIt;
```
ngxson, line 19: hmm, just wondering how this optimization could make sense.

normally `wordXpos[i]` should be compiled into `*wordXpos + i`, so an arithmetic op is needed each time we access the pointer (unless the compiler somehow optimizes it)

test/stubs/Bitmap.h
ngxson, line 1: IMO most of the stubs added in this PR is one-time usage. I bet the AI just try to get the test up and running.

Instead, we should implement these stubs in the HAL infrastructure instead.

Things like `Bitmap` doesn't make much sense to be a stub, because we do want to benchmark it in the future.

ngxson (commented):
Just my own opinion, but I don't really trust large optimizations made my AI. I have seen instances where even Claude Code intentionally modifies the testing results to make it look more appealing.

(fully disclosure: I'm one of the core maintainers of [llama.cpp](https://github.com/ggml-org/llama.cpp), and the instances I mentioned above is not a rare thing to us)

What I would suggest is that:
1. Make smaller optimizations, one PR for each, as @osteotek suggested above
2. Write a testing script that can run on device (maybe guarded under a `#define`), so that maintainers can test it. It's important to note that ESP32 CPU is much different than a computer CPU, as it lacks of cache and speculative/out-of-order execution. Testing on-device is still the most reliable metric.
3. Verify the compiled assembly code using something like `gdb` or `ghidra`

Also, note that some optimizations come with the cost of memory. So balancing between speed <> memory is also a consideration.

BlindBat:
@ngxson Thank you for your feedback! I thought it would be safe if it were covered with perf and unit tests. But I trust your experience. Actually, I use the version with these changes, and it works very fast. But I agree, it would be better to make it with small portions. Will do the decomposition later. I'll change the status of the PR to draft for the moment

osteotek:
@BlindBat I am sorry if I came across as a bit harsh 😃  I actually think these optimizations are valuable. I had same list->vector refactor sitting in my branch for a while. Tests also look good. Can we extract Page Load, Section Load and text layout optimizations separately, and I believe we can review and merge them in time

PR #768, opened 2026-02-08
feat: Tweak Lyra popup UI

bramschulting:
## Summary

I want to preface this PR by stating that the proposed changes are subjective to people's opinions. The following is just my suggestion, but I'm of course open to changes.

The popups in the currently implemented version of the Lyra theme feel a bit out of place. This PR suggests an updated version which looks a bit more polished and in line with the rest of the theme. 

I've also taken the liberty to remove the ellipsis behind the text of the popups, as they made the popup feel a bit off balance (example below).

With the applied changes, popups will look like this.

IMG_0012

The vertical position is (more or less) aligned to be in line with the sleep button. I'm aware the popup is used for other purposes aside from the sleep message, but this still felt like a good place. It's also a place where your eyes naturally 'rest'.

The popup has a small 2px white outline, neatly separating it from whatever is behind it.

### Alternatives considered and rationale behind proposal

Initially I started out worked off the Figma design for the Lyra theme, which [moves the popups](https://www.figma.com/design/UhxoV4DgUnfrDQgMPPTXog/Lyra-Theme?node-id=2011-19296&t=Ppj6B2MrFRfUo9YX-1) to the bottom of the screen. To me, this results in popups that are much too easy to miss:

IMG_0006

After this, I tried moving the popup back up (to the position of the sleep button), but to me it still kinda disappeared into the text of the book:

IMG_0008

Inverting the colors of the popup made things stand out the perfect amount in my opinion. The white outline separates the popup from what is behind it.

IMG_0011

This looked much better to me. The only thing that felt a bit off to me, was the balance due to the ellipsis at the end of the popup text. Also, "Entering Sleep..." felt a bit.. engineer-y. I felt something a bit more 'conversational' makes at all feel a bit more human-centric. But I'm no copywriter, and English is not even my native language. So feel free to chip in!

After tweaking that, I ended up with the final result:

_(Same picture as the first one shown in this PR)_

IMG_0012

## Additional Context

* Figma design: https://www.figma.com/design/UhxoV4DgUnfrDQgMPPTXog/Lyra-Theme?node-id=2011-19296&t=Ppj6B2MrFRfUo9YX-1

---

### AI Usage


bramschulting:
I realised I missed creating a `fillPopupProgress` implementation to work with the suggested popup. Once we've landed on what style we want to use, I will update my PR to make sure this works as intended.

bramschulting:
I've pushed a commit that implements `fillPopupProgress`. It works almost the same as in the BaseTheme, where a bar is rendered in the margin of the popup.

IMG_0019

Of course if the popup will change, this might need to change too. But having this PR merge ready before the end of the weekend worked best for me.

yagofarias:
Really lovely improvement! If I may, here some visual improvements i would do:
- I would bring the popup closer to the corners. Right now it doesn't seem to be aligned with anything. Decreasing the padding would make it easier to see.
- Loved the black version! For the Outline one i would increase the border size to make it more obvious when above book file similar to your case. Also you might consider adding outline white on black version to make it really "pop" from the content bellow.
- On popup with progress i would increase the padding to make it consistent with top alignment 

Despite from that it's a much better improvement from current version! Congrats

bramschulting:
@yagofarias 

> * I would bring the popup closer to the corners. Right now it doesn't seem to be aligned with anything. Decreasing the padding would make it easier to see.

Do you mean moving the entire popup to one of the corners? 

> * Loved the black version! For the Outline one i would increase the border size to make it more obvious when above book file similar to your case. Also you might consider adding outline white on black version to make it really "pop" from the content bellow.

Yep, I like the black one best too! It already has a white outline actually.

> * On popup with progress i would increase the padding to make it consistent with top alignment

Yeah I think I'll leave this as a followup PR, due to how this works in code now. Basically first a 'regular' popup is rendered, and then separately the progress bar is drawn inside it. They are two separate things drawn on top of each other. So we can't say: use a bigger margin in case there is a progress bar. Not without changing the API of these method at least.

yagofarias:
> Do you mean moving the entire popup to one of the corners?

Yes, positioning based on the Reading Orientation set by the user

bramschulting:
> Yes, positioning based on the Reading Orientation set by the user

Ah I see. Yeah maybe in landscape mode this could be an option. To position it in the top-right maybe, a but like notifications on MacOS? In portrait I think it might look strange, because of how narrow the screen is.

What do you think @daveallie? (Also of this PR in general)

daveallie:
The popup is currently positioned middle center regardless of reading orientation, and it's usually in response to a user operation, so there's no need to move it out of the way (top corner for example), I think the placement is fine. Regarding the style, I don't really mind either way tbh. @CaptainFrito has been running a lot of the theming so I'd like to lean on his opinion here.

CaptainFrito:
Looks great to me, thanks @bramschulting!

daveallie:
Apologies, with the merge of #728, updating strings has changed a bit, if you could update the translation files as part of resolving the conflicts that would be amazing.

bramschulting:
@daveallie No worries! The conflicts have been resolved, and I've reapplied the changes in the i18n system.

While I did change the English "Entering Sleep..." to "Going to sleep", I did not update the actual translations of other languages. I did remove the ellipsis, but the 'main' translations are unchanged. I don't speak any of these languages, and using Google Translate or something similar seems like the wrong choice in my opinion. But if preferred, I can still do that of course.

Issue #581, opened 2026-01-27
KOReader Sync Not Properly


getthething:
### Affected Version

0.16.0

### Describe the Bug

When I open the same book (downloaded via OPDS on my Kobo running KOReader, X4, and Readest on iOS) any progress pushed from the X4 is not recognized by Readest or KOReader. However, if I push/pull between KOReader and Readest, it works as expected. Progress pushed from KOReader/Readest isn't recognized on the X4 consistently. If it is, it will only sync to the correct chapter, not the exact spot (as it does with Readest and KOReader) 

If there are any logs or anything I can provide, I'm happy to. I just don't know how. 

### Steps to Reproduce

1. Open book on X4, read.
2. Push progress
3. Attempt to pull progress on KOReader

Or

1. Open book on KOReader, read.
2. Push progress
3. Attempt to pull progress on X4

### Expected Behavior

Pushed and pulled data to sync between all devices and sync to exact spot. 

### Relevant Log Output/Screenshots

```shell

```

bartiprime:
Yeah, I think I read on the commit that it is syncing via chapters only. I posted a message earlier and haven't received a response.

iandchasse:
The way KOSync works is that the exact index of document/chapter/paragraph/text is preserved and sent like a string of text. This string of text is related to how KOReader caches book locations. Since crosspoint caches its books in a different (and more memory conscious) way, these locations are not preserved. 

The chapters (sections), however, are the same for all KOReader books. So crosspoint takes the string for the chapter and sends it to KOSync, but because it only contains data for recent chapter, and not the paragraph/text offset, its point within the chapter is not preserved.  

This can be fixed in 2 ways:
- crosspoint caches it’s books in a way that can keep better track of its paragraph/div/text/etc elements, then maybe it can be mapped to somehow
- KOReader adds support to estimate a location in a chapter based on the progress variable that it receives instead of just using the exact text location (crosspoint preserves that variable, but KOReader doesn’t use it, in preference for the location string)

Both projects are open source, soooo anything’s possible

getthething:
> The way KOSync works is that the exact index of document/chapter/paragraph/text is preserved and sent like a string of text. This string of text is related to how KOReader caches book locations. Since crosspoint caches its books in a different (and more memory conscious) way, these locations are not preserved.
> 
[...]

PR #783, opened 2026-02-09
fix: Fix kosync repositioning issue

jpirnay:
## Summary

* Original implementation had inconsistent positioning logic:
  - When XPath parsing succeeded: incorrectly set pageNumber = 0 (always beginning of chapter)
  - When XPath parsing failed: used percentage for positioning (worked correctly)
  - Result: Positions restored to wrong locations depending on XPath parsing success
  - Mentioned in Issue #581 
* Solution
  - Unified ProgressMapper::toCrossPoint() to use percentage-based positioning exclusively for both spine identification and intra-chapter page calculation, eliminating unreliable XPath parsing entirely.

## Additional Context

* ProgressMapper.cpp: Simplified toCrossPoint() to always use percentage for positioning, removed parseDocFragmentIndex() function
* ProgressMapper.h: Updated comments and removed unused function declaration
* Tests confirmed appropriate positioning
* __Notabene: the syncing to another device will (most probably) end up at the current chapter of crosspoints reading position. There is not much we can do about it, as KOReader needs to have the correct XPath information - we can only provide an apporximate position (plus percentage) - the percentage information is not used in KOReaders current implementation__
---

### AI Usage

### DIFF
--- a/src/activities/reader/KOReaderSyncActivity.cpp
+++ b/src/activities/reader/KOReaderSyncActivity.cpp
@@ -129,18 +129,18 @@ void KOReaderSyncActivity::performSync() {
   CrossPointPosition localPos = {currentSpineIndex, currentPage, totalPagesInSpine};
   localProgress = ProgressMapper::toKOReader(epub, localPos);
 
-  xSemaphoreTake(renderingMutex, portMAX_DELAY);
-  state = SHOWING_RESULT;
+  {
+    RenderLock lock(*this);
+    state = SHOWING_RESULT;
 
-  // Default to the option that corresponds to the furthest progress
-  if (localProgress.percentage > remoteProgress.percentage) {
-    selectedOption = 1;  // Upload local progress
-  } else {
-    selectedOption = 0;  // Apply remote progress
+    // Default to the option that corresponds to the furthest progress
+    if (localProgress.percentage > remoteProgress.percentage) {
+      selectedOption = 1;  // Upload local progress
+    } else {
+      selectedOption = 0;  // Apply remote progress
+    }
   }
-
-  xSemaphoreGive(renderingMutex);
-  updateRequired = true;
+  requestUpdate();
 }
 
 void KOReaderSyncActivity::performUpload() {


osteotek:
Seems like it was built on top of overlay PR branch, better to separate them

jpirnay:
Rebased

iandchasse:
How well does this work? I was doing some light testing in #61 on the KOReader side of things by replacing its own sync function with a percentage based look-up. If you do this, then the syncing can technically be two-way preserved.

But, I think KOReader bases its % on ‘page index / total pages’, not book size as crosspoint does. So text based books should probably be pretty close in terms of syncing, but books with a lot of images might be a bit off. 

I don’t own an X4, so I could be very wrong.

jpirnay:
Caveat; I dont use a loot of books containing images, so YMMV
The relative position within a file will probably be different if the reader would include graphics as the rendered images will consume space.
For my use cases, it was close enough

osteotek:
> But, I think KOReader bases its % on ‘page index / total pages’, not book size as crosspoint does. So text based books should probably be pretty close in terms of syncing, but books with a lot of images might be a bit off.


GetCumulativeSize function that percentage calculation relies on uses uncompressed sizes of each spine item file (the XHTML/HTML/other spine href targets) and accumulates those sizes; it does not include images referenced inside those files. So it should work the same for image-heavy books as for non-image books

iandchasse:
@osteotek Do you happen to know if there are methods already implemented to achieve a ‘current page / total pages’ calculation instead of using the size? Or does the file have to be fully indexed before that can be done? I don't think that should replace all references to % as already done, but in this specific domain of KOSync, it's probably valuable to try and match up so that it'll work regardless of the file being synced.

src/activities/reader/KOReaderSyncActivity.cpp
```diff
@@ -121,15 +121,22 @@ void KOReaderSyncActivity::performSync() {
...
   state = SHOWING_RESULT;
-  selectedOption = 0;  // Default to "Apply"
+
+  // Default to the option that corresponds to the furthest progress
+  if (localProgress.percentage > remoteProgress.percentage) {
+    selectedOption = 1;  // Upload local progress
+  } else {
+    selectedOption = 0;  // Apply remote progress
+  }
+
```
osteotek: nice touch

osteotek:
@jpirnay I’ve tested this, and it appears to work better than before. At least, it doesn’t always sync to the beginning of the chapter on the Crosspoint side. However, syncing from crosspoint to the remote server and then fetching progress on another Koreader always puts me at the beginning of the next chapter (not the same one I’ve been reading on Crosspoint). Is this the same for you?

jpirnay:
> @jpirnay I’ve tested this, and it appears to work better than before. At least, it doesn’t always sync to the beginning of the chapter on the Crosspoint side. However, syncing from crosspoint to the remote server and then fetching progress on another Koreader always puts me at the beginning of the next chapter (not the same one I’ve been reading on Crosspoint). Is this the same for you?

You are pointing our a very important issue:
The XPath being handed over is a synthetic, chapter-level XPath pointing to the DocFragment corresponding to the current spine item (chapter). It is not the XPath of the closest text fragment, as CrossPoint does not preserve detailed HTML structure during parsing—only the linear spine and page counts.

We are assuming KOReader uses the XPath to identify the DocFragment and the percentage for fine positioning within it (thats how Crosspoint does it when reading something). However, based on KOReader's syncToProgress implementation, it only uses GotoXPointer(xPath) for EPUB, ignoring percentage. This results in positioning at the start of the specified DocFragment (chapter start), not an exact location.

Reality: Sync achieves approximate, chapter-level positioning, not pixel-perfect sync. KOReader lands at the beginning of the DocFragment because the XPath lacks detailed element paths (e.g., no `<p>` or `<div>` tags). Real KOReader XPaths are detailed (e.g., /body/DocFragment/body/div/div[4]/ul/li[3]...), but CrossPoint can't generate them.

If KOReader were modified to use percentage for intra-DocFragment positioning, sync would be more accurate. As-is, this is the best CrossPoint can do. It works for round-trip CrossPoint sync, and provides chapter-level KOReader positioning.

osteotek:
@jpirnay thanks for the analysis, seems like we can't solve this without modifications to how crosspoint stores text, or without reparsing html doc on every sync. PR in its current form looks like a good addition, I've tested it, works good.

osteotek (dismissed):

jpirnay:
For the brave of heart, @iandchasse pointed out a solution for better KOReader sync behaviour, which requires the patching of a single file though: https://github.com/crosspoint-reader/crosspoint-reader/discussions/61#discussioncomment-15770877

osteotek:
@jpirnay would it be possible to fix merge conflicts here?

jpirnay:
> @jpirnay would it be possible to fix merge conflicts here?

Done

src/activities/reader/KOReaderSyncActivity.cpp
```diff
@@ -123,18 +123,24 @@ void KOReaderSyncActivity::performSync() {
...
+  // Default to the option that corresponds to the furthest progress
+  if (localProgress.percentage > remoteProgress.percentage) {
+    selectedOption = 1;  // Upload local progress
+  } else {
+    selectedOption = 0;  // Apply remote progress
   }
-  requestUpdate();
+
+  xSemaphoreGive(renderingMutex);
+  updateRequired = true;
```
osteotek: SubActivities don't need to control mutex anymore, also need to switch to   requestUpdate(); 
See https://github.com/crosspoint-reader/crosspoint-reader/pull/774
That's why build is failing

Issue #423, opened 2026-01-19
Home page takes forever to load


ben25w:
### Affected Version

.0.1.4

### Describe the Bug

Going back to the home page takes an extensive amount of time, sometimes 30 seconds. I think it's because it's trying to(and failing) to launch my book cover as an image.

### Steps to Reproduce

Just going to the home page

### Expected Behavior

It shouldn't take 30 seconds or longer to load

### Relevant Log Output/Screenshots

```shell

```

drbourbon:
Confirm that. From serial port debug this seems to happen when the current reading book failed to generate book cover. Perhaps a fallback cover should be set for a book after a cover generation failure?

jonasdiemer:
IMO, even the "good" case is broken (i.e. when the cover does generate, it takes in my examples ~10s, so it appears the device is stuck).

I think we need to render the cover in the background and only display it when ready...

jonasdiemer:
Since background rendering could be difficult, I would propose (as short-term fix) we add an entry to the book menu "generate cover" (see PR #417) and show the cover on Home Screen only if it had been generated manually before.

jonasdiemer:
This is continuing to annoy me - each time I clear the cache (or read a new book), going back home takes ~10s without any user notification (device seems stuck).

@crosspoint-reader/firmware-maintainers - please comment if you're OK with above suggestion to "force users" to manually generate the covers. Or maybe we could/should generate the home cover when going to sleep (and only show the cover on home if it's been generated before, i.e. after a sleep cycle).

lukestein:
> comment if you're OK with above suggestion to "force users" to manually generate the covers. Or maybe we could/should generate the home cover when going to sleep (and only show the cover on home if it's been generated before, i.e. after a sleep cycle).

Personally, I'd def avoid making cover generations an explicit task. Fine with me to delay cover generation, but I think our main target user is someone who just wants to open a book and read it (which should include cover display if available), even if a one-time delay that occurs sometime.

jonasdiemer:
OK, so we then would defer the thumb generation to going to sleep? I.e. display the thumb on Home Screen only if it had been generated before and move the generation to going to sleep?

osteotek:
How about generating the cover when the book is first opened instead of on first time going to home?

jonasdiemer:
> How about generating the cover when the book is first opened instead of on first time going to home?

That would basically double the time it takes to load a book.

[...]

PR #788, opened 2026-02-09
perf: RFC / Proposal document for async thumb generation

jonasdiemer:
## Summary

Proposal to an initiate discussion/decision of async architecture for thumb generation etc.

---

### AI Usage

### DIFF
--- a/docs/proposal-async-thumb-generation.md
+++ b/docs/proposal-async-thumb-generation.md
@@ -1,0 +1,60 @@
+# Proposal: Background Thumbnail & Async SD Tasks
+
+## Current State
+
+- **Thumb generation takes ~10s** per cover (JPEG decode + Atkinson dither → 1bpp BMP), blocking the home screen.
+- We have similar issues with loading chapters (when entering a new chapter), generating sleep screen covers (blocks going to sleep, not critical).
+- **One SPI bus** shared between SD (CS=12) and e-ink display (CS=21). This concurrency needs to be considered, as all background tasks need access to SD.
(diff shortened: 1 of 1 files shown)


osteotek:
Papyrix fork uses background tasks for page caching, perhaps this model can be adopted- https://github.com/bigbag/papyrix-reader/blob/main/src/states/ReaderState.cpp#L844

jonasdiemer:
> Papyrix fork uses background tasks for page caching, perhaps this model can be adopted- https://github.com/bigbag/papyrix-reader/blob/main/src/states/ReaderState.cpp#L844

Thanks for the pointer. I had a quick look - at first glance, it seems to be a variant Option B - cooperative state machine.

The background task creates the cache in chunks of 10 pages. Then, the thumb is generated in 1 go from the cover BMP (not from JPEG). There are checks in between the chunks if it should stop. Stops are requested when the user turns pages - I assume to handle concurrency on SPI.

What's also interesting is that they are accepting the double-dithering issue of generating the 1bpp thumb from the 2bpp dithered cover image. They're using a dedicated functions for scaling and dithering to minimize the negative impact. Will comment in #423 - maybe it's an option for us as a workaround.

pablohc:
Perhaps this can partially solve the problem for those who want to directly avoid generation times or do it deferred on demand. https://github.com/crosspoint-reader/crosspoint-reader/pull/1842

itsthisjustin:
I think this belongs as a discussion and not a PR so in an effort to clean up our PRs I'm going to close this for now but please feel free to open as a discussion!

PR #789, opened 2026-02-09
feat: Add i18n internationalization system

brbla:
## Summary

- Add dynamic internationalization (i18n) system that scans the SD card for `.lang` files — adding a new language requires no firmware changes
- `TranslationManager` singleton with `T()` macro wraps all user-visible strings across ~30 source files
- Language setting exposed as `DynamicEnum` in Settings → System → Language
- Includes translations: Czech, French, German, Italian + English reference template
- **Web UI translated** via `/api/i18n` endpoint — all 3 web pages (Home, File Manager, Settings) fully support i18n

### How it works

- English strings serve as translation keys — `T("Settings")` returns `"Settings"` for English, `"Nastavení"` for Czech
- `.lang` files use a simple `key=value` format with `language.name=` metadata for display names
- Language preference stored as string (`languageCode`) in settings, persists across reboots
- Available languages are detected automatically from `/config/lang/*.lang` files on SD card

### Architecture

- **Build-time**: `tools/lang_compile.py` reads `en.lang`, extracts all translation keys, computes FNV-1a hashes, and generates a flash-resident hash table (`LangKeys.inc`) with open addressing and linear probing. The script runs automatically as a PlatformIO pre-build step with MD5-based staleness detection.
- **Runtime lookup**: `T("key")` computes FNV-1a hash → O(1) lookup in the flash hash table to get the key ID → direct index into the values array in RAM. No string comparisons at runtime.
- **Memory layout**: Only translated values are stored in RAM (~2-3 KB for 260+ keys). The hash table (6 bytes × 1024 slots) lives entirely in flash. English fallback returns the key itself with zero RAM cost.
- **Collision safety**: The build script verifies zero hash collisions at compile time — a collision would fail the build.
- Settings serialized as field 31 (`writeString`/`readString`) for backward compatibility.

### Web UI i18n

The 3 web pages (Home, File Manager, Settings) served via the built-in WiFi hotspot are also translated:

- **`/api/i18n` endpoint**: Returns a JSON object with translations for ~120 keys (navigation, page content, settings categories/names/values). For English, returns `{}` — zero overhead.
- **HTML pages**: Static text uses `data-i18n` attributes, dynamic JS-generated text uses a `t()` helper function. English text stays in HTML as fallback.
- **Settings page**: Category names, setting names, and enum option values are all translated via `t()` at render time.
- **No extra HTTP request latency**: The i18n fetch runs in parallel with page initialization.

### Performance characteristics

| Metric | Value |
|--------|-------|
| Lookup complexity | O(1) amortized (hash + linear probe) |
| Hash table | 1024 slots in flash, 30% load factor |
| RAM usage (non-English) | ~2-3 KB (values pool only, shrink-to-fit) |
| RAM usage (English) | 0 bytes (keys are their own values) |
| Translation keys | 309 |
| Hash collisions | 0 (verified at build time) |

### Optimization history

The initial implementation used a sorted array with binary search — O(log n) × `strcmp()` per lookup and ~6 KB RAM for both keys and values. After [feedback](https://github.com/crosspoint-reader/crosspoint-reader/pull/789#issuecomment-2836775945) about CPU and RAM cost on ESP32-C3 (380 KB SRAM), this was redesigned to the current flash hash table approach, which eliminates runtime string comparisons entirely and halves RAM usage by storing only values.

### Adding a new language

1. Copy `en.lang`, rename to language code (e.g. `es.lang`)
2. Set `language.name=Español`
3. Translate values on the right side of `=`
4. Place on SD card at `/config/lang/es.lang`

## Test plan

- [x] Build succeeds with `pio run -e default`
- [x] English UI works without any `.lang` files on SD card
- [x] Adding `cs.lang` to SD card shows "Čeština" in Settings → System → Language
- [x] Selecting Czech translates all UI strings (boot, sleep, home, settings, WiFi, reader)
- [x] Switching back to English restores original strings
- [x] Removing a `.lang` file removes the language from the list
- [x] Language preference persists across reboots
- [x] Web UI: Home page translated (status labels, navigation)
- [x] Web UI: File Manager translated (table headers, modals, buttons)
- [x] Web UI: Settings page translated (categories, setting names, enum values)
- [x] Web UI: English mode — no translation script injected, zero overhead
- [ ] Web API (`GET /api/settings`) returns language setting correctly

### DIFF
--- a/src/activities/settings/SettingsActivity.cpp
+++ b/src/activities/settings/SettingsActivity.cpp
@@ -26,7 +26,7 @@ void SettingsActivity::taskTrampoline(void* param) {
 
 void SettingsActivity::onEnter() {
   Activity::onEnter();
-  Serial.printf("[SETTINGS] onEnter, free heap: %u\n", static_cast<unsigned>(ESP.getFreeHeap()));
+  LOG_DBG("SET", "onEnter, free heap: %u", static_cast<unsigned>(ESP.getFreeHeap()));
   renderingMutex = xSemaphoreCreateMutex();
 
   // Build per-category vectors from the shared settings list
@@ -35,7 +35,7 @@ void SettingsActivity::onEnter() {
   controlsSettings.clear();
   systemSettings.clear();
 
-  Serial.println("[SETTINGS] Building settings list...");
+  LOG_DBG("SET", "Building settings list...");
   for (auto& setting : getSettingsList()) {
     if (!setting.category) continue;
     if (strcmp(setting.category, "Display") == 0) {
@@ -67,7 +67,7 @@ void SettingsActivity::onEnter() {
   currentSettings = &displaySettings;
   settingsCount = static_cast<int>(displaySettings.size());
 
-  Serial.printf("[SETTINGS] Settings built, free heap: %u\n", static_cast<unsigned>(ESP.getFreeHeap()));
+  LOG_DBG("SET", "Settings built, free heap: %u", static_cast<unsigned>(ESP.getFreeHeap()));
 
   // Trigger first update
   updateRequired = true;
--- a/src/i18n/TranslationManager.cpp
+++ b/src/i18n/TranslationManager.cpp
@@ -2,7 +2,7 @@
 
 #include <Esp.h>
 #include <HalStorage.h>
-#include <HardwareSerial.h>
+#include <Logging.h>
 
 #include <algorithm>
 #include <cstdlib>
@@ -31,7 +31,7 @@ TranslationManager::~TranslationManager() { freeAll(); }
 // ──────────────────────────────────────────────
 
 bool TranslationManager::init(const char* lang) {
-  Serial.println("[i18n] Initializing");
+  LOG_DBG("I18N", "Initializing");
 
   freeAll();
 
@@ -41,10 +41,10 @@ bool TranslationManager::init(const char* lang) {
     if (isValidLangCode(lang)) {
       sdLoaded = loadFromSD(lang);
       if (!sdLoaded) {
-        Serial.printf("[i18n] WARNING: Failed to load '%s', using English\n", lang);
+        LOG_ERR("I18N", "Failed to load '%s', using English", lang);
       }
     } else {
-      Serial.printf("[i18n] WARNING: Invalid language code '%s'\n", lang);
+      LOG_ERR("I18N", "Invalid language code '%s'", lang);
       sdLoaded = false;
     }
   }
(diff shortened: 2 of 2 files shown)


Uri-Tauber:
Hi @brbla,
Great work — your approach is definitely much easier to extend than mine (#728).
That said, I still believe my solution is better suited to the very limited resources of the ESP32-C3 (yes, I’m biased 😅).

Your method requires O(log n) × strcmp() for every UI string lookup, which can become noticeably CPU-heavy on the ESP32-C3. Also, keeping the active language’s strings in RAM means a persistent ~5 KB allocation — which is quite significant when you only have 380 KB total.
My approach obviously forces a firmware recompile to add or change languages, but all strings live in flash and — thanks to the enum + direct array indexing — string retrieval is pure O(1) with almost zero runtime cost.

I’ve just finished my PR and I’m now working on merging master into it.
I would really appreciate any feedback or thoughts you might have.
Thanks again for the your work!

brbla:
Thanks. I'm learning with this and every feedback helps me. I'll try to optimize it further and see where I can get. Thanks for the comments, at least I know what to study better! 😉

brbla:
Hi, @Uri-Tauber, thanks again for the feedback. Out of curiosity, I tried to adjust some things according to your comments. Reduce RAM requirements, make finding translations more efficient, and extend translation to the web interface (which I had completely forgotten about initially).

So thanks for the opportunity to get to know it a little more deeply than just as a device user.

Uri-Tauber:
That’s great to hear! If you learned something along the way, then it was definitely worth it 🙂
Going forward, I’d suggest checking whether someone is already working on a feature before investing too much time in it. When two people work on the same thing in parallel, a lot of effort can end up being wasted.
There are plenty of other open areas where your contribution would have a big impact.

brbla:
Hi, @Uri-Tauber. While testing and working on it, I came across a few flaws that showed up. This led to separate pull requests and if I'm not mistaken, it should lead to more stability. And at the same time I was able to use the translated strings for your branch = I hope I don't lose Czech in the reader UI! :-D

Today I tried the official firmware V5.0.3 and the fonts in the book still don't work well.

daveallie:
Hey @brbla, at this point, my preference is to progress @Uri-Tauber's PR for a few reasons, but namely, in the interest of keeping translations up to date. I am interesting to hear more opinions from @crosspoint-reader/firmware-maintainers, but my 2c is flash stored strings for all languages means that as new features are built we can ensure translations exist. Afaict your solution relies on users upating their .lang files as new strings are added to CrossPoint.

When you say "it should lead to more stability", could you be a little more specific? I'd love to continue looking at both if there's some specific additional information worth considering, but it looks to me that #728 is effectively compile-time safe?

ngxson:
I agree with the earlier comment from @Uri-Tauber about performance and memory issues with this approach.

brbla:
Hello @daveallie, as @ngxson explained to me in my PR #871 , my conclusion was wrong and my solution is qualitatively worse. The only benefit (maybe only for me) is that when editing the translation, it was enough to change the contents of the .lang file and quickly see the translation changes in the context of the reader environment without having to recompile and upload the firmware.

Thanks for your time, both of you, I appreciate it and don't want to take it away from other PRs who are in better shape.

daveallie:
I understand, it's always good to evaluate a few approaches with these large pieces, so I appreciate the effort here.
Going to move forward with #728 and close this one out to avoid further confusion.

Issue #739, opened 2026-02-07
Battery Status move while navigating the menu


lepislepis:
### Affected Version

1.0.0 RC

### Describe the Bug

When I try to navigating the menu on homepage, it appears the battery icon and status is moving and the value is not consistent (e.g. 100% and 99%)

<img width="574" height="876" alt="Image" src="https://github.com/user-attachments/assets/5d709b21-7f27-40ea-a8b9-f093749d64be" />

<img width="570" height="874" alt="Image" src="https://github.com/user-attachments/assets/2b0e4875-6b95-4f0e-9df5-50d3add35220" />

<img width="573" height="879" alt="Image" src="https://github.com/user-attachments/assets/dae12caf-80e6-4993-a9a3-a354d5e11cbc" />

### Steps to Reproduce

1. go to different menu

### Expected Behavior

battery status not moving

### Relevant Log Output/Screenshots

```shell

```

jonasdiemer:
Very similar to #729

Marbelymarble:
I looked into it a bit. We're currently just using the raw battery percentage from the SDK which is very much not consistent and likes to jump around. I'd love to take this issue on if nobody else is already on it :)

lukestein:
> I looked into it a bit. We're currently just using the raw battery percentage from the SDK which is very much not consistent and likes to jump around. I'd love to take this issue on if nobody else is already on it :)

In the works at #684

Issue #729, opened 2026-02-06
Battery symbol redrawing on Home Screen doesn't clear


jonasdiemer:
### Affected Version

0.16.0-dev (9f78fd3)

### Describe the Bug

Seems we're not clearing the screen behind the battery when redrawing it on the Home Screen (eg. when moving between options).

### Steps to Reproduce

1. Be on Home Screen (classic theme)
2. move up/down
3. Observer battery status "overdrawn" when redrawn.

### Expected Behavior

screen area is "cleared" before battery is drawn.

### Relevant Log Output/Screenshots

```shell

```

osteotek:
@jonasdiemer tangential, but what do you think about removing classic theme altogether? Supporting two UI options simultaneously is a headache

lukestein:
> @jonasdiemer tangential, but what do you think about removing classic theme altogether? Supporting two UI options simultaneously is a headache

I'm of both minds @osteotek. OTOH we're aiming for **focused reading experience** so one good theme should be enough. OTOH people like customization and if things are written with appropriate abstraction then Lyra serves an example of what else we can add later (or people can in a fork).

lukestein:
This may be as simple as redrawing on the Home Screen, but I think that this, #684 and #713 are all places where the battery icon could use some tlc

jonasdiemer:
> [@jonasdiemer](https://github.com/jonasdiemer) tangential, but what do you think about removing classic theme altogether? Supporting two UI options simultaneously is a headache

@osteotek Tbh, I switched to classic because I felt the new theme too cluttered (so not love at first sight ;)). Also, it is amplyfying the issue with generating thumbs for the Home Screen due to showing multiple books there (so my device would freeze for 10s each time I open a new book) - so we definitely need an option do change that.

jonasdiemer:
There is also an issue with the width of the text string - it jumps when the battery is between 99 and 100, which may look weird. Maybe we want to hard-code it to the width of "100 %".

Eloren1:
Classic theme should be here definitely

pablohc:
> There is also an issue with the width of the text string - it jumps when the battery is between 99 and 100, which may look weird. Maybe we want to hard-code it to the width of "100 %".

It could be improved by changing the layout of the icon and the charge percentage.
The battery icon on the right will always remain in the same place regardless of whether the charge has 3 or 2 digits.

[...]

PR #796, opened 2026-02-09
fix: align battery icon based on context (UI / Reader)

pablohc:
Issues solved: #729 and #739

## Summary

Currently, the battery icon and charge percentage were aligned to the left even for the UI, where they were positioned on the right side of the screen. This meant that when changing values of different numbers of digits, the battery would shift, creating a block of icons and text that was illegible.

- Add drawBatteryUi() method for right-aligned battery display in UI headers
- Keep drawBattery() for left-aligned display in reader mode
- Extract drawBatteryIcon() helper to reduce code duplication
- Battery icon now stays fixed at right edge regardless of percentage digits
- Text adjusts to left of icon in UI mode, to right of icon in reader mode

## Additional Context

* Add any other information that might be helpful for the reviewer 
* This fix applies to both themes (Base and Lyra).

---

### AI Usage

### DIFF
--- a/src/components/themes/BaseTheme.cpp
+++ b/src/components/themes/BaseTheme.cpp
@@ -61,11 +61,10 @@ void BaseTheme::drawBatteryRight(const GfxRenderer& renderer, Rect rect, const b
   // rect.x is already positioned for the icon (drawHeader calculated it)
   const uint16_t percentage = battery.readPercentage();
   const int y = rect.y + 6;
-  int textWidth = 0;
 
   if (showPercentage) {
     const auto percentageText = std::to_string(percentage) + "%";
-    textWidth = renderer.getTextWidth(SMALL_FONT_ID, percentageText.c_str());
+    const int textWidth = renderer.getTextWidth(SMALL_FONT_ID, percentageText.c_str());
     // Clear the area where we're going to draw the text to prevent ghosting
     const auto textHeight = renderer.getTextHeight(SMALL_FONT_ID);
     renderer.fillRect(rect.x - textWidth - batteryPercentSpacing, rect.y, textWidth, textHeight, false);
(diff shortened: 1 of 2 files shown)


jonasdiemer:
Seems to fix #739 - I would propose to move the whole battery+number on Home Screen etc. a bit further to the right to align it with the button edge (on base theme).

However, #729 still is an issue - to reproduce, go Home and press around when the battery is around 99-100%. It is redrawn each time the selection changes, but the old value is not cleared out, so 2 numbers are written on top of each other.

src/components/themes/BaseTheme.cpp
```diff
@@ -17,41 +17,61 @@ namespace {
...
+  // rect.x is already positioned for the icon (drawHeader calculated it)
+  const uint16_t percentage = battery.readPercentage();
+  const int y = rect.y + 6;
+  int textWidth = 0;
+
+  if (showPercentage) {
+    const auto percentageText = std::to_string(percentage) + "%";
+    textWidth = renderer.getTextWidth(SMALL_FONT_ID, percentageText.c_str());
+    // Draw text to the left of the icon
+    renderer.drawText(SMALL_FONT_ID, rect.x - textWidth - batteryPercentSpacing, rect.y, percentageText.c_str());
```
jonasdiemer, line 72: Clear the area where you're drawing the text, e.g. add
```diff
+    const auto textHeight = renderer.getTextHeight(SMALL_FONT_ID);
+    renderer.fillRect(rect.x - textWidth - batteryPercentSpacing, rect.y,
+                            textWidth, textHeight, false);
```

src/components/themes/BaseTheme.h
```diff
@@ -93,7 +93,10 @@ class BaseTheme {
 
   // Component drawing methods
   virtual void drawProgressBar(const GfxRenderer& renderer, Rect rect, size_t current, size_t total) const;
-  virtual void drawBattery(const GfxRenderer& renderer, Rect rect, bool showPercentage = true) const;
+  virtual void drawBattery(const GfxRenderer& renderer, Rect rect,
```
jonasdiemer: Maybe rename to left/right rather than "Ui"

pablohc:
Thanks for the feedback @jonasdiemer, it should be fixed now.
Please let me know if you see anything that could be improved.

jonasdiemer (approved):
Tested on device, looks good.

daveallie (approved):

PR #809, opened 2026-02-09
fix: Reduce MIN_SIZE_FOR_POPUP to 10KB

jonasdiemer:
Noticed that the Indexing... popup went missing despite 3-5 seconds delay. Reducing to 10KB, so we get a popup for delays > ~2s.

### AI Usage

Did you use AI tools to help write this code? no

### DIFF
--- a/lib/Epub/Epub/parsers/ChapterHtmlSlimParser.cpp
+++ b/lib/Epub/Epub/parsers/ChapterHtmlSlimParser.cpp
@@ -11,7 +11,7 @@ const char* HEADER_TAGS[] = {"h1", "h2", "h3", "h4", "h5", "h6"};
 constexpr int NUM_HEADER_TAGS = sizeof(HEADER_TAGS) / sizeof(HEADER_TAGS[0]);
 
 // Minimum file size (in bytes) to show indexing popup - smaller chapters don't benefit from it
-constexpr size_t MIN_SIZE_FOR_POPUP = 50 * 1024;  // 50KB
+constexpr size_t MIN_SIZE_FOR_POPUP = 10 * 1024;  // 10KB
 
 const char* BLOCK_TAGS[] = {"p", "li", "div", "br", "blockquote"};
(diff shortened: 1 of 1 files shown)


lukestein:
Requesting @osteotek review since I think maybe he was the one who cut out some of these popups since they took a while to draw?

daveallie:
Agree that I've had a handful of cases where the delay was large without feedback, keen to tighten this up as you've done, but also happy for @osteotek to weigh in

osteotek:
yes, with recent additions feels like chapter rendering time increased, so this timeout needs to be tightened up. But maybe set it to 20kb?

jonasdiemer:
> yes, with recent additions feels like chapter rendering time increased, so this timeout needs to be tightened up. But maybe set it to 20kb?

I tried 20kb first, still got "unpleasant" delays with ca 13kB file, so I decided to go down to 10. A 6kB file I tried still had noticable delay, but by the time you're wondering if the device is stuck, the delay was over. So I figured 10kB works well at the moment.

osteotek (approved):

osteotek:
> > yes, with recent additions feels like chapter rendering time increased, so this timeout needs to be tightened up. But maybe set it to 20kb?
> 
> I tried 20kb first, still got "unpleasant" delays with ca 13kB file, so I decided to go down to 10. A 6kB file I tried still had noticable delay, but by the time you're wondering if the device is stuck, the delay was over. So I figured 10kB works well at the moment.

I'm fine with 10kb too. It's just for small chapters that means time to text refresh increased by more than x2, just the fact of having to render popup becomes the biggest bottleneck. So we need to decide what is more unpleasant: increased time to render new chapter or smaller time but with no feedback

osteotek:
Just did some testing, opening same chapter size 18kb, time to first text:
- no popups -- 1226 ms
- 1 popup -- 1669 ms

so not much of a difference as I thought. Setting to 10 kb is probably good threshold

PR #810, opened 2026-02-09
feat: Extend python debugging monitor functionality (keyword filter / suppress)

jpirnay:
## Summary

* I needed the ability to filter and or suppress debug messages containig certain keywords (eg [GFX] for render related stuff)
* Update of debugging_monitor.py script for development work

## Additional Context
```
usage: debugging_monitor.py [-h] [--baud BAUD] [--filter FILTER] [--suppress SUPPRESS] [port]

ESP32 Monitor with Graph

positional arguments:
  port                 Serial port

options:
  -h, --help           show this help message and exit
  --baud BAUD          Baud rate
  --filter FILTER      Only display lines containing this keyword (case-insensitive)
  --suppress SUPPRESS  Suppress lines containing this keyword (case-insensitive)
```
* plus a couple of platform specific defaults (port, pip style)
---

### AI Usage

### DIFF
--- a/scripts/debugging_monitor.py
+++ b/scripts/debugging_monitor.py
@@ -1,3 +1,12 @@
+#!/usr/bin/env python3
+"""
+ESP32 Serial Monitor with Memory Graph
+
+This script provides a real-time serial monitor for ESP32 devices with
+integrated memory usage graphing capabilities. It reads serial output,
+parses memory information, and displays it in both console and graphical form.
+"""
+
 import sys
 import argparse
 import re
@@ -6,7 +15,7 @@ from datetime import datetime
 from collections import deque
 
 # Try to import potentially missing packages
-PACKAGE_MAPPING = {
+PACKAGE_MAPPING: dict[str, str] = {
     "serial": "pyserial",
     "colorama": "colorama",
     "matplotlib": "matplotlib",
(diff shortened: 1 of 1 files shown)


scripts/debugging_monitor.py
jonasdiemer, line 10: I think we should add something like this to the beginning of the file so it can be executed directly.
#!/usr/bin/env python3

scripts/debugging_monitor.py
```diff
@@ -160,55 +226,81 @@ def update_graph(frame):
...
     plt.tight_layout()
 
+
 def main():
     parser = argparse.ArgumentParser(description="ESP32 Monitor with Graph")
-    parser.add_argument("port", nargs="?", default="/dev/ttyACM0", help="Serial port")
+    if sys.platform.startswith("win"):
+        default_port = "COM8"
+    elif sys.platform.startswith("darwin"):
+        default_port = "/dev/tty.usbmodem14101"
```
jonasdiemer: On my machine (M1 MacBook Air), this one is the default:
/dev/cu.usbmodem101

Not sure this can be consistent...

jpirnay: For sure not - but being the egoistic pig I am I used my default for my windows machine and left the one for linux unchanged. Happy to extend that to a sensible darwin default

jonasdiemer: Looked into it, the number is dependent on USB topology, so we can't provide a default that works across different machines. Best default would be to see if there is only one tty.usbmodem* and use that (and stop if there are multiple). But that could be a separate PR IMO.

jonasdiemer: See #816

PR #779, opened 2026-02-08
perf: Improve large CSS files handling

jdk2pq:
## Summary

Closes #766. Thank you for the help @bramschulting!

**What is the goal of this PR?** 
- First and foremost, fix issue #766.
- Through working on that, I realized the current CSS parsing/loading code can be improved dramatically for large files and still had additional performance improvements to be made, even with EPUBs with small CSS.

**What changes are included?**
- Stream CSS parsing and reuse normalization buffers to cut allocations
- Add rule limits and selector validation to release rules and free up memory when needed
- Skip CSS parsing/loading entirely when "Book's Embedded Style" is off

## Additional Context

- My test EPUB has been updated [here](https://github.com/jdk2pq/css-test-epub) to include a very large CSS file to test this out

---

### AI Usage

While CrossPoint doesn't have restrictions on AI tools in contributing, please be transparent about their usage as it 
helps set the right context for reviewers.

Did you use AI tools to help write this code? _**YES**_, Codex

PR #815, opened 2026-02-10
fix: Fix heap fragmentation causing XTC page buffer allocation failure

rojohyd:
## Summary

Fixes XTC ebook page rendering failures on boot by pre-allocating the page buffer when heap fragmentation is minimal.

- Pre-allocate ~96KB page buffer in XtcReaderActivity::onEnter()
- Reuse buffer across renders instead of per-render [malloc()](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html)
- Fall back to dynamic allocation if pre-allocation fails
- Proper cleanup in onExit() to prevent leaks

## Additional Context

- **Root Cause Discovery**: Post-boot heap fragmentation was identified only after adding `--without-restart` flag to the debug monitor, which allowed capture of boot-time logs without device reset. Analysis revealed 178KB free memory split into fragments too small for 96KB allocation.
- **Solution**: Reserve buffer early when heap is contiguous, reuse for all renders
- **Impact**: Eliminates allocation failures; slight memory overhead (~96KB)
- **Backward Compatible**: Full fallback to `malloc()` if needed; no API changes
- **Testing**: Focus on boot rendering, activity re-entry, and fallback scenarios

---

### AI Usage

### DIFF
--- a/src/activities/reader/XtcReaderActivity.cpp
+++ b/src/activities/reader/XtcReaderActivity.cpp
@@ -10,6 +10,7 @@
 #include <FsHelpers.h>
 #include <GfxRenderer.h>
 #include <HalStorage.h>
+#include <esp_heap_caps.h>
 
 #include "CrossPointSettings.h"
 #include "CrossPointState.h"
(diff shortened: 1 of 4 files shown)


src/activities/reader/XtcReaderActivity.cpp
```diff
@@ -203,21 +242,36 @@ void XtcReaderActivity::renderPage() {
...
+  bool pageBufferIsPrealloc = false;
+  if (preallocPageBuffer && pageBufferSize <= preallocPageBufferSize) {
+    // Reuse the reserved buffer when the requested page fits within it.
+    // This avoids repeated large `malloc`/`free` operations during rendering.
+    pageBuffer = preallocPageBuffer;
+    pageBufferIsPrealloc = true;
+  } else {
+    // Allocate page buffer
+    pageBuffer = static_cast<uint8_t*>(malloc(pageBufferSize));
+    if (!pageBuffer) {
```
ngxson, line 256: I think an `std::vector<uint8_t>` is a better container here, rather than `malloc` manually

The vector already handle `size()` and `resize()` under the hood, and RAII automatically free the vector once the owner is destroyed.

ngxson: Btw we recently have quite some troubles with memory management recently (ref https://github.com/crosspoint-reader/crosspoint-reader/pull/779), I think at some points we should investigate if stack-allocated vector and string implementation, like chromium's [stack_container.h](https://chromium.googlesource.com/chromium/chromium/+/master/base/stack_container.h), can be beneficial or not. CC @crosspoint-reader/firmware-maintainers if someone interested

icannotttt:
I have also tried doing it this way. Although I haven't tested your PR yet, I would like to ask if it might lead to increased memory usage, and whether the EPUB functionality would be affected.

itsthisjustin:
I believe this issue has been solved in other ways at this point. I'm going to close for now in an effort to clean up our PRs but feel free to comment back if this is something we need to bring back

PR #816, opened 2026-02-10
feat: Add auto-detection of usb device to monitor

jonasdiemer:
## Summary

usb device on MacOS depends on USB topology, so adding an auto-detect that picks the one tty.usbmodem* if there's only one.

---

### AI Usage

Did you use AI tools to help write this code? yes

### DIFF
--- a/scripts/debugging_monitor.py
+++ b/scripts/debugging_monitor.py
@@ -184,11 +184,17 @@ def update_graph(frame):
     plt.tight_layout()
 
 def detect_serial_port():
-    """Auto-detect the serial port based on the platform."""
-    if platform.system() == "Darwin":
+    """Auto-detect the serial port on Linux and macOS."""
+    system = platform.system()
+    if system == "Darwin":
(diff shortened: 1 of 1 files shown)


jpirnay:
This will fail on Windows machines. pyserial.tools.list_ports would be one os-independent way, but from personal experience it will fail to reliably detect *all* com-ports

jonasdiemer:
> This will fail on Windows machines. pyserial.tools.list_ports would be one os-independent way, but from personal experience it will fail to reliably detect _all_ com-ports

Thanks for the feedback - added a check to apply this only on MacOS (tested) and Linux (guessing it works, too).

jpirnay:
While extending the debug utility with other features (PR #820), I copied and extended your auto-dection logic, so that it now covers all 3 main platforms

daveallie:
To confirm, this is now obsolete and can be closed?

jonasdiemer:
Looks like it.

Issue #579, opened 2026-01-27
Unable to Exit when Connected to File Transfer


getthething:
### Affected Version

0.16.0

### Describe the Bug

During a file transfer (which works successfully) the "Exit" button doesn't respond on the X4. I have to sleep the device to get it to disconnect and return to the home screen. 

### Steps to Reproduce

1. Initiate files transfer (I use join hotspot but I haven't tested the other methods)
2. Attempt to exit on the X4 using the exit button.

### Expected Behavior

Exit the file transfer view and return to the previous menu or home screen.

### Relevant Log Output/Screenshots

```shell

```

rolfsky1:
I noticed this a while back -- then just recently I realized you need to *hold* the back button (__long press__) to exit the file transfer screen.

getthething:
> I noticed this a while back -- then just recently I realized you need to _hold_ the back button (**long press**) to exit the file transfer screen.

I tried this as well. I'll try it again but it didn't work the first time.

lukestein:
I run into this too… one of the few real bugs I'm currently aware of.

Long-press "exit" often works when click doesn't.

@crosspoint-reader/firmware-maintainers, any idea whom we can assign this to?

gokulp01:
@lukestein happy to work on the issue

daveallie:
The root cause here is that gpio updates are not happening every 10ms when in the file transfer activity but instead every ~100ms so quickly tapping the button gets missed by the code. There's an assumption that the gpio will get refetched as part of the interations loop in CrossPointWebServerActivity:

https://github.com/crosspoint-reader/crosspoint-reader/blob/bf87a7dc600b94b39dc2c11e47dfcd7267c3df95/src/activities/network/CrossPointWebServerActivity.cpp#L351-L355

Going to drop this loop down to 64 iterations to return back to the main loop and ensure that gpio updates happen. This should make the button much more responsive but not impact uploads (will test in PR).

gokulp01:
Thanks! I've been trying to figure out why this is happening for a while now. This makes sense!

PR #819, opened 2026-02-10
fix: Manually trigger GPIO update in File Browser mode

daveallie:
## Summary

* Manually trigger GPIO update in File Browser mode
  * Previously just assumed that the GPIO data would update automatically (presumably via yield), the data is currently updated in the main loop (and now here as well during the middle of the processing loop).
  * This allows the back button to be correctly detected instead of only being checked once every 100ms or so for the button state.

## Additional Context

* Fixes https://github.com/crosspoint-reader/crosspoint-reader/issues/579

---

### AI Usage

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced input state detection in the web server interface for more responsive and accurate user command recognition during high-frequency operations.

### DIFF
--- a/src/activities/network/CrossPointWebServerActivity.cpp
+++ b/src/activities/network/CrossPointWebServerActivity.cpp
@@ -348,7 +348,7 @@ void CrossPointWebServerActivity::loop() {
         // Yield and check for exit button every 64 iterations
         if ((i & 0x3F) == 0x3F) {
           yield();
-          // Force trigger an update of which buttons are being pressed so be have accurate stat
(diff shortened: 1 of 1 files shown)


daveallie:
@coderabbitai review

lukestein:
Testing on device and quick clicks of the "Exit" button seem more reliably detected in both "Network Transfer" and "Hotspot Mode" screens.

I noticed that one time, clicking exit in the hotspot mode caused the device to reboot rather than just return home, but I can't reproduce that.

PR #820, opened 2026-02-10
feat: Allow screenshot retrieval from device

jpirnay:
## Summary

* Add a small loop in main to be able to receive external commands, currently being sent via the debugging_monitor
* Implemented command: cmd:SCREENSHOT sends the currently displayed screen to the monitor, which will then store it to screenshot.bmp

## Additional Context

I was getting annoyed with taking tilted/unsharp photos of the device screen, so I added the ability to press Enter during the monitor execution and type SCREENSHOT to send a command. Could be extended in the future
[screenshot.bmp](https://github.com/user-attachments/files/25213230/screenshot.bmp)

---

### AI Usage

### DIFF
--- a/scripts/debugging_monitor.py
+++ b/scripts/debugging_monitor.py
@@ -363,12 +363,20 @@ def get_auto_detected_port() -> list[str]:
     elif system == "Windows":
         from serial.tools import list_ports
 
+        # Be careful with this pattern list - it should be specific
+        # enough to avoid picking up unrelated devices, but broad enough
+        # to catch all common USB-serial adapters used with ESP32
+        # Caveat: localized versions of Windows may have different descriptions,
+        # so we also check for specific VID:PID (but that may not cover all clones)
         pattern_list = ["CP210x", "CH340", "USB Serial"]
         found_ports = list_ports.comports()
         port_list = [
             port.device
             for port in found_ports
             if any(pat in port.description for pat in pattern_list)
+            or port.hwid.startswith(
+                "USB VID:PID=303A:1001"
+            )  # Add specific VID:PID for XTEINK X4
         ]
 
     return port_list


ngxson:
May worth mentioning that I'm working on mostly the same feature on https://github.com/crosspoint-reader/crosspoint-reader/pull/715 , except that it allows much more than only screenshot (for example, also allow using a host directory as fake SD card)

In the long term, I also want to allow controlling buttons from host PC, so may you reconsider if this PR is still necessary?

jpirnay:
> May worth mentioning that I'm working on mostly the same feature on #715 , except that it allows much more than only screenshot (for example, also allow using a host directory as fake SD card)
> 
> In the long term, I also want to allow controlling buttons from host PC, so may you reconsider if this PR is still necessary?

I wasn't aware of your work - looking great - definitely way more advanced than anything in this simple monitor 🥇 
I do believe both solutions have their merit - I am used to taking my (work) laptop with me all the time (which I may use according to company directives for smaller non-work-related tasks outside office hours) - and here a full Docker setup would not be possible.

ngxson:
Sounds ok then, I think the debugging system is quite still in early stage and some duplicated functionalities can be acceptable.

Just a small clarification though: my PR doesn't need docker to work, has minimal python dependency because the GUI is just a web page. The docker version is only required on the QEMU branch, which is currently discontinued.

src/main.cpp
```diff
@@ -369,6 +369,21 @@ void loop() {
...
+    String line = Serial.readStringUntil('\n');
+    if (line.startsWith("CMD:")) {
+      String cmd = line.substring(4);
+      cmd.trim();
+      if (cmd == "SCREENSHOT") {
+        Serial.printf("SCREENSHOT_START:%d\n", HalDisplay::BUFFER_SIZE);
+        uint8_t* buf = display.getFrameBuffer();
+        Serial.write(buf, HalDisplay::BUFFER_SIZE);
+        Serial.printf("SCREENSHOT_END\n");
+      }
```
ngxson, line 383: Just a small note that the scheduler can always interrupt any this task between 2 Serial operations, this can lead to something like:

```
SCREENSHOT_START:123
(some other tasks printing stuff)
(then, screen data)
SCREENSHOT_END
```

This is a real problem I mentioned in [my other PR](https://github.com/crosspoint-reader/crosspoint-reader/pull/715), and I resolved it by simply give the current task highest priority. But probably it's not very necessary here, as the screenshot is always be taken on command, not automatically.

In the future, I think it will be cleaner to delegate any host-device communication to a dedicated class. At least on my PR, I added `EmulationUtils.cpp` for this purpose.

Issue #812, opened 2026-02-09
ePub causes crash in v1


reukiodo:
### Affected Version

v1.0.0

### Describe the Bug

When opening a specific epub, it causes CrossPoint v1.0.0 to crash.

This is a regression from v0.16.0 where the epub opened correctly.

### Steps to Reproduce

1. install CrossPoint v0.16.0 (and reboot)
2. copy Mass Effect Complete.epub to uSD card
3. open book
4. install CrossPoint v1.0.0 (and reboot)

### Expected Behavior

CrossPoint v1.0.0 should open all the epubs previous version was able to.

### Relevant Log Output/Screenshots

```shell
I still can't figure out how to collect logs from the X4.
```

Robin-Law:
Same issue with "The Lamb Will Slaughter the Lion" by Margaret Killjoy.

I'm attaching a crash log and the css file that it was reading when it crashed.

[crosspoint-crash-log.txt](https://github.com/user-attachments/files/25194191/crosspoint-crash-log.txt)
[epub.css](https://github.com/user-attachments/files/25194190/epub.css)

Tried a couple of solutions:
* Turning "Book's Embedded Style" off does not change anything; it still attempts to parse the css file withe same crash log.
* Removing the comment at the top of the css file did not make a difference, same crash log.

Robin-Law:
I tried removing all comments from the epub.css and that DID work. So I think the issue is one of two things: 

- Comment parsing in general is broken. This seems unlikely to me, but possible. This CSS has a mix large block comments and comments at the end of lines.
- Parsing of empty rule blocks is broken. This file contains these rules:

```css
*[class^="TOCPartNumber"],
*[class^="TOCFrontmatterHead"],
*[class^="TOCBackmatterHead"],
*[class^="PartNumber"],
*[class^="VolumeNumber"],
*[class^="CollectionBookNumber"] {
  /*background: #999;*/
  }
*[class^="TOCChapterNumber"],
*[class^="ChapNumber"],
*[class^="TableNumber"],
*[class^="ChartNumber"],
*[class^="FigureNumber"] {
  /*background: #eee;*/
/*  margin-top: 3em; */
  }
```

... which are empty except for the commented lines. This seems even more suspicious to me.

jonasdiemer:
Thanks for the bug report. I can reproduce, so will look into fixing.

jonasdiemer:
[...]

PR #821, opened 2026-02-10
fix: Refactor CSS parser to not read whole file.

jonasdiemer:
## Summary

Large CSS files cause OOM error, see #812. This PR refactors the CSS parser to avoid this.

@jdk2pq AFAICS, the original CSS parsing came from you, I would appreciate your review (the refactor is done by Claude Code).

---

### AI Usage

### DIFF
--- a/lib/Epub/Epub/css/CssParser.cpp
+++ b/lib/Epub/Epub/css/CssParser.cpp
@@ -7,7 +7,7 @@
 
 namespace {
 
-// Buffer size for reading CSS files in chunks
+// Buffer size for reading CSS files
 constexpr size_t READ_BUFFER_SIZE = 512;
 
 // Maximum CSS file size we'll process (prevent memory issues)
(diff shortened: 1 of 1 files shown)


jonasdiemer:
Pushed a more simplified fix that only moves the stripping of comments into `readFileContent` without touching the parser. Done mainly for simplicity - making the parser "chunked" would actually enable us to load even larger CSS files (64KB is the current limit AFAICS).

@jdk2pq, looking forward to your thoughts.

jdk2pq:
I'll plan to review this tonight! (US Eastern time)

jonasdiemer:
Just found #779 also fixes the issue - will review that one (it's probably the proper solution).

jonasdiemer:
#779 implements chunk-based parsing of CSS and efficient filtering of comments - it's the better solution (after some fixes commented there).

PR #832, opened 2026-02-11
fix: Auto calculate the settings size on serialization

jpirnay:
## Summary

* The constant SETTINGS_CONST was hardcoded and needed to be updated whenever an additional setting was added
* This is no longer necessary as the settings size will be determined automatically on settings persistence

## Additional Context

* New settings need to be added (as previously) in saveToFile - that's it

---

### AI Usage

### DIFF
--- a/src/CrossPointSettings.cpp
+++ b/src/CrossPointSettings.cpp
@@ -6,7 +6,6 @@
 
 #include <cstring>
 #include <string>
-#include <vector>
 
 #include "fontIds.h"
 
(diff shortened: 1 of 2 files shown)


src/CrossPointSettings.cpp
```diff
@@ -86,42 +87,50 @@ bool CrossPointSettings::saveToFile() const {
...
+      [&](FsFile& f) { serialization::writePod(f, frontButtonConfirm); },
+      [&](FsFile& f) { serialization::writePod(f, frontButtonLeft); },
+      [&](FsFile& f) { serialization::writePod(f, frontButtonRight); },
+      [&](FsFile& f) { serialization::writePod(f, fadingFix); },
+      [&](FsFile& f) { serialization::writePod(f, embeddedStyle); },
+      // New fields need to be added at end for backward compatibility
+  };
+
   serialization::writePod(outputFile, SETTINGS_FILE_VERSION);
-  serialization::writePod(outputFile, SETTINGS_COUNT);
```
osteotek, line 90: since we are not writing SETTINGS_COUNT anymore, will it be a problem for upgrading from old settings file to new one?

jpirnay: We are writing the count:
```
  serialization::writePod(outputFile, static_cast<uint8_t>(writers.size()));  <<<< SETTINGS_COUNT
  for (const auto& writer : writers) {
    writer(outputFile);
  }
```

osteotek: ah, thanks, missed that

src/CrossPointSettings.cpp
```diff
@@ -86,42 +87,50 @@ bool CrossPointSettings::saveToFile() const {
     return false;
   }
 
+  // Define the serialization writers in order
+  std::vector<std::function<void(FsFile&)>> writers = {
+      [&](FsFile& f) { serialization::writePod(f, sleepScreen); },
+      [&](FsFile& f) { serialization::writePod(f, extraParagraphSpacing); },
+      [&](FsFile& f) { serialization::writePod(f, shortPwrBtn); },
+      [&](FsFile& f) { serialization::writePod(f, statusBar); },
```
ngxson: This looks quite strange to use lambda function this way. IIRC each time a lambda is created, a bit of heap memory will be allocated to hold the bind params, so this is not very good in term of memory usage

Instead, you can simply add a wrapper for `serialization::writePod`, something like:

```cpp
template<typename T>
static void writePodSafe(FsFile& f, T val) {
  serialization::writePod(f, val);
  // add your code here
}
```

src/CrossPointSettings.cpp
```diff
@@ -86,42 +100,50 @@ bool CrossPointSettings::saveToFile() const {
     return false;
   }
 
+  // Define the serialization writers in order
+  std::vector<std::function<void(FsFile&)>> writers = {
+      [&](FsFile& f) { writePodSafe(f, sleepScreen); },
```
ngxson: what I mean is that you don't need lambda function here, just replace `serialization::writePod` with the new `writePodSafe`

src/CrossPointSettings.cpp
```diff
@@ -19,10 +21,22 @@ void readAndValidate(FsFile& file, uint8_t& member, const uint8_t maxValue) {
   }
 }
 
+// Template wrappers to avoid lambda overhead and allow for future extensions
+template <typename T>
+static void writePodSafe(FsFile& f, const T& val) {
+  serialization::writePod(f, val);
+  // add your code here (e.g., error checking, logging)
```
ngxson: huh??

src/CrossPointSettings.cpp
jpirnay: The lambda functions are intentional here - we need the writers vector to automatically calculate the number of settings via writers.size() (line 139). This ensures the count is always correct when we add new settings, without having to manually update a hardcoded constant. The lambdas capture each member variable and defer the actual write operation until the loop at lines 140-142.

src/CrossPointSettings.cpp
jpirnay: I think I misunderstood your earlier recommendation about templates, so I reverted that overhaed

src/CrossPointSettings.cpp
ngxson: IMO this pattern is not clean. std::vector is heap-allocated, and lambda function will compiled to huge overhead each time we create one. std::vector of lambda function is the worst of all I would say.

Instead, you should have a wrapper of `writePod` and `writeString` like this:

```cpp
class CrossPointSettings {
  bool is_counting = false;
  int item_count = 0;

  template <typename T>
  void writeItem(const T& value) {
    if (is_counting) {
      item_count++;
    } else {
      serialization::writePod(file, value);
    }
  }

  // same for writeItemString

  void writeSettings() {
    writeItem(sleepScreen);
    writeItem(extraParagraphSpacing);
    // etc
  }

  void saveToFile() {
    is_counting = true;
    writeSettings();
    // write header with item_count
    is_counting = false;
    writeSettings();
  }
```

The suggestion above use just ~8 more bytes of RAM and generates minimum assembly code when compiled.

ngxson:
This adds only ~100 bytes to the flash usage, compared to ~2500 bytes if you use the lambda version

PR #837, opened 2026-02-11
style: Enhance server HTML files (phase 2)

cdmoro:
## Summary

Overall improvement of the server HTML files, use of SVG (minified) instead of emojis, javascript enhacement, better support for mobile, etc.

### File size increase

I know that file size is very important due to device memory restrictions, so I calculated the size of the modified files.

|File|Original size|New size|Difference (%)
|:---|---:|---:|---:|
|HomePage|`4.18 KB`|`4.94 KB`|`+18.18%`|
|FilesPage|`43.7 KB`|`46.3 KB`|`+5.95%`|
|SettingsPage|`11.2 KB`|`12.1 KB`|`+8.04%`|

## Changes

- Add logo
- Enhance responsiveness
- Move from specific styles (`.upload-button`) to Bootstrap-like styles (`.btn-warning`)
- Clean JavaScript code (for example, use of `switch` instead of multiples `if`s)
- Use SVGs instead of emojis
- Improve header and footer

### Changes per file

- Home Page
  - Improve responsiveness
- Files Page
  - Highlight row when is selected
  - Use `dialog` (HTML element) instead of divs
  - Improve `folder-summary` legend, only shows what is in the folder:
    - Full: `1 folder, 5 files, 3.93 MB`
    - Only files `5 files, 3.93 MB`
    - Only folders `3 folders` 
- Settings Page
  - Hide setting fields when a setting type is invalid
  - Implement toast instead of alerts

### About the accent color

The green-ish color was taken from the title of this repository:
image

### Phases

Next steps/Phases:

1. Light/Dark themes [[MERGED](https://github.com/crosspoint-reader/crosspoint-reader/pull/1006)]
1. HTML enhancement, use SVG instead of emojis [THIS PR]
1. Use Vite + Typescript to improve DX and have better minification
2. Load external CSS file to avoid duplication

### Not included in this PR (hopefully more PRs will come)

- [ ] Favicon
- [ ] Load external CSS files (it would be super useful to avoid code duplication and reduce file size)
- [ ] Load external JS files (same as above)
- [ ] Internationalisation
- [ ] Use of LocalStorage to save potential settings (like theme, language, etc.)
- [ ] Light/Dark toggle button

Most of this are tied to the ability to import external files into the HTMLs.

### Home Page

Desktop

|Before|After (light)|After (dark)|
|---|---|---|
|image|image|image|

Mobile

|Before|After (light)|After (dark)|
|---|---|---|
|image|image|image|

### Files Page

Desktop

|Before|After (light)|After (dark)|
|---|---|---|
|image|image|image|

Mobile

|Before|After (light)|After (dark)|
|---|---|---|
|image|image|image|

### Settings Page

Desktop

|Before|After (light)|After (dark)|
|---|---|---|
|image|image|image|

Mobile

|Before|After (light)|After (dark)|
|---|---|---|
|image|image|image|

---

### AI Usage


jdk2pq:
@cdmoro Thank you for the contribution! Here are some of my thoughts:

- I'm a sucker for dark mode, so I'm always happy to see support for it. The actual CSS changes needed for this kind of support seem minor and not a significant factor in the increase in size of the HTML files.
- However, I am a bit hesitant to approve this as is due to the file size increase in the HTML files.
- Switching from an emoji glyph to an SVG icon has tradeoffs when it comes to file size: the emoji glyph is a text character and significantly smaller than its SVG equivalent, but emojis will render differently on different devices and SVGs solve that. Personally, I'm okay with the tradeoff of rendering emojis differently on different devices, since it results in a smaller footprint in these HTML files. I would suggest reverting back to emojis for that reason.
- We currently have duplicated CSS across the 3 HTML files. I _think_ we can factor these shared styles out into a .css file and include it with a `<link>` tag in each file to save on file size here.

TL;DR I do like this change and think it makes the Wifi Upload HTML look better, but I think we need some reduction in the file size increase that these changes come with before we can get this in.

cdmoro:
@jdk2pq I'm totally agree with you. In fact, in the PR description I mentioned some of your points. It would be nice if we could put the JS and the CSS files in separate files (but I don't know if the server could serve those files with the current implementation), the same with the SVG icons.

I could try to optimize the files even more, for example, in the FilesPage.html file there are a lot of repetitive functions that can be grouped using parameters (each modal has its own openModal and closeModal function).

But apart from this change, I think we can go even further, have a Vite + Vanilla JS + TS web project that build the files (minified by Vite, with inline CSS and JS if needed), which improve a lot the developer experience (I was doing some test running this locally and hitting 192.168.0.68 successfully, which is great as we could see the changes live without flashing the device). But I will leave this for a future PR :D

Btw, thanks for the feedback, I will work on it.

cdmoro:
@jdk2pq  regarding the svg icons, I could try to reduce then as much as I can, removing the unnecessary ones like the upload button, for instance.

Eloren1:
Header logo doesn't look like original one

cdmoro:
> Header logo doesn't look like original one

@Eloren1 you're right, but I just copied the svg from https://github.com/crosspoint-reader/crosspoint-reader/blob/master/src/images/logo.svg which doesn't have the white background, so in that case I'd need to edit that SVG myself and use the modified version. 

Good catch, thanks!

cdmoro:
BTW, @jdk2pq it would be really useful to have the ability of importing external CSSs and SVGs, this will avoid duplication and the files will be smaller :) would you think that is possible?

btw, I'm working on another PR, a simpler one (https://github.com/crosspoint-reader/crosspoint-reader/pull/1006) that only adds the light & dark themes, I think we could move forward with that in the meantime. Let me know what you think.

cdmoro:
> Header logo doesn't look like original one

Logo updated @Eloren1 :)

cdmoro:
@coderabbitai review

cdmoro:
@lukestein @Eloren1 @jdk2pq I really appreciate your review here 😃

cdmoro:
I minified the icons so I can also attach the unminified icons.svg to update the icons in the future, I used svgo to minify it.

Uri-Tauber:
Hi @cdmoro, this looks fantastic. Really strong work. I’d love to see this merged.

Regarding the concerns raised by @jdk2pq about the increased footprint of SVG icons compared to the original emojis: my view is that the emojis should be removed regardless. They make the interface looks cheap and AI-generated.

My suggestion is to swap your current SVGs for a lighter set, such as those from [Bytesize Icons](https://github.com/danklammer/bytesize-icons), which have a much smaller footprint.
Alternatively, you could incorporate the overflow menu from #619 and remove both the emojis and SVGs entirely.

Also, a few things have changed since this PR was opened, so it would be great if you could rebase on top of the current `master` when you have a chance—I know that can be a bit tedious.

Thanks again for the contribution!

cdmoro:
Hey, thanks for the comment @Uri-Tauber, I will update the PR and check the icon library you mentioned :)

itsthisjustin:
@cdmoro Can we get the merge conflicts solved and then get this merged asap? It looks good

Uri-Tauber:
> Hey, thanks for the comment @Uri-Tauber, I will update the PR and check the icon library you mentioned :)

Hey @cdmoro, do you think you will have the time to solve the conflicts here?

itsthisjustin:
Closing this as stale and I think if we are going to re-do the web UI we should do a real pass at the polish. Thanks so much for the effort put into this though @cdmoro

Issue #588, opened 2026-01-28
Regression of sleep cover image


jpirnay:
### Affected Version

0.16

### Describe the Bug

I have updated my x4 device to 0.16. For the attached epub the cover is not rendered for the sleep screen - 0.15 is doing it.

### Steps to Reproduce

1. Open attached book
2. Send X4 to sleep
3. Only the generic Xpoint logo appears

### Expected Behavior

[issue.zip](https://github.com/user-attachments/files/24907296/issue.zip)

### Relevant Log Output/Screenshots

```shell

```

LZando:
I suggest you to check as follow:
- Settings --> Display --> Sleep Screen

Then press "Toggle" till "Cover" option

edyf:
I have the same problem, the cover was shown in 0.15 but not in 0.16 anymore after I cleared the cache. The "cover" option is still set in the sleep screen settings.

LZando:
@edyf Does the problem persist with book uploads after the update to 0.16?

edyf:
Ok, it might be a problem with the epub file that I was reading when I cleared the cache. When I open other books then the cover shows up correctly.

jpirnay:
The root cause is most likely either an invalid thumbnail and / or a missing cover metadata information in the books opf.
I had encountered both variants and have provided a fix in: PR #830

steka:
I think this issue can be closed now. It seems to work just fine in [v1.2.0](https://github.com/crosspoint-reader/crosspoint-reader/releases/tag/1.2.0)

PR #838, opened 2026-02-11
fix: Use fallback logic for epub cover extraction

jpirnay:
## Summary

* Some epub covers don't get recognized and rendered - now we try common cover filenames when no official cover is declared
* Additionally corrupt cover thumbnails are detected and will be recreated

## Additional Context

*Added fallback logic to try cover.jpg and OEBPS/cover.jpg when coverItemHref is empty
* Made extension checking case-insensitive
* Added debug output for sleep screen cover detection
* Now both the home screen thumbnails and sleep screen covers will work with EPUBs that have cover.jpg files but don't declare them properly in the OPF metadata. The system will automatically find and use these covers for both display purposes.
* I have encountered both phenomenons (missing cover info in metadata.opf and corrupted/empty thumb) - also reported in #588 
---

### AI Usage

Did you use AI tools to help write this code? PARTIALLY mainly for issue detection

### DIFF
--- a/src/MappedInputManager.h
+++ b/src/MappedInputManager.h
@@ -15,6 +15,7 @@ class MappedInputManager {
 
   explicit MappedInputManager(HalGPIO& gpio) : gpio(gpio) {}
 
+  void update() const { gpio.update(); }
   bool wasPressed(Button button) const;
   bool wasReleased(Button button) const;
   bool isPressed(Button button) const;
--- a/src/RecentBooksStore.cpp
+++ b/src/RecentBooksStore.cpp
@@ -83,7 +83,7 @@ RecentBook RecentBooksStore::getDataFromBook(std::string path) const {
     lastBookFileName = path.substr(lastSlash + 1);
   }
 
-  Serial.printf("Loading recent book: %s\n", path.c_str());
+  Serial.printf("[%lu] [RBS] Loading recent book: %s\n", millis(), path.c_str());
 
   // If epub, try to load the metadata for title/author and cover
   if (StringUtils::checkFileExtension(lastBookFileName, ".epub")) {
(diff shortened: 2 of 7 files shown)


jpirnay:
Just looking at PR #830 - I believe we need both, proper opf analysis plus fallback and corruption check

osteotek:
would be interesting to compare how Papyrix fork detects covers - https://github.com/bigbag/papyrix-reader/blob/main/lib/Epub/Epub.cpp#L452

jpirnay:
> would be interesting to compare how Papyrix fork detects covers - https://github.com/bigbag/papyrix-reader/blob/main/lib/Epub/Epub.cpp#L452

I just added all variants that Papyrix checks, too

lib/Epub/Epub.cpp
```diff
@@ -458,9 +461,19 @@ std::string Epub::getCoverBmpPath(bool cropped) const {
...
+    // is this a valid cover or just an empty file we created to mark generation attempts?
+    invalid = !isValidThumbnailBmp(getCoverBmpPath(cropped));
+    if (invalid) {
+      // Remove the old invalid cover so we can attempt to generate a new one
+      Storage.remove(getCoverBmpPath(cropped).c_str());
+      Serial.printf("[%lu] [EBP] Previous cover generation attempt failed for %s mode, retrying\n", millis(),
+                    cropped ? "cropped" : "fit");
+    } else {
+      return true;
+    }
```
daveallie, line 476: This will cause sleep covered to attempt to generate every time even if it crashes out in previous runs, how do you feel about removing this and instead relying on users using the "Delete Book Cache" functionality within the book menu?

osteotek:
@jpirnay would it be possible to fix conflicts and address suggestions?

jpirnay:
I will close this one and provide a new one, as some things did change in the meantime

PR #843, opened 2026-02-11
feat: Add central logging pragma

jpirnay:
## Summary

* Definition and use of a central LOG function, that can later be extended or completely be removed (for public use where debugging information may not be required) to save flash by suppressing the -DENABLE_SERIAL_LOG like in the slim branch

## Additional Context
* By using the central logger the usual:
```
#include <HardwareSerial.h>
...
  Serial.printf("[%lu] [WCS] Obfuscating/deobfuscating %zu bytes\n", millis(), data.size());
```
would then become
```
#include <Logging.h>
...
  LOG_DBG("WCS", "Obfuscating/deobfuscating %zu bytes", data.size());
```
You do have ``LOG_DBG`` for debug messages, ``LOG_ERR`` for error messages and ``LOG_INF`` for informational messages. Depending on the verbosity level defined (see below) soe of these message types will be suppressed/not-compiled.

* The normal compilation (default) will create a firmware.elf file of 42.194.356 bytes, the same code via slim will create 42.024.048 bytes - 170.308 bytes less
* Firmware.bin : 6.469.984 bytes for default, 6.418.672 bytes for slim - 51.312 bytes less

### AI Usage

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -370,17 +370,18 @@ void loop() {
     lastMemPrint = millis();
   }
 
-  // Handle incoming serial commands
-  if (Serial.available() > 0) {
-    String line = Serial.readStringUntil('\n');
+  // Handle incoming serial commands,
+  // nb: we use logSerial from logging to avoid deprecation warnings
+  if (logSerial.available() > 0) {
+    String line = logSerial.readStringUntil('\n');
     if (line.startsWith("CMD:")) {
       String cmd = line.substring(4);
       cmd.trim();
       if (cmd == "SCREENSHOT") {
-        Serial.printf("SCREENSHOT_START:%d\n", HalDisplay::BUFFER_SIZE);
+        logSerial.printf("SCREENSHOT_START:%d\n", HalDisplay::BUFFER_SIZE);
         uint8_t* buf = display.getFrameBuffer();
-        Serial.write(buf, HalDisplay::BUFFER_SIZE);
-        Serial.printf("SCREENSHOT_END\n");
+        logSerial.write(buf, HalDisplay::BUFFER_SIZE);
+        logSerial.printf("SCREENSHOT_END\n");
       }
     }
   }


Uri-Tauber:
Very Nice!

lib/Logging/Logging.h
ngxson, line 1: I would suggest changing `LOG` --> `LOGI` for now (log info), and maybe have `LOGD` for debugging in the future. `LOGE` (error) can be added in a next PR, or the same PR if you like

I think in addition to `ENABLE_SERIAL_LOG`, we can also define `SERIAL_LOG_LEVEL` to filter out debugging log on release build in the future

lib/Logging/Logging.h
```diff
@@ -0,0 +1,15 @@
+#ifndef LOGGING_H
+#define LOGGING_H
```
ngxson:
```suggestion
#pragma once
```

making it aligned with the rest of the code base

jpirnay:
* I extended / renamed the logging macros to
  - LOG_ERR to be used for error messages
  - LOG_INF  to be used for informational messages
  - LOG_DBG to be used for debug messages
* Added a LOG_LEVEL define that controls log verbosity levels:
  - 0: ERR only
  - 1: ERR + INF
  - 2: ERR + INF + DBG
 If LOG_LEVEL is not defined, it defaults to 0
* The logging macros are now conditionally defined based on both ENABLE_SERIAL_LOG (to enable/disable logging entirely) and LOG_LEVEL (to control verbosity)

I did my best to assign LOG_ERR where I found them appropriate, but I am sure I missed quite a lot

jpirnay:
| Profile         | Verbosity         | .elf size | Delta | .bin size | Delta |
|-----------------|------------------|-----------|-------|-----------|-------|
| slim            | suppressed       | 42.024.044  | -    | 6.418.672   | -     |
| gh_release      | err              | 42.048.784  | +24.740 | 6.432.384   | +13.712 |
| gh_release_rc   | err +inf         | 42.058.268  | +34.224 | 6.432.736   | +14.064 |
| default         | err +inf +dbg    | 42.198.452  | +174.408| 6.473.376   | +54.704 |

platformio.ini
```diff
@@ -54,15 +54,31 @@ extends = base
...
+  -DENABLE_SERIAL_LOG
+  -DLOG_LEVEL=0 ; Set log level to error for release builds  
 
 [env:gh_release_rc]
 extends = base
 build_flags =
   ${base.build_flags}
   -DCROSSPOINT_VERSION=\"${crosspoint.version}-rc+${sysenv.CROSSPOINT_RC_HASH}\"
+  -DENABLE_SERIAL_LOG
+  -DLOG_LEVEL=1 ; Set log level to info for release candidate builds
```
ngxson, line 75: In general, I think both RC and release should have the same level=info, WDYT @daveallie ?

ngxson:
Asking for extra review from @coderabbitai since the change is large, just have a quick look to see if any log statement are mislabeled (for example, LOG_INF --> LOG_ERR and vice versa)

ngxson:
@jpirnay also as a safe-guard, I think we can define `Serial` to something else in one of the headers, so if any on-going PR try to log to `Serial`, an error will be triggered at compile time

```cpp
#ifdef Serial // in case Serial is defined in Arduino header
#undef Serial
#endif
#define Serial static_assert(false && "use LOG_* macro instead");
```

jpirnay:
> @jpirnay also as a safe-guard, I think we can define `Serial` to something else in one of the headers, so if any on-going PR try to log to `Serial`, an error will be triggered at compile time
> 
> ```c++
> #ifdef Serial // in case Serial is defined in Arduino header
> #undef Serial
> #endif
> #define Serial static_assert(false && "use LOG_* macro instead");
> ```
My naive attempt to do that in Logging.h like this:
```
...
HardwareSerial& logSerial = Serial;
....

#ifdef Serial  // in case Serial is defined in Arduino header
#undef Serial
#endif
#define Serial static_assert(false && "use LOG_* macro instead");
```
will cause framework libraries to fail compiling

ngxson:
Something like this will work:

```cpp

static HWCDC& logSerial = Serial;

...

class MySerialImpl : public Print {
 public:
  void begin(unsigned long baud) {
    logSerial.begin(baud);
  }
  __attribute__((deprecated("Use LOG_* macro instead")))
  size_t printf(const char * format, ...);
  size_t write(uint8_t b) override;
  size_t write(const uint8_t* buffer, size_t size) override;
  void flush() override;
  static MySerialImpl instance;
};

#define Serial MySerialImpl::instance

```

The version above is quite draft, please adapt it properly. Not sure if there are any `__attribute__` that throw an error on compile, instead of a warning.

jpirnay:
That works now as intended - thank you for your guidance

| Profile         | Verbosity      | firmware.elf | Delta    | firmware.bin | Delta    |
|-----------------|---------------|--------------|----------|--------------|----------|
| slim            | suppressed     | 42.031.068   | -       | 6.417.008    | -        |
| gh_release      | err            | 42.057.876   | 26.808   | 6.435.824    | 18.816   |
| gh_release_rc   | err +inf       | 42.063.268   | 32.200   | 6.436.192    | 19.184   |
| default         | err +inf +dbg  | 42.201.592   | 170.524  | 6.473.536    | 56.528   |

ngxson:
I managed to save ~10KB flash space via the last commit, please give it a try

ngxson (approved):

osteotek:
Having slim build is nice, but I think the default build with all debugging messages enabled will always be the common denominator

osteotek:
Build from current master branch gives me firmware.bin size 6302416 bytes, default build from this PR's branch - 6294336 bytes. About 8kb size reduction, nice

PR #857, opened 2026-02-12
feat: Add dictionary word lookup feature

kurtisgrant:
## Summary

- **Dictionary lookup from reader**: Select a word on the current page to look up its definition. Supports all screen orientations, hyphenated/wrapped words, and position-based row navigation with full wrapping.
- **StarDict 3 format**: Uses `.idx` (sorted binary index) and `.dict` (raw definitions) files. 
- **On first lookup**: builds a sparse offset table in RAM (~800 bytes) for O(log n) binary search + short linear scan. Subsequent lookups are near-instant.
- **Lookup history**: View and manage previously looked up words. Long-press select to delete entries.
- **Reader Menu Items Added**: "Lookup" and "Lookup History" appear when the expected dictionary files are found on the SD card root.

## Setup

Place `dictionary.idx`, `dictionary.dict`, and `dictionary.ifo` (StarDict 3 format) on the SD card root. The menu items appear automatically when the files are detected.

## Test plan

- [X] Look up words near the start, middle, and end of the dictionary
- [X] Look up a hyphenated/wrapped word by selecting either half
- [X] Verify word selection wrapping (left/right across rows, up/down wrap top-bottom)
- [X] Verify definition page turning in all orientations
- [X] Verify Lookup History label in menu and history screen
- [X] Test with CW, CCW, portrait, and inverted orientations

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/activities/network/CrossPointWebServerActivity.cpp
+++ b/src/activities/network/CrossPointWebServerActivity.cpp
@@ -348,6 +348,9 @@ void CrossPointWebServerActivity::loop() {
         // Yield and check for exit button every 64 iterations
         if ((i & 0x3F) == 0x3F) {
           yield();
+          // Force trigger an update of which buttons are being pressed so be have accurate state
+          // for back button checking
+          mappedInput.update();
           // Check for exit button inside loop for responsiveness
           if (mappedInput.wasPressed(MappedInputManager::Button::Back)) {
             onGoBack();
--- a/src/activities/network/CrossPointWebServerActivity.cpp
+++ b/src/activities/network/CrossPointWebServerActivity.cpp
@@ -348,9 +348,6 @@ void CrossPointWebServerActivity::loop() {
         // Yield and check for exit button every 64 iterations
         if ((i & 0x3F) == 0x3F) {
           yield();
-          // Force trigger an update of which buttons are being pressed so be have accurate state
-          // for back button checking
-          mappedInput.update();
           // Check for exit button inside loop for responsiveness
           if (mappedInput.wasPressed(MappedInputManager::Button::Back)) {
             onGoBack();
--- a/src/activities/network/CrossPointWebServerActivity.cpp
+++ b/src/activities/network/CrossPointWebServerActivity.cpp
@@ -37,7 +37,7 @@ void CrossPointWebServerActivity::taskTrampoline(void* param) {
 void CrossPointWebServerActivity::onEnter() {
   ActivityWithSubactivity::onEnter();
 
-  Serial.printf("[%lu] [WEBACT] [MEM] Free heap at onEnter: %d bytes\n", millis(), ESP.getFreeHeap());
+  LOG_DBG("WEBACT] [MEM", "Free heap at onEnter: %d bytes", ESP.getFreeHeap());
 
   renderingMutex = xSemaphoreCreateMutex();
 
@@ -58,7 +58,7 @@ void CrossPointWebServerActivity::onEnter() {
   );
 
   // Launch network mode selection subactivity
-  Serial.printf("[%lu] [WEBACT] Launching NetworkModeSelectionActivity...\n", millis());
+  LOG_DBG("WEBACT", "Launching NetworkModeSelectionActivity...");
   enterNewActivity(new NetworkModeSelectionActivity(
       renderer, mappedInput, [this](const NetworkMode mode) { onNetworkModeSelected(mode); },
       [this]() { onGoBack(); }  // Cancel goes back to home
@@ -68,7 +68,7 @@ void CrossPointWebServerActivity::onEnter() {
 void CrossPointWebServerActivity::onExit() {
   ActivityWithSubactivity::onExit();
 
-  Serial.printf("[%lu] [WEBACT] [MEM] Free heap at onExit start: %d bytes\n", millis(), ESP.getFreeHeap());
+  LOG_DBG("WEBACT] [MEM", "Free heap at onExit start: %d bytes", ESP.getFreeHeap());
 
   state = WebServerActivityState::SHUTTING_DOWN;
 
@@ -80,7 +80,7 @@ void CrossPointWebServerActivity::onExit() {
 
   // Stop DNS server if running (AP mode)
   if (dnsServer) {
-    Serial.printf("[%lu] [WEBACT] Stopping DNS server...\n", millis());
+    LOG_DBG("WEBACT", "Stopping DNS server...");
     dnsServer->stop();
     delete dnsServer;
     dnsServer = nullptr;
@@ -91,39 +91,39 @@ void CrossPointWebServerActivity::onExit() {
 
   // Disconnect WiFi gracefully
   if (isApMode) {
-    Serial.printf("[%lu] [WEBACT] Stopping WiFi AP...\n", millis());
+    LOG_DBG("WEBACT", "Stopping WiFi AP...");
     WiFi.softAPdisconnect(true);
   } else {
-    Serial.printf("[%lu] [WEBACT] Disconnecting WiFi (graceful)...\n", millis());
+    LOG_DBG("WEBACT", "Disconnecting WiFi (graceful)...");
     WiFi.disconnect(false);  // false = don't erase credentials, send disconnect frame
   }
   delay(30);  // Allow disconnect frame to be sent
 
-  Serial.printf("[%lu] [WEBACT] Setting WiFi mode OFF...\n", millis());
+  LOG_DBG("WEBACT", "Setting WiFi mode OFF...");
   WiFi.mode(WIFI_OFF);
   delay(30);  // Allow WiFi hardware to power down
 
-  Serial.printf("[%lu] [WEBACT] [MEM] Free heap after WiFi disconnect: %d bytes\n", millis(), ESP.getFreeHeap());
+  LOG_DBG("WEBACT] [MEM", "Free heap after WiFi disconnect: %d bytes", ESP.getFreeHeap());
 
   // Acquire mutex before deleting task
-  Serial.printf("[%lu] [WEBACT] Acquiring rendering mutex before task deletion...\n", millis());
+  LOG_DBG("WEBACT", "Acquiring rendering mutex before task deletion...");
   xSemaphoreTake(renderingMutex, portMAX_DELAY);
 
   // Delete the display task
-  Serial.printf("[%lu] [WEBACT] Deleting display task...\n", millis());
+  LOG_DBG("WEBACT", "Deleting display task...");
   if (displayTaskHandle) {
     vTaskDelete(displayTaskHandle);
     displayTaskHandle = nullptr;
-    Serial.printf("[%lu] [WEBACT] Display task deleted\n", millis());
+    LOG_DBG("WEBACT", "Display task deleted");
   }
 
   // Delete the mutex
-  Serial.printf("[%lu] [WEBACT] Deleting mutex...\n", millis());
+  LOG_DBG("WEBACT", "Deleting mutex...");
   vSemaphoreDelete(renderingMutex);
   renderingMutex = nullptr;
-  Serial.printf("[%lu] [WEBACT] Mutex deleted\n", millis());
+  LOG_DBG("WEBACT", "Mutex deleted");
 
-  Serial.printf("[%lu] [WEBACT] [MEM] Free heap at onExit end: %d bytes\n", millis(), ESP.getFreeHeap());
+  LOG_DBG("WEBACT] [MEM", "Free heap at onExit end: %d bytes", ESP.getFreeHeap());
 }
 
 void CrossPointWebServerActivity::onNetworkModeSelected(const NetworkMode mode) {
@@ -133,7 +133,7 @@ void CrossPointWebServerActivity::onNetworkModeSelected(const NetworkMode mode) 
   } else if (mode == NetworkMode::CREATE_HOTSPOT) {
     modeName = "Create Hotspot";
   }
-  Serial.printf("[%lu] [WEBACT] Network mode selected: %s\n", millis(), modeName);
+  LOG_DBG("WEBACT", "Network mode selected: %s", modeName);
 
   networkMode = mode;
   isApMode = (mode == NetworkMode::CREATE_HOTSPOT);
@@ -155,11 +155,11 @@ void CrossPointWebServerActivity::onNetworkModeSelected(const NetworkMode mode) 
 
   if (mode == NetworkMode::JOIN_NETWORK) {
     // STA mode - launch WiFi selection
-    Serial.printf("[%lu] [WEBACT] Turning on WiFi (STA mode)...\n", millis());
+    LOG_DBG("WEBACT", "Turning on WiFi (STA mode)...");
     WiFi.mode(WIFI_STA);
 
     state = WebServerActivityState::WIFI_SELECTION;
-    Serial.printf("[%lu] [WEBACT] Launching WifiSelectionActivity...\n", millis());
+    LOG_DBG("WEBACT", "Launching WifiSelectionActivity...");
     enterNewActivity(new WifiSelectionActivity(renderer, mappedInput,
                                                [this](const bool connected) { onWifiSelectionComplete(connected); }));
   } else {
@@ -171,7 +171,7 @@ void CrossPointWebServerActivity::onNetworkModeSelected(const NetworkMode mode) 
 }
 
 void CrossPointWebServerActivity::onWifiSelectionComplete(const bool connected) {
-  Serial.printf("[%lu] [WEBACT] WifiSelectionActivity completed, connected=%d\n", millis(), connected);
+  LOG_DBG("WEBACT", "WifiSelectionActivity completed, connected=%d", connected);
 
   if (connected) {
     // Get connection info before exiting subactivity
@@ -183,7 +183,7 @@ void CrossPointWebServerActivity::onWifiSelectionComplete(const bool connected) 
 
     // Start mDNS for hostname resolution
     if (MDNS.begin(AP_HOSTNAME)) {
-      Serial.printf("[%lu] [WEBACT] mDNS started: http://%s.local/\n", millis(), AP_HOSTNAME);
+      LOG_DBG("WEBACT", "mDNS started: http://%s.local/", AP_HOSTNAME);
     }
 
     // Start the web server
@@ -199,8 +199,8 @@ void CrossPointWebServerActivity::onWifiSelectionComplete(const bool connected) 
 }
 
 void CrossPointWebServerActivity::startAccessPoint() {
-  Serial.printf("[%lu] [WEBACT] Starting Access Point mode...\n", millis());
-  Serial.printf("[%lu] [WEBACT] [MEM] Free heap before AP start: %d bytes\n", millis(), ESP.getFreeHeap());
+  LOG_DBG("WEBACT", "Starting Access Point mode...");
+  LOG_DBG("WEBACT] [MEM", "Free heap before AP start: %d bytes", ESP.getFreeHeap());
 
   // Configure and start the AP
   WiFi.mode(WIFI_AP);
@@ -216,7 +216,7 @@ void CrossPointWebServerActivity::startAccessPoint() {
   }
 
   if (!apStarted) {
-    Serial.printf("[%lu] [WEBACT] ERROR: Failed to start Access Point!\n", millis());
+    LOG_ERR("WEBACT", "ERROR: Failed to start Access Point!");
     onGoBack();
     return;
   }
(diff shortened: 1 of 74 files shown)


pepastach:
As a non-native English speaker I would LOVE to have this feature. Really hoping this PR can make it into one of future releases.

kurtisgrant:
[eng-dictionary.zip](https://github.com/user-attachments/files/25271916/eng-dictionary.zip)
These are the dictionary files I used, generated I believe from the ~108k entry, GNU Collaborative International Dictionary of English

Uri-Tauber:
I’m really excited about this feature — it’s incredibly useful! Thank you for the effort. I hope it would be merged.

I tested it on my device and it works well. The lookup is fast enough for practical use (aside from the initial indexing), and I’ve confirmed that, beyond the ~1 kB required for that first indexing, there are no memory leaks.

lookup

**A few notes:** 
UX Improvement: When in lookup mode, opening the dictionary page for a specific word and then pressing Back should return to lookup mode, not to the main reader view.
This allows the user to search another word immediately; a second Back press would then return to reader mode.

Performance: Every time the device wake from sleep, the first lookup takes several seconds. Is it possible to persist the offset table across sessions?

kurtisgrant:
## New changes

### Back from definition returns to word select, persist index cache (Re: Feedback from @Uri-Tauber)
- Pressing **Back** from a definition now returns to word selection (lookup mode) so you can look up another word immediately. Pressing **Confirm** ("Done") exits back to the reader.
- `DictionaryWordSelectActivity` now extends `ActivityWithSubactivity` and manages the definition screen as its own subactivity — the `onLookup` callback to `EpubReaderActivity` is removed.
- The sparse index offset table is cached to `/dictionary.cache` on the SD card after first build. On subsequent lookups (including after wake), the cache is loaded instead of rescanning the full 1.8MB `.idx` file. The cache is invalidated automatically if the `.idx` file size changes.

### Stem fallback and similar word suggestions on lookup miss
When an exact match isn't found:
1. **Stemming** — tries suffix removal (plurals, past tense, progressive, comparative, adverb, derivational suffixes like -ness/-ment/-ful/-less) and prefix removal (un-, dis-, re-). First stem that exists in the dictionary auto-opens its definition.
2. **Similar words** — if no stem matches, scans ~1500 alphabetically nearby entries, computes Levenshtein edit distance (with early termination), and presents up to 6 closest matches in a "Did you mean?" selection screen. Back from definition returns to the suggestions list, back again returns to word selection.

### Split words on en-dash and em-dash
Words joined by en-dashes (–) or em-dashes (—) are now split into separate selectable words during extraction, with correct screen positions computed from prefix text width.

kurtisgrant:
During some testing I found:

- Text is obscured by button hints for definitions in inverted orientation.
- Text obscured by button hints in all orientations except portrait for "Did you mean?" screen.
- Inconsistent button labels/names compared to rest of UI. (Will use "up" and "down" in Lookup History and Did you mean?, and "prev" and "next" in the definition reader.)
- Selecting a hyphenated word wrapped to the next page will try to define only the portion of the word on this page, rather than getting the other half from the following page.

I will commit a fix later today. Otherwise it seems quite nice to use.

odessa2:
Are you aware the developers explicitly list such features as „out of scope“ [in this document](https://github.com/crosspoint-reader/crosspoint-reader/blob/master/SCOPE.md)? I think people should read the scope document before making pull requests.

kurtisgrant:
@odessa2 You make a good point lol. I should have checked, but the fact that I assumed it should be in scope some might say could serve as a protest to that decision. This is a community project after all. (I do apologize for the clutter if this appeal is rejected) 

I see this as a pretty important feature, I'd be interested to hear some additional opinions. 

Do you think it should be in scope if this implementation isn't found to cause any performance issues and the like? How do you find the UX?

Uri-Tauber:
While @kurtisgrant should have open a discussion before opening this PR, I think this specific feature deserves a re‑evaluation. 
@kurtisgrant, could you refactor the code so that it changes as little as possible in existing files and keeps all new logic in dedicated files? That would isolate the dictionary feature and keep the core codebase clean.

@daveallie, what are your thoughts?

src/activities/reader/DictionaryDefinitionActivity.cpp
```diff
@@ -0,0 +1,179 @@
...
+
+  const bool landscape = orientation == CrossPointSettings::ORIENTATION::LANDSCAPE_CW ||
+                         orientation == CrossPointSettings::ORIENTATION::LANDSCAPE_CCW;
+  const int sidePadding = landscape ? 50 : 20;
+  constexpr int titleY = 10;
+  const int lineHeight = renderer.getLineHeight(readerFontId);
+  constexpr int bodyStartY = 50;
+
+  // Title: the word in bold (use UI font for title)
+  renderer.drawText(UI_12_FONT_ID, sidePadding, titleY, headword.c_str(), true, EpdFontFamily::BOLD);
```
CaptainFrito: Any chance you could use the new GUI functions from UITheme please? That way this new screen will automatically support classic/lyra themes. You can use drawHeader for this title text.

src/activities/reader/DictionarySuggestionsActivity.cpp
```diff
@@ -0,0 +1,147 @@
...
+void DictionarySuggestionsActivity::renderScreen() {
+  renderer.clearScreen();
+
+  constexpr int sidePadding = 20;
+  constexpr int titleY = 15;
+  constexpr int subtitleY = 45;
+  constexpr int startY = 80;
+  constexpr int lineHeight = 30;
+
+  // Title
```
CaptainFrito: Same here, you can use GUI.drawHeader and GUI.drawList

osteotek:
@odessa2 @kurtisgrant I personally think it is good feature to have, if it doesn't degrade performance too much

CaptainFrito:
Amazing work @kurtisgrant, this is truly the kind of feature that will set Crosspoint apart! In the future it could even be possible to add a button in the settings to download the dictionary from Github, so that the feature can be enabled easily.

kurtisgrant:
Thank you @Uri-Tauber and @CaptainFrito for your recommendations, I will make these changes today. :)

Uri-Tauber:
@kurtisgrant, if I understand correctly, the `dictionary.cache` is created only once (unless the user replace the dictionary files).
In that case, it would make more sense to move the cache‑creation logic out of the firmware and into a Python script that runs on the PC — meaning four files per dictionary instead of three, while easing the load on the ESP32’s limited resources.

kurtisgrant:
> @kurtisgrant, if I understand correctly, the `dictionary.cache` is created only once (unless the user replace the dictionary files). In that case, it would make more sense to move the cache‑creation logic out of the firmware and into a Python script that runs on the PC — meaning four files per dictionary instead of three, while easing the load on the ESP32’s limited resources.

@Uri-Tauber I have to admit I'm largely vibecoding this with claude code due to my desire for this feature. My experience is with web development so parsing through the cpp and the performance restrictions is slightly out of my wheelhouse. That being said, I talked with claude about this and it confirmed my hunch that this would not meaningfully decrease any computational load and creates a lot of additional friction, overhead and documentation requirements.

<details>
  <summary>Read Claude's rationale here</summary>

  My take: I'd push back on the suggestion. A few reasons:

  1. User experience degrades significantly. Right now: drop 3 standard StarDict 3 files on the SD card, done. The firmware handles everything transparently. With a Python script, the user needs Python installed, needs to know to run the script, needs to run it again if they swap dictionaries. That's a meaningful friction increase for a device that's supposed to be simple.
  2. The cost is genuinely low. It's a sequential file read, not a computation problem. The ESP32-C3's bottleneck here is SD card read speed over SPI, and it's a one-time cost with a progress indicator already wired in. The resulting cache is under 1KB.
  3. Self-healing on dictionary swap. The current code validates the cache against the actual .idx file size (lines 79–82). If a user replaces their dictionary files, the cache auto-rebuilds. A PC-generated cache introduces a silent failure mode — mismatched cache and dictionary files — that the user would have no idea how to diagnose.
  4. Firmware complexity doesn't actually decrease much. You'd still need loadCachedIndex() and the validation logic. You'd only drop loadIndex() and saveCachedIndex() (~100 lines), but you'd add a Python script plus documentation explaining when/how to run it.
  5. Four files instead of three increases error surface. "Which files do I need?" is already a question for users dealing with StarDict format. Adding a fourth device-specific file makes the answer harder.

  The suggestion makes sense in the abstract — offload work to a more capable machine — but the work being offloaded is trivial, and the tradeoff in UX and robustness isn't worth it. If the index scan were doing something expensive (building a trie, computing hashes for a bloom filter, etc.), the calculus would be different.
</details>

kurtisgrant:
As mentioned by @odessa2 earlier, this PR is out of scope as it stands. I apologize for not having checked that before creating it. However, I would've been making this feature for myself regardless and if we can verify that it is not too taxing on the system, I think it is a wonderful UX and I'd love to get some feedback from @Uri-Tauber @pepastach @odessa2 @osteotek @CaptainFrito @daveallie and others. 

Thanks to those who provided feedback thus far, I've implemented most of it I believe. 

Feel free to use this english dictionary file set, added to the root of your SD card for testing. 
[eng-dictionary.zip](https://github.com/user-attachments/files/25310741/eng-dictionary.zip)

As @CaptainFrito mentioned, if this were merged, we might look at providing a download feature from the firmware in the future.

drbourbon:
Started discussion for reconsidering this feature as **"in scope"**:  #878

jpirnay:
> As @CaptainFrito mentioned, if this were merged, we might look at providing a download feature from the firmware in the future.

I had a similar demand for the html pages of the internal Webserver, so i added a check for specific files on the SD - card that will automatically download the files from github if they are missing (PR #881)

src/activities/network/CrossPointWebServerActivity.cpp
```diff
@@ -347,9 +347,6 @@ void CrossPointWebServerActivity::loop() {
         // Yield and check for exit button every 64 iterations
         if ((i & 0x3F) == 0x3F) {
           yield();
-          // Force trigger an update of which buttons are being pressed so be have accurate state
-          // for back button checking
-          mappedInput.update();
```
daveallie, line 352: Not sure if you have another fix for this, but without this change, back button from the server activity are pretty unreliable.

iandchasse:
I’m wondering if the UI for this feature (highlighting text on a page) could be merged with the ongoing discussion to implement bookmarking/clippings. Highlighting a single word, then having an option to place a bookmark/note or get the definition, or you could maybe highlight the start/end of a passage to better document specific text within a bookmark/note. @tnaftali what do you think?

CaptainFrito:
> I’m wondering if the UI for this feature (highlighting text on a page) could be merged with the ongoing discussion to implement bookmarking/clippings. Highlighting a single word, then having an option to place a bookmark/note or get the definition, or you could maybe highlight the start/end of a passage to better document specific text within a bookmark/note. @tnaftali what do you think?

This is a great idea!

src/activities/reader/DictionaryDefinitionActivity.h
```diff
@@ -0,0 +1,56 @@
...
+
+  // Orientation-aware layout gutters (computed in wrapText, used in renderScreen)
+  int leftPadding = 20;
+  int rightPadding = 20;
+  int hintGutterHeight = 0;
+  int contentX = 0;
+  int hintGutterWidth = 0;
+
+  TaskHandle_t displayTaskHandle = nullptr;
+  SemaphoreHandle_t renderingMutex = nullptr;
```
daveallie, line 50: Let's swap over to the new render pattern where we can drop a lot of this boilerplate.

src/main.cpp
```diff
@@ -426,13 +426,6 @@ void loop() {
...
     yield();  // Give FreeRTOS a chance to run tasks, but return immediately
   } else {
-    static constexpr unsigned long IDLE_POWER_SAVING_MS = 3000;  // 3 seconds
-    if (millis() - lastActivityTime >= IDLE_POWER_SAVING_MS) {
-      // If we've been inactive for a while, increase the delay to save power
-      delay(50);
-    } else {
-      // Short delay to prevent tight loop while still being responsive
-      delay(10);
-    }
```
daveallie, line 436: I assume you didn't mean to revert this change?

daveallie (changes requested):

daveallie:
@coderabbitai review

Anseryde:
Would it be possible to add a phrase lookup mode, or way to select multiple words as one word to look up a definition for? 

This would come in handy for CJK languages like Chinese, where words can take up multiple characters which here might register as individual words, in the future.

divinitycove:
Quite excited for this feature! A thought I had while reading over the discussion was that it might be better to read the dictionary files from a ".dict" folder (or something like that) to avoid root directory clutter, like how #948 handles the ".sleep" folder.

dragopieguy-bit:
Can Anyone help me add this feature to my Xteink to test it? I would really like to but I am new to GitHub and I don't know how to do so, I tried downloading the bin file and flashing it but I didn't get the feature, any help will be greatly appreciated as I really want to try this feature

Uri-Tauber:
@dragopieguy-bit You need to put the [dictionary files](https://github.com/user-attachments/files/25271916/eng-dictionary.zip) at the root of your SD card, and flash this firmware:
[firmware.bin.zip](https://github.com/user-attachments/files/26078351/firmware.bin.zip)

dragopieguy-bit:
@Uri-Tauber I am sorry but what order should I do it in? And how exactly should I put the dictionary files? I am sorry if my questions seem stupid but I literally created this account like 6 hours ago and I really want to try this feature. Thank you

Uri-Tauber:
@dragopieguy-bit No problem at all! I remember the first time I opened an issue on GitHub—I was struggling with installing some script, so I totally get it 🙂

Regarding your question: in my previous message, I attached two ZIP files. Please download and extract both of them. In one of the extracted folders, you’ll find three files starting with “directory.” Connect your SD card to your PC and copy those files into the main folder of the SD card (i.e., the root directory).

Next, connect the X4 reader to your PC using a USB cable and open the [flash tool](https://xteink.dve.al) (which I assume you’ve already used). By the way, did you save a full flash backup? If not, it’s definitely a good idea to do that first.

Previously, you likely flashed Crosspoint by pressing “Flash Crosspoint firmware.” This time, instead click “Select file,” choose the `.bin` file from the second ZIP I shared, and then click “Flash firmware from file.”

I hope this helps! Feel free to ask if anything isn’t clear or if you run into any issues.

P.S. For next time, it’s better to open an issue instead of commenting on a PR—it keeps things more organized.

dragopieguy-bit:
@Uri-Tauber thank you so much for this help, I really appreciate it. 

>P.S. For next time, it’s better to open an issue instead of commenting on a PR—it keeps things more organized.

Noted, thank you.its just that I am new and I don't even know what everything does.

Anseryde:
Hello, just wanted to add that there is a promising draft PR #1402 made by @mcrosson that appears to be a more developed version of this PR with added features like a dedicated 'dictionary' subfolder for dictionary files and per-book dictionary setting, but it is still a draft currently. Seems to have less eyes on it compared to this version of the PR

mcrosson:
> Hello, just wanted to add that there is a promising draft PR #1402 made by @mcrosson that appears to be a more developed version of this PR with added features like a dedicated 'dictionary' subfolder for dictionary files and per-book dictionary setting, but it is still a draft currently. Seems to have less eyes on it compared to this version of the PR

i do not recommend merging my code from pr #1402 . I'm using it on my device to ensure it works but I've not run a cleanup/verification pass against project norms or ram use, etc. This pr (857) works well as-is and trying to merge items from 1402 would likely be more problematic than it's worth. Given I've not run any cleanup, verification, etc yet there is a real risk of duplicated effort which I don't think anyone wants.

Anseryde:
okay

edit: Would PR #1402 or a future version of it theoretically replace the features added in this PR or add onto them if approved?

mcrosson:
> okay
> 
> edit: Would PR #1402 or a future version of it theoretically replace the features added in this PR or add onto them if approved?

yes, the only missing item is the lookup history and that is on my plan to add to 1402.

DustinHu:
(note: also left a similar comment [here](https://github.com/crosspoint-reader/crosspoint-reader/pull/1402#issuecomment-4167461472))

Hello! Apologies, I should have checked the open PRs a bit more closely; I've been working on [my own version of a Dictionary word-lookup feature](https://github.com/crosspoint-reader/crosspoint-reader/pull/1542/changes#diff-270ede15a59670852ad6e3ef0797a15445e897990f5c5747a6c6c94b2e033074R227-R373) without realizing that you already had a PR in progress. 

I really like your implementation! It's got a lot of cool features (search history, fuzzy searches, etc. are all great ideas that I hadn't thought of). Using StarDict in particular is something that I'd totally overlooked (I just had a half-baked, homebrewed custom dictionary format instead), and I've since updated my PR to use StarDict as well just so I could do some comparisons. 

That being said, because I had kind of brashly taken a completely separate approach, I've landed on a few different implementations that I think could be helpful to merge with your changes (definitely not trying to step on your toes here though). 

1. **Lookup speed optimizations** - Maybe it has to do with the dictionaries that I've downloaded for testing (I've been testing using the Websters English Dictionary Primarily, along with the GNU Collaborative International Dictionary of English) so let me know if your mileage has varied, but I've observed a good amount of delay between selecting a word and having the definition appear on my screen (around 1 - 1.5 seconds). In my implementation, I'd leveraged some architectural differences to get read times around ~275 - 325 ms, which would be a ~66% decrease in read speed if we could port the changes over. 
I hadn't tried importing any of it over yet, but I think the main improvements would come from:
    1. **Byte-by-byte SD reads in readWordInto()**: Each file.read() call has SPI bus overhead, so reading a 10-character word makes 10 separate SD card round-trips. A simple fix is wrapping it in a buffered reader that reads 512 bytes at a time and serves characters from memory, which allows us to minimize reads from the SD card. See: https://github.com/crosspoint-reader/crosspoint-reader/pull/1542/changes#diff-270ede15a59670852ad6e3ef0797a15445e897990f5c5747a6c6c94b2e033074R19-R61
    2. **Linear scan after sparse cache lookup**: The current lookup narrows down to a batch of words, then checks each one individually until it finds a match. Instead, in my approach I've been generating a secondary index file (.idx.cp) where every entry is a static 44 bytes. Because entries are the same size, you can calculate exactly where any entry is in the file by its index, so you can jump around and do a binary search more cleanly/efficiently without drilling into the actual words. See: https://github.com/crosspoint-reader/crosspoint-reader/pull/1542/changes#diff-270ede15a59670852ad6e3ef0797a15445e897990f5c5747a6c6c94b2e033074R227-R373
2. **Support for multiple installed dictionaries**: Rather than having the user select 1 active dictionary, I was leaning towards allowing the user to enable as many dictionaries as they wanted. The results would just show whichever dictionary's entry was found. When multiple entries were found for 1 word, they could scroll left/right to see multiple definitions. I'd personally appreciate this feature when reading books that frequently dip into a second language, e.g. if a mostly English book casually dropped French word it'd be nice to have a French -> English dictionary also installed so I could see a quick translation in addition to my usual English word lookups. 
See multi-dictionary search logic: https://github.com/crosspoint-reader/crosspoint-reader/pull/1542/changes#diff-9a1eadf3148e5a465f83a1fe9ece9309c7a00a2d93308cb97de21cf583ec33a2R318-R351
See displaying multiple results: https://github.com/crosspoint-reader/crosspoint-reader/pull/1542/changes#diff-94805a199a2123d5960663a35307148ec58e67ef987b1ad7c8568b4ef82d21fdR177-R178
3. **Hold-to-search**, rather than having to go into the menus: In my changes, I'd included the functionality to hold the "Confirm" button for 1 second to activate the "highlight word to define" mode, rather than having the user go into the menu and scroll down. To me this seems like a slightly more streamlined user experience (just 1 button press/screen refresh instead of ~3), though I could understand if we don't want to add top-level functionality to such an important button.
See: https://github.com/crosspoint-reader/crosspoint-reader/pull/1542/changes#diff-82798dedbe135495e619d4aa27a4bef560c70c7663cf43148b67a26ddde45682R242-R252
4. **Pre-generating the cache**: If we're supplying the user with the dictionary files already, and there cache that gets generated is deterministic/static, is there any reason we can't just include the cache in the .zip file, rather than having the device generate it the first time we look up a word?

mcrosson:
Given the lack of work on this PR and #1402 generally superseding it ; do any @crosspoint-reader/firmware-maintainers want to keep this open or can it be closed as a dupe?

itsthisjustin:
> #1402

I'm in favor of doing anything that cleans up stale PRs so you have my vote

Uri-Tauber:
I agree it should be closed.
Sorry, @kurtisgrant — but you can take comfort in knowing that #1402 was built on top of your work, so you absolutely deserve credit for starting this great feature.

mcrosson:
Credit has been given in the #1599 discussion as well.

PR #859, opened 2026-02-12
feat: Finetune addition of fonts

jpirnay:
## Summary

* Our flash memory is nearly fully consumed (96%) - this patch allows for individual compilation and inclusion of the three main font families.

## Additional Context

The implementation allows you to selectively disable font families to save flash memory by disabling the inclusion of font families in the platformio.ini file:

Current Configuration (All Enabled)
In ``platformio.ini`` all three reader fonts are currently enabled:
```
# Font family controls (comment out to disable and save flash)
# WARNING: At least ONE reader font must be enabled (Bookerly, NotoSans, or OpenDyslexic)
# Note: Ubuntu UI fonts are always included (used for menus and UI elements)
  -DENABLE_FONT_BOOKERLY=1
  -DENABLE_FONT_NOTOSANS=1
  -DENABLE_FONT_OPENDYSLEXIC=1
```

Memory Savings Available

- Disable OpenDyslexic only: Save ~1.5 MB (drop to 70-75% flash)
- Keep only Bookerly: Save ~1.8 MB (drop to 68-73% flash)
- Keep Bookerly + NotoSans: Save ~1.5 MB (drop to 70-75% flash)
- **NB:  At least one reader font is required for the device to function**

What's Protected

- Ubuntu fonts (UI elements): Always included
- NotoSans 8pt (small UI text): Always included
---

### AI Usage

Did you use AI tools to help write this code? Partially, I did let Claude analyze the biggest consumers of flash memory

Claudes flash analysis:
```
Your ESP32 flash is at 96.2% capacity (6.02 MB / 6.25 MB). The biggest culprit is fonts - they're consuming an estimated 3-4 MB of flash (60-65% of total usage)!

-  Critical Finding: FONTS
You're including 4 complete font families with multiple sizes and styles:

Bookerly (6 sizes × 4 styles = 24 variants)
NotoSans (6 sizes × 4 styles = 24 variants)
OpenDyslexic (6 sizes × 4 styles = 24 variants)
Ubuntu (partial set)
Each font variant has both glyph data (~12 KB) and bitmap data (50-200 KB), adding up to 3-4 MB total.

- Other Major Consumers
Language Dictionaries: ~280 KB

German (de_trie_data): 206 KB
Russian: 33 KB
English: 27 KB
Spanish: 13 KB
ESP-IDF Libraries: ~1.5-2 MB (WiFi, crypto, network stack)

HTML Assets: 41 KB (FilesPageHtml)
```

### DIFF
--- a/lib/Epub/Epub.cpp
+++ b/lib/Epub/Epub.cpp
@@ -2,8 +2,8 @@
 
 #include <FsHelpers.h>
 #include <HalStorage.h>
-#include <HardwareSerial.h>
 #include <JpegToBmpConverter.h>
+#include <Logging.h>
 #include <ZipFile.h>
 
 #include "Epub/parsers/ContainerParser.h"
@@ -17,7 +17,7 @@ bool Epub::findContentOpfFile(std::string* contentOpfFile) const {
 
   // Get file size without loading it all into heap
   if (!getItemSize(containerPath, &containerSize)) {
-    Serial.printf("[%lu] [EBP] Could not find or size META-INF/container.xml\n", millis());
+    LOG_ERR("EBP", "Could not find or size META-INF/container.xml");
     return false;
   }
 
@@ -29,13 +29,13 @@ bool Epub::findContentOpfFile(std::string* contentOpfFile) const {
 
   // Stream read (reusing your existing stream logic)
   if (!readItemContentsToStream(containerPath, containerParser, 512)) {
-    Serial.printf("[%lu] [EBP] Could not read META-INF/container.xml\n", millis());
+    LOG_ERR("EBP", "Could not read META-INF/container.xml");
     return false;
   }
 
   // Extract the result
   if (containerParser.fullPath.empty()) {
-    Serial.printf("[%lu] [EBP] Could not find valid rootfile in container.xml\n", millis());
+    LOG_ERR("EBP", "Could not find valid rootfile in container.xml");
     return false;
   }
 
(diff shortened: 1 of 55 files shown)


Uri-Tauber:
Thanks! this is really needed.

ngxson:
FWIW, we already had `-DOMIT_FONTS` that allows excluding all fonts except for Bookerly, it's mostly used for debugging though.

Personally I think further fine-tuning won't be beneficial as-is if the user can't have an easy way to build their own firmware. I assume large part of the user base doesn't know how to build it. [Many videos on youtube](https://www.youtube.com/results?search_query=crosspoint+xteink) already provided an easy entry point via the web flasher.

And even if we handle the build process, then now the question is how to distribute different versions of the firmware.

jpirnay:
A couple of comments:
* there are too many interesting ideas aka PRs flying around to integrate them all. So my idea is to make them optionally integratable and to allow the use to pick which options he would like to have - a bit like the grblhal web builder https://svn.io-engineering.com:8443/ - I will try to concoct a simple shell script later today to show the idea
* IMHO the real solution for this specific problem addressed in this PR is not to make fonts discardable but to change the way how we deal with fonts. I personally prefer the way how papyus is loading fonts on demand from sd.

jpirnay:
A (rather complete) mock of the functionality that I had in mind:
configurator

drbourbon:
> 
> * there are too many interesting ideas aka PRs flying around to integrate them all. So my idea is to make them optionally integratable and to allow the use to pick which options he would like to have - a bit like the grblhal web builder https://svn.io-engineering.com:8443/ - I will try to concoct a simple shell script later today to show the idea

Cherry picking features at compile time has been discussed here:
https://github.com/crosspoint-reader/crosspoint-reader/discussions/160#discussioncomment-15379015

The main point @daveallie brought is that this would potentially generate additional overhead with supporting this firmware (due to the profileration of builds).

> FWIW, we already had `-DOMIT_FONTS` that allows excluding all fonts except for Bookerly, it's mostly used for debugging though.

This also cuts out font sizes other than "Normal" from the firmware (just a matter of moving a couple of `#ifndef OMIT_FONTS` in `src/main.cpp` by the way)

pablohc:
Thanks @jpirnay In my particular case, I use a single font (Bookerly) and a single size (small), so having other sizes of the same font that are not going to be used may also be something to consider if there is a selector when flashing.

jpirnay:
Now with compressed fonts this is less of an issue

PR #861, opened 2026-02-12
feat: use pre-compressed HTML pages

ngxson:
## Summary

Pre-compress the HTML file to save flash space. I'm using `gzip` because it's supported everywhere (indeed, we are using the same optimization on [llama.cpp server](https://github.com/ggml-org/llama.cpp), our HTML page is huge 😅 ).

This free up ~40KB flash space.

Some users suggested using `brotli` which is known to further reduce 20% in size, but it doesn't supported by firefox (only supports if served via HTTPS), and some reverse proxy like nginx doesn't support it out of the box (unrelated in this context, but just mention for completeness)

```
PR:
RAM:   [===       ]  31.0% (used 101700 bytes from 327680 bytes)
Flash: [==========]  95.5% (used 6259244 bytes from 6553600 bytes)

master:
RAM:   [===       ]  31.0% (used 101700 bytes from 327680 bytes)
Flash: [==========]  96.2% (used 6302416 bytes from 6553600 bytes)
```

---

### AI Usage

Did you use AI tools to help write this code? **PARTIALLY**, only the python part

### DIFF
--- a/lib/Epub/Epub.cpp
+++ b/lib/Epub/Epub.cpp
@@ -2,8 +2,8 @@
 
 #include <FsHelpers.h>
 #include <HalStorage.h>
-#include <HardwareSerial.h>
 #include <JpegToBmpConverter.h>
+#include <Logging.h>
 #include <ZipFile.h>
 
 #include "Epub/parsers/ContainerParser.h"
@@ -17,7 +17,7 @@ bool Epub::findContentOpfFile(std::string* contentOpfFile) const {
 
   // Get file size without loading it all into heap
   if (!getItemSize(containerPath, &containerSize)) {
-    Serial.printf("[%lu] [EBP] Could not find or size META-INF/container.xml\n", millis());
+    LOG_ERR("EBP", "Could not find or size META-INF/container.xml");
     return false;
   }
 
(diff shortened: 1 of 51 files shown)


jpirnay:
I am just working on a different approach:
CrossPointWebserver does serve the html content from the sd-card, if the required files are not present, then it will fetch them from the github server. Currently testing and fixing...

ngxson:
@jpirnay That's way over complicated I think. If we already had access to the internet, just simply include all CSS and JS content to be fetched from internet from the browser:

```html
<link rel="stylesheet" href="https://crosspoint-reader.github.io/......">
<script src="https://crosspoint-reader.github.io/......" />
```

jpirnay:
> @jpirnay That's way over complicated I think. If we already had access to the internet, just simply include all CSS and JS content to be fetched from internet from the browser:
> 
> ```
> <link rel="stylesheet" href="https://crosspoint-reader.github.io/......">
> <script src="https://crosspoint-reader.github.io/......" />
> ```

Well, the idea behind it is not only to download html pages but all other non-essential data (like hyphenation data) from the web, too. And - it's working great :-)
Just for html:
```
RAM:   [===       ]  31.0% (used 101700 bytes from 327680 bytes)
Flash: [========= ]  94.7% (used 6204932 bytes from 6553600 bytes)
```

PR #869, opened 2026-02-13
fix: Close leaked file descriptors in SleepActivity and web server

brbla:
## Summary

- **SleepActivity.cpp**: Add missing `file.close()` calls in 3 code paths that open BMP files for sleep screen rendering but never close them before returning. Affects random custom sleep images, the `/sleep.bmp` fallback, and book cover sleep screens.
- **CrossPointWebServer.cpp**: Add missing `dir.close()` in the delete handler when `Storage.open()` returns a valid `FsFile` that is not a directory.

## Context

SdFat is configured with `DESTRUCTOR_CLOSES_FILE=0`, which means `FsFile` objects are **not** automatically closed when they go out of scope. Every opened file must be explicitly closed.

The SleepActivity leaks are particularly impactful because they occur on every sleep cycle. While ESP32 deep sleep clears RAM on wake, these leaks can still affect the current session if sleep screen rendering is triggered multiple times (e.g., cover preview, or if deep sleep fails to engage).

The web server leak in `handleDelete()` is a minor edge case (directory path that opens successfully but `isDirectory()` returns false), but it's still worth fixing for correctness.

## Test plan

- [x] Verify sleep screen still renders correctly (custom BMP, fallback, cover modes)
- [x] Verify folder deletion still works via the web UI
- [ ] Monitor free heap before/after sleep screen rendering to confirm no leak


daveallie:
Have you verified that folder deletion still works? I see it's unticked in your PR body?

brbla:
I tested it again just to be sure. I can create and delete a folder on the card via the web UI without any problems. Only empty folders can be deleted, if there is a file or subfolder in it, the message appears: Failed to delete: Folder is not empty. Delete contents first.

The third point is related to my mistake, which is related to my attempt at localization into Czech, and therefore will not be relevant.

ngxson:
I would be interested in doing researches about enabling `DESTRUCTOR_CLOSES_FILE` instead. Would appreciate if there are any insights on side effects if that's being enabled.

daveallie:
> I would be interested in doing researches about enabling DESTRUCTOR_CLOSES_FILE instead. Would appreciate if there are any insights on side effects if that's being enabled.

Looks to use about 5K of flash.

```
RAM:   [===       ]  31.5% (used 103100 bytes from 327680 bytes)
Flash: [=======   ]  68.9% (used 4513220 bytes from 6553600 bytes)
```

```
RAM:   [===       ]  31.5% (used 103100 bytes from 327680 bytes)
Flash: [=======   ]  68.9% (used 4518498 bytes from 6553600 bytes)
```

Functionally, we never actually intend to destruct a file without closing it, and indeed everywhere we open files we try to close them (this PR pointing out the few bad cases). I think we should move forward with enabling the flag, will open a new PR to do it in and merge this regardless.

PR #871, opened 2026-02-13
fix: Avoid stack overflow in getSettingsList() by using push_back

brbla:
## Summary

- Replace initializer_list construction in `getSettingsList()` with individual `push_back` calls and `reserve(30)`
- Fixes a **stack overflow crash** (`***ERROR*** A stack overflow in task loopTask has been detected`) when entering Settings

## Root cause

`getSettingsList()` returns a `std::vector<SettingInfo>` built via an initializer_list `return { item1, item2, ... }`. The compiler places all ~30 `SettingInfo` temporaries on the stack simultaneously before copying them into the vector.

Each `SettingInfo` is ~200 bytes (4× `std::function`, `std::vector<std::string>`, pointers), so the initializer_list alone consumes **~6 KB of stack**. The default `loopTask` stack on ESP32 is **8 KB**. Combined with the call chain depth (`loop()` → input handling → activity transition → `SettingsActivity::onEnter()` → `getSettingsList()`), this overflows the stack.

## How the fix works

With `push_back`, only **one** `SettingInfo` temporary (~200 bytes) exists on the stack at a time. It gets move-constructed into the heap-allocated vector, then the temporary is destroyed before the next one is created.

`reserve(30)` pre-allocates heap memory in a single call, so `push_back` never triggers reallocation.

## Impact

- **Memory**: Zero change — same total heap usage, just ~6 KB less peak stack usage
- **Performance**: Negligible — the function runs only when entering Settings or on web API calls
- **Behavior**: Identical output — same vector contents in the same order

## Test plan

- [x] Verified on hardware (ESP32-C3): Settings opens without crash
- [x] Tested repeatedly: entering/exiting Settings, sleep/wake cycles, web UI settings API
- [x] Build passes CI (clang-format, cppcheck)

### DIFF
--- a/src/SettingsList.h
+++ b/src/SettingsList.h
@@ -9,93 +9,105 @@
 // Shared settings list used by both the device settings UI and the web settings API.
 // Each entry has a key (for JSON API) and category (for grouping).
 // ACTION-type entries and entries without a key are device-only.
+//
+// Uses push_back instead of initializer_list to avoid placing all SettingInfo
+// temporaries (~200 bytes each) on the stack simultaneously. The initializer_list
+// pattern caused a stack overflow in the 8KB loopTask on ESP32.
 inline std::vector<SettingInfo> getSettingsList() {
-  return {
-      // --- Display ---
-      SettingInfo::Enum("Sleep Screen", &CrossPointSettings::sleepScreen,
-                        {"Dark", "Light", "Custom", "Cover", "None", "Cover + Custom"}, "sleepScreen", "Display"),
-      SettingInfo::Enum("Sleep Screen Cover Mode", &CrossPointSettings::sleepScreenCoverMode, {"Fit", "Crop"},
-                        "sleepScreenCoverMode", "Display"),
-      SettingInfo::Enum("Sleep Screen Cover Filter", &CrossPointSettings::sleepScreenCoverFilter,
-                        {"None", "Contrast", "Inverted"}, "sleepScreenCoverFilter", "Display"),
-      SettingInfo::Enum(
-          "Status Bar", &CrossPointSettings::statusBar,
-          {"None", "No Progress", "Full w/ Percentage", "Full w/ Book Bar", "Book Bar Only", "Full w/ Chapter Bar"},
-          "statusBar", "Display"),
-      SettingInfo::Enum("Hide Battery %", &CrossPointSettings::hideBatteryPercentage, {"Never", "In Reader", "Always"},
-                        "hideBatteryPercentage", "Display"),
-      SettingInfo::Enum("Refresh Frequency", &CrossPointSettings::refreshFrequency,
-                        {"1 page", "5 pages", "10 pages", "15 pages", "30 pages"}, "refreshFrequency", "Display"),
-      SettingInfo::Enum("UI Theme", &CrossPointSettings::uiTheme, {"Classic", "Lyra"}, "uiTheme", "Display"),
-      SettingInfo::Toggle("Sunlight Fading Fix", &CrossPointSettings::fadingFix, "fadingFix", "Display"),
+  std::vector<SettingInfo> s;
 
-      // --- Reader ---
-      SettingInfo::Enum("Font Family", &CrossPointSettings::fontFamily, {"Bookerly", "Noto Sans", "Open Dyslexic"},
-                        "fontFamily", "Reader"),
-      SettingInfo::Enum("Font Size", &CrossPointSettings::fontSize, {"Small", "Medium", "Large", "X Large"}, "fontSize",
-                        "Reader"),
-      SettingInfo::Enum("Line Spacing", &CrossPointSettings::lineSpacing, {"Tight", "Normal", "Wide"}, "lineSpacing",
-                        "Reader"),
-      SettingInfo::Value("Screen Margin", &CrossPointSettings::screenMargin, {5, 40, 5}, "screenMargin", "Reader"),
-      SettingInfo::Enum("Paragraph Alignment", &CrossPointSettings::paragraphAlignment,
-                        {"Justify", "Left", "Center", "Right", "Book's Style"}, "paragraphAlignment", "Reader"),
-      SettingInfo::Toggle("Book's Embedded Style", &CrossPointSettings::embeddedStyle, "embeddedStyle", "Reader"),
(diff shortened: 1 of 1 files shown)


src/SettingsList.h
```diff
@@ -9,93 +9,106 @@
...
-          {"None", "No Progress", "Full w/ Percentage", "Full w/ Book Bar", "Book Bar Only", "Full w/ Chapter Bar"},
-          "statusBar", "Display"),
-      SettingInfo::Enum("Hide Battery %", &CrossPointSettings::hideBatteryPercentage, {"Never", "In Reader", "Always"},
-                        "hideBatteryPercentage", "Display"),
-      SettingInfo::Enum("Refresh Frequency", &CrossPointSettings::refreshFrequency,
-                        {"1 page", "5 pages", "10 pages", "15 pages", "30 pages"}, "refreshFrequency", "Display"),
-      SettingInfo::Enum("UI Theme", &CrossPointSettings::uiTheme, {"Classic", "Lyra"}, "uiTheme", "Display"),
-      SettingInfo::Toggle("Sunlight Fading Fix", &CrossPointSettings::fadingFix, "fadingFix", "Display"),
+  std::vector<SettingInfo> s;
+  s.reserve(30);
```
ngxson: this is quite fragile IMO, there is nothing preventing contributors from adding some new settings and forget to update this number. instead, there should be an `assert()` to avoid passing the number of pre-allocated items.

brbla: Thanks for the feedback! You're right that a hardcoded `reserve(30)` is fragile.

I've updated the PR to **remove `reserve()` entirely**. The vector now grows organically via `push_back()` without any pre-allocation hint. This eliminates the magic number problem completely — contributors can add new settings freely without updating anything.

The performance cost is negligible: for ~29 items, the vector does ~6 internal reallocations on the heap (1→2→4→8→16→32), each taking microseconds. The vector is short-lived (built, consumed, destroyed), so there's no lasting fragmentation.

Regarding `assert()` — on an embedded device (ESP32 with no debugger attached), a failed assertion triggers a hard crash and reboot. That would make the device worse off than a few extra heap reallocations. The real fix here is the switch from `initializer_list` to `push_back`, which prevents all ~29 temporaries (~200 bytes each, ~6KB total) from being placed on the 8KB stack simultaneously.

Does this version look good to you?

ngxson: That's way worse. Heap will be fragmented on realloc.

ngxson: > Regarding `assert()` — on an embedded device (ESP32 with no debugger attached), a failed assertion triggers a hard crash and reboot.

It's either hard reboot or always works here. If it's always crash, that signals to the developer that things need to be fixed.

Beside, I don't feel comfortable when you just copy the answer from AI and paste it here. If you continue to do that with your next message, I will refuse to review this PR.

brbla: Thanks for the feedback. When I tried the firmware modifications, I encountered reboots when entering the settings (I added one item to the menu for my needs). But these were not reboots that would always occur. After uploading the firmware, the reader worked properly for some time, and after a few hours/days, the reader restarted when trying to enter the settings.

I read the cause from the serial communication and when I was looking for a possible solution, I tried this loading modification. Since then, the reader has not restarted after several days of operation, including my other modifications, and entering the settings always works for me.

Since this behavior can occur when an item is added to the settings (and from my use, the device does not always reset, but only after some time of operation), I found it useful to offer a modification that would handle these resets.

Since I am not a native English speaker, I use a translator. But I always read what I write before sending it and try to check that what I wanted to say is there.

Furthermore, I sincerely apologize for adding an edit that would have made things worse in other cases that I didn't see coming. If that's the case, then my edit is simply stupid and should be rejected.

Even though I have played with Arduino and similar kits in the past, I definitely do not consider myself a core developer in this regard and if my contributions are not perceived as useful (which is quite possible and then thank you for knowing), then I will refrain from them.

I do not take it personally, I just do not want to waste your time and it is quite possible that what I consider interesting and useful is just my excess and for others it is an unnecessary and harmful burden. I am not angry with you and on the contrary I appreciate that you write directly as it is.

ngxson: My hard rule is to disengage if a contributor only reply to me with AI-generated answer, the reason is because I suppose to speak human-to-human for discussion, not human-to-AI. No hard feeling in anyways, just my hard rule.

And btw, I'm not native english speaker either, you can see in this message I already made a lot of grammar errors. That's what make me a human.

Re. the quality of this PR, I can see all commits are done by Claude and it makes me feels like having to go back and forth between me <--> you <--> Claude. To make it straight, I also have Claude subscription, so doing this way is negative productivity for me.

Regarding the bug, I don't see any [issues](https://github.com/crosspoint-reader/crosspoint-reader/issues) about the same bug raised by other users. Could you check carefully to know why this problem only raise in your specific case?

brbla: I'm honestly writing this message myself :-) I understand the rule and it makes sense to me.

I'll investigate it even more thoroughly. The behavior occurred to me when I tried to translate the UI into Czech where I had just added the Language item to Settings → System. (my PR #789)

I extracted the edit into a separate PR because I found it useful and solved my problem (the reader seemingly "randomly" resetting when entering the settings after expanding the settings items) and it occurred to me that this could occur in the future regardless of my expanding the settings.

I use Claude, so I'm learning more advanced work with Git and I don't want to make some schoolboy mistake in the process. At the same time, I think it's right to admit that I used AI, because it significantly reduced the time for the first functional prototype of the translated UI.

Thank you again for your time.

ngxson: As explain via https://github.com/crosspoint-reader/crosspoint-reader/pull/789 , your approach has fundamental problems with excessive memory usage. The root cause is not due to stack-allocated `initializer_list`, but due to the fact that your approach copy the same string multiple times during execution.

In other words, this fix doesn't resolve the root cause of why you run into memory issues. Indeed, it's the reversed: it proves that https://github.com/crosspoint-reader/crosspoint-reader/pull/789 uses a lot more RAM space.

So unfortunately, I cannot accept this fix as-is.

brbla: Thanks for your time. It makes sense and I just misjudged it and I still have a lot to learn.

daveallie:
As per threaded comments, this is only relevant on top of #789, which has been closed. So therefore, going to mark this one as closed as well.

PR #874, opened 2026-02-13
fix: URL-decode filenames in file browser and HTTP uploads

jtvargas:
## Summary

- Files uploaded via the web interface's HTTP multipart upload path are saved to the SD card with URL-encoded names (e.g., `My%20Book%20%E2%80%94%20Title.epub` instead of `My Book — Title.epub`), because the ESP32 WebServer library does not decode the `Content-Disposition` filename header
- This PR adds a `urlDecode()` utility to `StringUtils` and applies it in two places:
  1. **File browser display** — filenames are URL-decoded for display only (the raw filename is preserved for filesystem access)
  2. **HTTP upload** — filenames are URL-decoded at upload time so new uploads are saved with clean names on the SD card

## Changes

- **`src/util/StringUtils.h` / `.cpp`** — Add `StringUtils::urlDecode()` that decodes `%XX` hex sequences and `+` as space
- **`src/activities/home/MyLibraryActivity.cpp`** — URL-decode filenames in the `render()` display lambda (display-only, does not affect file access)
- **`src/network/CrossPointWebServer.cpp`** — URL-decode `upload.filename` in `handleUpload()` so new files are saved with clean names
- **`.gitignore`** — Add `.venv` (Python virtual environment used for debugging scripts)

## AI Usage

Did you use AI tools to help write this code? yes, claude
Previous:
[screenshot_without_decoding.bmp](https://github.com/user-attachments/files/25301016/screenshot_without_decoding.bmp)

After:
[screenshot_with_decoding.bmp](https://github.com/user-attachments/files/25301019/screenshot_with_decoding.bmp)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,5 +7,6 @@ lib/EpdFont/fontsrc
 .vs
 build
 **/__pycache__/
+.venv
 /compile_commands.json
 /.cache
--- a/src/activities/home/MyLibraryActivity.cpp
+++ b/src/activities/home/MyLibraryActivity.cpp
@@ -242,7 +242,7 @@ void MyLibraryActivity::render() const {
   } else {
     GUI.drawList(
         renderer, Rect{0, contentTop, pageWidth, contentHeight}, files.size(), selectorIndex,
-        [this](int index) { return files[index]; }, nullptr, nullptr, nullptr);
+        [this](int index) { return StringUtils::urlDecode(files[index]); }, nullptr, nullptr, nullptr);
   }
 
   // Help text
--- a/src/network/CrossPointWebServer.cpp
+++ b/src/network/CrossPointWebServer.cpp
@@ -542,7 +542,7 @@ void CrossPointWebServer::handleUpload(UploadState& state) const {
     // Reset watchdog - this is the critical 1% crash point
     esp_task_wdt_reset();
 
-    state.fileName = upload.filename;
+    state.fileName = StringUtils::urlDecode(std::string(upload.filename.c_str())).c_str();
     state.size = 0;
     state.success = false;
     state.error = "";
--- a/src/util/StringUtils.cpp
+++ b/src/util/StringUtils.cpp
@@ -61,4 +61,28 @@ bool checkFileExtension(const String& fileName, const char* extension) {
   return localFile.endsWith(localExtension);
 }
 
+std::string urlDecode(const std::string& encoded) {
+  std::string decoded;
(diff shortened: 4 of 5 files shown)


jtvargas:
can this be merged? 👀  @ngxson @osteotek @jdk2pq

osteotek (approved):

jdk2pq:
Hey @jtvargas, thanks for the contribution! I'm trying to understand this PR a little before merging. I know you built the CrossX and Send To X4 extensions, both of which I've used on my device to send articles. Is the issue that when a file named something like `My Book — Title.epub` gets sent over from CrossX/Send To X4, CrossPoint is altering the file's name to something like `My%20Book%20%E2%80%94%20Title.epub`?

It seems slightly odd to me to URL decode a file's name in the file browser, since that's the real name of the file. But it does make sense to save the file with a URL decoded file name upon upload. If there's some logic on the CrossPoint side that's mangling the file name after upload, it'd be helpful to know where that is to see if we can fix that up.

jdk2pq:
Ah, after looking into it a bit more, we should be able to just use `WebServer::urlDecode` for this instead of rolling our own ([source here](https://github.com/espressif/arduino-esp32/blob/master/libraries/WebServer/src/Parsing.cpp#L591-L614)). 

However, I just tried sending a few things from CrossX and the Send To X4 extension, and I'm not seeing any issues with a URL encoded filename. Is there a specific instance you could share where this happens?

jtvargas:
@jdk2pq Hey, thanks for taking a look at this. I created this PR maybe the second day I got my X4 device. I used a tool (before I made CrossX) to transfer some files, and the tool was the one that was sending the files encoded. So it was the third-party tool's fault, not X4. I created this PR to have the device do the decoding during the transfer/receiving process, instead of expecting a third-party tool to do the decoding and send the files correctly. The process will happen on the device.

But now I’m doing some testing with some new features I’ve been working on for CrossX. I was using my version of firmware (the one with these changes), and it’s adding extra work to the pipeline process that’s basically better if we don’t add this lol, so we can close this PR.

jtvargas:
@jdk2pq One thing I found in the crosspoint firmware is that, due to some limitations with the device's poor Wi-Fi signal, if the device stays on the sharing network screen (the one that shows the QR) for too long, it freezes and won't go back. It needs to be put to sleep in order to go home again. Maybe we can add a back action if the device is not receiving any requests after x amount of time.

PR #879, opened 2026-02-14
feat: Allow a local configuration file for custom compiles

jpirnay:
## Summary

* platformio.ini is a repository based config for platformio and cannot be modified without constant nagging of git to include it into you commits
* PlatformIO allows you to split your configuration into multiple files using the extra_configs option in the [platformio] block. This effectively merges other .ini files into your main one. This will be silently ignored if such a file does not exist

## Additional Context

* Modifiy platformio.ini and add a .gitignore entry to ignore your local config
* eg my own ``platformio.local.ini``:
```
[env:custom]
extends = base
build_flags =
  ${base.build_flags}
  -DCROSSPOINT_VERSION=\"${crosspoint.version}-custom\"
  ; inclusion of additional fonts is disabled in custom builds to save space
  -DOMIT_FONTS
```
---

### AI Usage


jdk2pq:
I'm fine with this if `my_platformio.ini` is the name we want to use, but the pattern I've seen in other projects is to use the `*.local*` ignore pattern in the `.gitignore`, so something like `platformio.local.ini` would get ignored.

osteotek (approved):

PR #881, opened 2026-02-14
feat: Fetch of non-critical data from github on demand

jpirnay:
## Summary

* One attempt to reduce flash consumption is the (one-time) on-demand load of optional components from the web. This branch does not include any html file in the code but loads them from sd-card. If they are not present then it will download them from the internet aka this repo to fetch and store them for future local reuse.

## Additional Context

* Use a /.crosspoint/data/web cache directory to serve the httpserver content from disk
* Add a system setting routine to clean the system cache
* Initial time to download all 3 webserver files: 3sec
* This concept can (and should) be extended to load hyphenation data, future localisation etc. from sd card.
* The on-demand load releieves a user from coping content to the sd-card
Compilation sizes:
```
RAM:   [===       ]  31.0% (used 101708 bytes from 327680 bytes)
Flash: [==========]  96.1% (used 6300294 bytes from 6553600 bytes)
```
PR (including UI to deal with system cache)
```
RAM:   [===       ]  31.0% (used 101708 bytes from 327680 bytes)
Flash: [==========]  95.4% (used 6251256 bytes from 6553600 bytes)```
---
= minus 49.038 bytes

### AI Usage

Did you use AI tools to help write this code? _Partially - framework by AI, plus a lot of manual tweaking_

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,4 +9,4 @@ build
 **/__pycache__/
 /compile_commands.json
 /.cache
-my_platformio.ini
+*.local*
--- a/platformio.ini
+++ b/platformio.ini
@@ -1,6 +1,6 @@
 [platformio]
 default_envs = default
-extra_configs = my_platformio.ini
+extra_configs = platformio.local.ini
 
 [crosspoint]
 version = 1.0.0


odessa2:
Does this mean an Internet connection is required before beeing able to use the web interface? If so, this would render Crosspoint useless for people who only want to use the wifi hotspot of the device. This is bad for general accessability, and makes the out of the box experience worse.

jpirnay:
> Does this mean an Internet connection is required before beeing able to use the web interface? If so, this would render Crosspoint useless for people who only want to use the wifi hotspot of the device. This is bad for general accessability, and makes the out of the box experience worse.

Thank you for the feedback. I completely understand the concern regarding 'out-of-the-box' accessibility and the desire for a fully offline experience. I’d like to provide some context on the technical trade-offs and the intended workflow that led to this design:

1. Minimal Footprint & Resource Management
The ESP32-C3 is extremely memory-constrained (~380KB usable RAM). By downloading certain web components on demand rather than bundling them into the firmware image, we significantly reduce the initial flash footprint. This leaves more space for the core reading engine and future firmware features that cannot be offloaded.

2. The Hotspot vs. Library Use Case
While the device supports an AP (Access Point) mode, the primary intent of the web interface is the bulk upload of books. Accessing the web UI via the device's own hotspot is a 'circular' workflow for most: you connect your PC/Phone to the device to upload files that usually reside on a device that already has internet access.

3. Intentional Setup Flow
The 'out-of-the-box' experience assumes that for a user to get books onto the device via WiFi, they will be operating within a local network environment at least once. By fetching these components during the first connection, we ensure the device stays lean and high-performance for the actual reading experience, which remains 100% offline.

Proposed Compromise:
I recognize that 'silent' dependencies can be frustrating. We could look into:
- Pre-caching: Clearly indicating during the first WiFi setup that 'Web Interface Components' are being downloaded and cached to the SD card for permanent offline use.
- Minimal Fallback: Including a 'barebones' HTML interface in the binary for emergency AP-mode use, while the 'High-Performance' UI remains a downloaded component."

jdk2pq:
@jpirnay Thank you for your contribution! Here's some of my thoughts:

- I do think it's beneficial to offload the non-critical data from CrossPoint to the SD card to reduce flash space, and as we build more features into CrossPoint, it's something we'll have to be keenly aware of, especially given that we're nearing the limit already.
- That being said, I also agree with @odessa2 that requiring a wifi connection for features that are already built-in and users have come to expect by default may be a jarring change and cause some frustration, even if it is a short amount of time required to download these files and it's a one-time process.
- Separately, yesterday, we merged in a PR that pre-compresses the HTML files with gzip, which reduced the flash space to 95.5% used. There are likely more gains to be had here if we also add CSS and JS minification.
- A _potential_ alternative here could be changing our installation method to an install script that runs the default flash and then separately sets up the HTML files (and other non-critical assets) on the SD card in our `.crosspoint` dir (or a different directory since that gets cleared with the "Clear Cache" option), so that the device itself isn't required to connect to wifi to download those files upon usage of the features that require external files.

For right now, I'm not sure if this is the path we want to go down for reducing flash consumption. However, I'm only one person and don't want to be making a decision on behalf of all maintainers. Definitely happy to hear others' perspectives and reconsider!

jpirnay:
@jdk2pq 
Fine, I don't mind archiving this one. But we *need* to cut down flash usage, and 48k is 48k...

I am blown away by the amount of contributions from the community, but as exciting as they are as dangerous they are too.

People will feel (rightly) disappointed if (just one example) the hyphenation data for their native language will not / can not be included. So either you are a declaring a feature freeze for good or a methodology change needs to happen that will relocate lesser required data to the sd-card.

Agreed an immediate installation of relevant data would be preferrable, but that would require a different installation method than ~just~ flashing the firmware.

I am already building my own image, as a couple of relevant fixes (like the fix for the empty cover on recent & sleep screen) haven't found their way into main (yet) and I can exclude parts that are non-relevant to me (like additional fonts or non-native hyphenation) and just consume space. But I want to give back to the community.

Well this has become a discussion item more than a PR, so closing this one...

jdk2pq:
@jpirnay I really do appreciate your community-focused mindset, and I am very happy to have you as someone who is actively contributing to this project and the community at large! So a big thank you for that, and I apologize that my previous reply came off as harsh or anything less appreciative!

I think you're bringing up very valid concerns that we as a community must all be focused on. This would be a great issue to add and get some additional input and insights on. Like I said, I'm just one person, and I definitely _do not_ want to be dictating the direction this project moves in (or doesn't move in) without gaining consensus from the rest of the community first.

In conclusion, thanks again, and I do hope you continue contributing here and making CrossPoint even better!

PR #891, opened 2026-02-15
fix: Improved hyphenation for unsupported languages

jpirnay:
## Summary

* EPUBs in unsupported languages had NO hyphenation. 
* This PR will allow the fallback hyphenation to be used for unsupported languages, where no Liang data is available (reasonable hyphenation at any character boundary, respects minimum prefix/suffix constraints (2 chars each side))

## Additional Context

* Minimal changes, just using existing functionality by loosening the fallback application criteria

### AI Usage

Did you use AI tools to help write this code? _PARTIALLY_ came up during analysis of hyphenation memory requirements


osteotek:
This will result for very ugly hyphenation for unsupported languages, fallback hyphenation was not designed for that. I am not sure this is direction we need to go. We just need to merge font size improvements soon and include all hyphenation dictionaries from https://github.com/typst/hypher. All 30 languages will take only 1.1 MB

jpirnay:
> This will result for very ugly hyphenation for unsupported languages, fallback hyphenation was not designed for that. I am not sure this is direction we need to go. We just need to merge font size improvements soon and include all hyphenation dictionaries from https://github.com/typst/hypher. All 30 languages will take only 1.1 MB

I know, but now you have nothing if the user asks for hyphenation - but the real hyphenation is definitely better. 
Closing

PR #895, opened 2026-02-15
perf: Optimize hyphenation with stack-allocated scores

jpirnay:
## Summary

* Use stack allocation for words ≤64 chars (99% of cases)

## Additional Context
* Reduces heap allocations from 4→3 per word
* Falls back to heap for exceptionally long words (>64 chars)
* Benchmark: 1.78x faster (stack vs heap), zero memory leaks
* Cost: +550 bytes Flash
* 
---

### AI Usage

Did you use AI tools to help write this code? _YES_ validated in some benchmarks and tests


osteotek:
Was that benchmark performed on device?

jpirnay:
Yes it was, i compared the english hyphenation for 4 letter, 8 letter, 12 letter and "supercalifragilisticexpialigetic" :-)

Issue #896, opened 2026-02-15
[Feature Request] Larger Font Size for Keyboard


ariel-lindemann:
As per the dicsussion

_Originally posted by @cogfilms-3D in https://github.com/crosspoint-reader/crosspoint-reader/discussions/855_

Although Crosspoint firmware, I do have some ideas for improving the user experience.

1. The size of the keyboard screen for entering data is far too small. It needs to be at least the width of the screen for clear data entry. 
![crosspoint-keyboard-compair](https://github.com/user-attachments/assets/42a52ce4-9c10-4981-bd45-e841dde34f5f)

ariel-lindemann:
Update: for **Lyra Theme** users, this has been solved with the merging of #732. The larger keyboard will be available for **Classic Theme** as well once #897 is merged

PR #897, opened 2026-02-15
feat: increase keyboard font size for classic theme

ariel-lindemann:
## Summary

Adresses Feature Request #896 

Changed key dimensions, initial positions and margins.

## Additional Context

The keyboard now looks like this:
image

---

### AI Usage


jpirnay:
Commendable effort, but I think @CaptainFrito has this already covered in his PR #732

ariel-lindemann:
Thanks for the heads up, the layout with the keys at the bottom looks very nice.

But after having tested it on my device, I've noticed that his inplementation only applies with the Lyra theme. For classic it still has the old layout.

So I still think my PR has merit as long as the Classic theme is still supported, which afaik it is

ariel-lindemann:
I've updated this to incorporate the changes fom upstream and provide the larger keyboard to classic theme users as well

PR #903, opened 2026-02-15
perf: Read table of contents directly from zip stream (no temp-file)

jpirnay:
### TOC Zip file handling
* The current logic does decompress the table of content into a temporary file for further processing
* This intermediary step can be omitted by directly streaming the zip output

### LUT Caching
* Current: Every page load reads LUT offset from disk
* PR: Cache it in Section object after first read

## Additional Context

* Direct zip straming:

| Metric | Baseline (temp file) | Optimized (streaming) | Improvement |
| -- | -- | -- | -- |
| Time | 855 ms | 799 ms | 6.5% faster |
| Throughput | 103.56 us/byte | 96.78 us/byte | 6.5% faster |
| Heap | +8708 bytes | +6652 bytes | 24% less |

* LUT caching: around 5% faster on cache hit: 14,4ms vs 15,1 ms, 5 byte overhead
---

### AI Usage

Did you use AI tools to help write this code? _YES_

### DIFF
--- a/lib/Epub/Epub.cpp
+++ b/lib/Epub/Epub.cpp
@@ -21,9 +21,7 @@ class ParserStreamAdapter : public Print {
  public:
   explicit ParserStreamAdapter(Parser& p) : parser(p) {}
 
-  size_t write(uint8_t byte) override {
-    return write(&byte, 1);
-  }
+  size_t write(uint8_t byte) override { return write(&byte, 1); }
 
   size_t write(const uint8_t* buffer, size_t size) override {
     if (error) return 0;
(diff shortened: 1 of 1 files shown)


daveallie:
The reason for the middle stage is to allow us to free the DEFLATE buffers which are used when extracting the file before having to parse it. At the time it was written (things may be different now), there were plenty of books where TOC parsing would crash out due to OOM. I'd encourage you to test it with books with large TOCs, for example the Luther Bible.

Also, I notice there are two distinct improvements in here, would be good to split them up.

jpirnay:
Further investigation showed that this does not provide benefits

osteotek:
> Further investigation showed that this does not provide benefits

What about LUT caching?

PR #911, opened 2026-02-16
feat: Add ability to wrap text to UI elements

lkristensen:
## Summary
New function "wrappedText" added to the GfxRenderer

Similarly to how truncatedText it will take a text string and a width and return a truncated string, this function will wrap the text into a vector of strings. A maximum number of lines is specified, and the rest of the text (if any) is truncated using truncatedText

The function only wraps on spaces, as rules for how words are split varies by language.

## Additional Context

  specific areas to focus on).

This was requested by CaptainFrito in [732](https://github.com/crosspoint-reader/crosspoint-reader/pull/732#issuecomment-3883078980).

As part of the lyra theme.
Before:
before

After:
after

---

### AI Usage

Did you use AI tools to help write this code? Used chatgpt to validate that std::vector<std::string> was an appropriate return type

### DIFF
--- a/lib/GfxRenderer/GfxRenderer.cpp
+++ b/lib/GfxRenderer/GfxRenderer.cpp
@@ -679,78 +679,58 @@ std::string GfxRenderer::truncatedText(const int fontId, const char* text, const
   return item.empty() ? ellipsis : item + ellipsis;
 }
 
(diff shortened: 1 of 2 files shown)


lkristensen:
Closed as it has been merged as part of [1074](https://github.com/crosspoint-reader/crosspoint-reader/pull/1074)

PR #920, opened 2026-02-16
feat: Migrate binary settings to json

jpirnay:
## Summary

* This PR introduces a migration from binary file storage to JSON-based storage for application settings, state, and various credential stores. This improves readability, maintainability, and allows for easier manual configuration editing.
* Benefits:
  - Settings files are now JSON and can be easily read/edited manually
  - Easier to inspect application state and settings during development
  - JSON structure is more flexible for future changes
* Drawback: around 15k of additional flash usage
* Compatibility: Seamless migration preserves existing user data

## Additional Context
1. New JSON I/O Infrastructure files:
     - JsonSettingsIO: Core JSON serialization/deserialization logic using ArduinoJson library
     - ObfuscationUtils: XOR-based password obfuscation for sensitive data
2. Migrated Components (now use JSON storage with automatic binary migration):
     - CrossPointSettings (settings.json): Main application settings
     - CrossPointState (state.json): Application state (open book, sleep mode, etc.)
     - WifiCredentialStore (wifi.json): WiFi network credentials (Password Obfuscation: Sensitive data like WiFi passwords, uses XOR encryption with fixed keys. Note: This is obfuscation, not cryptographic security - passwords can be recovered with the key)
     - KOReaderCredentialStore (koreader.json): KOReader sync credentials
     - RecentBooksStore (recent.json): Recently opened books list
3. Migration Logic
     - Forward Compatibility: New installations use JSON format
     - Backward Compatibility: Existing binary files are automatically migrated to JSON on first load
     - Backup Safety: Original binary files are renamed with .bak extension after successful migration
     - Fallback Handling: If JSON parsing fails, system falls back to binary loading
4. Infrastructure Updates
     - HalStorage: Added rename() method for backup operations

---

### AI Usage

Did you use AI tools to help write this code? _** YES**_


src/JsonSettingsIO.cpp
```diff
@@ -0,0 +1,276 @@
...
+  store.matchMethod = static_cast<DocumentMatchMethod>(method);
+
+  LOG_DBG("KRS", "Loaded KOReader credentials for user: %s", store.username.c_str());
+  return true;
+}
+
+// ---- WifiCredentialStore ----
+
+namespace {
+constexpr uint8_t WIFI_OBFUSCATION_KEY[] = {0x43, 0x72, 0x6F, 0x73, 0x73, 0x50, 0x6F, 0x69, 0x6E, 0x74};
```
aplanas: Why no generate a random one per user, and persist it in the SD card?

jpirnay: In the end that would not make a lot difference, would it? If you have the SD-card, you have stuff...

aplanas: Indeed ... I was thinking on when some settings are shared in a forum, for example. In that case, then maybe the obfuscation can be dropped?

jpirnay: I am just checking whether the unique MAC-address of the chip would not work... Then you would need device + sd card, sd card alone (or publishing) would not provide an (immediate) breach

jpirnay: Yupp, that works 😄

osteotek: Does this mean editing json config to update wifi password is impossible? To me easy adding of wifi passwords beats obfuscation, tbh

jpirnay: Not my call to make - removal of obfuscation is easy, but someone needs to say what's more important - security or convenience.
(Although I could check whether a "password_clear" exists, read it, obfuscate and write back, delete it)

aplanas: > password_clear

that is a nice idea...

ngxson:
Personally I think it's not a good idea to use JSON. I'm pretty sure that the JSON parsing code will add quite a lot of overhead for both flash and RAM usage, especially when used with templated types. This PR seems to take >10KB flash already:

```
PR:
RAM:   [===       ]  31.0% (used 101692 bytes from 327680 bytes)
Flash: [==========]  97.3% (used 6376920 bytes from 6553600 bytes)

master:
RAM:   [===       ]  31.0% (used 101708 bytes from 327680 bytes)
Flash: [==========]  97.5% (used 6389766 bytes from 6553600 bytes)

```

Instead, I think we can implement a more future-proof binary format (can be shared for caching too).

src/JsonSettingsIO.cpp
ngxson: A more simple solution is to create a file `wifi.txt` that has first line as wifi name and second line as wifi password:

```
MyHomeWifi789
this_1s_the_passw0rd
```

jpirnay: The objective is not to modernize wifi or koreader settings save / restore, just for sake of easier external readibility. I want to step away from that binary fixed format setting file to ease the addition of new settings for future extensions / personalisation. I faced the challenge myself a couple of times during the few days I started contributing to this repo. Any new feature I wanted to be user-selectable faced the same challenge - when would the PR be merged aka in which position do I need to store / will I find my setting. Now  this is no longer relavant:  regardless where I define, when I define a new setting - it will _just_ end up in the json file.

jpirnay: And btw: I created my wifi.json file from scratch:
```
{"lastConnectedSsid":"PYSY","credentials":[{"ssid":"PYSY","password":"my_very_secret_password"}]
```
and after the next connection, ie read of the setting it got converted to:
```
{"lastConnectedSsid":"PYSY","credentials":[{"ssid":"PYSY","password_obf":"qSb/nO9XvDmR1tko"}]
```

jpirnay:
> Personally I think it's not a good idea to use JSON. I'm pretty sure that the JSON parsing code will add quite a lot of overhead for both flash and RAM usage, especially when used with templated types. This PR seems to take >10KB flash already:
> 
> ```
> PR:
> RAM:   [===       ]  31.0% (used 101692 bytes from 327680 bytes)
> Flash: [==========]  97.3% (used 6376920 bytes from 6553600 bytes)
> 
> master:
> RAM:   [===       ]  31.0% (used 101708 bytes from 327680 bytes)
> Flash: [==========]  97.5% (used 6389766 bytes from 6553600 bytes)
> ```
> 
> Instead, I think we can implement a more future-proof binary format (can be shared for caching too).

This wouldn't address my grievance I outlined above. Yes we are losing 10-12k (and you know how much I fight for flash size reduction) but we would gain a lot of flexibility.

ngxson (commented):
On second though, it can be a good idea to completely replace the settings REST API with this system, see `CrossPointWebServer`, should be done in a follow-up PR I think.

For now, the only thing I think should be improved (can be a follow-up PR) is the way settings are being defined. Imagine a new developer comes to the project, need to add a new setting, and now there are multiple places to be modified.

Instead, a better pattern is to have a centralized definition, much like `SettingsList.h`, something like:

```cpp
class SettingEntryUInt8 {
  const char* name;
  uint8_t valueDefault;
  uint8_t valueMin;
  uint8_t valueMax;
  bool read(uint8_t& output);
  void write();
};
class SettingEntryString {
  const char* name;
  bool read(std::string& output);
  void write();
};
```

This will also technically reduce some flash space, because less templated function instances are being compiled.

jpirnay:
> For now, the only thing I think should be improved (can be a follow-up PR) is the way settings are being defined. Imagine a new developer comes to the project, need to add a new setting, and now there are multiple places to be modified.

You are right: to add a single new uint8_t setting today, I need to touch 3 files, 4 locations:
 - CrossPointSettings.h: - add data member + enum constants
 - SettingsList.h - add UI/API registry entry
 - JsonSettingsIO.cpp - add to saveSettings() (line ~49-78) AND loadSettings() (line ~97-133)

The irony is that SettingsList.h already has almost everything needed for serialization (JSON key, member pointer, enum count via enumValues.size()), but JsonSettingsIO.cpp duplicates all of it manually.

What about JsonSettingsIO iterate over getSettingsList() instead of hardcoding each field. The existing SettingInfo struct already has:
- key — JSON key name
- valuePtr — member pointer to the uint8_t
- enumValues.size() — max valid value for clamping
- type — toggle/enum/value/string
- stringPtr/stringMaxLen — for string fields

This would reduce "add a new setting" to just 2 files (CrossPointSettings.h + SettingsList.h) with zero changes to serialization code.

It would need some additional logic for default values and obfuscating, though.

Can we move this to a new PR?

CaptainFrito:
How is the performance of JSON parsing on the device? This is the main advantage of using binary, the performance impact of reading files is minimal.

ngxson:
Given the complexity of the current setting, I think the difference will be in the order of < 10ms, making it too small to be noticeable.

But FWIW, in my experience working at [llama.cpp](https://github.com/ggml-org/llama.cpp), we try **not** to use json whenever possible as it is quite a double sword. Main problems are: templated functions can take huge amount of time to compile, takes up program space, developers tend to abuse JSON in many circumstances, making weak typing and difficult to do static code analysis, etc. If you look closer into the project, you will see that we have a format called GGUF which is an implementation of key-value but in binary format.

However, we still need a high-level system to make it accessible for everyone. That's why we recently add support for INI format: https://github.com/ggml-org/llama.cpp/issues/17850 . The main point of INI is that the parser is dead simple.

But honestly I have no preference if we stay on binary format or move to JSON or something else, the reason I explained above: we already allow changing settings from web API via a JSON request. That make it less likely to get rid of JSON. Otherwise, if we choose to use something other than JSON, we can update the API accordingly, but still the same amount of works I think.

jpirnay:
I chose json, as that library was alread used. Apart from that: I couldnt care less whether its yaml, json, ini or whatever. my usecase was the need to step away from a fixed order, fixed format setting file.

ngxson:
> my usecase was the need to step away from a fixed order, fixed format setting file.

Just my 2c, no intention to push back this PR, but I just want to clarify one or two things:

If "human-readable" is not one of your goals, the using json here is pretty much overkill. A simple example is: when you want to read a number, json parser need to parse text to number, then because it can be either a float or int, depending on the templated getter, it must return you the correct data.

If you need a simple out-of-order serialization protocol, just a simple mapping key ID --> value should be enough. Number types can be stored as-is without converting to text. Dynamic-length data like a string can be handled by appending an int to indicate the number of bytes to be read. IIRC gRPC work more or less this way.

src/JsonSettingsIO.cpp
```diff
@@ -0,0 +1,281 @@
...
+  s.shortPwrBtn = clamp(doc["shortPwrBtn"] | (uint8_t)S::IGNORE, S::SHORT_PWRBTN_COUNT, S::IGNORE);
+  s.orientation = clamp(doc["orientation"] | (uint8_t)S::PORTRAIT, S::ORIENTATION_COUNT, S::PORTRAIT);
+  s.sideButtonLayout =
+      clamp(doc["sideButtonLayout"] | (uint8_t)S::PREV_NEXT, S::SIDE_BUTTON_LAYOUT_COUNT, S::PREV_NEXT);
+  s.frontButtonBack =
+      clamp(doc["frontButtonBack"] | (uint8_t)S::FRONT_HW_BACK, S::FRONT_BUTTON_HARDWARE_COUNT, S::FRONT_HW_BACK);
+  s.frontButtonConfirm = clamp(doc["frontButtonConfirm"] | (uint8_t)S::FRONT_HW_CONFIRM, S::FRONT_BUTTON_HARDWARE_COUNT,
+                               S::FRONT_HW_CONFIRM);
+  s.frontButtonLeft =
+      clamp(doc["frontButtonLeft"] | (uint8_t)S::FRONT_HW_LEFT, S::FRONT_BUTTON_HARDWARE_COUNT, S::FRONT_HW_LEFT);
```
daveallie, line 114: For all these lines, what's the purpose of the logical or? Feels like it's going to change the value in a lot of cases. E.g. this case here, if you save `frontButtonLeft` as "Confirm", next time this is loaded, it's going to load `frontButtonLeft` as "Right" (as confirm is mapped to `1` and `FRONT_HW_LEFT` is `2`, so it's loaded as `1 | 2`).

Afaict, there's no need to OR each of the input values?

@coderabbitai is that right?

daveallie:
Given the reliance on JSON for the API, and ubiquity of it, I'm fine with moving ahead here with JSON here over other formats.

src/JsonSettingsIO.cpp
daveallie: TIL, thanks @coderabbitai

daveallie (approved):

PR #923, opened 2026-02-16
fix: add bresenham for arbitrary lines

jpirnay:
## Summary

* GfxRender did handle horizontal and vertical lines but had a TODO for arbitrary lines.
* Added integer based Bresenham line drawing 
  
## Additional Context

---

### AI Usage


jpirnay:
> Reads correct to me. I'd guess there's nowhere in the UI we exercise this right now, given that it logged an error before. Were you able to demonstrate any tests locally?

I encountered it when I wanted to draw an "X" on a bitmap indicating that a cover file could not be found.

jpirnay:
Just a test case

[screenshot.bmp](https://github.com/user-attachments/files/25366297/screenshot.bmp)

PR #928, opened 2026-02-16
chore: remove proper names translation

mrtnvgr:
## Summary

I think that proper names, such as: fonts, `CrossPoint` branding should not be translated.

## Additional Context

  specific areas to focus on).

---

### AI Usage


mrtnvgr:
This adds another issue that these values are in several locations and can differ. Also, the classes are expecting the `StrId`, not a string literal. This will need more refactoring in order to be implemented correctly.

Issue #921, opened 2026-02-16
Font text issue with ș and ț characters


PetreBogdan:
For a specific book I have on calibre, the characters are rendering weird. I have books with the same characters but they have not the same issue.

![image](https://github.com/user-attachments/assets/e4b93c6f-873e-45ba-b3b6-009fa7894848)

PetreBogdan:
I might found the issue. Crosspoint can render ş special character and this is found in all of my other books but it can't render the ș character which looks the same but is actually different.

jonasdiemer:
We likely have not included those glyphs in the fonts when creating them with lib/EpdFont/scripts/fontconvert.py

Eloren1:
> I might found the issue. Crosspoint can render ş special character and this is found in all of my other books but it can't render the ș character which looks the same but is actually different.

Well they look different

<img width="92" height="79" alt="Image" src="https://github.com/user-attachments/assets/c7b74bfa-51e6-4a1a-83c6-6e0b792e33a7" />

PR #940, opened 2026-02-16
feat: added special characters for extended latin and african languages

PetreBogdan:
## Summary

What is the goal of this PR?

I want to fix the font issue opened by me here: #921 (also wanting to come out of my shell by opening my first contribution ever or any project on the Github)

## Additional Context

I don't have a spare device to test if this is actually fixing my issue but I will look into the emulator someone shared for the community some days ago.

---

### AI Usage

Partially, for the right hex unicode codes.

### DIFF
--- a/lib/EpdFont/scripts/fontconvert.py
+++ b/lib/EpdFont/scripts/fontconvert.py
@@ -36,6 +36,9 @@ intervals = [
     ### Latin Extended-A ###
     # Eastern European and Baltic languages
     (0x0100, 0x017F),
+    ### Latin Extended-B ###
+    # Latin letters for Eastern European langauges and African languages
+    (0x0180, 0x024F),
     ### General Punctuation (core subset) ###
     # Smart quotes, en dash, em dash, ellipsis, NO-BREAK SPACE
     (0x2000, 0x206F),


bilalix:
Please let me know if you managed to make the emulator work on Windows with Crosspoint v1.0.0, I tried and got it partially working, but the components in the main UI were missing 

If you didn't succeed, you can flash your current X4 test it, and flash it back to the stable firmware (I'm still waiting for mine to arrive as well, it will take weeks)

znelson:
This change is incomplete. You'll need to run lib/EpdFonts/scripts/convert-builtin-fonts.sh, which will update the generated font headers in lib/EpdFonts/builtinFonts. Then run lib/EpdFont/scripts/build-font-ids.sh and copy the output into src/fontIds.h.

Unfortunately, after doing that, the additional character bitmaps push the firmware over the available flash memory:
```
Flash: [==========]  116.0% (used 7600642 bytes from 6553600 bytes)
Error: The program size (7600642 bytes) is greater than maximum allowed (6553600 bytes)
```

PetreBogdan:
So maybe just add the two characters are missing for my case?

PetreBogdan:
It seems it's not working for me because of the storage limitations.

ariel-lindemann:
@PetreBogdan, a lot of space has been free up in the meantime. I'm working on a romanian translation for the UI (#987) and it would be great to have this in, because right now `ș` and `ț` are not showing, so i had to replace them with the alternative characters

PR #948, opened 2026-02-17
feat: Prefer ".sleep" over "sleep" for custom image directory

jpirnay:
## Summary

* Custom sleep screen images now load from /.sleep directory (preferred), falling back to /sleep for backwards compatibility. The dot-prefix keeps the directory hidden from the file browser.
* Rewrote User Guide section 3.6 to document all six sleep screen modes, cover settings, and the updated custom image setup.

## Additional Context

* The sleep directoy entry while browsing files was distracting.

---

### AI Usage

### DIFF
--- a/USER_GUIDE.md
+++ b/USER_GUIDE.md
@@ -179,7 +179,7 @@ The **Sleep Screen** setting controls what is displayed when the device goes to 
 When using **Cover** or **Cover + Custom**, two additional settings apply:
 
 - **Sleep Screen Cover Mode**: **Fit** (scale to fit, white borders) or **Crop** (scale and crop to fill the screen).
-- **Sleep Screen Cover Filter**: **None** (grayscale), **Contrast** (black & white), or **Inverted Contrast** (inverted black & white).
+- **Sleep Screen Cover Filter**: **None** (grayscale), **Contrast** (black & white), or **Inverted** (inverted black & white).
 
 #### Custom images
 


Levrk:
conflicts with #887 which allows for images to be viewed from the file browser

osteotek:
If we merge this I assure you there will be feedback from users who preferred it the old way 😄

jpirnay:
> If we merge this I assure you there will be feedback from users who preferred it the old way 😄 

That's the reason I made it backward compatible, it will read the old directory

Levrk:
While I agree it is weird in the current state having the file browser show the sleep folder when it appears empty. I did just implement bmp viewing from the file browser in #887. I think that the file browser should have a bit more functionality moving forward, and so hiding these files long term doesn't make sense to me personally.

elvisgastelum:
If we merge this in, I suggest adding a way to show/hide hidden folders and files directly from the file manager from hotspot/network (in a quick or convenient way). I currently use this file manager to activate and deactivate sleep covers quickly, without having to remove the SD card and use a computer to move the files.

My current workflow is:

To activate a sleep cover, I move it from a custom folder called `sleep-archive` to the `sleep` folder.

To deactivate a sleep cover, I move it from the `sleep` folder back to `sleep-archive`.

I know I’m probably the only one doing this, but if we can toggle hidden folders/files on and off, I’ll still be able to manage my sleep covers on the go, without needing to remove the SD card from the Xteink.

This probably help more users in the future to be more comfortable with the system

osteotek:
@elvisgastelum `sleep` folder can still be used for images, and you your workflow should not change. @jpirnay correct me if I'm wrong

elvisgastelum:
> @elvisgastelum `sleep` folder can still be used for images, and you your workflow should not change. @jpirnay correct me if I'm wrong

Yeah i understood, i just suggest that feature because I would like to migrate to a dot folder instead of sleep, i see the .sleep pretty useful when im looking for the books in the internal file explorer and I don’t want to see any other noisy files or folders

jpirnay:
> @elvisgastelum `sleep` folder can still be used for images, and you your workflow should not change. @jpirnay correct me if I'm wrong
```
// Check if we have a /.sleep (preferred) or /sleep directory
  const char* sleepDir = nullptr;
  auto dir = Storage.open("/.sleep");
  if (dir && dir.isDirectory()) {
    sleepDir = "/.sleep";
  } else {
    if (dir) dir.close();
    dir = Storage.open("/sleep");
    if (dir && dir.isDirectory()) {
      sleepDir = "/sleep";
    }
  }
```
Will support both directory structures

mrtnvgr:
If `.sleep` is valid, it's logical that `.sleep.bmp` should also be :)

mrtnvgr:
Does this have any blockers?

src/activities/boot_sleep/SleepActivity.cpp
```diff
@@ -32,9 +32,20 @@ void SleepActivity::onEnter() {
...
+  auto dir = Storage.open("/.sleep");
   if (dir && dir.isDirectory()) {
+    sleepDir = "/.sleep";
+  } else {
+    if (dir) dir.close();
+    dir = Storage.open("/sleep");
+    if (dir && dir.isDirectory()) {
+      sleepDir = "/sleep";
+    }
+  }
```
znelson, line 46: Small thought to reduce repetition here:
```suggestion
  // Check if we have a /.sleep (preferred) or /sleep directory
  static constexpr const char* kSleepDirs[] = {"/.sleep", "/sleep"};
  const char* sleepDir = nullptr;
  FsFile dir;
  for (const auto* candidate : kSleepDirs) {
    dir = Storage.open(candidate);
    if (dir && dir.isDirectory()) {
      sleepDir = candidate;
      break;
    }
    if (dir) dir.close();
  }
```

Issue #947, opened 2026-02-17
Crash when epub contains large CSS file


sam-lord:
### Affected Version

v1.0.0

### Describe the Bug

Some poorly formed epub files contain extra CSS files, in my case the full CSS for a website (>200kb, styles buttons, modals, etc).

Opening an epub file like this will crash the reader. It would be better to skip large CSS files, since they're not likely to be relevant to the book. I can't include the ebook I'm testing with for copyright reasons, but the logs show that there's core.css (actual styles for the book) & style.css (styles for... the website of the publisher??).

I've tested skipping large files and it worked, the book opens just fine.

### Steps to Reproduce

1. Create an epub with a large (200kb+) CSS file in the zip
2. Transfer to the x4
3. Open with the browse files menu
4. x4 reboots

### Expected Behavior

The book should open successfully, ignoring any large CSS files since they won't be relevant to the display of text in a book.

### Relevant Log Output/Screenshots

```shell
12:43:20] [DBG] [COF] Entering guide state.
[12:43:20] [DBG] [COF] Found text reference in guide: OEBPS/cover.xhtml
[12:43:20] [ERR] [ZIP] Decompressed 5049 bytes into 30112 bytes
[12:43:20] [DBG] [EBP] Successfully parsed content.opf
[12:43:20] [DBG] [EBP] Parsing CSS file: OEBPS/core.css
[12:43:20] [ERR] [ZIP] Decompressed 3209 bytes into 15021 bytes
[12:43:21] [DBG] [CSS] Parsed 185 rules from 15021 bytes
[12:43:21] [DBG] [EBP] Parsing CSS file: OEBPS/style.css
[12:43:21] [ERR] [ZIP] Decompressed 35095 bytes into 259292 bytes
abort() was called at PC 0x4210b903 on core 0
Core  0 register dump:
MEPC    : 0x4038242a  RA      : 0x40387a44  SP      : 0x3fcacf50  GP      : 0x3fc8f000
TP      : 0x3f807ebc  T0      : 0x37363534  T1      : 0x7271706f  T2      : 0x33323130
S0/FP   : 0x3fcacf7c  S1      : 0x3fcacf7c  A0      : 0x3fcacf88  A1      : 0x3fcacf6a
A2      : 0x00000000  A3      : 0x3fcacfb5  A4      : 0x00000001  A5      : 0x3fca6000
[...]

PR #952, opened 2026-02-17
fix: Skip large CSS files to prevent crashes

sam-lord:
## Summary

**What is the goal of this PR?** (e.g., Implements the new feature for file uploading.)
* Fixes: https://github.com/crosspoint-reader/crosspoint-reader/issues/947

* Check to see if there's free heap memory before processing CSS (should we be doing this type of check or is it better to just crash if we exhaust the memory?)
* Skip CSS files larger than 128kb

## Additional Context

* I found that a copy of `Release it` contained a 250kb+ CSS file, from the homepage of the publisher. It has nothing to do with the epub, so we should just skip it
* Major question: Are there better ways to detect CSS that doesn't belong in a book, or is this size-based approach valid?
* Another question: Are there any epubs we know of that legitimately include >128kb CSS files?

Code changes themselves created with an agent, all investigation and write-up done by human. If you (the maintainers) would prefer a different fix for this issue, let me know.

---

### AI Usage

### DIFF
--- a/lib/Epub/Epub.cpp
+++ b/lib/Epub/Epub.cpp
@@ -258,6 +258,12 @@ bool Epub::parseTocNavFile() const {
 }
 
 void Epub::parseCssFiles() const {
+  // Maximum CSS file size we'll attempt to parse (uncompressed)
+  // Larger files risk memory exhaustion on ESP32
+  constexpr size_t MAX_CSS_FILE_SIZE = 128 * 1024;  // 128KB
+  // Minimum heap required before attempting CSS parsing
+  constexpr size_t MIN_HEAP_FOR_CSS_PARSING = 64 * 1024;  // 64KB
+
   if (cssFiles.empty()) {
     LOG_DBG("EBP", "No CSS files to parse, but CssParser created for inline styles");
   }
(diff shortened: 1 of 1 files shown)


osteotek:
CSS file bigger than 128kb sounds like malformed epub. We can probably just skip them without trying to parse. What do you think @jdk2pq ? I seem to remember you had a file limit in your recent PR but can't find it now

jdk2pq:
Thank you for the fix @sam-lord!!

> CSS file bigger than 128kb sounds like malformed epub. We can probably just skip them without trying to parse. What do you think @jdk2pq ? I seem to remember you had a file limit in your recent PR but can't find it now

Not necessarily malformed, but definitely poorly put together by the publisher (especially in @sam-lord's case where the publisher appears to be including their _website's_ CSS inside of the epub 🙃). This fix makes sense to me! 

As for the file limit that was in my PR previously, I removed that since [the thought was ](https://github.com/crosspoint-reader/crosspoint-reader/pull/779#discussion_r2782385637)we were already buffering the CSS file, and we'd hit the MAX_RULES limit anyway (and it looks like we do indeed 😄).

I tested the existing code out by making an EPUB with ~200KB of Tailwind's CSS, and I hit the same error. After the fix, it's working great. I'll add that test EPUB to my repo [here](https://github.com/jdk2pq/css-test-epub).

Longer-term, we should think about:
- only saving rules in `rulesBySelector_` if they contain supported CSS properties. This should be a relatively simple change that could be follow-on after this PR.
- dropping complex selectors and pseudo-selectors like `:hover`, `:focus`, etc. really aren't needed
- using [LibCSS](https://www.netsurf-browser.org/projects/libcss/) for a dedicated parsing library that should have better memory usage

jdk2pq (approved):

PR #960, opened 2026-02-18
feat: PoC of fonts on data partition

znelson:
## Summary

8 MB executable, 8 MB data/fonts. The UI fonts and selected reader font are mmap-ed at runtime.

Includes the Latin-B characters proposed in #940 to demonstrate expanded font capabilities.

## Additional Context

Flash with: 
```
pio run -t upload && pio run -t uploadfontdata
```

Firmware size w/o fonts:
```
Flash: [===       ]  26.4% (used 2217048 bytes from 8388608 bytes)
```

---

### AI Usage

Did you use AI tools to help write this code? _**YES Claude Opus 4.6**_

### DIFF
--- a/lib/EpdFont/FontPartition.cpp
+++ b/lib/EpdFont/FontPartition.cpp
@@ -8,9 +8,9 @@
 
 // ── On-flash structs (must match fontconvert.py v2 exactly) ─────────────────
 
-static constexpr uint32_t PARTITION_MAGIC   = 0x43504654;  // "CPFT"
+static constexpr uint32_t PARTITION_MAGIC = 0x43504654;  // "CPFT"
 static constexpr uint32_t PARTITION_VERSION = 2;
-static constexpr uint8_t  PARTITION_SUBTYPE = 0x40;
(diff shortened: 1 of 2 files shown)


lib/EpdFont/FontPartition.cpp
```diff
@@ -0,0 +1,268 @@
...
+  // Map the UI group
+  uiGroupIndex = findGroupIndex("ui");
+  if (uiGroupIndex < 0) {
+    LOG_ERR("FONT", "No 'ui' group in partition");
+    return false;
+  }
+
+  const auto& uiGroup = groupDir[uiGroupIndex];
+  const void* ptr = nullptr;
+  err = esp_partition_mmap(partition, uiGroup.dataOffset, uiGroup.dataSize, SPI_FLASH_MMAP_DATA, &ptr, &uiMmapHandle);
```
ngxson, line 186: Note that when working on https://github.com/crosspoint-reader/crosspoint-reader/pull/596 , I ran into the problem that we may not have enough data pages for mapping, so we may need to fallback to using instruction pages and having to access data with 4-byte alignment.

PR #963, opened 2026-02-18
fix: Correct word width and space calculations

znelson:
## Summary

This change fixes an issue I noticed while reading where occasionally, especially in italics, some words would have too much space between them. The problem was that word width calculations were including any negative X overhang, and combined with a space before the word, that can lead to an inconsistently large space.

## Additional Context

Screenshots of some problematic text:

| In CrossPoint 1.0 | With this change |
| -- | -- |
|  |  |

---

### AI Usage


osteotek:
@znelson I think that's a valid fix, what was the reason behind closing it?

znelson:
@osteotek I think it is correct, too. I had another bug in my working branch and for a moment there I thought it was caused by this change.

PR #967, opened 2026-02-18
feat: Avoid css crash, css rule selection logic

jpirnay:
## Summary

* I am well aware of the PR #952 which is definitely a solution to avoid memory crashes
* I am proposing a different approach instead of completely discarding larger css files
* Smaller css files will still be read via the existing memory parsing routine (btw around 20% faster than the existing one due to some optimisations)
* Larger css files will be read and processed in chunks (no size limit, tested with 512k 2500 rule css-file)
* Both variants will 
  - limit rules to 300 
  - categorize rules into essential / important / nice to have (based on a simple evaluation logic).
  - If the 300 limit is reached then low priority rules will be discarded / replaced by a newer higher priority rule
  - This is by no means foolproof but a viable plausibility check
 * Relevant rules will be loaded even for bigger css files that have been previously ignored
 
## Additional Context
* The priority score is exclusively used for cache eviction(highest → lowest)
  - element.class (e.g. p.note) ⇒ 9
  - class selector (e.g. .note) ⇒ 7
  - element selector (e.g. p, div) ⇒ 5
  - complex selectors (contain >, +, ~, :, [ ) ⇒ 1
  - empty selector ⇒ 0

---

### AI Usage

Did you use AI tools to help write this code? _**PARTIAL**_


osteotek:
@jpirnay code diff seems to include changes to how thumbnail images are generated which is not really in the scope of this PR? Might be merging gone wrong

daveallie:
I missed this PR and put up a pretty naive stripping of unused rules here: https://github.com/crosspoint-reader/crosspoint-reader/pull/1014

However keeping rules based on importance seems like a better long term solution.

jdk2pq (changes requested):
@jpirnay This PR appears to unintentionally revert the changes from here https://github.com/crosspoint-reader/crosspoint-reader/pull/779/changes I think this PR just need to be rebased on the latest changes from master. That should help clean up the diff for reviewing too.

jpirnay:
Outdated due to too many changes in the codebase

PR #977, opened 2026-02-18
fix: Don't extract unsupported formats

jpirnay:
## Summary

* During chapter parsing, every  tag triggered ZIP decompression and an SD card write regardless of whether the image format was supported. The mandatory delay(50) after each SD write compounded the cost. A chapter with 6 GIF images (a common decorative element in older EPUBs) wasted ~750 ms before any text rendering began.
Added an ``ImageDecoderFactory::isFormatSupported()`` check before any file I/O in the img-handler. Only JPEG and PNG proceed to extraction; all other formats (GIF, SVG, WebP, etc.) fall through immediately to alt-text rendering with no SD card access.
## Additional Context

Measured impact on a representative chapter with 6 GIF decorations:
|  | Before | After|
|-- | -- | --|
|Total parse time | ~882 ms | ~207 ms|
|Image handling | ~750 ms | ~76 ms|
---

### AI Usage


osteotek (approved):

insunaa:
Does the timing function have to be there in the final code?

Does crosspoint reader code have an equivalent to
```cpp
#ifdef NDEBUG
```
to exclude things like this from release builds?

I'm aware that the impact of this is trivial at best, I'm trying to learn how the project works.

daveallie (approved):

PR #978, opened 2026-02-18
perf: Improve font drawing performance

jpirnay:
## Summary

* ``renderChar`` checked ``is2Bit`` on every pixel inside the inner loop, even though the value is constant for the lifetime of a single glyph
* Moved the branch above both loops so each path (2-bit antialiased / 1-bit monochrome) runs without a per-pixel conditional
* Eliminates redundant work in the two inner loops that render font glyphs to the frame buffer, targeting ``renderChar`` and ``drawTextRotated90CW`` in ``GfxRenderer.cpp``

## Additional Context
* Measured on device using a dedicated framebuffer benchmark (no display refresh). 100 repetitions of "The quick brown fox jumps".

| Test            | Before          | After           | Change  |
|-----------------|-----------------|-----------------|---------|
| drawText UI12   | 1,337 µs/call | 1,024 µs/call | −23%|
| drawText Bookerly14 | 2.174 µs / call | 1,847 µs/call | −15% |

---

### AI Usage

Did you use AI tools to help write this code? _**PARTIALLY**_ Claude did the analysis and wrote the benchmarks

### DIFF
--- a/lib/GfxRenderer/GfxRenderer.cpp
+++ b/lib/GfxRenderer/GfxRenderer.cpp
@@ -79,13 +79,12 @@ static void renderCharImpl(const GfxRenderer& renderer, GfxRenderer::RenderMode 
 
   const EpdFontData* fontData = fontFamily.getData(style);
   const bool is2Bit = fontData->is2Bit;
-  const uint32_t offset = glyph->dataOffset;
   const uint8_t width = glyph->width;
   const uint8_t height = glyph->height;
   const int left = glyph->left;
   const int top = glyph->top;
 
-  const uint8_t* bitmap = &fontData->bitmap[offset];
+  const uint8_t* bitmap = renderer.getGlyphBitmap(fontData, glyph);
 
   if (bitmap != nullptr) {
     // For Normal:  outer loop advances screenY, inner loop advances screenX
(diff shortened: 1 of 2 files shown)


lib/GfxRenderer/GfxRenderer.cpp
```diff
@@ -1001,9 +1014,17 @@ void GfxRenderer::renderChar(const EpdFontFamily& fontFamily, const uint32_t cp,
...
+      }
+    } else {
+      for (int glyphY = 0; glyphY < height; glyphY++) {
+        const int screenY = *y - glyph->top + glyphY;
+        int pixelPosition = glyphY * width;
+        for (int glyphX = 0; glyphX < width; glyphX++, pixelPosition++) {
+          const int screenX = screenXBase + glyphX;
+
+          const uint8_t byte = bitmap[pixelPosition >> 3];
+          const uint8_t bit_index = 7 - (pixelPosition & 7);
```
znelson: The amount of duplication between `renderChar` and `drawTextRotated90CW` is unfortunate. Consider factoring them out into one? 

Here's an example using `if constextpr` to avoid any additional runtime overhead: https://github.com/znelson/crosspoint-reader/commit/7913eaab0b075758ff7590115e65bc1b18429318

jpirnay: Done, the refactor is actually marginally faster:
|Test | Before | After | Change|
|-- | -- | -- | --|
|drawText UI12 | 1,044 µs/call | 1,024 µs/call | −2%|
|drawText Bookerly14 | 1,941 µs/call | 1,847 µs/call | −5%|

Within noise for UI12, and a small but real improvement for Bookerly14 — likely because the renderCharImpl template structure made it easier for the compiler to optimise the two paths independently.

znelson (dismissed):

daveallie (dismissed):

PR #979, opened 2026-02-18
feat: Implement silent pre-indexing for the next chapter in EpubReaderActivity

LSTAR1900:
## Summary

* A simple tweak to pre-index the next chapter silently during normal reading.
* Triggers silent pre-indexing of the next chapter when the penultimate page of a chapter is rendered to reduce visible interruptions.
* Keeps existing indexing with popup when a reader jumps directly into an unindexed chapter.

## Additional Context

* Reader input is temporarily blocked during silent indexing to avoid navigation/index state conflicts.
* The penultimate page is used because readers typically spend longer there than on the final page.
* This change optimizes linear reading flow while preserving reliable indexing for non-linear navigation.

## Possible Improvements

* Add a setting for First Page Indexing vs Penultimate Page Pre-indexing
* Display an indexing icon in the status bar instead of using a popup that overlaps book text.

Tested on device:
https://www.dropbox.com/scl/fi/29g5kjqgsi5e4hgujv38u/Silent-Indexing.MOV?rlkey=yemi4mosmev5vicaa7gpe49qw&dl=0

---

### AI Usage

### DIFF
--- a/src/activities/reader/EpubReaderActivity.cpp
+++ b/src/activities/reader/EpubReaderActivity.cpp
@@ -684,14 +684,16 @@ void EpubReaderActivity::silentIndexNextChapterIfNeeded(const uint16_t viewportW
   Section nextSection(epub, nextSpineIndex, renderer);
   if (nextSection.loadSectionFile(SETTINGS.getReaderFontId(), SETTINGS.getReaderLineCompression(),
                                   SETTINGS.extraParagraphSpacing, SETTINGS.paragraphAlignment, viewportWidth,
-                                  viewportHeight, SETTINGS.hyphenationEnabled, SETTINGS.embeddedStyle)) {
+                                  viewportHeight, SETTINGS.hyphenationEnabled, SETTINGS.embeddedStyle,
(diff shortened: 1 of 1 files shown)


LSTAR1900:
After searching, this had been discussed in #72

iandchasse:
I really like this change, works great it seems!

znelson:
The one part of this that makes me a little uncomfortable is the lack of visibility leading to what might feel like glitches or hangs if you don't know exactly what's going on. 

When I read on my kindle at night (with its front light) and my X4 during the day, I end up doing a little bit of skipping ahead to find the spot I left off on the other device. With this change, a few rapid page turns on the X4 which cross a chapter boundary will now have the section indexing hang on a page turn that doesn't initially appear to correspond with a chapter boundary.

Ultimately I think the best experience would be to have something like this _and_ improve section indexing to be interruptible, so it can still handle button input and redraw. That's a deeper change, though.

Under normal reading circumstances with no skipping ahead, I think this change is a net reading experience improvement.

Issue #993, opened 2026-02-19
[1.1.0-rc] PNG in EPUB causes crash


allgoewer:
### Affected Version

1.0.0-rc+402e8807

### Describe the Bug

Hey all,

This version of [The Maltese Falcon](https://standardebooks.org/ebooks/dashiell-hammett/the-maltese-falcon) (Scroll down and download the "Compatible epub") hard-crashes my X4.

Looks like some kind of double-free after the failed PNG decode to me?

```
[18:27:13] [IMG] Rendering image at 8,261: /.crosspoint/epub_1065909436/img_0_0.png (463x145)
[18:27:13] [IMG] File does not exist: /.crosspoint/epub_1065909436/img_0_0.pxc
[18:27:13] [IMG] Decoding and caching: /.crosspoint/epub_1065909436/img_0_0.png
[18:27:13] [IMG] Using PNG decoder
[18:27:13] [PNG] Decoding PNG: /.crosspoint/epub_1065909436/img_0_0.png
[18:27:13] [PNG] PNG 1400x440 -> 463x145 (scale 0.33), bpp: 8
[18:27:13] [IMG] Allocated cache buffer: 16820 bytes for 463x145
[18:27:13] [PNG] Decode failed: 2
CORRUPT HEAP: Bad tail at 0x3fcbe67c. Expected 0xbaad5678 got 0x00000000
assert failed: multi_heap_free multi_heap_poisoning.c:259 (head != NULL)
Core  0 register dump:
MEPC    : 0x4038242a  RA      : 0x40387a44  SP      : 0x3fcb1ff0  GP      : 0x3fc8ee00
TP      : 0x3f81bb68  T0      : 0x37363534  T1      : 0x7271706f  T2      : 0x33323130
S0/FP   : 0x3fcb2168  S1      : 0x0000007f  A0      : 0x3fcb2054  A1      : 0x3fc9189d
A2      : 0x00000001  A3      : 0x00000029  A4      : 0x00000001  A5      : 0x3fca6000
A6      : 0x7a797877  A7      : 0x76757473  S2      : 0x3fcb2048  S3      : 0x00000001
S4      : 0x3fcb2048  S5      : 0x4038cc7e  S6      : 0x00001a33  S7      : 0x00001a43
S8      : 0x3fcae504  S9      : 0x3fcb224c  S10     : 0x000001cf  S11     : 0x00000091
[...]

PR #995, opened 2026-02-19
fix: Increase PNGdec buffer size to support wide images

osteotek:
## Summary

* Increased `PNG_MAX_BUFFERED_PIXELS` from 6402 to 16416 in `platformio.ini` to support up to 2048px wide RGBA images
* adds a check to abort decoding and log an error if the required PNG scanline buffer exceeds the configured `PNG_MAX_BUFFERED_PIXELS`, preventing possible buffer overruns.
* fixes https://github.com/crosspoint-reader/crosspoint-reader/issues/993

---

### AI Usage

### DIFF
--- a/lib/Epub/Epub/converters/PngToFramebufferConverter.cpp
+++ b/lib/Epub/Epub/converters/PngToFramebufferConverter.cpp
@@ -341,7 +341,7 @@ bool PngToFramebufferConverter::decodeToFramebuffer(const std::string& imagePath
     delete png;
     return false;
   }
-  
+
   if (png->getBpp() != 8) {
     warnUnsupportedFeature("bit depth (" + std::to_string(png->getBpp()) + "bpp)", imagePath);
   }


osteotek:
Increasing PNG scanline buffer fixes crash on book from https://github.com/crosspoint-reader/crosspoint-reader/issues/993 
Haven't tested beyond that though

osteotek:
> The first issue being the PNG decode failing and the second one being the crash caused after the decode fails. There will most likely always be PNGs which are too big to fit, but that error should be handled gracefully.

@allgoewer Correct. Added buffer check, overruning buffer size will stop png decoding with log message

```
[92240] [ERR] [PNG] PNG row buffer too small: need 11234 bytes for width=1400 type=6, configured PNG_MAX_BUFFERED_PIXELS=6402
[92240] [ERR] [PNG] Aborting decode to avoid PNGdec internal buffer overflow
```

martinbrook:
@osteotek what was the png size and type which was causing the crash ? Maybe I can add another page to the test epub generator (https://github.com/crosspoint-reader/crosspoint-reader/blob/master/scripts/generate_test_epub.py) so we can catch any future regressions ?

osteotek:
@martinbrook PNG with width=1400, it's on the first page of this epub - https://standardebooks.org/ebooks/dashiell-hammett/the-maltese-falcon/downloads/dashiell-hammett_the-maltese-falcon.epub

osteotek:
@martinbrook I may have fixed it incorrectly, I haven't been really following how images subsystems work. Please feel free to fix it properly, if needed.

PR #1007, opened 2026-02-19
feat: add KOReader sync server registration menu option

jpirnay:
## Summary

* The current KOReaderSyncClient only implements authentication (login), ie using an existing account
* Users currently need to register accounts through other means (web interface, different KOReader app, etc.)
* Adds a dedicated "Register" entry to the KOReader sync settings menu, allowing users to create a new account on their sync server directly from the device 

## Additional Context

grafik

---

### AI Usage


.skills/SKILL.md
osteotek, line 1: This file out of scope of this PR

jpirnay: Apologies, that was an oversight. Removed

jpirnay:
Too much out of sync - will be incorporated in a new PR

PR #1012, opened 2026-02-19
feat: add Switch Firmware setting to change OTA boot partition

tent4kel:
Adds a "Switch Firmware (reboot)" entry to the System settings tab. When selected, calls esp_ota_set_boot_partition() on the inactive OTA slot (app0 ↔ app1) then esp_restart(). Useful for dual-firmware setups where app1 holds a separate application

## Summary

enabling users to have a focused apps with different features. Or stable and dev versions on the same device.

only 14 lines of code

## Additional Context

* successfully tested on device.

---

### AI Usage

Did you use AI tools to help write this code? _**YES **_


Eloren1:
What if a regular user who doesn't install anything else clicks this? What will be the behavior?

tent4kel:
> What if a regular user who doesn't install anything else clicks this? What will be the behavior?

Good point!
esp_ota_set_boot_partition() validates the image header (magic byte0xE9) before writing the OTA metadata. On an empty partition (all0xFF) it should return ESP_ERR_OTA_VALIDATE_FAILED, which the current code already catches, it logs an error and doesn't reboot.

itsthisjustin:
As we now support SD flashing. What I don't want is a user switching to another firmware that DOESN'T support this and then getting stuck. I'm going to close this for now with the assumption that this is too technical for our average user and if they want to switch firmwares there are other methods that are better. Thanks for the effort on this though, it's a really cool feature.

PR #1013, opened 2026-02-19
fix: Fixed book title in home screen

DestinySpeaker:
## Summary

* The goal is to fix the title of books in the Home Screen. 

Before
IMG_8867

After:
IMG_8868

## Additional Context

  specific areas to focus on).

---

### AI Usage

Did you use AI tools to help write this code? YES, Cursor

### DIFF
--- a/src/RecentBooksStore.cpp
+++ b/src/RecentBooksStore.cpp
@@ -11,7 +11,7 @@
 #include "util/StringUtils.h"
 
 namespace {
-constexpr uint8_t RECENT_BOOKS_FILE_VERSION = 4;
+constexpr uint8_t RECENT_BOOKS_FILE_VERSION = 3;
 constexpr char RECENT_BOOKS_FILE[] = "/.crosspoint/recent.bin";
 constexpr int MAX_RECENT_BOOKS = 10;
 }  // namespace
@@ -86,11 +86,11 @@ RecentBook RecentBooksStore::getDataFromBook(std::string path) const {
   LOG_DBG("RBS", "Loading recent book: %s", path.c_str());
 
   // If epub, try to load the metadata for title/author and cover.
-  // Use buildIfMissing=true so we build the metadata cache when missing (e.g. cache cleared);
-  // otherwise getTitle()/getAuthor() stay blank until the book is opened.
+  // Use buildIfMissing=false to avoid heavy epub loading on boot; getTitle()/getAuthor() may be
+  // blank until the book is opened, and entries with missing title are omitted from recent list.
   if (StringUtils::checkFileExtension(lastBookFileName, ".epub")) {
     Epub epub(path, "/.crosspoint");
-    epub.load(true, true);
+    epub.load(false, true);
     return RecentBook{path, epub.getTitle(), epub.getAuthor(), epub.getThumbBmpPath()};
   } else if (StringUtils::checkFileExtension(lastBookFileName, ".xtch") ||
              StringUtils::checkFileExtension(lastBookFileName, ".xtc")) {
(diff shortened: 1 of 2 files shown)


src/RecentBooksStore.cpp
```diff
@@ -11,7 +11,7 @@
 #include "util/StringUtils.h"
 
 namespace {
-constexpr uint8_t RECENT_BOOKS_FILE_VERSION = 3;
+constexpr uint8_t RECENT_BOOKS_FILE_VERSION = 4;
```
daveallie: The structure of this file isn't changing as far as I can tell so there's no need for this bump or the migration below, unless I'm missing something here?

src/RecentBooksStore.cpp
```diff
@@ -146,15 +171,35 @@ bool RecentBooksStore::loadFromFile() {
...
+      // If title is missing (e.g. file was saved before metadata was available), load from book
+      if (title.empty()) {
+        RecentBook fresh = getDataFromBook(path);
+        if (!fresh.title.empty() || !fresh.author.empty() || !fresh.coverBmpPath.empty()) {
+          title = fresh.title;
+          author = fresh.author;
+          coverBmpPath = fresh.coverBmpPath;
+          metadataRefreshed = true;
+        }
+      }
```
daveallie: I think in this case, we should just omit the book from the `recentBooks` vector instead of attempting to load it again.

src/RecentBooksStore.cpp
```diff
@@ -85,10 +85,12 @@ RecentBook RecentBooksStore::getDataFromBook(std::string path) const {
...
   LOG_DBG("RBS", "Loading recent book: %s", path.c_str());
 
-  // If epub, try to load the metadata for title/author and cover
+  // If epub, try to load the metadata for title/author and cover.
+  // Use buildIfMissing=true so we build the metadata cache when missing (e.g. cache cleared);
+  // otherwise getTitle()/getAuthor() stay blank until the book is opened.
   if (StringUtils::checkFileExtension(lastBookFileName, ".epub")) {
     Epub epub(path, "/.crosspoint");
-    epub.load(false, true);
+    epub.load(true, true);
```
daveallie: I actually consider this being false a plus. If epub loading breaks for a book having his run and potentially crash the device on boot is a bad time.

src/components/themes/lyra/LyraTheme.cpp
```diff
@@ -480,13 +482,73 @@ void LyraTheme::drawRecentBookCover(GfxRenderer& renderer, Rect rect, const std:
...
+    std::vector<std::string> words;
+    words.reserve(8);
+    size_t pos = 0;
+    while (pos < lastBookTitle.size()) {
+      while (pos < lastBookTitle.size() && lastBookTitle[pos] == ' ') ++pos;
+      if (pos >= lastBookTitle.size()) break;
+      const size_t start = pos;
+      while (pos < lastBookTitle.size() && lastBookTitle[pos] != ' ') ++pos;
+      words.emplace_back(lastBookTitle.substr(start, pos - start));
+    }
```
daveallie, line 496: Nit: I think this could be cleaned up a bit

```cpp
std::vector<std::string> words;
std::string::size_type wordStart = 0;
std::string::size_type wordEnd = 0;

// find_first_not_of skips leading/interstitial spaces
while ((wordStart = book.title.find_first_not_of(' ', wordEnd)) != std::string::npos) {
    wordEnd = book.title.find(' ', wordStart);
    words.emplace_back(book.title.substr(start, end - start));
}
```

src/components/themes/lyra/LyraTheme.cpp
```diff
@@ -480,13 +482,73 @@ void LyraTheme::drawRecentBookCover(GfxRenderer& renderer, Rect rect, const std:
...
+    const int maxLineWidth = textWidth;
+    const int spaceWidth = renderer.getSpaceWidth(UI_12_FONT_ID, EpdFontFamily::BOLD);
+    std::vector<std::string> titleLines;
+    std::string currentLine;
+    for (auto& w : words) {
```
CaptainFrito, line 501: It would be great if the text wrapping function could be centralized elsewhere, like in https://github.com/crosspoint-reader/crosspoint-reader/pull/911/changes 
That way we could also clean up the classic theme home screen implementation.

src/RecentBooksStore.cpp
DestinySpeaker: I reverted the version change :)

src/components/themes/lyra/LyraTheme.cpp
DestinySpeaker: Would it require us to refactor the entire Lyra theme?

daveallie: Agree that following https://github.com/crosspoint-reader/crosspoint-reader/pull/911, we should amend these implementations to reduce code duplication and improve readability, but I am also keen to get this one into 1.1, so will merge as is and we can clean up all the implementations once #911 is merged.

PR #1019, opened 2026-02-20
feat: Display file extensions in File Browser

CaptainFrito:
## Summary

IMG_8136 Medium
IMG_8137 Medium
IMG_8138 Medium

## Additional Context

Do we want a setting to toggle this?

---

### AI Usage

### DIFF
--- a/src/activities/home/FileBrowserActivity.cpp
+++ b/src/activities/home/FileBrowserActivity.cpp
@@ -244,6 +244,14 @@ std::string getFileName(std::string filename) {
   return filename.substr(0, pos);
 }
 
+std::string getFileExtension(std::string filename) {
+  if (filename.back() == '/') {
+    return "";
+  }
+  const auto pos = filename.rfind('.');
+  return filename.substr(pos);
+}
+
 void FileBrowserActivity::render(RenderLock&&) {
   renderer.clearScreen();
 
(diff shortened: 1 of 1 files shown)


src/activities/home/FileBrowserActivity.cpp
```diff
@@ -214,7 +222,8 @@ void MyLibraryActivity::render(Activity::RenderLock&&) {
     GUI.drawList(
         renderer, Rect{0, contentTop, pageWidth, contentHeight}, files.size(), selectorIndex,
         [this](int index) { return getFileName(files[index]); }, nullptr,
-        [this](int index) { return UITheme::getFileIcon(files[index]); });
+        [this](int index) { return UITheme::getFileIcon(files[index]); },
+        [this](int index) { return getFileExtension(files[index]); }, false);
```
CaptainFrito, line 274: I'm starting to think that all these lambda functions that get the same file from the array separately and perform string operations are not the most optimized way to go about this. Perhaps a single function that returns a struct with all these elements would work better?

Eloren1:
How does it look on long filenames?

el:
It looks too crowded this way. Maybe we can show extensions for only non-epub files?

CaptainFrito:
> How does it look on long filenames?

IMG_8163 Medium

Long filenames are truncated but the extension is still shown

znelson:
I acknowledge @el's concern that it looks crowded and that we should perhaps omit the extension for .epub files (assuming those are the most common). Let's make that a separate change.

PR #1020, opened 2026-02-20
feat: Themed language screen

CaptainFrito:
## Summary

Added theme support to the language screen

IMG_8139 Medium

## Additional Context

---

### AI Usage

### DIFF
--- a/src/activities/settings/LanguageSelectActivity.cpp
+++ b/src/activities/settings/LanguageSelectActivity.cpp
@@ -30,15 +30,16 @@ void LanguageSelectActivity::loop() {
     return;
   }
 
-  if (mappedInput.wasPressed(MappedInputManager::Button::Up) ||
-      mappedInput.wasPressed(MappedInputManager::Button::Left)) {
-    selectedIndex = (selectedIndex + totalItems - 1) % totalItems;
+  // Handle navigation
+  buttonNavigator.onNextRelease([this] {
+    selectedIndex = ButtonNavigator::nextIndex(static_cast<int>(selectedIndex), totalItems);
     requestUpdate();
-  } else if (mappedInput.wasPressed(MappedInputManager::Button::Down) ||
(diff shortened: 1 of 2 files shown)


lukestein:
No code review but tested on device and seems to work great!

ariel-lindemann:
@CaptainFrito, thanks for the new screen, it looks very nice!

I noticed you changed the localisation string form `STR_ON_MARKER` to `STR_SET` and I wondered if there is a particular reason for that. Is it because of the brackets so it looks nicer?

Because the new string is translated quite diffferently in the translations, since "set" is a bit ambiguous . Currently this is also the only place in the code this is used as far as I could tell. So translators also lacked context maybe. At least for the languages I understand it looks a bit off.

I thought I'd check with you before I open a PR. My proposal would be to make a string called `Selected` which is less ambuguous and it matches the button (`Select`)

CaptainFrito:
> @CaptainFrito, thanks for the new screen, it looks very nice!
> 
> I noticed you changed the localisation string form `STR_ON_MARKER` to `STR_SET` and I wondered if there is a particular reason for that. Is it because of the brackets so it looks nicer?
> 
> Because the new string is translated quite diffferently in the translations, since "set" is a bit ambiguous . Currently this is also the only place in the code this is used as far as I could tell. So translators also lacked context maybe. At least for the languages I understand it looks a bit off.
> 
> I thought I'd check with you before I open a PR. My proposal would be to make a string called `Selected` which is less ambuguous and it matches the button (`Select`)

Thanks for the feedback! I only changed it to try to reduce the number of different strings we have in the app, and Set is already used elsewhere. But I understand if it's not suitable we can change it back, or change STR_ON_MARKER to something else :)

ariel-lindemann:
Ok, thank you that helps. I'll open a PR later today and we can discuss the rest there

Issue #439, opened 2026-01-20
WebDAV /FTP/ SAMBA support


j-xella:
It would be nice if the web server could support the above mentioned protocols, so that the e-reader's file system could be mounted and accessed by file managers.

n4ts:
The WebDav protocol would be great, its implementation is quite simple and would allow easy file transfer on both Windows and Linux (plus synchronization with Calibre or other scripts).

daveallie:
WebDAV support added in #1047, going to mark this closed

PR #1030, opened 2026-02-20
feat: Add WebDAV Class 1 server for SD card access

dexif:
## Summary

PR for issue #439 

Implement WebDAV support so the device can be mounted as a network drive in Windows Explorer, macOS Finder, Linux Nautilus, and mobile file managers.

Methods: OPTIONS, PROPFIND, GET, HEAD, PUT, DELETE, MKCOL, MOVE,  COPY, LOCK, UNLOCK. All handled via onNotFound() to avoid conflicts with existing routes.

## Additional Context

The WebServer library is vendored into lib/WebServer/ because the framework version buffers the entire PUT body in RAM before calling the handler. On ESP32-C3 (~150KB free heap) this makes file uploads over ~100KB impossible. The vendored copy has a minimal patch in Parsing.cpp (~14 lines) that skips body buffering for PUT requests with no registered handler, allowing the WebDAV handler to stream the body from the TCP buffer in 4KB chunks.

### DIFF
--- a/src/network/WebDAVHandler.cpp
+++ b/src/network/WebDAVHandler.cpp
@@ -316,6 +316,7 @@ void WebDAVHandler::handlePut() {
     return;
   }
 
+  bool writeOk = true;
   if (contentLength > 0) {
     uint8_t buf[4096];
     size_t remaining = contentLength;
@@ -323,13 +324,28 @@ void WebDAVHandler::handlePut() {
       esp_task_wdt_reset();
       size_t toRead = remaining < sizeof(buf) ? remaining : sizeof(buf);
       size_t bytesRead = client.readBytes(buf, toRead);
-      if (bytesRead == 0) break;
-      file.write(buf, bytesRead);
+      if (bytesRead == 0) {
+        writeOk = false;
+        break;
+      }
+      size_t written = file.write(buf, bytesRead);
+      if (written != bytesRead) {
+        writeOk = false;
+        break;
+      }
       remaining -= bytesRead;
     }
+    if (remaining > 0) writeOk = false;
   }
 
   file.close();
+
+  if (!writeOk) {
+    Storage.remove(path.c_str());
+    _server->send(500, "text/plain", "Write failed - incomplete upload or disk full");
+    return;
+  }
+
   clearEpubCacheIfNeeded(path);
 
   _server->send(existed ? 204 : 201);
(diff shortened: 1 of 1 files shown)


el:
I like the idea, but the implementation seems pretty heavy. Maybe it could be easier to add FTP support instead?

dexif:
> I like the idea, but the implementation seems pretty heavy. Maybe it could be easier to add FTP support instead?

I made a [pull request](https://github.com/espressif/arduino-esp32/pull/12381) to the library (once they approve the fix, we can remove the library files). There aren't really many changes, as I tried to avoid pulling in third-party libraries. However, I had to patch the existing one to be able to load files larger than 100 KB.

And I tested it and everything works great. It seems to me that ftp will be significantly more expensive in terms of resources and code.

src/network/CrossPointWebServer.cpp
```diff
@@ -159,6 +159,12 @@ void CrossPointWebServer::begin() {
   server->onNotFound([this] { handleNotFound(); });
   LOG_DBG("WEB", "[MEM] Free heap after route setup: %d bytes", ESP.getFreeHeap());
 
+  // Collect WebDAV headers and create handler
+  const char* davHeaders[] = {"Depth", "Destination", "Overwrite", "If", "Lock-Token", "Timeout"};
+  server->collectHeaders(davHeaders, 6);
+  davHandler.reset(new WebDAVHandler(server.get()));
```
el, line 165: This automatically starts the WebDav server when "File Transfer" is used afaics. Is that something we want? I assumed we would have a 4th option in "File Transfer" menu as WebDav Server.

dexif: Essentially, we're already running the web server. In this case, I don't see the point in creating a separate section for enabling webdav, given that processing is happening there as well.

Jason2866:
> > I like the idea, but the implementation seems pretty heavy. Maybe it could be easier to add FTP support instead?
> 
> I made a [pull request](https://github.com/espressif/arduino-esp32/pull/12381) to the library (once they approve the fix, we can remove the library files). There aren't really many changes, as I tried to avoid pulling in third-party libraries. However, I had to patch the existing one to be able to load files larger than 100 KB.
> 
> And I tested it and everything works great. It seems to me that ftp will be significantly more expensive in terms of resources and code.

Your current platformio.ini does **not** use actual Arduino core. If your PR in espressif32 Arduino core is accepted you have to change to switch from your currently used Arduino core 2.0.17 to actual 3.3.x
Core 2.0.17 is outdated and out of support.

dexif:
> Core 2.0.17 is outdated and out of support.

If the pull request to the framework is accepted, I was planning to work on the update.

src/network/CrossPointWebServer.cpp
el: Webserver is already pretty slow and resource intensive, wouldnt this put more pressure on it?

dexif: I can’t imagine what could create load here. Essentially, it’s just a few additional request handlers.

src/network/WebDAVHandler.cpp
```diff
@@ -0,0 +1,754 @@
...
+  }
+  if (name.isEmpty()) return false;
+
+  if (name.startsWith(".")) return true;
+
+  for (size_t i = 0; i < HIDDEN_ITEMS_COUNT; i++) {
+    if (name.equals(HIDDEN_ITEMS[i])) return true;
+  }
+
+  return false;
```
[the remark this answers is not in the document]
dexif: fixed

src/network/WebDAVHandler.cpp
```diff
@@ -0,0 +1,754 @@
...
+      file.write(buf, bytesRead);
+      remaining -= bytesRead;
+    }
+  }
+
+  file.close();
+  clearEpubCacheIfNeeded(path);
+
+  _server->send(existed ? 204 : 201);
+  LOG_DBG("DAV", "PUT complete: %s (%d bytes)", path.c_str(), contentLength);
```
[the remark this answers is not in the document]
dexif: fixed

dexif:
I'll open a new pull request. It will include library updates and a webdav implementation. Just to make it a clean pull request.

Issue #998, opened 2026-02-19
Minor hyphenation and text flow issues


cbix:
### Affected Version

1.1.0-rc+402e887

### Describe the Bug

In some cases for 2 German epubs, I could observe text flow issues with hyphenation enabled.

**1. Missing expected word-break**
Not sure if this is expected, apparently there's (rightfully) a non-breaking space between _200_ and _Quadratkilometer_ but I don't know how to check which word breaks CrossPoint introduces here or how to inspect the underlying hyphenation trie. Is it possible the non-breaking space prevents the following word from breaking at all? Because that would actually be a software bug and not just explained through missing hyphenation data.

Additionally, I'm not sure if it's expected that the non-breaking space has a fixed small width while regular spaces stretch for justification.
<img width="480" height="800" alt="Image" src="https://github.com/user-attachments/assets/51ad6672-abd5-4155-95c2-0be52f4b398b" />

**2. Hyphenated word flows to next line**
This should not happen, the hyphen is inserted at the right place but the word then flows to the next line _including_ the hyphen.
<img width="480" height="800" alt="Image" src="https://github.com/user-attachments/assets/f16a00b0-ab06-48e6-b519-eee9a414f3d3" />

**3. Quoted word not hyphenated**
(See https://github.com/crosspoint-reader/crosspoint-reader/issues/998#issuecomment-3940533510)

### Steps to Reproduce

Due to the epubs' copyright I unfortunately can't provide them publicly, maybe this issue can serve as a place to collect further examples from public domain books where this happens.
(Update: html excerpts in https://github.com/crosspoint-reader/crosspoint-reader/issues/998#issuecomment-3934584552)

### Expected Behavior

For comparison, here's how these paragraphs look in landscape mode:

<img width="800" height="480" alt="Image" src="https://github.com/user-attachments/assets/49d97a3b-23ec-4acb-84f1-ad01860ac269" />
<img width="800" height="480" alt="Image" src="https://github.com/user-attachments/assets/125045ba-913a-4ecb-b80d-3e5a73f979ec" />

### Relevant Log Output/Screenshots

```shell
(No valuable info in logs, screenshots are pasted above)
```

jpirnay:
"Qaudatkilometer" should have proper hyphenation, "Ölinfrastruktur" might suffer from a bug in the hypenation logic that would treat composed characters -  instead of "Ö" an "O&#x308;" (an O followed by the combining diaeresis U+0308) - incorrectly.
You can see this here

[...]

PR #1037, opened 2026-02-20
fix: Fix hyphenation and rendering of decomposed characters

jpirnay:
## Summary

* This PR fixes decomposed diacritic handling end-to-end:
  - Hyphenation: normalize common Latin base+combining sequences to precomposed codepoints before Liang pattern matching, so decomposed words hyphenate correctly
  - Rendering: correct combining-mark placement logic so non-spacing marks are attached to the preceding base glyph in normal and rotated text rendering paths, with corresponding text-bounds consistency updates.
  - Hyphenation around non breaking space variants have been fixed (and extended)
  - Hyphenation of terms that already included of hyphens were fixed to include Liang pattern application (eg "US-Satellitensystem" was *exclusively* broken at the existing hyphen)

## Additional Context

* Before

* After
fix1
fix2

* Note 1: the hyphenation fix is not a 100% bullet proof implementation. It adds composition of *common* base+combining sequences (e.g. O + U+0308 -> Ö) during codepoint collection. A complete solution would require implementing proper Unicode normalization (at least NFC, possibly NFKC in specific cases) before hyphenation and rendering, instead of hand-mapping a few combining marks. That was beyond the scope of this fix.

* Note 2: the render fix should be universal and not limited to the constraints outlined above: it properly x-centers the compund glyph over the previous one, and it uses at least 1pt of visual distance in y.

Before:
Image

After: 
Image

* This should resolve the issues described in #998 
---

### AI Usage

### DIFF
--- a/lib/EpdFont/EpdFont.cpp
+++ b/lib/EpdFont/EpdFont.cpp
@@ -4,17 +4,6 @@
 
 #include <algorithm>
 
-namespace {
-
-bool isCombiningMark(const uint32_t cp) {
-  return (cp >= 0x0300 && cp <= 0x036F)      // Combining Diacritical Marks
-         || (cp >= 0x1DC0 && cp <= 0x1DFF)   // Combining Diacritical Marks Supplement
-         || (cp >= 0x20D0 && cp <= 0x20FF)   // Combining Diacritical Marks for Symbols
-         || (cp >= 0xFE20 && cp <= 0xFE2F);  // Combining Half Marks
-}
-
-}  // namespace
-
 void EpdFont::getTextBounds(const char* string, const int startX, const int startY, int* minX, int* minY, int* maxX,
                             int* maxY) const {
   *minX = startX;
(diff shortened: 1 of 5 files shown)


jpirnay:
Testfile: 
[hyphenation_demo.zip](https://github.com/user-attachments/files/25443728/hyphenation_demo.zip)

lukestein:
Thank you very much for this! Tagging in our hyphenation expert @osteotek. I've been really hoping we could get the fix that this promises:
> Hyphenation of terms that already included of hypehns were fixed to include Liang pattern application (eg "US-Satellitensystem" was exclusively broken at the existing hyphen)

I..e., allowing additional hyphenation of words that also include an explicit hyphen so that the occasional multiply-hyphenated-phrases have more places to get broken. (I use large or extra-large fonts, so I'm particularly sensitive to hyphenation quality.)

osteotek:
Ok, since popular opinion seems to be in favor of breaking already hyphenated words, i'm OK with it too 😃

Issue #902, opened 2026-02-15
[feature request] xsmall font


alveitia:
Would love to have xsmall font like the offering from papyrix firmware.

jonasdiemer:
Should be addressed with #930

ariel-lindemann:
@jonasdiemer, the Issue you opened groups a lot of loosely related issues regarding fonts. 

The request here is much more focued and specific and can be adressed independently of what you wrote in yours.

The font size isn't even mentioned in #930, so I don't see how it can be a duplicate. Or if it's a duplicate of some other issue could you please refer to that so we don't misunderstand?

Otherwise I think it's best to reopen this. Could you do that please? I don't have the permissions for that.

hedgepips:
> Would love to have xsmall font like the offering from papyrix firmware.

This would be awesome! I hope it is considered for future update

jonasdiemer:
> [@jonasdiemer](https://github.com/jonasdiemer), the Issue you opened groups a lot of loosely related issues regarding fonts.
> 
> The request here is much more focued and specific and can be adressed independently of what you wrote in yours.
> 
> The font size isn't even mentioned in [#930](https://github.com/crosspoint-reader/crosspoint-reader/issues/930), so I don't see how it can be a duplicate. Or if it's a duplicate of some other issue could you please refer to that so we don't misunderstand?
> 
> Otherwise I think it's best to reopen this. Could you do that please? I don't have the permissions for that.

Thanks for these comments, let me clarify:

1. Font sizes are "hard coded" into the pre-rendered font. So in order to add a font size, we'd need to internally add a separate copy of each of the fonts. 
2. Hence, adding a size is basically like adding a font. Let me also clarify this in #930
3. We're already maxed out the flash storage (of the current partitioning scheme). So adding fonts or sizes is no longer straight-forward.

Hence I decided to link all requests that ask for new fonts to #930 because other than juggling around fonts (e.g. replace an existing font / size with another one) - which is completely personal preference - we have no option than implement user-loadable fonts.

ariel-lindemann:
> > [@jonasdiemer](https://github.com/jonasdiemer), the Issue you opened groups a lot of loosely related issues regarding fonts.
> > The request here is much more focued and specific and can be adressed independently of what you wrote in yours.
[...]

Issue #863, opened 2026-02-12
Font face, type and size request


gibblett06:
I know sizes are limited because the device itself is small. These are a few additions that could help people with minor vision issues. 

- a larger size font
- the ability to bold the font
- font: atkinson hyperlegible or something similar

Currently reading on bookerly at the largest size. I'm not concerned about the UI font at all, just for the reader. Thanks!

flbn:
hell yea. would be nice but the opposite for me. the letters are too big!

jonasdiemer:
We need to implement user-fonts to address this.

jonasdiemer:
Should be addressed with #930

Issue #398, opened 2026-01-16
Extra Small font size


testeriosandandroid1-boop:
Can someone please implement a smaller font size than the current one?
I would like something a bit more compact than the current Small font.
Thank you.

reukiodo:
IMHO, the whole UI could be quite a bit more compact and give more attention to the book rather than the rest of the UI

testeriosandandroid1-boop:
Agreed. I hoped to see the extra small font option would be implemented on v0.15, but I don't see it. Maybe in the next version?

daveallie:
It may be one more version after the next, we're butting up against a hard(ish) limit for the firmware size, I have some changes to fonts coming which will lower the flash usage a lot and will add the small fonts as part of that

reukiodo:
I have been trying to find the partition layout, to maybe look at ways to change this a bit... I think a good goal would be to get a single scalable universal unicode font (like https://unifoundry.com/unifont/index.html ) into a separate partition, while having the app0/app1 for A/B OTA firmware updates. The firmware then only containing the compiled code. Additional fonts could be stored and loaded from the uSD card, where the universal unicode font would be default/fallback font when custom fonts aren't available.

reukiodo:
Another way to tackle this is repartition for a small basic recovery bootloader, in its own partition, which does a few things:

- display a boot screen (X4)
- checks uSD card for firmware image, if present, show progress bar and write it to SPI flash's firmware partition, verify, delete the firmware image from uSD, and reboot
- try to boot from firmware partition, and if fails, display error screen and initialize USB for firmware flashing

This would leave the rest of the SPI flash for the entire firmware - no app0/app1 A/B, just a full ~15MB SPI partition.

This could still be used in conjunction with separate (compressed?) font partition.

osteotek:
There are already forks / PRs that implement loading custom fonts from SD card, without partition shenanigans

reukiodo:
Even on small, it is nearly double what I use on phones with much larger screens:
![20260130_175907.jpg](https://github.com/user-attachments/assets/400eaff2-cf69-4f13-9748-ba4f9522dbb0)

DavidOrtmann:
But doesn't the massively smaller resolution become an issue here? Hard to compare both screen types.

reukiodo:
No? 480x800 is less than was common to use on computers ~30yrs ago with 640x480 being a very common resolution and displaying a lot more text. It was readable then and is still readable now.

[...]

PR #1043, opened 2026-02-20
feat: add extra small font sizes

ariel-lindemann:
## Summary

Add a smaller font size for each of the built in fonts, as requested in #398,  #902 and [here](https://github.com/crosspoint-reader/crosspoint-reader/issues/863#issuecomment-3902330878)

This is what it looks like for
Bookerly

image

The fonts themselves:
+ Bookerly size 10 
+ NotoSans size 10
+ OpenDyslexic size 6

And the appropriate menu entry

## Additional Context

The only thing still needed is some localization strings. I've done those of the languages i know:, but these still remain,
Tagging the translators from  [`translators.md`](https://github.com/crosspoint-reader/crosspoint-reader/blob/master/docs/translators.md)

- [ ] Czech (@brbla)
- [ ] Portuguese (@yagofarias)
- [x] Russian (@madebyKir or @mrtnvgr)
- [x] Spanish (@yeyeto2788 or @Skrzakk)
- [ ] Swedish (@dawiik)

---

### AI Usage

Did you use AI tools to help write this code? _**< PARTIALLY >**_ (mainly with the adjustments to `build_font_ids.sh`


lib/I18n/translations/english.yaml
```diff
@@ -173,6 +173,7 @@ STR_NEXT_PREV: "Next/Prev"
 STR_BOOKERLY: "Bookerly"
 STR_NOTO_SANS: "Noto Sans"
 STR_OPEN_DYSLEXIC: "Open Dyslexic"
+STR_X_SMALL: "X Small"
```
yeyeto2788, line 176: Shouldn't this also be added to all language files? Seems like it is missing on some. Maybe adding the English even on the Spanish should make it work right? Or is there a mechanism so in case it is missing it falls back to English?

yeyeto2788: In case you will be adding the string to other languages, the correct for Spanish would be `Extra pequeña` 😃

ariel-lindemann: yes, I intentonally left it out and it seems to fall back on the default.

Thanks for the translation! Will be adding it in

DARKFiB3R:
If only my eyes worked as well as they once did 🙁

I really wish there were even larger fonts.

I think more granular control of font size could really help with readability while also trying to avoid it displaying only 2 words per line 🤣

So if font size could increase by 0.5 (or even less), that would be great.

Is that something you could help with?

I have no idea what is actually involved to achieve this, so I do apologise if it's a lot to ask. 

I had presumed font size was arbitrary and wondered why the options were so limited. I mean, if someone wants XXL fonts, resulting in 1 word per line, they should be able to do that, right? I guess there is more to it than I thought.

ariel-lindemann:
@DARKFiB3R, that's definetely doable, but probably better to start an issue or discussion on that. I've started this PR because I personally would use the smaller text and it was also my sense from discussions and issues around the repo and on reddit/twitter that it is has been common wish for a while and now we have the opportunity to implement it because some space has been freed up.

The reason we have limited options is because we need an entirely new set of fonts (see the diff). This change is about 5% of the available storage, which is quite a lot.

So I will be concentrating on this for now, but if the idea for an XXL font gains enough traction that would be something to consider separately.

lib/I18n/translations/english.yaml
osteotek: Translation to all languages is not required from PRs, there will be fallback to english in case of missing translations.

ariel-lindemann: good to know. I wasn't sure what the consensus is. But thankful for the translations provided.

Besides that, the PR is ready. I've been using it on my device for the last two days without issue.

ariel-lindemann:
@daveallie @osteotek could we please get this merged in the near future?

I see that the gains we recently made with regards to space are being quickly eaten up by recent PRs (many newer than this) so I feel like the window is closing on this and I think this should have a certain priority given it has been requested so often and it is a direct imrovement of the core reading experience (the guiding principle according to `SCOPE.md`)

I have been testing it on my device since Friday and rebasing every day and I had no issues.

DARKFiB3R:
@ariel-lindemann

Thanks for the reply.

So a new font would be needed because fonts are created with a fixed set of sizes? They can't just scale like a vector image?

jonasdiemer:
> So a new font would be needed because fonts are created with a fixed set of sizes? They can't just scale like a vector image?

Correct, fonts are pre-rendered and baked in with fixed sizes at compile time.

znelson:
Closing due to inactivity. Please feel free to join the font discussion in #1693.

PR #1056, opened 2026-02-21
perf: font-compression improvements

adriancaruana:
## Purpose

This PR includes some preparatory changes that are needed for an upcoming performant CJK font feature. The changes have no impact on render time and heap allocation for latin text. **Despite this, I think these changes stand on their own as a better font compression/decompression implementation.**

## Summary

 - Font decompressor rewrite: Replaced the 4-slot LRU group cache with a two-tier system — a page buffer (glyphs prewarmed before rendering begins) and a hot-group fallback (last decompressed group retained for non-prewarmed      
  glyphs). 
  - Byte-aligned compressed bitmap format: Glyph bitmaps within compressed groups are now stored row-padded rather than tightly packed before DEFLATE compression, improving compression ratios by making identical pixel rows produce 
  identical byte patterns. Glyphs are compacted back to packed format on demand at render time. Reduces flash size by 155 KB.
  - Page prewarm system: Added `Page::collectText` and `Page::getDominantStyle` to extract per-style glyph requirements before rendering, and `GfxRenderer::prewarmFontCache` to pre-decompress only the groups needed for the dominant style
   — eliminating mid-render decompression for the common case.
  - UTF-8 robustness fixes: `utf8NextCodepoint` now validates continuation bytes and returns a replacement glyph on malformed input; `ChapterHtmlSlimParser` correctly preserves incomplete multi-byte sequences across word-buffer flush
  boundaries rather than splitting them.

---

### AI Usage

Did you use AI tools to help write this code? _**YES**_ Architecture and design was done by me, refined a bit by Claude. Code mostly by Claude, but not entirely.

### DIFF
--- a/lib/GfxRenderer/FontCacheManager.cpp
+++ b/lib/GfxRenderer/FontCacheManager.cpp
@@ -13,14 +13,18 @@ void FontCacheManager::clearCache() {
   if (fontDecompressor_) fontDecompressor_->clearCache();
 }
 
-void FontCacheManager::prewarmCache(int fontId, const char* utf8Text, EpdFontFamily::Style style) {
+void FontCacheManager::prewarmCache(int fontId, const char* utf8Text, uint8_t styleMask) {
   if (!fontDecompressor_ || fontMap_.count(fontId) == 0) return;
-  const EpdFontData* data = fontMap_.at(fontId).getData(style);
-  if (!data || !data->groups) return;
-  int missed = fontDecompressor_->prewarmCache(data, utf8Text);
-  if (missed > 0) {
-    LOG_DBG("FCM", "prewarmCache: %d glyph(s) not cached for style %d; hot-group fallback in use", missed,
-            static_cast<int>(style));
+
+  for (uint8_t i = 0; i < 4; i++) {
+    if (!(styleMask & (1 << i))) continue;
+    auto style = static_cast<EpdFontFamily::Style>(i);
+    const EpdFontData* data = fontMap_.at(fontId).getData(style);
+    if (!data || !data->groups) continue;
+    int missed = fontDecompressor_->prewarmCache(data, utf8Text);
+    if (missed > 0) {
+      LOG_DBG("FCM", "prewarmCache: %d glyph(s) not cached for style %d", missed, i);
+    }
   }
 }
 
@@ -63,14 +67,14 @@ void FontCacheManager::PrewarmScope::endScanAndPrewarm() {
   manager_->scanMode_ = ScanMode::None;
   if (manager_->scanText_.empty()) return;
 
-  // Determine dominant style from scan counts
-  uint8_t dominantStyle = 0;
-  for (uint8_t i = 1; i < 4; i++) {
-    if (manager_->scanStyleCounts_[i] > manager_->scanStyleCounts_[dominantStyle]) dominantStyle = i;
+  // Build style bitmask from all styles that appeared during the scan
+  uint8_t styleMask = 0;
+  for (uint8_t i = 0; i < 4; i++) {
+    if (manager_->scanStyleCounts_[i] > 0) styleMask |= (1 << i);
   }
+  if (styleMask == 0) styleMask = 1;  // default to regular
 
-  manager_->prewarmCache(manager_->scanFontId_, manager_->scanText_.c_str(),
-                         static_cast<EpdFontFamily::Style>(dominantStyle));
+  manager_->prewarmCache(manager_->scanFontId_, manager_->scanText_.c_str(), styleMask);
 
   // Free scan string memory
   manager_->scanText_.clear();
--- a/lib/GfxRenderer/FontCacheManager.h
+++ b/lib/GfxRenderer/FontCacheManager.h
@@ -15,7 +15,7 @@ class FontCacheManager {
   void setFontDecompressor(FontDecompressor* d);
 
   void clearCache();
-  void prewarmCache(int fontId, const char* utf8Text, EpdFontFamily::Style style = EpdFontFamily::REGULAR);
+  void prewarmCache(int fontId, const char* utf8Text, uint8_t styleMask = 0x0F);
   void logStats(const char* label = "render");
   void resetStats();
 


ngxson:
Design-wise, I think these should be improved before we can implement further optimizations:
1. Because the font decompression cache is designed based on the assumption that we will predominantly use one single typeface per render pass, we should have a RAII-style API to indicate that "I'm gonna use this one single typeface for this screen"
2. We should not add too many patches to the epub logic to get the info for pre-warm, it will be a huge pain for future developers who want to extend the epub reader system. Instead, you should implement a "defer" API for `GfxRenderer`, something like this:

```cpp
renderer.startAsyncTextRender();
for (const auto word& : text) {
  renderer.drawText(fontId, ..., ...); // temporary track it, but does not actually draw
}
renderer.endAsyncTextRender(); // gather cache, pre-warm, etc, then draw, then finally clear cache
```

And with RAII style:

```cpp
{
  const auto async = renderer.async(); // will automatically stop async on out-of-scope
  for (const auto word& : text) {
    renderer.drawText(fontId, ..., ...); // temporary track it, but does not actually draw
  }
}
```

And even better, stack-allocated cache:

```cpp
GfxRenderer::AsyncScope GfxRenderer::async() {
  uint8_t cache[MAX_CACHE_SIZE];
  fontDecompressor.setCache(cache, MAX_CACHE_SIZE);
  ...
}
```

ngxson:
Btw, adding a dedicated comment for clarity: I'm doing some tests on https://github.com/crosspoint-reader/crosspoint-reader/pull/831 and I realized that cache is cleared once the page is finish rendering.

So, should we allocate the font cache on stack instead of heap?

adriancaruana:
@ngxson 

  1. RAII scope (const auto async = renderer.async()): Good idea                                                                                                                                                                                                                                                                                                                                                                                                                                                                                    
  2. Defer API (`drawText` tracks, `endAsync` draws + clears): Also a good idea. I didn't implement it quite the way you suggested with your pseudocode. ("gather cache, pre-warm, then draw, then clear cache" of `endAsyncTextRender()` doesn't work cleanly due to multipass grayscale rendering (BW, LSB, MSB) each with different `renderMode`, and non-text ops (images, underlines) also need to render in each pass.) Instead, `endScanAndPrewarm()` only prewarns. The caller then makes normal `page->render()` calls for each pass, all hitting the warm cache. 
  3. Stack-allocated cache (`uint8_t cache[MAX_CACHE_SIZE]`): Not feasible: render task stack is 8KB, cache buffers can be much larger (e.g., from 50 KB to 100+ KB). But the RAII scope improves the lifetime guarantees.

lib/GfxRenderer/GfxRenderer.cpp
```diff
@@ -225,6 +295,20 @@ void GfxRenderer::drawText(const int fontId, const int x, const int y, const cha
     return;
   }
 
+  if (deferredMode_ == DeferredMode::Scanning) {
+    scanText_ += text;
```
ngxson: I'm not sure how this will affect the performance, but I think in general, string concat with `+` operator should be avoided as it will create a copy of the string.

Instead, I think we can implement `scanText_` as a stack-allocated buffer of size, for example 1024 or 2048 bytes, much larger than the number of characters can show on the screen.

For the edge case where it's filled up, we can force doing rendering right away (i.e. flush the buffer)

src/activities/reader/EpubReaderActivity.cpp
```diff
@@ -695,20 +715,42 @@ void EpubReaderActivity::renderContents(std::unique_ptr<Page> page, const int or
...
 
     // Render and copy to MSB buffer
     renderer.clearScreen(0x00);
     renderer.setRenderMode(GfxRenderer::GRAYSCALE_MSB);
     page->render(renderer, SETTINGS.getReaderFontId(), orientedMarginLeft, orientedMarginTop);
     renderer.copyGrayscaleMsbBuffers();
+    const auto tGrayMsb = millis();
 
     // display grayscale part
     renderer.displayGrayBuffer();
```
ngxson, line 762: > 2\. Defer API (`drawText` tracks, `endAsync` draws + clears): Also a good idea. I didn't implement it quite the way you suggested with your pseudocode. ("gather cache, pre-warm, then draw, then clear cache" of `endAsyncTextRender()` doesn't work cleanly due to multipass grayscale rendering (BW, LSB, MSB) each with different `renderMode`, and non-text ops (images, underlines) also need to render in each pass.) Instead, `endScanAndPrewarm()` only prewarns. The caller then makes normal `page->render()` calls for each pass, all hitting the warm cache.

Can we force rendering when we encounter a `displayGrayBuffer()` ?

It's the same logic as "instruction buffer is full" in the context of my last comment.

It can also be triggered when non-text element is drawn. But that should not be a problem because in reality, we should always see a lot of consecutive calls to `drawText()`

src/activities/reader/TxtReaderActivity.cpp
```diff
@@ -421,7 +420,12 @@ void TxtReaderActivity::renderPage() {
...
   };
 
-  // First pass: BW rendering
+  // Deferred rendering: scan pass accumulates text, then prewarm, then real render
+  auto scope = renderer.deferTextRendering();
+  renderLines();  // scan pass — text accumulated, no drawing
+  scope.endScanAndPrewarm();
+
+  // BW rendering
   renderLines();
```
ngxson, line 404: A bit doubt if the 2 passes will negatively impact the performance, especially when `renderLines` can potentially do a lot of stuff under the hood (same for epub)

Instead, we should implement a set of "instructions" like this:

```cpp
class TextInstruction {
  EpdGlyph* glyph;
  uint16_t x;
  uint16_t y;
};
```

Then have a fixed, stack-allocated buffer:

```
class DeferredRenderScope {
  TextInstruction inst[1024];
  size_t instSize = 0;
};
```

Actual rendering will kick in if one of the conditions are met:
1. `instSize == 1024`
2. Or, going out of scope
3. Or, a non-text draw function is called

ngxson:
Btw, before you proceed further with my comments above, I think we should extract the defer render API to a dedicated API. I think it can already be useful for the font decompression version on `master`

src/activities/reader/TxtReaderActivity.cpp
ngxson: Also, I think the usage in the current form doesn't fit the definition of "defer" or "async"

The current implementation is more like `GatherMetricsScope`; True defer means that we accumulate instructions along the way, then execute them (or "flush") then in batch, as I suggested in the last comment

osteotek:
> utf8NextCodepoint now validates continuation bytes and returns a replacement glyph on malformed input;

this fix is discussed separately - https://github.com/crosspoint-reader/crosspoint-reader/pull/1134/

src/activities/reader/TxtReaderActivity.cpp
adriancaruana: I explored this direction carefully and hit three practical barriers:

  1. Stack allocation isn't feasible. The render task stack is 8 KB (Activity.cpp:22). 1024 instructions × 8 bytes = 8 KB (i.e., the entire stack). Even 256 entries (2 KB) is risky with the deep call chains through `page->render()` -> `PageLine::render()` -> `TextBlock::render()` -> `drawText()`. This is the same reason the font cache itself can't be stack-allocated.
  2. Images bypass `GfxRenderer`'s draw methods. `ImageBlock::render()` writes pixels via `drawPixelWithRenderMode()` (in DitherUtils.h), calling `renderer.drawPixel()` per pixel — not through a single interceptable method like `drawBitmap()`. So the "flush when a non-text draw is called" trigger can't be detected inside `GfxRenderer` without either a) Adding a scan-mode check to drawPixel itself (called millions of times -- for every text glyph pixel too), or b) Adding awareness of deferred rendering into the epub layer, which is what we're trying to avoid.
  3. Multi-pass grayscale still needs `page->render()`. With anti-aliasing enabled, we call `page->render()` for BW, LSB, and MSB passes. Each pass renders images and underlines too (with different render modes), so `page->render()` is needed regardless. Record-and-replay only saves 1 of 4 passes (i.e., the scan) which is the cheapest pass (font map lookup + string append + early return per word, no pixel work).

src/activities/reader/EpubReaderActivity.cpp
adriancaruana: See point 3 [here](https://github.com/crosspoint-reader/crosspoint-reader/pull/1056#discussion_r2843364357).

lib/GfxRenderer/GfxRenderer.cpp
adriancaruana: `std::string` concat using `s = s + i` [does a copy](https://en.cppreference.com/w/cpp/string/basic_string/operator+), but `s += i` [does not](https://en.cppreference.com/w/cpp/string/basic_string/operator+=) -- it's O(1), and grows geometrically. `reserve(x)` seems worse to me in two ways: Will often alloc more than is necessary and adds complexity about what to do when `x` is exhausted. The abstraction seems genuinely useful here.

I prefer to keep this as-is.

src/activities/reader/TxtReaderActivity.cpp
ngxson: Hmm ok, agree that we cannot get away with true deferring logic due to memory constraint.

I'm now having another concern though, because the epub/txt reader now have to do a prewarm, I wonder how much time it takes to run the prewarm, compared to the time it takes if we don't do a prewarm then miss the cache.

To illustrate what I mean, without prewarm:

```cpp
page->render(...); // may result in a lot of cache miss
```

And with prewarm:

```cpp
auto scope = renderer.createFontPrewarmScope();
page->render(renderer, SETTINGS.getReaderFontId(), orientedMarginLeft, orientedMarginTop);  // scan pass
scope.endScanAndPrewarm();

page->render(...);
// less cache miss now, but we called page->render() twice 
```

This is the main reason why I wanted to do full defer at first, because `page->render()` may invoke sdcard read, which can be much slower than a cache miss in font rendering

adriancaruana: That's a valid concern, but it's not an issue. With the pre-warm, `page->render()` doesn't access the SD Card. For `TxtReaderActivity`, all text is loaded from memory via `loadPageAtOffset()` which happens _before_ `renderPage()`. It's a similar story for `EpubReaderActivity`, with the exception of image rendering, which can read from the SD Card during `page->render()`, but _not_ with a prewarm-scoped renderer.

I hope these benchmarking stats I compiled dispel your concerns about the prewarm overhead.

Phase | Master | New branch
-- | -- | --
prewarm | -- | 5-7 ms
bw_render | 50-67 ms | 34-49 ms
display | 442 ms | 442 ms
gray_lsb | 39-43 ms | 33-43 ms
gray_msb | 40-45 ms | 35-46 ms
gray_display | 63 ms | 62-63 ms
total | 664-674 ms | 632-669 ms

---

As for whether the prewarm logic could be embedded into the first render pass to avoid the extra complexity: it can't, without a performance cost. With the font decompressor only holding one decompressed glyph group in memory at a time (minimises heap), rendering a page with glyphs from multiple groups without a prewarmed cache causes repeated decompression of the same groups as the single hot slot thrashes. The scan pass is cheap (just string concatenation, no I/O or decompression) and lets the prewarm batch-extract all needed glyphs upfront, which is why `bw_render` is consistently faster on the new branch. 

While the overall performance  in these benchmarks from latin text isn't much improved with these changes, for CJK text the difference is huge (~500ms to 1s).

ngxson: Hmm ok. I'm working on a fallback logic on https://github.com/crosspoint-reader/crosspoint-reader/pull/1163 and just realized that now if I mix the fallback from non-compressed fonts (UI fonts) to compressed fonts, there will be a big performance gap as we cannot do a prewarm pass for UI rendering. The use case was that we can use Ubuntu as main typeface for UI, and fallback to Noto Sans whenever we encounter a glyph that ubuntu doesn't support.

The fallback logic could also be useful in the future if we want to mix between a CJK font on sdcard, and built-in fonts in firmware (same page can use both fonts for example). Though I could be wrong, because probably the sdcard font can already contain everything we need to render a complete page.

With this warmup logic, do you think it's still relevant to have runtime fallback? Or you think we should not do fallback at all? Also, do you think the warmup can potentially help in the case where font is stored in sdcard?

lib/GfxRenderer/GfxRenderer.cpp
ngxson: I don't quite concern about the performance, but what I concern more is that string reallocation can potentially fragment the heap

ngxson: Otherwise, maybe just add `reserve(1024)` here?

src/activities/reader/TxtReaderActivity.cpp
adriancaruana: > The use case was that we can use Ubuntu as main typeface for UI, and fallback to Noto Sans whenever we encounter a glyph that ubuntu doesn't support.

In this case, if a pre-warm is truly not possible (it might be very impractical if UI rendering methods are stateful and have side effects; though this can always be solved with refactoring), then we might have to accept degraded rendering performance. This could be mitigated with a multi-slot LRU cache (like on master), but I think this would be a problem for another PR.

> With this warmup logic, do you think it's still relevant to have runtime fallback? Or you think we should not do fallback at all?

I think we should always have a fallback, since the UX of a slower render is better than rendering replacement glyphs.

> Also, do you think the warmup can potentially help in the case where font is stored in sdcard?

Definitely it will help when the fonts are on the SD Card. The main benefit of the prewarm is that is allows you to have the best of both worlds when it comes to: compressed font glyph groups, minimum heap overhead, and fast rendering, since each group that's needed does not need to be decompressed more than once, and each rendered glyph hits a warm cache. So, even if it is slower to retrieve the compressed font glyphs from the SD Card, the impact is minimised since there's no duplicated reads for the same group.

ngxson: Sounds good then, I think a LRU cache is still needed, but that can be considered later on.

Re. the ubuntu fallback --> noto sans, I'm thinking about another direction is to get rid of the ubuntu font altogether, and just use noto sans for UI. Mainly because ubuntu doesn't support multi language and maybe on the way doing so, we could finally support CJK inside the firmware. But I'll push a dedicated discussion for this idea. For now, the conclusion is: fallback is still needed, but less priority than sdcard fonts.

lib/EpdFont/FontDecompressor.cpp
```diff
@@ -1,134 +1,476 @@
...
+    stats.cacheMisses++;
+    const EpdFontGroup& group = fontData->groups[groupIndex];
+
+    free(hotGroupData);
+    hotGroupData = static_cast<uint8_t*>(malloc(group.uncompressedSize));
+    if (!hotGroupData) {
+      LOG_ERR("FDC", "Failed to allocate %u bytes for hot group %u", group.uncompressedSize, groupIndex);
+      hotGroupFont = nullptr;
+      hotGroupIndex = UINT16_MAX;
+      hotGroupSize = 0;
```
ngxson: IMO it's better to use `std::vector<uint8_t>` for `hotGroup`

- `malloc` --> `hotGroup.reserve()` or `resize()` depending on how it's used
- `hotGroupData` --> `hotGroup.data()`
- `hotGroupSize` --> `hotGroup.size()`

And maybe wrap it inside `std::optional` if you need to check if it's available or not

lib/EpdFont/FontDecompressor.cpp
```diff
@@ -1,134 +1,476 @@
...
+    stats.cacheHits++;
   }
 
-  // Cache miss - decompress
-  entry = findEvictionCandidate();
-  if (!decompressGroup(fontData, groupIndex, entry)) {
+  // Compact just the requested glyph from byte-aligned data into scratch buffer
+  if (glyph->dataLength > hotGlyphBufSize) {
+    free(hotGlyphBuf);
+    hotGlyphBuf = static_cast<uint8_t*>(malloc(glyph->dataLength));
```
ngxson: same for other buffers, can be std::vector

src/activities/reader/EpubReaderActivity.cpp
```diff
@@ -653,11 +653,29 @@ void EpubReaderActivity::saveProgress(int spineIndex, int currentPage, int pageC
...
                                         const int orientedMarginRight, const int orientedMarginBottom,
                                         const int orientedMarginLeft) {
+  const auto t0 = millis();
+  renderer.resetFontStats();
+
+  // Font prewarm: scan pass accumulates text, then prewarm, then real render
+  const uint32_t heapBefore = esp_get_free_heap_size();
+  auto scope = renderer.createFontPrewarmScope();
+  page->render(renderer, SETTINGS.getReaderFontId(), orientedMarginLeft, orientedMarginTop);  // scan pass
+  scope.endScanAndPrewarm();
```
ngxson:
```suggestion
{
  FontPrewarmScope scope(renderer);
  page->render(renderer, SETTINGS.getReaderFontId(), orientedMarginLeft, orientedMarginTop);  // scan pass
  // FontPrewarmScope deconstructor automatically trigger endScanAndPrewarm
}
```

lib/GfxRenderer/GfxRenderer.h
```diff
@@ -1,10 +1,13 @@
 #pragma once
 
 #include <EpdFontFamily.h>
-#include <FontDecompressor.h>
 #include <HalDisplay.h>
 
+class FontDecompressor;
```
ngxson: IMO it will be a bit cleaner if `FontDecompressor` become singleton and can be accessed from anywhere. I done that in https://github.com/crosspoint-reader/crosspoint-reader/pull/1163 , but you can copy it here

adriancaruana: I agree this would be good, but I think this should be out of scope for this PR. We can add this later. 😄

ngxson (approved):

adriancaruana:
Am I able to get a second review here? @osteotek @daveallie

ngxson:
Cc @znelson too if you have a bit of time. Let's try to unblock this so that @adriancaruana can proceed with the SD card font support

lib/Utf8/Utf8.cpp
```diff
@@ -13,20 +13,46 @@ uint32_t utf8NextCodepoint(const unsigned char** string) {
...
   if (bytes == 1) {
+    (*string)++;
     return chr[0];
   }
 
+  // Validate continuation bytes before consuming them
+  for (int i = 1; i < bytes; i++) {
+    if ((chr[i] & 0xC0) != 0x80) {
+      // Missing or invalid continuation byte — skip only the lead byte
+      (*string)++;
```
znelson: This should probably be:

```suggestion
      *string += bytes;
```

Without that, we'd replace every byte in a bad UTF-8 multi-byte sequence with a replacement glyph.

I had similar UTF-8 improvements in #1134. Yours here is a little more thorough. 

In my opinion, we should separate core UTF-8 handling changes from other changes, for independent review.

lib/Epub/Epub/parsers/ChapterHtmlSlimParser.cpp
```diff
@@ -744,9 +744,44 @@ void XMLCALL ChapterHtmlSlimParser::characterData(void* userData, const XML_Char
...
+    // If we're about to run out of space, then cut the word off and start a new one.
+    // For CJK text (no spaces), this is the primary word-breaking mechanism.
+    // We must avoid splitting multi-byte UTF-8 sequences across word boundaries,
+    // otherwise the trailing bytes become orphaned continuation bytes that the
+    // decoder can't interpret.
     if (self->partWordBufferIndex >= MAX_WORD_SIZE) {
-      self->flushPartWordBuffer();
+      // Walk back past continuation bytes (10xxxxxx) to find the lead byte
+      // of the last character in the buffer
+      int leadPos = self->partWordBufferIndex - 1;
```
znelson: All of this UTF-8 logic should be extracted to functions in Utf8.cpp.

lib/GfxRenderer/GfxRenderer.cpp
```diff
@@ -9,12 +10,82 @@ const uint8_t* GfxRenderer::getGlyphBitmap(const EpdFontData* fontData, const Ep
...
+
+void GfxRenderer::logFontStats(const char* label) {
+  if (fontDecompressor) fontDecompressor->logStats(label);
+}
+
+void GfxRenderer::resetFontStats() {
+  if (fontDecompressor) fontDecompressor->resetStats();
+}
+
+// --- FontPrewarmScope implementation ---
```
znelson: Let's move this out of GfxRenderer.

lib/Epub/Epub/parsers/ChapterHtmlSlimParser.cpp
znelson: FWIW: #1380 needs the same UTF-8 safe trimming operation

lib/GfxRenderer/GfxRenderer.cpp
adriancaruana: This is a good idea, I'm happy to take this on.

adriancaruana:
@ngxson @znelson I plan to make no further changes to this branch unless there are some showstoppers. I've described the two recent changes below

1. I decided to extract the font caching from the renderer before merging, mainly so that the changes in the SD card font branch stem from this better implementation rather than something that will need to be unpicked later. See https://github.com/crosspoint-reader/crosspoint-reader/pull/1056/commits/ca2e876603da1bdd6849e67f68600583b0ae1af7. 
2. While dogfooding the SD card branch on my commute this morning, I noticed an issue with the current 'dominant style' approach of the prewarm caching -- it's not very good if the page contains a 51-49 split between two styles (e.g., regular and italic). Upon reflection, I think it's best to simply prewarm all glyphs that appear on the page regardless of which style is dominant. See https://github.com/crosspoint-reader/crosspoint-reader/pull/1056/commits/a295c4b00ae272bf9d171e46c2b0a23fd3d62a53.

znelson (approved):

lib/Utf8/Utf8.cpp
```diff
@@ -31,8 +31,8 @@ uint32_t utf8NextCodepoint(const unsigned char** string) {
   // Validate continuation bytes before consuming them
   for (int i = 1; i < bytes; i++) {
     if ((chr[i] & 0xC0) != 0x80) {
-      // Missing or invalid continuation byte — skip only the lead byte
-      (*string)++;
+      // Missing or invalid continuation byte — skip all bytes consumed so far
+      *string += i;
```
znelson, line 35: Should this be `bytes` instead of `i`?
```suggestion
      *string += bytes;
```
(I'm fine with fixing edge cases in followup PRs instead of revising this one further.)

adriancaruana: No, I think it should be `i`. When the loop fails at position `i`, bytes `0` through `i-1` are the broken sequence (lead + validated continuations), and byte `i` is something else entirely -- possibly a valid lead byte. Skipping `i` bytes consumes exactly the broken part; skipping bytes would silently swallow unexamined bytes that may start new valid sequences.

znelson:
@adriancaruana @osteotek @ngxson I'm noticing that this change is really problematic for diacritic-heavy text. I'm working with the [Czech prayer book](https://stahuj.kancional.cz/e-kniha/kancional.epub) linked in #383 to improve some ToC navigation shortcomings.

That [kancional.epub](https://stahuj.kancional.cz/e-kniha/kancional.epub) has heavy use of Czech diacritics (á, č, ď, é, ě, í, ň, ó, ř, š, ť, ú, ů, ý, ž). Page turns take 3-5 seconds with anti-aliasing enabled.

The logs show:

```
[bw_render] hits=327 misses=102 (76.2% hit rate)
[bw_render] decompress=963ms groups_accessed=0
getBitmap: 429 calls, 990184us total, 2308us/call avg
```

groups_accessed=0 means no DEFLATE decompression is happening during render: the hot group is already loaded. The 963ms labeled "decompress" is entirely spent in compactSingleGlyph, doing per-pixel bit-shuffling from the byte-aligned format back to packed format. Each of the 102 cache misses costs ~2.3ms of pure CPU.

This happens 3 times per page (BW + grayscale LSB + grayscale MSB), so compaction alone adds ~900ms × 3 = ~2.7 seconds per page turn.

With the changes in #1449 reverting back to the earlier font compression behavior, it goes back to < 1 second per page turn.

adriancaruana:
Thanks for flagging @znelson. I'll take a look and prepare a fix.

adriancaruana:
@znelson Fix is up here, please take a look when you get a chance: https://github.com/crosspoint-reader/crosspoint-reader/pull/1451. FYI, I believe the issue with that Czech book was **not** the diatrics, but that most pages contained multiple styles (e.g., regular + bold).

PR #1058, opened 2026-02-21
fix: Correct bottom space for "Book Bar Only"

### LINKED ISSUE #1024: Bug: `Progress Bar Only` adds empty space
### 1.1.0-rc+8f33bee

When the setting `Progress Bar Only` is selected, one line disappears and creates blank space, although the progress bar takes up a minimal number of pixels

Image Image

znelson:
## Summary

Corrected bottom margin calculation so the "Book Bar Only" option doesn't reserve more space than it needs. Fixes #1024.

## Additional Context

There's a lot of duplication between TxtReaderActivity and EpubReaderActivity. We should factor out the common logic in a separate change.

---

### AI Usage


osteotek:
I believe https://github.com/crosspoint-reader/crosspoint-reader/pull/733 will bring similar fix

znelson:
Same fix in #733

PR #1059, opened 2026-02-21
feat: CJK Font Support

adriancaruana:
## Summary

Strawman: CJK Font Support (Not intended to be merged as-is).

## Additional Context

- A single CJK Font has been added (NotoSans CJK, 14pt Regular).
- Storing compressed groups of glyphs, grouped by occurence frequency
- Byte-aligning the rows of glyphs before compression, for better compression ratio
- Sensible caching after retrieving for rendering
- Fast decoding is a must (`miniz` > `uzlib`)

## Notes

- Note: This PR is stacked on: https://github.com/crosspoint-reader/crosspoint-reader/pull/1056, and builds on https://github.com/crosspoint-reader/crosspoint-reader/pull/831
- https://github.com/crosspoint-reader/crosspoint-reader/pull/951 is probably a pre-requisite to this. With a 15 MB app partition, we could support about 4 CJK fonts with ease.

## Some of the current limitations

- Even with compression, each CJK font still uses >2 MB of flash in compressed form. So, that's 2 MB per size and weight).
- Each page of CJK text contains more unique glyphs than latin text, so there's more decompression work to do per-page. This makes rendering a bit slower (about 30% slower in my testing) and adds more heap pressure (11 KB for the decompressor, and more depending on how many glyphs are needed to render the page).
- I had to remove some other fonts (about 400 KB worth) to get Noto Sans CJK in 14 pt to fit in the current firmware.

---

### AI Usage

Did you use AI tools to help write this code? **_YES_** Architecture and design was done by me, refined a bit by Claude. Code mostly by Claude, but not entirely.


ngxson:
> With a 15 MB app partition, we could support about 4 CJK fonts with ease.

Would appreciate if you can get any info on how much flash space is left after that.

Even with a bigger flash space to work with, we should still leave enough space for Latin-Ext and future features (explained in https://github.com/crosspoint-reader/crosspoint-reader/pull/951#issuecomment-3920105345)

Also, I think you mean 4 CJK variants, not fonts, right? Remember that to make it actually usable, you will need at least 6 combinations: `(serif, sans) x (bold, regular) x (12pt, 14pt)`, and maybe plus an extra for UI fonts

Uri-Tauber:
@adriancaruana Since CJK fonts take up a significant amount of space, it might make sense to store them outside of the main firmware. That way, users who don’t need Chinese/Japanese/Korean support wouldn’t be impacted by the additional size.

Perhaps the CJK fonts could be loaded into the (currently unused) SPIFFS partition, or handled in a similar way to #455? This could help keep the core firmware lighter while still providing full support for those who need it.

adriancaruana:
@ngxson 

> Would appreciate if you can get any info on how much flash space is left after that.
> 
> Even with a bigger flash space to work with, we should still leave enough space for Latin-Ext and future features (explained in https://github.com/crosspoint-reader/crosspoint-reader/pull/951#issuecomment-3920105345)
> 
> Also, I think you mean 4 CJK variants, not fonts, right? Remember that to make it actually usable, you will need at least 6 combinations: (serif, sans) x (bold, regular) x (12pt, 14pt), and maybe plus an extra for UI fonts
> 

No, I mean four fonts. See the first paragraph of https://en.wikipedia.org/wiki/Typeface.

[NotoSans CJK 14 pt is 2.2 MB](https://raw.githubusercontent.com/adriancaruana/crosspoint-reader/489fdec9eec9dba9075677a35b540afe02f9b58b/lib/EpdFont/builtinFonts/notosanscjk_14_regular.h). I compiled a firmware with NotoSans: 14pt Regular, 14pt Bold, 16pt Regular, and 16pt Bold. I don't think the X4 has the DPI to support 12pt CJK. They require: 2.2, 2.3, 2.5, and 2.7 MB respectively, **or 9.8 MB total**. I hope that helps.

@Uri-Tauber 

> @adriancaruana Since CJK fonts take up a significant amount of space, it might make sense to store them outside of the main firmware. That way, users who don’t need Chinese/Japanese/Korean support wouldn’t be impacted by the additional size.
> 
> Perhaps the CJK fonts could be loaded into the (currently unused) SPIFFS partition, or handled in a similar way to https://github.com/crosspoint-reader/crosspoint-reader/pull/455? This could help keep the core firmware lighter while still providing full support for those who need it.

Definitely this is also a possibility. The point I want to get across is that with font compression + extra flash space, I think we can add CJK support.

ngxson:
> [NotoSans CJK 14 pt is 2.2 MB](https://raw.githubusercontent.com/adriancaruana/crosspoint-reader/489fdec9eec9dba9075677a35b540afe02f9b58b/lib/EpdFont/builtinFonts/notosanscjk_14_regular.h). I compiled a firmware with NotoSans: 14pt Regular, 14pt Bold, 16pt Regular, and 16pt Bold. I don't think the X4 has the DPI to support 12pt CJK. They require: 2.2, 2.3, 2.5, and 2.7 MB respectively, **or 9.8 MB total**. I hope that helps.

@adriancaruana The latest build on `master` branch is ~4.2MB, with NotoSans SC, that will be 14MB in total, leaving 1MB for future features. Personally I think it can be quite risky to do this way, we should also leave some spaces for things like i18n, icons and hyphenation.

Also, as `fontconvert.py` indicates, this only contains core ideograph (chinese + taiwanese) + hangul (korean), there seems to be no more space for japanese hiragana/katakana? And btw vietnamese is not in CJK, we use [Latin-Ext](https://en.wikipedia.org/wiki/Vietnamese_alphabet), so the current implementation won't fit both vietnamese+chinese+japanese+korean. I think externally-loaded fonts will still be necessary.

Probably with both compression + external fonts, we can improve the speed a bit by allow caching more glyphs on RAM.

---

@Uri-Tauber The SPIFFS will be repurposed after https://github.com/crosspoint-reader/crosspoint-reader/pull/951 , basically it will be part of the distributed firmware, i.e. being the same `firmware.bin` to be flashed to device. Unfortunately, it will make the partition not swappable

adriancaruana:
@ngxson 

I believe they do contain Japanese, (see `--additional-intervals 0x3040,0x309F` (Hiragana) and `--additional-intervals 0x30A0,0x30FF` (Katakana) [here](https://github.com/adriancaruana/crosspoint-reader/blob/489fdec9eec9dba9075677a35b540afe02f9b58b/lib/EpdFont/scripts/convert-builtin-fonts.sh#L50)).

But yes you make a good point. It seems wasteful to allocate such a large proportion of the flash to a single typeface. I think it is becoming clear that if we want comprehensive support -- including many fonts and typefaces -- for non-latin text then we will have to lean on the SD Card. I'll start looking into this when I get a chance.

PR #1060, opened 2026-02-21
perf: optimize text rendering with byte-level framebuffer writes

jpirnay:
### Optimisation of glyph rendering for 1 bit bw fonts = UI - fonts
* Replace the per-pixel `drawPixel()` loop in the 1-bit BW text rendering path with direct byte-level framebuffer writes via `bitmapExtract()` +   `writeRowBits()`, eliminating `rotateCoordinates()`, bounds checking, and  OOB-logging overhead for every set pixel.
* Landscape orientations process 8 glyph pixels per loop iteration with purely sequential bitmap reads, yielding a 2.5–3.1× speedup.
* Portrait orientations use an 8×8 bit-matrix transpose (`transpose8x8`,  Warren "Hacker's Delight" §7-3) to convert sequentially-read glyph rows into   column bytes for physical-row writes, yielding a ~2× speedup.  
* PortraitInverted avoids a per-column `reverseBits8` by packing rows in reverse order before transposing.
* The fast path activates only for 1-bit fonts in BW render mode with no text rotation (`TextRotation::None`).  All other paths (2-bit/grayscale fonts, rotated text via `drawTextRotated90CW`) are unchanged.

### Optimisation of 2-bit text rendering
* Added non-rotated 2-bit fast glyph rendering path using chunked framebuffer mask writes.

### UI
10 x "The quick brown fox jumps over the lazy dog" with ubuntu 12 regular aka UI font

| Orientation              | Legacy  | Fast   | Speedup |
|--------------------------|---------|--------|---------|
| LandscapeCounterClockwise| 1757 µs | 561 µs |  3.1×   |
| LandscapeClockwise       | 1702 µs | 694 µs |  2.5×   |
| PortraitInverted         | 1750 µs | 883 µs |  2.0×   |
| Portrait                 | 1686 µs | 882 µs |  1.9×   |

### Reader
* Measured impact (Bookerly 14)
  - BW full-page render: 105679us -> 91609us (-13.3%)
  - BW+AA pipeline: 340114us -> 311612us (-8.4%)
  - LSB pass+copy: 106321us -> 100349us (-5.6%)
  - MSB pass+copy: 110823us -> 102216us (-7.8%)
  - Heap after benchmark unchanged (201444)
---

### AI Usage

Did you use AI tools to help write this code? _**YES **_

### DIFF
--- a/lib/GfxRenderer/GfxRenderer.cpp
+++ b/lib/GfxRenderer/GfxRenderer.cpp
@@ -376,30 +376,28 @@ static inline uint8_t get2BitPixel(const uint8_t* const bitmap, const int stride
   return get2BitPixel(bitmap, row * stride + col);
 }
 
-static inline uint8_t drawMaskFor2BitMode(const GfxRenderer::RenderMode mode) {
-  switch (mode) {
-    case GfxRenderer::BW:
-      return 0x0E;  // draw raw {1,2,3}
-    case GfxRenderer::GRAYSCALE_MSB:
-      return 0x06;  // draw raw {1,2}
-    case GfxRenderer::GRAYSCALE_LSB:
-    default:
-      return 0x04;  // draw raw {2}
-  }
+template <GfxRenderer::RenderMode mode>
+static constexpr uint8_t drawMaskFor2BitMode() {
+  if constexpr (mode == GfxRenderer::BW)
+    return 0x0E;  // draw raw {1,2,3}
+  else if constexpr (mode == GfxRenderer::GRAYSCALE_MSB)
+    return 0x06;  // draw raw {1,2}
+  else
+    return 0x04;  // GRAYSCALE_LSB: draw raw {2}
 }
 
 // 2-bit pipeline — fused gather+threshold (X axis): the 2-bit analog of extractGlyphBlock, but
 // gather and threshold are collapsed into one pass. The threshold (2-bit raw value → 1-bit on/off)
 // is information-lossy, so no contiguous 2-bit intermediate block can be formed mid-pipeline.
 // The resulting 1-bit mask feeds writeRowBits directly (scatter). build2BitColMask is the Y-axis counterpart.
+template <GfxRenderer::RenderMode mode>
 static inline uint8_t build2BitRowMask(const uint8_t* const bitmap, const int rowStartPixel, const int glyphXStartOrEnd,
-                                       const int count, const bool reverseXInChunk,
-                                       const GfxRenderer::RenderMode renderMode) {
+                                       const int count, const bool reverseXInChunk) {
   // drawMask uses raw 2-bit glyph values directly from font bitmaps:
   // raw 0=white, 1=light gray, 2=dark gray, 3=black.
   // Bit N set means: draw/update when raw==N.
-  // This avoids per-pixel remap (bmpVal = 3 - raw) and branch chains in the hot loop.
-  const uint8_t drawMask = drawMaskFor2BitMode(renderMode);
+  // Compile-time constant lets the compiler reduce (drawMask >> raw) & 1 to a single comparison.
+  constexpr uint8_t drawMask = drawMaskFor2BitMode<mode>();
 
   uint8_t mask = 0;
   for (int i = 0; i < count; i++) {
@@ -410,13 +408,64 @@ static inline uint8_t build2BitRowMask(const uint8_t* const bitmap, const int ro
   return mask;
 }
 
+// Fast-path 2-bit mask builder for 8 byte-aligned pixels.
+//
+// The 2-bit glyph bitmap stores 4 pixels per byte, MSB-first:
+//   byte b = [p0.msb p0.lsb  p1.msb p1.lsb  p2.msb p2.lsb  p3.msb p3.lsb]
+//
+// For each render mode the draw decision collapses to a two-bit boolean:
+//   BW            (draw if raw ≠ 0):      msb | lsb
+//   GRAYSCALE_MSB (draw if raw ∈ {1,2}):  msb ^ lsb
+//   GRAYSCALE_LSB (draw if raw == 2):     msb & ~lsb
+//
+// Derivation for one byte:
+//   msb_bits = b & 0xAA  →  bits 7,5,3,1 hold p0.msb … p3.msb; bits 6,4,2,0 = 0
+//   lsb_bits = (b & 0x55) << 1  →  same positions hold p0.lsb … p3.lsb
+//   draw_bits = msb_bits OP lsb_bits  →  bits 7,5,3,1 are the per-pixel draw flags
+//
+// compact4: squeezes those 4 draw flags from bit positions 7,5,3,1
+//   into the top nibble (bits 7,6,5,4 → pixels 0,1,2,3).
+//
+// Two bytes b0 (pixels 0–3) and b1 (pixels 4–7) are combined:
+//   mask = compact4(draw(b0)) | (compact4(draw(b1)) >> 4)
+//
+// This avoids the 8-iteration per-pixel loop in build2BitRowMask and
+// processes the full 8-pixel chunk in ~16 ALU ops instead of ~56.
+// The caller is responsible for only calling this when pixelStart is
+// 4-pixel (1-byte) aligned (pixelStart & 3 == 0) and count == 8.
+template <GfxRenderer::RenderMode mode>
+static inline uint8_t build2BitRowMaskFromTwoBytes(const uint8_t b0, const uint8_t b1) {
+  const uint8_t msb0 = b0 & 0xAA;
+  const uint8_t lsb0 = (b0 & 0x55) << 1;
+  const uint8_t msb1 = b1 & 0xAA;
+  const uint8_t lsb1 = (b1 & 0x55) << 1;
+
+  uint8_t draw0, draw1;
+  if constexpr (mode == GfxRenderer::BW) {
+    draw0 = msb0 | lsb0;
+    draw1 = msb1 | lsb1;
+  } else if constexpr (mode == GfxRenderer::GRAYSCALE_MSB) {
+    draw0 = msb0 ^ lsb0;
+    draw1 = msb1 ^ lsb1;
+  } else {  // GRAYSCALE_LSB
+    draw0 = msb0 & ~lsb0;
+    draw1 = msb1 & ~lsb1;
+  }
+
+  // Compact each nibble's draw flags from bit positions 7,5,3,1 → 7,6,5,4.
+  auto compact4 = [](const uint8_t d) -> uint8_t {
+    return (d & 0x80) | ((d & 0x20) << 1) | ((d & 0x08) << 2) | ((d & 0x02) << 3);
+  };
+  return compact4(draw0) | (compact4(draw1) >> 4);
+}
+
 // 2-bit pipeline — fused gather+threshold (Y axis): column-direction counterpart to build2BitRowMask.
 // Samples count pixels down glyph column glyphX starting at row glyphYStart; reverseRows implements
 // a negative-stride view along Y (reads bottom-to-top), needed for PortraitInverted.
+template <GfxRenderer::RenderMode mode>
 static inline uint8_t build2BitColMask(const uint8_t* const bitmap, const int glyphWidth, const int glyphX,
-                                       const int glyphYStart, const int count, const bool reverseRows,
-                                       const GfxRenderer::RenderMode renderMode) {
-  const uint8_t drawMask = drawMaskFor2BitMode(renderMode);
+                                       const int glyphYStart, const int count, const bool reverseRows) {
+  constexpr uint8_t drawMask = drawMaskFor2BitMode<mode>();
   uint8_t mask = 0;
   for (int i = 0; i < count; i++) {
     const int row = reverseRows ? (glyphYStart + count - 1 - i) : (glyphYStart + i);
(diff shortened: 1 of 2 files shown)


lib/GfxRenderer/GfxRenderer.cpp
```diff
@@ -61,6 +61,369 @@ static inline void rotateCoordinates(const GfxRenderer::Orientation orientation,
...
+                                const bool pixelState, const GfxRenderer::Orientation orientation,
+                                const GfxRenderer::RenderMode renderMode) {
+  // Non-rotated text fast path for 2-bit glyphs. Writes compact masks directly to framebuffer rows.
+  // TextRotation::Rotated90CW keeps the legacy per-pixel fallback path for safety and readability.
+  const bool writeState = (renderMode == GfxRenderer::BW) ? pixelState : false;
+  const uint8_t drawMask =
+      (renderMode == GfxRenderer::BW) ? 0x0E : ((renderMode == GfxRenderer::GRAYSCALE_MSB) ? 0x06 : 0x04);
+
+  switch (orientation) {
+    case GfxRenderer::LandscapeCounterClockwise: {
```
ngxson, line 511: Just wondering if we can template these cases for clarity.

I thought about the same idea (byte-aligned) a while back then. But I didn't proceed as it may make things too complicated for other maintainers.

This implementation seems to be roughly equivalent to doing tensor ops in machine learning, where each element is from `GF(2)`

The nested for loop `glyphY++` `glyphX += 8` (and reversed case below) can be considered as stride for each dimension, so on a quick look, it seems like we are using 3D tensors here.

jpirnay: - Great point. I avoided templating orientation to keep code size and maintenance complexity under control, since orientation is runtime-configurable and already split in one explicit switch for traceability.
- I did a readability update by replacing the ternary draw-mask expression with a named helper and expanded comments around raw 2-bit semantics.”
- If we still want templating, I’d scope it to ``RenderMode`` only (3 variants) and keep orientation runtime-switched; that gives some clarity/perf without exploding complexity.
```
template <GfxRenderer::RenderMode Mode>
static constexpr uint8_t drawMaskForMode() {
  if constexpr (Mode == GfxRenderer::BW) {
    return 0x0E;  // raw {1,2,3}
  } else if constexpr (Mode == GfxRenderer::GRAYSCALE_MSB) {
    return 0x06;  // raw {1,2}
  } else {  // GRAYSCALE_LSB
    return 0x04;  // raw {2}
  }
}

template <GfxRenderer::RenderMode Mode>
static inline uint8_t build2BitRowMaskT(const uint8_t* bitmap, int rowStartPixel, int glyphXStartOrEnd, int count,
                                        bool reverseXInChunk) {
  constexpr uint8_t drawMask = drawMaskForMode<Mode>();
  uint8_t mask = 0;
  for (int i = 0; i < count; i++) {
    const int logicalX = reverseXInChunk ? (glyphXStartOrEnd - i) : (glyphXStartOrEnd + i);
    const int pixelPosition = rowStartPixel + logicalX;
    const uint8_t byte = bitmap[pixelPosition >> 2];
    const uint8_t bitIndex = (3 - (pixelPosition & 3)) * 2;
    const uint8_t raw = static_cast<uint8_t>((byte >> bitIndex) & 0x3);
    if ((drawMask >> raw) & 0x01) {
      mask |= static_cast<uint8_t>(1u << (7 - i));
    }
  }
  return mask;
}

template <GfxRenderer::RenderMode Mode>
static void renderGlyphFast2BitT(uint8_t* frameBuffer, const uint8_t* bitmap, int glyphWidth, int glyphHeight,
                                 int screenXBase, int screenYBase, bool pixelState,
                                 GfxRenderer::Orientation orientation) {
  constexpr bool kIsBw = (Mode == GfxRenderer::BW);
  const bool writeState = kIsBw ? pixelState : false;

  switch (orientation) {
    case GfxRenderer::LandscapeCounterClockwise: {
      for (int glyphY = 0; glyphY < glyphHeight; glyphY++) {
        const int phyY = screenYBase + glyphY;
        if (phyY < 0 || phyY >= HalDisplay::DISPLAY_HEIGHT) continue;
        uint8_t* row = frameBuffer + phyY * HalDisplay::DISPLAY_WIDTH_BYTES;
        const int rowStartPixel = glyphY * glyphWidth;
        for (int glyphX = 0; glyphX < glyphWidth; glyphX += 8) {
          const int count = std::min(8, glyphWidth - glyphX);
          const uint8_t mask = build2BitRowMaskT<Mode>(bitmap, rowStartPixel, glyphX, count, false);
          if (mask == 0) continue;
          const int phyBitPos = screenXBase + glyphX;
          if (phyBitPos + count <= 0 || phyBitPos >= HalDisplay::DISPLAY_WIDTH) continue;
          writeRowBits(row, phyBitPos, mask, writeState);
        }
      }
      break;
    }
    // ...same orientation cases as today, but using build2BitRowMaskT<Mode>(...)
    default:
      break;
  }
}

// Runtime dispatch at callsite:
if constexpr (rotation == TextRotation::None) {
  switch (renderMode) {
    case GfxRenderer::BW:
      renderGlyphFast2BitT<GfxRenderer::BW>(renderer.getFrameBuffer(), bitmap, width, height,
                                            innerBase, outerBase, pixelState, renderer.getOrientation());
      break;
    case GfxRenderer::GRAYSCALE_MSB:
      renderGlyphFast2BitT<GfxRenderer::GRAYSCALE_MSB>(renderer.getFrameBuffer(), bitmap, width, height,
                                                       innerBase, outerBase, pixelState, renderer.getOrientation());
      break;
    case GfxRenderer::GRAYSCALE_LSB:
    default:
      renderGlyphFast2BitT<GfxRenderer::GRAYSCALE_LSB>(renderer.getFrameBuffer(), bitmap, width, height,
                                                       innerBase, outerBase, pixelState, renderer.getOrientation());
      break;
  }
  *cursorX += glyph->advanceX;
  return;
}
```

jpirnay:
> 🧹 Nitpick comments (1)
> 
> > lib/GfxRenderer/GfxRenderer.cpp (1)> `385-433`: **Consider transpose optimization for 2-bit portrait modes (optional future improvement).**
> > > The Portrait and PortraitInverted cases iterate column-by-column, which is less cache-friendly than the 8×8 transpose approach used in `renderGlyphFastBW`. This still provides speedup over per-pixel `drawPixel()`, but could be further optimized using a similar transpose technique adapted for 2-bit pixels.
> > > This is acceptable for now given the added complexity of transposing 2-bit data, but worth noting for future optimization if 2-bit portrait rendering becomes a bottleneck.

Agreed, and thanks for calling this out.
For 2-bit portrait paths we currently use a column-wise loop because it keeps the implementation understandable while still removing the heavy [drawPixel()](vscode-file://vscode-app/c:/Users/jpirn/AppData/Local/Programs/Microsoft%20VS%20Code/072586267e/resources/app/out/vs/code/electron-browser/workbench/workbench.html) overhead, which already gives a measurable speedup.

We considered a transpose-style approach (similar to the 1-bit path), but for packed 2-bit glyph data it adds significant complexity (extract/repack rules, orientation edge handling, and maintainability risk). Given current benchmark results, we chose the simpler/safer implementation for this PR.

We’ll treat a 2-bit transpose path as a follow-up optimization if profiling shows portrait 2-bit rendering as the next dominant bottleneck.

lib/GfxRenderer/GfxRenderer.cpp
ngxson: What I mean by saying templating is, for example, move the stride size to template:

```
template<int strideX, int strideY>
...
for (int glyphY = 0; glyphY < glyphHeight; glyphY += strideY) {
for (int glyphX = 0; glyphX < glyphWidth; glyphX += strideX) {
...
```

Even more advanced, but you can also put a function as template param: https://stackoverflow.com/questions/1174169/function-passed-as-template-argument . This is a very common practice when writing low-level tensor op kernel.

The reason why I suggest this is because as I explained earlier, tensor ops are pretty much based on stride. For example, in most tensor library, transposing a matrix is simply modifying the stride, no data done is involved until a `contigous` function is called.

By extracting it to more tensor-like ops, like `reshape`, `view`, etc, we may make it more obvious to look at, and potentially come up with an even better optimization in the future.

lib/GfxRenderer/GfxRenderer.cpp
```diff
@@ -61,6 +61,379 @@ static inline void rotateCoordinates(const GfxRenderer::Orientation orientation,
...
+          for (int n = 0; n < rowCount; n++, bitStart += glyphWidth) {
+            pack |= static_cast<uint64_t>(bitmapExtract(bitmap, bitStart, colCount)) << (56 - 8 * (rowCount - 1 - n));
+          }
+          pack = transpose8x8(pack);
+          for (int k = 0; k < colCount; k++) {
+            const uint8_t cols_k = static_cast<uint8_t>(pack >> (56 - 8 * k));
+            if (cols_k == 0) continue;
+            const int phyY = screenXBase + glyphX + k;
+            if (phyY < 0 || phyY >= HalDisplay::DISPLAY_HEIGHT) continue;
+            writeRowBits(frameBuffer + phyY * HalDisplay::DISPLAY_WIDTH_BYTES, phyBitPos, cols_k, pixelState);
```
ngxson: To illustrate my last comment: Here what you're doing in term of tensor ops is to get a contiguous 8x8 bits matrix from a view. It's better to extract into function like this:

```cpp
uint64_t viewContiguous(tensor, offsetX, offsetY);
```

ngxson: Same idea for the loop of `writeRowBits`, you're copying data from a contiguous 8x8 to a bigger (maybe non-contiguous) view of a bigger tensor; the bigger tensor is `frameBuffer` in this case

lib/GfxRenderer/GfxRenderer.cpp
jpirnay: @ngxson I did update the documentation and the structure to be as close to that analogy as possible
```
---------------------------------------------------------------------------
// Fast glyph render pipeline
// ---------------------------------------------------------------------------
// Both 1-bit (BW) and 2-bit (antialiased) paths share the same structure:
//
//   gather → [reindex] → scatter
//
// The glyph bitmap is a row-major 2D tensor [glyphHeight][glyphWidth].
// The framebuffer is a row-major 2D tensor [DISPLAY_HEIGHT][DISPLAY_WIDTH_BYTES]
// (1 bpp) with a fixed row stride of DISPLAY_WIDTH_BYTES bytes.
//
// Non-rotated (Landscape): glyph rows map 1-to-1 to framebuffer rows.
// Reindex is a no-op; the pipeline is a tight per-row gather+scatter loop.
//
// Rotated 90° (Portrait): glyph rows become framebuffer columns.
// A row↔column axis swap (reindex) is required before scattering.
//
// 1-bit pipeline
//   gather  : extractGlyphBlock        reads an 8×8 glyph tile into a
//                                      contiguous uint64_t block
//                                      (≈ glyphTensor[tile].contiguous())
//   reindex : transpose8x8             swaps row↔column axes in the uint64_t;
//                                      pure index transform, no data movement
//   scatter : scatterBlockToFrameBuffer → writeRowBits
//                                      writes each column-byte to its row
//
// 2-bit pipeline (why it differs)
//   The glyph stores 4 gray levels (0–3). Rendering reduces these to a 1-bit
//   draw/skip decision via a render-mode threshold. That reduction is
//   information-lossy, so gather and threshold cannot be separated — there is
//   no contiguous 2-bit block to transpose. The two steps are fused:
//
//   gather+threshold : build2BitRowMask  Landscape — samples along glyph X
//                      build2BitColMask  Portrait  — samples along glyph Y
//                      both return a 1-bit mask ready for writeRowBits
//   scatter          : writeRowBits      same atom as the 1-bit path
// ---------------------------------------------------------------------------
```
The reason there's no uint64_t intermediate for 2-bit is that the threshold (raw→1bit) is information-lossy — so a pure bitwise 2D block copy followed by transpose isn't possible. The gather and reduce must happen together. This is the actual constraint that prevents the 2-bit portrait case from getting the 8×8 block transpose optimization that the 1-bit case get

jpirnay:
The synthetic benchmark provides now:
[853] [DBG] [MAIN] Running render character benchmark
[853] [INF] [BENCH] === renderChar benchmark ===
[853] [INF] [BENCH] Text: "The quick brown fox jumps over the lazy dog", 10 iterations per orientation
[853] [INF] [BENCH] --- 1-bit (BW fast path vs legacy per-pixel) ---
[879] [INF] [BENCH]   1-bit Portrait                legacy: 1665 us   fast: 980 us   speedup: 1.7x
[907] [INF] [BENCH]   1-bit PortraitInverted        legacy: 1735 us   fast: 925 us   speedup: 1.9x
[931] [INF] [BENCH]   1-bit LandscapeClockwise      legacy: 1720 us   fast: 744 us   speedup: 2.3x
[955] [INF] [BENCH]   1-bit LandscapeCounterClockwise  legacy: 1756 us   fast: 593 us   speedup: 3.0x
[955] [INF] [BENCH] --- 2-bit (antialiased fast path ---
[1026] [INF] [BENCH]   2-bit Portrait                legacy: 4616 us   fast: 2460 us   speedup: 1.9x
[1086] [INF] [BENCH]   2-bit PortraitInverted        legacy: 3423 us   fast: 2467 us   speedup: 1.4x
[1149] [INF] [BENCH]   2-bit LandscapeClockwise      legacy: 3523 us   fast: 2777 us   speedup: 1.3x
[1212] [INF] [BENCH]   2-bit LandscapeCounterClockwise  legacy: 3596 us   fast: 2705 us   speedup: 1.3x

Issue #1066, opened 2026-02-21
Some epubs crash Crosspoint


williamsdb:
### Affected Version

1.0.0

### Describe the Bug

I have a series of ePubs in the MIT Press Essential Knowledge series which, on opening, cause Crosspoint to restart. These are not DRM protected and open fine elsewhere. Would love to be able to read these on the X4. 

### Steps to Reproduce

1. unzip attached file: [happiness_mit.epub.zip](https://github.com/user-attachments/files/25460001/happiness_mit.epub.zip)
2. upload file to x4
3. open

### Expected Behavior

I'd expect the book to open and not reboot Crosspoint.

### Relevant Log Output/Screenshots

```shell

```

Uri-Tauber:
I tested the attached book in version 1.1.0, and it opened correctly. I was able to browse through it without any issues.

However, the serial monitor shows an error that, while not affecting the reading experience, shouldn't happen. When opening the book for the first time, approximately 2,500 lines of the following message are generated within about two seconds:
`[20:25:19] [ERR] [GFX] !! Outside range (482, 335) -> (335, -3)`
(The numbers vary slightly.)

Here is the serial log:
```
[20:25:09] [SD] SD card detected
[20:25:09] [DBG] [CPS] Settings loaded from file
[I18N] Loaded language: 0
[20:25:09] [KRS] File does not exist: /.crosspoint/koreader.bin
[20:25:09] [DBG] [KRS] No credentials file found
[20:25:09] [DBG] [UI] Using Lyra theme
[20:25:09] [DBG] [MAIN] Starting CrossPoint version 1.0.0-dev
[20:25:09] EInkDisplay: begin() called
[20:25:09]   Static frame buffer (48000 bytes = 48KB)
[20:25:09]   Initializing e-ink display driver...
[20:25:09]   SPI initialized at 40 MHz, Mode 0
[20:25:09]   GPIO pins configured
[20:25:09]   Resetting display...
[20:25:09]   Display reset complete
[20:25:09]   Initializing SSD1677 controller...
[20:25:09]   Wait complete:  CMD_SOFT_RESET (2 ms)
[20:25:09]   Clearing RAM buffers...
[20:25:09]   Wait complete:  CMD_AUTO_WRITE_BW_RAM (5 ms)
[20:25:09]   Wait complete:  CMD_AUTO_WRITE_RED_RAM (5 ms)
[20:25:09]   SSD1677 controller initialized
[...]

PR #1068, opened 2026-02-21
fix: Correct hyphenation of URLs

Uri-Tauber:
## Summary

* **What is the goal of this PR?** Handle hyphenation of long URLs.

## Additional Context

See #1066 for more info.

---

### AI Usage


lukestein:
Noting existing discussion as #305 and #425 that resulted in a decision to not break after slashes. It's good for URLs but not for fractions or dates, and perhaps ambiguous in this/that constructions.

I'm open to revisiting but tagging in @osteotek.

More generally, our line breaking around punctuation should definitely be improved. I ran into last night a book with a bunch of things like

> …“so he said”—and we all wondered if…

(And related patterns from an author who loves em dashes) which should break after the dash but don't.

osteotek:
Text that doesn't fit on one line should break using fallback breaks whether it has dashes or not. Need to check why it isn't happening, might be a bug

osteotek (dismissed):

osteotek (approved):

jonasdiemer:
> Text that doesn't fit on one line should break using fallback breaks whether it has dashes or not. Need to check why it isn't happening, might be a bug

@osteotek are you still on it? I saw you approved this PR, but it does seem to introduce breaks in places that can be problematic outside of URLS (e.g. dates). So I'd agree with your statement and @lukestein that it would be preferred to not introduce `/` as a hyphen character and rather "hard break" stuff that doesn't fit on the screen otherwise.

itsthisjustin:
@Uri-Tauber is this still a good PR or should we close?

Uri-Tauber:
@itsthisjustin I think the problem this address still exist, although I should check to be sure.
The reason this wasn't merged yet is what @osteotek said: the proper fix should be forcing break at the edge of the screen whether the text has dashes or not.

I was planning to find more general solution, but I totally forgot about this...

itsthisjustin:
Ok sounds good. I do agree with the proper fix suggestion

osteotek:
I'll try to test this

jonasdiemer:
I am also still interested in having this fixed. Force-breaking if lines overflow is for sure needed. Still, I would also suggest to consider breaking on `/` and `-` when not surrounded by numbers (i.e. dates), so that we break stuff like "this/list/of/words/separated/by/slashes".

lukestein:
> I am also still interested in having this fixed. Force-breaking if lines overflow is for sure needed. Still, I would also suggest to consider breaking on `/` and `-` when not surrounded by numbers (i.e. dates), so that we break stuff like "this/list/of/words/separated/by/slashes".

Better hyphenation would help here too. I have some multiply/slashed-and-dashed examples in the test epub included with #2165 ;)

PR #1070, opened 2026-02-21
fix: prevent infinite render loop in Calibre Wireless after file transfer

pablohc:
## Summary

Fix an infinite render loop bug in CalibreConnectActivity that caused the e-ink display to refresh continuously every ~421ms after a file transfer completed.

- Added lastProcessedCompleteAt member variable to track which server completion timestamp has already been processed
- Modified the completion status update logic to only accept new values from the server, preventing re-processing of old timestamps
- Added clarifying comments explaining the fix

## Problem Description
After receiving a file via Calibre Wireless, the activity displays "Received: [filename]" for 6 seconds, then clears the message. However, the web server's wsLastCompleteAt timestamp persists indefinitely and is never cleared.

This created a race condition:

After 6 seconds, lastCompleteAt is set to 0 (timeout)
In the next loop iteration, status.lastCompleteAt (still has the old timestamp) ≠ lastCompleteAt (0)
The code restores lastCompleteAt from the server value
Immediately, the 6-second timeout condition is met again
This creates an infinite cycle causing unnecessary e-ink refreshes

## Solution
The fix introduces lastProcessedCompleteAt to track which server timestamp value has already been processed:

Only accept a new status.lastCompleteAt if it differs from lastProcessedCompleteAt
Update lastProcessedCompleteAt when processing a new value
Do NOT reset lastProcessedCompleteAt when the 6-second timeout clears lastCompleteAt
This prevents re-processing the same old server value after the timeout.

## Testing
Tested on device with multiple file transfer scenarios:

✅ File received message appears correctly after transfer
✅ Message clears after 6 seconds as expected
✅ No infinite render loop after timeout
✅ Multiple consecutive transfers work correctly
✅ Exiting and re-entering Calibre Wireless works as expected

## Performance Impact
Before: Infinite refreshes every ~421ms after timeout (high battery drain, display wear)
After: 2-3 refreshes after timeout, then stops (normal behavior)

## Additional Context
This is a targeted fix that only affects the Calibre Wireless file transfer screen. The root cause is the architectural difference between the persistent web server state (wsLastCompleteAt) and the per-activity display state (lastCompleteAt).

An alternative fix would be to clear wsLastCompleteAt in the web server after some timeout, but that would affect all consumers of the web server status. The chosen solution keeps the fix localized to CalibreConnectActivity.

---

### AI Usage

### DIFF
--- a/src/activities/network/CalibreConnectActivity.cpp
+++ b/src/activities/network/CalibreConnectActivity.cpp
@@ -28,6 +28,7 @@ void CalibreConnectActivity::onEnter() {
   currentUploadName.clear();
   lastCompleteName.clear();
   lastCompleteAt = 0;
+  lastProcessedCompleteAt = 0;
   exitRequested = false;
 
   if (WiFi.status() != WL_CONNECTED) {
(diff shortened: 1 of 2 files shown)


pablohc:
Hi @itsthisjustin, could you validate this fix? This was my first time transferring a book from Calibre, and I started noticing the screen behaving strangely. I investigated further, and Claude found the solution.

itsthisjustin:
Will take a look tonight!

itsthisjustin:
> Hi @itsthisjustin, could you validate this fix? This was my first time transferring a book from Calibre, and I started noticing the screen behaving strangely. I investigated further, and Claude found the solution.

Yes I saw the same thing. Locked up and wouldn't let me go back either. Was going to jump into this but patched the plugin yesterday instead. Good to see it's approved.

pablohc:
Hi @daveallie, could this PR get a review?

After a file transfer finishes in Calibre Wireless, a render loop starts, leaving ghost text on the screen that doesn’t clear, even after a full refresh.

This occurs after a transfer finishes and the device goes to sleep (~10 min.).

Screenshot_2026-02-27-02-01-20-05_99c04817c0de5652397fc8b56c3b3817

PR #1081, opened 2026-02-22
feat: Inverted ui orientation

badretdinov:
## Summary

Adds a new UI Orientation setting (Portrait / Inverted) that rotates the entire non-reader UI 180°. This enables mounting the Xteink X4 upside-down on an iPhone Air via a MagSafe case - the only physically viable orientation due to the MagSafe attachment geometry - while keeping the interface fully readable and usable.

IMG_0703

- New setting (uiOrientation): a binary Portrait / Inverted toggle in the Display category, serialized to settings storage with backward-compatible ordering. Separate from the existing 4-way reader orientation setting, since the UI only makes sense in portrait or inverted portrait.
- GfxRenderer: added drawTextRotated180() for rendering upside-down glyph text (used for button hint labels in inverted mode). Also fixed drawIcon() to draw pixel-by-pixel, which is required for correct rendering under coordinate transformation.
- MappedInputManager: added setButtonInversion() / buttonInverted flag. When inverted, the four front buttons are physically mirrored and Up/Down side buttons are swapped, so logical navigation remains correct.
- UITheme: introduced getContentTopY(renderer) and getContentBottomMargin(renderer) helpers that return orientation-aware layout values. In inverted mode top and bottom margins are swapped. All UI activities updated to use these instead of hardcoded metrics.topPadding / metrics.buttonHintsHeight.
- Battery indicator (BaseTheme + LyraTheme): in inverted mode, the battery is rendered at the logical bottom (physical top of the panel), accounting for the non-viewable margin (VIEWABLE_MARGIN_TOP), to avoid overlapping button hints.
- Button hints (BaseTheme + LyraTheme): in inverted mode, label text is drawn with drawTextRotated180() so it appears right-side-up relative to the user.
- Reader activities (EpubReaderActivity, TxtReaderActivity, XtcReaderActivity): on exit, orientation is restored to uiOrientation (instead of hardcoded Portrait). Button inversion is also applied/cleared on reader enter/exit.
- Settings live preview: changing UI Orientation in the settings menu takes effect immediately without restarting.
- Startup: UI orientation and button inversion are applied right after display initialization in main.cpp.
- i18n: added STR_UI_ORIENTATION with translations in 8 languages (EN, RU, DE, FR, ES, PT, CS, SV).

## Additional Context

- The inverted mode is designed specifically for use cases where the device is mounted physically upside-down (e.g., on an iPhone Air via MagSafe), not as a general-purpose rotation feature - hence only two options rather than four.
- No landscape options are exposed for UI orientation; landscape in the UI would require significant additional layout work and is out of scope here.
- The reader orientation setting is unaffected: it continues to support all 4 directions independently.
- Button inversion logic is fully decoupled from the display rotation - MappedInputManager handles it separately so physical button mappings stay correct regardless of the visual transformation.

---

### AI Usage


pablohc:
Thanks for this PR, but I wouldn't limit it to just Portrait and Inverted Portrait as the only valid orientations.
I see total sense in a landscape view, with the Power and side buttons at the top and the front buttons on the right side. Using it with a stand would be a great addition.

badretdinov:
I thought about this, but the main problem was that the side buttons were essentially drawn 90 degrees off.
How about I change the side button names to something like: ⌫ ↵ ← → (back, select, left, right)?
Or, you can suggest other options.

badretdinov:
I tried implementing at least a few screens and found that a lot of them needed to be redesigned.
I suggest either merging the PR and having only an inverted one at this stage, or canceling the PR and doing it as part of a larger redesign.

th0m4sek:
Polish string:
STR_UI_ORIENTATION: "Orientacja interfejsu"

ariel-lindemann:
String for Romanian:

`Orientare interfață`

itsthisjustin:
This seems incredibly useful given the description but risk for a bad merge is high and it touches a ton of files. I'd recommend moving this to a discussion where we can have the community vote on how needed this is, for now I'm closing due to risk.

PR #1083, opened 2026-02-22
Sleep as trmnl

drbourbon:
## Summary

## Additional Context

  specific areas to focus on).

---

### AI Usage

### DIFF
--- a/USER_GUIDE.md
+++ b/USER_GUIDE.md
@@ -28,7 +28,7 @@ Welcome to the **CrossPoint** firmware. This guide outlines the hardware control
 
 ## 1. Hardware Overview
 
-The device utilises the standard buttons on the Xtink X4 (in the same layout as the manufacturer firmware, by default):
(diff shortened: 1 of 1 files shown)


iandchasse:
This would be awesome, do you think maybe it could take presence in the settings menu instead of on the homescreen? It seems out of place if the only non-sleep functionality is setup, like OPDS and KOSync

drbourbon:
> This would be awesome, do you think maybe it could take presence in the settings menu instead of on the homescreen? It seems out of place if the only non-sleep functionality is setup, like OPDS and KOSync

Sure! I lazy coded this knowing that it wouldn't be approved being out of scope, but I'm moving it as setting anyway.

iandchasse:
I think it'd be a very useful feature, I don't really see the benefit of flashing a whole firmware for the functionality of TRMNL tbh. If it doesn't take up that much more flash/RAM I'd think it could definitely be approved.

PR #1086, opened 2026-02-22
refactor: Simplify new setting introduction

jpirnay:
## Summary

* **What is the goal of this PR?** Eliminate the 3-file / 4-location overhead for adding a new setting. Previously, every new setting required manually editing JsonSettingsIO.cpp in two places (save + load), duplicating knowledge already present in SettingsList.h. After this PR, JsonSettingsIO.cpp never needs to be touched again for standard settings.
  * `SettingInfo` (in `SettingsActivity.h`) gains one new field: `bool obfuscated` (base64 save/load for passwords), with a fluent builder method `.withObfuscated()`. The previously proposed `defaultValue`/`withDefault()` approach was dropped in favour of reading the struct field's own initializer value as the fallback (see below).
  * `SettingsList.h` entries are annotated with `.withObfuscated()` on the OPDS password entry. The list is now returned as a `static const` singleton (`const std::vector<SettingInfo>&`), so it is constructed exactly once. A missing `key`/`category` on the `statusBarProgressBarThickness` entry was also fixed — it was previously skipped by the generic save loop, so changes were silently lost on restart.
  * `JsonSettingsIO::saveSettings` and `loadSettings` replace their ~90 lines of manual per-field code with a single generic loop over `getSettingsList()`. The loop uses `info.key`, `info.valuePtr`/`info.stringOffset`+`info.stringMaxLen` (for char-array string fields), `info.enumValues.size()` (for enum clamping), and `info.obfuscated`.
  * **Default values**: instead of a duplicated `defaultValue` field in `SettingInfo`, `loadSettings` reads `s.*(info.valuePtr)` *before* overwriting it. Because `CrossPointSettings` is default-constructed before `loadSettings` is called, this captures each field's struct initializer value as the JSON-absent fallback. The single source of truth for defaults is `CrossPointSettings.h`.
  * One post-loop special case remains explicitly: the four `frontButton*` remap fields (managed by the RemapFrontButtons sub-activity, not in SettingsList) and `validateFrontButtonMapping()`.
  * One pre-loop migration guard handles legacy settings files that predate the status bar refactor: if `statusBarChapterPageCount` is absent from the JSON, `applyLegacyStatusBarSettings()` is called first so the generic loop picks up the migrated values as defaults and applies its normal clamping.
  * OPDS password backward-compat migration (plain `opdsPassword` → obfuscated `opdsPassword_obf`) is preserved inside the generic obfuscated-string path.

## Additional Context
Say we want to add a new `bookmarkStyle` enum setting with options `DOT`, `LINE`, `NONE` and a default of `DOT`:

1. `src/CrossPointSettings.h` — add enum and member:
```cpp
enum BOOKMARK_STYLE { BOOKMARK_DOT = 0, BOOKMARK_LINE = 1, BOOKMARK_NONE = 2 };
uint8_t bookmarkStyle = BOOKMARK_DOT;
```

2. `lib/I18n/translations/english.yaml` — add display strings:
```yaml
STR_BOOKMARK_STYLE: "Bookmark Style"
STR_BOOKMARK_DOT: "Dot"
STR_BOOKMARK_LINE: "Line"
```
(Other language files will fall back to English if not translated. Run `gen_i18n.py` to regenerate `I18nKeys.h`.)

3. `src/SettingsList.h` — add one entry in the appropriate category:
```cpp
SettingInfo::Enum(StrId::STR_BOOKMARK_STYLE, &CrossPointSettings::bookmarkStyle,
                  {StrId::STR_BOOKMARK_DOT, StrId::STR_BOOKMARK_LINE, StrId::STR_NONE_OPT},
                  "bookmarkStyle", StrId::STR_CAT_READER),
```

That's it — no default annotation needed anywhere, because `bookmarkStyle = BOOKMARK_DOT` in the struct already provides the fallback. The setting will automatically persist to JSON on save, load with clamping on boot, appear in the device settings UI under the Reader category, and be exposed via the web API — all with no further changes.

---

### AI Usage

Did you use AI tools to help write this code? _**< PARTIALLY>**_


src/SettingsList.h
```diff
@@ -27,17 +27,20 @@ inline std::vector<SettingInfo> getSettingsList() {
...
+          .withDefault(CrossPointSettings::FULL),
       SettingInfo::Enum(StrId::STR_HIDE_BATTERY, &CrossPointSettings::hideBatteryPercentage,
                         {StrId::STR_NEVER, StrId::STR_IN_READER, StrId::STR_ALWAYS}, "hideBatteryPercentage",
                         StrId::STR_CAT_DISPLAY),
       SettingInfo::Enum(
           StrId::STR_REFRESH_FREQ, &CrossPointSettings::refreshFrequency,
           {StrId::STR_PAGES_1, StrId::STR_PAGES_5, StrId::STR_PAGES_10, StrId::STR_PAGES_15, StrId::STR_PAGES_30},
-          "refreshFrequency", StrId::STR_CAT_DISPLAY),
+          "refreshFrequency", StrId::STR_CAT_DISPLAY)
+          .withDefault(CrossPointSettings::REFRESH_15),
```
znelson: Are defaults like this now duplicated here and in CrossPointSettings.h?

src/SettingsList.h
ngxson, line 1: Just wondering if we can somehow build this as a static list. The current `getSettingsList` firstly construct the list on stack, then copy it to heap, so a bit wasteful in term of memory.

One idea is to reimplement `SettingInfo::Enum`, `SettingInfo::Value`, etc as macros, and store the instance as `static const`

src/SettingsList.h
jpirnay: Yes (initial state of PR) and no: no longer. Changed it

src/SettingsList.h
jpirnay: Changed it

src/JsonSettingsIO.cpp
```diff
@@ -170,42 +193,10 @@ bool JsonSettingsIO::loadSettings(CrossPointSettings& s, const char* json, bool*
...
-    pass = doc["opdsPassword"] | "";
-    if (!pass.empty() && needsResave) *needsResave = true;
-  }
-  strncpy(s.opdsPassword, pass.c_str(), sizeof(s.opdsPassword) - 1);
-  s.opdsPassword[sizeof(s.opdsPassword) - 1] = '\0';
+
   LOG_DBG("CPS", "Settings loaded from file");
 
-  if (doc.containsKey("statusBarChapterPageCount")) {
+  if (!doc["statusBarChapterPageCount"].isNull()) {
```
znelson: Should this block land before the `for (const auto& info : getSettingsList()) {` loop, so the values read here get clamped? (Probably not critical in practice.)

src/SettingsList.h
znelson: This is missing the "statusBarProgressBarThickness" key so the generic settings loop will currently skip it:

```
SettingInfo::Enum(StrId::STR_PROGRESS_BAR_THICKNESS, &CrossPointSettings::statusBarProgressBarThickness,
                  {StrId::STR_PROGRESS_BAR_THIN, StrId::STR_PROGRESS_BAR_MEDIUM, StrId::STR_PROGRESS_BAR_THICK},
                  "statusBarProgressBarThickness", StrId::STR_CUSTOMISE_STATUS_BAR),
```

jpirnay:
Needs rework

src/SettingsList.h
jpirnay: Fixed

src/JsonSettingsIO.cpp
jpirnay: Rewrote the logic:
- Normal case: the migration check is a no-op, the generic loop loads and clamps all values correctly
- Legacy migration case: applyLegacyStatusBarSettings populates s with migrated values, which the generic loop then picks up as the fieldDefault (line 175) and applies proper clamping on top

znelson (approved):
Tedious stuff, but really helpful in the long run. Thanks for taking it on!

Issue #982, opened 2026-02-18
[Feature Request] A QR code function for copying on screen text


Befbo:
I have no idea how many other people would use such a feature, but it would be kind of neat to have the ability to generate a QR code with all of the text currently on screen, much like the way Koreader will let you do for copying text onto your phone. Since there isn't currently a way to highlight text, maybe it could just do all of the text that is currently on the screen. That way you could store notes on your phone, or send little clips to people. I understand that this may not be feasible, or something that anybody would even want or use.

Uri-Tauber:
Closing as solved.

PR #1055, opened 2026-02-20
perf: byte-level framebuffer writes for fillRect and axis-aligned drawLine

jpirnay:
## Summary

* **What is the goal of this PR?** Eliminate per-pixel drawPixel calls for solid fills and axis-aligned lines by writing directly to the framebuffer at byte granularity, exploiting the fact that one logical dimension always maps to a contiguous physical row in the framebuffer.
* **What changes are included?**
  - GfxRenderer: new private fillPhysicalHSpan(phyY, phyX_start, phyX_end, state) helper that writes a solid horizontal span to a physical framebuffer row with MSB-first bit packing, partial-byte masking at both edges, and memset for the aligned middle section.
  - fillRect: replaced the per-row drawLine loop with orientation-specific fast paths — each logical axis that maps to a constant physical row is iterated as the outer loop, and the perpendicular extent is written as a single fillPhysicalHSpan call. Covers all 4 orientations.
  - drawLine (axis-aligned cases): logical vertical lines in Portrait/PortraitInverted and logical horizontal lines in Landscape now route through fillPhysicalHSpan. Bresenham path for diagonal lines is unchanged.

## Additional Context

Why per-pixel was expensive: drawPixel calls rotateCoordinates() (switch + arithmetic), a bounds check, a byte-index multiply, and a read-modify-write on the framebuffer — for every pixel. fillRect(480, 800) previously issued 384,000 such calls.

Geometry: In Portrait orientation, a logical column (fixed x, varying y) maps to a single physical row (phyY = 479 - x), with phyX = y. Eight consecutive y values share one framebuffer byte. The same complementary relationship holds for each other orientation, so all four orientations have a fast path for fillRect. For drawLine, only the orientation where the line runs along the fast axis gets the byte-write path; the other axis still falls back to per-pixel.

Measured results on hardware (ESP32-C3, Portrait orientation):
|Operation | Before | After | Speedup|
|-- | -- | -- | --|
|fillRect(480×800) | 125,577 µs | 519 µs | 242×|
|drawLine vertical 800px | 261 µs | 3 µs | 87×|
|drawLine horizontal 480px | 153 µs | 162 µs | ~1× (no fast path, expected)|

The fill speedup is near-theoretical: for a full-screen Portrait fill, the inner body reduces to memset(row, 0x00, 100) × 480 rows - 48 KB of aligned writes at ~93 MB/s SRAM bandwidth.

[...]

PR #1099, opened 2026-02-22
feat: Current page as QR

el:
## Summary

* **What is the goal of this PR?** Implements QR text of the current page

## Additional Context

I saw this feature request at #982 
It made sense to me so I implemented. But if the team thinks it is not necessary please let me know and we can close the PR. 

| Page | Menu | QR |
|------|-------|----|
| IMG_6601.bmp | IMG_6599.bmp | IMG_6600.bmp |

---

### AI Usage

Did you use AI tools to help write this code? _** YES

### DIFF
--- a/src/util/QrUtils.cpp
+++ b/src/util/QrUtils.cpp
@@ -2,9 +2,12 @@
 
 #include <qrcode.h>
 
+#include <algorithm>
+#include <memory>
+
 #include "Logging.h"
 
-void QrUtils::drawQrCode(GfxRenderer& renderer, const Rect& bounds, const std::string& textPayload) {
+void QrUtils::drawQrCode(const GfxRenderer& renderer, const Rect& bounds, const std::string& textPayload) {
   // Dynamically calculate the QR code version based on text length
   // Version 4 holds ~114 bytes, Version 10 ~395, Version 20 ~1066, up to 40
   // qrcode.h max version is 40.
--- a/src/util/QrUtils.h
+++ b/src/util/QrUtils.h
@@ -9,6 +9,6 @@
 namespace QrUtils {
 
 // Renders a QR code with the given text payload within the specified bounding box.
-void drawQrCode(GfxRenderer& renderer, const Rect& bounds, const std::string& textPayload);
+void drawQrCode(const GfxRenderer& renderer, const Rect& bounds, const std::string& textPayload);
 
 }  // namespace QrUtils


src/activities/reader/QrDisplayActivity.cpp
```diff
@@ -0,0 +1,90 @@
...
+    const int yOff = startY + (availableHeight - qrDisplaySize) / 2;
+
+    // Draw the QR Code
+    for (uint8_t cy = 0; cy < qrcode.size; cy++) {
+      for (uint8_t cx = 0; cx < qrcode.size; cx++) {
+        if (qrcode_getModule(&qrcode, cx, cy)) {
+          renderer.fillRect(xOff + px * cx, yOff + px * cy, px, px, true);
+        }
+      }
+    }
```
znelson: There's not a ton of duplication here, but it might be good eventually to consolidate this and `drawQRCode` in CrossPointWebServerActivity.cpp to a single QR code drawing routine, especially as folks work on drawing performance changes like #1055 and #1060.

lib/Epub/Epub/blocks/TextBlock.h
```diff
@@ -18,6 +18,7 @@ class TextBlock final : public Block {
   BlockStyle blockStyle;
 
  public:
+  const std::vector<std::string>& getWords() const { return words; }
```
znelson: Style nit-pick: can we move this below the constructor with other public methods?

src/activities/reader/QrDisplayActivity.cpp
```diff
@@ -0,0 +1,90 @@
...
+  const size_t len = textPayload.length();
+  int version = 4;
+  if (len > 114) version = 10;
+  if (len > 395) version = 20;
+  if (len > 1066) version = 30;
+  if (len > 2110) version = 40;
+
+  // Make sure we have a large enough buffer on the heap to avoid blowing the stack
+  uint32_t bufferSize = qrcode_getBufferSize(version);
+  uint8_t* qrcodeBytes = new uint8_t[bufferSize];
```
znelson: You could save the manual `delete` later with 
```suggestion
  auto qrcodeBytes = std::make_unique<uint8_t[]>(bufferSize);
```

znelson (approved):

lukestein:
I think this is very cool. Haven't tested on device yet, but I noticed that the demo screenshot seems to have hyphenation included in the QR encoded text.

Can we avoid that and just represent the original, un-hyphenated text?

osteotek:
> I think this is very cool. Haven't tested on device yet, but I noticed that the demo screenshot seems to have hyphenation included in the QR encoded text.
> 
> Can we avoid that and just represent the original, un-hyphenated text?

Not really possible without "reindexing" whole document again at the moment of creating QR (or doing something creative like running hyphenation algorithm in reverse)

PR #985, opened 2026-02-19
fix: Remember selected item when returning to Home

longlostnick:
## Summary

* **What is the goal of this PR?**  
  Small little improvement to the Home screen UX by remembering/restoring the last selected cursor position when returning to Home. This was bothering me whenever I went back to home. It would reset back to the first option (typically the first book in the very upper left) and I'd have to move the cursor all the way back down to the option I wanted (e.g. Settings)

* **What changes are included?**
  - Added `HomeSelectorMemory` helper (`src/activities/home/HomeSelectorMemory.h`) to store/restore the last Home selector index with bounds clamping.
  - Updated `HomeActivity` to:
    - persist selection on exit,
    - restore selection on enter,
    - apply a **pending/deferred restore** during early loop/render until menu size is stable (prevents losing intended selection if available items change),
    - stop deferred restore once user navigates (so user input always wins).
  - Added/updated host-side tests:
    - `test/host/home_selector_memory/HomeSelectorMemoryTest.cpp`
    - includes coverage for clamp/restore behavior and stored-index access.
  - Introduced an extensible host test runner:
    - `test/run_host_tests.sh` (auto-discovers `test/host/**/*Test.cpp`, supports optional `.sources` and `.cxxflags` manifests).
  - CI update:
    - `host-tests` job now runs `bash test/run_host_tests.sh`.
  - Documentation update:
    - `test/README` now explains host test conventions and how to add new tests.

## Test Plan
- Automated tests
- Manual testing by flashing my device

---

### AI Usage

Did you use AI tools to help write this code? _**YES**_

PR #1103, opened 2026-02-22
feat: remember selected menu item when returning to Home

dexif:
## Summary

When navigating back to Home always reset the cursor to the top. Now each activity passes its corresponding HomeMenuItem enum value so the cursor returns to the item the user navigated from.


ngxson:
This change will likely be overwritten by https://github.com/crosspoint-reader/crosspoint-reader/pull/1016

I think it's better to delegate this task `ActivityManager` rather adding too many things on top of `main.cpp`

Probably we may convert the `HomeActivity` into a "launcher", keep it on background like what Android does.

daveallie:
A different approach was put together in https://github.com/crosspoint-reader/crosspoint-reader/pull/985

znelson:
Superseded by #1629

Issue #1052, opened 2026-02-20
Feature: select sleep image


oolong:
It's really nice being able to select your sleep screen on the device. The default firmware actually does this pretty well! 

I'm not sure how we're meant to find out how to do it right now, but it took me quite a bit of googling...

inseend:
https://github.com/crosspoint-reader/crosspoint-reader/blob/master/USER_GUIDE.md#36-sleep-screen 
It's in the user guide. If you place sleep.bmp in the root folder, it will always use that as the sleep screen image.

And if you place multiple bmp files in a folder called sleep, it will randomly pick one of those images.

Maybe the user guide should be more front and center.

oolong:
> https://github.com/crosspoint-reader/crosspoint-reader/blob/master/USER_GUIDE.md#36-sleep-screen It's in the user guide. If you place sleep.bmp in the root folder, it will always use that as the sleep screen image.

Thanks, yes, I did find this eventually, but the request stands. It would still be really nice if we could see BMP files in the navigation interface, and put them in the appropriate place from there.

I guess I struggled to find the user guide because it's embedded quite a long way down a large amount of text in the readme file. Perhaps a supplemental feature request here would be to include the user guide somewhere near the top?

Chaseagle96:
I agree having a way to access these files natively would be nice, or selecting specific images to feature on the lock screen. Being able to manage my sleep wallpapers and view them on device.

el:
Now that we have BMP viewer #887  
I am thinking we can add a button to it that sets it as a sleep image.
so it would:
- Copy image to /sleep.png
- Rename /sleep directory if there is one
- Set the Sleep Screen to "Custom"

I think this is a valid case and I can create a PR for this.

osteotek:
@el there was similar functionality in https://github.com/crosspoint-reader/crosspoint-reader/pull/968, may be the code can be reused

itsthisjustin:
Closing as out of scope for priorities right now

PR #968, opened 2026-02-18
feat: navigate BMPs and set sleep image from viewer

barbarhan:
## Summary

Goal:
Refine the BMP viewer workflow by making it easier to navigate images in a folder and set a specific BMP as the sleep image, while keeping the settings UI aligned with the actual selected image.

Changes:
- Added left/right navigation between BMP files in the current folder from the BMP viewer.
- Added a `Set Sleep` action in the BMP viewer to save the current BMP as the sleep image.
- Added support for rendering a specifically selected BMP as the sleep image.
- Polished the Settings UI so the selected BMP filename (for example `ariane.bmp`) is shown only when that specific-image sleep mode is active.
- Hid the generic image-from-folder mode from normal cycling unless a specific BMP has already been selected.
- Removed the redundant extra sleep image path row from settings.

AI Usage:
Did you use AI tools to help write this code? YES, OpenClaw / GPT-5.4.

PR #909, opened 2026-02-16
feat: Long Click for File Deletion through File Browser

Levrk:
## Summary

* **What is the goal of this PR?** (e.g., Implements the new feature for file uploading.)

Allow users to better manage their epub library by offloading unwanted or finished books and other files. Resolves #893 

* **What changes are included?**

Added Delete Book shortcut in the fil browser. Delete function implements the new ConfirmationActivity to show file name and solicit user interaction before either returning to the file browser on a press of the back button, or proceeding to delete. Delete function then deletes the file and returns user to the file browser menu at the current directory. Video of it working on my machine attached here:

https://github.com/user-attachments/assets/329b0198-9e97-45ad-82aa-c39894351667

## Additional Context

* Add any other information that might be helpful for the reviewer (e.g., performance implications, potential risks, 
  specific areas to focus on).

Certainly potential risks associated with file deletion. Please let me know if there are any concerns that need to be better addressed. I think this is a very good feature to have to go along with the new screenshots so you don't get stuck with a bunch of extra files on your device. Also I did add this to the user guide.

---

### AI Usage

While CrossPoint doesn't have restrictions on AI tools in contributing, please be transparent about their usage as it 
helps set the right context for reviewers.

Did you use AI tools to help write this code? _**YES**_

PR #1104, opened 2026-02-22
feat: Set sleep cover from BMP viewer

el:
## Summary

* **What is the goal of this PR?** Implements setting a cover image in the device

- Added 2 more buttons to the BMP viewer:
- Set Cover copies the current BMP file to /sleep.bmp
- Delete lets you delete a BMP file (with a confirmation dialog)
- /sleep.bmp now takes precedence over the sleep/ folder since we now have a way of deleting it to go back to the cover.
- Volume buttons now let you switch to the previous or next image in the folder #968

## Additional Context

| screenshot-102386.bmp |
| --- |

Addresses #1052 

---

### AI Usage


Levrk:
Awesome!

osteotek:
We will probably have file deletion in file manager via https://github.com/crosspoint-reader/crosspoint-reader/pull/909 not sure if it will be needed here

btw, what happens if /sleep.bmp is deleted

el:
> We will probably have file deletion in file manager via #909 not sure if it will be needed here
> 
> btw, what happens if /sleep.bmp is deleted

I think it is ok to have in both places. As long as we share the Prompt and deletion logic. Thats why I moved the Prompt (that uses a Toast) to shareable activity.

When sleep.bmp is deleted, it falls-back to sleep/ directory. Previously sleep/ folder had a precedence over /sleep.bmp I  just changed the priority.

osteotek:
> > We will probably have file deletion in file manager via #909 not sure if it will be needed here
> > btw, what happens if /sleep.bmp is deleted
> 
> I think it is ok to have in both places. As long as we share the Prompt and deletion logic. Thats why I moved the Prompt (that uses a Toast) to shareable activity.
> 

I just don't like reusing front nav button for something else. We try to keep consistency in behavior for front and side navigation buttons everywhere. Here we break that consistency, one of the front nav button will behave differently from side nav button

el:
> > > We will probably have file deletion in file manager via #909 not sure if it will be needed here
> > > btw, what happens if /sleep.bmp is deleted
> > 
> > 
> > I think it is ok to have in both places. As long as we share the Prompt and deletion logic. Thats why I moved the Prompt (that uses a Toast) to shareable activity.
> 
> I just don't like reusing front nav button for something else. We try to keep consistency in behavior for front and side navigation buttons everywhere. Here we break that consistency, one of the front nav button will behave differently from side nav button

removed the delete file logic

znelson:
This looks valuable, but needs attention. I see @Uri-Tauber you self-assigned. Will leave this open for now.

Uri-Tauber:
@znelson Yeah. I’d like to review it, but I’ve been busy with my RTL support work. I’m hoping to get to it in the next few days.

@el, if you could rebase it onto `master`, that would really help.

el:
> @znelson Yeah. I’d like to review it, but I’ve been busy with my RTL support work. I’m hoping to get to it in the next few days.
> 
> @el, if you could rebase it onto `master`, that would really help.

should be ready

Uri-Tauber (dismissed):
Great work overall! I tested this on my device and it works exactly as expected. 
I noticed one small opportunity for deduplication: there is logic to find the last slash and extract the directory path repeated in three places within `BmpViewerActivity.cpp` (specifically in `loadSiblingImages()` and the Up/Down button logic). This logic is almost identical to what `extractFolderPath()` do.
Could you swap out the manual parsing for that function call?

el:
Done

Uri-Tauber (approved):
Nice work.

Uri-Tauber:
@osteotek @znelson @mcrosson Can you take a look at this PR. it’s a nice QoL improvement worth adding.

PR #1105, opened 2026-02-22
fix: added romanian translation to new strings

ariel-lindemann:
## Summary

Added Romanian translations for newly addded strings

Just the translations in the localisation file.

## Additional Context

  specific areas to focus on).

---

### AI Usage


lib/I18n/translations/romanian.yaml
```diff
@@ -256,6 +256,7 @@ STR_STATUS_BAR_FULL_CHAPTER: "Complet cu bara de capitol"
 STR_UI_THEME: "Tema UI"
 STR_THEME_CLASSIC: "Clasic"
 STR_THEME_LYRA: "Lyra"
+STR_THEME_LYRA_EXTENDED: "Lyra Extended"
```
znelson, line 259: Should we be translating "Extended"?

ariel-lindemann: In all languages except Spanish it is left as `Lyra Extended` so I decided to follow the majority

But maybe we could decide on a common policy for that

znelson (approved):

PR #1107, opened 2026-02-22
feat: set WiFi hostname to CrossPoint-Reader-XXXXXXXXXXXX

dexif:
## Summary

Replace the default esp32-XXXXXXXXXXXX hostname with CrossPoint-Reader-AABBCCDDEEFF (full MAC address) so the device is easily identifiable on the router's client list.

### DIFF
--- a/src/activities/network/WifiSelectionActivity.cpp
+++ b/src/activities/network/WifiSelectionActivity.cpp
@@ -219,6 +219,12 @@ void WifiSelectionActivity::attemptConnection() {
 
   WiFi.mode(WIFI_STA);
 
+  // Set hostname so routers show "CrossPoint-Reader-AABBCCDDEEFF" instead of "esp32-XXXXXXXXXXXX"
+  String mac = WiFi.macAddress();
+  mac.replace(":", "");
+  String hostname = "CrossPoint-Reader-" + mac;
+  WiFi.setHostname(hostname.c_str());
+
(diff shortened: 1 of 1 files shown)


lukestein:
Hi @dexif. Please check out discussion at #671 about concerns this may raise regarding workflows that rely on `crosspoint.local`. Not sure whether your implementation raises the same concerns, but want to make you aware of them.

dexif:
Hi @lukestein.
The concern from #671 was about changing the mDNS name, which would break crosspoint.local. This PR doesn't do that. 
                                                                                                                                                                                                                                            
WiFi.setHostname() sets the DHCP hostname - what the router displays in its client list. 
At the moment this hostname looks like esp32s3-23CDF0

The mDNS name is registered separately via MDNS.begin("crosspoint") in CrossPointWebServerActivity.cpp and is not affected by the DHCP hostname.
                                         
These are two independent mechanisms:
DHCP hostname (router client list): `CrossPoint-Reader-AABBCCDDEEFF` - set by WiFi.setHostname() before WiFi.begin()
mDNS name (service discovery): `crosspoint.local` -set by MDNS.begin("crosspoint") when the web server starts

crosspoint.local continues to work as before. Existing workflows (curl, Apple Shortcuts, etc.) are unaffected.

dexif:
Devices built on ESP often don’t change this hostname, so on the router you end up seeing a bunch of ESPs and it’s not clear what kind of device it actually is.

PR #1124, opened 2026-02-23
feat: Vietnamese support

danoooob:
## Summary

  Add Vietnamese (Tiếng Việt) language support to CrossPoint Reader — both UI translations and font glyph coverage.

  **1. Font: Vietnamese glyph support**
  - Added Vietnamese-specific Unicode ranges to `fontconvert.py`:
    - Latin Extended-B subset (`U+01A0–01B0`: Ơ/ơ, Ư/ư)
    - Combining Diacritical Marks (`U+0300–0303`, `U+0309`, `U+0323`)
    - Vietnamese Extended precomposed characters (`U+1EA0–1EF9`)
  - Only the minimal required codepoints were selected to keep flash usage as low as possible.
  - Added **NotoSans as fallback font** for Ubuntu UI font, since Ubuntu font does not contain Vietnamese glyphs. The font conversion script (`convert-builtin-fonts.sh`) now passes NotoSans as a secondary font source for Ubuntu variants.

  **2. I18n: Vietnamese translation**

### Memory Impact

| Metric | Before | After | Delta |
|:--|--:|--:|--:|
| Flash Data (`.rodata`) | 2,969,212 B | 3,348,828 B | **+379,616 B (+12.8%)** |
| Flash Code (`.text`) | 1,478,248 B | 1,478,264 B | +16 B |
| Total image size | 4,660,137 B | 5,039,769 B | **+379,632 B (+8.1%)** |
| Flash usage | 69.1% | 74.9% | **+5.8 pp** |
| RAM usage | 29.0% | 29.0% | unchanged |

> The ~371 KB flash increase is entirely from additional font glyph bitmaps in `.rodata`. RAM is unaffected. Flash still has ~25% headroom remaining.

### Notes
- Vietnamese characters were selected conservatively — only the precomposed forms and essential combining marks were included to minimize flash footprint.
- The Ubuntu font family lacks Vietnamese glyphs, so NotoSans is used as a fallback font source during font conversion. Other font families (Bookerly, NotoSans, OpenDyslexic) already contain the needed Vietnamese glyphs natively.

---

### AI Usage


lib/EpdFont/scripts/fontconvert.py
```diff
@@ -37,6 +37,20 @@
...
+    ### Latin Extended-B (Vietnamese subset only) ###
+    # Only Ơ/ơ (U+01A0-01A1), Ư/ư (U+01AF-01B0) for Vietnamese
+    (0x01A0, 0x01B0),
+    ### Combining Diacritical Marks (Vietnamese essential) ###
+    # Grave, Acute, Circumflex, Tilde (U+0300-0303)
+    (0x0300, 0x0303),
+    # Hook Above - Vietnamese hỏi tone (U+0309)
+    (0x0309, 0x0309),
+    # Dot Below - Vietnamese nặng tone (U+0323)
+    (0x0323, 0x0323),
```
ngxson: These requires going back to last character and render this character on top of it, but if I understand correctly, crosspoint font rendering only support advancing along X dimension, not going backwards.

Please verify if this actually work or not.

Edit: yes, we do support `combiningGlyph`

danoooob: Tested this — it does work, but there are a couple of rendering issues:
- Dot below (U+0323) renders as dot above instead of below the character.
- Multiple combining marks stack on top of each other above the base character, rather than being positioned correctly (e.g. circumflex + dot).

See test screenshots below:
IMG_001
IMG_002
IMG_003

ngxson: No idea why, but I think the easier way now is to simply have a logic to map a sequence of codepoints to a glyph. Example: `o^~` --> `ỗ`

ngxson: Otherwise, you can also try patching the `combiningGlyph` logic: https://github.com/crosspoint-reader/crosspoint-reader/blob/13fc8b94b0fbce0b0a38ea53959fe32e8bdbd186/lib/GfxRenderer/GfxRenderer.cpp#L233-L248

But this is above my head, probably better to exact it to another PR for visibility

danoooob: Or maybe we can skip supporting combining diacritics altogether? Nowadays, almost everyone uses precomposed Unicode. Finding an ebook that still uses combining Unicode is very rare, mostly just files from the early 2000s or before. I think the effort to support converting from combining to precomposed isn't worth it for such edge cases.

ngxson: > Or maybe we can skip supporting combining diacritics altogether

Yes that's what I suggested when I say:

> Support for more language-specific codepoints should be implemented as sdcard fonts, it's a planned feature

It's safe to delete this `Combining Diacritical Marks` block then?

danoooob: Yes, safe to delete. Though I just realized there's already an overlapping subset defined right below:
https://github.com/crosspoint-reader/crosspoint-reader/blob/57250b97e44191d20fcaa45563659a386d7c40ec/lib/EpdFont/scripts/fontconvert.py#L50-L52

lib/EpdFont/scripts/convert-builtin-fonts.sh
```diff
@@ -46,8 +46,10 @@ for size in ${UI_FONT_SIZES[@]}; do
   for style in ${UI_FONT_STYLES[@]}; do
     font_name="ubuntu_${size}_$(echo $style | tr '[:upper:]' '[:lower:]')"
     font_path="../builtinFonts/source/Ubuntu/Ubuntu-${style}.ttf"
+    # Use NotoSans as fallback for Vietnamese characters not in Ubuntu
+    fallback_path="../builtinFonts/source/NotoSans/NotoSans-${style}.ttf"
     output_path="../builtinFonts/${font_name}.h"
-    python fontconvert.py $font_name $size $font_path > $output_path
+    python fontconvert.py $font_name $size $font_path $fallback_path > $output_path
     echo "Generated $output_path"
```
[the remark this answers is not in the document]
ngxson: I think it's a bit waste of space to have a fallback like this. There will be the same set of glyphs appear twice in the flash memory.

Instead, better to reduce the scope of your PR to only add the font support for reading books (no UI and i18n changes). And then open another PR for UI and i18n.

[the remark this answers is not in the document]
ngxson: @coderabbitai no I'm not talking with you, talking with @danoooob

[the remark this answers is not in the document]
danoooob: @ngxson Got it, that makes sense. I'll split this into smaller PRs to make it easier to review and keep the discussion more focused.

Issue #1118, opened 2026-02-23
OPDS. Non-latin fonts in books names not saved correctly


a-serpichev:
### Affected Version

1.0.0, 1.1.0

### Describe the Bug

When I download a book form Calibre Web and book name is in cyrilic, I get book saved to root folder with "_.epub" name instead of right name. Although name was displayed correctly in the OPDS catalogue and during download progress
I think this could be relevant to all non-latin characters

### Steps to Reproduce

1. Open OPDS Browser (OPDS catalogue already connected in settings to calibre web in local network)
2. Go to "Recently added books"
3. Select any book with cyrillic name to download
4. Wait for download to finish, See the right name on the screen during progress bar visualization
5. Check file name of downloaded file:

Expected behaviour: You are able to read file name in cyrillic
Current behaviour: file saved as "_.epub"  (probably replacing all cyrillic symbols with blank ones and space symbol with "_"

### Expected Behavior

Expected behaviour: You are able to read file name in cyrillic after downloading file from OPDS Directory

### Relevant Log Output/Screenshots

```shell

```

spl3g:
Yeah, downloading removes all non-ASCII characters..
https://github.com/crosspoint-reader/crosspoint-reader/blob/57250b97e44191d20fcaa45563659a386d7c40ec/src/activities/browser/OpdsBookBrowserActivity.cpp#L321
https://github.com/crosspoint-reader/crosspoint-reader/blob/57250b97e44191d20fcaa45563659a386d7c40ec/src/util/StringUtils.cpp#L11-L20

osteotek:
Fixed with https://github.com/crosspoint-reader/crosspoint-reader/pull/1132

PR #1132, opened 2026-02-23
fix: Handle non-ASCII characters in sanitizeFilename

znelson:
## Summary

Probable fix for #1118. `sanitizeFilename` was only passing through ASCII characters from filenames. It now maintains valid UTF-8 codepoints, including non-ASCII multibyte sequences. Truncation happens at a maximum number of bytes, rather than characters, to prevent filenames with many multibyte sequences from unexpectedly exceeding FAT32 limits.

---

### AI Usage

Did you use AI tools to help write this code? _**YES, I described #1118 to Claude and it suggested sanitizeFilename as the likely cause**_

### DIFF
--- a/lib/Utf8/Utf8.cpp
+++ b/lib/Utf8/Utf8.cpp
@@ -15,28 +15,14 @@ uint32_t utf8NextCodepoint(const unsigned char** string) {
 
   const int bytes = utf8CodepointLen(**string);
   const uint8_t* chr = *string;
+  *string += bytes;
 
   if (bytes == 1) {
-    *string += 1;
     return chr[0];
   }
 
-  for (int i = 1; i < bytes; ++i) {
-    // Validate each continuation byte starts with expected 0b10 sequence.
-    // This also handles unexpected null string terminator in the middle of
-    // a UTF-8 multibyte sequence.
-    if ((chr[i] & 0xC0) != 0x80) {
-      *string += 1;
-      return REPLACEMENT_GLYPH;
-    }
-  }
+  uint32_t cp = chr[0] & ((1 << (7 - bytes)) - 1);  // mask header bits
 
-  *string += bytes;
-
-  // Mask off the 0b110, 0b1110, or 0b11110 header bits from the first byte.
-  uint32_t cp = chr[0] & ((1 << (7 - bytes)) - 1);
-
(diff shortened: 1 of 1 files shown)


osteotek (dismissed):
Fixed OPDS downloading of files with non-latin names

src/util/StringUtils.cpp
```diff
@@ -1,35 +1,45 @@
...
+      if (result.length() + 1 > maxBytes) break;
       result += '_';
-    } else if (c >= 32 && c < 127) {
-      // Keep printable ASCII characters
-      result += c;
+    } else if (cp >= 128 || (cp >= 32 && cp < 127)) {
+      const size_t cpBytes = text - cpStart;
+      if (result.length() + cpBytes > maxBytes) break;
+      result.append(reinterpret_cast<const char*>(cpStart), cpBytes);
     }
```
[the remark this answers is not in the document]
znelson: Fixed with 7add13a84f9d71aec9c90d3e1bf295e9802cc1be, please confirm

lib/Utf8/Utf8.cpp
```diff
@@ -15,14 +15,28 @@ uint32_t utf8NextCodepoint(const unsigned char** string) {
...
   const uint8_t* chr = *string;
-  *string += bytes;
 
   if (bytes == 1) {
+    *string += 1;
     return chr[0];
   }
 
-  uint32_t cp = chr[0] & ((1 << (7 - bytes)) - 1);  // mask header bits
+  for (int i = 1; i < bytes; ++i) {
```
osteotek: utf8NextCodepoint() is used heavily during text layout, and runs on every character of the text. Are we sure we want to introduce additional computation here instead of simple shift operation

osteotek:
@znelson I am not sure if your last commit is necessary, worked fine without it

lib/Utf8/Utf8.cpp
znelson: Sure, let's consider this as a separate change. There's a potential existing risk for this code to blindly move the `**string` pointer beyond a null terminator, if the input string was incorrectly terminated mid-multibyte sequence. But I agree it's not directly related to the original fix and better to keep separate.

znelson:
Thanks @osteotek, I moved the `utf8NextCodepoint` change to #1134 for independent discussion

PR #1134, opened 2026-02-23
fix: Handle null terminator in the middle of UTF-8 multibyte sequence

znelson:
## Summary

Raised as a concern by @coderabbitai in #1132, `utf8NextCodepoint` could blindly move the `string` pointer beyond a null terminator if the input string was incorrectly terminated in the middle of a UTF-8 multibyte sequence.

---

### AI Usage

Did you use AI tools to help write this code? _**PARTIALLY, CodeRabbit suggested fixing this and I made the changes**_


lib/Utf8/Utf8.cpp
```diff
@@ -15,14 +15,28 @@ uint32_t utf8NextCodepoint(const unsigned char** string) {
...
-  uint32_t cp = chr[0] & ((1 << (7 - bytes)) - 1);  // mask header bits
+  for (int i = 1; i < bytes; ++i) {
+    // Validate each continuation byte starts with expected 0b10 sequence.
+    // This also handles unexpected null string terminator in the middle of
+    // a UTF-8 multibyte sequence.
+    if ((chr[i] & 0xC0) != 0x80) {
+      *string += 1;
+      return REPLACEMENT_GLYPH;
+    }
+  }
```
znelson, line 32: The question here is whether it's worth the extra work to look out for improperly terminated UTF-8 multibyte sequences?

znelson:
Covered by #1056

PR #1074, opened 2026-02-22
feat: Add text wrapping to Lyra Extended theme

iandchasse:
## Summary

* **What is the goal of this PR?** (e.g., Implements the new feature for file uploading.)
Wrap the text of the 3-cover Lyra Theme to follow the work from #1013 but preceding the work to be done continuing from #911
* **What changes are included?**
Changes font size for 3 cover theme, works nicely, wraps up to 3 lines of text for each book. Still maintains good spacing.
## Additional Context

image

* Add any other information that might be helpful for the reviewer (e.g., performance implications, potential risks, 
  specific areas to focus on).

---

### AI Usage

While CrossPoint doesn't have restrictions on AI tools in contributing, please be transparent about their usage as it 
helps set the right context for reviewers.

Did you use AI tools to help write this code? _**< YES >**_

PR #1141, opened 2026-02-23
feat: wrapped text in GfxRender, implemented in themes so far

iandchasse:
## Summary

Conrgegate the changes of #1074 , #1013 , and extended upon #911 by @lkristensen
New function implemented in GfxRenderer.cpp
```C++
std::vector<std::string> GfxRenderer::wrappedText(const int fontId, const char* text, const int maxWidth,
                                                  const int maxLines, const EpdFontFamily::Style style) const
```
Applied logic to all uses in Lyra, Lyra Extended, and base theme (continue reading card as pointed out by @znelson 

## Additional Context

IMG_8604
IMG_8605
IMG_8606

---

### AI Usage


lib/GfxRenderer/GfxRenderer.h
```diff
@@ -115,6 +117,11 @@ class GfxRenderer {
   int getLineHeight(int fontId) const;
   std::string truncatedText(int fontId, const char* text, int maxWidth,
                             EpdFontFamily::Style style = EpdFontFamily::REGULAR) const;
+  /// Word-wrap \p text into at most \p maxLines lines, each no wider than
+  /// \p maxWidth pixels. Overflowing words and excess lines are UTF-8-safely
+  /// truncated with an ellipsis (U+2026).
+  std::vector<std::string> wrappedText(int fontId, const char* text, int maxWidth, int maxLines,
```
ngxson, line 128: nits, but should we have a `drawWrappedText()` wrapper for this? the idea is like `drawCenteredText()`, that will simply downstream code

ngxson: btw, is it possible to also return a bool to indicate if the text is truncated? something like `bool& truncated`

iandchasse: I can do that, let me see

iandchasse: @ngxson just getting around to looking, what do you mean by a wrapper?

CaptainFrito (approved):
Thank you so much for this!!

CaptainFrito:
Very cool device by the way

lib/GfxRenderer/GfxRenderer.h
ngxson: I mean the `wrappedText` only tell how to wrap the text, but not drawing on screen. A wrapper like `drawWrappedText` will call `wrappedText` under the hood, and also draw the text to screen

ngxson: (just optional though because I will need that for another PR)

ngxson (approved):

PR #1142, opened 2026-02-23
feat: Reading progress and time on recent books

confus-3d:
## Summary

Improve "Recent books" view on all themes, adding a new progress bar per book, and reading time/estimated reading time remaining.

It changes the layout on RecentBooksActivity. Working with classic and Lyra.
It adds a new file on the cache for each book, where it stores the total reading time of each book.
Changes in EPUB-TXT-XTC-ReaderActivity to track time.

## Additional Context
It keeps track of time on the reading screen only, without background processes.
Only tracks time once the feature has been activated.
If no file is found (book never opened with this feature), shows 0h 00m - ..

---

### AI Usage

Did you use AI tools to help write this code?  YES 

Imagen

### DIFF
--- a/src/activities/home/RecentBooksActivity.cpp
+++ b/src/activities/home/RecentBooksActivity.cpp
@@ -8,188 +8,10 @@
 #include <Txt.h>
 #include <Xtc.h>
 
-#include <algorithm>
-
 #include "MappedInputManager.h"
 #include "RecentBooksStore.h"
 #include "components/UITheme.h"
 #include "fontIds.h"
-#include "util/StringUtils.h"
-
-namespace {
-constexpr uint32_t TXT_INDEX_MAGIC = 0x54585449;  // "TXTI"
-constexpr uint8_t TXT_INDEX_VERSION_MIN = 1;
-constexpr uint8_t TXT_INDEX_VERSION_MAX = 2;
-constexpr char READING_TIME_FILE_NAME[] = "/reading_time.bin";
-
-uint32_t readSecondsFromCachePath(const std::string& cachePath) {
-  uint32_t seconds = 0;
-  FsFile f;
-  if (!Storage.openFileForRead("RBA", cachePath + READING_TIME_FILE_NAME, f)) {
-    return 0;
-  }
-
-  uint8_t data[4];
-  if (f.read(data, sizeof(data)) == 4) {
-    seconds = data[0] | (data[1] << 8) | (data[2] << 16) | (data[3] << 24);
-  }
-  f.close();
-  return seconds;
-}
-
-uint32_t loadReadSecondsForPath(const std::string& path) {
-  if (StringUtils::checkFileExtension(path, ".epub")) {
-    Epub epub(path, "/.crosspoint");
-    return readSecondsFromCachePath(epub.getCachePath());
-  }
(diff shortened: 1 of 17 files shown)


Levrk:
a lot of overlap with #1127, I do really like the bars though

src/activities/home/RecentBooksActivity.cpp
```diff
@@ -101,8 +311,40 @@ void RecentBooksActivity::render(Activity::RenderLock&&) {
...
+          if (index < 0 || static_cast<size_t>(index) >= readingMetrics.size()) return std::string();
+          const auto& m = readingMetrics[index];
+          const std::string percent =
+              (m.progressPercent >= 0) ? (std::to_string(clampProgressPercent(m.progressPercent)) + "%") : "--%";
+          const std::string readPart = formatDuration(m.readSeconds);
+          const std::string leftPart = m.hasRemainingEstimate ? formatDuration(m.remainingSeconds) : std::string("--");
+          return percent + " • " + readPart + " • " + leftPart;
+        },
         [this](int index) { return UITheme::getFileIcon(recentBooks[index].path); });
+
```
CaptainFrito, line 109: If we want to add support for additional UI elements in the list, let's add that logic to the list component please, instead of duplicating list drawing logic here

src/activities/home/RecentBooksActivity.cpp
```diff
@@ -13,7 +17,178 @@
...
+  uint8_t data[4];
+  if (f.read(data, sizeof(data)) == 4) {
+    seconds = data[0] | (data[1] << 8) | (data[2] << 16) | (data[3] << 24);
+  }
+  f.close();
+  return seconds;
+}
+
+uint32_t loadReadSecondsForPath(const std::string& path) {
```
CaptainFrito: This logic should not be in the Activity, but in RecentBookStore please.
I really dig the read timer idea, but could we please save it in the RecentBook struct along with the progress, rather than in the progress file for each book format? That way we won't need to find and open 10 different files upon entering this screen, and on the home screen.

confus-3d: Tried to change all the logic, now some draw logic is in themes and the math is o RecentBookStore

confus-3d:
Hey @CaptainFrito I did some changes in a new branch following your approach, but have some concerns about the implementation before deleting this PR and making a newone.

If we save reading time on RecentBookStore we can lose track of books already read.

If the book falls to 11 position, everything is deleted.

I think is better to keep track inside each book json and just read this 10 progress, it works fine and you don't lose the time (im saving the reading time) of each book in case someone wants to use that info for another feature? (Reader profile, reading achievements, whatever)

What do you think?

BTW i have another PR about Dark Mode in the reader mode in case you want to take a look =)

znelson:
We have a number of reading reading stats and time related PRs in #1142, #1196, #1233, #1251.

I think it would be good for us to have a discussion at https://github.com/crosspoint-reader/crosspoint-reader/discussions to come to a consensus on direction. Probably something like: 
- Which stats add real value?
- How do we collect and store those stats?
- Where do we display those stats?

itsthisjustin:
Considering the sheer amount of opportunity at a mistake in the merge with upstream I'm going to close this for now. Aesthetically I'll add that we should probably add more padding above the bars and align them to the icons properly if we do try this style UI again. I'd also recommend this as a setting instead of being default on (which I think it is?) as it definitely muddies up the UI.

PR #1156, opened 2026-02-24
chore: Removed generated language headers

znelson:
## Summary

I18nKeys.h and I18nStrings.h are generated by gen_i18n.py prior to each build, so we do not need to maintain a checked-in copy of these files.

---

### AI Usage


ngxson:
Btw, do we also need to add it to `.gitignore` ?

PR #1163, opened 2026-02-24
feat: runtime font fallback

ngxson:
## Summary

As mentioned in https://github.com/crosspoint-reader/crosspoint-reader/pull/1147#issuecomment-3952907426 : Ubuntu does not support vietnamese and other latin-ext-B languages, so we need to fallback to noto sans.

The problem is that fallback at conversion will result in having a set of the same glyphs being duplicated inside the firmware, wasting precious space (see https://github.com/crosspoint-reader/crosspoint-reader/pull/1124#discussion_r2841933291)

Better to be merged after https://github.com/crosspoint-reader/crosspoint-reader/pull/1157 

---

### AI Usage


mcrosson:

itsthisjustin:
Going to close this as SD fonts are merged and Noto Sans is the default UI font anyway now.

PR #1169, opened 2026-02-24
fix: clarity issue with ambiguous string `SET`

ariel-lindemann:
## Summary

Fixes a clarity issue regarding the translation string `STR_SET`. The issue lies in the fact that the english word can have different meanings. 

The only time the string is used is in the language selectio screen, where it has the meaning of _selected_.  (As in _The language has been **set** to French_).

Another meaning can be _configured_. (As in _The KOReader username has been __set__). This is the meaning many of the translations have taken. The reason that the string is right above `STR_NOT_SET` (which is meant as _not configured_).

With this PR I propose to explicitly use the term "_Selected_". There are two good reasons for this:
+ it removes the confusion and the misleading translations
+ it is consistent with the button label `Select`, communicating the link between the two (the row will be marked `Selected` if you press the buttpn `Select`. Much clearer than now)

Removed the unused strings and added translations for the new string `STR_SELECTED` for the languages I know.

tagging the translators for feedback:
fr: @Spigaw @CaptainFrito 
de: @DavidOrtmann
cs: @brbla 
pt: @yagofarias
it: @andreaturchet @fargolinux
ru: @madebykir @mrtnvgr 
es: @yeyeto2788 @Skrzakk @pablohc 
sv: @dawiik
ca: @angeldenom 
uj: @mirus-ua 
be: @dexif 

## Additional Context

  specific areas to focus on).

the Issue was introduced in #1020. Previously, if a language was selected it was marked with `[ON]` (`STR_ON_MARKER`). I considered reverting it back to that, but the solution I described above seemed superior.

---

### AI Usage


znelson (approved):

lib/I18n/translations/russian.yaml
mrtnvgr, line 228:
```suggestion
STR_SELECT: "Выбрать"
STR_SELECTED: "Выбран"
```

lib/I18n/translations/french.yaml
```diff
@@ -237,7 +237,6 @@ STR_YES: "Oui"
 STR_NO: "Non"
 STR_STATE_ON: "ON"
 STR_STATE_OFF: "OFF"
-STR_SET: "Défini"
 STR_NOT_SET: "Non défini"
 STR_DIR_LEFT: "Gauche"
```
CaptainFrito, line 241: "Sélectionné" would be the right word but that's very long, alternatively we could use "Actif" in this context. Not a fan of either honestly. What do you think @Spigaw ?

lib/I18n/translations/ukrainian.yaml
mirus-ua, line 228:
```suggestion
STR_SELECT: "Вибрати"
STR_SELECTED: "Вибрано"
```

PR #1184, opened 2026-02-25
refactor: Introduce HalNetwork abstraction for WiFi on/off control

jpirnay:
## Summary

* **What is the goal of this PR?** Replace the scattered per-file WiFi on/off calls with a proper HAL abstraction (`HalNetwork`), following the existing pattern of `HalDisplay`, `HalPowerManager`, `HalStorage`, and `HalGPIO`. This both unifies duplicate code and positions the codebase for future porting (e.g. X3) or emulator stubs — callers never see the ESP32 WiFi driver directly.

  * New `lib/hal/HalNetwork` class (singleton `network`) with four methods:
    * `enable()` — set STA mode (replaces bare `WiFi.mode(WIFI_STA)` calls)
    * `disable()` — graceful disconnect + delays + power off (replaces the previously duplicated teardown sequence)
    * `isConnected()` — canonical two-condition check (`WL_CONNECTED` **and** valid IP), replacing inconsistent per-file checks where some callers tested only `WiFi.status()` and others also checked the IP
    * `isActive()` — true when the radio is on in any mode (STA or AP); used by `HalPowerManager` to suppress CPU throttling
  * `HalPowerManager` now queries `network.isActive()` instead of calling `WiFi.getMode()` directly, completing its WiFi-free dependency graph.
  * `OpdsBookBrowserActivity` — two raw `WiFi.mode(WIFI_OFF)` calls replaced with `network.disable()`. One was missing the graceful disconnect entirely (bug fix); the other was missing the shutdown delays.
  * `KOReaderAuthActivity`, `KOReaderSyncActivity`, `CalibreConnectActivity` — `WiFi.status()` checks replaced with `network.isConnected()`, upgrading partial checks to the canonical two-condition form.
  * `KOReaderSyncActivity`, `KOReaderAuthActivity`, `OpdsBookBrowserActivity` — `#include <WiFi.h>` removed; these files now have no direct WiFi driver dependency.
  * `WifiHelpers.{h,cpp}` deleted — superseded entirely.

* **Intentionally left unchanged:**
  * `CrossPointWebServerActivity.cpp` — its shutdown path branches between `WiFi.softAPdisconnect()` (AP mode) and STA disconnect; this does not map cleanly to the current `HalNetwork` interface without AP-specific methods.
  * `WifiSelectionActivity.cpp` — scan, credential handling, `WiFi.begin()`, hostname setup: application-level logic, not raw hardware control.
  * `CrossPointWebServer.cpp` — checks both STA-connected and AP-active as separate conditions; the AP half requires `WiFi.softAPgetStationNum()` which `HalNetwork` does not yet abstract.
  * `CalibreConnectActivity.cpp` — `WiFi.localIP()` / `WiFi.SSID()` are read-only info queries used to display connection details, not hardware control.
  * `KOReaderSyncClient.cpp` — `#include <WiFi.h>` retained for `WiFiClient`/`WiFiClientSecure` type definitions.
  * SNTP lifecycle (`syncTimeWithNTP()`, `esp_sntp_stop()`) — remains in `KOReaderSyncActivity`; embedding it in `HalNetwork::disable()` would tie the HAL to an ESP32 service-SDK header, which is the opposite of the porting goal.

---

### AI Usage

### DIFF
--- a/lib/hal/HalNetwork.h
+++ b/lib/hal/HalNetwork.h
@@ -46,40 +46,4 @@ class HalNetwork {
    * two-condition test (status + valid IP) is applied consistently everywhere.
    */
   bool isConnected() const;
-
-  /**
-   * RAII guard: enables the network on construction, disables it on destruction.
-   *
-   * Prefer this over paired enable()/disable() calls so that the radio is
-   * guaranteed to be switched off even on early returns.
-   *
-   * Short-lived use (single function):
-   *   HalNetwork::Guard g(network);
-   *
-   * Activity lifetime (onEnter / onExit):
-   *   std::optional<HalNetwork::Guard> networkGuard_;
-   *   networkGuard_.emplace(network);  // onEnter
-   *   networkGuard_.reset();           // onExit
-   */
-  class Guard {
-   public:
-    explicit Guard(HalNetwork& net) : net_(&net) { net_->enable(); }
-    ~Guard() {
-      if (net_) net_->disable();
-    }
-    Guard(Guard&& other) noexcept : net_(other.net_) { other.net_ = nullptr; }
-    Guard& operator=(Guard&& other) noexcept {
-      if (this != &other) {
-        if (net_) net_->disable();
-        net_ = other.net_;
-        other.net_ = nullptr;
-      }
-      return *this;
-    }
-    Guard(const Guard&) = delete;
-    Guard& operator=(const Guard&) = delete;
-
-   private:
-    HalNetwork* net_;
-  };
 };
--- a/src/activities/settings/KOReaderAuthActivity.cpp
+++ b/src/activities/settings/KOReaderAuthActivity.cpp
@@ -51,10 +51,8 @@ void KOReaderAuthActivity::performAuthentication() {
 void KOReaderAuthActivity::onEnter() {
   Activity::onEnter();
 
-  // Enable WiFi and keep it on for the duration of this activity via RAII guard.
-  // Authentication may involve multiple requests and user interactions (e.g. captive portal).
-  // The guard calls network.disable() automatically when reset or destroyed.
-  networkGuard_.emplace(network);
+  // Turn on WiFi
+  network.enable();
 
   // Check if already connected
   if (network.isConnected()) {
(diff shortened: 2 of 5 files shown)


osteotek (approved):

ngxson:
While working on https://github.com/crosspoint-reader/crosspoint-reader/pull/1016 , I'm thinking that we could better converting our `WifiHelper` object to a RAII style object where constructor == wifiOn and destructor == wifiOff

The main point of this idea is that we can have an `ActivityResult` that contains `std::unique_ptr<WifiHelper>`, which enforce the `std::move(wifi)` object from a subactivity to main activity:

```cpp
// WifiSelectionActivity
result.wifi = std::move(wifi);
setResult(std::move(result));

// Other activities
startActivityForResult(std::make_unique<WifiSelectionActivity>(renderer, mappedInput),
                         [this](const ActivityResult& result) {
                           if (!result.isCancelled) {
                              // Move the object so it will be destroyed along with the current activity
                              this->selectedWifi = std::move(result.wifi);
                           }
                         });
```

If this sounds good to you, maybe we can wait until ActivityManager is merged?

src/network/WifiHelpers.h
ngxson, line 1: I think it can be a good opportunity to move this into a `HalNetwork`, and no longer calling it "wifi". The wifi is only one of the implementation of `HalNetwork` abstraction

For now, to simplify, the `HalNetwork` can contain only control on/off state. But in the future, we may be able to stub some responses, making it easier to be emulated

Same idea for other HAL classes already existing, it make porting crosspoint to other devices (like X3) and making emulator easier in the future.

jpirnay: Makes a lot of sense, so I picked up the idea

jpirnay:
> Not exactly what I have in mind but ok as a first step.
> 
> I would be more interested in have a RAII object where the destructor is the equivalent to `HalNetwork::disable()`

I have added the lifetime guard. Any feedback welcome, this is slightly out of my comfort zone

ngxson:
Hmm sorry I looked deeper into the code and realized that it's not that simple. `CalibreConnectActivity` has a logic to reuse wifi connection if it's already connected, that breaks my model of "activity must be the ownership of its network connection." Maybe other places has the same logic, so better to refactor them altogether.

I need more time to reflect on the over architecture. For now, could you revert https://github.com/crosspoint-reader/crosspoint-reader/pull/1184/commits/1ba425c12218ada06d53d9d046d5e2bb0fcb3f5f ? The version I reviewed before that commit was already good enough.

ngxson (approved):

PR #1188, opened 2026-02-25
fix: broken translations in status bar settings

ariel-lindemann:
## Summary

The strings for `Show` and `Hide` were always showing in English, regardless of which language was selected.

Replace the variables in the lambda by direct calls to the `tr` macro

## Additional Context

  specific areas to focus on).

---

### AI Usage

Did you use AI tools to help write this code? _**<NO >**_


whyte-j:
Hi Ariel,

~Can you double check that the languages you have used to identify this bug are ones that have the hide/show strings? The new strings for the status bar were only added for English in my PR (The others were agreed to be added at RC, but I see that a couple of language PRs have been merged since). If a string does not exist for a given language, it falls back to English - this may be what you're seeing here, rather than an error with the code itself.~

I've just tested myself, and you're absolutely right! I'm not entirely sure why this would be the case - I do wonder if it would work if we moved the translated string call to the render function? It wouldn't save as many calls as I was hoping, but it would save us from calling multiple times for the same string when we render.

Thanks for fixing!

PR #1194, opened 2026-02-26
perf: Optimize HTML entities lookup to O(log(n))

Uri-Tauber:
## Summary

**What is the goal of this PR?** Replace the linear scan of `lookupHtmlEntity` with a simple binary search to improve lookup performance.

`lib/Epub/Epub/Entities/htmlEntities.cpp`: 
 - Sorted the `ENTITY_LOOKUP` array.
 - Added a compile-time assertion to guarantee the array remains sorted.
 - Rewrote `lookupHtmlEntity` to use a binary search.

## Additional Context

Benchmarked on my x64 laptop (probably will be different on RISC-V)
```
=== Benchmark (53 entities x 10000 iterations) ===

Version           Total time   Avg per lookup
----------------------------------------------
linear          236.97 ms total     447.11 ns/lookup
binary search    22.09 ms total      41.68 ns/lookup

=== Summary ===

Binary search is 10.73x faster than linear scan.
```

This is a simplified alternative to #1180, focused on keeping the implementation clean, and maintainable.

### AI Usage

### DIFF
--- a/lib/Epub/Epub/htmlEntities.cpp
+++ b/lib/Epub/Epub/htmlEntities.cpp
@@ -1,4 +1,4 @@
-// from
+// based on
 // https://github.com/atomic14/diy-esp32-epub-reader/blob/2c2f57fdd7e2a788d14a0bcb26b9e845a47aac42/lib/Epub/RubbishHtmlParser/htmlEntities.cpp
 
 #include "htmlEntities.h"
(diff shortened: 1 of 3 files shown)


lib/Epub/Epub/parsers/ChapterHtmlSlimParser.cpp
```diff
@@ -685,7 +685,7 @@ void XMLCALL ChapterHtmlSlimParser::characterData(void* userData, const XML_Char
 void XMLCALL ChapterHtmlSlimParser::defaultHandlerExpand(void* userData, const XML_Char* s, const int len) {
   // Check if this looks like an entity reference (&...;)
   if (len >= 3 && s[0] == '&' && s[len - 1] == ';') {
-    const char* utf8Value = lookupHtmlEntity(s, len);
+    const char* utf8Value = lookupHtmlEntity(s, (size_t)len);
```
znelson: Minor nit-pick, `static_cast` is better for modern C++ code to avoid accidentally doing any dangerous casting. (There's no real risk here, just a thing that I've had to pay attention to in other codebases.)

```suggestion
    const char* utf8Value = lookupHtmlEntity(s, static_cast<size_t>(len));
```

If you make lookupHtmlEntity take a `string_view`, this would become:
```c++
const char* utf8Value = lookupHtmlEntity({s, static_cast<size_t>(len)});
```

lib/Epub/Epub/htmlEntities.cpp
```diff
@@ -10,67 +10,105 @@ struct EntityPair {
...
+static constexpr bool isTableSorted() {
+  for (size_t i = 1; i < ENTITY_LOOKUP_COUNT; i++) {
+    if (constexprStrcmp(ENTITY_LOOKUP[i - 1].key, ENTITY_LOOKUP[i].key) >= 0) return false;
+  }
+  return true;
+}
+static_assert(isTableSorted(), "ENTITY_LOOKUP must be sorted lexicographically by key");
+
+// Lookup a single HTML entity and return its UTF-8 value.
+const char* lookupHtmlEntity(const char* entity, size_t len) {
```
znelson, line 84: If you want, you can rewrite this with `std::lower_bound` binary search, and to take a `string_view` param. (I will do the same std algorithm changes for kerning and ligature binary searches in EpdFont.cpp.)

```c++
const char* lookupHtmlEntity(std::string_view entity) {
  if (entity.empty()) {
    return nullptr;
  }

  // lower_bound: exact-key lookup. Finds the first entry with key >= entity,
  // then the equality check confirms an exact match exists.
  const auto it = std::lower_bound(
      std::begin(ENTITY_LOOKUP), std::end(ENTITY_LOOKUP), entity,
      [](const EntityPair& entry, std::string_view value) {
        return std::string_view(entry.key) < value;
      });

  if (it != std::end(ENTITY_LOOKUP) && std::string_view(it->key) == entity) {
    return it->value;
  }

  return nullptr;
}
```

(`string_view` `operator==` does the equivalent of `memcmp`.)

znelson (dismissed):

znelson (approved):

lib/Epub/Epub/htmlEntities.cpp
Uri-Tauber: I was under the impression that the firmware was compiled using the gnu++11 standard, but it seems I was mistaken.
Since this isn’t really introducing a change, I would prefer to leave the code as it is. In my opinion, the manual binary search is more readable than using `std::lower_bound` (no need to check documentation).

osteotek (approved):

Issue #1191, opened 2026-02-26
Feature request: Indication in the file browser if a book have been read/completed


Rufaim:
Currently, when I open a browser, I cannot distinguish if I read a book or not. I request to add some indication. I believe, it will improve usability.

In the simplest case it could look like a checkmark if a book file has ever been opened. And perhaps a percentage of completion.

If this feature is implemented in a more elaborate way, there may be a way for the user to indicate if the book is finished. Possibly with a button or automatically after they reach the end of the file. Thus, in the browser, one will be able to see (i) if they finished the book and (ii) what percentage they finished.

PR #1201, opened 2026-02-26
feat: add 'opened' indicator to on-device file browser

Nosh118:
## Summary
Addresses #1191 (in part)
  Adds a visual indicator to the on-device `MyLibraryActivity` file browser to show which books have been previously opened.

  * Created `ReadProgressUtil` to centralise cache path resolution (`.crosspoint/<type>_<hash>/progress.bin`).
  * Extended `MyLibraryActivity` to track opened status for files during directory scanning.
  * Passed a `rowValue` callback to `GUI.drawList()` to render a bullet (`•`) indicator on the right side of the list row for opened books.

## Additional Context

**Implementation Design / Missing Features**
* **Why no completion percentage?** Currently, it's not possible to accurately calculate a completion percentage without loading the full book metadata into memory. While `progress.bin` exists for all formats, the schema varies by reader (EPUB is 6 bytes holding chapter info, TXT/XTC are 4 bytes holding absolute pages). The EPUB reader needs to parse spine item sizes from `book.bin` to calculate a true book-level percentage. 
* **Why no "Finished" state?** Similarly, detecting a "completed" state purely from the file system isn't currently possible without either modifying the native reader activities to save a separate `finished` marker file when the end-of-book is reached, or modifying the core `progress.bin` schemas.
* **Why duplicate hash logic?** `ReadProgressUtil` duplicates the `std::hash<std::string>{}(filepath)` cache path generation logic found in the `Epub`, `Txt`, and `Xtc` constructors. This was done intentionally to keep the change footprint minimally invasive, allowing us to perform quick file-existence checks during directory scanning without requiring us to instantiate the heavier reader classes or refactor their internals.

---

### AI Usage

### DIFF
--- a/src/activities/home/MyLibraryActivity.cpp
+++ b/src/activities/home/MyLibraryActivity.cpp
@@ -9,6 +9,7 @@
 #include "MappedInputManager.h"
 #include "components/UITheme.h"
 #include "fontIds.h"
+#include "util/ReadProgressUtil.h"
 #include "util/StringUtils.h"
 
 namespace {
@@ -100,6 +101,19 @@ void MyLibraryActivity::loadFiles() {
   }
   root.close();
   sortFileList(files);
+
+  // Look up opened status for each file
+  fileOpened.clear();
+  for (const auto& filename : files) {
+    if (filename.back() == '/') {
+      fileOpened.push_back(false);  // Directories have no progress
+    } else {
+      std::string fullPath = basepath;
+      if (fullPath.back() != '/') fullPath += "/";
+      fullPath += filename;
+      fileOpened.push_back(ReadProgressUtil::hasBeenOpened(fullPath));
+    }
+  }
 }
 
 void MyLibraryActivity::onEnter() {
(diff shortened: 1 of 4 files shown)


Nosh118:
Happy to attempt to implement % tracking - could be done by saving % complete on page turn in a separate binary file, but otherwise involves modifying the structure of progress.bin to include it. Similarly changes to progress.bin would be needed to store an end of book flag (which could also be manually triggered via the options menu in reader).

Wanted to keep changes light for now before modifying the structure.

Nosh118:
Tested on device, appears to work perfectly, incl. with overflow text.

Keen to hear what the consensus is around modifying the progress.bin file for % etc., but I guess dependent on #1142 & #1127 in terms of implementation method.

itsthisjustin:
@Nosh118 I love the intent of this PR but considering the lack of movement since Feb I'm going to close for now. I think the community would prefer we just show progress vs a simple opened state and if we do move forward with that, it solves this same issue. I super appreciate the thought you put into why not progress indicators as I think that will help drive future development in this area

PR #1199, opened 2026-02-26
chore: new Ukrainian translation lines

mirus-ua:
## Summary

* **What is the goal of this PR?** 
Update translation after https://github.com/crosspoint-reader/crosspoint-reader/pull/733

---

### AI Usage

While CrossPoint doesn't have restrictions on AI tools in contributing, please be transparent about their usage as it 
helps set the right context for reviewers.

Did you use AI tools to help write this code? _**NO**_

PR #1203, opened 2026-02-26
docs: update readme.md with i18n checklist

mirus-ua:
## Summary

Introduce a new code of conduct rule when a developer adds new translation strings

### The goals are:
- reduce the number of new PRs with minor changes in the i18n string
- call to action for all translators
- couple a feature with translations

Only`README.md`

#### Why do I think this is important?
New features, such https://github.com/crosspoint-reader/crosspoint-reader/pull/733 introduce a lot of changes for i18n strings, and it follows with many new PR, like these https://github.com/crosspoint-reader/crosspoint-reader/pull/1199 https://github.com/crosspoint-reader/crosspoint-reader/pull/1198 

IMO, it brings unnecessary cognitive complexity for the maintainers to validate each and every PR with i18n strings instead of focusing on the PRs with new functionality

My suggestion is to embrace the case https://github.com/crosspoint-reader/crosspoint-reader/pull/1169, where in a PR with a new i18n string, the contributor tags all the translators from the https://github.com/crosspoint-reader/crosspoint-reader/blob/master/docs/translators.md file, and then translators can [suggest](https://github.com/crosspoint-reader/crosspoint-reader/pull/1169#discussion_r2851521004) correct translations within a single PR

Sure, we need to set a limited timeframe for the translators to suggest new strings. This is required to prevent unnecessary PR merge blocking. On the other hand, we minimize cases when a new firmware version is released without proper translation until the next version 

Open to a discussion, maybe we can come up with a better solution instead of this one, but the most important thing is to meet the goals
---

### AI Usage

### DIFF
--- a/README.md
+++ b/README.md
@@ -169,6 +169,11 @@ principles, please see [GOVERNANCE.md](GOVERNANCE.md).
 3. Make changes
 4. Submit a PR
 
+#### i18n checklist
+- your new translation string at the bottom of `english.yaml`
+- you tagged the [translators](https://github.com/crosspoint-reader/crosspoint-reader/blob/master/docs/translators.md) in your PR 
+- accept `Suggested change` related to other languages
+
 ---
 
 CrossPoint Reader is **not affiliated with Xteink or any manufacturer of the X4 hardware**.


ariel-lindemann:
Hey mirius,
 I have been thinking about the same problem these days as there has been a lot of activity around the i18n strings lately and we don't have a set policy yet of how to handle it the best.

Thank you for taking the time to write a constructive solutuion. 

I am wondering if we can't add a degree of automation to this. Maybe a github actions trigger to tag the translators automatically when a new string is added or deleted from `english.yml` for example. Because in general I think we want to make it as easy as possilble for people to contribute and if we add another rule maybe that's a bit too much for some (or maybe they won't even read it).

Another thing we already have: the  `gen_i18n.py` script already lists which strings need translations in which languages. Maybe we could use the output of that to have some sort of status badge or translations dashboard

This is just my perspective. Happy to hear what thers have to say.

mirus-ua:
> Hey mirius, I have been thinking about the same problem these days as there has been a lot of activity around the i18n strings lately and we don't have a set policy yet of how to handle it the best.
> 
> Thank you for taking the time to write a constructive solutuion.
> 
> I am wondering if we can't add a degree of automation to this. Maybe a github actions trigger to tag the translators automatically when a new string is added or deleted from `english.yml` for example. Because in general I think we want to make it as easy as possilble for people to contribute and if we add another rule maybe that's a bit too much for some (or maybe they won't even read it).
> 
> Another thing we already have: the `gen_i18n.py` script already lists which strings need translations in which languages. Maybe we could use the output of that to have some sort of status badge or translations dashboard
> 
> This is just my perspective. Happy to hear what thers have to say.

This is a nice perspective as well, and most likely we can combine the two ideas, like that:

Instead of forcing a contributor to ping all translators, we can take the output of `gen_i18n.py` and automatically ping people. Most likely, we can do it even w/o an extra GHA and use CodeRabbit with a proper prompt. 

On the other hand, I still think that coupling as many translation strings as possible within OG PR is the right approach

Uri-Tauber:
What if we configured CodeRabbit/Copilot to automatically merge PRs that only affect `yaml` files? What do you think?
It will be less work for the contributor, while removing the maintainers extra work.

mirus-ua:
> What if we configured CodeRabbit/Copilot to automatically merge PRs that only affect `yaml` files? What do you think? It will be less work for the contributor, while removing the maintainers extra work.

BTW, this is also a good idea, IMO 

Only one question: do you see how we can keep the translation percentage high for releases? One thing that I wish to avoid is a disconnect between a feature release and a translation for it

Levrk:
I would think it would be a bit of a burden on translators to have to review every single PR with translation implications, as well being a bottleneck in the review process. Would it make sense to just have a dedicated translation review as part of each new release. Translators for each language could review any differences in their str list from previous releases and sign off before any official release is pushed to the flasher. 

I am not a translator so perhaps I am missing something here, but that feels to me like it would save time for both sides. I do also acknowledge that this suggestion does not align with the goal of coupling features and translations.

mirus-ua:
> I would think it would be a bit of a burden on translators to have to review every single PR with translation implications, as well being a bottleneck in the review process. Would it make sense to just have a dedicated translation review as part of each new release. Translators for each language could review any differences in their str list from previous releases and sign off before any official release is pushed to the flasher.
> 
> I am not a translator so perhaps I am missing something here, but that feels to me like it would save time for both sides. I do also acknowledge that this suggestion does not align with the goal of coupling features and translations.

Don't get me wrong. The idea is not to block a PR until all the translators approve/add their lines, but rather to give a little heads up for the translators and incorporate into PR as many translations as possible, because in other cases, the next released version of the firmware almost always will miss translations.

Uri-Tauber:
> BTW, this is also a good idea, IMO
> 
> Only one question: do you see how we can keep the translation percentage high for releases? One thing that I wish to avoid is a disconnect between a feature release and a translation for it

We could approach this on two fronts:

* Configure CodeRabbit to automatically approve and merge PRs that only modify .yaml files.

* Have CodeRabbit maintain and continuously update a list of translators based on previous YAML-related PRs. When a new PR introduces additional strings, CodeRabbit would automatically tag the relevant translators from that list to notify them. They could then either suggest translations directly on the PR or open a follow-up PR with the necessary updates.

@coderabbitai do you think you up to this?

Issue #632, opened 2026-01-31
OPDS file download not working


tilleule:
### Affected Version

1.60.0

### Describe the Bug

I can log into my OPDS server (I'm using RSSPUB) but every download fails.
 
I can transfer the same file via Web-upload and it is displayed fine on the reader. So the issue seems to be OPDS. Are there any logs I can submit?

### Steps to Reproduce

1. go to OPDS Browser
2. connect to server and see list of files provided
3. select file and press Download
4. for a split second the screen reads "downloading ..." but immediately turns into "Error: Download failed"

### Expected Behavior

I'd expect the file to download, obviously.

### Relevant Log Output/Screenshots

```shell

```

tilleule:
Additional information: the files are automatically named "rss_digest_20260201_100645.epub", can it be the long file name that prevents the download? I wish there were logs to check.

teranex:
Is this fix included in v1.1.1?
I'm running 1.1.1 and just tried to setup the OPDS server (through the web ui, yay!). I use Readeck (https://readeck.org/en/) as a read-it-later tool, which supports OPDS out of the box to deliver each article as an epub. Browsing the articles on the X4 works, but downloading then is stuck on 0% until it says the download failed. Urls in the atom xml from Readeck also contain absolute url's to the epubs.

osteotek:
@teranex Fix is included in 1.1.1, yes

You can see logs online using this website https://pineterm.link/ (only works in Chrome-based browsers). Connect your X4 while it's in sleeping mode to a computer, power it on, press `Connect UART button`, select serial port. After that you should be able to see logs in the window below, press `ASCII View` to view logs in text format. Try downloading book from OPDS and share logs here

teranex:
That is a cool tool :)

These are the logs from the moment I pressed the download button until teh X4 told me the download failed. (Due to the huge amount of similar lines I have cut out part of it):

```
[...]

PR #1207, opened 2026-02-26
fix: use HTTPClient::writeToStream for downloading files from OPDS

osteotek:
## Summary

* Refactored `HttpDownloader::downloadToFile` to use `FileWriteStream` and `HTTPClient::writeToStream`, removing manual chunked downloading logic, which was error-prone.
* Fixes https://github.com/crosspoint-reader/crosspoint-reader/issues/632

## Additional Context

* Tested downloading files from OPDS with a size up to 10 mb.

---

### AI Usage

### DIFF
--- a/src/network/HttpDownloader.cpp
+++ b/src/network/HttpDownloader.cpp
@@ -157,12 +157,17 @@ HttpDownloader::DownloadError HttpDownloader::downloadToFile(const std::string& 
   }
 
   // Let HTTPClient handle chunked decoding and stream body bytes into the file.
-  const size_t total = contentLength > 0 ? contentLength : 0;
-  FileWriteStream fileStream(file, total, progress);
-  http.writeToStream(&fileStream);
+  FileWriteStream fileStream(file, contentLength, progress);
+  const int writeResult = http.writeToStream(&fileStream);
 
   file.close();
   http.end();
+
+  if (writeResult < 0) {
+    LOG_ERR("HTTP", "writeToStream error: %d", writeResult);
+    Storage.remove(destPath.c_str());
+    return HTTP_ERROR;
+  }
 
   const size_t downloaded = fileStream.downloaded();
   LOG_DBG("HTTP", "Downloaded %zu bytes", downloaded);
--- a/src/network/HttpDownloader.h
+++ b/src/network/HttpDownloader.h
@@ -38,7 +38,4 @@ class HttpDownloader {
    */
   static DownloadError downloadToFile(const std::string& url, const std::string& destPath,
                                       ProgressCallback progress = nullptr);
-
- private:
-  static constexpr size_t DOWNLOAD_CHUNK_SIZE = 1024;
 };


src/network/HttpDownloader.cpp
```diff
@@ -112,56 +156,30 @@ HttpDownloader::DownloadError HttpDownloader::downloadToFile(const std::string&
...
-  if (!stream) {
-    LOG_ERR("HTTP", "Failed to get stream");
-    file.close();
-    Storage.remove(destPath.c_str());
-    http.end();
-    return HTTP_ERROR;
-  }
-
-  // Download in chunks
-  uint8_t buffer[DOWNLOAD_CHUNK_SIZE];
```
znelson, line 126: `DOWNLOAD_CHUNK_SIZE` is now unused, can be removed from HttpDownloader.h

src/network/HttpDownloader.cpp
```diff
@@ -112,56 +156,30 @@ HttpDownloader::DownloadError HttpDownloader::downloadToFile(const std::string&
...
-  size_t downloaded = 0;
+  // Let HTTPClient handle chunked decoding and stream body bytes into the file.
   const size_t total = contentLength > 0 ? contentLength : 0;
```
znelson: It's not new in this change, but if contentLength is unsigned (size_t), isn't this equivalent to just assigning to total without the ternary expression?

src/network/HttpDownloader.cpp
```diff
@@ -112,56 +156,30 @@ HttpDownloader::DownloadError HttpDownloader::downloadToFile(const std::string&
...
-    }
-
-    downloaded += bytesRead;
-
-    if (progress && total > 0) {
-      progress(downloaded, total);
-    }
-  }
+  FileWriteStream fileStream(file, total, progress);
+  http.writeToStream(&fileStream);
```
znelson: Should we consider handling `writeToStream` failure?

```suggestion
  const int writeResult = http.writeToStream(&fileStream);

  file.close();
  http.end();

  if (writeResult < 0) {
    LOG_ERR("HTTP", "writeToStream error: %d", writeResult);
    Storage.remove(destPath.c_str());
    return HTTP_ERROR;
  }
```

znelson (approved):

src/network/HttpDownloader.cpp
osteotek: removed

src/network/HttpDownloader.cpp
osteotek: yes, removed `total` variable

src/network/HttpDownloader.cpp
osteotek: thank you, added

osteotek:
@znelson thank you for the suggestions, addressed all of them

znelson (approved):

ngxson (approved):

PR #1209, opened 2026-02-26
feat: Support for multiple OPDS servers

### LINKED ISSUE #1178: Support for multiple OPDS catalogs
Having support for multiple OPDS catalogs would be great. For instance, users could add their personal Calibre, Project Gutenberg, and Standard Ebooks all at once.

osteotek:
## Summary

* Add support for configuring and using multiple OPDS servers, replacing the previous single-server limitation. Closes https://github.com/crosspoint-reader/crosspoint-reader/issues/1178
* New OpdsServerStore singleton (modeled after WifiCredentialStore) that persists up to 8 OPDS servers to /.crosspoint/opds.json with MAC-based password obfuscation.
* One-time migration from legacy single-server fields in CrossPointSettings to the new store on first boot.
* New OpdsServerListActivity for the device UI — works in two modes: a settings list (add/edit/delete servers) and a picker (select which server to browse). When only one server is configured, the picker is skipped automatically.
* Renamed CalibreSettingsActivity → OpdsSettingsActivity for clarity. It now edits individual OpdsServer entries (name, URL, username, password, delete).
* OpdsBookBrowserActivity now receives an OpdsServer at construction and uses its credentials for all fetches/downloads, and shows the server name in the header.
* HttpDownloader::fetchUrl and downloadToFile accept optional per-call username/password parameters instead of reading from global settings.
* REST API endpoints on CrossPointWebServer: GET /api/opds, POST /api/opds, POST /api/opds/delete — passwords are never exposed over the API (only a hasPassword flag), and omitting the password field on update preserves the existing one.
* Web UI (SettingsPage.html) with dynamic OPDS server management cards — add, edit, save, and delete servers from the browser.
SCR-20260416-stvu

## Additional Context

* The OpdsServerStore JSON format and obfuscation scheme are identical to WifiCredentialStore, so the same JsonSettingsIO infrastructure handles both.
* The web API uses POST /api/opds/delete instead of DELETE /api/opds because the ESP32 WebServer doesn't support the DELETE method with a request body.
* Existing single-server configurations are migrated automatically — no user action required. After migration the legacy CrossPointSettings fields are cleared so it only runs once.
* The HttpDownloader changes are backward-compatible: the credential parameters default to empty strings, so existing callers are unaffected.

---

### AI Usage


jetpacktuxedo:
Would loved to see this merged! I'll probably pull this branch and flash it myself to test in the next few days!

osteotek:
@crosspoint-reader/firmware-maintainers team, this PR is ready for review

znelson (approved):
@osteotek thanks for being so patient and diligent in keeping this up-to-date. I'm not an OPDS user, but this looks to me like a solid improvement for users. Similar to @pablohc and the keyboard stuff, I'll look to you to help with anything that might pop up as a result.

Issue #1137, opened 2026-02-23
Crash on releasing SPI mutex from incorrect task


znelson:
### Affected Version

1.1.1

### Describe the Bug

**_Issue analysis and crash symbolication with help from Claude:_**

### Summary

The device crashes with a FreeRTOS mutex ownership assertion when `EpubReaderChapterSelectionActivity::render()` reads TOC entries from the SD card. The render function, running on the dedicated `ActivityManagerRender` FreeRTOS task, performs SD card I/O via SPI during rendering. The SD card and the e-ink display share the same SPI bus, which is protected by a FreeRTOS mutex. The crash occurs when the SdFat library's `SPI.endTransaction()` tries to release the SPI bus mutex, but the mutex is owned by a different task (the main `loopTask`).

### Symbolicated Stack Trace

Crash on the **render task** (`ActivityManagerRender`, priority 1):

```
ActivityManager::renderTaskLoop()                         — ActivityManager.cpp:38
  └─ EpubReaderChapterSelectionActivity::render()         — EpubReaderChapterSelectionActivity.cpp:126
       └─ Epub::getTocItem(int)                           — Epub.cpp:781
            └─ BookMetadataCache::getTocEntry(int)        — BookMetadataCache.cpp:428
                 └─ BookMetadataCache::readTocEntry()     — BookMetadataCache.cpp:442
                      └─ serialization::readString()      — Serialization.h:49
                           └─ FatFile::readPrivate()      — FatFile.cpp:837
                                └─ FsCache::prepare()     — FsCache.cpp:41
                                     └─ SdSpiCard::readSectors()  — SdSpiCard.cpp:494
                                          └─ SdSpiCard::readData()  — SdSpiCard.cpp:402
                                               └─ SdSpiCard::spiStop()  — SdSpiCard.cpp:601
                                                    └─ SPIClass::endTransaction()  — SPI.cpp:195
                                                         └─ xQueueGenericSend  — queue.c:832  ← ASSERTION FAILURE
```

FreeRTOS assertion:
```
pxQueue->pcHead != NULL || pxQueue->u.xSemaphore.xMutexHolder == NULL ||
pxQueue->u.xSemaphore.xMutexHolder == xTaskGetCurrentTaskHandle()
```

Translation: the calling task tried to release (`xSemaphoreGive`) a mutex that is currently held by a **different** task.

### The Two Tasks and SPI

[...]

PR #1212, opened 2026-02-26
fix: make file system operations thread-safe (HalFile)

ngxson:
## Summary

Fix https://github.com/crosspoint-reader/crosspoint-reader/issues/1137

Introducing `HalFile`, a thin wrapper around `FsFile` that uses a global mutex to protect file operations.

To test this PR, place the code below somewhere in the code base (I placed it in `onGoToRecentBooks`)

```cpp
static auto testTask = [](void* param) {
  for (int i = 0; i < 10; i++) {
    String json = Storage.readFile("/.crosspoint/settings.json");
    LOG_DBG("TEST_TASK", "Read settings.json, bytes read: %u", json.length());
  }
  vTaskDelete(nullptr);
};
xTaskCreate(testTask, "test0", 8192, nullptr, 1, nullptr);
xTaskCreate(testTask, "test1", 8192, nullptr, 1, nullptr);
xTaskCreate(testTask, "test2", 8192, nullptr, 1, nullptr);
xTaskCreate(testTask, "test3", 8192, nullptr, 1, nullptr);
delay(1000);
```

It will reliably lead to crash on `master`, but will function correctly with this PR.

A macro renaming trick is used to avoid changing too many downstream code files.

---

### AI Usage

Did you use AI tools to help write this code? **PARTIALLY**, only to help with tedious copy-paste tasks

### DIFF
--- a/lib/hal/HalStorage.cpp
+++ b/lib/hal/HalStorage.cpp
@@ -129,10 +129,14 @@ bool HalStorage::removeDir(const char* path) { HAL_STORAGE_WRAPPED_CALL(removeDi
   assert(impl != nullptr);                 \
   return impl->file.method(__VA_ARGS__);
 
+#define HAL_FILE_FORWARD_CALL(method, ...) \
+  assert(impl != nullptr);                 \
+  return impl->file.method(__VA_ARGS__);
+
 void HalFile::flush() { HAL_FILE_WRAPPED_CALL(flush, ); }
 size_t HalFile::getName(char* name, size_t len) { HAL_FILE_WRAPPED_CALL(getName, name, len); }
-size_t HalFile::size() { return impl->file.size(); }          // already thread-safe, no need to wrap
-size_t HalFile::fileSize() { return impl->file.fileSize(); }  // already thread-safe, no need to wrap
+size_t HalFile::size() { HAL_FILE_FORWARD_CALL(size, ); }          // already thread-safe, no need to wrap
+size_t HalFile::fileSize() { HAL_FILE_FORWARD_CALL(fileSize, ); }  // already thread-safe, no need to wrap
 bool HalFile::seek(size_t pos) { HAL_FILE_WRAPPED_CALL(seekSet, pos); }
 bool HalFile::seekCur(int64_t offset) { HAL_FILE_WRAPPED_CALL(seekCur, offset); }
 bool HalFile::seekSet(size_t offset) { HAL_FILE_WRAPPED_CALL(seekSet, offset); }
@@ -143,13 +147,9 @@ int HalFile::read() { HAL_FILE_WRAPPED_CALL(read, ); }
 size_t HalFile::write(const void* buf, size_t count) { HAL_FILE_WRAPPED_CALL(write, buf, count); }
 size_t HalFile::write(uint8_t b) { HAL_FILE_WRAPPED_CALL(write, b); }
 bool HalFile::rename(const char* newPath) { HAL_FILE_WRAPPED_CALL(rename, newPath); }
-bool HalFile::isDirectory() const { return impl->file.isDirectory(); }  // already thread-safe, no need to wrap
+bool HalFile::isDirectory() const { HAL_FILE_FORWARD_CALL(isDirectory, ); }  // already thread-safe, no need to wrap
 void HalFile::rewindDirectory() { HAL_FILE_WRAPPED_CALL(rewindDirectory, ); }
-bool HalFile::close() {
-  HalStorage::StorageLock lock;
-  if (!impl) return false;
-  return impl->file.close();
-}
+bool HalFile::close() { HAL_FILE_WRAPPED_CALL(close, ); }
 HalFile HalFile::openNextFile() {
   HalStorage::StorageLock lock;
   assert(impl != nullptr);


lib/hal/HalStorage.h
```diff
@@ -1,10 +1,13 @@
 #pragma once
 
-#include <FS.h>  // need to be included before SdFat.h for compatibility with FS.h's File class
-#include <SDCardManager.h>
+#include <common/FsApiConstants.h>  // for oflag_t
 
+#include <memory>
+#include <string>
 #include <vector>
```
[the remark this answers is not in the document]
ngxson: added Print.h, WString will be included automatically

lib/hal/HalStorage.h
```diff
@@ -25,31 +30,71 @@ class HalStorage {
...
+  bool openFileForWrite(const char* moduleName, const char* path, HalFile& file);
+  bool openFileForWrite(const char* moduleName, const std::string& path, HalFile& file);
+  bool openFileForWrite(const char* moduleName, const String& path, HalFile& file);
   bool removeDir(const char* path);
 
   static HalStorage& getInstance() { return instance; }
 
+  class StorageLock;  // private class, used internally
+
  private:
```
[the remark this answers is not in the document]
ngxson: that means the implementation of the class is private

lib/hal/HalStorage.h
```diff
@@ -25,31 +30,71 @@ class HalStorage {
...
+  SemaphoreHandle_t storageMutex = nullptr;
 };
 
 #define Storage HalStorage::getInstance()
 
+class HalFile : public Print {
+ public:
+  class Impl;
+  std::shared_ptr<Impl> impl;
+
```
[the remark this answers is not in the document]
ngxson: it's to allow `fromFsFile` to construct it. otherwise, `FsFile` need to be on the constructor of `HalFile`, meaning it must be defined in `HalStorage.h`, and we don't want that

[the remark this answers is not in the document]
ngxson: `fromFsFile(void* fsFile)`, the `void *` looks ugly and I don't think `std::move` can be used here, we should keep it for clarity

lib/hal/HalStorage.h
```diff
@@ -25,31 +30,73 @@ class HalStorage {
...
+  bool isOpen() const;
+  operator bool() const;
+};
+
+// Only do renaming FsFile to HalFile if this header is included by downstream code
+// The renaming is to allow using the thread-safe HalFile instead of the raw FsFile, without needing to change the
+// downstream code
+#ifndef HAL_STORAGE_IMPL
+#define FsFile HalFile
+#endif
```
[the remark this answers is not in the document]
ngxson: it's an acceptable trade-off for now, a proper renaming will be done in the future

lib/hal/HalStorage.cpp
```diff
@@ -1,67 +1,157 @@
...
 
-FsFile HalStorage::open(const char* path, const oflag_t oflag) { return SDCard.open(path, oflag); }
+class HalFile::Impl {
+ public:
+  Impl(FsFile&& fsFile) : file(std::move(fsFile)) {}
+  FsFile file;
+};
 
-bool HalStorage::mkdir(const char* path, const bool pFlag) { return SDCard.mkdir(path, pFlag); }
+HalFile::HalFile() : impl(std::make_shared<HalFile::Impl>(FsFile())) {}
```
znelson: Dynamic allocation on construction is not ideal. Could you leave impl empty unless we call `fromFsFile`?

lib/hal/HalStorage.h
```diff
@@ -25,31 +30,73 @@ class HalStorage {
...
 };
 
 #define Storage HalStorage::getInstance()
 
+class HalFile : public Print {
+ public:
+  // The impl class is public to allow fromFsFile() to construct it
+  // Otherwise, we will need to include FsFile in this header, which is undesirable
+  class Impl;
+  std::shared_ptr<Impl> impl;
```
znelson: Do you mean for the pimpl to be shared ownership or should we use `unique_ptr`?

If you do mean for it to be shared, what does it mean for a copied HalFile to close the FsFile, e.g.:
```c++
HalFile a = Storage.open("/test.txt");
HalFile b = a;
b.close();
a.read(buf, 10);
```

If you don't mean for it to be shared, we should either delete or override the copy constructor and copy assignment operator.

lib/hal/HalStorage.h
```diff
@@ -25,31 +30,73 @@ class HalStorage {
...
+
+  explicit HalFile();
+  ~HalFile() = default;
```
znelson: If you use a unique_ptr for pimpl, this `= default` will cause a compile problem because it can't see the Impl destructor here. The fix is to move the `= default` to the .cpp, and change this to:

```suggestion
  ~HalFile();
```

lib/hal/HalStorage.h
```diff
@@ -25,31 +30,73 @@ class HalStorage {
...
   bool initialized = false;
+  SemaphoreHandle_t storageMutex = nullptr;
 };
 
 #define Storage HalStorage::getInstance()
 
+class HalFile : public Print {
+ public:
+  // The impl class is public to allow fromFsFile() to construct it
+  // Otherwise, we will need to include FsFile in this header, which is undesirable
```
znelson: This comment isn't correct. You could make Impl and impl private without including FsFile in this header, but you'd need to add `friend class HalStorage;`. Probably a reasonable tradeoff.

lib/hal/HalStorage.cpp
```diff
@@ -1,67 +1,157 @@
...
+  return impl->file.isDirectory();
+}
+void HalFile::rewindDirectory() { HAL_FILE_WRAPPED_CALL(rewindDirectory, ); }
+bool HalFile::close() { HAL_FILE_WRAPPED_CALL(close, ); }
+HalFile HalFile::openNextFile() {
+  HalStorage::StorageLock lock;
+  assert(impl != nullptr);
+  return fromFsFile(impl->file.openNextFile());
+}
+bool HalFile::isOpen() const { HAL_FILE_WRAPPED_CALL(isOpen, ); }
```
znelson: Up to you whether you want to change this to be less careful based on some internal knowledge of FsFile: `isOpen`, `isDirectory`, `size`, `fileSize`, `position`, and `available` all just return cached values from with FsFile, never actually access the SPI bus, so don't technically need the StorageLock.

znelson:
Some of the thoughts above, in a patch: 
[crosspoint-reader-11-18-24 AM.patch](https://github.com/user-attachments/files/25609996/crosspoint-reader-11-18-24.AM.patch)

lib/hal/HalStorage.cpp
ngxson: it's mostly to mirror bug-for-bug the behavior of FsFile, although I'm open for suggestions

lib/hal/HalStorage.h
ngxson: yeah agree, I aimed to mirror the exact behavior of FsFile, but let me check if any places in the code actually copy FsFile

lib/hal/HalStorage.cpp
ngxson: I'm not quite sure about `position` and `available` though, as they can technically change if the same file is read by multiple threads at the same, though agree that it's unlikely to happen in practice. I'm keeping mutex on them anyway.

OK to remove mutex for the rest that you mentioned

lib/hal/HalStorage.h
ngxson: should be ok with your patch: https://github.com/crosspoint-reader/crosspoint-reader/pull/1212/commits/bb7096a92026484b082daf353a57bab5fda6aad0

lib/hal/HalStorage.h
ngxson: ok so after applied your patch, it still compiles, so I assume that we never use the copy constructor in downstream code

znelson (approved):
This is a great one to get fixed, thanks for taking it on!

PR #1213, opened 2026-02-27
feat: update language code validation and abbreviation logic in scripts/gen_i18n.py

victordomingos:
…i18n.py

Refactor language abbreviation handling and validation in scripts/gen_i18n.py in order to be able to support language codes with more than 2 characters, e.g. language variations like PTPT - Portuguese (Portugal) and PTBR - Portuguese (Brasil). Now it will generate correctly the C++ items for those cases.

## Summary

Refactored language abbreviation handling and validation in scripts/gen_i18n.py in order to be able to support language codes with more than 2 characters, e.g. language variations like PTPT - Portuguese (Portugal) and PTBR - Portuguese (Brasil). Now it will generate correctly the C++ items for those cases.
Also some additional validation is performed for duplicate language codes.

---

### AI Usage

Did you use AI tools to help write this code? _**< YES (some help from ChatGPT 5.2) >**_

### DIFF
--- a/scripts/gen_i18n.py
+++ b/scripts/gen_i18n.py
@@ -161,8 +161,17 @@ def load_translations(
         name = data.get("_language_name")
         if not code or not name:
             raise ValueError(f"{fname}: missing _language_code or _language_name")
+        code = code.strip().upper()
+        if not re.match(r"^[A-Za-z_][A-Za-z0-9_]*$", code):
+            raise ValueError(f"{fname}: invalid _language_code '{code}'.")
         language_codes.append(code)
         language_names.append(name)
+
+    # Ensure generated abbreviations are unique
+    abbrevs = [get_lang_abbreviation(c, n) for c, n in zip(language_codes, language_names)]
+    duplicates = sorted({a for a in abbrevs if abbrevs.count(a) > 1})
+    if duplicates:
+        raise ValueError(f"Duplicate language codes: {', '.join(duplicates)}.")
 
     # String keys come from English (order matters)
     english_data = parsed[english_file]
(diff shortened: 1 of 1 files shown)


osteotek:
Forgive me my ignorance, but are Portuguese (Portugal) and Portuguese (Brasil) different enough to warrant different translations? Honest question

victordomingos:
It’s a fair question, and I agree that the decision ultimately depends on priorities such as maintenance cost and contributor availability.

That said, from a linguistic and usability perspective, European Portuguese and Brazilian Portuguese are not interchangeable in practice — especially in technical terminology. While the languages are mutually intelligible, there are consistent differences in vocabulary and usage that directly affect user experience.

For example:
	-	“file” → arquivo (BR) vs ficheiro (PT)
	-	“user” → usuário (BR) vs utilizador (PT)
In Portugal, usuário is rarely used in software contexts and can even be confused with usurário (which has a negative meaning).
	-	“screen” → tela (BR) vs ecrã (PT)

Most major software platforms localize to both variants (pt-BR and pt-PT) precisely because these differences are noticeable to end users and can affect perceived quality and adoption.

If there is no significant technical or maintenance burden, providing both variants side by side would improve inclusivity and usability for Portuguese users in different regions.

Of course, I fully respect whichever direction the project decides to take, and I’m happy to contribute accordingly.

victordomingos:
I would like to get some feedback on this PR, in order to be able to start contributing with translations. I am particularly interested in European Portuguese, but first we need a clarification if the language codes proposed here for variants like European/Brazilian Portuguese are the way to go.

ariel-lindemann:
if we agree fo support this, then i suggest following [BCP 47](https://developer.mozilla.org/en-US/docs/Glossary/BCP_47_language_tag) and documenting that in `i18.md`.

Just so future contributors have a reference and we don't need a new discussion for each proposal.

Linking https://github.com/crosspoint-reader/crosspoint-reader/pull/1506#issuecomment-4130245356

PR #1218, opened 2026-02-27
fix: properly implement requestUpdateAndWait()

ngxson:
## Summary

Properly implement `requestUpdateAndWait()` using freeRTOS direct task notification.

FWIW, I think most of the current use cases of `requestUpdateAndWait()` are redundant, better to be replaced by `requestUpdate(true)`. But just keeping them in case we can find a proper use case for it in the future.

---

### AI Usage

Did you use AI tools to help write this code? **YES**, it's trivial, so I asked an AI to write the code


src/activities/network/WifiSelectionActivity.cpp
```diff
@@ -462,13 +462,6 @@ std::string WifiSelectionActivity::getSignalStrengthIndicator(const int32_t rssi
 }
 
 void WifiSelectionActivity::render(RenderLock&&) {
-  // Don't render if we're in PASSWORD_ENTRY state - we're just transitioning
-  // from the keyboard subactivity back to the main activity
-  if (state == WifiSelectionState::PASSWORD_ENTRY) {
-    requestUpdateAndWait();
-    return;
-  }
```
znelson, line 470: While `requestUpdateAndWait` is bad here because it leads to a deadlock, I think the early return was still valuable. When the keyboard subactivity exits and render fires before `loop` has advanced the state, there's no point drawing the full wifi network list just to immediately overwrite it.

src/activities/ActivityManager.cpp
```diff
@@ -225,6 +234,22 @@ void ActivityManager::requestUpdate(bool immediate) {
...
   }
 }
+void ActivityManager::requestUpdateAndWait() {
+  if (!renderTaskHandle) {
+    return;
+  }
+  // There should never be the case where 2 tasks are waiting for a render at the same time
+  assert(!waitingTaskHandle && "Already waiting for a render to complete");
+  // Render task cannot call requestUpdateAndWait() or it will cause a deadlock
+  assert(xTaskGetCurrentTaskHandle() != renderTaskHandle && "Render task cannot call requestUpdateAndWait()");
```
znelson: I think the @coderabbitai observation here is correct, there's a really small potential for deadlock if two tasks happened to call `requestUpdateAndWait` and interrupt back and forth in the precise sequence to overwrite `waitingTaskHandle`. 

```suggestion
  // Render task cannot call requestUpdateAndWait() or it will cause a deadlock
  static portMUX_TYPE waitingTaskMutex = portMUX_INITIALIZER_UNLOCKED;
  taskENTER_CRITICAL(&waitingTaskMutex);
  bool alreadyWaiting = (waitingTaskHandle != nullptr);
  if (!alreadyWaiting) {
    waitingTaskHandle = xTaskGetCurrentTaskHandle();
  }
  taskEXIT_CRITICAL(&waitingTaskMutex);
  assert(!alreadyWaiting && "Already waiting for a render to complete");
  if (alreadyWaiting) {
    return;
  }
```

Or maybe clearer if you make `std::atomic<TaskHandle_t> waitingTaskHandle{nullptr};` and:
```c++
  TaskHandle_t expected = nullptr;
  TaskHandle_t self = xTaskGetCurrentTaskHandle();
  if (!waitingTaskHandle.compare_exchange_strong(expected, self)) {
    assert(false && "Already waiting for a render to complete");
    return;
  }
```

And then the clear side in `renderTaskLoop`:
```c++
  if (TaskHandle_t waiter = waitingTaskHandle.exchange(nullptr)) {
    xTaskNotify(waiter, 1, eIncrement);
  }
```

znelson (dismissed):

src/activities/ActivityManager.cpp
ngxson: I decided to go with the `taskENTER_CRITICAL` instead of `std::atomic` as it looks cleaner in the particular case, see https://github.com/crosspoint-reader/crosspoint-reader/pull/1218/commits/882e327d8e17f82d3aa92afb03f0009bd6b00841

src/activities/ActivityManager.cpp
```diff
@@ -225,6 +234,36 @@ void ActivityManager::requestUpdate(bool immediate) {
...
+
+  // There should never be the case where 2 tasks are waiting for a render at the same time
+  assert(!alreadyWaiting && "Already waiting for a render to complete");
+
+  // Cannot call while holding RenderLock or it will cause a deadlock
+  assert(!holdingRenderLock && "Cannot call requestUpdateAndWait() while holding RenderLock");
+
+  xTaskNotify(renderTaskHandle, 1, eIncrement);
+  ulTaskNotifyTake(pdTRUE, portMAX_DELAY);
+}
```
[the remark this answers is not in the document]
ngxson: assertions are always enabled, it's fine here. this should be a hard-crash so that it's visible to developers

daveallie (approved):

PR #1219, opened 2026-02-27
feat: Auto Page Turn for Epub Reader

GenesiaW:
## Summary

  - Implements auto page turn feature for epub reader in the reader submenu

  - added auto page turn feature in epub reader in the submenu
  - currently there are 5 settings, `OFF, 1, 3, 6, 12` pages per minute

## Additional Context
  specific areas to focus on).
  - Replacement PR for #723 
  - when auto turn is enabled, space reserved for chapter title will be used to indicate auto page turn being active
  - Back and Confirm button is used to disable it

---

### AI Usage

Did you use AI tools to help write this code? _**Partially (mainly code reviews)**_

### DIFF
--- a/src/activities/reader/EpubReaderActivity.cpp
+++ b/src/activities/reader/EpubReaderActivity.cpp
@@ -132,7 +132,6 @@ void EpubReaderActivity::loop() {
     if (mappedInput.wasReleased(MappedInputManager::Button::Confirm) ||
         mappedInput.wasReleased(MappedInputManager::Button::Back)) {
       automaticPageTurnActive = false;
-      updateLastPageTurnTime = false;
       // updates chapter title space to indicate page turn disabled
       requestUpdate();
       return;
@@ -143,26 +142,14 @@ void EpubReaderActivity::loop() {
       return;
     }
 
-    const unsigned long currentTime = millis();
-
-    // Updates lastPageTurnTime while waiting for render/indexing
-    if (updateLastPageTurnTime) {
-      lastPageTurnTime = currentTime;
+    // Skips page turn if renderingMutex is busy
+    if (RenderLock::peek()) {
+      lastPageTurnTime = millis();
       return;
     }
 
-    if ((currentTime - lastPageTurnTime) >= pageTurnDuration) {
-      lastPageTurnTime = currentTime;
-      updateLastPageTurnTime = true;
-      if (section->currentPage < section->pageCount - 1) {
-        section->currentPage++;
-      } else {
-        RenderLock lock(*this);
-        nextPageNumber = 0;
-        currentSpineIndex++;
-        section.reset();
-      }
-      requestUpdate();
+    if ((millis() - lastPageTurnTime) >= pageTurnDuration) {
+      pageTurn(true);
       return;
     }
   }
@@ -236,8 +223,7 @@ void EpubReaderActivity::loop() {
   const bool skipChapter = SETTINGS.longPressChapterSkip && mappedInput.getHeldTime() > skipChapterMs;
 
   if (skipChapter) {
-    // if chapter was skipped manually, reset time elapsed
-    resetLastPageTurnTime();
+    lastPageTurnTime = millis();
     // We don't want to delete the section mid-render, so grab the semaphore
     {
       RenderLock lock(*this);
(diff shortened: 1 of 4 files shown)


osteotek:
Apologies for recent refactorings that put additional strain on contributors with open PRs. We will try to review this soon

src/activities/reader/EpubReaderActivity.cpp
```diff
@@ -126,6 +128,45 @@ void EpubReaderActivity::loop() {
     return;
   }
 
+  if (automaticPageTurnActive) {
+    if (mappedInput.wasReleased(MappedInputManager::Button::Confirm) ||
+        mappedInput.wasReleased(MappedInputManager::Button::Back)) {
+      automaticPageTurnActive = false;
+      updateLastPageTurnTime = false;
```
ngxson: Just wondering, instead of keeping track of this `updateLastPageTurnTime` which seems to be quite fragile, can we simply consider a page turn (either auto or manual) as just one single action?

Pseudo code:

```cpp
void loop() {
  if (mappedInput.wasReleased(RIGHT)) {
    pageTurn(true);
  }
  if (mappedInput.wasReleased(LEFT)) {
    pageTurn(false);
  }
  if (lastPageTurnTime > 0 && millis() - lastPageTurnTime >= autoTurnDelay) {
    pageTurn(true);
  }
}

void pageTurn(bool next) {
  if (next) {
    // Code to turn to the next page
  } else {
    // Code to turn to the previous page
  }
  lastPageTurnTime = millis();  // always update this
}
```

Also the current logic assumes that a render == a page turn, which is ok for now, but I just doubt if that will no longer true in the future (?)

GenesiaW: Yes, I believe that should be possible, and will probably result in cleaner code.

There was an edge case raised in the previous PR where the indexing of the next chapter was longer than the autoTurnDelay. This resulted in the first page of the next chapter being skipped when auto turn is active. 

Hence, the `updateLastPageTurnTime` flag was added to determine completion of render along with a constant update of `lastPageTurnTime `

ngxson: In such case, I think a logic like: if rendering is happening, then we update lastPageTurnTime, but does not actually turn the page:

```cpp
void loop() {
  // ... existing code
  auto currLastTurnTime = lastPageTurnTime;
  if (lastPageTurnTime > 0 && millis() - lastPageTurnTime >= autoTurnDelay) {
    if (RenderLock::peak()) {
      // If we're still rendering, we can skip this turn and wait for the next loop
      // also update lastPageTurnTime to avoid trying to turn again immediately after rendering finishes
      lastPageTurnTime = millis();
      return;
    }
    pageTurn(true);
  }
}
```

`RenderLock::peak()` does not exist yet, but it's trivial to add: https://forums.freertos.org/t/how-to-peek-a-mutex/5544/3

GenesiaW: Cool! I will make the changes accordingly

src/activities/reader/EpubReaderActivity.cpp
```diff
@@ -126,6 +128,32 @@ void EpubReaderActivity::loop() {
     return;
   }
 
+  if (automaticPageTurnActive) {
+    if (mappedInput.wasReleased(MappedInputManager::Button::Confirm) ||
+        mappedInput.wasReleased(MappedInputManager::Button::Back)) {
+      automaticPageTurnActive = false;
+      // updates chapter title space to indicate page turn disabled
+      requestUpdate();
+      return;
```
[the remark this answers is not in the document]
GenesiaW: This is intentional as layout remains stable when user disables auto page turn, which provides a smoother user experience.

jiddum:
Hello!
I was just wondering about the reasoning behind the options used for pages per minute. Why does it skip 2? 1 goes straight to 3 then straight to 6 then 12.
I had tried crosspet before which had 2 pages per minute auto flip so I was confused at the skipped options and seemingly arbitrary skips between 1, 3, 6, and 12.
Thank you in advance for your reply.

GenesiaW:
> Hello! I was just wondering about the reasoning behind the options used for pages per minute. Why does it skip 2? 1 goes straight to 3 then straight to 6 then 12. I had tried crosspet before which had 2 pages per minute auto flip so I was confused at the skipped options and seemingly arbitrary skips between 1, 3, 6, and 12. Thank you in advance for your reply.

Previously I was thinking in terms of how much time is spent reading on each page which was how I landed with 1,3,6,12. Those options correspond with 60 seconds, 20 seconds, 10 seconds, and 5 seconds on each page

jiddum:
> > Hello! I was just wondering about the reasoning behind the options used for pages per minute. Why does it skip 2? 1 goes straight to 3 then straight to 6 then 12. I had tried crosspet before which had 2 pages per minute auto flip so I was confused at the skipped options and seemingly arbitrary skips between 1, 3, 6, and 12. Thank you in advance for your reply.
> 
> Previously I was thinking in terms of how much time is spent reading on each page which was how I landed with 1,3,6,12. Those options correspond with 60 seconds, 20 seconds, 10 seconds, and 5 seconds on each page

oh ! I see. is there a reason 2 pages per minute was not included? that would still be 30 seconds on each page right?

PR #1222, opened 2026-02-27
docs: ActivityManager migration guide

znelson:
## Summary

Added overview and migration guide for ActivityManager changes in #1016. Thanks for the suggestion, @drbourbon!

---

### AI Usage

Did you use AI tools to help write this code? _**YES, fully written by Claude 4.6**_

### DIFF
--- a/docs/activity-manager.md
+++ b/docs/activity-manager.md
@@ -266,7 +266,7 @@ There are two tasks relevant to the activity system:
 
 Both tasks run at priority 1. Since the ESP32-C3 is single-core, they alternate execution: the main task runs `loop()`, then at the end of the loop iteration, notifies the render task if an update was requested. The render task wakes, acquires the mutex, calls `render()`, releases the mutex, and blocks again.
 
-Avoid creating additional FreeRTOS tasks inside activities. Most work that seems to need a background task can be done directly in `loop()` using non-blocking patterns or `requestUpdateAndWait()`. There are currently only two activities that create their own tasks (for network operations), and these are considered technical debt — arbitrary `xTaskCreate` is error-prone because the task must be carefully tied to the activity's lifecycle and cleaned up in `onExit()`. If a background task is genuinely unavoidable, it should be wrapped in a lifecycle-aware `Worker` helper that ties creation and deletion to `onEnter()` / `onExit()`, rather than using raw `xTaskCreate` / `vTaskDelete`. Such a wrapper does not exist yet — if you find yourself needing one, that's the right time to build it. Background tasks must never call `render()` directly; use `requestUpdate()` or `requestUpdateAndWait()` instead.
+Do not use `xTaskCreate` inside activities. If you have a use case that seems to require a background task, open a discussion to propose a lifecycle-aware `Worker` abstraction first.
 
 ### The Render Mutex and RenderLock
 


znelson:
You can view the [rendered document here](https://github.com/znelson/crosspoint-reader/blob/8644e710d1cd77dfce0ae56772d2016008773a6c/docs/activity-manager.md)

docs/activity-manager.md
```diff
@@ -0,0 +1,472 @@
...
+The `enterNewActivity()` / `exitActivity()` subactivity pattern is replaced by a stack with typed results:
+
+```cpp
+// BEFORE
+void MyActivity::launchWifi() {
+  enterNewActivity(new WifiSelectionActivity(renderer, mappedInput,
+      [this](bool connected) { onWifiDone(connected); }));
+}
+// Child calls: onComplete(true); // triggers callback, which may call exitActivity()
+
```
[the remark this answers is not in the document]
znelson: Neither `ConfirmationActivity` nor `pendingSubactivityExit` exist in this codebase. These appear to be references from CodeRabbit's learning data (a previous PR #909 review), not actual code in the current tree. The `ActivityWithSubactivity` class in our codebase has no `pendingSubactivityExit` flag, and there is no `ConfirmationActivity` class.

Since the code being referenced doesn't exist here, there's nothing to clarify in the doc. Adding a note about a pattern that isn't present would confuse readers rather than help them. No change needed.

docs/activity-manager.md
```diff
@@ -0,0 +1,472 @@
...
+// AFTER
+class SettingsActivity : public Activity {
+public:
+  SettingsActivity(GfxRenderer& r, MappedInputManager& m)
+      : Activity("Settings", r, m) {}
+  // Use finish() to go back, activityManager.goHome() to go home
+};
+```
+
+This removes `std::function` overhead (~2-4KB per unique signature) and eliminates lifetime risks from captured `this` pointers.
```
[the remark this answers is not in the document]
znelson: The `~2-4KB per unique signature` claim comes directly from this project's own engineering guidelines in `CLAUDE.md`:

> `std::function<void()>` adds ~2–4 KB per unique signature and heap-allocates its closure. Avoid both in library code and any path called from the render loop.

The number refers to binary code bloat from template instantiation (each distinct `std::function<Signature>` generates separate code), and it's the project's documented rationale for avoiding `std::function` on this constrained target. Making the statement vaguer would lose the concrete motivation that matters to developers working on a 380KB RAM device.

No change needed — the doc is consistent with the project's own guidelines.

docs/activity-manager.md
```diff
@@ -0,0 +1,472 @@
...
+  state = DONE;
+}
+
+// CORRECT — release lock before blocking
+void MyActivity::doNetworkStuff() {
+  {
+    RenderLock lock;
+    state = LOADING;
+  }
+  requestUpdate();               // render "Loading..." while we work
```
ngxson:
```suggestion
  requestUpdate(true);               // render "Loading..." while we work
```

Here we will also need `immediate` rendering, otherwise the render will only be called after loop finishes (aggregate multiple calls to `requestUpdate()`)

docs/activity-manager.md
```diff
@@ -0,0 +1,472 @@
...
+## Migration Checklist
+
+### 1. Change Base Class
+
+If your activity extended `ActivityWithSubactivity`, change it to extend `Activity`:
+
+```cpp
+// BEFORE
+class MyActivity final : public ActivityWithSubactivity {
+  MyActivity(GfxRenderer& r, MappedInputManager& m, std::function<void()> goBack)
```
ngxson, line 91: note that `GfxRenderer` and `MappedInputManager` are pretty much singletons at this point, I think we can remove them from activity argument in near future

docs/activity-manager.md
```diff
@@ -0,0 +1,472 @@
...
+│ - pending actions    │     │ - RenderLock (mutex)     │
+│ - sleep/power mgmt   │     │ - activity->render()     │
+│ - requestUpdate →────┼─────┼─► xTaskNotify()          │
+│   (end of loop)      │     │                          │
+└──────────────────────┘     └──────────────────────────┘
+```
+
+Both tasks run at priority 1. Since the ESP32-C3 is single-core, they alternate execution: the main task runs `loop()`, then at the end of the loop iteration, notifies the render task if an update was requested. The render task wakes, acquires the mutex, calls `render()`, releases the mutex, and blocks again.
+
+Activities may also create their own background tasks (e.g., for network operations). These tasks can call `requestUpdate()` or `requestUpdateAndWait()` to trigger renders, but must never call `render()` directly.
```
ngxson: Actually there are only 2 places in the code that creates a new task, but I would say it's very error-prone and indeed, they can be converted to simple loop() code.

I think we should consider arbitrary `xTaskCreate` as an anti-pattern at this point, until someone really need it (at such point, I think we should better have a `Worker` wrapper that is tied to the life cycle of the activity)

docs/activity-manager.md
```diff
@@ -0,0 +1,472 @@
...
+
+Avoid creating additional FreeRTOS tasks inside activities. Most work that seems to need a background task can be done directly in `loop()` using non-blocking patterns or `requestUpdateAndWait()`. There are currently only two activities that create their own tasks (for network operations), and these are considered technical debt — arbitrary `xTaskCreate` is error-prone because the task must be carefully tied to the activity's lifecycle and cleaned up in `onExit()`. If a background task is genuinely unavoidable, it should be wrapped in a lifecycle-aware `Worker` helper that ties creation and deletion to `onEnter()` / `onExit()`, rather than using raw `xTaskCreate` / `vTaskDelete`. Such a wrapper does not exist yet — if you find yourself needing one, that's the right time to build it. Background tasks must never call `render()` directly; use `requestUpdate()` or `requestUpdateAndWait()` instead.
```
ngxson: I think just simple mention of this style should be good: about (1) do not use `xTaskCreate` and (2) open a discussion for `Worker` when someone need it.

The `Worker` does not exist yet, but it would be helpful if someone find a legitimate use case for it and make a proposal first.

The 2 places currently using `xTaskCreate` will be removed soon so I think there is no need to mention in the docs

osteotek (approved):

osteotek:
I get benefits of RAII wrapper for locks, but its implicit behavior always seemed less obvious to me. But I'm just old-fashioned I guess

PR #1225, opened 2026-02-28
feat: Add git branch to version information on settings screen

jpirnay:
## Summary

* **What is the goal of this PR?** During my development I am frequently jumping from branch to branch flashing test versions on my device. It becomes sometimes quite difficult to figure out which version of the software I am currently looking at. 

  - Dev builds now display the current git branch in the version string shown on the Settings screen (e.g. 1.1.0-dev+feat-my-feature), making it easier to identify which firmware is running on the device when switching between branches frequently.
  - Release, RC, and slim builds are unaffected — they continue to set their version string statically in platformio.ini.
after

## Additional Context

A new PlatformIO pre-build script (scripts/git_branch.py) runs automatically before every dev build. It reads the base version from the [crosspoint] section of platformio.ini, queries git rev-parse --abbrev-ref HEAD for the current branch, and injects the combined string as the CROSSPOINT_VERSION preprocessor define. In a detached HEAD state it falls back to the short commit SHA. If git is unavailable it warns and falls back to unknown.

The script can also be run directly with python scripts/git_branch.py for validation without triggering a full build.

---

### AI Usage

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -3,7 +3,7 @@ default_envs = default
 extra_configs = platformio.local.ini
 
 [crosspoint]
-version = 1.1.0
+version = 1.1.1
 
 [base]
 platform = https://github.com/pioarduino/platform-espressif32/releases/download/55.03.37/platform-espressif32.zip


znelson (approved):
Nice! This sounds useful because I find myself in the same situation, questioning which version I most recently flashed. What would you think about adding the short commit SHA, too, like "30d8a8d"?

PR #1229, opened 2026-02-28
fix: Hanging indent (negative text-indent) and em-unit sizing

jpirnay:
## Summary

* **What is the goal of this PR?** Fixing two independent CSS rendering bugs combined to make hanging-indent list styles
(e.g. margin-left:3em; text-indent:-1em) render incorrectly:

  1. Negative text-indent was silently ignored

      Three guards in ParsedText.cpp (computeLineBreaks, computeHyphenatedLineBreaks,
      extractLine) conditioned firstLineIndent on blockStyle.textIndent > 0, so any
      negative value collapsed to zero. Additionally, wordXpos was uint16_t, which
      cannot represent negative offsets — a cast of e.g. −18 would wrap to 65518 and
      render the word far off-screen.

   2. extraParagraphSpacing suppressed hanging indents

      Even after removing the > 0 guard, the existing !extraParagraphSpacing condition
      would still suppress all text-indent when that setting is on (its default). Positive
      text-indent is a decorative paragraph indent that the user can reasonably replace with
      vertical spacing — negative text-indent is structural (it positions the list marker)
      and must always apply.

   3. em unit was calibrated against line height, not font size

      emSize was computed as getLineHeight() * lineCompression (the full line advance).
      CSS em units are defined relative to the font-size, which corresponds to the
      ascender height — not the line height. Using line height makes every em-based
      margin/indent ~20–30% wider than a browser would render it, and is especially
      noticeable for CSS that uses font-size: small (which we do not implement).

## Additional Context

Test case
```
.lsl1 { margin-left: 3em; text-indent: -1em; }

<div class="lsl1">• First list item that wraps across lines</div>
<div class="lsl1">• Short item</div>
```
Before: all lines of all items started at 3 em from the left edge (indent ignored).

After: the bullet marker hangs at 2 em; continuation lines align at 3 em.

before
after

---

### AI Usage

### DIFF
--- a/lib/Epub/Epub/Section.cpp
+++ b/lib/Epub/Epub/Section.cpp
@@ -10,7 +10,7 @@
 #include "parsers/ChapterHtmlSlimParser.h"
 
 namespace {
-constexpr uint8_t SECTION_FILE_VERSION = 14;
+constexpr uint8_t SECTION_FILE_VERSION = 16;
 constexpr uint32_t HEADER_SIZE = sizeof(uint8_t) + sizeof(int) + sizeof(float) + sizeof(bool) + sizeof(uint8_t) +
                                  sizeof(uint16_t) + sizeof(uint16_t) + sizeof(uint16_t) + sizeof(bool) + sizeof(bool) +
                                  sizeof(uint32_t);
--- a/lib/Epub/Epub/ParsedText.cpp
+++ b/lib/Epub/Epub/ParsedText.cpp
@@ -144,9 +144,12 @@ std::vector<size_t> ParsedText::computeLineBreaks(const GfxRenderer& renderer, c
     return {};
   }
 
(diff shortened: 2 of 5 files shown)


lib/Epub/Epub/Section.cpp
```diff
@@ -10,7 +10,7 @@
 #include "parsers/ChapterHtmlSlimParser.h"
 
 namespace {
-constexpr uint8_t SECTION_FILE_VERSION = 14;
+constexpr uint8_t SECTION_FILE_VERSION = 16;
```
jdk2pq, line 13: minor: should this be 15 instead of 16?

znelson: Good catch!

jdk2pq: Looks like it got merged as is, so all good. Version numbers are "free", I suppose 😄

PR #1245, opened 2026-02-28
feat: footnote anchor navigation

Uri-Tauber:
## Summary: Enable footnote anchor navigation in EPUB reader

  This PR extracts the core anchor-to-page mapping mechanism from PR #1143 (TOC fragment navigation) to provide immediate footnote navigation support. By merging this focused subset first, users get a complete footnote experience now while simplifying the eventual review and merge of the full #1143 PR.

  ---

## What this extracts from PR #1143

  PR #1143 implements comprehensive TOC fragment navigation for EPUBs with multi-chapter spine files. This PR takes only the anchor resolution infrastructure:

   - Anchor-to-page mapping in section cache: During page layout, ChapterHtmlSlimParser records which page each HTML id attribute lands on, serializing the map into the .bin cache file.
   - Anchor resolution in `EpubReaderActivity`: When navigating to a footnote link with a fragment (e.g., `chapter2.xhtml#note1`), the reader resolves the anchor to a  page number and jumps directly to it.
   - Section file format change: Bumped to version 15, adds anchor map offset in header.

  ---

## Simplified scope vs. PR #1143

  To minimize conflicts and complexity, this PR differs from #1143 in key ways:

* **Anchors tracked**
  * **Origin:** Only TOC anchors (passed via `std::set`)
  * **This branch:** All `id` attributes

* **Page breaks**
  * **Origin**: Forces new page at TOC chapter boundaries
  * **This branch:** None — natural flow

* **TOC integration**
  * **Origin**: `tocBoundaries`, `getTocIndexForPage()`, chapter skip
  * **This branch:** None — just footnote links

* **Bug fix**
  * **This branch:** Fixed anchor page off-by-1/2 bug

  The anchor recording bug (recording page number before `makePages()` flushes previous block) was identified and fixed during this extraction. The fix uses a deferred `pendingAnchorId` pattern that records the anchor after page completion.

  ---

## Positioning for future merge

  Changes are structured to minimize conflicts when #1143 eventually merges:

   - `ChapterHtmlSlimParser.cpp` `startElement()`: Both branches rewrite the same if `(!idAttr.empty())` block. The merged version will combine both approaches (TOC anchors get page breaks + immediate recording; footnote anchors get deferred recording).
   - `EpubReaderActivity.cpp` `render()`: The `pendingAnchor` resolution block is positioned at the exact same insertion point where #1143 places its `pendingTocIndex` block (line 596, right after `nextPageNumber` assignment). During merge, both blocks will sit side-by-side.
   
  ---

## Why merge separately?

   1. Immediate user value: Footnote navigation works now without waiting for the full TOC overhaul
   2. Easier review: ~100 lines vs. 500+ lines in #1143 
   3. Bug fix included: The page recording bug is fixed here and will carry into #1143 
   4. Minimal conflicts: Structured for clean merge — both PRs touch the same files but in complementary ways
---

### AI Usage

Did you use AI tools to help write this code? _**< YES >**_ Done by Claude Opus 4.6


Uri-Tauber:
@znelson I’d really appreciate your thoughts on this.

BTW, I’m wondering whether using `std::map` might be overkill for storing anchors. I wouldn’t expect more than 20–30 anchors per section, so maybe something like `std::vector<std::pair<std::string, uint16_t>>` would be sufficient and simpler.

What do you think?

znelson:
Thanks @Uri-Tauber! I won't get through this tonight but I will do my best to do so tomorrow. 🙂

lib/Epub/Epub/parsers/ChapterHtmlSlimParser.h
```diff
@@ -117,4 +124,5 @@ class ChapterHtmlSlimParser {
   ~ChapterHtmlSlimParser() = default;
   bool parseAndBuildPages();
   void addLineToPage(std::shared_ptr<TextBlock> line);
+  const std::map<std::string, uint16_t>& getAnchorPageMap() const { return anchorPageMap; }
```
znelson: I do think we'd be better off with `std::vector<std::pair<std::string, uint16_t>>` versus `std::map`. My intuition is that the dynamic allocation for each node in the map will be more detrimental than the linear search over a vector, especially for the relatively low number of anchors in a typical chapter.

During this ChapterHtmlSlimParser pass I think we use the anchor structure to append only, gathering [anchor, page] pairs to serialize, so a vector of pairs would be perfectly sufficient.


```suggestion
  const std::vector<std::pair<std::string, uint16_t>>& getAnchors() const { return anchorData; }
```

lib/Epub/Epub/parsers/ChapterHtmlSlimParser.cpp
```diff
@@ -133,11 +133,21 @@ void ChapterHtmlSlimParser::startNewTextBlock(const BlockStyle& blockStyle) {
       // This handles cases like <div style="margin-bottom:2em"><h1>text</h1></div> where the
       // div's margin should be preserved, even though it has no direct text content.
       currentTextBlock->setBlockStyle(currentTextBlock->getBlockStyle().getCombinedBlockStyle(blockStyle));
+
+      if (!pendingAnchorId.empty()) {
+        anchorPageMap[pendingAnchorId] = static_cast<uint16_t>(completedPageCount);
+        pendingAnchorId.clear();
```
znelson: When you switch to a vector, you can move the `pendingAnchorId` string into the pair:

```suggestion
        anchorData.push_back({std::move(pendingAnchorId), static_cast<uint16_t>(completedPageCount)});
        pendingAnchorId.clear();
```

lib/Epub/Epub/parsers/ChapterHtmlSlimParser.cpp
```diff
@@ -151,19 +161,28 @@ void XMLCALL ChapterHtmlSlimParser::startElement(void* userData, const XML_Char*
...
   std::string styleAttr;
+  std::string idAttr;
   if (atts != nullptr) {
     for (int i = 0; atts[i]; i += 2) {
       if (strcmp(atts[i], "class") == 0) {
         classAttr = atts[i + 1];
       } else if (strcmp(atts[i], "style") == 0) {
         styleAttr = atts[i + 1];
+      } else if (strcmp(atts[i], "id") == 0) {
+        idAttr = atts[i + 1];
```
znelson: Do you need `idAttr`, or can you just assign directly to `pendingAnchorId`?

```suggestion
        self->pendingAnchorId = atts[i + 1];
```

(Otherwise, you could `std::move` or `std::swap` below to avoid copying from `idAttr` to `pendingAnchorId`.)

lib/Epub/Epub/Section.cpp
```diff
@@ -262,3 +273,33 @@ std::unique_ptr<Page> Section::loadPageFromSectionFile() {
   file.close();
   return page;
 }
+
+std::map<std::string, uint16_t> Section::readAnchorMap(const std::string& sectionPath) {
```
znelson: I think I'd probably keep the anchor data structure details private to `Section` and change this API to something like:

```suggestion
std::optional<uint16_t> Section::getPageForAnchor(const std::string& sectionPath, const std::string& anchor) {
```

lib/Epub/Epub/Section.cpp
```diff
@@ -262,3 +273,33 @@ std::unique_ptr<Page> Section::loadPageFromSectionFile() {
...
+  f.seek(anchorMapOffset);
+  uint16_t count;
+  serialization::readPod(f, count);
+  std::map<std::string, uint16_t> result;
+  for (uint16_t i = 0; i < count; i++) {
+    std::string key;
+    uint16_t page;
+    serialization::readString(f, key);
+    serialization::readPod(f, page);
+    result.emplace(std::move(key), page);
```
znelson: If this API changed to `getPageForAnchor`, this loop could return early as soon as it finds a matching key.

src/activities/reader/EpubReaderActivity.cpp
```diff
@@ -593,6 +593,19 @@ void EpubReaderActivity::render(RenderLock&& lock) {
       section->currentPage = nextPageNumber;
     }
 
+    if (!pendingAnchor.empty()) {
+      const auto anchorMap =
+          Section::readAnchorMap(epub->getCachePath() + "/sections/" + std::to_string(currentSpineIndex) + ".bin");
```
znelson: This "/sections/" path logic currently exists only in the `Section` constructor, but now leaks into EpubReaderActivity. Maybe we should keep Section as the authority on its cache file structure by moving this path logic into e.g. `Section::getPageForAnchor`?

And actually we have the `section` in this context. We could switch to a class method (instead of static), and the section already has its `filePath` member populated, so we wouldn't need to duplicate any path logic.

Uri-Tauber:
Thanks @znelson for the feedback! I’ve incorporated all of your suggestions.

I haven’t had a chance to test it on the device yet since I left it at my apartment and won’t be back for three days, but it should work without any issues.

lib/Epub/Epub/Section.cpp
```diff
@@ -234,10 +235,20 @@ bool Section::createSectionFile(const int fontId, const float lineCompression, c
...
-  file.seek(HEADER_SIZE - sizeof(uint32_t) - sizeof(pageCount));
+  // Write anchor-to-page map for fragment navigation (e.g. footnote targets)
+  const uint32_t anchorMapOffset = file.position();
+  const auto& anchors = visitor.getAnchors();
+  serialization::writePod(file, static_cast<uint16_t>(anchors.size()));
+  for (const auto& [anchor, page] : anchors) {
+    serialization::writeString(file, anchor);
+    serialization::writePod(file, page);
+  }
+
```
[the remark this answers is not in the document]
Uri-Tauber: Impossible. even big chapter won't have more than 20 notes, and maybe a few sub-chapters. In fact, we can reduce it to uint8_t and it still will work fine for 99% of  epubs.

lib/Epub/Epub/Section.cpp
```diff
@@ -262,3 +273,35 @@ std::unique_ptr<Page> Section::loadPageFromSectionFile() {
...
+    serialization::readPod(f, page);
+    if (key == anchor) {
+      f.close();
+      return page;
+    }
+  }
+
+  f.close();
+  return std::nullopt;
+}
```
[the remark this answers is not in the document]
Uri-Tauber: What I added: A single anchorMapOffset >= fileSize bounds check (line 286). This catches SD card corruption where the
  offset points past EOF.

  What I did NOT add (and why):

   - readString length validation / malloc cap: serialization::readString blindly does s.resize(len) from disk — but this is
   the same pattern used by ~30 call sites across the codebase (BookMetadataCache, TextBlock, KOReaderCredentialStore, 
  etc.). Hardening readString itself would be the right fix, but that's a cross-cutting concern outside this PR's scope.
   - Per-entry position checks / string length cap: The section cache is written by our own code moments before, versioned 
  (v15), and invalidated on any parameter change. Defending against every possible corruption byte-by-byte adds complexity 
  disproportionate to the risk for a local cache file. The existing SECTION_FILE_VERSION check already rejects 
  stale/mismatched files.

lib/Epub/Epub/parsers/ChapterHtmlSlimParser.cpp
```diff
@@ -133,11 +133,21 @@ void ChapterHtmlSlimParser::startNewTextBlock(const BlockStyle& blockStyle) {
...
       return;
     }
 
     makePages();
   }
+  // Record deferred anchor after previous block is flushed
+  if (!pendingAnchorId.empty()) {
+    anchorData.push_back({std::move(pendingAnchorId), static_cast<uint16_t>(completedPageCount)});
+    pendingAnchorId.clear();
+  }
```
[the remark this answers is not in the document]
Uri-Tauber: This is not a real problem. Here's why:

  The overwrite scenario requires two consecutive id attributes on elements that don't trigger `startNewTextBlock()`. Elements that call `startNewTextBlock` (and flush `pendingAnchorId`) include all block tags: `<p>`, `<div>`, `<blockquote>`, `<li>`, `<br>`, `<h1>`–`<h6>`, `<table>`/`<td>`, and `<aside>`.

  Footnote targets — the anchors we actually navigate to — are virtually always block elements:
```
   <aside id="note1" epub:type="footnote"><p>...</p></aside>
   <p id="note1">Footnote text</p>
   <div id="fn1">...</div>
```
  The only way to lose an ID is two inline elements with IDs in the same block without a block break, e.g. `<span id="a">`...`<span id="b">`. In that case:

   1. Both are on the same page (no page break between them)
   2. Both would record the same completedPageCount value
   3. In practice, footnote back-references (`<a id="ref1">`) are the only inline IDs, and we never navigate to those

  Adding a vector introduces heap allocation overhead on every block boundary for a scenario that doesn't occur in real EPUB footnote markup. No change needed.

lib/Epub/Epub/Section.cpp
```diff
@@ -13,7 +13,7 @@ namespace {
 constexpr uint8_t SECTION_FILE_VERSION = 16;
```
znelson: aff93f1 took SECTION_FILE_VERSION 16, so this needs to be bumped up again

znelson (commented):
This looks good to me! But the SECTION_FILE_VERSION change washed out in the last merge and needs to be updated quick

lib/Epub/Epub/Section.cpp
Uri-Tauber: done

znelson (dismissed):

Uri-Tauber:
Solved merge conflicts.
I will appreciate if someone will test this on the device, since I don't have mine right now.

lib/Epub/Epub/Section.cpp
```diff
@@ -13,7 +13,7 @@ namespace {
 constexpr uint8_t SECTION_FILE_VERSION = 17;
```
znelson: The SECTION_FILE_VERSION bump got lost in the last merge

Uri-Tauber: Nice find. fixed.

znelson (approved):

Uri-Tauber:
@osteotek Can you review this PR?

osteotek (approved):

PR #1253, opened 2026-03-01
fix: Hide unusable button hints when viewing empty directory

Levrk:
## Summary

Increase accuracy of button hints and text description in the file browser when viewing empty directory.
 
Adjusted button label hint rendering logic in file browser to hide the "Open", "Up", and "Down" hints when the they are not available due to an empty directory. 

I also changed the NO_BOOKS_FOUND string to NO_FILES_FOUND and updated translations. File browser shows more than just books so seeing "No Books Found" really doesn't make sense. 

## Additional Context

Very Simple change, here is what that looks like on my device.

Untitled (7)

---

### AI Usage

### DIFF
--- a/src/activities/home/MyLibraryActivity.cpp
+++ b/src/activities/home/MyLibraryActivity.cpp
@@ -266,8 +266,9 @@ void MyLibraryActivity::render(RenderLock&&) {
   }
 
   // Help text
-  const auto labels = mappedInput.mapLabels(basepath == "/" ? tr(STR_HOME) : tr(STR_BACK),
-                                            files.empty() ? "" : tr(STR_OPEN), tr(STR_DIR_UP), tr(STR_DIR_DOWN));
+  const auto labels =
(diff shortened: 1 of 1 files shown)


ariel-lindemann:
@Levrk  thank you for the contribution,

In the future would you please tag the translators, so we can give our input on the translations?

Or you can also leave them unchanged and we will see they need translating. 

But the way it's done here there is the danger that we miss it completely and some translations might be wrong. At least the Romanian one I would have done differently. And also the German and French are techinically correct but not in line with the overall style of the translations. So you're introducing changes which are out of scope and the ones responsible aren't notified.

I'm sure you meant well and thank you for your work

Levrk:
@ariel-lindemann Absolutely, I will be more mindful of this moving forward. Is there a standard method to notify translators, or would I tag folks individually? Perhaps we should have an ongoing discussion post where PR's with translation updates can be linked for visibility.

ariel-lindemann:
Thanks @Levrk. There is no established method yet, but we are discussing it in #1203 if you want to follow or give your perspective. I think there's multiple viable solutions but it's important we agree on something

PR #1254, opened 2026-03-01
fix: improve and add Spanish translations

DaniPhii:
## Summary

    - Improve and add the latest missing Spanish translations

    - Add missing spaces and remove extra unneeded ones (spaces at the end of certain strings and others, i.e. the one introduced in the string `Smart Device`; actually, `SmartDevice` is the correct Calibre plugin name)
    - Normalise the use of caps in certain strings
    - Adapting the translation to the one found in related third-party software (i.e. Spanish translation for the word `plugin` in Calibre is `complemento`)
    - Shortening some translations to make them smaller and fit better in screen
    - Rewording ambiguous translations (i.e. `Volver a inicio` could mean to go back to Home, but also to go back to the first page of the current book, so I changed it for a more specific action, `Volver al menú Inicio`)

## Additional Context

* **Missing spaces caused a lack of clarity** 

    - My main motivation for this PR was the following:
        <details>

        <summary>Screenshots:</summary>
        
        In English:
        English.bmp
        
        In Spanish:
        Spanish.bmp

        </details>

---

### AI Usage


DaniPhii:
Hi @yeyeto2788, @Skrzakk, @pablohc, feel free to check my adjustments.

Maybe some of my changes are not needed; I initially wanted to fix the issue I showed in the screenshots, but then I started making more changes that to me felt like improvements, as the use of caps is more consistent throughout the whole UI when set in Spanish. Also, some strings caused several log errors, for example after installing an update, as shown [here](https://github.com/crosspoint-reader/crosspoint-reader/issues/943#issuecomment-3948095912).

Looking forward to your feedback. 👍🏻

znelson (approved):

yeyeto2788:
Hey @DaniPhii! Thanks for making the PR.

I do have mix feelings as I do see OK some of them, let me explain my thinking (doesn't mean it has to be done as I say)

1. `"Función especial botón Power"` for me would be `"Función especial botón encendido"` which is a mix between what is was and your change.

2. These phrases sound better than you proposal (Perosnal feeling):
    - `"URL del servidor de sinc."` 
    - `"cuando se vuelvan a abrir."`

3. Some of your proposed solution sound better (also personal taste):
    - `"Mantenga pulsado botón de inicio para reiniciar"`
    - `"1 min."` better for the small space in the UI
 
4. `STR_FOOTNOTES: "Notas al pie"` shouldn't it be `STR_FOOTNOTES: "Pie de página"`

Also we need to take into account that we might be having different opinions on what sound better based on our Spanish dialect. I guess as a community we should agree at least on what would be the best to use and keep.

Let's open a discussion and talk about it, virtual coffees are also welcomed.

Cheers.

DaniPhii:
Hi @yeyeto2788, thanks for your reply!

1. I replaced `Función especial` with `Toque simple` as it's the shortest string I could come up with, but at the same time tried to improve the translation; when I read the original translation for the very first time, I understood I was meant to configure the secondary/alternate/special function for the Power button, but I couldn't be sure how it would be triggered once set, as it could be long pressing it or just a single click. But `Pulsación simple botón encendido` would've been too long, so I went with `Toque` instead of `Pulsación` (but maybe it could be even shorter if we say `Clic` and keep it even closer to the original string in English).

2. I'm quite neutral about those two, so I reverted my changes there.

3. While reviewing all what you mentioned, I thought it could even be clearer, so I slightly reworded it.

4. I totally agree with you, I think I wanted to say `Notas al pie de página` but trimmed it incorrectly… 😅

I hope I covered everything, thanks again for your help!

lukestein (approved):

PR #1256, opened 2026-03-01
feat: Rework "Cover + Custom" sleep screens to show covers only when currently reading

iandchasse:
## Summary

The cover/custom setting was a bit misleading to me. Usually after you start reading a book you never see your customs again unless a cover fails to render for some reason. With this, you can easily show your custom images by just sleeping from the menus instead of the reader

## Additional Context

  specific areas to focus on).

---

### AI Usage


znelson (approved):

lukestein (approved):
No code review from me, but tested on device and works well, as expected

PR #1267, opened 2026-03-01
feat: replace UI font `Ubuntu` with `NotoSans`

ariel-lindemann:
## Summary

Replaces the Ubuntu UI font with NotoSans. Credit to @Uri-Tauber for the idea (see https://github.com/crosspoint-reader/crosspoint-reader/pull/1255#issuecomment-3979379429)

+ A new size for NotoSans
+ Removed Ubuntu
+ amdended scripts and docs

## Additional Context

  specific areas to focus on).

The Idea originated under PR #1255 

## What it would look like

image

---

### AI Usage


lib/EpdFont/builtinFonts/all.h
```diff
@@ -17,6 +17,10 @@
 #include <builtinFonts/bookerly_18_italic.h>
 #include <builtinFonts/bookerly_18_regular.h>
 #include <builtinFonts/notosans_8_regular.h>
+#include <builtinFonts/notosans_10_bold.h>
+#include <builtinFonts/notosans_10_bolditalic.h>
+#include <builtinFonts/notosans_10_italic.h>
+#include <builtinFonts/notosans_10_regular.h>
```
znelson, line 21: The system UI only ever uses regular and bold, so I imagine the italic and bold italic are just consuming flash space right now.

znelson (commented):
I certainly had this thought when I was proposing Inter and IBM Plex, but I came to the conclusion that a distinct font for the system UI gives some separation between the reader interface and the system interface. 

If others agree that this looks good (no need for distinction between system UI and reader), and does a good job of covering the Thai, Vietnamese, and other localized ranges, I'm fine with going this direction.

Uri-Tauber:
Master:
`Flash: [========= ]  91.5% (used 5996835 bytes from 6553600 bytes)`

Branch:
`Flash: [========= ]  89.9% (used 5892291 bytes from 6553600 bytes)`

Nice flash saving.

lib/EpdFont/scripts/convert-builtin-fonts.sh
```diff
@@ -44,8 +44,8 @@ UI_FONT_STYLES=("Regular" "Bold")
 
 for size in ${UI_FONT_SIZES[@]}; do
   for style in ${UI_FONT_STYLES[@]}; do
-    font_name="ubuntu_${size}_$(echo $style | tr '[:upper:]' '[:lower:]')"
-    font_path="../builtinFonts/source/Ubuntu/Ubuntu-${style}.ttf"
+    font_name="notosans_${size}_$(echo $style | tr '[:upper:]' '[:lower:]')"
+    font_path="../builtinFonts/source/NotoSans/NotoSans-${style}.ttf"
```
znelson, line 48: This loop is currently stomping the size 12 2-bit Noto Sans reader font with the 1-bit variation rendered for the UI.

lib/EpdFont/builtinFonts/notosans_12_regular.h
```diff
@@ -2,2891 +2,3058 @@
  * generated by fontconvert.py
  * name: notosans_12_regular
  * size: 12
- * mode: 2-bit  compressed: true
- * Command used: fontconvert.py notosans_12_regular 12 ../builtinFonts/source/NotoSans/NotoSans-Regular.ttf --2bit --compress
+ * mode: 1-bit
```
znelson, line 5: You can see here the conversion from 2-bit to 1-bit for Noto Sans size 12

lib/EpdFont/builtinFonts/all.h
ariel-lindemann: good catch! fixed it now

lib/EpdFont/scripts/convert-builtin-fonts.sh
ariel-lindemann: ok I hadnt thought about that. But that would mean that we generate the 1 bit versions on top and we can't reuse size 12 NotoSans, right?

Or alternatively we would use 2 bit for the UI as well?

znelson: Yeah, I'm not sure what the original reasoning was for using 1-bit fonts for the system UI. 

We could do something like this, making the 1-bit UI version of Noto output to a different file: 
[crosspoint-reader-8-23-02 PM.patch](https://github.com/user-attachments/files/25671444/crosspoint-reader-8-23-02.PM.patch)

But then we lose the flash space benefit of reusing Noto. 

One other difference is that the 1-bit UI fonts aren't compressed. Compression was added by @adriancaruana in #831.

adriancaruana: > Yeah, I'm not sure what the original reasoning was for using 1-bit fonts for the system UI.

I think the reason is because the UI doesn't have the two-pass AA rendering that the epub activity does. Not 100% sure about this.

> One other difference is that the 1-bit UI fonts aren't compressed.

Ensure you benchmark the UI rendering time before/after this change to ensure that UI rendering performance doesn't degrade massively due to decompression. Also benchmark free heap at various points in the UI to ensure the overhead from the decompressor doesn't make the chip go OOM.

znelson:
Closing for now. The overlap in 1-bit and 2-bit fonts makes this less of a clear win than it might otherwise be. Please join the font discussion in #1693 if you're interested.

PR #1273, opened 2026-03-02
feat: Page overlay sleep screen mode

jpirnay:
## Summary

* **What is the goal of this PR?** Adds a new "Page overlay" option to the sleep screen setting that composites an image on top of the current reader page when the device goes to sleep, instead of replacing the screen entirely.

## Additional Context

New sleep screen mode (OVERLAY = 6 in SLEEP_SCREEN_MODE):

* The reader page remains visible beneath the overlay image — no "Entering sleep" popup is shown so the frame buffer is preserved
* Overlay image is selected using the same logic as the Custom mode: random pick from /sleep/ directory, or /sleep.bmp / /sleep.png fallback
* BMP overlays: white pixels are transparent (existing drawBitmap behaviour), black pixels are composited on top
* PNG overlays (new): transparent pixels (alpha channel or tRNS colour key) show through to the reader page; opaque pixels are drawn in their grayscale brightness
* Supports all PNG pixel types: TRUECOLOR_ALPHA, GRAY_ALPHA, INDEXED+tRNS, TRUECOLOR+tRNS colour key, GRAYSCALE+tRNS
* PNGdec (~42 KB) is heap-allocated with a 60 KB free-heap guard
* tRNS colour-key transparency is initialised lazily on the first draw callback, because PNGdec parses the tRNS chunk during decode(), not open()
* Re-render from non-reader screens: when going to sleep from the home screen or settings (not from inside a reader), the last reading position is re-rendered to the frame buffer using new static helpers:

| Reader Screen | Mask image | Result |
| --- | --- | --- |
|  | cat |   |
|  | bookcover |  |
|  |snoopy | |

---

### AI Usage


pablohc:
I was trying to achieve the same thing but failed miserably, so thanks for this PR!!!

th0m4sek:
Polish string:
STR_PAGE_OVERLAY: "Nakładka"

trilwu (changes requested):
dismissed

FireMotion:
Suggestion, as a nice-to-have:
I think it would be nice to show "**Sleeping**", to replace the bottom left battery indicators in the status bar.
* This makes it easier to see that the device is sleeping
* You won't be surprised of the battery level change when turning on the device again

efenner:
Why did this get closed?

PR #1286, opened 2026-03-02
fix: use same file name as KOReader for OPDS downloads

spfenwick:
This allows KOSync to work for files downloaded from OPDS when set to use filename matching

## Summary

* **What is the goal of this PR?** Brings the names of files downloaded via OPDS into line with the KOReader OPDS plugin so the files sync when KOSync is set to filename matching
* **What changes are included?** Change the filename generated by the OPDS browser from "\<title\> - \<author\>" to "\<author\> - \<title\>" so it matches KOReader.

## Additional Context

  specific areas to focus on).

Both KOReader and crosspoint sanitise the filename by removing special characters. This step seems to already use the same rules.

---

### AI Usage


znelson (dismissed):

osteotek:
@spfenwick Would it be possible to fix merge conflicts?

spfenwick:
> @spfenwick Would it be possible to fix merge conflicts?

Done. The code has changed quite a bit and it looks like I inadvertently closed the PR in the process of re-syncing but it should be good now.

Uri-Tauber (approved):
Nice

znelson (approved):

PR #1293, opened 2026-03-02
perf: Eliminate per-pixel overheads in image rendering

martinbrook:
## Summary

Replace per-pixel getRenderMode() + rotateCoordinates() + bounds checks with a DirectPixelWriter struct that pre-computes orientation and render mode state once per row. Use bitwise ops instead of division/modulo for cache pixel packing. Skip PNG cache allocation when buffer exceeds 48KB (framebuffer size) since PNG decode is fast enough that caching provides minimal benefit, and the large buffer competes with the 44KB PNG decoder for heap.

## Additional Context
Measured improvements on ESP32-C3 @ 160MHz:
- JPEG decode: 5-7% faster (1:1 scale)
- PNG decode: 15-20% faster (1:1 scale)
- Cache renders: 3-6% faster across both formats
- Eliminates "Failed to allocate cache buffer" errors for large PNGs

---

### AI Usage

Did you use AI tools to help write this code? _**<  PARTIALLY >**_

### DIFF
--- a/lib/Epub/Epub/blocks/ImageBlock.cpp
+++ b/lib/Epub/Epub/blocks/ImageBlock.cpp
@@ -4,7 +4,7 @@
 #include <Logging.h>
 #include <Serialization.h>
 
-#include "../converters/DitherUtils.h"
+#include "../converters/DirectPixelWriter.h"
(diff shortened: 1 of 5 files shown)


martinbrook:
Ready for review

Issue #1183, opened 2026-02-25
File Manager - Cannot delete empty folder if contains dotfiles.


razzoTech:
### Affected Version

1.1.1

### Describe the Bug

File manager shows no file inside a folder (0 B, 0 files) that contained files in the past.

<img width="860" height="504" alt="Image" src="https://github.com/user-attachments/assets/0258e9fb-7846-408b-92ca-868b2a837c10" />

but when trying to delete the folder itself, shows the error: "Failed to delete: Folder is not empty. Delete contents first."

<img width="808" height="400" alt="Image" src="https://github.com/user-attachments/assets/9f1b57c7-ab35-4511-ac53-38eb43c986d6" />

There's some hidden file (dot files)  that locks this operation. If i create an empty folder, with or without moving files on it, from File Manager, the deletion works perfectly. Differently if it was created directly for file explorer / finder on the SD Card.

### Steps to Reproduce

1. Create a folder with a file explorer directly on sd card. Place some file.
2. open xteink 4 file manager from browser.
3. move all files from folder to another.
4. try to delete the folder.

### Expected Behavior

File manager must ignore dotfiles, or it have to show them in order to allow user to delete them.

### Relevant Log Output/Screenshots

```shell

```

razzoTech:
moving with terminal on SD Card i can see mac temp files.

```
$ls -a Volumes/NO NAME/Books 
  .
  ._somebook1.epub
  ._somebook2.epub
  ..
```

osteotek:
We can't just ignore hidden files and delete them silently. We probably need an option to show hidden files

Uri-Tauber:
Closing this issue. Solved in #1288.

PR #1288, opened 2026-03-02
feat: Show hidden directories in browser

jpirnay:
## Summary

* **What is the goal of this PR?** Add setting to display hidden files / directories in filebrowser / web file browser
* **What changes are included?**

## Additional Context

-

---

### AI Usage

While CrossPoint doesn't have restrictions on AI tools in contributing, please be transparent about their usage as it 
helps set the right context for reviewers.

Did you use AI tools to help write this code? _**< NO >**_

PR #1296, opened 2026-03-03
feat: Allow Web Server To Handle Hidden Files

jb567:
## Summary

* Allow hidden files to be seen and edited via the web server
* This was requested by #1183 

## Additional Context

* This does expose protected crosspoint files via the API now, but just flags them as to not be displayed
  * I'm totally open to changing this, it was trying to make less of a change to the end points

---

### AI Usage


jb567:
Seems #1288 was accepted instead.

PR #1298, opened 2026-03-03
fix: use sleep routine from the original firmware

### LINKED ISSUE #1263: Hang on sleep with battery drain
### Affected Version

1.1.1

### Describe the Bug

So,

since, i think 1.0, i've now repeatedly noticed that my X4 sometimes, after going to sleep wont wake back up when pressing the power button.
I then press reset and hold power to turn it back on.

The battery then has noticeably drained, depending on how long since I put the device to sleep.

I'm relatively sure that I've never seen the issue pre-1.0.

I'm assuming that something goes wrong during going to sleep which leads the device to not actually enter sleep mode and then hanging completely. That's probably why I need to do a hard reset.

Anyone else noticed this issue?

### Steps to Reproduce

Can't really reproduce, but

1. Read a book
2. Go to sleep
3. Sometimes it won't come back by pressing power, battery drained depending on how long since sleep

### Expected Behavior

Does not hang and drain battery

```shell

```

ngxson:
## Summary

Fixes #1263

I spent half of my day(-off) reverse engineering the stock english firmware V3.1.1, it's more or less like solving a sudoku with some known pieces (like debug strings, known static addresses, known compiled function, etc) and then the task is to guess the rest.

Long story short, this is the sleep routine that they use:

image

From the code above:
- They pull down GPIO13 (value = 0xd) before sleep
- They verify that power button is released by doing a delay loop of 50ms, similar to what we're doing
- `esp_sleep_config_gpio_isolate` is called but I'm not 100% sure why
- Pull up power button, note that it's likely redundant because power button should already pulled up by `InputManager`
- `param1` and `param2` means enabling front/side buttons for wake up, but it doesn't used in the code in reality. But I think it's physically impossible, see the explanation below
- `param3` means "wake up from power button"
- `esp_sleep_start` is used; there is a logic to handle if it fails to sleep, then retry recursively (no idea why!)

My observation is that they use GPIO13 so that it will be on HIGH state when the chip is powered on, without any user space code to keep it on that state. And once going to deep sleep, it goes into FLOATING by default. That may explain why it need to be in LOW state before going to sleep. (Nice trick btw)

Looking again at the circuit diagram provided [here](https://github.com/sunwoods/Xteink-X4/blob/main/readme-img/sch.jpg) (note: it's not official):

image

It kinda make sense as the GPIO13 and VBUS (USB VCC) have the same role, they are part of a simple "battery protection" cirtuit

Now, we may wonder, how the device wake up when there is no battery at all?

image

It seems like power button is not just a simple switch between GPIO3 and ground, but it also linked the POWER_CTRL, which leads to nowhere on the diagram, but I suppose it connects the battery back for a short amount of time, just enough for the MCU to wake up, and GPIO13 goes HIGH again. It may also explain why power button becomes non-responsive for ~1 second after power on, as it's being pulled up by the current from battery (remind: high = not pressed, low = pressed)

To test the theory above, I simply **comment out** the `esp_deep_sleep_enable_gpio_wakeup`:
- On battery, power button works as nothing happen
- On USB, it doesn't wake up, I need to press RST

---

Important things about my analysis:
1. I had to name every function on the code above **manually**, but I'm 99% confident about it. The only function that I'm not sure is `esp_wifi_bt_power_domain_off` ; Edit: it was indeed mislabeled, see https://github.com/crosspoint-reader/crosspoint-reader/pull/1298#discussion_r2879670852
2. Some logic inside the stock firmware looks very strange, there is almost no mention to "arduino" in the hardware, suggesting that they may just call esp-idf functions directly, bypassing the arduino abstraction.

---

### AI Usage

### DIFF
--- a/lib/hal/HalPowerManager.cpp
+++ b/lib/hal/HalPowerManager.cpp
@@ -53,20 +53,25 @@ void HalPowerManager::setPowerSaving(bool enabled) {
 }
 
 void HalPowerManager::startDeepSleep(HalGPIO& gpio) const {
-  // Pre-sleep routines from the original firmware
-  constexpr gpio_num_t GPIO_SPIWP = GPIO_NUM_13;
-  gpio_set_direction(GPIO_SPIWP, GPIO_MODE_OUTPUT);
-  gpio_set_level(GPIO_SPIWP, 0);
   // Ensure that the power button has been released to avoid immediately turning back on if you're holding it
   while (gpio.isPressed(HalGPIO::BTN_POWER)) {
     delay(50);
     gpio.update();
   }
+  // Pre-sleep routines from the original firmware
+  // GPIO13 is connected to battery latch MOSFET, we need to make sure it's low during sleep
+  // Note that this means the MCU will be completely powered off during sleep, including RTC
+  constexpr gpio_num_t GPIO_SPIWP = GPIO_NUM_13;
+  gpio_set_direction(GPIO_SPIWP, GPIO_MODE_OUTPUT);
+  gpio_set_level(GPIO_SPIWP, 0);
   esp_sleep_config_gpio_isolate();
   gpio_deep_sleep_hold_en();
   gpio_hold_en(GPIO_SPIWP);
   pinMode(InputManager::POWER_BUTTON_PIN, INPUT_PULLUP);
   // Arm the wakeup trigger *after* the button is released
+  // Note: this is only useful for waking up on USB power. On battery, the MCU will be completely powered off, so the
+  // power button is hard-wired to briefly provide power to the MCU, waking it up regardless of the wakeup source
+  // configuration
   esp_deep_sleep_enable_gpio_wakeup(1ULL << InputManager::POWER_BUTTON_PIN, ESP_GPIO_WAKEUP_GPIO_LOW);
   // Enter Deep Sleep
   esp_deep_sleep_start();


ngxson:
@jpirnay I cannot confirm if the stock firmware hold the CS pin to low or not, but I think it's quite unlikely. However, I appreciate if you can verify my theory here when you get the board. I ordered one but it won't arrive in 1-2 weeks.

lib/hal/HalPowerManager.cpp
```diff
@@ -53,11 +53,17 @@ void HalPowerManager::setPowerSaving(bool enabled) {
 }
 
 void HalPowerManager::startDeepSleep(HalGPIO& gpio) const {
+  // Pre-sleep routines from the original firmware
+  constexpr gpio_num_t GPIO_SPIWP = GPIO_NUM_13;
+  gpio_set_level(GPIO_SPIWP, 0);
```
znelson: Do you need to set direction first?


```suggestion
  gpio_set_direction(GPIO_SPIWP, GPIO_MODE_OUTPUT);
  gpio_set_level(GPIO_SPIWP, 0);
```

ngxson:
> what would be fixed by this change? #1263?

Yes, quite likely. However, I would appreciate if someone with better electronic knowledge to confirm my finding about the battery protection circuit.

lib/hal/HalPowerManager.cpp
ngxson: I was also wondering why they don't call `gpio_set_direction`, I suppose it's already output by default. But adding an explicit `gpio_set_direction` won't hurt

iandchasse:
EE here, sound analysis I think. Bit of a weird design on their part, I think the effort done to to design a latch/un-latching mechanism for the ESP is probably more effort than what's needed to just ensure low current draw in deep sleep WITHOUT completely severing power.

lib/hal/HalPowerManager.cpp
```diff
@@ -53,11 +53,18 @@ void HalPowerManager::setPowerSaving(bool enabled) {
...
+  gpio_set_direction(GPIO_SPIWP, GPIO_MODE_OUTPUT);
+  gpio_set_level(GPIO_SPIWP, 0);
   // Ensure that the power button has been released to avoid immediately turning back on if you're holding it
   while (gpio.isPressed(HalGPIO::BTN_POWER)) {
     delay(50);
     gpio.update();
   }
+  esp_sleep_config_gpio_isolate();
+  gpio_hold_en(GPIO_SPIWP);
+  pinMode(InputManager::POWER_BUTTON_PIN, INPUT_PULLUP);
```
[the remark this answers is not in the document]
ngxson: hmm, nice find. indeed, the `esp_wifi_bt_power_domain_off` in my analysis was mislabeled, that was `gpio_deep_sleep_hold_en` which is compiled into `_DAT_6000808c = _DAT_6000808c & 0xffffbfff | 0x800;`

lib/hal/HalPowerManager.cpp
```diff
@@ -53,11 +53,19 @@ void HalPowerManager::setPowerSaving(bool enabled) {
 }
 
 void HalPowerManager::startDeepSleep(HalGPIO& gpio) const {
+  // Pre-sleep routines from the original firmware
+  constexpr gpio_num_t GPIO_SPIWP = GPIO_NUM_13;
+  gpio_set_direction(GPIO_SPIWP, GPIO_MODE_OUTPUT);
+  gpio_set_level(GPIO_SPIWP, 0);
```
ngxson: interestingly, switching this to `gpio_set_level(GPIO_SPIWP, 1)` and now RTC memory is preserved across sleep, cc @osteotek

iandchasse: I believe this has been found before, RTC is preserved but idle sleep current from the battery is 3-4mA, so it's not feasible to keep it enabled. TRMNL firmware for the X4 does this (I think?), but the battery only lasts for a few days

ngxson: yeah it will definitely use more battery, the current leakage to other components can be significant. just saying that now we know why/how

I'm not entire sure how TRMNL keep the device "awake", but in my experience, a light sleep also does the trick. although, it consume significantly more power than deep sleep

edit: they don't use deep sleep at all: https://github.com/usetrmnl/trmnl-firmware/issues/313

iandchasse: https://discord.com/channels/1440809779944624179/1440810116394782843/1465307061834354710

Discussion from Larry Bank in discord here

ngxson: hmm ok, so they use the same trick: https://github.com/usetrmnl/trmnl-firmware/blob/98e96ac3d341dd5535518570b6c499998c442b2a/src/bl.cpp#L1912-L1915

the difference is that the pin value is not explicitly set, so it will hold the previous value, which is high (meaning battery always connected)

that further comfirms my finding here though, gpio 13 high = battery kept connected, low = disconnect battery

lib/hal/HalPowerManager.cpp
```diff
@@ -58,7 +58,20 @@ void HalPowerManager::startDeepSleep(HalGPIO& gpio) const {
...
+  // Pre-sleep routines from the original firmware
+  // GPIO13 is connected to battery latch MOSFET, we need to make sure it's low during sleep
+  // Note that this means the MCU will be completely powered off during sleep, including RTC
+  constexpr gpio_num_t GPIO_SPIWP = GPIO_NUM_13;
+  gpio_set_direction(GPIO_SPIWP, GPIO_MODE_OUTPUT);
+  gpio_set_level(GPIO_SPIWP, 0);
+  esp_sleep_config_gpio_isolate();
+  gpio_deep_sleep_hold_en();
+  gpio_hold_en(GPIO_SPIWP);
+  pinMode(InputManager::POWER_BUTTON_PIN, INPUT_PULLUP);
```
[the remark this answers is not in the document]
ngxson: should be fine here, it should never fail in practice, even official fw has no check for these

znelson:
I've been using this change for many days and haven't had a hang on sleep; in my experience it seems to fix #1263. I think we should move forward with this change. @crosspoint-reader/firmware-maintainers

znelson:
Not a lot of activity from reviewers lately. I'm going to move ahead with this one and hopefully we can get some feedback from folks on the other side of #1263.

PR #1306, opened 2026-03-04
refactor: Compress I18n string data (~88 KB flash savings)

jpirnay:
## Summary

* **What is the goal of this PR?** All 17 language string tables are now stored as zlib-compressed blobs in flash instead of raw const char* arrays. The active language is decompressed once at boot (and again on language change) into a small static RAM buffer — the tr() / I18N.get() API is completely unchanged.
  - gen_i18n.py gains a compression step using Python's built-in zlib module (no new dependencies). It also prints a per-language size table and flash/RAM summary on every build.
  - Uses the existing InflateReader + uzlib libraries already in the project (used by ZIP, PNG, and font decompression).

## Additional Context
* RAM cost at runtime: ~12 KB  (string buffer 10 KB + pointer table 1356 B)
* Compressed format: all strings for one language concatenated in StrId order, each null-terminated, then zlib.compress(level=9). 
* Decompressed with InflateReader::init(false) + skipZlibHeader() + read() (one-shot, no heap allocation).
* String table rebuilt by scanning null terminators in the flat decompressed blob — O(n) once at init, then O(1) per lookup.
* I18N_MAX_DECOMPRESSED_SIZE is computed by the generator and baked into I18nKeys.h to size the static buffer exactly.
* Adding a new language automatically updates the buffer size and LANG_DATA table on the next build.

```
Loaded 17 languages, 339 string keys
Compressing string data...
...
Flash usage (estimated compiled binary):
  Before: 139,516 B  (136 KB)  [strings + pointer arrays]
  After:   48,407 B  (47 KB)  [compressed blobs + LANG_DATA table] 
  Saved:   91,109 B  (88 KB)
  RAM cost at runtime: ~12 KB  (string buffer 10 KB + pointer table 1356 B)
  ```
 Real saving:
```
RAM:   [===       ]  32.8% (used 107492 bytes from 327680 bytes)   
Flash: [========= ]  90.2% (used 5910989 bytes from 6553600 bytes) 
```
vs old
```
RAM:   [===       ]  29.0% (used 95084 bytes from 327680 bytes)    
Flash: [========= ]  91.5% (used 5999763 bytes from 6553600 bytes) 
```

---

### AI Usage


ngxson:
Just my opinion but trading 1.3% flash for 3.8% RAM is not a good deal.. RAM pricing is high recently, pun intended

jpirnay:
> Just my opinion but trading 1.3% flash for 3.8% RAM is not a good deal.. RAM pricing is high recently, pun intended

I get you but with the amount of languages flowing in the problem grows...

I would still prefer to have just a basic amount of languages baked in and anything else would require a conscious 'load additional languages' from a internet source

ngxson:
> I get you but with the amount of languages flowing in the problem grows...

Will RAM usage also grow when more strings are added? (I mean strings, not languages)

> I would still prefer to have just a basic amount of languages baked in and anything else would require a conscious 'load additional languages' from a internet source

Is compression still relevant if languages are downloaded from the internet?

jpirnay:
I have a working PR that implements the load on demand: #1314 

| Variant | Flash | Delta | Ram | Delta |
| -- | -- | -- | -- | -- |
| all  languages | 5.999.763 | - | 95.084 | - |
| compressed (PR #1306 ) | 5.910.989 | -            88.774 | 107.492 | +          12.408|
| core languages (this PR) | 5.919.329 | -            80.434 | 96.492 | +            1.408|

No compression applied +1.4k generic ram consumption +15k when a non core language is used. 

Advantages: still 80k flash savings, new languages do no longer require a new firmware build, as they are loaded dynamically on request

https://github.com/user-attachments/assets/4d35f3ac-ea9f-4823-880e-16b9eb1232ba

(apologies for the distorted image, will need to do it again, but gives an idea of the workflow)

jpirnay:
Converted to draft, as a lot of new ideas emerge from this. Linking to discussion #1316

jpirnay:
Competes with a couple of other PRs

PR #1310, opened 2026-03-04
fix: Use font metrics for combining mark positioning

znelson:
## Summary

**What is the goal of this PR?** (e.g., Implements the new feature for file uploading.)

Combining diacritical marks (U+0300–U+036F) were positioned using a heuristic that centered them at the midpoint of the base glyph's **advance width**. This worked acceptably for Bookerly but produced visibly off-center marks for Noto Sans due to a fundamental difference in how the two fonts design their combining mark metrics.

Builds on the work of #1037.

## The problem

The two built-in body fonts encode combining mark `left` offsets with very different conventions:

| Mark | Bookerly `left` | Noto Sans `left` |
|---|---|---|
| U+0301 (acute) | -2 | -10 |
| U+0300 (grave) | -5 | -15 |
| U+0302 (circumflex) | -5 | -5 |
| U+0323 (dot below) | -2 | -11 |

Noto Sans uses large negative `left` values because its marks are designed for placement at the post-advance cursor position, with `left` pulling the bitmap back over the base glyph. Bookerly uses small offsets because its marks sit closer to the glyph origin. The old `advance/2` centering split the difference poorly — it happened to land close to correct for Bookerly but placed Noto Sans marks roughly 6px left of center on a typical lowercase letter.

There was also a bug in the vertical gap heuristic. It unconditionally computed a `raiseBy` value to prevent above-baseline marks from colliding with tall base glyphs, but it applied the same logic to **below-baseline** marks like cedilla (U+0327), dot below (U+0323), and ogonek (U+0328). For those marks, the math produced a large positive raise (e.g., 24px for dot-below on 'a'), launching them above the x-height instead of keeping them below the baseline.

## The fix

**Horizontal positioning**: Instead of centering at `advance/2`, align the mark bitmap's visual midpoint directly over the base glyph bitmap's visual midpoint. This uses the base glyph's actual `left` and `width` rather than its advance width, producing correct results regardless of how the font encodes its mark offsets.

**Vertical positioning**: The raise heuristic now checks `markTop - markHeight > 0` and skips below-baseline marks entirely, leaving them at their font-designed position.

**Consolidation**: The shared math is extracted into two `constexpr` helpers (`combiningMark::centerOver` and `combiningMark::raiseAboveBase`) in `EpdFontData.h`, eliminating the previously triplicated inline calculations across `drawText`, `drawTextRotated90CW`, and `getTextBounds`. The `MIN_COMBINING_GAP_PX` constant is also centralized as `combiningMark::MIN_GAP_PX`.

| Before | After |
| -- | -- |
|  |  |
|  |  |

---

### AI Usage

Did you use AI tools to help write this code? _**YES to analyze differences between Noto Sans and Bookerly font metrics**_


znelson:
~~I think this can be done better by combining with some of the improvements I'm learning about for Thai support.~~

Actually, maybe better to iterate. This is clearly an improvement over the existing behavior because combining marks in Noto Sans are no longer shifted way left, and lower combining marks in both fonts now stay at the bottom.

andrewb1269 (dismissed):

Uri-Tauber (approved):

Uri-Tauber:
Hi @znelson, I’ve updated this and believe I successfully merged your work here with the changes from #1413. Could you please review my updates and confirm everything works correctly on your device?

znelson:
Thanks @Uri-Tauber! I'll try it out today

znelson:
The merge + subsequent fixes look good to me. This change is imperfect, but an improvement over master in all of my tests. I think we can iterate on it going forward.

Issue #1186, opened 2026-02-25
fix: hyphenation broken for words containing apostrophes (e.g. Italian/French contractions)


DanieleDDM:
### Affected Version

v 1.1.1

## Problem

Words containing apostrophes are never hyphenated. This causes the preceding line to have excessive spacing when justification is enabled, because the entire token (e.g. "all'improvviso") is treated as an unsplittable unit.

Example: in Italian text, "all'improvviso" cannot be broken, so the line before it stretches badly.

![Image](https://github.com/user-attachments/assets/d59b0bff-e2d4-4b2f-9f08-73307597d948)

## Root cause

In `LiangHyphenation.cpp`, `buildAugmentedWord()` rejects any word that contains a non-letter codepoint. An apostrophe (both ASCII `'` 0x27 and typographic `'` U+2019) triggers this rejection, causing the entire token to be silently skipped with zero hyphenation points returned.

## Fix

Treat apostrophes between two alphabetic characters as segment separators, similar to how explicit hyphens are already handled in `Hyphenator::breakOffsets()`.

Changes:
- Add `isApostrophe(uint32_t cp)` helper in `HyphenationCommon`
- In `breakOffsets()`, detect contractions and run Liang patterns on each alphabetic segment around the apostrophe
- Only add a break point *at* the apostrophe position if the prefix before it is ≥ 3 characters, to avoid short articles like "l'" or "d'" being stranded alone at the end of a line

Files changed:
- `lib/Epub/Epub/hyphenation/HyphenationCommon.h`
- `lib/Epub/Epub/hyphenation/HyphenationCommon.cpp`
- `lib/Epub/Epub/hyphenation/Hyphenator.cpp`

This also benefits French ("l'homme", "d'accord") and any other language that uses apostrophe-based contractions.

I am not an experienced developer — this analysis and fix were generated with the help of an AI assistant (Claude code). I hope the information is useful and accurate enough to help resolve the issue.

### Steps to Reproduce

1. Set the reader language to Italian (or French)
2. Enable hyphenation in the reader settings
3. Open any EPUB with Italian text that contains words with apostrophes e.g. "all'improvviso", "l'uomo", "dell'arte")
4. Observe that lines containing those words have excessive spacing — the word after the apostrophe is never hyphenated and is treated as an unsplittable block

### Expected Behavior

[...]

PR #1318, opened 2026-03-05
fix: Add special handling for apostrophe hyphenation

jpirnay:
## Summary

* **What is the goal of this PR?** Fixing / extending the hyphenation logic to deal with words containing an apostophe as raised in #1186 

## Additional Context

---

### AI Usage

Did you use AI tools to help write this code? _**PARTIALLY**_ (as the user provided a thorough analysis that I followed)

### DIFF
--- a/lib/Epub/Epub/hyphenation/Hyphenator.cpp
+++ b/lib/Epub/Epub/hyphenation/Hyphenator.cpp
@@ -143,7 +143,7 @@ void sortAndDedupeBreakInfos(std::vector<Hyphenator::BreakInfo>& infos) {
 
   infos.erase(std::unique(infos.begin(), infos.end(),
                           [](const Hyphenator::BreakInfo& a, const Hyphenator::BreakInfo& b) {
-                            return a.byteOffset == b.byteOffset && a.requiresInsertedHyphen == b.requiresInsertedHyphen;
+                            return a.byteOffset == b.byteOffset;
                           }),
               infos.end());
 }


lib/Epub/Epub/hyphenation/Hyphenator.cpp
```diff
@@ -59,6 +59,83 @@ std::vector<Hyphenator::BreakInfo> buildExplicitBreakInfos(const std::vector<Cod
...
+          isAlphabetic(cps[i + 1].value)) {
+        size_t leftPrefixLen = 0;
+        for (size_t j = segmentStart; j < i; ++j) {
+          if (isAlphabetic(cps[j].value)) {
+            ++leftPrefixLen;
+          }
+        }
+
+        // Avoid stranding very short clitics like "l'" or "d'" at line end.
+        if (leftPrefixLen >= 3) {
```
znelson: Will English contractions like "can't" and "wouldn't" now sometimes split to "can' | t" and "wouldn' | t"?

jpirnay: Good catch, changed it

drbourbon:
Hope this gets merged soon, because it's a *huge* page rendering enhancement, especially when using big fonts with apostrophe based languages. Thanks!

IMG_1034

lukestein:
> This looks reasonable to me. Thanks for fixing the English contraction case. I think it'd be good to move forward with this and get user feedback around whether it's working as expected in languages with apostrophes in more common usage.

I see @znelson referred to "fixing the English contraction case" but alas @jpirnay I'm getting problems with contractions getting broken at the apostrophe. See screenshots in #1403.

znelson:
@lukestein indeed, should be a simple tweak, hopefully 🙂

PR #1325, opened 2026-03-05
chore: change label while on settings tab actions

jpirnay:
## Summary

* **What is the goal of this PR?** The "Toggle" label while on the "tab" actions of settings screen was misleading. Will show "Select" now , while "Toggle" remains in place for all 'real' settings

## Additional Context

---

### AI Usage

Did you use AI tools to help write this code? _**NO **_

### DIFF
--- a/src/activities/settings/SettingsActivity.cpp
+++ b/src/activities/settings/SettingsActivity.cpp
@@ -243,8 +243,10 @@ void SettingsActivity::render(RenderLock&&) {
       true);
 
   // Draw help text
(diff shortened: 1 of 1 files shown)


znelson:
I agree that "Toggle" doesn't feel right, but I'm not sure about "Select" either. "Select" seems like it would go _down_ to dive into the category of settings. 

"Next"? "Right"? "Reader"/"Controls"/"System"/"Display" (the name of the next section)?

th0m4sek:
Maybe "Change" ?

PR #1326, opened 2026-03-06
feat: add Greek and Letterlike Symbols to all reader fonts

BryanFran:
## Summary

* **What is the goal of this PR?** Enables Noto Sans, Bookerly, and OpenDyslexic to render Greek letters (φ, λ, θ, α, Δ, π, ρ, β, ψ, Φ, μ, ϵ, γ, Λ, ζ) and Letterlike Symbols (ℓ, ℝ, ℤ, ℂ, ℚ) that previously displayed as replacement characters (▯).

  - Uncomment Greek & Coptic (U+0370–U+03FF) in `fontconvert.py` — the range was disabled even though all three source font files already contain these glyphs.
  - Add Letterlike Symbols (U+2100–U+214F) to `fontconvert.py` (ℓ, ℝ, ℤ, ℂ, ℚ, etc.).
  - Regenerate all 49 reader-font headers: Bookerly (16), NotoSans (17), OpenDyslexic (16).
  - Grow `app0`/`app1` partitions by 256 KB each in `partitions.csv` to accommodate the additional glyph data.

## Additional Context

The ~354 KB of new glyph data pushed the firmware past the previous 6.25 MB app partition limit (the project was already at ~97.5% before this change). The partition table is updated as follows:

| Partition | Before | After |
|-----------|--------|-------|
| app0 / app1 | 6.25 MB each | 6.50 MB each |
| spiffs | 3.375 MB | 2.875 MB |

SPIFFS holds only small settings files (font choice, reading progress), so the 512 KB reduction has no practical impact on users. Total flash stays at 16 MB. New firmware headroom: ~69 KB.

**Note for existing users:** changing the partition table requires a full flash (not OTA). OTA updates will continue to work normally between firmware versions that share this new layout.

Fonts are `static const` arrays in flash — **no RAM impact** on the 380 KB DRAM budget.

All 48 compressed font files verified with `verify_compression.py` (48 passed, 0 failed).

The motivation came from reading a mathematical geodesy EPUB (*Map of the World: An Introduction to Mathematical Geodesy*, Vermeer & Rasila, CRC Press 2019) that makes heavy use of Greek letters and letterlike symbols throughout its inline text.

---

### AI Usage

This PR was developed with Claude Code (claude-sonnet-4-6). The AI identified the root cause (commented-out Unicode range in `fontconvert.py`), verified TTF/OTF glyph coverage using fontTools across all three font families, regenerated all 49 font headers, validated the output with `verify_compression.py`, and diagnosed/fixed the partition overflow.


Uri-Tauber:
Changing the partition map like this will break the OTA updates.

znelson:
More discussion in #951 around partition changes

mcrosson:

itsthisjustin:
So as we are completely re-doing the fonts system, I am going to close this for now. The sd fonts system is capable of adding these symbols in a way that doesn't add to our flash constraints.

PR #1327, opened 2026-03-06
feat: Back-end firmware support for SD card fonts

### LINKED ISSUE #930: Support for User Fonts
We need support for custom fonts:

1. Ability to select user-provided font in binary format (pre-rendered for specific size(s))
2. Tool to convert regular ttf fonts into binary format (we already have python scripts that can be used for that) at specified sizes
3. Potentially some glyph caching
4. Potentially use SPIFFS (on-device flash) to store the font data.
5. Potentially we need to refactor the font family logic to allow loading a full font family
6. Potentially we want to add more size options or allow any size font to be loaded.

Current status:
1. There is some detailed discussion in #160.
2. There is a PR draft: #589 
3. It seems this fork already has work in this direction: https://github.com/ngxson/pluspoint-reader

adriancaruana:
## Summary

Some necessary back-end firmware changes to support SD card fonts. Note, this PR does not include the SD Card fonts themselves -- this is deferred to https://github.com/crosspoint-reader/crosspoint-reader/pull/1392.

- A custom binary format for pre-generated glyph data `.cpfont`, to be stored on the SD card
- Dynamic font list that supports automatic discovery of fonts on the SD card at `.crosspoint/fonts/`
- Changes to the rendering pipeline to support glyph retrieval either from Flash or the SD card.
- Some changes to ensure that SD card-based rendering is as performant as Flash-based rendering. Rendering performance of Bookerly small is basically identical to Flash.

## Some things we need to consider

- We need to implement a way for fonts to be:
  - Compiled (additional build step, since we probably don't want to add these to the git tree, like I have)
  - Distributed (i.e., as part of a release)
  - and to be added to the SD card (Manual copying works fine, but for many users, potentially it would be more convenient to have a method via WIFI/Bluetooth that downloads them directly to the SD card).
- I believe we cannot remove Flash-based support entirely, since we need a font in case there are none on the SD card.

---

### AI Usage

Did you use AI tools to help write this code? Partially.


Uri-Tauber:
It sound like two seperate PRs combined. You should separate them. 
BTW, how your implementation differ from #455?

adriancaruana:
@Uri-Tauber, thank you for the consideration and feedback. Not sure what you mean by separating into two PRs. There's nothing different about CJK fonts except the number of glyphs supported by the typeface. I guess you could say CJK is a feature of this branch by default.

As for how this branch differs to #455: Aside from superficial differences (e.g., binary font format), the main difference is that the LRU-cache approach of #455 is a non-starter for languages that don't have a small, common set of glyphs (e.g., basically anything non-Latin). Take CJK, for example: For the LRU cache to be effective, its capacity would need to be much larger than the RAM can support. Instead, this branch builds on the 'prewarm' approach (implemented in #1056, which is on the verge of being merged). Prewarm bulk-reads all glyphs needed for a page in a single sequential SD card pass, then serves them from RAM -- no per-glyph seeks during rendering. This minimises SD card reads (or Flash reads and decompressions for non-SD card fonts), maximises cache-reuse, minimises heap usage, and crucially also maintains solid performance for languages with extended glyph sets. For example, with this branch and Bookerly Small, I get basically the same render performance of Flash vs SD-card (697ms vs 681ms).

Just to clarify, #1056 is not a CJK-feature, it's just a better approach than the LRU cache that's currently on master.

Uri-Tauber:
I discussed the LRU approach with `@ngxson` in #596—please take a look there for the full context. The conclusion was that an LRU cache with 128 slots should work well for about 99% of languages. Chinese may be the main exception where LRU might not perform as effectively.

The core idea behind using LRU isn’t to keep all glyphs for a language in RAM, but rather to maintain a low miss rate. If rendering a page results in, say, around five SD card reads, that would already be a very good outcome.

Also, I’m not entirely sure why we would still need CJK glyphs stored in flash if `SD card fonts` are already implemented. Doesn’t that somewhat defeat the purpose of freeing up flash space for other features?

I'm not saying your approach is less good. Just wondering if using LRU will result in even better performance.

happyjake:
Loading font glyphs from sdcard is way to slow. I think it's a dead end. I tried that in [my fork of papyrix-reader](https://github.com/happyjake/papyrix-reader). It does not feel snappy like when I'm reading English books with crosspoint readers.

adriancaruana:
Thank you @Uri-Tauber for sharing the additional context in #596. Let me share my experience with trying to make the LRU work. I hope this clears things up.

As you suggest, for most languages, an LRU works well. I should know, I implemented the LRU that's currently on master. The problem is that cache misses are very expensive; for a cache to be effective, it needs close to a 100% hit rate. But for Chinese, Japanese (kanji, hiragana, katakana), and to a lesser extent Korean (has slightly better cache hit rate than the formers), a cache hit rate of about 50-60% is not good enough, and rendering becomes dominated by the cost of cache misses.

Conceptually, the prewarm approach should be the optimal solution for expensive glyph reads: 1. Figure out what glyphs are needed for a page, 2. Read/decode each glyph once only, 3. Construct the cache from these glyphs, and use for all rendering phases. This is the solution I came up with in #1056 because I just couldn't make an LRU work for CJK text (I allude to this in its PR description).

adriancaruana:
I'm not following your comment @happyjake. I get equivalent performance on Latin text for Flash vs SD-card fonts with this PR. Please let me know if you are not seeing this on your end. Chinese text is about 32% slower (which is pretty incredible tbh). NB: I upgraded my SD card from the one that came with the device.

happyjake:
Hi, @adriancaruana . I flashed and tried your code. It's way better than I expected—much better performance than my approach. I can still feel the lag when turning pages, but it's indeed usable. I verified it on two devices side by side with the same Chinese books. (love X4, so I bought both colors.)

I'm trying a new approach: to squeeze the entire GB2312 character set (~7,000 hanzi covering 99.5%+ of Simplified Chinese) at two sizes—10 pt for UI and 14 pt for book rendering—into the 3.3MB SPIFFS partition, loaded via mmap. This eliminates loading glyphs from the SD card when turning pages. It achieves performance on par with Latin rendering. I know this is more limited in character set coverage than what this PR provides. Just exploring the idea to see what's possible.

adriancaruana:
@happyjake I'm glad you found the implementation performant. (I'm still waiting for my second colour to arrive!)

If you measure the performance directly in the firmware, I think you'll find that flash (fast reads, but slow decompression) and SD card (slow reads, but no decompression) are roughly equivalent. 

As for storing CJK fonts in flash: it's definitely possible (as I explored in https://github.com/crosspoint-reader/crosspoint-reader/pull/1059), but @ngxson convinced me that even if we use font compression and as much flash as possible for fonts (e.g., 10 MB of 15 MB) we would only be able to provide full glyph coverage for four fonts, which didn't provide a very good UX. Not too mention that it is just kind of wasteful to allocate so much space for fonts that some people won't even use. I think it is better to use flash for software features, not for assets.

lib/I18n/translations/belarusian.yaml
```diff
@@ -173,6 +173,7 @@ STR_NEXT_PREV: "Наперад/Назад"
 STR_BOOKERLY: "Bookerly"
 STR_NOTO_SANS: "Noto Sans"
 STR_OPEN_DYSLEXIC: "Open Dyslexic"
+STR_NOTO_SANS_CJK: "Noto Sans CJK"
```
ngxson: this newly-added string doesn't seem to be used

lib/EpdFont/scripts/output/Bookerly-SD/Bookerly-SD_12_bold.cpfont
ngxson, line 1: should we bundle all variants into one big file? IMO single file is a better UX overall

src/main.cpp
```diff
@@ -192,6 +196,71 @@ void enterDeepSleep() {
...
+  } else {
+    LOG_DBG("MAIN", "SD font family not found: %s (clearing)", wantedFamily);
+    SETTINGS.sdFontFamilyName[0] = '\0';
+  }
+}
+
+// Map fontSize enum (SMALL=0, MEDIUM=1, LARGE=2, EXTRA_LARGE=3) to point sizes.
+// SD card fonts use the standard sizes: 12, 14, 16, 18.
+static constexpr uint8_t FONT_SIZE_TO_PT[] = {12, 14, 16, 18};
+static constexpr int FONT_SIZE_TO_PT_COUNT = sizeof(FONT_SIZE_TO_PT) / sizeof(FONT_SIZE_TO_PT[0]);
```
ngxson: UX-wise, if the font conversion script always produce the same set of size, we can simplify by remove this check.

And if font is generated into one single file as I suggested above, then there will be no chance of missing a font size

src/main.cpp
```diff
@@ -31,6 +33,8 @@ MappedInputManager mappedInputManager(gpio);
 GfxRenderer renderer(display);
 ActivityManager activityManager(renderer, mappedInputManager);
 FontDecompressor fontDecompressor;
+SdCardFontRegistry fontRegistry;
+SdCardFontManager sdFontManager;
```
ngxson, line 40: from an API designer perspective, it's better to hide all of these implementation details behind something like `FontManager`

we should only have `FontManager.begin()` in `main.cpp` as the entrypoint for the whole system

src/main.cpp
```diff
@@ -221,6 +290,25 @@ void setupDisplayAndFonts() {
   renderer.insertFont(UI_10_FONT_ID, ui10FontFamily);
   renderer.insertFont(UI_12_FONT_ID, ui12FontFamily);
   renderer.insertFont(SMALL_FONT_ID, smallFontFamily);
+
+  // Discover SD card font families
+  fontRegistry.discover();
+
+  // Register SD font ID resolver so CrossPointSettings can look up font IDs
+  SETTINGS.sdFontIdResolver = resolveSdFontId;
```
ngxson: IMO a callback function like this is usually not recommended. If we design it to be a callback, that means we may set it dynamically to another value at runtime. But it's not the case here.

Instead, better to just include font system inside settings, and the linker will be responsible to link the function and do further optimizations if needed.

ngxson (commented):
Since this PR is still draft, here is my quick review to refine the direction of this PR

lib/EpdFont/scripts/output/Bookerly-SD/Bookerly-SD_12_bold.cpfont
adriancaruana: Yeah, this is a good idea. I've implemented this.

adriancaruana:
This is now ready for review @znelson @ngxson

mcrosson:
@crosspoint-reader/firmware-maintainers Is anyone available to review this PR? This is a pre-req for #1392 (sd card font support) which was on @znelson 's list of 'ambitious but desirable' PRs to get into Crosspoint at some point.

Uri-Tauber:
@mcrosson I believe this PR is the actual font support, while #1392 introduces a range of additional features around the same area (some of which feel a bit over-engineered to me).
I’d really like to see this PR merged, but it’s quite large and honestly beyond what I can thoroughly review right now. I’ll try it out on my reader and share feedback on how the UX is affected when loading fonts from the SD card.

hugovincent:
> I believe we cannot remove Flash-based support entirely

Is there any difference to either power consumption, or performance given SD card variability? While I don't have measurements, I suspect the flash path is going to consume less energy and offer a longer battery life, but it might be marginal. Secondly, and as you allude to, the bundled SD cards aren't the best quality, and in any case, all SD cards vary in performance depending on both how much wear they've experienced, and how much spare capacity they have available. It would be nice to compare performance with a near-full bundled SD card, and not just an upgraded high quality one.

itsthisjustin:
> > I believe we cannot remove Flash-based support entirely
> 
> Is there any difference to either power consumption, or performance given SD card variability? While I don't have measurements, I suspect the flash path is going to consume less energy and offer a longer battery life, but it might be marginal. Secondly, and as you allude to, the bundled SD cards aren't the best quality, and in any case, all SD cards vary in performance depending on both how much wear they've experienced, and how much spare capacity they have available. It would be nice to compare performance with a near-full bundled SD card, and not just an upgraded high quality one. 

This has been a thing said a lot in the community. I have had 4 of these personally and 1 to a friend, never changed the SD card from what came with the device. I don't know if other users are filling them up more, writing to them more, or what the difference is (could truly be a hw different too I guess) but I've never seen any issues with the bundled cards. My 2 cents. It does fly in the face of community recommendation to upgrade the card though.

mcrosson:
Unless folk are putting forth very real, documented differences in 'included sd card vs other models' I think going round and round re sd card performance is putting the cart before the horse.

If I've missed some actual discussion with numbers/data, please point it out but I've not seen anything pop up yet.

Screen refresh, indexing pages, rending pages and the simple fact we have to load glyphs per page are the bigger delays currently. Even if you had 1ms or lower response times from an sd card (this wont happen in the real world on an esp32), you'll still run into the fact nothing is 'fast' and we have to manage code paths very carefully and balance speed vs ram use. Also bear in mind the esp32 will never match the actual performance of a 'fast sd card'. The device itself is just too slow.

This is an endemic concern with this PR and the dictionary ( #1402 ) PR.

Additionally: we are dealing with user provided font files. There are a lot of fonts and more than a few can and will cause problems (as recently shown in the #1693 discussion). On top of that, it's pretty clear SPIFFS wont be big enough to hold some CJK fonts. No matter what we do sd card *must* be used for this feature. Thats been made pretty clear in #1693 already.

Can we put the sd card performance, SPIFFS performance, flash performance theorization stuff elsewhere and leave this PR to the work that needs doing for sd card based font support?

itsthisjustin:
> Unless folk are putting forth very real, documented differences in 'included sd card vs other models' I think going round and round re sd card performance is putting the cart before the horse.
> 
> If I've missed some actual discussion with numbers/data, please point it out but I've not seen anything pop up yet.
> 
> Screen refresh, indexing pages, rending pages and the simple fact we have to load glyphs per page are the bigger delays currently. Even if you had 1ms or lower response times from an sd card (this wont happen in the real world on an esp32), you'll still run into the fact nothing is 'fast' and we have to manage code paths very carefully and balance speed vs ram use. Also bear in mind the esp32 will never match the actual performance of a 'fast sd card'. The device itself is just too slow.
> 
> This is an endemic concern with this PR and the dictionary ( #1402 ) PR.
> 
> Additionally: we are dealing with user provided font files. There are a lot of fonts and more than a few can and will cause problems (as recently shown in the #1693 discussion). On top of that, it's pretty clear SPIFFS wont be big enough to hold some CJK fonts. No matter what we do sd card _must_ be used for this feature. Thats been made pretty clear in #1693 already.
> 
> Can we put the sd card performance, SPIFFS performance, flash performance theorization stuff elsewhere and leave this PR to the work that needs doing for sd card based font support?

100% agree with this

mcrosson:
@adriancaruana can i ask why you chose `8` for `OVERFLOW_CAPACITY`?

Note: I'm only curious, I don't have a problem with the value.

lib/EpdFont/scripts/fontconvert_sdcard.py
osteotek, line 1: we probably don't need two separate scripts for fontconverting and generating fonts (I mean fontconvert.py and fontconvert_sdcard.py , convert-builtin-fonts.sh and generate-sd-fonts.sh)

mcrosson: Agreed. However, for now these two scripts serve very different purposes. One is related to sd font 'stuff' and the other the built in fonts.

I think it unwise to try to tinker with these at the moment. Especially since it opens up a much bigger discussion around UI fonts and built in font support.

I think we should focus on sd fonts, stabilizing where needed and then looping around to UI and built in font items so we can stay focused.

znelson:
Combined with #1392 and updated in #1884.

PR #1349, opened 2026-03-07
fix: long-press back navigates to book's directory (#1117)

### LINKED ISSUE #1117: Long-pressing back button takes you to home directory instead of the open book’s directory
### Affected Version

v1.1.0

### Describe the Bug

I had to revert back to v1.0.0 because the new release doesn’t allow you to return to a book’s directory via use of the back button, it only allows back to the main directory or to the home screen.

This is probably a BUG and not as intended, since it also shows the name of the book’s directory at the top of the main directory, where it would normally say “SD CARD.”

### Steps to Reproduce

1. Open a book
2. Long-press the back button
3. The home directory appears and has the name of the book’s directory displayed at the top

### Expected Behavior

Closing a book via long-pressing the back button should return you to the book’s directory (ideally with the current book still selected) and not to the home directory.

```shell

```

Cache8063:
## Summary

- Long-pressing the back button while reading now navigates to the book's parent directory instead of root
- Previously, the full book file path was passed to `goToFileBrowser()`, which failed to open it as a directory

Fixes #1117

## Root cause

Each reader activity (Epub, Txt, Xtc) passed the full book path (e.g., `/Books/SciFi/MyBook.epub`) to `goToFileBrowser()`. The file browser tried to open this as a directory, failed, and showed an empty listing. The correct `extractFolderPath()` logic already existed in `ReaderActivity::goToLibrary()` but wasn't used by the individual reader activities.

## Change

In `EpubReaderActivity.cpp`, `TxtReaderActivity.cpp`, and `XtcReaderActivity.cpp`: extract the parent directory from the book path before passing to `goToFileBrowser()`.

## Test plan

- [ ] Open a book in a subdirectory (e.g., `/Books/SciFi/MyBook.epub`)
- [ ] Long-press the back button
- [ ] File browser opens showing `/Books/SciFi/` contents (not root)
- [ ] Works for EPUB, TXT, and XTC file types

### DIFF
--- a/src/activities/reader/EpubReaderActivity.cpp
+++ b/src/activities/reader/EpubReaderActivity.cpp
@@ -183,7 +183,8 @@ void EpubReaderActivity::loop() {
     if (epub) {
       auto path = epub->getPath();
       auto lastSlash = path.find_last_of('/');
-      activityManager.goToFileBrowser(lastSlash == std::string::npos || lastSlash == 0 ? "/" : path.substr(0, lastSlash));
+      activityManager.goToFileBrowser(lastSlash == std::string::npos || lastSlash == 0 ? "/"
+                                                                                       : path.substr(0, lastSlash));
     } else {
       activityManager.goToFileBrowser("");
     }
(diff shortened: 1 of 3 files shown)


Cache8063:

PR #1350, opened 2026-03-07
fix: preserve CSS text-align:center when user sets Justify (#1029)

### LINKED ISSUE #1029: CSS styled element is not centered (When Paragraph Alignment is set to Justify)
### Affected Version

22b96ec22abbd31f9a94f16045edaee378b887ad

### Describe the Bug

Certain text elements styled with CSS are not centered despite CSS saying it should be. Example screenshots posted below.

Edit: Happens both with Embedded Style `ON` and `OFF`, but disappears when Reader Paragraph Alignment is set to `Book's Style` instead of `Justify`
In my opinion `Justify` should not override centering of paragraphs; only `Left`, `Center` and `Right`

The CSS in question:
```css
/* Additional */
.orn {text-indent: 0; text-align: center; margin: 1em 0 1em 0}
```

As seen in Calibre
Image

Source of the segment in question in calibre
Image

Source of the CSS
Image

As seen on the X4
Image

### Steps to Reproduce

1. Open an Epub with a `<p class="randcenter">*</p>`
2. Style that epub with a css that centers this class ( `.randcenter {text-align: center;}` )
3. The `*` should not be centered.

( Minimally viable epub to repro: https://drive.google.com/file/d/1ziT9hd4hZ8P6QYLJyfTzRl15V4G_mBi_/view?usp=sharing )

### Expected Behavior

The `*` should be centered.

```shell

```

Cache8063:
## Summary

- Preserve CSS `text-align: center` even when the user's paragraph alignment is set to Justify, Left, or Right
- Centered text almost always serves a structural/decorative purpose (ornamental dividers, title pages, dedications) and should not be overridden

Fixes #1029

## Root cause

`BlockStyle::fromCssStyle()` unconditionally replaced the CSS text-align with the user's paragraph alignment setting, except in "Book's Style" mode. This meant `text-align: center` from the book's CSS was ignored.

## Change

3-line addition in `lib/Epub/Epub/blocks/BlockStyle.h` — add an `else if` branch that preserves `CssTextAlign::Center` from the book's stylesheet, regardless of the user's alignment setting.

## Test plan

- [ ] EPUB with CSS `text-align: center` (ornamental dividers, title pages) renders centered with Justify setting
- [ ] Normal paragraphs still respect user's Justify/Left/Right setting
- [ ] "Book's Style" mode is unaffected

### DIFF
--- a/lib/Epub/Epub/blocks/BlockStyle.h
+++ b/lib/Epub/Epub/blocks/BlockStyle.h
@@ -86,9 +86,13 @@ struct BlockStyle {
       blockStyle.textIndentDefined = true;
     }
     blockStyle.textAlignDefined = cssStyle.hasTextAlign();
-    // User setting overrides CSS, unless "Book's Style" alignment setting is selected
+    // User setting overrides CSS, unless "Book's Style" alignment setting is selected.
+    // Exception: CSS text-align:center is always respected, as it typically serves a
+    // structural/decorative purpose (ornamental dividers, title pages, etc.)
     if (paragraphAlignment == CssTextAlign::None) {
       blockStyle.alignment = blockStyle.textAlignDefined ? cssStyle.textAlign : CssTextAlign::Justify;
+    } else if (blockStyle.textAlignDefined && cssStyle.textAlign == CssTextAlign::Center) {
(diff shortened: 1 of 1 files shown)


Cache8063:

PR #1352, opened 2026-03-07
fix: correct QR code version capacity thresholds (#1309)

### LINKED ISSUE #1309: Page as QR code - issue with decoding
### Affected Version

build from master 88594077aa8d0ddc842f9244dff9bcc20ab37a1f

### Describe the Bug

In books written in Polish (and I bet any other languages with special characters like ąśćłżź) with font size set to small QR code is generated but most QR code readers have problems with decoding it. Some of them display error, some 'content 0', others some dummy values like few random digits.
The problem appears only with pages full of text, when there is fewer text on the page it's encoded and decoded correctly with special characters.

One of the online decoder I found was able to read the content but with wrong decoding of special charachters (ex. Nie musimy reagowaÄ bojaÅºliwie i trzymaÄ jÄzyka za zÄbami, tylko dlatego, Å¼e po drugiej stronie stojÄ pieniÄdze i wpÅywy)

There is no problem with same font size setting and string length for texts in English.

Here's one of the problematic QR code:

Image

### Steps to Reproduce

1. Have a text in language with special charachters (ex. Polish)
2. Set font size to small and find a page full of text
3. Generate QR code
4. Try to read it with any QR code scanner
5. QR code can't be decoded

### Expected Behavior

Text is correctly decoded

Cache8063:
## Summary

- Fix incorrect QR version-to-capacity thresholds that caused corrupt/undecodable QR codes
- Add explicit overflow guard with user-visible error when text exceeds QR v40 maximum capacity

Fixes #1309

## Root cause

The version selection thresholds in `QrUtils.cpp` assumed higher capacities than the actual byte-mode limits at ECC_LOW. For example, version 4 was used for up to 114 bytes, but it can only hold ~78 bytes. The upstream `ricmoo/QRCode` library has a known missing bounds check (`// @TODO: Return error if data is too big`), so oversized data silently corrupts the QR structure.

## Changes

| Before | After | Actual capacity |
|--------|-------|----------------|
| v4: 114B | v4: 78B | 78 bytes |
| v10: 395B | v10: 271B | 271 bytes |
| v20: 1066B | v20: 858B | 858 bytes |
| v30: 2110B | v30: 1732B | 1732 bytes |
| v40: unlimited | v40: 2953B max | 2953 bytes |

Also shows "Text too large for QR code" error instead of a corrupt QR when the payload exceeds the maximum capacity.

## Test plan

- [ ] Small page text → QR code scans correctly
- [ ] Medium page text → QR version scales up, still scans
- [ ] Very large page (dense small font) → shows error message instead of corrupt QR
- [ ] Polish/UTF-8 text → QR encodes and scans correctly (multi-byte chars use more capacity)

### DIFF
--- a/src/util/QrUtils.cpp
+++ b/src/util/QrUtils.cpp
@@ -6,18 +6,24 @@
 #include <memory>
 
 #include "Logging.h"
+#include "fontIds.h"
 
 void QrUtils::drawQrCode(const GfxRenderer& renderer, const Rect& bounds, const std::string& textPayload) {
-  // Dynamically calculate the QR code version based on text length
-  // Version 4 holds ~114 bytes, Version 10 ~395, Version 20 ~1066, up to 40
-  // qrcode.h max version is 40.
-  // Formula: approx version = size / 26 + 1 (very rough estimate, better to find best fit)
+  // Dynamically calculate the QR code version based on text length.
+  // Capacities are for ECC_LOW byte mode, with overhead for mode indicator and length field.
   const size_t len = textPayload.length();
+
+  if (len > 2953) {
+    LOG_ERR("QR", "Text payload (%d bytes) exceeds QR v40 max capacity", (int)len);
+    renderer.drawCenteredText(UI_12_FONT_ID, bounds.y + bounds.height / 2, "Text too large for QR code", true);
+    return;
+  }
+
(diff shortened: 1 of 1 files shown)


Cache8063:

PR #1354, opened 2026-03-07
fix: back button in settings returns to tab bar first

### LINKED ISSUE #797: [Feature Request] Back Button in Settings Should Return to Tab Menu
Hi

1.0 is looking awesome so far! Love the changes. One thing that would be nice is a faster way to return to the tabbed categories across the top of settings. If you've scrolled down to a lower setting, you have to click up several times to return. Currently the back button takes you out of settings, but I think it would make sense for it to take jump focus to the tabs. Then back one more time would take you out of settings. (or long hold)

Image

Cache8063:
## Summary
- Pressing Back while browsing settings within a category now jumps focus to the category tab bar
- Pressing Back again from the tab bar exits settings to home
- Previously, Back always exited directly to home regardless of scroll position

Closes #797

## Test plan
- [ ] Scroll deep into a settings category → press Back → tab bar is focused
- [ ] Press Back again from tab bar → exits to home screen
- [ ] Use category switching (continuous hold) → still works as before
- [ ] Settings are saved on exit (not on tab-bar jump)

### DIFF
--- a/src/activities/settings/SettingsActivity.cpp
+++ b/src/activities/settings/SettingsActivity.cpp
@@ -89,8 +89,13 @@ void SettingsActivity::loop() {
   }
 
   if (mappedInput.wasPressed(MappedInputManager::Button::Back)) {
-    SETTINGS.saveToFile();
-    onGoHome();
+    if (selectedSettingIndex > 0) {
+      selectedSettingIndex = 0;
+      requestUpdate();
+    } else {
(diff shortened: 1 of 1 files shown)


znelson (approved):

Cache8063:

znelson:
I've confirmed on device, works as expected and seems more intuitive to me than the existing behavior.

PR #1355, opened 2026-03-07
fix: hide system directories (sleep, XTCache) in file browser (#350)

### LINKED ISSUE #350: Hide "sleep" folder or any directory without valid files (.epub / .xtc)
### Affected Version

v0.13.1

### Describe the Bug

When browsing the contents of the memory card, directories that are not useful for the eBook reader are displayed. For example, the sleep folder should not be visible in the file explorer because it does not contain any .epub / .xtc files.
The same applies to any other directories that do not contain valid eBook files (.epub / .xtc). These folders have no functional purpose for the user and should be hidden to improve navigation and usability.

### Steps to Reproduce

1. Copy several directories to the memory card (including folders without .epub files, e.g., sleep).
2. Insert the memory card into the device.
3. Open the file explorer on the device.
4. Notice that directories appear which, when accessed, show a message indicating there are no .epub/.xtc files inside.

### Expected Behavior

Only directories containing valid eBook files (.epub / .xtc) should be displayed in the file explorer

```shell

```

Cache8063:
## Summary
- Hide `sleep/` and `XTCache/` directories from the file browser
- These contain sleep screen images and system cache, not books
- Matches the filtering already applied in the WebDAV handler
- Dot-prefixed versions (`.sleep`, `.crosspoint`) were already hidden

Closes #350

## Test plan
- [ ] `sleep/` folder no longer appears in file browser
- [ ] `XTCache/` folder no longer appears in file browser
- [ ] Normal book folders still appear
- [ ] Dot-prefixed folders still hidden (`.crosspoint`, `.sleep`)
- [ ] `System Volume Information` still hidden

### DIFF
--- a/src/activities/home/FileBrowserActivity.cpp
+++ b/src/activities/home/FileBrowserActivity.cpp
@@ -83,8 +83,8 @@ void FileBrowserActivity::loadFiles() {
   char name[500];
   for (auto file = root.openNextFile(); file; file = root.openNextFile()) {
     file.getName(name, sizeof(name));
-    if (name[0] == '.' || strcmp(name, "System Volume Information") == 0 ||
-        strcmp(name, "sleep") == 0 || strcmp(name, "XTCache") == 0) {
+    if (name[0] == '.' || strcmp(name, "System Volume Information") == 0 || strcmp(name, "sleep") == 0 ||
+        strcmp(name, "XTCache") == 0) {
       file.close();
       continue;
     }


Uri-Tauber:
This was addressed in https://github.com/crosspoint-reader/crosspoint-reader/pull/948.

I think it’s quite possible that a user might have a books folder named `sleep`. Using `.sleep` for custom images was intended to address this while ensuring we don’t accidentally hide a books folder.

You may also want to take a look at #1296.

Cache8063:
Closing in favor of #948. The maintainer correctly pointed out that  could be a valid user books folder name — good call. The  hiding is already covered by #948 so this PR is redundant. Thanks for the review!

PR #1357, opened 2026-03-08
fix: Fix bootloop logging crash

jpirnay:
## Summary

* **What is the goal of this PR?** On a cold boot (or after a crash that corrupts RTC RAM), logHead contains garbage. Then addToLogRingBuffer does: ``strncpy(logMessages[logHead], message, MAX_ENTRY_LEN - 1); `` With garbage logHead, this computes a completely invalid address. The % MAX_LOG_LINES guard on line 16 only runs after the bad store, which is too late. The fix is to clamp logHead before use.

## Additional Context

  specific areas to focus on).

---

### AI Usage

Did you use AI tools to help write this code? _**NO**_ (did use claude for the magic hash value)

### DIFF
--- a/lib/Logging/Logging.cpp
+++ b/lib/Logging/Logging.cpp
@@ -14,10 +14,7 @@ RTC_NOINIT_ATTR size_t logHead = 0;
 // value is only set by clearLastLogs(), so its absence means the buffer was
 // never properly initialized.
 RTC_NOINIT_ATTR uint32_t rtcLogMagic;
-static constexpr uint32_t fnv1a32(const char* s, uint32_t h = 2166136261u) {
-  return *s ? fnv1a32(s + 1, (h ^ static_cast<uint32_t>(*s)) * 16777619u) : h;
-}
-static constexpr uint32_t LOG_RTC_MAGIC = fnv1a32("crosspoint-reader");
+static constexpr uint32_t LOG_RTC_MAGIC = 0xDEADBEEF;
 
 void addToLogRingBuffer(const char* message) {
   // Add the message to the ring buffer, overwriting old messages if necessary.


ngxson:
Does it make more sense to refine the logic inside HalSystem::begin instead? It made for this exact case, but I think it's incomplete:

https://github.com/crosspoint-reader/crosspoint-reader/blob/170cc257744f50006ed6bfae8a0d5446f4ef1b07/lib/hal/HalSystem.cpp#L73-L80

jpirnay:
> Does it make more sense to refine the logic inside HalSystem::begin instead? It made for this exact case, but I think it's incomplete:
> 
> https://github.com/crosspoint-reader/crosspoint-reader/blob/170cc257744f50006ed6bfae8a0d5446f4ef1b07/lib/hal/HalSystem.cpp#L73-L80

That would not help for static-init windows - things that happen before setup calls begin() - so I think it needs to stay here.

Still I think your comment holds merit, so I added a helper to be used by getPanicInfo to prevent garbage in crash reports.

jpirnay:
Commit 18b36ef ("feat: dump crash report to sdcard") introduced two changes that combined to cause a bootloop on cold power-on:

- logHead and logMessages were changed to RTC_NOINIT_ATTR — meaning they are not zero-initialized on cold boot (only preserved across warm resets)
- The if (!logSerial) return early-exit guard in logPrintf was removed, since logging was now always desired

This meant that for the first time, logPrintf calls from static constructors (e.g. UITheme / LyraTheme constructors that run before setup()) would reach addToLogRingBuffer with a garbage logHead value. On cold boot, logHead could be any arbitrary value — if it was >= MAX_LOG_LINES, the write logMessages[logHead][...] would store to an out-of-bounds address, causing a store access fault and immediate reset. The device would loop indefinitely since each cold boot repeats the same path.

lib/hal/HalSystem.cpp
```diff
@@ -76,6 +76,10 @@ void begin() {
   // `clearPanic()` to clear it after dumping
   if (!isRebootFromPanic()) {
     clearPanic();
+  } else {
+    // Panic reboot: preserve logs and panic info, but clamp logHead in case the
+    // panic occurred before begin() ever ran (e.g. in a static constructor).
+    sanitizeLogHead();
```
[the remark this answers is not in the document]
jpirnay: Valid, fixed

lib/Logging/Logging.cpp
```diff
@@ -67,12 +73,21 @@ std::string getLastLogs() {
...
   return output;
 }
 
+bool sanitizeLogHead() {
+  if (logHead >= MAX_LOG_LINES) {
+    logHead = 0;
+    return true;
+  }
+  return false;
+}
```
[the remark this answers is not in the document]
jpirnay: Please check the last commit for a possible fix

ngxson:
> That would not help for static-init windows - things that happen before setup calls begin() - so I think it needs to stay here.

But we never call any `LOG_*` before `begin()`, right?

Otherwise, if we want to make sure it's running before everything, we can also make `HalSystem` a constructor that can be initialized statically, much like singleton `GfxRenderer` or `HalStorage`

Checking it on each `addToLogRingBuffer` is a bit overkill I think

lib/Logging/Logging.cpp
```diff
@@ -8,9 +8,26 @@
...
+// Magic word written alongside logHead to detect uninitialized RTC memory.
+// RTC_NOINIT_ATTR is not zeroed on cold boot, so logHead may appear in-range
+// (0..MAX_LOG_LINES-1) by chance even though logMessages is garbage. The magic
+// value is only set by clearLastLogs(), so its absence means the buffer was
+// never properly initialized.
+RTC_NOINIT_ATTR uint32_t rtcLogMagic;
+static constexpr uint32_t fnv1a32(const char* s, uint32_t h = 2166136261u) {
+  return *s ? fnv1a32(s + 1, (h ^ static_cast<uint32_t>(*s)) * 16777619u) : h;
+}
+static constexpr uint32_t LOG_RTC_MAGIC = fnv1a32("crosspoint-reader");
```
ngxson, line 18: Having a magic number of ok, but IMO this is a bit overkill

Something like this should be enough:

```suggestion
RTC_NOINIT_ATTR uint32_t rtcLogMagic;
static constexpr uint32_t LOG_RTC_MAGIC = 0xDEADBEEF;
```

Mathematically say, if `RTC_NOINIT` is uniformly random, then any values will get the same `1 / (2^32)` chance of collision, no matter if the value comes from a hash or was manually picked (by collision, I mean the chance of the randomly-init value matches the magic)

ngxson:
In junction with what you're adding here, I would suggest removing the `HalSystem::begin()` if it's not actually effective. Instead, move it to a constructor so that it will be invoked before the main `setup()`: https://github.com/ngxson/crosspoint-reader/commit/8f9f88261e7ffe93e8f739b273961e73dccdd074

lib/Logging/Logging.cpp
jpirnay: Any value would be fine, agree, was just an attempt to do someting remotely funny ;-) And as this is done on compile time, why not?

ngxson: I mean, the code is to be read by other contributors and a function like this may mislead someone to think that we actually need the FNV hash for a specific reason.

I have no problems with adding funny thing to the code, but as long as it's kept the code easy to understand, then I'm fine. You may want to look at this list too: https://nedbatchelder.com/text/hexwords

jpirnay:
> In junction with what you're adding here, I would suggest removing the `HalSystem::begin()` if it's not actually effective. Instead, move it to a constructor so that it will be invoked before the main `setup()`: [ngxson@8f9f882](https://github.com/ngxson/crosspoint-reader/commit/8f9f88261e7ffe93e8f739b273961e73dccdd074)

I assume you will raise that PR?

ngxson:
> I assume you will raise that PR?

Yes I will do that later

PR #1361, opened 2026-03-08
fix: Fixes for unitialized structures

jpirnay:
## Summary

* **What is the goal of this PR?** Fixes for 3 edge cases

## Additional Context

* addLineToPage null page crash — When the first text block in a chapter exceeded 750 words, the early-flush path in characterData called addLineToPage before any page had been created. Unlike makePages(), addLineToPage had no guard for a null currentPage, causing a null pointer dereference. Fixed by adding the same if (!currentPage) initialization guard that makePages() already had.

* PageLine::deserialize propagates null TextBlock — TextBlock::deserialize returns nullptr when it detects a corrupt word count (wc > 10000), but PageLine::deserialize constructed a PageLine with the null block anyway. When that page was later rendered or re-serialized, block->render() / block->serialize() crashed. Page::deserialize also pushed the result without checking. Fixed by returning nullptr from PageLine::deserialize on failure and checking the result in Page::deserialize before pushing.

* readString uninitialized len on truncated read — len was declared but not initialized before readPod. If the file read returned fewer than 4 bytes (truncated/corrupt section cache), len held stack garbage and s.resize(len) would attempt a massive heap allocation, causing OOM on the ESP32. Fixed by initializing len = 0 and adding a 4096-byte sanity cap that clears the string and returns early on corrupt data.

---

### AI Usage

Did you use AI tools to help write this code? _** Partially **_ Asked claude to look got unitialized structures

### DIFF
--- a/lib/Epub/Epub/Page.cpp
+++ b/lib/Epub/Epub/Page.cpp
@@ -22,6 +22,10 @@ std::unique_ptr<PageLine> PageLine::deserialize(FsFile& file) {
   serialization::readPod(file, yPos);
 
   auto tb = TextBlock::deserialize(file);
+  if (!tb) {
+    LOG_ERR("PGE", "Failed to deserialize TextBlock");
+    return nullptr;
+  }
   return std::unique_ptr<PageLine>(new PageLine(std::move(tb), xPos, yPos));
 }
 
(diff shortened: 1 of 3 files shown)


jpirnay:
Getting out of hand

PR #1363, opened 2026-03-09
fix: enable DESTRUCTOR_CLOSES_FILE flag

laird:
Enables the `DESTRUCTOR_CLOSES_FILE` flag so file handles are properly closed when they go out of scope, preventing resource leaks on early returns.


laird:
@coderabbitai review

znelson:
Same as #1075?

PR #1365, opened 2026-03-09
fix: restore WiFi after screenshot tour regardless of Danger Zone state

laird:
WiFi is now correctly restored after a screenshot tour completes, even if Danger Zone was not active when the tour started.

> ⚠️ **Depends on #1331** (PULSR theme) — this PR includes Antonio/PULSR font registration in main.cpp and should be reviewed/merged after #1331 lands.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -6,9 +6,87 @@
 #include <HalGPIO.h>
 #include <HalPowerManager.h>
 #include <HalStorage.h>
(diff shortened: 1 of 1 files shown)


laird:
Closing — Danger Zone is a claw-fork feature and isn't part of upstream CrossPoint. This fix will be maintained in the claw branch.

PR #1366, opened 2026-03-09
chore: remove miniz and modularise inflation logic

laird:
- Removes the bundled miniz copy and replaces it with a modular inflation implementation
- Makes `INFLATE_DICT_SIZE` constexpr
- Handles one-shot `readAtMost` usage cleanly


laird:
@coderabbitai review

znelson:
This is the same as #1073

PR #1367, opened 2026-03-09
feat: device name setting and /reader-feed.xml HTTP/1.1 proxy endpoint

laird:
- Adds a Device Name setting (default: `crosspoint`) used for mDNS hostname (`crosspoint.local`)
- Adds `/reader-feed.xml` proxy endpoint for HTTP/1.1 compatibility with RSS feed readers

### DIFF
--- a/feed-server/feed_server.py
+++ b/feed-server/feed_server.py
@@ -1,0 +1,243 @@
+#!/usr/bin/env python3
+"""
+CrossPoint content feed server.
+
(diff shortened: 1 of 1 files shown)


laird:
Closing — the device name setting and /reader-feed.xml endpoint are tied to Danger Zone / the claw-fork feed pipeline, which isn't upstream. Will maintain in the claw branch.

PR #1368, opened 2026-03-09
feat: Danger Zone — web-based device management and screenshot tour

laird:
Adds an optional Danger Zone mode (disabled by default) that enables advanced device management features via the web interface:

- **Screenshot tour**: captures all screens as BMPs and serves them via the web UI for remote device inspection
- **WiFi management**: connect/disconnect WiFi from the web interface  
- **Firmware flash via web UI**: upload firmware.bin directly from a browser
- **WiFi restore fix**: WiFi is correctly restored after a screenshot tour completes, regardless of whether DZ was already active

Danger Zone is off by default and must be explicitly enabled in Settings → System. When disabled, none of these endpoints are accessible.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -6,9 +6,87 @@
 #include <HalGPIO.h>
 #include <HalPowerManager.h>
 #include <HalStorage.h>
+#include <HalSystem.h>
 #include <I18n.h>
 #include <Logging.h>
 #include <SPI.h>
+#include <Update.h>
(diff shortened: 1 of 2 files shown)


itsthisjustin:
I really love the idea of doing firmware type things over the web server but due to the failed checks on this PR and potential overflow into other competing work, I'm going to close this. Love the ideas though @laird please stay engaged!

PR #1370, opened 2026-03-09
fix: load access fault crash

Uri-Tauber:
## Summary

Fixes a crash (load access fault) when opening EPUB chapters whose first text block exceeds 750 words.

## Changes

* **Crash fix (`addLineToPage`)**: Added a null guard for `currentPage`. If `makePages()` hasn't run yet (which can happen when the first block triggers the "text block too long" split path), the page is now created on demand.
* **Layout fix (`characterData`)**: The early-split path previously used `viewportWidth`, ignoring CSS margins and padding. It now computes `effectiveWidth` using `totalHorizontalInset()`, consistent with `makePages()`.

## Additional Context

* Crash signature: `MCAUSE=0x5` (load access fault), `A0=0x0` (`Page*` null), `MTVAL=0x4 / 0x8` (offsets into `Page::elements`).
* Confirmed in two user reports reported in #1328
* Tested on PR #1357 (not on `master`).

---

### AI Usage

### DIFF
--- a/lib/Epub/Epub/parsers/ChapterHtmlSlimParser.cpp
+++ b/lib/Epub/Epub/parsers/ChapterHtmlSlimParser.cpp
@@ -772,8 +772,12 @@ void XMLCALL ChapterHtmlSlimParser::characterData(void* userData, const XML_Char
   // Spotted when reading Intermezzo, there are some really long text blocks in there.
   if (self->currentTextBlock->size() > 750) {
     LOG_DBG("EHP", "Text block too long, splitting into multiple pages");
+    const int horizontalInset = self->currentTextBlock->getBlockStyle().totalHorizontalInset();
+    const uint16_t effectiveWidth = (horizontalInset < self->viewportWidth)
+                                        ? static_cast<uint16_t>(self->viewportWidth - horizontalInset)
+                                        : self->viewportWidth;
     self->currentTextBlock->layoutAndExtractLines(
-        self->renderer, self->fontId, self->viewportWidth,
+        self->renderer, self->fontId, effectiveWidth,
         [self](const std::shared_ptr<TextBlock>& textBlock) { self->addLineToPage(textBlock); }, false);
   }
 }
@@ -1019,6 +1023,11 @@ bool ChapterHtmlSlimParser::parseAndBuildPages() {
 
 void ChapterHtmlSlimParser::addLineToPage(std::shared_ptr<TextBlock> line) {
   const int lineHeight = renderer.getLineHeight(fontId) * lineCompression;
+
+  if (!currentPage) {
+    currentPage.reset(new Page());
+    currentPageNextY = 0;
+  }
 
   if (currentPageNextY + lineHeight > viewportHeight) {
     completePageFn(std::move(currentPage));


lib/Epub/Epub/parsers/ChapterHtmlSlimParser.cpp
```diff
@@ -1020,6 +1024,11 @@ bool ChapterHtmlSlimParser::parseAndBuildPages() {
 void ChapterHtmlSlimParser::addLineToPage(std::shared_ptr<TextBlock> line) {
   const int lineHeight = renderer.getLineHeight(fontId) * lineCompression;
 
+  if (!currentPage) {
+    currentPage.reset(new Page());
+    currentPageNextY = 0;
```
znelson, line 1029: Do we need similar logic to `makePages()`? 
```suggestion
    // Apply top spacing before the paragraph (stored in pixels)
    currentPageNextY = 0;
    const BlockStyle& blockStyle = currentTextBlock->getBlockStyle();
    if (blockStyle.marginTop > 0) {
      currentPageNextY += blockStyle.marginTop;
    }
    if (blockStyle.paddingTop > 0) {
      currentPageNextY += blockStyle.paddingTop;
    }
```

(A `BlockStyle::topInset` similar to `BlockStyle::leftInset` could simplify things.)

lib/Epub/Epub/parsers/ChapterHtmlSlimParser.cpp
```diff
@@ -772,8 +772,12 @@ void XMLCALL ChapterHtmlSlimParser::characterData(void* userData, const XML_Char
   // Spotted when reading Intermezzo, there are some really long text blocks in there.
   if (self->currentTextBlock->size() > 750) {
     LOG_DBG("EHP", "Text block too long, splitting into multiple pages");
+    const int horizontalInset = self->currentTextBlock->getBlockStyle().totalHorizontalInset();
+    const uint16_t effectiveWidth = (horizontalInset < self->viewportWidth)
+                                        ? static_cast<uint16_t>(self->viewportWidth - horizontalInset)
+                                        : self->viewportWidth;
```
znelson, line 778: I see this matches the other call to `layoutAndExtractLines` now, which is excellent. 

In a followup change, I wonder if we ought to bound to some minimum width. e.g.

```
constexpr uint16_t MIN_EFFECTIVE_LINE_WIDTH = 100;
inline uint16_t getEffectiveWidth(int viewportWidth, int horizontalInset) {
  const int remainingWidth = viewportWidth - horizontalInset;
  return remainingWidth < MIN_EFFECTIVE_LINE_WIDTH ? MIN_EFFECTIVE_LINE_WIDTH : static_cast<uint16_t>(remainingWidth);
}
```

znelson (approved):

lib/Epub/Epub/parsers/ChapterHtmlSlimParser.cpp
Uri-Tauber: The null guard in `addLineToPage()` is hit mid-block — while the XML parser is still accumulating words into `currentTextBlock`. After parsing continues and reaches the block's closing tag (e.g. `</p>`), `startNewTextBlock()` fires and calls `makePages()` for the same block's remaining lines. `makePages()` will always applies `marginTop/paddingTop` before it lays out its lines.

lib/Epub/Epub/parsers/ChapterHtmlSlimParser.cpp
Uri-Tauber: It a good defensive idea against malformed CSS file. Could definitely be another PR since this one handle crashing.

znelson: I agree, follow up PR is best because this is fixing a real blocking issue for certain books.

lib/Epub/Epub/parsers/ChapterHtmlSlimParser.cpp
znelson: Nice, good to know!

ngxson (approved):

PR #1373, opened 2026-03-09
feat: add 6 Unicode symbols to Bookerly and OpenDyslexic fonts

alemohamad:
## Summary

Add support for ~22 Unicode Miscellaneous Symbols (★☆ ☺☻☼ ♀♂ ♠♡♢♣♤♥♦♧ ♩♪♫♬♭♮♯) across all three reader font families~ 6 Unicode musical symbols (♩♪♫♬♭♮) to the Bookerly and OpenDyslexic font families. These previously rendered as the replacement character (�).

1. Added curated symbol intervals (~U+2605-2606, U+263A-263C, U+2640, U+2642, U+2660-2667,~ U+2669-266E) to `fontconvert.py`.
2. ~Added `NotoSansSymbols-Regular.ttf` as a font stack fallback for **Noto Sans** in `convert-builtin-fonts.sh`, since Noto Sans lacks these glyphs natively.~
3. Regenerated all 48 compressed font headers and font IDs.

Noto Sans is now unchanged. It does not contain these glyphs natively, so no fallback fonts are used.

## Additional Context

  specific areas to focus on).

1. Flash impact: ~+68 KB (+1.14%)~ ~+100 KB (+1.68%)~ +12.5 KB (+0.20%) on gh_release build.
2. ~The 22 symbols were chosen to match what Bookerly and OpenDyslexic support natively, avoiding unnecessary flash usage from the full 256-codepoint Misc Symbols block.~
3. Font ID changes will automatically invalidate the on-device section cache (.crosspoint/), triggering a one-time re-render on first book open.

---

### AI Usage


Uri-Tauber:
Just my opinion, but I don’t think it’s a good time to add 100 KB of flash for non-essential glyphs when we already have so little free flash remaining.
Maybe it would be better to save this for later, after one of the PRs for custom fonts is merged.

Also, do the missing glyphs actually affect the reading experience in practice? I would expect them to be quite rare, so even if they appear occasionally it probably won't be real problem.

alemohamad:
> Just my opinion, but I don’t think it’s a good time to add 100 KB of flash for non-essential glyphs when we already have so little free flash remaining. Maybe it would be better to save this for later, after one of the PRs for custom fonts is merged.
> 
> Also, do the missing glyphs actually affect the reading experience in practice? I would expect them to be quite rare, so even if they appear occasionally it probably won't be real problem.

@Uri-Tauber That's a fair point! I originally added this because while reading "Project Hail Mary", there are passages where music notes appear and they all show up as �. Same thing happens with Chinese characters. It was a little confusing. 😅 

That said, I totally understand the flash concern. If you feel the space is better reserved for future development, I'm happy to close this PR. No worries at all.

alemohamad:
@Uri-Tauber I've scaled this PR down significantly. Instead of all 22 symbols, it now only adds the 7 musical symbols (♩♪♫♬♭♮♯) to Bookerly and OpenDyslexic (which support them natively). The Noto Sans fallback fonts have been removed entirely.

Now flash impact dropped from +100 KB down to just +14 KB (+0.23%). This was the original motivation for the PR: music notes in books like Project Hail Mary rendering as �, so let me know if you believe it's worth adding this to the project or not.

Uri-Tauber:
Hi @alemohamad, I’ve also read *Project Hail Mary*, so I know the passage you’re referring to 🙂

Personally, I don’t think the absence of those glyphs would significantly affect the reading experience. That said, this is just my opinion—let’s see what other contributors think as well.

mcrosson:

PR #1378, opened 2026-03-09
feat: Add an option to the settings for the short power button click to toggle blocking the front buttons

makaGeorge:
## Summary

There are situations where you can accidentally press a front button. Some examples might be:

* maybe when reading/holding the device in your hand, while temporarily pausing reading.
* if you close the cover but don't want to turn the device off
* if you keep your reader in a notebook cover with your notebooks and want to switch to one of them to write while you're reading

This pull request adds an option to the settings screen for the Short Power Button Click to toggle a block to all front buttons while reading. You then can turn pages with the side buttons and when you're ready just toggle back the functionality with another click.

### Changes
 This adds the item to the settings menu and changes the logic of button detection for the different readers.

---

### AI Usage

I used ChatGPT to translate the text. This was meant as a temporary measure as I can't speak most languages included in the repo.


trilwu (approved):
dismissed

itsthisjustin:
Due to the checks failing and the high risk for merge conflicts I'm going to close this PR. I do agree in the principal behind this but I don't think we should be disabling buttons on a device like this. Thanks for the idea though and we should surface this to the community for voting so it's not just me deciding it's not in scope.

PR #1381, opened 2026-03-11
fix: use esp_http_client for KOSync to prevent TLS OOM on ESP32-C3

### LINKED ISSUE #581: KOReader Sync Not Properly

trilwu:
## Summary

* **What is the goal of this PR?** Fix KOReader sync failing on HTTPS servers due to TLS out-of-memory on ESP32-C3
  - Replace `WiFiClientSecure`/`HTTPClient` with `esp_http_client` (ESP-IDF native API) for all KOSync HTTP requests
  - Use 2KB TLS buffers instead of the default 16KB — KOSync payloads are tiny JSON (<1KB), so this is more than sufficient
  - Use `esp_crt_bundle_attach` for proper TLS certificate verification (replaces `setInsecure()`)
  - Strip trailing slashes from server URL to prevent double-slash in API paths (e.g. `https://server.com//users/auth`)
  - Add `lastHttpCode` static field for diagnostics
  - Add free heap logging to help debug memory issues

### Problem

The ESP32-C3 has ~46KB free heap after WiFi is initialized. `WiFiClientSecure` allocates 16KB for TLS RX + 16KB for TLS TX = 32KB just for the TLS buffers, leaving almost no room for the actual TLS handshake (which needs additional dynamic allocations). This causes KOReader sync to:

1. Fail silently with network errors on HTTPS servers (including the default `sync.koreader.rocks`)
2. Occasionally crash with heap exhaustion

### Solution

`esp_http_client` (the ESP-IDF native HTTP client) allows configuring `buffer_size` and `buffer_size_tx` independently. Setting both to 2KB (total 4KB) leaves plenty of heap for the TLS handshake while still being sufficient for KOSync's small JSON payloads.

This also fixes the `setInsecure()` anti-pattern — `esp_crt_bundle_attach` provides proper certificate verification using the ESP-IDF's built-in CA bundle, so credentials and reading history are no longer sent over unverified TLS connections.

### Files changed

| File | Change |
|------|--------|
| `lib/KOReaderSync/KOReaderSyncClient.cpp` | Replace WiFiClientSecure/HTTPClient with esp_http_client; add ResponseBuffer, base64 encoder, createClient helper |
| `lib/KOReaderSync/KOReaderSyncClient.h` | Add `lastHttpCode` static field |
| `lib/KOReaderSync/KOReaderCredentialStore.cpp` | Strip trailing slashes from base URL |

## Additional Context

Tested on a CrossPoint X4 (ESP32-C3 with 4MB flash). Before this change, KOSync auth to `sync.koreader.rocks` (HTTPS) would fail ~80% of the time. After: works reliably.

The `base64Encode` helper is needed because `esp_http_client` doesn't have a built-in `setAuthorization()` method like `HTTPClient` does. This is used for the HTTP Basic Auth header required by Calibre-Web-Automated KOSync servers.

Fixes #581

---

### AI Usage

Did you use AI tools to help write this code? _**PARTIALLY**_ AI assisted with the esp_http_client migration pattern and base64 encoder. The root cause analysis (TLS buffer OOM) and solution design were manual.

### DIFF
--- a/lib/KOReaderSync/KOReaderSyncClient.cpp
+++ b/lib/KOReaderSync/KOReaderSyncClient.cpp
@@ -2,6 +2,7 @@
 
 #include <ArduinoJson.h>
 #include <Logging.h>
+#include <base64.h>
 #include <esp_crt_bundle.h>
 #include <esp_http_client.h>
 
@@ -54,25 +55,6 @@ esp_err_t httpEventHandler(esp_http_client_event_t* evt) {
   return ESP_OK;
 }
 
-// Base64 encode for HTTP Basic Auth
-std::string base64Encode(const std::string& input) {
-  static const char table[] = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/";
-  std::string out;
-  out.reserve(((input.size() + 2) / 3) * 4);
-  int val = 0, valb = -6;
-  for (unsigned char c : input) {
-    val = (val << 8) + c;
-    valb += 8;
-    while (valb >= 0) {
-      out.push_back(table[(val >> valb) & 0x3F]);
-      valb -= 6;
-    }
-  }
-  if (valb > -6) out.push_back(table[((val << 8) >> (valb + 8)) & 0x3F]);
-  while (out.size() % 4) out.push_back('=');
-  return out;
-}
-
 // Create configured esp_http_client with small TLS buffers
 esp_http_client_handle_t createClient(const char* url, ResponseBuffer* buf,
                                       esp_http_client_method_t method = HTTP_METHOD_GET) {
@@ -90,20 +72,30 @@ esp_http_client_handle_t createClient(const char* url, ResponseBuffer* buf,
   if (!client) return nullptr;
 
   // KOSync auth headers
-  esp_http_client_set_header(client, "Accept", "application/vnd.koreader.v1+json");
-  esp_http_client_set_header(client, "x-auth-user", KOREADER_STORE.getUsername().c_str());
-  esp_http_client_set_header(client, "x-auth-key", KOREADER_STORE.getMd5Password().c_str());
+  if (esp_http_client_set_header(client, "Accept", "application/vnd.koreader.v1+json") != ESP_OK ||
+      esp_http_client_set_header(client, "x-auth-user", KOREADER_STORE.getUsername().c_str()) != ESP_OK ||
+      esp_http_client_set_header(client, "x-auth-key", KOREADER_STORE.getMd5Password().c_str()) != ESP_OK) {
+    LOG_ERR("KOSync", "Failed to set auth headers");
+    esp_http_client_cleanup(client);
+    return nullptr;
+  }
 
   // HTTP Basic Auth for Calibre-Web-Automated compatibility
   std::string credentials = KOREADER_STORE.getUsername() + ":" + KOREADER_STORE.getPassword();
-  std::string authHeader = "Basic " + base64Encode(credentials);
-  esp_http_client_set_header(client, "Authorization", authHeader.c_str());
+  String encoded = base64::encode(reinterpret_cast<const uint8_t*>(credentials.data()), credentials.size());
+  std::string authHeader = "Basic " + std::string(encoded.c_str());
+  if (esp_http_client_set_header(client, "Authorization", authHeader.c_str()) != ESP_OK) {
+    LOG_ERR("KOSync", "Failed to set Authorization header");
+    esp_http_client_cleanup(client);
+    return nullptr;
+  }
 
   return client;
 }
 }  // namespace
 
 KOReaderSyncClient::Error KOReaderSyncClient::authenticate() {
+  lastHttpCode = 0;
   if (!KOREADER_STORE.hasCredentials()) {
     LOG_DBG("KOSync", "No credentials configured");
     return NO_CREDENTIALS;
@@ -131,6 +123,7 @@ KOReaderSyncClient::Error KOReaderSyncClient::authenticate() {
 
 KOReaderSyncClient::Error KOReaderSyncClient::getProgress(const std::string& documentHash,
                                                           KOReaderProgress& outProgress) {
+  lastHttpCode = 0;
   if (!KOREADER_STORE.hasCredentials()) {
     LOG_DBG("KOSync", "No credentials configured");
     return NO_CREDENTIALS;
@@ -178,6 +171,7 @@ KOReaderSyncClient::Error KOReaderSyncClient::getProgress(const std::string& doc
 }
 
 KOReaderSyncClient::Error KOReaderSyncClient::updateProgress(const KOReaderProgress& progress) {
+  lastHttpCode = 0;
   if (!KOREADER_STORE.hasCredentials()) {
     LOG_DBG("KOSync", "No credentials configured");
     return NO_CREDENTIALS;
@@ -203,8 +197,12 @@ KOReaderSyncClient::Error KOReaderSyncClient::updateProgress(const KOReaderProgr
   esp_http_client_handle_t client = createClient(url.c_str(), &buf, HTTP_METHOD_PUT);
   if (!client) return NETWORK_ERROR;
 
-  esp_http_client_set_header(client, "Content-Type", "application/json");
-  esp_http_client_set_post_field(client, body.c_str(), body.length());
+  if (esp_http_client_set_header(client, "Content-Type", "application/json") != ESP_OK ||
+      esp_http_client_set_post_field(client, body.c_str(), body.length()) != ESP_OK) {
+    LOG_ERR("KOSync", "Failed to set request body");
+    esp_http_client_cleanup(client);
+    return NETWORK_ERROR;
+  }
 
   esp_err_t err = esp_http_client_perform(client);
   const int httpCode = esp_http_client_get_status_code(client);
--- a/lib/KOReaderSync/KOReaderSyncClient.cpp
+++ b/lib/KOReaderSync/KOReaderSyncClient.cpp
@@ -1,31 +1,104 @@
 #include "KOReaderSyncClient.h"
 
 #include <ArduinoJson.h>
-#include <HTTPClient.h>
 #include <Logging.h>
-#include <WiFi.h>
-#include <WiFiClientSecure.h>
+#include <esp_crt_bundle.h>
+#include <esp_http_client.h>
 
 #include <ctime>
 
 #include "KOReaderCredentialStore.h"
+
+int KOReaderSyncClient::lastHttpCode = 0;
 
 namespace {
 // Device identifier for CrossPoint reader
 constexpr char DEVICE_NAME[] = "CrossPoint";
 constexpr char DEVICE_ID[] = "crosspoint-reader";
 
-void addAuthHeaders(HTTPClient& http) {
-  http.addHeader("Accept", "application/vnd.koreader.v1+json");
-  http.addHeader("x-auth-user", KOREADER_STORE.getUsername().c_str());
-  http.addHeader("x-auth-key", KOREADER_STORE.getMd5Password().c_str());
+// Small TLS buffers to fit in ESP32-C3's limited heap (~46KB free after WiFi).
+// KOSync payloads are tiny JSON (<1KB), so 2KB buffers are sufficient.
+// Default 16KB buffers cause OOM during TLS handshake.
+constexpr int HTTP_BUF_SIZE = 2048;
 
-  // HTTP Basic Auth (RFC 7617) header. This is needed to support koreader sync server embedded in Calibre Web Automated
-  // (https://github.com/crocodilestick/Calibre-Web-Automated/blob/main/cps/progress_syncing/protocols/kosync.py)
-  http.setAuthorization(KOREADER_STORE.getUsername().c_str(), KOREADER_STORE.getPassword().c_str());
+// Response buffer for reading HTTP body
+struct ResponseBuffer {
+  char* data = nullptr;
+  int len = 0;
+  int capacity = 0;
+
+  ~ResponseBuffer() { free(data); }
+
+  bool ensure(int size) {
+    if (size <= capacity) return true;
+    char* newData = (char*)realloc(data, size);
+    if (!newData) return false;
+    data = newData;
+    capacity = size;
+    return true;
+  }
+};
+
+// HTTP event handler to collect response body
+esp_err_t httpEventHandler(esp_http_client_event_t* evt) {
+  auto* buf = static_cast<ResponseBuffer*>(evt->user_data);
+  if (evt->event_id == HTTP_EVENT_ON_DATA && buf) {
+    if (buf->ensure(buf->len + evt->data_len + 1)) {
+      memcpy(buf->data + buf->len, evt->data, evt->data_len);
+      buf->len += evt->data_len;
+      buf->data[buf->len] = '\0';
+    }
+  }
+  return ESP_OK;
 }
 
-bool isHttpsUrl(const std::string& url) { return url.rfind("https://", 0) == 0; }
+// Base64 encode for HTTP Basic Auth
+std::string base64Encode(const std::string& input) {
+  static const char table[] = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/";
+  std::string out;
+  out.reserve(((input.size() + 2) / 3) * 4);
+  int val = 0, valb = -6;
+  for (unsigned char c : input) {
+    val = (val << 8) + c;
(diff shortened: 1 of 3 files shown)


lib/KOReaderSync/KOReaderSyncClient.cpp
```diff
@@ -1,31 +1,106 @@
 #include "KOReaderSyncClient.h"
 
 #include <ArduinoJson.h>
-#include <HTTPClient.h>
 #include <Logging.h>
-#include <WiFi.h>
-#include <WiFiClientSecure.h>
+#include <esp_crt_bundle.h>
+#include <esp_http_client.h>
```
znelson, line 7: There's a [comment in OtaUpdater.cpp about this header](https://github.com/crosspoint-reader/crosspoint-reader/blob/master/src/network/OtaUpdater.cpp#L17-L24):

```c++
/*
 * When esp_crt_bundle.h included, it is pointing wrong header file
 * which is something under WifiClientSecure because of our framework based on arduno platform.
 * To manage this obstacle, don't include anything, just extern and it will point correct one.
 */
extern "C" {
extern esp_err_t esp_crt_bundle_attach(void* conf);
}
```

Do you know if that still applies? Maybe not relevant after we updated the platform in a610568f8c4f0c11f7dc7ce1142b397eb62678e9 from espressif32 6.12.0 (Arduino Core 2.0.17) to pioarduino 55.03.37 (Arduino Core 3.3.7, ESP-IDF 5.5.2).

cc @embedded4ever

znelson: For what it's worth, Claude's analysis: 

> The evidence is clear:
> 
> **In the old framework** (espressif32 6.12.0 / Arduino Core 2.0.17), the `WiFiClientSecure` library shipped its own `esp_crt_bundle.c` with a *different signature*:
> 
> ```c
> // WiFiClientSecure's version (Arduino Core 2.x) — note: missing size parameter
> void esp_crt_bundle_set(const uint8_t *x509_bundle)
> ```
> 
> vs. the ESP-IDF version:
> 
> ```c
> // ESP-IDF's version — correct signature
> esp_err_t esp_crt_bundle_set(const uint8_t *x509_bundle, size_t bundle_size);
> ```
> 
> When you did `#include <esp_crt_bundle.h>`, the build system would pick up WiFiClientSecure's version, giving you the wrong function declarations. That's why OtaUpdater used the `extern "C"` workaround.
> 
> **After a610568** (pioarduino 55.03.37 / Arduino Core 3.3.7 / ESP-IDF 5.5.2):
> 
> 1. `WiFiClientSecure` was renamed to `NetworkClientSecure`
> 2. There's no `esp_crt_bundle.h` anywhere inside `framework-arduinoespressif32` — my search confirmed zero hits
> 3. Both `NetworkClientSecure`'s source files and user code resolve `#include "esp_crt_bundle.h"` to the *same* ESP-IDF header in `framework-arduinoespressif32-libs/esp32c3/include/mbedtls/esp_crt_bundle/include/`
> 
> So the OtaUpdater comment and `extern "C"` workaround are stale. The direct `#include <esp_crt_bundle.h>` is actually the correct approach for the current framework, and it's OtaUpdater that should be cleaned up to match.
> 
> A follow-up cleanup opportunity: remove the `extern "C"` hack in OtaUpdater and replace it with a normal include, now that the header conflict is gone.

lib/KOReaderSync/KOReaderSyncClient.cpp
```diff
@@ -1,31 +1,106 @@
...
+
+  // KOSync auth headers
+  esp_http_client_set_header(client, "Accept", "application/vnd.koreader.v1+json");
+  esp_http_client_set_header(client, "x-auth-user", KOREADER_STORE.getUsername().c_str());
+  esp_http_client_set_header(client, "x-auth-key", KOREADER_STORE.getMd5Password().c_str());
+
+  // HTTP Basic Auth for Calibre-Web-Automated compatibility
+  std::string credentials = KOREADER_STORE.getUsername() + ":" + KOREADER_STORE.getPassword();
+  std::string authHeader = "Basic " + base64Encode(credentials);
+  esp_http_client_set_header(client, "Authorization", authHeader.c_str());
```
znelson: Not critical but I agree it would probably be good to have some error checking on these `esp_http_client_set_header` calls. You can see what OtaUpdater does [here](https://github.com/crosspoint-reader/crosspoint-reader/blob/master/src/network/OtaUpdater.cpp#L96-L101).

lib/KOReaderSync/KOReaderSyncClient.cpp
```diff
@@ -1,31 +1,106 @@
...
+    }
+  }
+  return ESP_OK;
+}
 
-  // HTTP Basic Auth (RFC 7617) header. This is needed to support koreader sync server embedded in Calibre Web Automated
-  // (https://github.com/crocodilestick/Calibre-Web-Automated/blob/main/cps/progress_syncing/protocols/kosync.py)
-  http.setAuthorization(KOREADER_STORE.getUsername().c_str(), KOREADER_STORE.getPassword().c_str());
+// Base64 encode for HTTP Basic Auth
+std::string base64Encode(const std::string& input) {
```
znelson: You should be able to use `#include <mbedtls/base64.h>` and `base64::encode()`. See [ObfuscationUtils.cpp for an example](https://github.com/crosspoint-reader/crosspoint-reader/blob/master/lib/Serialization/ObfuscationUtils.cpp#L45).

lib/KOReaderSync/KOReaderSyncClient.cpp
```diff
@@ -1,31 +1,106 @@
...
-#include <WiFi.h>
-#include <WiFiClientSecure.h>
+#include <esp_crt_bundle.h>
+#include <esp_http_client.h>
 
 #include <ctime>
 
 #include "KOReaderCredentialStore.h"
 
+int KOReaderSyncClient::lastHttpCode = 0;
```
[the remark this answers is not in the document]
znelson: Sounds like a reasonable suggestion to me.

znelson (dismissed):
A few possible tweaks, but looks reasonable to me. @mcrosson [mentioned](https://github.com/crosspoint-reader/crosspoint-reader/pull/1483#issuecomment-4142302835) it's been a big QOL improvement for them.

itsthisjustin:
So this library appears to not support TLS 1.3 which means KO Sync servers like kosync.ak-team.com:3042 fail to authenticate. I have a fix that I can push, but I don't want to railroad this PR with my own code without a thumbs up. 

The TLS handshake fails with `-0x7780` (`MBEDTLS_ERR_SSL_FATAL_ALERT_MESSAGE`)

- **`-D` build flag**: `CONFIG_MBEDTLS_SSL_PROTO_TLS1_3` is a Kconfig option, not a preprocessor define — has no effect on the pre-compiled library
- **`custom_sdkconfig`**: Triggers a full framework recompile that fails due to unrelated managed component dependencies (`esp_insights` certificate embedding, `__wrap_log_printf` symbol mismatches)
- **Library swap** (`libmbedtls_2.a`): Ships with TLS 1.3 source files but they're compiled as empty stubs — PSA crypto prerequisites were disabled at compile time

None of these work without rebuilding ESP-IDF from source, which is out of scope for the Arduino framework.

### Solution

A Cloudflare Worker at `kosync.crosspointreader.com` acts as a TLS version bridge. The firmware tries a direct connection first, and only falls back to the proxy if the TLS handshake fails — so servers that support TLS 1.2 are never affected.

**Flow:**
```
ESP32-C3 --[TLS 1.2]--> Cloudflare Worker --[TLS 1.3]--> KOSync server
```

### kosync.crosspointreader.com

```js
const TARGET_HEADER = "X-Target-Origin";

export default {
  async fetch(request) {
    const origin = request.headers.get("X-Target-Origin");
    if (!origin) {
      return new Response("Missing X-Target-Origin header", { status: 400 });
    }

    if (!origin.startsWith("https://")) {
      return new Response("X-Target-Origin must be HTTPS", { status: 400 });
    }

    const url = new URL(request.url);
    const target = `${origin.replace(/\/+$/, "")}${url.pathname}${url.search}`;

    // Forward all headers except proxy-specific and sensitive ones
    const headers = new Headers(request.headers);
    headers.delete("X-Target-Origin");
    headers.delete("Authorization"); // Strip Basic Auth (contains raw password)
    headers.set("Host", new URL(origin).host);

    const response = await fetch(target, {
      method: request.method,
      headers,
      body: request.body,
    });

    return new Response(response.body, {
      status: response.status,
      statusText: response.statusText,
      headers: response.headers,
    });
  },
};
```

### Firmware changes (`KOReaderSyncClient.cpp`)

Added a `performRequest()` helper that wraps all HTTP calls with automatic fallback:

```cpp
// TLS 1.3 bridge: ESP32-C3 only supports TLS 1.2, so servers requiring TLS 1.3
// are unreachable directly. On TLS handshake failure, requests are retried via
// this Cloudflare Worker which bridges TLS 1.2 -> 1.3.
constexpr char TLS_BRIDGE_URL[] = "https://kosync.crosspointreader.com";
```

When `createClient()` receives a `targetOrigin`, it rewrites the URL to go through the bridge and sets the `X-Target-Origin` header:

```cpp
esp_http_client_handle_t createClient(const char* url, ResponseBuffer* buf,
                                      esp_http_client_method_t method = HTTP_METHOD_GET,
                                      const char* targetOrigin = nullptr) {
  std::string proxyUrl;
  if (targetOrigin) {
    const char* path = strstr(url, "://");
    if (path) {
      path = strchr(path + 3, '/');
    }
    proxyUrl = std::string(TLS_BRIDGE_URL) + (path ? path : "/");
  }

  esp_http_client_config_t config = {};
  config.url = targetOrigin ? proxyUrl.c_str() : url;
  // ... same config as before ...

  if (targetOrigin) {
    esp_http_client_set_header(client, "X-Target-Origin", targetOrigin);
  }
  // ...
}
```

`performRequest()` tries direct first, retries via bridge on TLS failure:

```cpp
HttpResult performRequest(const char* url, esp_http_client_method_t method = HTTP_METHOD_GET,
                          const char* contentType = nullptr, const char* body = nullptr,
                          int bodyLen = 0) {
  HttpResult result = {};

  // First attempt: direct connection
  esp_http_client_handle_t client = createClient(url, &result.buf, method);
  // ...
  result.err = esp_http_client_perform(client);
  result.httpCode = esp_http_client_get_status_code(client);
  esp_http_client_cleanup(client);

  // If TLS handshake failed, retry via the TLS bridge
  if (result.err != ESP_OK && result.httpCode == 0) {
    std::string origin;
    const char* schemeEnd = strstr(url, "://");
    if (schemeEnd) {
      const char* pathStart = strchr(schemeEnd + 3, '/');
      origin.assign(url, pathStart ? (size_t)(pathStart - url) : strlen(url));
    }

    if (!origin.empty()) {
      LOG_DBG("KOSync", "TLS failed, retrying via bridge (heap: %u)", (unsigned)ESP.getFreeHeap());
      result.buf.len = 0;

      esp_http_client_handle_t proxy = createClient(url, &result.buf, method, origin.c_str());
      // ... set headers, perform, cleanup ...
    }
  }

  return result;
}
```
### Security

The Cloudflare Worker strips the `Authorization` header (which contains the base64-encoded raw password for Calibre-Web compatibility) before forwarding to the upstream server. Only the KOSync-specific `x-auth-user` and `x-auth-key` headers (username + MD5 hash) pass through the proxy. The raw password never transits Cloudflare.

All three API methods (`authenticate`, `getProgress`, `updateProgress`) now use `performRequest()` instead of managing `esp_http_client` directly.

@znelson @trilwu would love to get this in, just say the word

mcrosson:
Would there be a way to disable the cloudflare worker bridge access? I self-host my kosync instance and would very much not use the sync feature if it was routed through CloudFlare.

itsthisjustin:
> Would there be a way to disable the cloudflare worker bridge access? I self-host my kosync instance and would very much not use the sync feature if it was routed through CloudFlare.

It won't use it at all if your self hosted server is using TLS 1.2. If the connection works, it doesn't even attempt to connect to it. If your self hosted server requires TLS 1.3 it wont work anyway in the current code without this proxy

mcrosson:
> > Would there be a way to disable the cloudflare worker bridge access? I self-host my kosync instance and would very much not use the sync feature if it was routed through CloudFlare.
> 
> It won't use it at all if your self hosted server is using TLS 1.2. If the connection works, it doesn't even attempt to connect to it. If your self hosted server requires TLS 1.3 it wont work anyway in the current code without this proxy

I'd still want the option to disable the CloudFlare pass through. Even if my server were updated to use tls 1.3. The whole reason I self-host is to avoid CloudFlare infra maximally, even if that means breakage of tls for sync. I'm sure I'm not alone in this. 

I'd also be well beyond grumpy if I discovered my traffic was being routed through a CloudFlare tls proxy and I had no way to turn off that routing explicitly.

itsthisjustin:
> > > Would there be a way to disable the cloudflare worker bridge access? I self-host my kosync instance and would very much not use the sync feature if it was routed through CloudFlare.
> > 
> > 
> > It won't use it at all if your self hosted server is using TLS 1.2. If the connection works, it doesn't even attempt to connect to it. If your self hosted server requires TLS 1.3 it wont work anyway in the current code without this proxy
> 
> I'd still want the option to disable the CloudFlare pass through. Even if my server were updated to use tls 1.3. The whole reason I self-host is to avoid CloudFlare infra maximally, even if that means breakage of tls for sync. I'm sure I'm not alone in this.
> 
> I'd also be well beyond grumpy if I discovered my traffic was being routed through a CloudFlare tls proxy and I had no way to turn off that routing explicitly.

OK but then your kosync wouldn't work at all haha. It's not that it's just "oh we dont use tls 1.3", it's that kosync wont auth period if you updated that. You'd be free to do that, but it wouldnt work on xteink/crosspoint at all then. I can add a setting for it or some text for transparency, but if you turned it off, your kosync would no longer function

mcrosson:
> OK but then your kosync wouldn't work at all haha. It's not that it's just "oh we dont use tls 1.3", it's that kosync wont auth period if you updated that. You'd be free to do that, but it wouldnt work on xteink/crosspoint at all then. I can add a setting for it or some text for transparency, but if you turned it off, your kosync would no longer function

I'm aware. That doesn't change my desire to avoid CloudFlare proxies and to go without if that's my only option for avoiding CloudFlare.

A setting + warning would be great. That'll give everyone options and make it clear what's going on. There could also be a note in the KOSync docs of Crosspoint to explain the how/why/need for a proxy at times as well as the risk of disabling the use of a proxy.

itsthisjustin:
> > OK but then your kosync wouldn't work at all haha. It's not that it's just "oh we dont use tls 1.3", it's that kosync wont auth period if you updated that. You'd be free to do that, but it wouldnt work on xteink/crosspoint at all then. I can add a setting for it or some text for transparency, but if you turned it off, your kosync would no longer function
> 
> I'm aware. That doesn't change my desire to avoid CloudFlare proxies and to go without if that's my only option for avoiding CloudFlare.
> 
> A setting + warning would be great. That'll give everyone options and make it clear what's going on. There could also be a note in the KOSync docs of Crosspoint to explain the how/why/need for a proxy at times as well as the risk of disabling the use of a proxy.

fwiw I'd also be happy to provide the code in case you wanted to run this on your own. Maybe in the future we just have an input for TLS Proxy where you can accept the default or put your own.

mcrosson:
> fwiw I'd also be happy to provide the code in case you wanted to run this on your own. Maybe in the future we just have an input for TLS Proxy where you can accept the default or put your own

I appreciate the offer but I should be ok without the proxy code. It may be wise to add it somewhere for folk to audit but I'll leave that to others to decide.

My only concern is being forced to use the proxy. So long as it can be disabled, I'm happy.

I would agree, long-term, it might be nice to add a tunable proxy but I have a feeling that's more effort than it's worth currently.

itsthisjustin (dismissed):

itsthisjustin:
@Uri-Tauber can you approve this again? I made @znelson's changes but now it needs 1 more approval again

znelson (approved):

mcrosson (approved):
LGTM

PR #1404, opened 2026-03-15
fix: Don't break English contractions at apostrophes

### LINKED ISSUE #1403: Lines getting broken after apostrophes in English (after #1318)
### Affected Version

main branch

### Describe the Bug

English language contractions are getting broken mid-word after an apostrophe. Screenshots below.

Image

Source is
```
<p class="calibre10">Within a couple of months, we now have built technology, we have better data to analyze what’s happening with delays and all the things that are happening, put cameras up, use AI, I use AI to figure out what’s going on, far better data than any airline in the world. And because of that, it also made our culture infinitely better. We took that off of the plate and told employees, “We want you to just do the right thing. You’re a team, work as a team. It doesn’t matter, there’s this delay, it doesn’t matter whose fault it is, you’re not going to get blamed”.</p>
```

Image

Source is
```
<p class="calibre10"><strong class="calibre15">This is perhaps an odd venue for you given this is a site about technology, and you are an odd subject for me given you’re an airline executive, but this is actually an interview I’ve been looking forward to for a very long time for reasons that I think will become clear in our conversation. I am, however, like many of my readers, a bit of an aviation geek, and for what it’s worth, I’ve been a <a href="https://www.united.com/en/us/fly/mileageplus/premier.html">Premier 1K member</a> for many years.</strong></p>
```

### Steps to Reproduce

1. Open ebook
2. Read

### Expected Behavior

Keep full word on one line (or break with hyphen at allowable points)

```shell

```

BillionClaw:
## Summary

Fixes #1403 - English contractions like "what's", "you're" are no longer incorrectly broken at the apostrophe.

## Problem

PR #1318 added support for hyphenating words with apostrophes (for Italian/French contractions like "all'improvviso"). However, this introduced a regression where English contractions were incorrectly broken at the apostrophe position.

## Solution

Added detection for common English contraction suffixes ('s, 're, 've, 'll, 'd, 't, 'm) and prevents line breaks at those positions while still allowing hyphenation within the segments for Italian/French-style contractions.

## Changes

- Added `isEnglishContractionSuffix()` helper function
- Modified `appendApostropheContractionBreaks()` to skip breaks for English contractions

## Testing

- Verified English contractions (what's, you're, I've, don't, etc.) no longer break at apostrophes
- Verified Italian words (all'improvviso) still get hyphenation points
- Verified existing hyphenation evaluation tests still pass (english: 99.1%, italian: 99.9%)

---

**AI Disclosure:** This fix was implemented by an AI agent (ClawOSS). The issue analysis, solution design, and code changes were generated autonomously.

### DIFF
--- a/lib/Epub/Epub/hyphenation/Hyphenator.cpp
+++ b/lib/Epub/Epub/hyphenation/Hyphenator.cpp
@@ -99,6 +99,31 @@ void appendSegmentPatternBreaks(const std::vector<CodepointInfo>& cps, const Lan
   }
 }
 
+// Checks if the codepoints after the apostrophe form a common English contraction suffix.
+// These should NOT be line-broken (e.g., "you're", "what's", "don't").
+// Common suffixes: 's, 're, 've, 'll, 'd, 't, 'm
+bool isEnglishContractionSuffix(const std::vector<CodepointInfo>& cps, size_t apostropheIdx) {
+  if (apostropheIdx + 1 >= cps.size()) {
+    return false;
+  }
+
+  // Collect the alphabetic characters after the apostrophe
+  std::string suffix;
+  for (size_t j = apostropheIdx + 1; j < cps.size() && !isSegmentSeparator(cps[j].value); ++j) {
+    if (isAlphabetic(cps[j].value)) {
+      // Convert to lowercase for comparison
+      const uint32_t lower = (cps[j].value >= 'A' && cps[j].value <= 'Z') ? cps[j].value - 'A' + 'a' : cps[j].value;
+      if (lower <= 0x7F) {
+        suffix.push_back(static_cast<char>(lower));
+      }
+    }
+  }
+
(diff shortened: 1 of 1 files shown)


znelson:
Didn't see this until I had also made #1405. Do you think there are cases where we want to break on a two character tail that's not a common English contraction, or should we just bump the tail threshold to 3 characters?

jpirnay:
> Didn't see this until I had also made #1405. Do you think there are cases where we want to break on a two character tail that's not a common English contraction, or should we just bump the tail threshold to 3 characters?

The logic in hyphenator.cpp appendApostropheContractionBreaks will allow breaking if the left or right part is >= the given thresholds (left: 3, right: 2) -> so the right threshold is too generous (>=2 will allow "you're" to be broken at the apostrophe).

So an adjustment of the values should be good enough, IMHO.

lukestein:
Thanks, but closing in favor of #1405

PR #1405, opened 2026-03-16
fix: Prevent line breaks on common English contractions

### LINKED ISSUE #1403: Lines getting broken after apostrophes in English (after #1318)

znelson:
## Summary

Follow up to 3dabd30287fd164282fccca3f6a04aa736e05d4a: prevent line breaks in common English contractions with two tail characters. Prior to this change, I saw a line break inserted in "they've" as "they'" | "ve".

Fixes #1403

---

### AI Usage

### DIFF
--- a/lib/Epub/Epub/hyphenation/Hyphenator.cpp
+++ b/lib/Epub/Epub/hyphenation/Hyphenator.cpp
@@ -102,7 +102,7 @@ void appendSegmentPatternBreaks(const std::vector<CodepointInfo>& cps, const Lan
 void appendApostropheContractionBreaks(const std::vector<CodepointInfo>& cps,
                                        std::vector<Hyphenator::BreakInfo>& outBreaks) {
   constexpr size_t kMinLeftSegmentLen = 3;
-  constexpr size_t kMinRightSegmentLen = 2;
+  constexpr size_t kMinRightSegmentLen = 3;
   size_t segmentStart = 0;
 
   for (size_t i = 0; i < cps.size(); ++i) {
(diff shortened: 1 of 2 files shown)


znelson:
Tagging @jpirnay for feedback

jpirnay:
> Tagging @jpirnay for feedback


The logic in hyphenator.cpp appendApostropheContractionBreaks will allow breaking if the left or right part is >= the given thresholds (left: 3, right: 2) -> so the right threshold is too generous (>=2 will allow "you're" to be broken at the apostrophe).

So an adjustment of the values should be good enough, IMHO.

lukestein:
I'll test this, but sounds like the logic should be good for English. I know the other (AI-generated) PR uses a list of suffixes, but if the length-based approach still works ok for other languages then it's probably better?!

The minimum prefix length here should keep us ok with "o'clock" I think.

PR #1410, opened 2026-03-16
feat: Display available space on SDCard

jpirnay:
## Summary

This requires https://github.com/open-x4-epaper/community-sdk/pull/30 to be merged!

* **What is the goal of this PR?** Display available space on SDCard in webserver

## Additional Context
* SDCardManager.h/.cpp: totalBytes() and usedBytes() using sd.vol() cluster/sector math from SdFat
* HalStorage.h/.cpp: sdTotalBytes() / sdUsedBytes() wrappers with the storage mutex
* CrossPointWebServer.cpp: sdTotal and sdUsed added to the /api/status JSON
* HomePage.html: New "SD Card" row showing 1.23 GB / 16.00 GB format
* New SystemInformation screen:

grafik

### AI Usage

Did you use AI tools to help write this code? _**<  NO >**_


zgredex:
IMG_4714
Congrats on being faster! I was just trying to nail the icon and margins.

jpirnay:
This needs more testing, it disconnected my SD card after several consecutive uses...

zgredex:
@jpirnay I believe I understand why it happens. You called freeClusterCount() on every loop iteration to compute free space. This function performs a full FAT walk to count free clusters and it holds the SD mutex for the entire duration.
                                                                                                                                                                                                                     The problem gets worse and worse because stuff like OPDS or Calibre) runs concurrently and also needs the SD mutex for file I/O. With the FAT walk running every loop tick, the mutex was almost never available. The web server would time out waiting for it, which caused the driver to enter an error state and drop the card entirely. 
                                                                                                                                                                                                                     
My solution:                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                                                                                                                                                                                         Single background task, not a polling loop. FreeRTOS task (sdFreeUpdateTask) runs the FAT walk once every 60 seconds and stores the result in volatile uint32_t sdFreeMB. The UI thread never waits for the walk — it reads the cached value instantly. The background task competes for the mutex like any normal SD operation, so there is no starvation and no crashes.
                                                                                                                                                                                                                                                                                                                                                                                                                         [You can take a look here](https://github.com/zgredex/crosspoint-reader/tree/feature/sd-space-indicator-clean)
                                                                                                                                                                                                                                                                                                                                                                                                                         I left you a PR that should fix it, hopefully for good.
                                                                                                                                                                                                                                                                                                                                                                                                                         
![IMG_4748](https://github.com/user-attachments/assets/fe1fc99b-a23e-4d80-9568-68298b352c1d)

jpirnay:
#1466 is an extended PR that should be considered instead

PR #1411, opened 2026-03-16
feat: show full path bar in file browser

zgredex:
obraz

## Summary

Adds a full path display at the bottom of the file browser with a separator line matching the header style. Path uses the small font, left-truncates to always show the deepest folder when path is too long. Toggleable via Settings > System > Show Full Path (default on).

### AI Usage


lib/I18n/translations/english.yaml
```diff
@@ -341,3 +341,4 @@ STR_LINK: "[link]"
 STR_SCREENSHOT_BUTTON: "Take screenshot"
 STR_AUTO_TURN_ENABLED: "Auto Turn Enabled: "
 STR_AUTO_TURN_PAGES_PER_MIN: "Auto Turn (Pages Per Minute)"
+STR_SHOW_FULL_PATH: "Show Full Path"
```
znelson: This seems generally useful. I'd vote to just make it the default behavior and remove the preference.

src/activities/home/FileBrowserActivity.cpp
```diff
@@ -265,6 +269,33 @@ void FileBrowserActivity::render(RenderLock&&) {
...
+      const char ellipsis[] = "\xe2\x80\xa6";  // UTF-8 ellipsis (…)
+      const int ellipsisWidth = renderer.getTextWidth(SMALL_FONT_ID, ellipsis);
+      const int available = pathMaxWidth - ellipsisWidth;
+      // Walk forward from the end until the suffix fits
+      const char* p = pathStr + strlen(pathStr);
+      while (p > pathStr) {
+        if (renderer.getTextWidth(SMALL_FONT_ID, p) <= available) break;
+        --p;
+        while (p > pathStr && (static_cast<unsigned char>(*p) & 0xC0) == 0x80) --p;
+      }
```
znelson: Can you check that this is truncating correctly? It looks like if any truncation is necessary, it will break immediately, because `p` initially points to the null terminator, and truncate to just an ellipsis.

I think it should be like:
```suggestion
      const char* p = pathStr;
      while (*p) {
        if (renderer.getTextWidth(SMALL_FONT_ID, p) <= available) break;
        ++p;
        while (*p && (static_cast<unsigned char>(*p) & 0xC0) == 0x80) ++p;
      }
```

src/activities/home/FileBrowserActivity.cpp
znelson: Do we need to change the `pageItems` calculation here when some of the page is used for the path bar? Can you test with a directory with many items that requires scrolling to get to later items?

zgredex: I think we do. Without it we would overshoot by one position

src/components/UITheme.h
```diff
@@ -19,7 +19,7 @@ class UITheme {
   void reload();
   void setTheme(CrossPointSettings::UI_THEME type);
   static int getNumberOfItemsPerPage(const GfxRenderer& renderer, bool hasHeader, bool hasTabBar, bool hasButtonHints,
-                                     bool hasSubtitle);
+                                     bool hasSubtitle, int extraReserved = 0);
```
znelson: NBD but maybe want to distinguish that this is reserved _height_:
```suggestion
                                     bool hasSubtitle, int extraReservedHeight = 0);
```

src/SettingsList.h
```diff
@@ -75,6 +75,7 @@ inline const std::vector<SettingInfo>& getSettingsList() {
                         StrId::STR_CAT_CONTROLS),
 
       // --- System ---
+      // STR_NONE_OPT: serialized by JsonSettingsIO but UI ordering is handled in SettingsActivity
```
znelson: One more line to revert here

znelson (dismissed):

src/components/UITheme.cpp
```diff
@@ -61,7 +61,7 @@ int UITheme::getNumberOfItemsPerPage(const GfxRenderer& renderer, bool hasHeader
   if (hasButtonHints) {
     reservedHeight += metrics.verticalSpacing + metrics.buttonHintsHeight;
   }
-  const int availableHeight = renderer.getScreenHeight() - reservedHeight;
+  const int availableHeight = renderer.getScreenHeight() - reservedHeight - extraReserved;
```
znelson:
```suggestion
  const int availableHeight = renderer.getScreenHeight() - reservedHeight - extraReservedHeight;
```

src/components/UITheme.cpp
```diff
@@ -49,7 +49,7 @@ void UITheme::setTheme(CrossPointSettings::UI_THEME type) {
 }
 
 int UITheme::getNumberOfItemsPerPage(const GfxRenderer& renderer, bool hasHeader, bool hasTabBar, bool hasButtonHints,
-                                     bool hasSubtitle) {
+                                     bool hasSubtitle, int extraReserved) {
```
znelson:
```suggestion
                                     bool hasSubtitle, int extraReservedHeight) {
```

znelson (dismissed):

znelson (dismissed):

src/components/UITheme.cpp
zgredex: Thanks. Sorry for being so chaotic. Still learning this stuff.

znelson: All good 🙂

zgredex:
@znelson master changed and some stuff needed to be solved

lukestein (approved):
No code review, but tested on device and works well, as expected

znelson (approved):

PR #1412, opened 2026-03-17
fix: Build with -fno-exceptions

znelson:
## Summary

Until today, I sincerely thought we were building without exception support. The codebase is not set up with any exception handling infrastructure. The SKILL.md file specifies no exceptions. 

I just learned that we actually were building with exceptions enabled, with `-fexceptions` coming from ~/.platformio/packages/framework-arduinoespressif32-libs/esp32c3/pioarduino-build.py. 

This change removes the `-fexceptions` flag and adds `-fno-exceptions`. The result is **53,670 bytes in flash savings**.

---

### AI Usage

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -36,9 +36,11 @@ build_flags =
   -DPNG_MAX_BUFFERED_PIXELS=16416
   -Wno-bidi-chars
   -Wl,--wrap=panic_print_backtrace,--wrap=panic_abort
+  -fno-exceptions
 
 build_unflags =
   -std=gnu++11
+  -fexceptions
 
 ; Board configuration
 board_build.flash_mode = dio


osteotek (approved):

znelson:
@ngxson thanks for the comment! I ran your comment through Claude for a sanity check and here's what it had to say: 

> Short answer: The pre-compiled libs are safe.
> 
> I scanned all 135 pre-compiled .a files in the ESP32-C3 framework package using `riscv32-esp-elf-nm`. Only one library references any exception symbol: `libcxx.a`, and it's a single weak/undefined reference to `__gxx_personality_v0` from `cxx_guards.cpp.obj` (the thread-safe static local initialization guards). It doesn't define or call `__cxa_throw`, `_Unwind_Resume`, `__cxa_begin_catch`, or any other exception-throwing symbol.
> 
> Here's why this is a non-issue:
> 1. Almost all pre-compiled libs are C code (FreeRTOS, lwIP, mbedTLS, NVS, driver HALs, etc.). C code doesn't use C++ exceptions at all.
> 2. ABI compatibility: Pre-compiled code built with `-fexceptions` and user code built with `-fno-exceptions` link together fine. The difference is that `-fexceptions` code has unwind tables (`.eh_frame`) and landing pads, while `-fno-exceptions` code omits them. The calling convention and data layout are identical -- it's not like mixing `-fno-rtti` objects that disagree on vtable layout.
> 3. What `-fno-exceptions` actually does to our code: It omits the unwind tables from our compiled translation units (which is where the 53KB flash savings comes from). It also makes throw a compile error in our code, and turns any uncaught exception path into a call to `std::terminate()`/`abort()`.
> 4. If a pre-compiled lib hypothetically did throw: It would propagate up until it hits a frame compiled without unwind info, at which point `std::terminate()` is called. But none of the ESP-IDF pre-compiled libs actually throw C++ exceptions -- they all use C-style error codes.
> 5. The build itself is proof: If any linked pre-compiled object required exception infrastructure that was missing, the linker would fail with an undefined symbol error. It doesn't.
> 
> The `__gxx_personality_v0` weak reference in `cxx_guards.cpp.obj` is satisfied by libgcc (the compiler runtime, always linked), and the dead code eliminator simply discards the exception-handling paths that are never reached.

Overall, that makes me feel better. Let me know if you have any remaining concerns. 🙂

ngxson:
A quick decompile reveals that quite a lot of places still want to throw an exception (as part of std library), but it seems like new will now go straight to `std::terminate()` which makes sense, and still align to the behavior with exceptions. So it seems good to me.

image

Just one more thing that I haven't been able to test though, does the exception name still get logged? Maybe try calling an invalid `std::function`, does it still log "bad function call" to serial log?

znelson:
@ngxson Shoot, sorry, I may have jumped the gun on merging this. I'll try to follow up on the logging concern when I can today. If you think we should revert for now, let me know. 

Claude thinks the exception name will still get logged, with this as the analysis:

> **Yes, the exception name and `what()` message will still get logged.** The diagnostic quality is preserved.
> 
> Here's the traced call chain for an unbound `std::function` call:
> 
> 1. `std::function::operator()` on empty target calls `std::__throw_bad_function_call()` -- this lives in `functional.o` inside the **pre-compiled** `libstdc++.a`, which was built **with** exceptions enabled.
> 
> 2. `__throw_bad_function_call()` does `__cxa_allocate_exception(4)` to heap-allocate a `bad_function_call` object, constructs it with the vtable, then calls `__cxa_throw(obj, typeinfo_for_bad_function_call, destructor)`.
> 
> 3. `__cxa_throw` invokes `_Unwind_RaiseException` to start unwinding. It walks the stack, but our frames (compiled with `-fno-exceptions`) have no unwind tables / landing pads, so unwinding fails.
> 
> 4. On unwind failure, `__cxa_throw` calls `std::terminate()`, which calls the registered terminate handler.
> 
> 5. The default handler is `__gnu_cxx::__verbose_terminate_handler()` (in `vterminate.o`). I disassembled it -- here's what it does:
>    - Calls `__cxa_current_exception_type()` to get the in-flight exception's type_info
>    - Calls `__cxa_demangle()` on the mangled type name
>    - `fputs` to stderr: `"terminate called after throwing an instance of '"` + demangled name + `"'"`
>    - Calls `__cxa_begin_catch`, then calls `what()` through the vtable (offset 8)
>    - `fputs` to stderr: `"  what():  "` + the `what()` string (i.e., "bad function call")
>    - `fputc('\n')`, then `__cxa_end_catch`, then `abort()`
> 
> So the serial output would look the same as before:
> 
> ```
> terminate called after throwing an instance of 'std::bad_function_call'
>   what():  bad function call
> ```
> 
> followed by the ESP-IDF abort/panic handler with a backtrace.
> 
> This works because the entire throw-and-terminate path lives in the **pre-compiled libstdc++**, which was built with exceptions. Our `-fno-exceptions` flag only strips unwind tables from **our** translation units (where the 53KB savings come from). The library code that constructs, throws, catches-for-diagnostic, and prints the exception is all pre-compiled and unchanged.

ngxson:
Yeah that's not a big problem, the message is a nice-to-have anyway. At worst we can just wrap the `__cxa_throw` and inject our own message inside.

znelson:
@ngxson Maybe I'm doing this differently than you expected, so it would be great if you could check my work. I added a small patch to call the following on entering one of the activities from home:

```c++
std::function<void()> bad;
bad();
```

Both with and without exceptions, over serial I get a simple log like `abort() was called at PC 0x42146d3d on core 0`, but I don't see the exception message even with exceptions.

[no_exceptions.txt](https://github.com/user-attachments/files/26094018/no_exceptions.txt)
[with_exceptions.txt](https://github.com/user-attachments/files/26094019/with_exceptions.txt)

ngxson:
Sounds good then, I remembered once seeing `bad_alloc` exception in the serial log, though not quite sure if I actually see while working on crosspoint or another project 😅 

In either way, I think allow logging the exception name would be useful, we will need to make a wrapper similar to `__wrap_panic_abort` for that. I'll have a look in another PR

PR #1418, opened 2026-03-17
refactor: Added utilities for non-throwing memory allocation

znelson:
## Summary

Following up on b5df6cb2b5f4415ec7ffa49b06c853a78b741f68. Added lib/Memory/Memory.h with `makeUniqueNoThrow<T>` a `nothrow` wrapper for `std::make_unique` that return `nullptr` on OOM instead of calling `abort()` (the behavior of bare `new` with `-fno-exceptions`)

---

### AI Usage

Did you use AI tools to help write this code? _**PARTIALLY, to update the SKILL.md file**_


znelson:
Related: #1412

lib/JpegToBmpConverter/JpegToBmpConverter.cpp
```diff
@@ -288,24 +289,15 @@ bool JpegToBmpConverter::jpegFileToBmpStreamInternal(FsFile& jpegFile, Print& bm
...
-      delete[] rowAccum;
-      delete[] rowCount;
-      delete atkinsonDitherer;
-      delete fsDitherer;
-      delete atkinson1BitDitherer;
-      free(mcuRowBuffer);
-      free(rowBuffer);
-    }
-  } cleanup{rowBuffer, mcuRowBuffer, atkinsonDitherer, fsDitherer, atkinson1BitDitherer, rowAccum, rowCount};
+  const ScopedCleanup cleanup{[&]() {
```
ngxson: IMO this seems to be an anti pattern now, because all of these can be replaced with the newly added `makeUniqueNoThrow`

ngxson (commented):
Probably we can make the rule a bit more explicit? Something like:
- Heap allocation with custom error handling --> use `makeUniqueNoThrow`
- If a hard crash is expected on OOM, it is acceptable to use normal std types like `std::string` or `std::vector`
- It is not recommend to use raw `malloc` / `free`

Also, just note that there are multiple places still using the `new[]` operator without `std::nothrow`, should we migrate them in the same PR?

Image

And even more places where `new` is used, but I suspect in most places we expect a hard crash on OOM:

(Edit: it seems like these cases are `make_unique(new ...)` or `reset(new ...)`)

Image

lib/JpegToBmpConverter/JpegToBmpConverter.cpp
znelson: > this seems to be an anti pattern now

That's a good point. malloc is nothrow and safe if used correctly, but being explicit that there's one right way to do things in the project (makeUniqueNoThrow) will make it easier to get right in the future.

I'll move this back to draft for now and take another pass at making this more cohesive.

znelson: @ngxson Curious to get your thoughts now. I mechanically fixed every use of `new` and `malloc`, but the changes grow quite extensive. Some cases can still lead to null ptr deref, but now we now consistently log the failure immediately after the failed allocation.

ngxson: Yes I think it's much more reasonable now.

Going a step further, `makeUniqueNoThrow` can also log or keep track of `__LINE__` and `__FILE__`, so we can better report the place where we can counter OOM, in case it crashes. But I'll experiment with this in another PR.

lib/hal/HalStorage.cpp
```diff
@@ -90,8 +96,18 @@ bool HalStorage::openFileForRead(const char* moduleName, const char* path, HalFi
   StorageLock lock;  // ensure thread safety for the duration of this function
   FsFile fsFile;
   bool ok = SDCard.openFileForRead(moduleName, path, fsFile);
-  file = HalFile(std::make_unique<HalFile::Impl>(std::move(fsFile)));
-  return ok;
+  if (!ok) {
+    return false;
+  }
```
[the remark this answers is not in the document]
znelson: This is a non-issue. No fix needed.

The `file` output parameter is a `HalFile&` — a reference to the caller's existing variable. On the failure paths (lines 99-101 and 104-107), the function returns `false`, and `file` is left unchanged. That's correct behavior:

- If `file` was default-constructed by the caller (the typical pattern), it's already an empty/invalid `HalFile` with `impl == nullptr`, so `isOpen()` returns `false`. No stale handle.
- If the caller is reusing a `file` variable that previously held a valid handle, the caller already has ownership of that handle and is responsible for closing it before reuse. Resetting it silently on a failed open would leak the previously-held file descriptor.

The callers always check the `bool` return value before using `file`. The existing code is safe.

lib/hal/HalStorage.cpp
```diff
@@ -153,7 +179,12 @@ bool HalFile::close() { HAL_FILE_WRAPPED_CALL(close, ); }
 HalFile HalFile::openNextFile() {
   HalStorage::StorageLock lock;
   assert(impl != nullptr);
-  return HalFile(std::make_unique<Impl>(impl->file.openNextFile()));
+  auto nextImpl = makeUniqueNoThrow<Impl>(impl->file.openNextFile());
+  if (!nextImpl) {
+    LOG_ERR("HAL", "OOM HalFile::openNextFile");
+    return HalFile();  // Return empty/invalid file
+  }
```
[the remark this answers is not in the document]
znelson: No fix needed. This is overengineering for a scenario that doesn't warrant it.

`openNextFile()` returns an empty `HalFile` on both EOF and OOM. The OOM case already logs via `LOG_ERR("HAL", "OOM HalFile::Impl for openNextFile")`, so the failure is visible in serial output for debugging. The callers (directory listing loops) simply stop iterating when they get an empty `HalFile` — which is the correct behavior in both cases:

- **EOF**: Done iterating, stop.
- **OOM**: System is under severe memory pressure. Stopping iteration and proceeding with whatever entries were already collected is the safest thing to do. Continuing to allocate more entries would make the OOM worse.

Changing the API signature would touch `HalFile` (a public class), `HalStorage.h`, and every caller across the codebase — all to distinguish an error case where the correct recovery action is identical to the normal termination case (stop iterating). The OOM is already logged for diagnostics. No behavioral change would result from the refactor.

ngxson (commented):
To avoid too much changes, I think this PR is pretty much OK to be merged as-is.

Some places where early return could be wrong as coderabbit pointed out, I think they can be converted to `assert(pointer != nullptr)` for now. Without proper UI messages to tell user what's going on, these early-return paths can cause confusions when OOM happen.

I'm working on an error report display on boot (in case reboot from crash), so some places do indeed expect a hard crash.

ngxson:
(Adding to my last comment), we could also add a flag to `makeUniqueNoThrow`, something like:

```
auto ptr = makeUniqueNoThrow(..., CRASH_ON_OOM);
```

osteotek (approved):

znelson:
Superseded by #1832

Issue #1390, opened 2026-03-13
.BMP images do not work


Jepl4r:
I downloaded some `.bmp` images, created a folder named `.sleep` like the user guide said and set the Sleep Screen option to Custom but none of the images show up.

Edit: renamed the folder to `sleep` without the dot and now it works. Why does the folder with the dot not work? It says in the user guide that it should work.

xChikyx:
same here. It .sleep didn't work, but without the dot it worked perfectly

cautious-bread:
I came looking to post this and see if anyone else had the same issue.
Just adding my voice to this issue - are you able to change the title of the post?  It should be "**_.sleep folder not working as expected_**" rather than being a .BMP file issue.

Venus2121:
My sleep folder does not work either, QAQ

znelson:
I believe the bug is here: https://github.com/greiman/SdFat/blob/master/src/FatLib/FatFile.cpp#L669-L671

The SdFat library used by CrossPoint says it intends to skip '.' and '..', but the heuristic is broad enough to skip any file or folder where the first character is '.'.

I wonder how #948 was tested... @jpirnay?

znelson:
Claude suggests we could patch the SdFat library locally like: #1421.

Can anyone here give it a try? You can either apply the [diff](https://github.com/crosspoint-reader/crosspoint-reader/pull/1421.diff) locally or the [firmware artifact](https://github.com/crosspoint-reader/crosspoint-reader/actions/runs/23208570396/artifacts/5970714532) from CI.

jpirnay:
> I believe the bug is here: https://github.com/greiman/SdFat/blob/master/src/FatLib/FatFile.cpp#L669-L671
> 
> The SdFat library used by CrossPoint says it intends to skip '.' and '..', but the heuristic is broad enough to skip any file or folder where the first character is '.'.
> 
> I wonder how [#948](https://github.com/crosspoint-reader/crosspoint-reader/pull/948) was tested... [@jpirnay](https://github.com/jpirnay)?

Strange, I can see and can browse to the .sleep folder (admittedly, I am on top of #1288 and have both .sleep and sleep on my SD) - and it does pick a picture from .sleep

artlan:
There was a fix for .sleep preference on March 2nd in the commits, but it is not in the current 1.1.1 release.
https://github.com/crosspoint-reader/crosspoint-reader/pull/948
[...]

PR #1421, opened 2026-03-17
fix: Patch SdFat to support dot-prefixed directories like .sleep

znelson:
## Summary

SdFat's openNext() and LFN open() skip any directory entry whose 8.3 short filename starts with '.', intending to hide the '.' and '..' meta-entries. This is too broad: with a dot-prefixed directory like '.sleep' with a literal leading dot in the SFN, SdFat discards the entry and the directory becomes inaccessible. 

This pre-build patch narrows the check to only match the actual '.' and '..' entries (SFN fields padded with spaces), allowing user-created dot-prefixed directories to resolve normally through LFN matching.

This will also make dot-prefixed files visible at the filesystem level, but FileBrowserActivity.cpp skips dot-prefixed files at the application level.

Potential fix for #1390.

---

### AI Usage

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -49,6 +49,7 @@ extra_scripts =
   pre:scripts/build_html.py
   pre:scripts/gen_i18n.py
   pre:scripts/patch_jpegdec.py
+  pre:scripts/patch_sdfat.py
   pre:scripts/git_branch.py
 
 ; Libraries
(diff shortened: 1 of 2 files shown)


jpirnay:
Do we have different libs? I can see and browse to .crosspoint or .sleep - either with the integrated browser and #1288 in place, or using the filebrowser on the webinterface and manually adjusting the foldername on the url when browsing (even wo #1288)

PR #1425, opened 2026-03-18
feat: Adjust Navigation at End of Book

nscheung:
## Summary

* **What is the goal of this PR? (e.g., Implements the new feature for file uploading.)**
Currently, pressing forward at the end of a book loops back to the last page. This change will instead sends you to the home page instead. 

Applies the change to the three supported format: EPUB, XTC, TXT

## Additional Context

* This is more of a QOL improvement than a new feature. If there's interest, we could extend this to track a completed state for ebooks.

---

### AI Usage

### DIFF
--- a/src/activities/reader/TxtReaderActivity.cpp
+++ b/src/activities/reader/TxtReaderActivity.cpp
@@ -85,7 +85,6 @@ void TxtReaderActivity::loop() {
       requestUpdate();
     } else {
       onGoHome();
-      return;
     }
   }
 }


src/activities/reader/TxtReaderActivity.cpp
```diff
@@ -79,9 +79,14 @@ void TxtReaderActivity::loop() {
...
-  } else if (nextTriggered && currentPage < totalPages - 1) {
-    currentPage++;
-    requestUpdate();
+  } else if (nextTriggered) {
+    if (currentPage < totalPages - 1) {
+      currentPage++;
+      requestUpdate();
+    } else {
+      onGoHome();
+      return;
```
znelson: I don't think you need this `return;`

nscheung: Fixed!

PR #1427, opened 2026-03-18
feat: battery charging indicator (mirroring PR #537)

jpirnay:
## Summary

* **What is the goal of this PR?** All praise goes to @didacta for his PR #537. Just picked up the reviewer comments to contain the changes as suggested (there was no response for more than 6 weeks, so I wanted to reanimate this feature).

Just one addition: should recognize usb cable plug ins / retractions and update the icon immediately

## Additional Context

see #537 

---

### AI Usage

### DIFF
--- a/src/components/themes/lyra/LyraTheme.cpp
+++ b/src/components/themes/lyra/LyraTheme.cpp
@@ -1,6 +1,7 @@
 #include "LyraTheme.h"
 
 #include <GfxRenderer.h>
+#include <HalGPIO.h>
 #include <HalPowerManager.h>
 #include <HalStorage.h>
 #include <I18n.h>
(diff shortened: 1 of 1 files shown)


src/main.cpp
```diff
@@ -379,6 +379,12 @@ void loop() {
     return;
   }
 
+  // Refresh the battery icon when USB is plugged or unplugged.
+  // Placed after sleep guards so we never queue a render that won't be processed.
+  if (gpio.wasUsbStateChanged()) {
+    activityManager.requestUpdate();
+  }
+
```
ngxson, line 387: I suspect that force-requesting activity update may break certain activities and also make the UX less responsive overall, because eink display takes ~100ms to update even with FAST_REFRESH

Also, since this is an eink display, continuously refreshing one spot may cause permanent burn-in

ngxson: Ah nevermind, sorry I misread it. If the update is only triggered on usb changed event, then it's fine

PR #1428, opened 2026-03-18
feat: add orientation-aware popups for reader activities

mrtnvgr:
## Summary

Make popups (like "Going to sleep") respect the current screen orientation when shown from reader activities.

- Apply reader orientation in SleepActivity before showing popup when `lastSleepFromReader` is true
- Make popup Y-position proportional to screen height (7.5% for BaseTheme, 16.5% for LyraTheme) instead of hardcoded pixel values, ensuring correct positioning in both portrait and landscape modes.
- Add `isReaderActivity()` override to all reader sub-screens (menu, chapter selection, percent selection, footnotes, QR display, KOReader sync), so sleep popups rotate correctly when entering sleep from any reader context.

## Additional Context

---

### AI Usage

Did you use AI tools to help write this code? _**YES**_ (Claude Opus 4.5)

### DIFF
--- a/src/activities/reader/EpubReaderChapterSelectionActivity.h
+++ b/src/activities/reader/EpubReaderChapterSelectionActivity.h
@@ -32,4 +32,5 @@ class EpubReaderChapterSelectionActivity final : public Activity {
   void onExit() override;
   void loop() override;
   void render(RenderLock&&) override;
+  bool isReaderActivity() const override { return true; }
 };
--- a/src/activities/boot_sleep/SleepActivity.cpp
+++ b/src/activities/boot_sleep/SleepActivity.cpp
@@ -10,13 +10,22 @@
 
 #include "CrossPointSettings.h"
 #include "CrossPointState.h"
+#include "activities/reader/ReaderUtils.h"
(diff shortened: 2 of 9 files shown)


mrtnvgr:
@coderabbitai review

lukestein (approved):
No code review, but tested on device and seems to work well

src/activities/reader/EpubReaderChapterSelectionActivity.h
```diff
@@ -32,4 +32,5 @@ class EpubReaderChapterSelectionActivity final : public Activity {
   void onExit() override;
   void loop() override;
   void render(RenderLock&&) override;
+  bool isReaderActivity() const override { return true; }
```
znelson, line 35: We could remove a bunch of these `isReaderActivity` overrides in a future change by making ActivityManager walk the activity stack, e.g.:

```c++
bool ActivityManager::isReaderActivity() const {
  for (const auto& activity : stackActivities) {
    if (activity->isReaderActivity()) return true;
  }
  return false;
}
```

znelson (approved):

PR #1433, opened 2026-03-20
refactor: RAII scoped open/close for ZipFile

znelson:
## Summary

Added `ScopedOpenClose` RAII guard to eliminate repetitive `wasOpen`/`close()` boilerplate across all ZipFile methods.

---

### AI Usage


ngxson (approved):
This is ok for a bug-for-bug fix, but the design of `ZipFile` seems strange IMO.

I would expect a design pattern like this:
- File is automatically opened inside constructor
- Automatically close file inside destructor
- If any of the operators is called during file closed, we return error

jpirnay (approved):
lgtm

PR #1437, opened 2026-03-20
refactor: Deduplicate battery drawing code and fix Lyra charging indicator

znelson:
## Summary

Following up on #1427:
- Extracted shared battery drawing logic, including lightning bolt, to reduce duplication. 
- In Lyra with segmented battery the lightning bolt was hard to see, so when charging Lyra now uses a solid battery fill.

---

### AI Usage


jpirnay (dismissed):
Have tested it in my build works as advertised

znelson:
| Before | After |
| -- | -- |
|  |  |

PR #1445, opened 2026-03-20
feat: Local overrides for css and img display

jpirnay:
## Summary

* **What is the goal of this PR?** Allow per book override of css- / image-use

## Additional Context
grafik

---

### AI Usage


iandchasse:
I like this, some books of mine have quirky css. I wonder if an additional PR could be done for the font style/size for each book. It'd be a good excuse to have the ability to change font from reader menu instead of the settings menu.

PR #1031, opened 2026-02-20
feat: slim footnotes support

Uri-Tauber:
## Summary
**What is the goal of this PR?** Implement support for footnotes in epub files.
   It is based on #553, but simplified — removed the parts which complicated the code and burden the CPU/RAM. This version supports basic footnotes and lets the user jump from location to location inside the epub.

**What changes are included?**
    - `FootnoteEntry` struct — A small POD struct (number[24], href[64]) shared between parser, page storage, and UI.
    - Parser: `<a href>` detection (`ChapterHtmlSlimParser`) — During a single parsing pass, internal epub links are detected and collected as footnotes. The link text is underlined to hint navigability. Bracket/whitespace normalization is applied to the display label (e.g. [1] → 1).
    - Footnote-to-page assignment (`ChapterHtmlSlimParser`, `Page`) — Footnotes are attached to the exact page where their anchor word appears, tracked via a cumulative word counter during layout, surviving paragraph splits and the 750-word mid-paragraph safety flush.
    - Page serialization (`Page`, `Section`) — Footnotes are serialized/deserialized per page (max 16 per page). Section cache version bumped to 14 to force a clean rebuild.
    - Href → spine resolution (`Epub`) — `resolveHrefToSpineIndex()` maps an href (e.g. `chapter2.xhtml#note1`) to its spine index by filename matching.
    - Footnotes menu + activity (`EpubReaderMenuActivity`, `EpubReaderFootnotesActivity`) — A new "Footnotes" entry in the reader menu lists all footnote links found on the current page. The user scrolls and selects to navigate.
    - Navigate & restore (`EpubReaderActivity`) — `navigateToHref()` saves the current spine index and page number, then jumps to the target. The Back button restores the saved position when the user is done reading the footnote.

  **Additional Context**

**What was removed vs #553:** virtual spine items (`addVirtualSpineItem`, `isVirtualSpineItem`), two-pass parsing, `<aside>` content extraction to temp HTML files, `<p class="note">` paragraph note extraction, `replaceHtmlEntities` (master already has `lookupHtmlEntity`), `footnotePages` / `buildFilteredChapterList`, `noterefCallback` / `Noteref` struct, and the stack size increase from 8 KB to 24 KB (not needed without two-pass parsing and virtual file I/O on the render task).
 
[...]

PR #1455, opened 2026-03-21
feat: TOC anchor navigation for multi-chapter spine files

### LINKED ISSUE #1131: Incorrect chapter length
### Affected Version

1.1.0

### Describe the Bug

I'm reading It by Stephen King. In chapter 3 crosspoint thinks chapter 3 is 160 pages long but then when you press next page, the chapter continues and is now another 142 pages. See the screenshots below of progress. 

Image

Image

Image

### Steps to Reproduce

Read book

### Expected Behavior

Chapter 3 should be 202 pages long.

```shell

```

### LINKED ISSUE #1604: Epub ToC parsing improvements
Top-level category for various epub ToC parsing improvements targeted for the next release.

znelson:
## Summary

- Fix for #383: epub books where multiple TOC items point into a single spine file via fragment anchors.
- Builds off of work from @Uri-Tauber in #1031 and #1245.
- Brings forward earlier work from #1143, updated for all of the recent work in main.
- First half of #1447, where the second half will fix #1131.

EPUBs that pack multiple TOC chapters into a single XHTML spine file use fragment anchors (e.g. `chapter1.xhtml#section-5`) to distinguish sub-chapters. The reader previously ignored these, causing three bugs:
- TOC navigation jumped to the start of the spine file instead of the target chapter
- Chapter skip (long-press) jumped by spine boundaries instead of TOC entries
- The status bar showed the wrong chapter name

## Changes

**Parser:** When an HTML `id` matches a TOC entry's anchor, forces a page break so chapters start on fresh pages. All `id` attributes are recorded with their page numbers in the section cache for later lookup (shared with footnote anchor navigation from #1245). Page-break and anchor-recording logic is centralized in `startNewTextBlock` via the existing `pendingAnchorId` deferred pattern.

**Section:** On load, builds a small in-memory `tocBoundaries` vector by scanning the on-disk anchor data (written by #1245) for just the TOC-relevant anchors. Provides `getTocIndexForPage()` and `getPageForTocIndex()` for in-memory lookups with no file I/O per page turn. No `std::map` or `std::set`, all storage uses flat vectors.

**Navigation:** Chapter skip walks TOC entries instead of spine indices. Same-spine jumps resolve in memory; cross-spine jumps use a deferred `pendingTocIndex` resolved after the target section loads. Guards against `curTocIndex < 0` for orphan spines (cover pages, end-matter). The chapter selector passes `tocIndex` back via `ChapterResult` so the reader lands on the correct sub-chapter.

**Status bar:** Uses per-page `getTocIndexForPage()` for the chapter title so the correct sub-chapter name is shown.

**Documentation:** `docs/epub-toc-navigation.md` covers the architecture, edge cases, and performance characteristics.

**Test epub:** `scripts/generate_spine_toc_edges_epub.py` generates a test book exercising all spine/TOC relationship patterns (multi-TOC-per-spine, spine with no TOC entry, mid-file anchor, nested hierarchy, normal 1:1).

---

### AI Usage


jonvex:
Will this fix work for unnamed chapters/ chapters that are multiple files? I can try to fix that if it doesn't

znelson:
@jonvex I'm not sure about that. Do you have a document I could test with?

znelson:
@jonvex I just updated from main, so if you want to give it try locally please do.

jonvex:
@znelson thanks I'll test in a bit and let you know

jonvex:
@znelson, I tested and the behavior seems to be the same as before. I emailed you the file

lib/Epub/Epub/Section.cpp
```diff
@@ -410,7 +410,7 @@ void Section::buildTocBoundariesFromFile(FsFile& f) {
 
   // Single pass through on-disk anchors, matching against cached TOC anchors.
   // Stop early once all TOC anchors are resolved.
-  f.seek(HEADER_SIZE - sizeof(uint32_t));
+  f.seek(HEADER_SIZE - sizeof(uint32_t) * 2);
```
pablohc: CRITICAL BUG: Header seek offset in buildTocBoundariesFromFile

buildTocBoundariesFromFile does:

f.seek(HEADER_SIZE - sizeof(uint32_t) * 2);
uint32_t anchorMapOffset;

But the current header layout has 4 uint32_t fields at the end (lutOffset, anchorMapOffset, paragraphLutOffset, liLutOffset):
- HEADER_SIZE - sizeof(uint32_t) * 2 points to paragraphLutOffset, not anchorMapOffset
- The correct offset is HEADER_SIZE - sizeof(uint32_t) * 3

Commit 4ce37a4 claims to fix this, but the diff I fetched from GitHub still shows * 2. This needs immediate verification — if the offset is wrong, it silently reads garbage or a different field, causing incorrect TOC boundaries or crashes.

Compare with getPageForAnchor() at Section.cpp:326 which correctly uses HEADER_SIZE - sizeof(uint32_t) * 3.

lib/Epub/Epub/parsers/ChapterHtmlSlimParser.cpp
```diff
@@ -140,15 +140,35 @@ void ChapterHtmlSlimParser::startNewTextBlock(const BlockStyle& blockStyle) {
...
+            currentPage.reset(new Page());
+            currentPageNextY = 0;
+          }
+        }
         anchorData.push_back({std::move(pendingAnchorId), static_cast<uint16_t>(completedPageCount)});
         pendingAnchorId.clear();
       }
+      wordsExtractedInBlock = 0;
       return;
     }
```
pablohc, line 151: There are two code paths that repeat the same pattern:
1. Empty block reuse path (~line 143 in the PR diff) — when currentTextBlock->isEmpty() is true:
if (!pendingAnchorId.empty()) {
    if (std::find(tocAnchors.begin(), tocAnchors.end(), pendingAnchorId) != tocAnchors.end()) {
        // force page break...
    }
    anchorData.push_back(...);
    pendingAnchorId.clear();
}
return;
2. Non-empty block path (~line 155 in the PR diff) — after makePages():
if (!pendingAnchorId.empty() &&
    std::find(tocAnchors.begin(), tocAnchors.end(), pendingAnchorId) != tocAnchors.end()) {
    // force page break... (same logic)
}
if (!pendingAnchorId.empty()) {
    anchorData.push_back(...);
    pendingAnchorId.clear();
}
Both do the same three things: check if pendingAnchorId is a TOC anchor, conditionally force a page break, then record the anchor and clear pendingAnchorId.

pablohc:
Some feedback by Claude:

Overall Architecture
The approach is sound and well-suited for ESP32-C3:
- Parser: forces page breaks at TOC anchors during pagination
- Section: in-memory tocBoundaries vector for O(1) lookups with no file I/O
- Navigation: chapter skip by TOC index instead of spine index
- Status bar: per-page chapter title instead of per-spine
The choice of flat vectors over std::map/std::set is the right call for ESP32-C3 to avoid heap fragmentation.

<details>

```
CRITICAL BUG: Header seek offset in buildTocBoundariesFromFile
buildTocBoundariesFromFile does:
f.seek(HEADER_SIZE - sizeof(uint32_t) * 2);
uint32_t anchorMapOffset;
But the current header layout has 4 uint32_t fields at the end (lutOffset, anchorMapOffset, paragraphLutOffset, liLutOffset):
- HEADER_SIZE - sizeof(uint32_t) * 2 points to paragraphLutOffset, not anchorMapOffset
- The correct offset is HEADER_SIZE - sizeof(uint32_t) * 3
Commit 4ce37a4 claims to fix this, but the diff I fetched from GitHub still shows * 2. This needs immediate verification — if the offset is wrong, it silently reads garbage or a different field, causing incorrect TOC boundaries or crashes.
Compare with getPageForAnchor() at Section.cpp:326 which correctly uses HEADER_SIZE - sizeof(uint32_t) * 3.
```

```
MEDIUM: Duplicated page-break logic in startNewTextBlock
The force-page-break + record-anchor block is duplicated across the empty-block path and the non-empty-block path:
// Path 1: empty block reuse
if (!pendingAnchorId.empty()) {
    if (std::find(tocAnchors.begin(), tocAnchors.end(), pendingAnchorId) != tocAnchors.end()) {
        // force page break...
    }
    anchorData.push_back({std::move(pendingAnchorId), ...});
    pendingAnchorId.clear();
}
// Path 2: after makePages
if (!pendingAnchorId.empty() && std::find(...)) {
    // force page break... (exact same logic)
}
if (!pendingAnchorId.empty()) {
    anchorData.push_back({std::move(pendingAnchorId), ...});
    pendingAnchorId.clear();
}
Recommendation: Extract into a private helper like flushPendingAnchor(bool forcePageBreak) to prevent the two paths from diverging in future changes.
```

```
MEDIUM: std::move on const object (valid CodeRabbit nitpick)
In buildTocBoundariesFromFile:
const auto entry = epub->getTocItem(i);
tocAnchorsToResolve.push_back({i, std::move(entry.anchor)});
std::move on a const object's field doesn't move — it copies (produces const string&&, which can only bind to a copy). Should either make entry non-const or drop std::move.
```

```
MINOR: Missing #include <algorithm>
Section.cpp uses std::sort and std::upper_bound after this PR but doesn't include <algorithm>. While it likely works via transitive includes, it's not portable and should be explicit.
```

```
MINOR: getTocIndexForPage() behavior before first boundary
When tocBoundaries is non-empty and the page falls before the first startPage, getTocIndexForPage returns tocBoundaries[0].tocIndex. This is likely correct (early pages belong to the first chapter in this spine), but should be documented explicitly.
```

</details>

Positives
1. Zero overhead for the common case — early exit when unresolvedCount == 0 (1:1 TOC-to-spine), no boundaries stored, all queries fall back to spine-level methods
2. buildTocBoundariesFromFile caches TOC anchor strings before the disk scan loop — avoids repeated file I/O to BookMetadataCache per on-disk entry
3. Defensive sort of tocBoundaries — safety net for malformed EPUBs where TOC entries are out of document order
4. Excellent documentation in docs/epub-toc-navigation.md — covers architecture, edge cases, performance trade-offs, and implementation pitfalls
5. Comprehensive test epub covering multi-TOC-per-spine, multi-spine-per-TOC, orphan spines, mid-file anchors, and nested hierarchy
6. pendingAnchorId overwrite is honestly documented as a known limitation inherited from PR #1245, with a clear explanation of why it doesn't affect TOC navigation in practice

lib/Epub/Epub/Section.cpp
```diff
@@ -279,6 +298,157 @@ std::unique_ptr<Page> Section::loadPageFromSectionFile() {
...
+          break;
+        }
+      }
+    }
+    tocBoundaries.push_back({i, page});
+  }
+
+  // Defensive sort in case TOC entries are out of document order in a malformed epub
+  std::sort(tocBoundaries.begin(), tocBoundaries.end(),
+            [](const TocBoundary& a, const TocBoundary& b) { return a.startPage < b.startPage; });
```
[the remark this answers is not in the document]
Uri-Tauber: @znelson mind solving this copilot comment?
maybe something like 
```suggestion
  // Defensive sort in case TOC entries are out of document order in a malformed epub
  std::sort(tocBoundaries.begin(), tocBoundaries.end(), [](const TocBoundary& a, const TocBoundary& b) {
   if (a.startPage != b.startPage) return a.startPage < b.startPage;
   return a.tocIndex < b.tocIndex;
 });
```

lib/Epub/Epub/Section.cpp
```diff
@@ -303,6 +322,157 @@ std::unique_ptr<Page> Section::loadPageFromSectionFile() {
...
+    const auto entry = epub->getTocItem(i);
+    tocBoundaries.push_back({i, 0});
+    if (!entry.anchor.empty()) {
+      tocAnchorsToResolve.push_back({i, std::move(entry.anchor)});
+    }
+  }
+
+  // Single pass through on-disk anchors, matching against cached TOC anchors.
+  // Stop early once all TOC anchors are resolved.
+  f.seek(HEADER_SIZE - sizeof(uint32_t) * 2);
```
Uri-Tauber: Not sure about the Claude comment, but it seems to me this could benefit from clearer naming:
```suggestion
 constexpr uint32_t ANCHOR_MAP_OFFSET_FIELD_POS = HEADER_SIZE - sizeof(uint32_t) * 2;
 f.seek(ANCHOR_MAP_OFFSET_FIELD_POS);
```

src/activities/reader/EpubReaderActivity.cpp
```diff
@@ -850,28 +917,25 @@ void EpubReaderActivity::renderStatusBar() const {
   const float bookProgress = epub->calculateProgress(currentSpineIndex, sectionChapterProg) * 100;
 
   std::string title;
-
   int textYOffset = 0;
 
   if (automaticPageTurnActive) {
     title = tr(STR_AUTO_TURN_ENABLED) + std::to_string(60 * 1000 / pageTurnDuration);
 
-    // calculates textYOffset when rendering title in status bar
```
Uri-Tauber, line 873: Any reason why you removed this comment?

src/activities/reader/EpubReaderActivity.cpp
```diff
@@ -850,28 +917,25 @@ void EpubReaderActivity::renderStatusBar() const {
...
     const uint8_t statusBarHeight = UITheme::getInstance().getStatusBarHeight();
-
-    // offsets text if no status bar or progress bar only
```
Uri-Tauber, line 876: and this?

scripts/generate_spine_toc_edges_epub.py
```diff
@@ -0,0 +1,2096 @@
+#!/usr/bin/env python3
+"""
+Generate a test EPUB that exercises edge cases in spine/TOC relationships.
```
Uri-Tauber, line 3: I can't think of any reason why someone will run this script ever.
If I would want to change the epub, it would be much quicker to manually edit it...
That said, I don't really mind this.

Uri-Tauber (changes requested):

scripts/generate_spine_toc_edges_epub.py
znelson: Hmm... the other sample documents have Python scripts to produce them so this is following an existing pattern. Is it faster to unzip, modify HTML, and carefully re-zip (with mimetype first) than to run a script? Do you have a good workflow for that process?

I would definitely support pulling the content for these out into version-controlled files, and consolidating to a single script to generate all of the test documents.

lib/Epub/Epub/Section.cpp
znelson: Yeah the offset calculations in this file are error prone. I have another branch where all of the header offsets are calculated into named constexpr values.

src/activities/reader/EpubReaderActivity.cpp
znelson: Just mistakes after months of conflicts, I guess...

src/activities/reader/EpubReaderActivity.cpp
znelson: Just mistakes after months of conflicts, I guess...

lib/Epub/Epub/Section.cpp
znelson: Thanks, #1986 hopefully makes these bugs less likely in the future

lib/Epub/Epub/Section.cpp
znelson: Something like #1986 will probably help going forward

scripts/generate_spine_toc_edges_epub.py
Uri-Tauber: You're right there are several of those stale scripts...
I'm using Calibre edit. it skips the entire process you mention - just need to modify the html and save...

Uri-Tauber:
@znelson thanks for the quick fixes. I will test on my device after Shabbat.

jspmarc:
@znelson thank you for this fix! Out of curiosity, is there a reason to close this PR? this PR seems to work perfectly with the epub I tested

znelson:
@jspmarc Thanks for testing. I still believe the fix itself is valuable. I closed this and my other pending PRs because I'm not in a position to actively maintain or shepherd queued changes through the review and release process right now.

DianaNites:
Shouldnt they remain open or have some sort of "postponed" label, in that case, for easy re-evaluation triage later vs rejected or obsolete PRs?

znelson:
@DianaNites The new project leadership (@itsthisjustin) can decide whether they want to revive, reimplement, or revisit any of these changes later.

PR #1224, opened 2026-02-27
feat: integrated epub optimizer

pablohc:
## Problem

Many e-ink readers have limited image decoder support natively.
EPUBs with images in other formats than **baseline JPEG** frequently cause:

- **Broken images**: pages render as blank, corrupted noise, or never load
- **Slow rendering**: unoptimized images cause severe delays on e-ink hardware, up to 7 seconds per page turn, with cover images taking up to 59 seconds to render
- **Broken covers**: the book thumbnail never generates

Fixing this today requires external tools before uploading.

---

## What this PR does

Adds an **optional, on-demand EPUB optimizer** to the file upload flow. When enabled,
it converts all images to baseline JPEG directly in the browser — no server, no internet,
no external tools needed.

**Conversion is opt-in. The standard upload flow is unchanged.**

---

## Real-world impact

The optimizer was applied in batch to **61 EPUBs**:
- 60 standard EPUBs: 198 MB → 55 MB (**−72.2%**, 143 MB saved)
  - Text-dominant books: 8–46% smaller (covers and inline images converted)
  - Image-heavy / illustrated books: 65–93% smaller
- 1 Large manga volume (594 MB): 594 MB → 72 MB (**−87.8%**, 522 MB saved)
- EPUB structural integrity fully maintained — zero new validation issues introduced across all 61 books

*Size and integrity analysis: [epub-comparator](https://github.com/pablohc/epub-comparator)*

From that set, **17 books were selected** as a representative sample covering different content
types: image-heavy novels, pure manga, light novels with broken images, and text-dominant books.
Each was benchmarked on two devices running in parallel, one on `master` and one
on `PR#1224` — measuring render time across ~30 pages per book on average.

### Rendering bugs fixed

| Book | Problem (original) | After optimization |
|------|--------------------|--------------------|
| Fairy Tale — Stephen King | Cover took **59.7 s** to render | 2.1 s (−96%) |
| Cycle of the Werewolf — Stephen King | Cover took **23.3 s** to render | 1.7 s (−93%) |
| Tomie: Complete Deluxe Ed. — Junji Ito | Cover took **18.3 s** to render | 2.0 s (−89%) |
| Joel Dicker — El tigre (Ed. Ilustrada) | Cover took **14.5 s** to render | 1.4 s (−90%) |
| Jackson, Holly — Asesinato para principiantes | Cover failed completely (blank) | 2.0 s ✓ |
[...]

PR #1462, opened 2026-03-22
feat: add OPDS search support & next/prev page navigation

rxmmah:
## Summary

Adds OPDS search support, allowing users to search a catalog directly from the book browser when the server exposes an OpenSearch template.

- `OpdsParser`: parses the OpenSearch template URL from feed-level `<link rel="search">` elements and exposes it via `getSearchTemplate()`
- `OpdsBookBrowserActivity`: fetches and stores the search template after each feed load; shows a Search hint on the Left button when a template is available; launches the existing `KeyboardEntryActivity` for query input; URL-encodes the query and fetches the result feed
- Absolute search result URLs are handled correctly in `fetchFeed` (skips prepending the server base URL)
- A `consumeConfirm` guard prevents the Confirm release that submits the keyboard from immediately triggering a book download on the first browsing frame after search results load

## Additional Context

- Search is silently unavailable if the server does not advertise an OpenSearch template — no UI change in that case
- Tested against a Calibre-Web OPDS endpoint which exposes `<link rel="search" type="application/opensearchdescription+xml">`
- The inline URL encoder in `performSearch` was necessary as `StringUtils` has no such utility; worth considering extracting to `StringUtils` in a follow-up
- No new dependencies introduced

---

### AI Usage


itsthisjustin:
@rxmmah thanks for doing this. I cleaned up the button to work with the new localization system, made it only work as search when you're on the first item (otherwise it was impossible to use the up button) and added support for next and previous navigation

itsthisjustin (approved):

itsthisjustin:
@znelson @pablohc can I get a second review on this?

pablohc:
Hey @itsthisjustin, I wish I could approve but I don't have the permissions for that. That said, I'd genuinely love to contribute to the project, I'm not a C++ guy, so code review isn't really my lane, but if @znelson finds it useful I'm happy to validate improvements from PRs directly on the device and share real-world usage feedback and log analysis, as I've been doing in PR #1224.

itsthisjustin:
> Hey @itsthisjustin, I wish I could approve but I don't have the permissions for that. That said, I'd genuinely love to contribute to the project, I'm not a C++ guy, so code review isn't really my lane, but if @znelson finds it useful I'm happy to validate improvements from PRs directly on the device and share real-world usage feedback and log analysis, as I've been doing in PR #1224.

Sorry for the eroneous tag there @pablohc. Tagging in @martinbrook @osteotek

lib/I18n/translations/ukrainian.yaml
```diff
@@ -45,7 +45,7 @@ STR_OPEN_URL_HINT: "Відкрийте цю URL-адресу у вашому б
 STR_OR_HTTP_PREFIX: "або http://"
 STR_SCAN_QR_HINT: "або відскануйте QR-код телефоном:"
 STR_CALIBRE_WIRELESS: "Calibre бездротовий"
-STR_CALIBRE_WEB_URL: "URL Calibre Web"
+STR_CALIBRE_WEB_URL: "OPDS URL"
```
znelson, line 48: Looks like we could remove this from localization?

znelson (approved):
Added @pablohc to the maintainers group. 🚀

lib/I18n/translations/ukrainian.yaml
itsthisjustin: Probably yes. I didn't want to break anything and it doesn't really "hurt" in case we do want to rename this later

PR #1467, opened 2026-03-23
feat: Show progress bar during chapter indexing

jpirnay:
## Summary
- Replaces the static "Indexing" popup with a progress bar that fills as the chapter HTML is parsed
- Progress is estimated from bytes read vs total file size, reported in 5% increments to limit e-ink refreshes
- Reuses the existing `drawPopup`/`fillPopupProgress` infrastructure (already used by HomeActivity and BmpViewerActivity)

## Changes
- `popupFn` callback (`std::function<void()>`) changed to `progressFn` (`std::function<void(int)>`) across `ChapterHtmlSlimParser`, `Section`, and `EpubReaderActivity`
- Parser tracks `bytesRead` in the Expat parse loop and calls back with progress 0-100
- `EpubReaderActivity` draws the popup on first callback, then updates the progress bar on subsequent ones
- Silent pre-indexing (next chapter) is unaffected — still passes `nullptr`

### DIFF
--- a/lib/Epub/Epub/Section.cpp
+++ b/lib/Epub/Epub/Section.cpp
@@ -135,7 +135,7 @@ bool Section::clearCache() const {
 bool Section::createSectionFile(const int fontId, const float lineCompression, const bool extraParagraphSpacing,
                                 const uint8_t paragraphAlignment, const uint16_t viewportWidth,
                                 const uint16_t viewportHeight, const bool hyphenationEnabled, const bool embeddedStyle,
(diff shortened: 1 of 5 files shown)


jpirnay:
clang-format fails as the command to install clang-format fails in the action - modules should be properly formatted

znelson (commented):
LGTM. I re-ran the clang-format CI step and it passed. :+1:

znelson (approved):

PR #1474, opened 2026-03-23
feat: Add remove button to Recent Books screen

jpirnay:
## Summary

* **What is the goal of this PR?** Allow users to remove individual books from the Recent Books list directly from the Recent Books screen.
  * Added `RecentBooksStore::removeBook()` method to remove a book by path and persist the change
  * Repurposed the Left button in `RecentBooksActivity` from navigation (scroll up) to a "Remove" action that shows a confirmation dialog before removing the selected book
  * Added `STR_REMOVE` i18n key ("Remove") and updated button hints to display it
  * Removed unused previous-button navigation (up/left) and `GO_HOME_MS` constant from `RecentBooksActivity`

## Additional Context

* Follows the same confirmation dialog pattern used in `FileBrowserActivity` for file deletion
* Navigation through the list is now down/right only (single direction scroll), which is sufficient for a short list (max 10 items)
* The book is only removed from the recent list, not deleted from storage

---

### AI Usage


src/activities/home/RecentBooksActivity.cpp
```diff
@@ -59,27 +56,44 @@ void RecentBooksActivity::loop() {
...
 
   buttonNavigator.onNextRelease([this, listSize] {
     selectorIndex = ButtonNavigator::nextIndex(static_cast<int>(selectorIndex), listSize);
     requestUpdate();
   });
 
-  buttonNavigator.onPreviousRelease([this, listSize] {
-    selectorIndex = ButtonNavigator::previousIndex(static_cast<int>(selectorIndex), listSize);
-    requestUpdate();
-  });
```
znelson, line 72: The Left button repurposing looks good, but _I think_ removing `onPreviousRelease` and `onPreviousContinuous` also kills backward navigation via the Up side-button. 

`ButtonNavigator::getPreviousButtons()` returns `Up, Left`, and since the new explicit Left handler already returns out of `loop()` before the `buttonNavigator` calls are reached, there's no conflict. Restoring both of these (`onPreviousRelease` and `onPreviousContinuous`) _should_ give Up-button backward scrolling back for free while keeping Left for Remove.

jpirnay: You are right, well spotted.

znelson (approved):

PR #682, opened 2026-02-03
feat: enhance file deletion functionality with multi-select

Jessica765:
## Summary

* **What is the goal of this PR?** Enhances the file manager with multi-select deletion functionality and improved UI formatting.
* **What changes are included?**
	* Added multi-select capability for file deletion in the web interface
	* Fixed formatting issues in file table for folder rows
	* Updated [.gitignore] to exclude additional build artifacts and cache files
	* Refactored CrossPointWebServer.cpp to support batch file deletion
	* Enhanced FilesPage.html with improved UI for file selection and deletion

## Additional Context

* The file deletion endpoint now handles multiple files in a single request, improving efficiency when removing multiple files
* Changes are focused on the web file manager component only

---

### AI Usage

While CrossPoint doesn't have restrictions on AI tools in contributing, please be transparent about their usage as it 
helps set the right context for reviewers.

Did you use AI tools to help write this code? _**PARTIALLY**_

PR #1475, opened 2026-03-24
fix: webserver /delete API backward compatibility

DianaNites:
## Summary

This fixes #682 by restoring API compatibility for external users, eg deleting on-device books in bulk via calibre using the crosspoint plugin works now.

The `/delete` webserver API now accepts the old `path` argument and maps it to a JSON array element. It is made an error to provide both arguments.

## Additional Context

  specific areas to focus on).

I have compiled, flashed, and tested this on master at commit 0245972132b54563976525a5841a0e9c49e8e498. Before this change, calibre cannot delete books, either individually or in bulk.

After this change, it can do so successfully.

---

### AI Usage

### DIFF
--- a/src/network/CrossPointWebServer.cpp
+++ b/src/network/CrossPointWebServer.cpp
@@ -959,16 +959,31 @@ void CrossPointWebServer::handleMove() const {
 }
 
 void CrossPointWebServer::handleDelete() const {
-  // Check if 'paths' argument is provided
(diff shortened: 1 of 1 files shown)


Uri-Tauber (approved):
I haven’t tested it on my own device since I don’t use Calibre for transferring ebooks, but the code looks solid and compiles without any issues.

znelson (approved):

PR #1476, opened 2026-03-24
fix: swedish translation

steka:
## Summary

  fix swedish translation
  lib\I18n\translations\swedish.yaml
## Additional Context

  specific areas to focus on).

---

### AI Usage


lib/I18n/translations/swedish.yaml
```diff
@@ -224,7 +225,7 @@ STR_BATTERY: "Batteri"
 STR_UI_THEME: "Användargränssnittstema"
 STR_THEME_CLASSIC: "Klassisk"
 STR_THEME_LYRA: "Lyra"
-STR_THEME_LYRA_EXTENDED: "Lyra Extended"
+STR_THEME_LYRA_EXTENDED: "Lyra utökad"
```
znelson, line 228: Generally we haven't been translating the theme names (they should probably be removed from translation), but it doesn't hurt anything

lukestein: Agreed. I'll approve anyway @znelson but leave to you to merge if either this is removed or before next release (better done than perfect).

lukestein (approved):

steka:
Just so you know STR_THEME_LYRA_EXTENDED has already been transtaded in some of the other languages.
However I do understand that it maybe shouldn't be translated, but in that case it shouldn't be listed at all as a STR_...

PR #1499, opened 2026-03-25
feat: Reading ruler for EPUB reader

jpirnay:
## Summary

* **What is the goal of this PR?** Add a reading ruler feature that helps readers track their current line while reading EPUBs on the e-ink display.
  - Single-line underline ruler drawn below the current line, navigated line-by-line with page turn buttons (short press)
  - Long press falls through to normal page turn / chapter skip behavior
  - Auto-progress: ruler advances automatically after a configurable idle timeout (off / 3 / 5 / 8 / 10 / 15 seconds)
  - Per-book ruler state (on/off + line index) persisted in `progress.bin` (backward-compatible: 9 bytes, up from 6)
  - Toggle in reader menu when global setting is enabled; per-book default follows the global setting
  - New settings: "Reading Ruler" toggle and "Ruler Auto-Progress" interval selector
  - I18n strings added for English and German

## Additional Context

* The ruler is rendered as a 2px horizontal line below the current text line, spanning the content area
* `progress.bin` format is extended but backward-compatible — older files (4/6 bytes) still load correctly
* Page transitions at ruler boundaries (last line forward, first line backward) work seamlessly
* Ruler state survives menu open/close without page re-render

---

### AI Usage

### DIFF
--- a/src/activities/reader/EpubReaderActivity.cpp
+++ b/src/activities/reader/EpubReaderActivity.cpp
@@ -32,6 +32,7 @@ namespace {
 constexpr unsigned long skipChapterMs = 700;
 // pages per minute, first item is 1 to prevent division by zero if accessed
 const std::vector<int> PAGE_TURN_LABELS = {1, 1, 3, 6, 12};
+constexpr int READING_RULER_LINE_WIDTH = 2;
 
 int clampPercent(int percent) {
   if (percent < 0) {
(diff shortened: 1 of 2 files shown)


iandchasse:
This is an interesting feature, I see it's mainly on the Kindle app but not on their e-readers (?) How well does it work? I'd imagine that doing a fast refresh for every line on a page might result in some fading by the end of it, no?

jpirnay:
It's actually quite okay:
https://drive.google.com/file/d/1xGjhxsRJLuvYG_DE8HDt3xHztuue3-fP/view?usp=sharing

iandchasse:
Ohhh I see. It refreshes the entirety of the screen. I see it also does the anti-aliasing every step, which is a bit jarring, but I suppose if it didn't do that we might see more of the fading? It's an interesting feature for sure

jpirnay:
> Ohhh I see. It refreshes the entirety of the screen. I see it also does the anti-aliasing every step, which is a bit jarring, but I suppose if it didn't do that we might see more of the fading? It's an interesting feature for sure

It can and should be improved - I was trying to get it functional first to see whether there is a usecase for people with visual challenges

jpirnay:
Looking smoother now

https://drive.google.com/file/d/1XT6C4ivLvHbTP8SSQEQrjamJepQgqqIu/view?usp=sharing

PR #1502, opened 2026-03-25
fix: remove zero-byte guard in HttpDownloader::downloadToFile that rejects legitimate empty 200 OK responses

jpirnay:
## Summary

* **What is the goal of this PR?** Dealing with issue #1485

## Additional Context

  specific areas to focus on).

---

### AI Usage

### DIFF
--- a/src/activities/browser/OpdsBookBrowserActivity.cpp
+++ b/src/activities/browser/OpdsBookBrowserActivity.cpp
@@ -333,6 +333,20 @@ void OpdsBookBrowserActivity::downloadBook(const OpdsEntry& book) {
       });
 
   if (result == HttpDownloader::OK) {
+    FsFile downloadedFile;
+    if (!Storage.openFileForRead("OPDS", filename, downloadedFile) || downloadedFile.size() == 0) {
+      LOG_ERR("OPDS", "Downloaded file is empty or unreadable: %s", filename.c_str());
+      if (downloadedFile) {
+        downloadedFile.close();
+      }
+      Storage.remove(filename.c_str());
+      state = BrowserState::ERROR;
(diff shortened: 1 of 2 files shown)


znelson (approved):
It sounds like this fixes something, but I still can't quite figure out what. Is a zero-length file used as a sentinel for some workflow?

src/activities/browser/OpdsBookBrowserActivity.cpp
```diff
@@ -333,6 +333,20 @@ void OpdsBookBrowserActivity::downloadBook(const OpdsEntry& book) {
       });
 
   if (result == HttpDownloader::OK) {
+    FsFile downloadedFile;
+    if (!Storage.openFileForRead("OPDS", filename, downloadedFile) || downloadedFile.size() == 0) {
```
Uri-Tauber, line 337: This re-introduces the same bug this PR meant to fix, just at a higher layer. A legitimate zero-byte file downloaded from WebDAV will pass `HttpDownloader::OK` and then get silently deleted here because `size() == 0`.

Also, conflating "unreadable" and "empty" into a single `||` condition conflates two very different situations — an unreadable file is a real error, an empty file may be intentional.

Unless I completely missed something here, this post-download size check should be removed.

Uri-Tauber (changes requested):

src/activities/browser/OpdsBookBrowserActivity.cpp
znelson: My interpretation is that some other use of HttpDownloader needs to allow 0 byte files, but OdpsBookBrowserActivity still wants to consider them failures.

It would be great to hear from @jpirnay or @mmajewsk with specifics about which workflow this unblocks.

PR #1503, opened 2026-03-25
chore: update Dutch translations

basvdploeg:
## Summary

Added new Dutch translations
New Dutch translations

## Additional Context

  specific areas to focus on).

---

### AI Usage


lib/I18n/translations/dutch.yaml
```diff
@@ -66,6 +66,10 @@ STR_SLEEP_COVER_MODE: "Slaapscherm omslag-modus"
 STR_HIDE_BATTERY: "Batterij % verbergen"
 STR_EXTRA_SPACING: "Extra regelafstand alinea"
 STR_TEXT_AA: "Tekst Anti-Aliasing"
+STR_IMAGES: "Afbeeldingen"
+STR_IMAGES_DISPLAY: "Weergave"
+STR_IMAGES_PLACEHOLDER: "Plaatshouder"
```
veypatch: "Placeholder" is probably better for dutch then "Plaatshouder", which as far as I know is not even a real dutch word.

basvdploeg: Good point. Changed it to "Placeholder" since that makes more sense.

lib/I18n/translations/dutch.yaml
```diff
@@ -77,6 +81,7 @@ STR_SCREEN_MARGIN: "Schermmarge lezer"
 STR_PARA_ALIGNMENT: "Uitlijning alinea lezer"
 STR_HYPHENATION: "Woordafbreking"
 STR_TIME_TO_SLEEP: "Tijd tot slaapstand"
+STR_SHOW_HIDDEN_FILES: "Toon verborgen bestanden"
```
FireMotion, line 84: Grammatically, "Toon verborgen bestanden" uses the imperative, but the other translations don't.
"Verborgen bestanden tonen" would be more consistent with the rest of the translations.

PR #1506, opened 2026-03-25
chore: Introduce ptPt 

jpirnay:
## Summary

* **What is the goal of this PR?** A couple of Grammar fixes for the Brezilian variant of Portuguese and introduction of the European variant of Portuguese. 

## Additional Context

  specific areas to focus on).

---

### AI Usage

Did you use AI tools to help write this code? _**< PARTIALLY> **_ (Term translation)


ariel-lindemann:
no comments  on the translation specifics but the PR title is misleading, because `PT` is the correct code.

the language codes we use are according to [ISO 639](https://en.wikipedia.org/wiki/List_of_ISO_639_language_codes). See the i18n guide [here](https://github.com/crosspoint-reader/crosspoint-reader/blob/master/docs/i18n.md).

But since ISO 639 doesn't differentiate between the two variants, we could go with [BCP 47 language tags](https://developer.mozilla.org/en-US/docs/Glossary/BCP_47_language_tag).

In that case those would be:
+ [PT-BR] for Braziliam Portuguese
+ [PT-PT] for European Portuguese

And then best to specify that in `i18n.md`as well. 

I think it's important to keep the codes aligned to a common standard, to make it easier for contributors.

Or if you are referring to a certain standard or objective measure where the codes are like this, then please update the docs as well, because right now it doesn't align.

jpirnay:
> no comments on the translation specifics but the PR title is misleading, because `PT` is the correct code.
> 
> the language codes we use are according to [ISO 639](https://en.wikipedia.org/wiki/List_of_ISO_639_language_codes). See the i18n guide [here](https://github.com/crosspoint-reader/crosspoint-reader/blob/master/docs/i18n.md).
> 
> But since ISO 639 doesn't differentiate between the two variants, we could go with [BCP 47 language tags](https://developer.mozilla.org/en-US/docs/Glossary/BCP_47_language_tag).
> 
> In that case those would be:
> 
>     * [PT-BR] for Braziliam Portuguese
> 
>     * [PT-PT] for European Portuguese
> 
> 
> And then best to specify that in `i18n.md`as well.
> 
> I think it's important to keep the codes aligned to a common standard, to make it easier for contributors.
> 
> Or if you are referring to a certain standard or objective measure where the codes are like this, then please update the docs as well, because right now it doesn't align.

I intentionally sticked to the two letter abbreviation convention crosspoint has *for now* - I did not want to introduce another change on top of the trivial yaml edit, the full ISO compliant changes would have implied code changes, too.

jpirnay:
Additionally: there was a bug, when picking Russian or Portuguese (brazil) both of them had the same language id. This coud lead to an unexpected language after a restart.

ariel-lindemann:
@jpirnay the bug you mentioned is one introduced by this PR. there are no duplicate ids in master

The russian id should stay at 6 and the new language should get an id that is free (i think 19 is the next one). Changing the id of Russian breaks backward compatibility (exactly what you described)

ariel-lindemann:
> I intentionally sticked to the two letter abbreviation convention crosspoint has for now - I did not want to introduce another change on top of the trivial yaml edit, the full ISO compliant changes would have implied code changes, too.

Thanks for clarifying. But this confirms my point about the title. It is not that the code before was wrong but a concious compromise to support a language variant (imo we should be clear about those things and document them properly)

jpirnay:
> > I intentionally sticked to the two letter abbreviation convention crosspoint has for now - I did not want to introduce another change on top of the trivial yaml edit, the full ISO compliant changes would have implied code changes, too.
> 
> Thanks for clarifying. But this confirms my point about the title. It is not that the code before was wrong but a concious compromise to support a language variant (imo we should be clear about those things and document them properly)



> @jpirnay the bug you mentioned is one introduced by this PR. there are no duplicate ids in master
> 
> The russian id should stay at 6 and the new language should get an id that is free (i think 19 is the next one). Changing the id of Russian breaks backward compatibility (exactly what you described)

You are correct - fixed

jpirnay:
> > I intentionally sticked to the two letter abbreviation convention crosspoint has for now - I did not want to introduce another change on top of the trivial yaml edit, the full ISO compliant changes would have implied code changes, too.
> 
> Thanks for clarifying. But this confirms my point about the title. It is not that the code before was wrong but a concious compromise to support a language variant (imo we should be clear about those things and document them properly)

Title updated. 
btw: when #1408 gets merged, there is no longer a need to have "pt" and "po", you can then use "pt_PT" and "pt_BR"

ariel-lindemann:
> btw: when https://github.com/crosspoint-reader/crosspoint-reader/pull/1408 gets merged, there is no longer a need to have "pt" and "po", you can then use "pt_PT" and "pt_BR"

that's very good! I was just looking thouhg the code and thinging the same. Thank you for for implementing it

znelson (dismissed):

Uri-Tauber:
@jpirnay quick question: are you a native Portuguese speaker? I saw on your profile that you’re lived in Germany.
Happy to add new strings, as long as they’re checked by a native speaker 🙂.
Thanks!

gama20000:
I am native Portuguese (Pt-pt). If you point me in their direction, i can check them for you.

Uri-Tauber:
@gama20000 Press "Files changed" on the top:
image

and go over the two files there.
Make sure to compare with the english strings (which are in `/lib/I18n/translations/english.yaml`) to ensure every translation is accurate.

Thx.

goncaloMgomes:
@Uri-Tauber Native speaker here. I checked the PT-PT translations. They are perfect.

Uri-Tauber:
@goncaloMgomes Thanks.
Can you also check the PT-BR translations - see the changes are right?

goncaloMgomes:
@Uri-Tauber Yes they are also correct.

lib/I18n/translations/portuguese_pt.yaml
```diff
@@ -0,0 +1,292 @@
+_language_name: "Português (Portugal)"
+_language_code: "PT"
+_order: "19"
```
osteotek:
```suggestion
_order: "22"
```

itsthisjustin:
Looks like there's no conflicts that need to be fixed but sounds like once they are this should get merged?

CookieCaptainD:
Hello.

I translated to the latest version 1.4.0 - can you please check to be included in the next version of crosspoint? 
Also change the filenames to have portuguese_BR and portuguese_PT

[portuguese-PT.yaml](https://github.com/user-attachments/files/29974249/portuguese-PT.yaml)

PR #1507, opened 2026-03-26
fix: render <br> between paragraphs as a visible section break

brooksilg:
## Summary

Fix `<br>` elements used as scene/section separators between paragraphs being silently discarded, making chapter section breaks invisible on the reader.

This is a common pattern in Calibre-exported EPUBs (e.g. `<br class="calibre1"/>`) and in many commercially produced EPUBs where authors use blank lines to divide scenes within a chapter.

| Before | After | After (no extra space) |
|--------|--------|--------|
|  | | |

**Root cause**

The `<br>` handler called `startNewTextBlock()` with the previous paragraph's `BlockStyle`, producing an empty `ParsedText` block. When the following `<p>` opened, `startNewTextBlock()` detected the empty block and took the merge path (`getCombinedBlockStyle`), carrying only ~2pt of `marginTop` forward — invisible on screen.

**Fix**

Add a `fromBrElement` flag to `BlockStyle`. When the `<br>` creates its empty block, this flag is set. In `startNewTextBlock()`'s empty-block merge path, if the flag is present, one full line-height of `marginTop` is injected into the incoming paragraph's style before merging. The flag is explicitly reset in `getCombinedBlockStyle` so it cannot leak into sibling blocks.

Inline `<br>` inside a paragraph is unaffected: text content is added to the block before the next block element opens, making it non-empty and routing it through `makePages()` rather than the merge path.

**Page-boundary behaviour**

The extra `marginTop` is applied as a `currentPageNextY` increment before any line is placed. If this causes the first line of the following paragraph to overflow `viewportHeight`, `addLineToPage()` triggers a page break and resets `currentPageNextY = 0` — no blank space leaks to the top of a new page.

## Cases handled

| Case | Expected | Status |
|---|---|---|
| Standalone `<br>` between `<p>` elements | Full blank-line gap | ✅ Fixed |
| `<br class="..."/>` (Calibre style) | Full blank-line gap | ✅ Fixed |
| Multiple consecutive `<br>` | N blank lines for N breaks | ✅ Fixed |
| Inline `<br>` inside `<p>` | Line break only, no extra gap | ✅ Unaffected |
| `<br>` near page boundary | No spurious gap at top of new page | ✅ Correct by construction |
| `<br>` after a heading | Full blank-line gap | ✅ Fixed |

## Testing

Added `scripts/generate_br_section_break_epub.py` and the generated `test/epubs/test_br_section_break.epub` with 6 chapters, each covering one of the cases above with embedded PASS criteria for visual verification on device.

## Notes

> **Potential conflict with #1454** — PR #1454 (`fix: Use block-style stack to prevent margin/alignment leaking between siblings`) refactors `startNewTextBlock()` and the `<br>` handler using a new block-style stack. The two PRs touch the same code paths. If #1454 merges first, this PR will need a rebase against its approach. The fix here is conceptually compatible — the `fromBrElement` flag mechanism can be applied to whichever version of `startNewTextBlock()` is current.

---

### AI Usage

Did you use AI tools to help write this code? **Yes** — Claude was used to analyse the rendering pipeline, identify the root cause, and implement the fix. The logic and test cases were reviewed and verified manually.


itsthisjustin:
I love the intent behind this PR. @brooksilg if you can solve the merge conflicts I think this is a good candidate to get merged as soon as they are solved.

Uri-Tauber:
I solved the conflicts here and merge to `develop`. 
Thank you @brooksilg for your work here.

I had problems to push directly into this PR, so I opened new one. I'm closing this one as merged.

PR #1510, opened 2026-03-26
feat: Add basic clock support

jpirnay:
## Summary

* **What is the goal of this PR?** Adds wall-clock time support to the device. The ESP32-C3 has no battery-backed RTC, so time is lost on every sleep/power cycle. This PR introduces HalClock, a three-layer persistence strategy to maintain time across deep sleep and cold boot, along with UI to display and sync the clock
  - Clock persistence (HalClock)
    - NTP sync via HalClock::syncNtp() — centralises the SNTP logic previously duplicated in KOReaderSyncActivity
    - RTC memory (RTC_NOINIT_ATTR) — survives deep sleep; uses esp_clk_rtc_time() (LP timer) to compute elapsed sleep duration and correct the stored epoch on wake
    - NVS (flash key-value store) — survives power cycles as a fallback when RTC memory is lost (cold boot, no elapsed correction)
    - formatTime() — shared formatter with 12h/24h support, ~ prefix for approximate time, --:-- when never synced
   formatLogTime() — wall-clock timestamp for serial logs
  - New settings
    - Clock Format (System) — 24h (14:00) or 12h (2:00pm)
    - Keep Clock Alive (System) — keeps GPIO13 HIGH during deep sleep so the LP timer continues running (~3-4 mA extra draw), enabling accurate time restoration on wake
    - Clock (Status Bar) — toggle clock display in the reader status bar
    - Sync Time (System action) — connects WiFi, performs NTP sync, displays result
  - UI changes
    - Clock displayed in the header on all screens (home, settings, etc.) — shows --:-- until first NTP sync, ~ prefix when approximate
    - Optional clock in the reader status bar (toggle in Status Bar settings)
    - Serial log timestamps now include wall-clock time when available: [12345 14:30:05] instead of [12345]
 
## Additional Context

home screen
system settings
clock settings
status bar
reader
sync

---

### AI Usage

### DIFF
--- a/lib/hal/HalClock.cpp
+++ b/lib/hal/HalClock.cpp
@@ -9,6 +9,7 @@
 #include <sys/time.h>
 #include <time.h>
 
+#include <cmath>
 #include <cstdlib>
 
 // ---- RTC-memory state (survives deep sleep, not cold boot) ----------------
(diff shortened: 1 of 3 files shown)


jpirnay:
Properly deals with UTC, next step: timezone support

iandchasse:
I have a version of this in my fork for [my open-source hardware](https://www.github.com/iandchasse/de-link), which supports RTC without draining the battery. Having the clock is useful, but even when it's alive it shifts at least a second per minute when asleep. Without syncing for a day and being in deep sleep, the clock drifted ahead 20 minutes. I'm testing out a calibration which should sort of aid this, but it won't be perfect even then. See [here](https://www.youtube.com/watch?v=fZAR8WTKiSg) for a good summary.

With the X4, enabling that keep clock alive option it should be noted aggressively to the user how it drains battery, otherwise it would run out of charge in ~3 days regardless if you're using it or not. It's a real shame.

jpirnay:
> I have a version of this in my fork for [my open-source hardware](https://www.github.com/iandchasse/de-link), which supports RTC without draining the battery. Having the clock is useful, but even when it's alive it shifts at least a second per minute when asleep. Without syncing for a day and being in deep sleep, the clock drifted ahead 20 minutes. I'm testing out a calibration which should sort of aid this, but it won't be perfect even then. See [here](https://www.youtube.com/watch?v=fZAR8WTKiSg) for a good summary.
> 
> With the X4, enabling that keep clock alive option it should be noted aggressively to the user how it drains battery, otherwise it would run out of charge in ~3 days regardless if you're using it or not. It's a real shame.

I agree, this is a feature with a lot of caveats and restrictions - I did it more for feature completeness than to satisfy an urgent need of mine.

(btw I would really like to get my finger on one of your devices, are they available yet?)

iandchasse:
One suggestion I have from my branch, I added some choices to the "hide battery" option that either: replace the battery with the wall clock entirely, or swap between battery and wall clock every other page. Good compromise for taking up less room of the status bar. Only downside is looking down to expect the clock only to have to get to the next page to see it.

jpirnay:
> Without syncing for a day and being in deep sleep, the clock drifted ahead 20 minutes.

I did use the basic calibration from ``esp_clk_slowclk_cal_get`` with that I do have about 10sec drift / hour (translates to 4-5 min/day)

PR #1517, opened 2026-03-27
refactor: replace picojpeg with JPEGDEC for cover art conversion

jpirnay:
## Summary

- Removes the vendored `picojpeg` library and rewrites `JpegToBmpConverter` to use the already-present `JPEGDEC` (bitbank2) dependency
- Eliminates the redundancy of having two JPEG decoders in the firmware
- All BMP output (headers, fixed-point scaling, Atkinson/Floyd-Steinberg dithering) is identical to before — cached cover BMPs are unaffected

## Size impact

| | Before | After | Delta |
|---|---|---|---|
| Flash | 5,754,089 bytes (87.8%) | 5,744,777 bytes (87.7%) | **−9,312 bytes** |
| RAM | 95,212 bytes (29.1%) | 92,852 bytes (28.3%) | **−2,360 bytes** |

## Implementation notes

- `bmpDrawCallback` receives MCU-sized blocks from JPEGDEC (up to 16 rows × MCU-width), accumulates them into a pre-allocated `mcuBuf`, and applies the same scaling + dithering logic once each MCU row is complete
- File I/O uses a file-scope static `FsFile*` (safe in single-threaded embedded context) via JPEGDEC's open/read/seek callbacks — same pattern as `JpegToFramebufferConverter`
- Added a 52 KB free-heap guard before allocating the JPEGDEC object (~17 KB)
- `lib/picojpeg/` deleted (2,087 lines of C removed)

## Test plan

- [ ] Build compiles without warnings
- [ ] Cover art BMP cache regenerates correctly for EPUB books
- [ ] Home screen thumbnails (1-bit BMP path) render correctly
- [ ] Custom-size thumbnails (`jpegFileToBmpStreamWithSize`) render correctly


martinbrook (approved):
LGTM. And drops a bunch of code.

znelson (approved):

znelson:
Nice to get down to just one JPEG decoder

itsthisjustin:
I'm not 100% sure if it's this commit but as of me pulling in the latest a book that always failed cover art but did at least show fallback now shows as a blank image in both sleep and home screen cover

znelson:
@itsthisjustin Oh, shoot. It almost certainly is. Thanks for the heads up. It was a bad call on my part to merge this older change without more rigorous testing. 

I'll confirm locally now and, if it's not something obvious, revert this one.

znelson:
@itsthisjustin I misread your message, I thought you said a book whose cover _did_ work before was _failing_ now. Reading the actual symptoms, that sounds like a regression but less critical to revert immediately.

Is there anything interesting you can discern from serial logs? Or could we connect to share the file and I can try to investigate?

itsthisjustin:
> @itsthisjustin I misread your message, I thought you said a book whose cover _did_ work before was _failing_ now. Reading the actual symptoms, that sounds like a regression but less critical to revert immediately.
> 
> Is there anything interesting you can discern from serial logs? Or could we connect to share the file and I can try to investigate?

Well the critical part for me is that now the cover is an empty box istead of the fallback art which means the sleep screen is just blank instead of the fallback sleep screen. Makes it look super broken or even powered off.

Aside from that I'm getting super bad ghosting on my local build now as well. I think we regressed somehow on the screen too (not sure if related to this)

znelson:
@itsthisjustin understood. Not the clear and obvious win it appeared to be. Revert is at #1689.

itsthisjustin:
> @itsthisjustin understood. Not the clear and obvious win it appeared to be. Revert is at #1689.

I'll check it and validate that's definitely the fix. All I know is I pulled upstream and it immediately was obvious something had broken

znelson:
Hey @itsthisjustin what are your thoughts a few days later?

PR #1520, opened 2026-03-28
fix: Honor CSS text-align

jpirnay:
## Summary

* **What is the goal of this PR?** Respects explicit text-align css information, eg centered

## Additional Context

* When a global alignment setting (e.g. Justify, Left) was set, it unconditionally overrides the CSS text-align for all non-header block elements (<p>, <div>, etc.).
* The included fix now restores the CSS text-align for non-header block elements when the book's embedded stylesheet explicitly defines it — the same pattern already used for headers

---

### AI Usage


osteotek (approved):

itsthisjustin:
Seems like this was approved but then just never got merged and is now in conflict. @jpirnay is this still a good fix and worthy of merge? I'll handle your conflicts for you so you don't need to jump back into your old commits that got ignored and went stale (no shade to maintainers, there's a LOT to handle here, I just feel sorry for folks like @jpirnay who worked really hard early on and then got frustrated by slow reviews)

PR #1540, opened 2026-03-30
refactor: Refactor drawArc / fillArc for faster execution

jpirnay:
## Summary

* **What is the goal of this PR?** Replace the o(r^2) routines with a o(r) scanline logic - will make fillArc roughly 50% faster and drawArc roughly 5x faster. Still probably unnoticeable.

## Additional Context

---

### AI Usage

### DIFF
--- a/lib/GfxRenderer/GfxRenderer.cpp
+++ b/lib/GfxRenderer/GfxRenderer.cpp
@@ -354,7 +354,9 @@ void GfxRenderer::drawArc(const int maxRadius, const int cx, const int cy, const
     while (xOuter > 0 && (xOuter * xOuter + dy * dy) > outerRadiusSq) {
       --xOuter;
     }
-    while (xInner > 0 && (xInner * xInner + dy * dy) > innerRadiusSq) {
+    // Keep the smallest x that still lies outside/at the inner radius,
+    // i.e. (x^2 + y^2) >= innerRadiusSq.
+    while (xInner > 0 && ((xInner - 1) * (xInner - 1) + dy * dy) >= innerRadiusSq) {
       --xInner;
     }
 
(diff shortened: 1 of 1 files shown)


lib/GfxRenderer/GfxRenderer.cpp
```diff
@@ -472,15 +490,39 @@ void GfxRenderer::fillRectDither(const int x, const int y, const int width, cons
...
-      const int distSq = dx * dx + dy * dy;
-      const int px = cx + xDir * dx;
-      const int py = cy + yDir * dy;
-      if (distSq <= radiusSq) {
-        drawPixelDither<color>(px, py);
-      }
+    while (x > 0 && (x * x + dy * dy) > radiusSq) {
+      --x;
+    }
+    if (x < 0) break;
```
znelson, line 507: Minor note to clean up later: x is initialized to maxRadius (positive after the early return) and the while (x > 0 && ...) loop guarantees x >= 0, so this branch can never execute.

lib/GfxRenderer/GfxRenderer.cpp
```diff
@@ -343,17 +338,40 @@ void GfxRenderer::drawArc(const int maxRadius, const int cx, const int cy, const
...
+      continue;
+    }
+
+    const int x0 = cx + xDir * xInner;
+    const int x1 = cx + xDir * xOuter;
+    const int left = std::min(x0, x1);
+    const int width = std::abs(x1 - x0) + 1;
+    const int py = cy + yDir * dy;
+
+    if (width > 0) {
```
znelson, line 373: Minor note to clean up later: width = std::abs(x1 - x0) + 1 is always >= 1, so this check can never fail.

znelson (approved):

osteotek (approved):

PR #1546, opened 2026-03-31
refactor: logPrintf and predefined log level strings

CSCMe:
## Summary

* More consistent string formatting in logPrintf
* Moved [ and ] from log level strings into new format string
* Behaviour change: Early exit if user string format fails

## Additional Context

* Should not have performance implications, debug monitor etc work as before
* Clamp may be unnecessary due to information snprintf currently never being able to exceed max buffer length, but not having it would bug me

---

### AI Usage

Did you use AI tools to help write this code? _**PARTIALLY, to reason about correctness**_

### DIFF
--- a/lib/Logging/Logging.cpp
+++ b/lib/Logging/Logging.cpp
@@ -38,37 +38,26 @@ void logPrintf(const char* level, const char* origin, const char* format, ...) {
   va_start(args, format);
   char buf[MAX_ENTRY_LEN];
   char* c = buf;
-  // add the timestamp
+  // add timestamp, level and origin
   {
     unsigned long ms = millis();
-    int len = snprintf(c, sizeof(buf), "[%lu] ", ms);
+    int len = snprintf(c, sizeof(buf), "[%lu] [%s] [%s] ", ms, level, origin);
+    // erro while writing => return
     if (len < 0) {
-      return;  // encoding error, skip logging
+      va_end(args);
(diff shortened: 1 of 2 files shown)


jpirnay (changes requested):
This will break the scripts\debugging_monitor.py script

CSCMe:
@jpirnay  Could you please elaborate on what breaks? It appeared to work fine when I tested it but it seems like I missed something

jpirnay:
> @jpirnay Could you please elaborate on what breaks? It appeared to work fine when I tested it but it seems like I missed something

I was mistaken, apologies for the confusion

lib/Logging/Logging.cpp
```diff
@@ -38,37 +38,26 @@ void logPrintf(const char* level, const char* origin, const char* format, ...) {
   va_start(args, format);
   char buf[MAX_ENTRY_LEN];
   char* c = buf;
-  // add the timestamp
+  // add timestamp, level and origin
   {
     unsigned long ms = millis();
-    int len = snprintf(c, sizeof(buf), "[%lu] ", ms);
+    int len = snprintf(c, sizeof(buf), "[%lu] [%s] [%s] ", ms, level, origin);
+    // erro while writing => return
```
znelson, line 45: small typo here
```suggestion
    // error while writing => return
```

PR #1558, opened 2026-04-03
fix: generate paragraph-level XPath for KOSync push

rubalo:
## Summary
- When pushing progress to KOSync, CrossPoint generated `/body/DocFragment[N]/body` — KOReader would jump to the **start** of the chapter instead of the actual position
- Now uses the section element map (cache v19+) to reverse-lookup the first paragraph on the current page, producing `/body/DocFragment[N]/body/p[X]`
- Falls back to the chapter-level XPath when element map is unavailable (page 0 or older cache)
- The reverse direction (KOReader → CrossPoint) already parses `p[N]` via `parseXPathElement`, so both directions now have paragraph-level precision


jpirnay:
Duplicate of #1217

rubalo:
Sorry, I wanted to push these on my fork !

PR #1560, opened 2026-04-03
fix: long-press back navigates to book's directory

### LINKED ISSUE #1117: Long-pressing back button takes you to home directory instead of the open book’s directory

Cache8063:
## Summary

Long-pressing the back button while reading now navigates to the book's parent directory instead of root.

Routes long-press BACK through the existing `goToLibrary()` helper (which uses `extractFolderPath()`) instead of passing the full book path directly to `goToFileBrowser()`. The file browser was receiving a file path and failing to open it as a directory.

Fixes #1117
Supersedes #1349

## Change

One-line fix in each of the three reader activities (`EpubReaderActivity`, `TxtReaderActivity`, `XtcReaderActivity`): replace `activityManager.goToFileBrowser(path)` with `goToLibrary(path)`.

## Test plan

- [ ] Open a book in a subdirectory (e.g., `/Books/SciFi/MyBook.epub`)
- [ ] Long-press the back button
- [ ] File browser opens showing the book's parent directory contents (not root)
- [ ] Works for EPUB, TXT, and XTC file types

---

### AI Usage

### DIFF
--- a/src/activities/reader/EpubReaderActivity.cpp
+++ b/src/activities/reader/EpubReaderActivity.cpp
@@ -161,7 +161,7 @@ void EpubReaderActivity::loop() {
 
   // Long press BACK (1s+) goes to file selection
   if (mappedInput.isPressed(MappedInputManager::Button::Back) && mappedInput.getHeldTime() >= ReaderUtils::GO_HOME_MS) {
-    activityManager.goToFileBrowser(epub ? epub->getPath() : "");
+    goToLibrary(epub ? epub->getPath() : "");
     return;
   }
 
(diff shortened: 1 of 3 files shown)


Uri-Tauber:
Closing this PR since I can't reproduce the bug. Also, `goToLibrary` is just a thin wrapper around `goToFileBrowser`, so I don't see how this change helps.

PR #1214, opened 2026-02-27
feat: OPDS Browser - Allow Saving In Author Folders

jb567:
## Summary

* Implements optional save location for OPDS files to be based on author name
* Adds a new setting to control whether or not to save to the author folder

## Additional Context

* This is my first C++ open source contribution, so I'd appreciate any feedback. I expect there to be issues, especially style issues
*  I tested this against Calibre-Web OPDS, and it seemed to work well

---

### AI Usage

While CrossPoint doesn't have restrictions on AI tools in contributing, please be transparent about their usage as it 
helps set the right context for reviewers.

Did you use AI tools to help write this code?  No

PR #1562, opened 2026-04-04
feat: Use author name as directory to store books when using the OPDS…

chamaeleon:
## Summary

* **What is the goal of this PR?** Store books in a directory using the author's name when using the OPDS browser
* **What changes are included?** The `OpdsBookBrowserActivity::downloadBook()` function was modified to create directory based on the full path (currently `/author/title.epub`) by taking using a substring that ends at the last `/` before downloading the file.

## Additional Context

* The code unconditionally uses the author name as directory if available (if not, the root directory is used using only the title, like before). It is possible users may not want this behavior, and a setting might be desired to toggle the behavior on or off.
* Deleting all books in an author folder leaves an empty folder that can't currently be deleted by using the device only. This is a bit unsatisfactory. Either the directory should be deleted automatically, or the delete action should be expanded to handled directories as well.

---

### AI Usage


chamaeleon:
Looks like the only usage of `sanitizeFilename()` has been in the the code I modified for saving a OPDS download. Given the return value of `"book"` if the sanitizing results in an empty string, doesn't work since I needed to sanitize author and title separately, rather as a single combined string, to retain the `/` directory separator between then. If it was only up to me, I'd change that function to return an empty string if that is the result of sanitizing.

houstoncelluloidfilms:
Literally just logged on to suggest this feature in discussion!  The randomly throwing the files on top level is annoying! Thanks

chamaeleon:
> Literally just logged on to suggest this feature in discussion! The randomly throwing the files on top level is annoying! Thanks

I have a small change that also removes the directory when the last file in it is deleted. That removal excludes the root directory (which would presumably fail to be removed, but never hurts to be careful), `.crosspoint`, `.sleep`, and `sleep`. As I'm very new to the code base, and the whole ESP32 ecosystem, I didn't want to just include that right now in this PR without more testing and making sure there's nothing else that should at all costs be excluded.

znelson:
This looks like a good change, but the two @coderabbitai comments sound correct, at least worth following up on. Can you take a look before we merge this?

znelson (dismissed):
My mistake, I missed the intermediate commits. LGTM

chamaeleon:
@znelson

Thank you for your feedback! I changed `LOG_DBG()` to `LOG_ERR()` which makes sense.

src/activities/browser/OpdsBookBrowserActivity.cpp
```diff
@@ -316,12 +316,21 @@ void OpdsBookBrowserActivity::downloadBook(const OpdsEntry& book) {
...
 
-  // Create sanitized filename: "Title - Author.epub" or just "Title.epub" if no author
-  std::string baseName = book.title;
-  if (!book.author.empty()) {
-    baseName += " - " + book.author;
+  // Create sanitized filename: "Author/Title.epub" or just "Title.epub" if no author
+  std::string author = StringUtils::sanitizeFilename(book.author);
+  std::string title = StringUtils::sanitizeFilename(book.title);
+
+  if (title == "") title = "book";
```
Uri-Tauber, line 277: If a user downloads multiple books with invalid titles by the same author (or no author), they will all attempt to save as `/book.epub` or `/Author/book.epub`. The code does not check if the file already exists, meaning it will overwrite the previous book.

chamaeleon: I could envision using a counter and appending it to the title as long as an existing file is found with that name, incrementing the counter until the filename becomes unique.

Uri-Tauber: That would be a good solution.

chamaeleon: Something along the lines of (untested, but compiles (ship it!)) code below. Static function, not a member of the `OpdsBookBrowserActivity` class (but could be made one of course)
```cpp
static std::string filePath(const std::string& author, const std::string& title, int collisionCounter) {
  std::string titleWithCounter = title + (collisionCounter > 0 ? std::to_string(collisionCounter) : "");

  return "/" + (author != "" ? (author + "/" + titleWithCounter) : titleWithCounter) + ".epub";
}
```
and
```cpp
...
  if (title == "") title = "book";

  if (author != "") {
    if (!Storage.exists(author.c_str()) && !Storage.mkdir(author.c_str())) {
      LOG_ERR("OPDS", "Failed to create directory '%s'!\n", author.c_str());
      author = "";
    }
  }

  int collisionCounter = 0;
  std::string filename = filePath(author, title, collisionCounter);

  while (Storage.exists(filename.c_str())) {
    filename = filePath(author, title, ++collisionCounter);
  }
...
```
Maybe there's a neater way of handling it. Anyway... Not sure how far everyone would want to take this

Uri-Tauber: Maybe a better UX would be to fall back to the file’s actual name if we can’t extract the book title from the metadata. What do you think?

chamaeleon: Since we're talking about OPDS here, there's not necessarily a filename available to the client, I would think. The server is a black box we know nothing much about and we need to rely on the OPDS protocol. As far as the client is concerned, the server could be using a database and no individual files for books behind the scenes.

This is all we know about it, when requesting from Calibre (added a couple of debug lines to verify what was stored exactly in the OpdsEntry structture). Not sure to what extent other servers would differ with respect toid and href (that is, I'm not sure what the OPDS protocol dictates).

```
[237699] [DBG] [OPDS] OpdsEntry author [John Scalzi]
[237699] [DBG] [OPDS] OpdsEntry title [Redshirts]
[237699] [DBG] [OPDS] OpdsEntry id [urn:uuid:ad3ee106-83ca-402b-84cc-fe3cf9a0e530]
[237700] [DBG] [OPDS] OpdsEntry href [/get/epub/1812/Calibre]
```
In this case, `1812` would be a unique identifier for Calibre, but does the href necessarily look like that for any other server? I haven't tried any others.

Uri-Tauber: Forgive my ignorance, I’m not familiar with the OPDS protocol. But since we just need a unique identifier for a book without author or title, could we use the `id` field? Specifically, the part after `uuid:` might be enough, which could avoid introducing the collisionCounter logic.

What do you think?

chamaeleon: Without knowing if other servers necessarily follow the format (OPDS seems to be using Atom feed format to a large extent, but no idea how rigorous the wording is for this particular field), sure, it would be possible to strip off the starting `urn:uuid:` from the id field.

Btw, keep in mind that using simply `/book.epub` for the path was the case before I made any changes at all. All I wanted to achieve with my PR was an author directory being created to store the file in.

itsthisjustin:
I think there are two open PRs doing the same thing so one of them needs to be closed. #1214

itsthisjustin:
@chamaeleon Can you solve the merge conflicts so we can get this in?

Uri-Tauber:
@chamaeleon Something went wrong with your merge. 

image

chamaeleon:
@Uri-Tauber This PR should probably be scrapped and a new one created if it is desired. I really tried to get the PR branch into a state that it would not have that, but clearly failed miserably, due to having to resolve the merge conflict and the state of the branch before (which might not have been handled properly to accommodate another merge issue)

itsthisjustin:
Closing in favor of #2571

PR #1090, opened 2026-02-22
feat: Add push progress and sleep option

imedgar:
## Summary

* **What is the goal of this PR?** (e.g., Implements the new feature for file uploading.)
Implements feature [issue-914](https://github.com/crosspoint-reader/crosspoint-reader/issues/914).
Introduces a new option to "Push Progress & Sleep" in the reading menu, which allows the user to Push the progress and put the device to sleep without any interaction.
* **What changes are included?**

1. New menu option
2. Entry in i18n English
3. Extended KOReaderSyncActivity to handle Push only (avoiding interactive flow)

Refactored the initial implementation due to the introduction of ActivityManager model [PR #1016](https://github.com/crosspoint-reader/crosspoint-reader/pull/1016).

## Additional Context

Expected behavior:
- No KOReader credentials -> Same as in SYNC action, action does nothing
- WiFi fails/Document Hash fails/Upload fails -> pendingSubactivityExit and returns to reader
- Success -> onPushComplete(true) sets pendingSleep and enterDeepSleep

---

### AI Usage

While CrossPoint doesn't have restrictions on AI tools in contributing, please be transparent about their usage as it 
helps set the right context for reviewers.

Did you use AI tools to help write this code? _**PARTIALLY**_
Used partially to validate implementation idea + conflict solving after the introduction of [PR #1016](https://github.com/crosspoint-reader/crosspoint-reader/pull/1016).

demo

PR #1571, opened 2026-04-05
feat: KOReader auto-sync background progress tracking

lukehoersten:
## Summary

Adds background auto-sync of reading progress to KOReader-compatible sync servers, building on the existing manual **Sync Progress** feature.

- **Auto KOReader Sync** setting (Off / 10 pages / 15 pages) triggers a background sync after the configured number of page turns
- A 5-minute timer also triggers a sync if the position has moved and no page-threshold sync has fired recently
- On book exit (Back button or sleep/auto-sleep), a final sync uploads the current position; a **"Syncing progress..."** overlay is shown during this
- **Go Home** in the reader menu exits without syncing, for quick switching between books
- A sync-failure indicator appears on the status bar if WiFi is unavailable; all further auto-sync attempts are suppressed for the session so the reader is never interrupted by retry attempts. Only a successful manual sync or a sleep/wake cycle clears the failure state
- A small tick mark above the progress bar shows the last successfully synced position

## Memory / storage design notes

This is running on an ESP32-C3 with ~380KB usable RAM and no PSRAM, so care was taken to avoid any new heap allocations in the sync path:

- **No new heap allocations.** `tryAutoSync()` works entirely with stack-local variables and existing `KOReaderProgress` / `KOReaderDocumentId` structs already used by the manual sync path
- **Sync state is 8 scalar fields** (`float`, `bool`, `unsigned long`) added to the activity — negligible DRAM cost
- **Last-synced position is persisted to SD card** in a 4-byte float file (`ko_autosync.bin`) in the book's existing `.crosspoint` cache directory. This means the sync marker survives across reading sessions and device restarts without touching NVS or adding a new settings field
- **`pendingSyncFailed` is a file-scope static** (not an instance field) so it persists across book open/close cycles within a power session and resets for free on deep-sleep wake, with no SD writes needed to track it
- WiFi is connected on-demand and immediately disconnected after each sync — no persistent WiFi state is held while reading

## Test plan

- [ ] Auto-sync fires after configured page count, sync marker appears on progress bar
- [ ] 5-minute timer sync fires when page threshold hasn't triggered
- [ ] Exit sync shows "Syncing progress..." popup and uploads on Back / sleep / auto-sleep
- [ ] Go Home skips exit sync
- [ ] Sync failure indicator appears when WiFi unavailable; no further auto-sync attempts that session
- [ ] Manual Sync Progress clears the failure indicator
- [ ] `ko_autosync.bin` written to book cache dir; marker position survives book close/reopen
- [ ] No regression on manual Sync Progress flow

---

### AI Usage


itsthisjustin:
I am going to close this for now due to the risk in having this running all the time and the fact that (as far as I know) it conflicts with how KOReader itself works at least on Android. Wifi is very expensive on these devices and #1090 I think is a more elegant user demand way to do this

itsthisjustin:
Also only because it's 3am and I don't want to come across as emotionless, @lukehoersten thanks so much for the contribution. I'd super encourage you to open a discussion on this topic so we can see which version of "auto" sync the community prefers

lukehoersten:
It's off by default and has to be enabled by the user. The "auto" semantics are basically that whenever you stop reading it syncs. Where can I open a community discussion issue? I'm not familiar with that? Thanks @itsthisjustin

itsthisjustin:
Just at the top of the repo there's a discussions tab

lukehoersten:
@itsthisjustin https://github.com/crosspoint-reader/crosspoint-reader/discussions/2219 if you have any thoughts/feedback.

PR #1572, opened 2026-04-05
fix: epub images not rendering correctly on x3

itsthisjustin:
Replace hardcoded DISPLAY_WIDTH, DISPLAY_HEIGHT, and DISPLAY_WIDTH_BYTES constants with runtime values from display object to support multiple device models (X3 and X4) with different screen dimensions.

### DIFF
--- a/lib/Epub/Epub/converters/DirectPixelWriter.h
+++ b/lib/Epub/Epub/converters/DirectPixelWriter.h
@@ -15,6 +15,7 @@
 struct DirectPixelWriter {
   uint8_t* fb;
   GfxRenderer::RenderMode mode;
+  uint16_t displayWidthBytes;  // Runtime framebuffer stride (X4: 100, X3: 99)
 
   // Orientation is collapsed into a linear transform:
   //   phyX = phyXBase + x * phyXStepX + y * phyXStepY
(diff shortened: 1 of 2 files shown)


itsthisjustin:
> `79-86`: _⚠️ Potential issue_ | _🟠 Major_
> 
> **Increase `kMaxRowSize` to support X4 device screenshots.**
> 
> The buffer capacity (68 bytes) is sized only for X3's 528px display height. After the 90° rotation, `phyWidth = height`, so for X4 with `DISPLAY_HEIGHT = 800`, the BMP row size becomes 100 bytes and exceeds the buffer. Screenshots will fail on X4 with the "Row size exceeds buffer capacity" error at lines 81–86.
> 
> The runtime check safely prevents overflow but silently breaks the feature. Either confirm X4 is unsupported by screenshots, or increase `kMaxRowSize` to 100 (or dynamically allocate) to support both devices:
> 
> * 528px (X3) → 68 bytes
> * 800px (X4) → 100 bytes

This is incorrect.  Look at the rotation on lines 43-44:                                                   
                 
  int phyWidth = height;   // X4: 480, X3: 528
  int phyHeight = width;   // X4: 800, X3: 792               
                                                             
  The BMP is rotated 90 degrees. phyWidth (which determines  
  row size) is the display height, not width:                
                                                             
  - X4: phyWidth = 480 → (480+31)/32*4 = 60 bytes            
  - X3: phyWidth = 528 → (528+31)/32*4 = 68 bytes           
                                                             
  68 covers both devices. CodeRabbit confused DISPLAY_WIDTH
   (800) with phyWidth (480 after rotation).

PR #457, opened 2026-01-20
fix: prevent OOM crash when loading large EPUBs (2000+ chapters)

DChells:
## Summary

Fixes #134 - Large EPUBs with 2000+ chapters (common in webnovels) crash the device during initial indexing.

## Problem

When loading a large EPUB, `buildBookBin()` calls `loadAllFileStatSlims()` which pre-loads ALL ZIP central directory entries into a `std::unordered_map<std::string, FileStatSlim>`. For EPUBs like "Shadow Slave" (2773 entries) or "Lord of Mysteries" (1400+ entries), this exhausts ESP32-C3's ~380KB RAM and triggers `abort()`.

Serial log evidence:
```
[61679] [EBP] Loading ePub: /Books/Shadow Slave...
[567904] [BMC] Wrote 2768 spine, 2768 TOC entries
abort() was called at PC 0x420e61ad on core 0
```

## Solution

Remove the `loadAllFileStatSlims()` call. The existing `loadFileStatSlim()` function already handles individual lookups by scanning the ZIP central directory per-file when the cache is empty.

**Trade-off**: This is O(n×m) instead of O(n) for lookups, making initial indexing slower for large EPUBs. However, it prevents memory exhaustion and allows the book to load successfully. A future optimization could cache only spine item stats rather than all ZIP entries.

## Testing

- [ ] Build passes
- [ ] Normal EPUBs still load correctly
- [ ] Large EPUB (2000+ chapters) no longer crashes during initial load
- [ ] Large EPUB indexes successfully (may take several minutes)

## Notes

This addresses the TODO comment that was already in the code acknowledging this issue.

PR #1574, opened 2026-04-05
fix: epubs with more than 2000 chapters/toc items crashing

CSCMe:
## Why draft?
* I'm currently working on file-backed arrays that will a) make sure we reliably don't run out of memory while parsing (at least from causes I have identified so far)
* After that we will be able to parse large-enough epubs to for me to call this final

## Summary

* Increase the maximum supported spine size for epubs
* Initially intended to provide support for books up to 65000 entries long, however this would require a larger restructuring of the contents.opf parser
* Motivation: I have books with more than 2.5k chapter entries

* Batch-based processing of toc and spine items
* Switched to deques instead of vectors for parsing data, making use more available memory
* Added generate_mega_epub.py to generate test books
* Now supports books with 10000 chapters/toc entries, limit (not confirmed but mused to be) ~12000 entries instead of ~2500
* These limits are not guaranteed => Log message when potential OOM when parsing (not added for parsing content.opf)

## Additional Context

* The overwrite argument was added to writeToCEntry to keep the ordering of toC entries. Normal parsing behaves as before, batch parsing first writes every ToC entry in order then selectively overwrites entries with missing spine connections if found
* Removed length from itemIndexEntry because matches are later string compared, allowing us to halve the required memory for parsing items

---

### AI Usage

### DIFF
--- a/lib/Epub/Epub.cpp
+++ b/lib/Epub/Epub.cpp
@@ -396,29 +396,36 @@ bool Epub::load(const bool buildIfMissing, const bool skipLoadingCss) {
 
   // TOC Pass - try EPUB 3 nav first, fall back to NCX
   const uint32_t tocStart = millis();
+
   if (!bookMetadataCache->beginTocPass()) {
     LOG_ERR("EBP", "Could not begin writing toc pass");
     return false;
   }
 
-  bool tocParsed = false;
+  // Do a ToC pass
+  // A toc pass reads as much of the spine as fits into memory and then reads the entire ToC
+  // An initial spine read is performed in beginTocPass
+  // After every pass the tocFile position is reset so the next pass may write in positions missed by the first
+  do {
+    bool tocParsed = false;
+    // Try EPUB 3 nav document first (preferred)
+    if (!tocNavItem.empty()) {
+      LOG_DBG("EBP", "Attempting to parse EPUB 3 nav document");
+      tocParsed = parseTocNavFile();
+    }
 
-  // Try EPUB 3 nav document first (preferred)
-  if (!tocNavItem.empty()) {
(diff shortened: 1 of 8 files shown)


CSCMe:
Regarding benchmarks: The overhead from switching to deque is mostly negligible ~2 sec for large files (from 25 to 27 sec for the 1000.epub) and I didn't really notice a difference for small files
5000.epub crashes without these changes and takes ~2.5 minutes to load now
10000.epub crashes without these changes and takes ~5 minutes to load now

znelson:
> Motivation: I have books with more than 2.5k chapter entries

Can you share what type of books this is targeting? It feels like the chapter browsing experience will be pretty poor with books this long.

CSCMe:
Specifically, this is targeting my epub-converted flashcard where each individual card is a ToC entry for quick skipping.
It worked fine in 1.1 but started crashing after some commit between 1.1 and 1.2
[Karteikarten-Musik.zip](https://github.com/user-attachments/files/26601265/Karteikarten-Musik.zip)

While I do not have a legitimate book with this many chapters/ToC entries, #457 mentions Shadow Slave with 2.5k entries.
I don't have a copy of that book but I'd presume this would also crash now, given that the flashcards have a similar chapter count.

The chapter browsing experience itself wasn't the goal of this PR, just loading the book itself.
(Although I have tested it and navigating the ToC list with long presses isn't actually too bad so I'd say its fine)

CSCMe:
Alternatively to the approach in this PR, I'm currently working on switching to file-backed arrays for the epub lookup tables so we don't run out of memory even with 65k ToC/Spine entries but I can't say anything about performance yet (might be ok, might not be)
But that approach would, in theory, also not affect loading times except for currently unloadable books

Issue #1026, opened 2026-02-20
The caption below an image appears centered with Justify Paragragh and Embedded Style OFF


pablohc:
### Affected Version

1.1.0-rc+8f33bee

### Describe the Bug

When Embedded Style is disabled, the caption below an image appears centered, without respecting the justified alignment.

BLACK Embedded Style OFF ----//----- WHITE Embedded Style ON:

![Image](https://github.com/user-attachments/assets/3526963b-8ebe-421f-8af8-b1fb9922bd48)

Example 1:
![Image](https://github.com/user-attachments/assets/8e81c6c0-9433-4a6a-a151-a43aa170881a)
Extracted from ePub:
<img width="990" height="600" alt="Image" src="https://github.com/user-attachments/assets/f4773974-9b79-487b-a9dc-39364249f12e" />

Example 2:
![Image](https://github.com/user-attachments/assets/803d3d67-f9a5-4f7c-b3a0-7a919b7c976c)
Extracted from ePub:
<img width="921" height="622" alt="Image" src="https://github.com/user-attachments/assets/818c265e-c8a6-478b-b603-7076f7e6461f" />

### Steps to Reproduce

1. Go to Settings > Reader
2. Select Justify in Paragraph alignment
3. Desactivate Embedded Style
4. Open ePub with images and caption
5. Validate if caption respects the defined paragraph alignment.

### Expected Behavior

Respects the defined paragraph alignment to image caption with Embedded Style is OFF

### Relevant Log Output/Screenshots

```shell

```

Eloren1:
Looks like correct behavior, captions under images are usually in the center

pablohc:
> Looks like correct behavior, captions under images are usually in the center

Captions must respect the selected settings.

osteotek:
@pablohc what happens when Parapgraph Alignment is set to Book's Style?

pablohc:
> [@pablohc](https://github.com/pablohc) what happens when Parapgraph Alignment is set to Book's Style?

Hi @osteotek, has the same behavior:

<img width="1342" height="569" alt="Image" src="https://github.com/user-attachments/assets/e5377464-b75a-4fc9-b962-137afe90eefb" />

<img width="1380" height="752" alt="Image" src="https://github.com/user-attachments/assets/1e654506-82db-4b15-a571-3c90524183cf" />

daveallie:
Definitely a bug, image centering is bleeding into the next text block

[...]

PR #1581, opened 2026-04-06
fix: Reset block style completely when leaving element

daveallie:
## Summary

- Extends the empty text block style reset (introduced in #1026) to also clear all margins and padding, not just text alignment
- Prevents horizontal margins/padding from a closed block from bleeding into the next sibling element's empty text block

## Additional Context

Consider the following example:
```html
<div style="margin:1em 40%">
  <img/>
</div>
<p>text</p>
```

Before this fix, the `<p>` would inherit the div's 40% side margins, pushing text inward extremely far. After this fix, the div style is completely reset when leaving the element, preventing the bleed into the p element.

This does highlight an interesting issue where nested block elements with styles would leak incorrectly, but now no longer stack correctly. I think this is preferred, but happy for opinions.

e.g.

```html
<div class="c1">          <!-- marginLeft: 10px -->
  <div class="c2">        <!-- marginLeft: 20px -->
    <p class="c3">text</p>  <!-- marginLeft: 5px -->
  </div>
  <p class="c4">text2</p>   <!-- marginLeft: 5px -->
</div>
```

Expected: `c3` gets 10+20+5=35px, `c4` gets 10+5=15px.
Original code: `c2` closes with empty block (margins 30 preserved) →  `c4` merges styles → 35px. Wrong (should be 15).
This PR: `c2` closes with empty block → margins reset to 0 → c4 merges styles → 5px. Also wrong (should be 15).

The complete fix for this would be a full style stack. However this fix imo is slightly better than the old implementation. Throwing away styles feels like a better compromise rather than leaking them.

---

### AI Usage

Did you use AI tools to help write this code? Yes

### DIFF
--- a/lib/Epub/Epub/parsers/ChapterHtmlSlimParser.cpp
+++ b/lib/Epub/Epub/parsers/ChapterHtmlSlimParser.cpp
@@ -955,17 +955,29 @@ void XMLCALL ChapterHtmlSlimParser::endElement(void* userData, const XML_Char* n
     self->currentCssStyle.reset();
     self->updateEffectiveInlineStyle();
 
-    // Reset alignment on empty text blocks to prevent stale alignment from bleeding
+    // Reset block style on empty text blocks to prevent stale styling from bleeding
     // into the next sibling element. This fixes issue #1026 where an empty <h1> (default
     // Center) followed by an image-only <p> causes Center to persist through the chain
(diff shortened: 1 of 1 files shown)


daveallie:
Thinking of combining this and #1580 into a proper fix for inheriting block styles

daveallie:
Addressing in #1582

PR #1583, opened 2026-04-06
feat: force contrast filter for sleep screen based on filename

Telemaniaka:
## Summary

I have a mix of sleep screens, some of which benefit from the grayscale rendering, and some simpler ones that look washed out with it.

In order to have best of both, this change checks if the filename contains the word `contrast` and acts as if I have set the sleep screen filter "contrast"  for this particular image. 

### AI Usage


Telemaniaka:
I've been testing https://github.com/crosspoint-reader/crosspoint-reader/pull/1614 and its a big improvement in the rendering. 
No need for this silly PR.

PR #1612, opened 2026-04-08
feat: X3 clock display with DS3231 RTC and NTP sync

juicecultus:
## Summary

Enables the clock feature on the X3 device by reading the onboard DS3231 RTC and displaying the time in the reader status bar. Includes a dedicated UTC offset picker, 12h/24h format toggle, debounced NTP sync, and manual resync action.

## Changes

* **HalClock module** (`lib/hal/HalClock.h/.cpp`) — singleton that reads hours/minutes from the DS3231 over I2C with 10-second caching. Handles both 12h and 24h BCD register modes. Stores UTC on the RTC.
* **NTP sync** — automatically syncs the RTC from `pool.ntp.org` / `time.nist.gov` on the first WiFi connection only (in `WifiSelectionActivity`). Subsequent connects skip the sync (the DS3231 drifts ~2 ppm so one initial sync is enough), keeping WiFi connect latency unchanged on every-day use.
* **Manual sync action** — new `ClockSyncActivity` reachable from Settings > Customise Status Bar > "Sync clock now" (X3 only). Provides explicit user-initiated resync after crossing time zones or if accuracy is in doubt. Resetting the `clockHasBeenSynced` flag from the web UI also forces a re-sync on next WiFi connect.
* **Status bar rendering** — displays `HH:MM` (or `H:MM AM`/`HH:MM PM`) right-aligned next to the page count in `BaseTheme::drawStatusBar()`. Properly reserves space so the centered title doesn't overlap.
* **Settings** under Customise Status Bar (X3 only, hidden entirely on X4):
  * **Clock** (toggle) — show/hide the clock display
  * **Clock Format** (24-hour / 12-hour)
  * **Clock UTC Offset** — launches a dedicated `ClockOffsetActivity` (see below)
  * **Sync clock now** — launches `ClockSyncActivity`
* **ClockOffsetActivity** — replaces the previous linear cycler (52 entries) with a three-field picker (sign / hours / minutes) modelled on @znelson's mock. Confirm cycles fields, Up/Down adjust the active one, with a live wall-clock preview underneath. Quarter-hour granularity (15-min steps) supports oddball zones like Nepal (UTC+5:45), India (UTC+5:30), and Chatham (UTC+12:45).
* **Timezone handling** — RTC always stores UTC; the offset is applied at display time only, so the RTC value is never ambiguous.

## Device scope

**X3 only.** All clock UI is gated on `halClock.isAvailable()` (DS3231 presence). On X4 devices the clock entries are not even present in the Customise Status Bar menu, so there are no dead toggles or hidden state to worry about.

## Storage / migration note

The offset field was renamed `clockUtcOffset` -> `clockUtcOffsetQ` (quarter-hour steps, biased by 48, range 0..104) to support 15-min granularity. The JSON key rename means any persisted half-hour values from earlier test builds are dropped cleanly and the default UTC+0 is used. Users who had set a custom offset on the old builds will need to set it once more.

## Testing

* Verified on X3 hardware: NTP sync sets correct UTC time on first WiFi connect, debouncing skips it on subsequent connects, manual "Sync clock now" forces a fresh sync, UTC offset activity supports all 15-min steps, 12h/24h toggle works, clock persists across reboots.
* X4 build verified: clock entries do not appear in the Customise Status Bar menu.

## Outstanding items (out of scope for this PR)

* Showing the clock on the home screen and other non-reader views (per @leecming82's suggestion)
* README / USER_GUIDE updates — to land alongside the merged release per @znelson's note

## AI Disclosure

Yes, AI was used to assist with the development of these changes.


lib/hal/HalClock.cpp
```diff
@@ -0,0 +1,162 @@
...
+    hour = _cachedHour;
+    minute = _cachedMinute;
+    return true;  // return stale data on error
+  }
+  Wire.requestFrom(I2C_ADDR_DS3231, (uint8_t)3);
+  if (Wire.available() < 3) {
+    _lastPollMs = now;
+    hour = _cachedHour;
+    minute = _cachedMinute;
+    return true;
```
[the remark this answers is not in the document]
juicecultus: fixed, _hasCachedTime guard - getTime() now returns false on I2C error if no successful read has ever occurred

juicecultus:
Addressed all CodeRabbit feedback in c638af7

wildfire070:
Just flashed and tested for the X3 from your site. Works great! I see that there's a 12h and 24h mode with this PR, but I can't seem to change to 12h in settings after flashing to my X3?

mcrosson:
Is there anything preventing this from being used in an "NTP Only" fashion?

leecming82:
~~I opened #1882 to continue this work against current master, preserving the original commits and adding master merge and follow-up fixes.~~

juicecultus:
> Just flashed and tested for the X3 from your site. Works great! I see that there's a 12h and 24h mode with this PR, but I can't seem to change to 12h in settings after flashing to my X3?

@wildfire070 Great catch. The underlying RTC code reads both 12h and 24h BCD modes, but I never wired up a user-facing format toggle. Just pushed a commit adding a "Clock Format" entry under Customise Status Bar with 12-hour / 24-hour options.

juicecultus:
> Is there anything preventing this from being used in an "NTP Only" fashion?

@mcrosson Nope. DS3231 is what makes the clock feel "set and forget" as it persists across cold boots, full discharges, and firmware flashes via its coin cell. Without it, every reboot would either show stale time (relying on the ESP32-C3's internal RTC, which only survives deep sleep, not power-off and drifts ~1 min/day) or require WiFi-up + a 2–3s NTP handshake on every boot. That's a real battery hit on a device whose primary mode is offline reading. An "NTP at boot if WiFi configured" opt-in for X4 would be a reasonable follow-up, but maybe it warrants its own PR rather than expanding this one's scope?

juicecultus:
> I opened #1882 to continue this work against current master, preserving the original commits and adding master merge and follow-up fixes.

@leecming82 Thanks for picking this up while I was away! This PR wasn't delayed in being merged due to it being stale but rather due to more foundational considerations from the reviewers - as to impact on device battery, etc. I've now rebased [#1612](https://github.com/juicecultus/crosspoint-reader/issues/1612) onto current master, resolved conflicts, and addressed open feedback (12h/24h toggle for @wildfire070). Since [#1612](https://github.com/juicecultus/crosspoint-reader/issues/1612) is the original branch with the full review history and is now in the same state as [#1882](https://github.com/juicecultus/crosspoint-reader/issues/1882), would you mind closing [#1882](https://github.com/juicecultus/crosspoint-reader/issues/1882) in favour of this one? Happy to cherry-pick anything additional you added there if I missed it.

juicecultus:
@znelson can this PR be included in the next release?

itsthisjustin:
> @znelson can this PR be included in the next release?

Its a VERY hotly requested feature and a reason users stay on stock currently

znelson:
@itsthisjustin Can you please test on device and provide review feedback on the changes here?

Befbo:
Does this affect battery life at all?

juicecultus:
> Does this affect battery life at all?

Have been running this for over a month (I keep a build at https://einkhub.com) with next to zero impact on battery as the NTP only fires once at first setting switch on to set the time straight and then the time is kept on device.

itsthisjustin:
Working as intended. I did have a book suddenly jump to end of book and now every time I open it it does this...but other books are fine so I'm not sure if that's related to this branch and I can't imagine how it would be related

itsthisjustin:
@znelson a point of readme, i had no idea how to activate this feature and had to read the PR notes. Worth mentioning how x3 users can use the clock and calling out that x4 users wont get it

leecming82:
Maybe not necessary for the upcoming RC but would be good to show the clock in the home screen and other places outside of the reader view

znelson:
> @znelson a point of readme, i had no idea how to activate this feature and had to read the PR notes. Worth mentioning how x3 users can use the clock and calling out that x4 users wont get it

@itsthisjustin If we put it in the README.md today, people will look for it in the latest release build (1.2.0). The PR notes are where you'd look to figure out how a proposed change works, right?

I agree when it's in a release we should document it in the readme. 👍

itsthisjustin:
> > @znelson a point of readme, i had no idea how to activate this feature and had to read the PR notes. Worth mentioning how x3 users can use the clock and calling out that x4 users wont get it
> 
> @itsthisjustin If we put it in the README.md today, people will look for it in the latest release build (1.2.0). The PR notes are where you'd look to figure out how a proposed change works, right?
> 
> I agree when it's in a release we should document it in the readme. 👍

Ha yes that's what I meant. Do we have a pr for readme updates that will fold in with the release?

pablohc:
I would still recommend that every PR includes an update to either the README or the USER_GUIDE, reflecting the improvements implemented in each PR. This ensures that responsibility for keeping these documents up to date is shared among us.

itsthisjustin:
> I would still recommend that every PR includes an update to either the README or the USER_GUIDE, reflecting the improvements implemented in each PR. This ensures that responsibility for keeping these documents up to date is shared among us.

Every is a bit much but feature prs maybe

src/activities/network/WifiSelectionActivity.cpp
```diff
@@ -248,6 +249,9 @@ void WifiSelectionActivity::checkConnectionStatus() {
     connectedIP = ipStr;
     autoConnecting = false;
 
+    // Sync RTC from NTP if available (blocks up to ~5s)
+    halClock.syncFromNTP();
```
znelson: I'm a little concerned about potentially blocking for up to 5s for NTP sync on every wifi connection. On the X3, at least, the time should be quite accurate after it has synced once, right? Should syncing time be a user-initiated action?

src/activities/settings/StatusBarSettingsActivity.cpp
```diff
@@ -11,14 +12,35 @@
...
-                                     StrId::STR_XTC_STATUS_BAR};
+                                     StrId::STR_XTC_STATUS_BAR,
+                                     StrId::STR_CLOCK,
+                                     StrId::STR_CLOCK_UTC_OFFSET,
+                                     StrId::STR_CLOCK_FORMAT};
+
+constexpr int CLOCK_FORMAT_ITEMS = 2;
+const StrId clockFormatNames[CLOCK_FORMAT_ITEMS] = {StrId::STR_CLOCK_FORMAT_24H, StrId::STR_CLOCK_FORMAT_12H};
+
+// UTC offset range: 0 = UTC-12:00, 24 = UTC+0, 52 = UTC+14:00 (half-hour steps)
```
znelson: Nepal is UTC+5:45. I think we need to represent the offset in a more flexible format to allow for oddball timezones.

src/activities/settings/StatusBarSettingsActivity.cpp
```diff
@@ -11,14 +12,35 @@
...
                                      StrId::STR_BOOK_PROGRESS_PERCENTAGE,
                                      StrId::STR_PROGRESS_BAR,
                                      StrId::STR_PROGRESS_BAR_THICKNESS,
                                      StrId::STR_TITLE,
                                      StrId::STR_BATTERY,
-                                     StrId::STR_XTC_STATUS_BAR};
+                                     StrId::STR_XTC_STATUS_BAR,
+                                     StrId::STR_CLOCK,
+                                     StrId::STR_CLOCK_UTC_OFFSET,
+                                     StrId::STR_CLOCK_FORMAT};
```
znelson: Are these options shown on X4?

src/activities/settings/StatusBarSettingsActivity.cpp
```diff
@@ -121,6 +151,22 @@ void StatusBarSettingsActivity::handleSelection() {
...
+    // Show Clock (X3 only — DS3231 RTC required)
+    if (!halClock.isAvailable()) return;
+    SETTINGS.statusBarClock = (SETTINGS.statusBarClock + 1) % 2;
+  } else if (selectedIndex == 8) {
+    // UTC Offset (cycle in half-hour steps)
+    if (!halClock.isAvailable()) return;
+    if (SETTINGS.clockUtcOffset >= UTC_OFFSET_MAX) {
+      SETTINGS.clockUtcOffset = UTC_OFFSET_MIN;
+    } else {
+      SETTINGS.clockUtcOffset++;
```
znelson: This is a pretty rough experience, flipping through 52 offsets linearly. One extra click and it's a long way back to the offset you wanted.

It feels like we need an activity to select the offset, with three fields:

```
┌──────────────────────────────┐
│   Clock UTC Offset           │
│                              │
│       UTC +  5 : 45          │
│           ^ ^^   ^^          │
│                              │
│  ◄ Back  ● Toggle  ▲▼ Adj    │
└──────────────────────────────┘
```

- `+ / -`
- Hour: 0-14 positive, 0-12 negative
- Minute: 0, 15, 30, 45 (quarter hour steps)

itsthisjustin: Yeah I was going to comment the same thing but honestly it's not like you change it often. Felt like a silly thing to over index on.

juicecultus: This is non-blocking - see the Changes section of the PR: “NTP sync automatically syncs the RTC from pool.ntp.org / time.nist.gov when WiFi connects (in WifiSelectionActivity). Takes ~2-3s”. If user has never connected to WIFI the time shown will be what it was stored on it before the flash but if users has Stock firmware that’s already updated and stored.

src/activities/settings/StatusBarSettingsActivity.cpp
juicecultus: Agreed, the offset was always a little hacky - will redo as activity and recommit.

itsthisjustin:
@juicecultus let me know when you want this reviewed again and I'll test

juicecultus:
@itsthisjustin Almost certainly unrelated to this PR. Nothing in here touches reading position, EPUB parsing, or progress state, the only files touched are HalClock, the status bar renderer, and a new ClockOffsetActivity. If you can reproduce reliably with a specific book, please open it as a separate issue with the file (or its hash) so it can be repro'd against master and bisected to the real culprit.

Worth also clearing `.crosspoint/` for that book on the SD card first in case the cache got corrupted by an unrelated abort.

juicecultus:
@itsthisjustin Ready for another pass, thanks! Three new things to try on device since you last looked:

1. **Customise Status Bar > Clock UTC Offset** now launches a dedicated picker with three editable fields (sign, hours, minutes). Confirm cycles fields, Up/Down adjust. Supports 15-minute steps so zones like UTC+5:45 work.
2. **Customise Status Bar > Clock Format** toggles between 24-hour and 12-hour (AM/PM).
3. **Customise Status Bar > Sync clock now** forces a manual NTP resync. Auto-sync on WiFi connect now only fires the first time, so reconnects no longer add latency.

On X4 the Clock, Clock Format, Clock UTC Offset, and Sync clock now entries are hidden entirely (not just inert), so should also verify nothing appears there.

@znelson this should now address all your inline feedback (offset picker UX, 15-min granularity, X4 visibility, blocking NTP sync on every connect). Flagging for review when you have a moment.

itsthisjustin (approved):

itsthisjustin (approved):

Issue #1595, opened 2026-04-07
Screenshot shortcut also act as next chapter shortcut


Naia-love:
### Affected Version

v1.2.0

### Describe the Bug

When I do the screenshot shortcut (which is not indicated anywhere and not customisable, I just assumed it. Customisation would be nice, I'd like up and down at the same time instead of power and down):
- the screenshot animation plays
- the screen reset (is this a bug too that screen reset after *all* screenshot?)
- the player skips to next chapter like If I used the "long press on next" shortcut (despite not pressing next long)

### Steps to Reproduce

1. takes a screenshot via the power + "volume down" shortcut

### Expected Behavior

to only take a screenshot. For now I have to do with the menu screenshot button, which is pretty slow, but can be faster than having to get back to x pages after a skip.
also clear indication of those shorcut somewhere and/or customisation of it would be appriecated (like. was the power + down shortcut already existing before 1.2.0 but only to skips chapters ? If it dId, I did not know)

### Relevant Log Output/Screenshots

```shell

```

Mraulio:
Can confirm the same behavior, was about to write an Issue about it

Capelare:
I can confirm the issue as well.

Mraulio:
Made a fix here https://github.com/crosspoint-reader/crosspoint-reader/pull/1625, works fine for me and doesn't break the other Power button functions

Mraulio:
You can close this issue, my fix was merged on https://github.com/crosspoint-reader/crosspoint-reader/commit/5c12f2f01ea744e5e453d5237e4e90244614202a

PR #1625, opened 2026-04-09
fix: avoid skipping chapter after screenshot

Mraulio:
## Summary

Fixes skipping chapter when it's enabled in Settings and you take a Screenshot.
A simple return if Power and Down were released before the skipChapter.

## Additional Context

* There is an Issue related #1595.
* Checked that it kept the Power button funcionality for Next page and Suspend.

---

### AI Usage


Uri-Tauber (approved):
Neat! Works on my device.

PR #1626, opened 2026-04-09
feat: enable manual screen refresh on power button short press

### LINKED ISSUE #550: Feature Request: Add refresh screen on power button short press
After seeing multiple issues on reddit with fading screens, it would be great if we can add a option to the power button short press, where it refreshes the screen.

I got the idea from release 0.13 where there was an addition
"Add page turn on power button press #286."

bdeshi:
## Summary

Adds an option to allow manual screen refresh on power button short press.

## Additional Context

there's an option to refresh the screen after a set number of pages. but sometimes a manual refresh is needed to clear up stale ink pixels. this works everywhere both in and out of reading mode.

resolves #550 

---

### AI Usage

Did you use AI tools to help write this code? _**PARTIALLY**_, to understand the project structure.


Uri-Tauber (approved):
Code looks fine, and tested on my device.

mcrosson (approved):
lgtm

mcrosson:
@Uri-Tauber Can you merge this, I'm not sure of project norms as a reviewer and don't want to merge wrong

Uri-Tauber:
@mcrosson I believe we both have the same permissions here.

I’m not entirely sure myself what the guidelines are around merging and timing, so I’ve just been trying to use my best judgment. At the end of the day, we’re all volunteers—no one’s going to hold it against you if something isn’t perfect. Worst case, we can always roll back to a previous commit.

lukestein:
> @mcrosson I believe we both have the same permissions here.

@Uri-Tauber, we're all figuring it out, best I can tell ;)

My view is that if (1) something has multiple maintainer review approvals, (2) the comments suggest that it's been tested on device and the code has been reviewed, (3) it shouldn't add much to flash, which perhaps remains constrained, and (4) it's clearly in scope, simple enough not to break much stuff, and there don't seem to be competing implementations or other sources of likely controversy… go for it.

PR #1627, opened 2026-04-09
fix: Wild pointer crash in JPEGDEC MCU_SKIP handling

itsthisjustin:
Adds a PlatformIO pre-build script to patch JPEGDEC library. When decoding progressive JPEGs with AC coefficients, MCU_SKIP (-8) causes array index 0xFFFFF8, creating a wild pointer ~33MB past the sMCUs array. The patch redirects pMCU to sMCUs[0] when MCU_SKIP is active, preventing store-access faults while maintaining correct behavior for JPEG_SCALE_EIGHTH decoding. Devices with larger framebuffers (like the x3) (792×528 = 52,272 bytes vs 800×480 = 48,000 bytes) have less free heap, shifting the allocation and changing where the wild pointer lands.

Commit 8628297 guarded the DC coefficient write (pMCU[0]) with if (iMCU >= 0), which prevents crashes for progressive JPEGs whose first scan is DC-only (iScanEnd == 0). However, if the first scan includes AC coefficients (iScanEnd > 0), the AC decode loop still writes through the wild pointer and crashes.

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -51,6 +51,7 @@ extra_scripts =
   pre:scripts/build_html.py
   pre:scripts/gen_i18n.py
   pre:scripts/git_branch.py
+  pre:scripts/patch_jpegdec.py
 
 ; Libraries
 lib_deps =
--- a/scripts/patch_jpegdec.py
+++ b/scripts/patch_jpegdec.py
@@ -1,0 +1,68 @@
+"""
+PlatformIO pre-build script: patch JPEGDEC for MCU_SKIP wild pointer crash.
+
+Problem:
+  JPEGDecodeMCU_P computes pMCU = &sMCUs[iMCU & 0xffffff].  When iMCU is
+  MCU_SKIP (-8), the bitmask produces index 0xFFFFF8 (16 777 208), creating a
+  pointer ~33 MB past the 392-entry sMCUs array.  If the progressive JPEG's
+  first scan includes AC coefficients (iScanEnd > 0), the AC decode loop writes
+  through this wild pointer and crashes with a store-access fault.
+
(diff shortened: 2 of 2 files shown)


itsthisjustin:
@martinbrook I believe this was originally something you contributed to if you could take a look

Uri-Tauber:
@itsthisjustin would you mind opening a PR for this bug in the upstream [JPEGDEC](https://github.com/bitbank2/JPEGDEC) library?
I think it’d be cleaner than adding a custom script here to fix it.

itsthisjustin:
> @itsthisjustin would you mind opening a PR for this bug in the upstream [JPEGDEC](https://github.com/bitbank2/JPEGDEC?rgh-link-date=2026-04-09T19%3A01%3A16Z) library? I think it’d be cleaner than adding a custom script here to fix it.

I absolutely can and will. Any idea how fast that would get taken care of though? My fear is that without this, x3 users will have crashes on books. It's a hard crash too.

Uri-Tauber:
@itsthisjustin Not sure, but @bitbank2 is very active on github, so he will probably see it soon.

itsthisjustin:
> JPEGDEC

Issue opened. https://github.com/bitbank2/JPEGDEC/issues/119

martinbrook:
> 

Nice bug report @itsthisjustin , Larry was responsive when addressing our last report but I did use a local patch while waiting for the fix so we have a precedent there.

martinbrook (approved):
I don't have a device to test this but approving. We do have a precedent of patching the upstream while waiting for Larry.

znelson (approved):

znelson:
Nicely done - thanks for the thorough documentation on the upstream issue.

PR #1629, opened 2026-04-09
fix: small QoL: return to the last selected menu location

Uri-Tauber:
## Summary

* **What is the goal of this PR?** Small UX improvement to the Home screen by preserving the last selected cursor position when returning to it.

It supersedes #985 and #1103, which are both significantly outdated and hundreds of commits behind master.

---

### AI Usage


src/activities/ActivityManager.cpp
```diff
@@ -196,7 +196,23 @@ void ActivityManager::goToFullScreenMessage(std::string message, EpdFontFamily::
...
+    } else if (activityName == "OpdsBookBrowser") {
+      initialMenuItem = HomeMenuItem::OPDS_BROWSER;
+    } else if (activityName == "CrossPointWebServer") {
+      initialMenuItem = HomeMenuItem::FILE_TRANSFER;
+    } else if (activityName == "Settings") {
+      initialMenuItem = HomeMenuItem::SETTINGS_MENU;
+    }
+  }
+  replaceActivity(std::make_unique<HomeActivity>(renderer, mappedInput, initialMenuItem));
+}
```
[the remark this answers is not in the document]
Uri-Tauber: I tested the feature on my device and its working without any problem. could you suggest a test to check your problem?

src/activities/home/HomeActivity.cpp
```diff
@@ -113,11 +113,32 @@ void HomeActivity::onEnter() {
   // Check if OPDS browser URL is configured
   hasOpdsUrl = strlen(SETTINGS.opdsServerUrl) > 0;
 
-  selectorIndex = 0;
-
   const auto& metrics = UITheme::getInstance().getMetrics();
   loadRecentBooks(metrics.homeRecentBooksCount);
 
+  const auto base = static_cast<int>(recentBooks.size());
```
jonvex, line 119: This logic duplicates the logic at https://github.com/crosspoint-reader/crosspoint-reader/blob/14ec53a33556b9084600b066fcc299ffd222783b/src/activities/home/HomeActivity.cpp#L189

You should try to refactor common logic

src/activities/home/HomeActivity.h
```diff
@@ -19,6 +19,7 @@ class HomeActivity final : public Activity {
   bool coverRendered = false;      // Track if cover has been rendered once
   bool coverBufferStored = false;  // Track if cover buffer is stored
   uint8_t* coverBuffer = nullptr;  // HomeActivity's own buffer for cover image
+  const HomeMenuItem initialMenuItem;
```
jonvex: initialSelectorMenuItem

src/activities/ActivityManager.cpp
```diff
@@ -196,7 +196,23 @@ void ActivityManager::goToFullScreenMessage(std::string message, EpdFontFamily::
   replaceActivity(std::make_unique<FullScreenMessageActivity>(renderer, mappedInput, std::move(message), style));
 }
 
-void ActivityManager::goHome() { replaceActivity(std::make_unique<HomeActivity>(renderer, mappedInput)); }
+void ActivityManager::goHome(HomeMenuItem initialMenuItem) {
+  if (initialMenuItem == HomeMenuItem::NONE && currentActivity) {
```
jonvex, line 211: Instead of this, maybe we can make HomeMenuItem like this? And then we can just call fromString:

```
struct HomeMenuItem {
    enum Value { NONE, FILE_BROWSER, RECENTS, OPDS_BROWSER, FILE_TRANSFER, SETTINGS_MENU };

    Value value;

    HomeMenuItem(Value v) : value(v) {}

    // Forward: enum → string
    constexpr std::string_view toString() const {
        switch (value) {
            case NONE:          return "None";
            case FILE_BROWSER:  return "File Browser";
            case RECENTS:       return "Recents";
            case OPDS_BROWSER:  return "OPDS Browser";
            case FILE_TRANSFER: return "File Transfer";
            case SETTINGS_MENU: return "Settings Menu";
        }
        return "Unknown";
    }

    // Inverse: string → enum
    static HomeMenuItem fromString(std::string_view s) {
        if (s == "None")          return NONE;
        if (s == "File Browser")  return FILE_BROWSER;
        if (s == "Recents")       return RECENTS;
        if (s == "OPDS Browser")  return OPDS_BROWSER;
        if (s == "File Transfer") return FILE_TRANSFER;
        if (s == "Settings Menu") return SETTINGS_MENU;
        return NONE; // fallback
    }
};

// Usage
HomeMenuItem item = HomeMenuItem::fromString("Recents");
item.toString(); // "Recents"
```

jonvex (commented):
logic seems good, but a few suggestions to improve maintainability

src/activities/ActivityManager.cpp
Uri-Tauber: Nice suggestion! Thx.

src/activities/home/HomeActivity.cpp
Uri-Tauber: Will do!

src/activities/home/HomeActivity.cpp
```diff
@@ -113,11 +113,12 @@ void HomeActivity::onEnter() {
...
+  selectorIndex = initialMenuItem == HomeMenuItem::NONE ? 0 : base + initialMenuItem.toIndex(hasOpdsUrl);
```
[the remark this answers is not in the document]
Uri-Tauber: There is only one recent book, and it correctly return to the first position.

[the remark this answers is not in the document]
Uri-Tauber: You should flag such small bugs as "minor", not " 🟠 Major". Assume the author always tested the feature on their device and its works well. Only flag problem as "Major" if there is a chance for OOM, invalid pointer, etc.. things which may not surface in casual use.

src/activities/home/HomeActivity.cpp
```diff
@@ -196,16 +197,27 @@ void HomeActivity::loop() {
...
-      onRecentsOpen();
-    } else if (menuSelectedIndex == opdsLibraryIdx) {
-      onOpdsBrowserOpen();
-    } else if (menuSelectedIndex == fileTransferIdx) {
-      onFileTransferOpen();
-    } else if (menuSelectedIndex == settingsIdx) {
-      onSettingsOpen();
+    } else {
+      const int menuIndex = selectorIndex - static_cast<int>(recentBooks.size());
+      switch (HomeMenuItem::fromIndex(menuIndex, hasOpdsUrl).value) {
```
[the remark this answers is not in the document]
Uri-Tauber: fixed

src/activities/ActivityManager.cpp
```diff
@@ -196,7 +196,23 @@ void ActivityManager::goToFullScreenMessage(std::string message, EpdFontFamily::
...
+    const auto& activityName = currentActivity->name;
+    if (activityName == "FileBrowser") {
+      initialMenuItem = HomeMenuItem::FILE_BROWSER;
```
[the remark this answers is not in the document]
Uri-Tauber: fixed

src/activities/ActivityManager.h
```diff
@@ -16,6 +16,43 @@
 class Activity;    // forward declaration
 class RenderLock;  // forward declaration
 
+struct HomeMenuItem {
```
znelson: Overall, I like the QoL improvement of this change, but I don't think `HomeMenuItem` belongs in `ActivityManager`. This is coupling `HomeActivity` internal behavior to the public ActivityManager.h.

Can you find a way to do this so this helper data remains private to `HomeActivity`?

Uri-Tauber: You're right. I moved the actual logic into 'HomeActivity', and kept in 'ActivityManager' only an enum class which can be used by other activities.

znelson (commented):
I haven't tested this, but do you think we could get away without externalizing any of the HomeActivity's details about menu ordering by using a private static?

[private-static.patch](https://github.com/user-attachments/files/27420629/crosspoint-reader-7-57-17.PM.patch)

Uri-Tauber:
@znelson, I can't access your patch.

Uri-Tauber:
Anyways, menu ordering is already private to `HomeActivity`,  What’s exposed is the semantic target (`HomeMenuItem`) so callers can request a destination.

znelson:
Try this? https://gist.github.com/znelson/f739c4944abe662beee5c6f8320306d0

My point is: after this change, if `HomeActivity` gets a new menu item, we need to modify ActivityManager.h. However, `HomeActivity` is really the conceptual owner of `HomeMenuItem`, because it displays and handles those menu items. 

The patch is an attempt to keep `HomeMenuItem` as a private `HomeActivity` concept.

Uri-Tauber:
@znelson

```diff
-  void onGoHome(HomeMenuItem item = HomeMenuItem::OPDS_BROWSER);
+  void onGoHome();
```

I think the responsibility for deciding the landing destination should remain with each activity, rather than being handled inside `HomeActivity`.

If `HomeActivity` owns that state, it can become inconsistent with the actual navigation flow. For example: if a user opens a file from *Recent Books*, then long-presses back to return to the File Browser, and finally goes back to Home — it should land on File Browser, not Recent Books.

I haven’t tested your patch yet, but from reading the change, it looks like this scenario might regress.

What do you think?

Uri-Tauber:
It’s true that if this PR is merged, adding a new MenuItem will require changes in four different files, which isn’t very convenient. However, this isn’t something that happens often, so it’s probably an acceptable trade-off.

itsthisjustin:
@Uri-Tauber are you still in favor of getting this merged? can you help me understand the user value here? It's probably just late and I'm missing something simple

Uri-Tauber:
@itsthisjustin 

> Small little improvement to the Home screen UX by remembering/restoring the last selected cursor position when returning to Home. This was bothering me whenever I went back to home. It would reset back to the first option (typically the first book in the very upper left) and I'd have to move the cursor all the way back down to the option I wanted

itsthisjustin (approved):

znelson (approved):

PR #1633, opened 2026-04-10
feat: add additional refresh rate and line spacing options

jonvex:
## Summary

Add refresh rates of 100 pages and 1000 pages to the settings

Add "Very Tight" line spacing

## Additional Context

I'm not really seeing any ghosting after 30 pages so I am introducing larger options. 

The "Tight" line spacing isn't compact enough imo. So I added a very tight option

---

### AI Usage

No


src/SettingsList.h
```diff
@@ -44,7 +44,8 @@ inline const std::vector<SettingInfo>& getSettingsList() {
                         {StrId::STR_SMALL, StrId::STR_MEDIUM, StrId::STR_LARGE, StrId::STR_X_LARGE}, "fontSize",
                         StrId::STR_CAT_READER),
       SettingInfo::Enum(StrId::STR_LINE_SPACING, &CrossPointSettings::lineSpacing,
-                        {StrId::STR_TIGHT, StrId::STR_NORMAL, StrId::STR_WIDE}, "lineSpacing", StrId::STR_CAT_READER),
+                        {StrId::STR_TIGHT, StrId::STR_NORMAL, StrId::STR_WIDE, StrId::STR_VERY_TIGHT}, "lineSpacing",
```
jonvex, line 47: Can't change the ordering here because it goes off the enum value. And that is saved to file, so it would cause backwards compatibility issues to change the order

znelson:
Can you share a few screenshots of the new text spacing option in the different fonts?

Does 100 and 1000 page refresh work OK in practice, without getting smeary?

jonvex:
1. I'm thinking maybe we change the config to SettingInfo::Value instead of enum so that the user can just choose any spacing in 0.05 or 0.01 increments. 

2. I've been using this for a week and read 4 or 5 books and haven't noticed any issues with ghosting or smudginess. My relevant configs are:
- Refresh Frequency: 1000 pages
- Sunlight fading fix: off
- Reader Font Family: Bookerly
- Reader Font Size: X Large
- Reader Line Spacing: Very Tight
- Reader Paragraph Alignment: Center
- Embedded Style: OFF
- Hyphenation: OFF
- Reading Orientation: Portrait
- Extra Paragraph Spacing: On
- Text Anti-Aliasing: On
- Images: Placeholder

jonvex:
Added a numeric config for the spacing:
https://github.com/crosspoint-reader/crosspoint-reader/pull/1682

will create a new pr for the refresh rate

PR #1638, opened 2026-04-11
fix: increase loadable epub size

CSCMe:
## Summary

* **What is the goal of this PR?** Slightly increase the OOM limit when loading epubs
* **What changes are included?** Switched vectors for parsing epubs to deques, allowing use of more memory

## Additional Context

* Increases loadable epub size from 2000+ chapter/ToC entries loadable to 5000+ chapter/ToC entries
* #1574, but without the complicated stuff

---

### AI Usage

### Testing
| Commit | Book | Time |
|----------|------|-------|
| 83cd96bc2fd90e5f5a8b0ceaed2022a16a017a1c | 1000.epub | ~30 sec |
| 83cd96bc2fd90e5f5a8b0ceaed2022a16a017a1c | 5000.epub | crash |
| PR | 1000.epub | ~29 sec |
| PR | 5000.epub | ~2 min 20 sec |

=> No actual loading time regressions
[tested_epubs.zip](https://github.com/user-attachments/files/26645243/tested_epubs.zip)

### DIFF
--- a/lib/Epub/Epub/BookMetadataCache.cpp
+++ b/lib/Epub/Epub/BookMetadataCache.cpp
@@ -4,7 +4,7 @@
 #include <Serialization.h>
 #include <ZipFile.h>
 
-#include <vector>
+#include <deque>
 
 #include "FsHelpers.h"
 
@@ -50,7 +50,7 @@ bool BookMetadataCache::beginTocPass() {
 
   if (spineCount >= LARGE_SPINE_THRESHOLD) {
     spineHrefIndex.clear();
-    spineHrefIndex.reserve(spineCount);
+    spineHrefIndex.resize(spineCount);
     spineFile.seek(0);
     for (int i = 0; i < spineCount; i++) {
       auto entry = readSpineEntry(spineFile);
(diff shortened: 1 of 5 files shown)


jpirnay:
That sounds like a good idea, thank you - one feedback:
The ContentOpfParser.cpp uses itemIndex with reserve() + push_back() in its build path (same pattern as spineHrefIndex and targets in BookMetadataCache.cpp). The PR changes itemIndex to deque in the header but doesn't update the .cpp to use resize() + direct assignment. That's not a bug — deque supports push_back() fine — but it leaves the "don't pre-allocate a contiguous block" benefit partially on the table, since reserve() on a deque doesn't actually do anything useful (it's a no-op on deque). Changing those to resize() + index access in ContentOpfParser.cpp would be consistent with the rest of the PR.

CSCMe:
Thanks for pointing that out!

Regarding `ContentOpfParser.cpp`, I don’t think it actually follows the same `reserve() + push_back()` pattern used elsewhere (like `spineHrefIndex` or `targets`)—at least not in the same way those cases do.

More importantly, even if we wanted to switch to `resize() + direct assignment`, that pattern isn’t really applicable here. `ContentOpfParser` builds `itemIndex` incrementally while parsing, and the final size isn’t known up front. Because of that, we can’t safely `resize()` and assign by index without either:

* doing a separate pre-pass to determine the size, or
* introducing additional bookkeeping that would complicate the parsing logic.

Given that, `push_back()` remains the natural and correct approach in this case, regardless of whether the container is a `std::vector` or `std::deque`.

So while I agree with the general goal of avoiding unnecessary contiguous pre-allocation, I think `ContentOpfParser` is a case where the pattern doesn’t translate cleanly.

Happy to revisit if you had a specific spot in mind that I might have missed.

jpirnay:
Thanks for the detailed explanation — you're right. I went back and checked: ContentOpfParser.cpp never called reserve() on itemIndex to begin with (the reserve() calls in that file are only on local strings inside stripHtml()). So there's nothing to update there. The push_back() pattern is the correct approach when size isn't known upfront, and the deque already avoids the large contiguous allocation. The adaptation is fine as-is.

jpirnay (approved):
Looking good to me

znelson (approved):

PR #1653, opened 2026-04-13
docs: Update README esptool.py firmware flashing instructions

ryneches:
Added instructions for flashing firmware using esptool.py.

## Summary

Add instructions to README.md for flashing the firmware using `esptool.py` from the terminal.

## Additional Context

Not everyone has Chrome or Chromium installed, so it would be nice not to have to hunt for the instructions for using `esptool.py`.

---

### AI Usage


ryneches:
Added `docs:` the PR title, since this is a documentation only suggestion.

PR #1662, opened 2026-04-14
fix: Accumulate glyph positions in fixed-point for smoother text spacing

### LINKED ISSUE #1182: Kerning leads to uneven character spacing
### Affected Version

1.1.0

### Describe the Bug

In some constellation, the current kerning seems to lead uneven character spacing. Eg the word "drew" on Bookerly has a gap between d and r, with the rest of the characters being quite condensed. 

@znelson  maybe you have an idea. Maybe some fine tuning is needed to move characters with serifs closer together?

### Steps to Reproduce

Look for uneven character spacing

### Expected Behavior

Soils be more even

```shell

```

ingpaschke:
## Summary

* **What is the goal of this PR?**  Complete the fix for #1182 — eliminate remaining character spacing unevenness that persisted after #1168 introduced fixed-point advances.
  - `drawText()`, `drawTextRotated90CW()`, and `getTextBounds()` now accumulate the cursor position in 12.4 fixed-point and only snap to integer pixels at render time
  - Previously, each glyph advance was rounded to the nearest pixel before being added to the cursor, discarding fractional precision at every step

## Background                                             
                                                                                                                                                         
PR #1168 by @znelson introduced 12.4 fixed-point advances and differential rounding, which was a significant improvement. However, as @jonasdiemer noted in #1182:

  > On Small (just tried recently), I still feel there's room for improvement, but can't pinpoint it yet                                                 
  
He attributed the remaining unevenness to screen resolution limits, but it was actually residual rounding error from the differential rounding approach, which still discarded the fractional remainder at each step.

## How it works

The old code rounded and added each step:
`cursor (int) += round(advance + kern)   // fraction lost every character
`
The new code accumulates without rounding:
```
cursor (fixed-point) += advance + kern  // no precision loss
snapX = round(cursor)                   // snap only for rendering

```
This ensures each glyph is placed at the rounded version of its true cumulative position, rather than a chain of individually rounded steps. The fractional error never compounds.

## Impact

- **6,030 pixels shifted** on a single test page (Bookerly 14pt, justified text)
- Zero performance cost — same number of operations, same types
- Binary is 12 bytes smaller (removed a `toPixel()` call from the advance step)
- Measurement (`getTextBounds`) updated to match, so line breaks stay consistent

Tested on X4 hardware. 

Side-by-side with highlights (left = old/red, right = new/blue):

<img width="964" height="800" alt="font-diff-highlighted" src="https://github.com/user-attachments/assets/0492b7ed-fcf9-4075-be9a-a20323afc059" />

Fixes #1182
---

### AI Usage

The fix was developed collaboratively with Claude Code (Opus 4.6). The approach (fixed-point accumulator vs per-step rounding) is a well-known technique in text rendering — Claude identified the rounding pattern in the existing code and implemented the fix.

### DIFF
--- a/lib/EpdFont/EpdFont.cpp
+++ b/lib/EpdFont/EpdFont.cpp
@@ -15,10 +15,9 @@ void EpdFont::getTextBounds(const char* string, const int startX, const int star
     return;
   }
 
-  int lastBaseX = startX;
-  int lastBaseAdvanceFP = 0;  // 12.4 fixed-point
+  int32_t cursorFP = fp4::fromPixel(startX);  // accumulate in 12.4 fixed-point
+  int lastBaseAdvanceFP = 0;                  // 12.4 fixed-point (for combining mark centering)
   int lastBaseTop = 0;
-  int32_t prevAdvanceFP = 0;  // 12.4 fixed-point: prev glyph's advance + next kern for snap
   constexpr int MIN_COMBINING_GAP_PX = 1;
   uint32_t cp;
   uint32_t prevCp = 0;
@@ -31,9 +30,7 @@ void EpdFont::getTextBounds(const char* string, const int startX, const int star
 
     const EpdGlyph* glyph = getGlyph(cp);
     if (!glyph) {
-      lastBaseX += fp4::toPixel(prevAdvanceFP);  // flush pending advance before resetting
       prevCp = 0;
-      prevAdvanceFP = 0;
       continue;
     }
 
@@ -45,12 +42,14 @@ void EpdFont::getTextBounds(const char* string, const int startX, const int star
       }
     }
 
+    // Add kerning to the accumulator (no rounding yet)
     if (!isCombining && prevCp != 0) {
       const auto kernFP = getKerning(prevCp, cp);  // 4.4 fixed-point kern
-      lastBaseX += fp4::toPixel(prevAdvanceFP + kernFP);
+      cursorFP += kernFP;
     }
 
-    const int glyphBaseX = isCombining ? (lastBaseX + fp4::toPixel(lastBaseAdvanceFP / 2)) : lastBaseX;
+    const int snapX = fp4::toPixel(cursorFP);
+    const int glyphBaseX = isCombining ? fp4::toPixel(cursorFP + lastBaseAdvanceFP / 2) : snapX;
     const int glyphBaseY = startY - raiseBy;
 
     *minX = std::min(*minX, glyphBaseX + glyph->left);
@@ -61,7 +60,7 @@ void EpdFont::getTextBounds(const char* string, const int startX, const int star
     if (!isCombining) {
       lastBaseAdvanceFP = glyph->advanceX;  // 12.4 fixed-point
       lastBaseTop = glyph->top;
-      prevAdvanceFP = lastBaseAdvanceFP;
+      cursorFP += lastBaseAdvanceFP;  // advance accumulator without rounding
       prevCp = cp;
     }
   }
--- a/lib/GfxRenderer/GfxRenderer.cpp
+++ b/lib/GfxRenderer/GfxRenderer.cpp
@@ -1011,8 +1011,7 @@ int GfxRenderer::getTextAdvanceX(const int fontId, const char* text, EpdFontFami
 
   uint32_t cp;
   uint32_t prevCp = 0;
-  int widthPx = 0;
-  int32_t prevAdvanceFP = 0;  // 12.4 fixed-point: prev glyph's advance + next kern for snap
+  int32_t cursorFP = 0;  // accumulate in 12.4 fixed-point, matching drawText
   const auto& font = fontIt->second;
   while ((cp = utf8NextCodepoint(reinterpret_cast<const uint8_t**>(&text)))) {
     if (utf8IsCombiningMark(cp)) {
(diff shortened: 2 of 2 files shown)


znelson:
Hey @ingpaschke, I appreciate having another set of eyes on the subtle spacing tweaks to get great typographic layout!

I see the PR notes only call out #1168, not the tweaks for differential rounding in #1413. Can you help me compare this to 1398aeb1edb3a2819e90d52bd96f339c8e073491? 

I was trying to fix the case where I'd see common pairs of characters, e.g. "oo", with slightly different spacing across a page. My analysis was that it was happening because each character snaps to a position based on accumulated fixed-point error, and so spacing between specific glyph pairs can vary. For example, say the spacing for "oo" is 8.875 px in fixed-point. Depending on the cumulative error, sometimes that can snap to 8px, other times to 9px. On a relatively low density display like this, and with small font, the pixel snapping difference starts to become obvious, and felt off to me.

The "typographic consensus" is largely shaped by systems that have sub-pixel rendering, anti-aliasing, and high DPI displays, luxuries the X4 doesn't have. My math is that a 1px difference on "oo", for example, works out to a 100% variance in layout: either 1px or 2px space.

Maybe the "non-smooth" cases you observe primarily show up with justified line layout, because the right edge becomes slightly ragged due to the accumulated error of the current differential rounding algorithm? I admit I'm primarily testing with left-aligned text which intentionally leaves a ragged-right.

If #1413 gets us "even" rendering for words, in the sense that pairs of characters get consistent spacing (by using fixed-point differential rounding from the x-advance and kerning space), but leaves us with line-length accumulated error (with ragged edges when justified as a remaining pain point), perhaps we should distribute accumulated line error into inter-word spacing for justified layout?

I'd love to hear from @jonasdiemer too, with his observations regarding #1182 after 1398aeb1edb3a2819e90d52bd96f339c8e073491 (which is currently in main but not in 1.2.0 release).

ingpaschke:
Hi @znelson, thanks for the detailed analysis! You're right that I should have referenced #1413 — the differential rounding was the foundation this builds on.                                                                                                                                             
                                                                                                                                                         
To answer your question about "oo" consistency: yes, with the accumulator approach, identical pairs can get different pixel spacing depending on accumulated fractional position. I understand why differential rounding was designed to prevent this. I like the accumulator approach of the PR better (at least with justified text). But I agree that's probably personal preference. It would be easy to make this a config option.

The fundamental problem is that we calculate sub-pixel positions, but with our pre-rendered pixel font that's antialiased for whole pixels, we can only render at whole pixel positions.

I've spent the last two days experimenting extensively with sub-pixel positioning approaches to find the best trade-off for this hardware. 

Here's what I tested:                                                                                                                                                                                      
                                                            
1. 2x horizontal oversampled fonts — Render at 300 DPI horizontal, 1-bit, downsample at runtime. Half-pixel positioning with runtime AA. Horizontal spacing was great, but vertical edges lost AA quality. Flash dropped to 85% (1-bit compresses better), but the rendering felt less sharp than the pre-baked 2-bit fonts.
                                                                                                                                                         
2. Per-glyph sub-pixel blending — Use the existing 2-bit data, blend each pixel with its neighbor weighted by fractional cursor offset. But the linear blend doesn't capture how FreeType reshapes glyphs at different sub-pixel    positions. Only 14.6% of pixels in a shifted-original matched FreeType's actual 0.5px render. That didn't look good.
                                                                                                                                                         
3. Line-buffer compositing — Render all glyphs into a 2x-wide temp buffer (~4KB), downsample the composited line. Spacing was spot-on and glyph interactions handled correctly, but there's a slight softening vs native hinting.                                                                      
                                                            
4. Dual-position glyphs — Store each glyph twice: rendered by FreeType at 0.0 and 0.5px offset. Pick the closer variant at render time. This gives native hinting sharpness at both positions — best visual result by far. Storage cost is ~25KB compressed per font for the alternates.
                                                                                                                                                         
With the dual-position approach identical pairs always render with the same variant (since the threshold is deterministic) AND each variant is properly hinted for its grid position. No blending, no softening.
                                                                                                                                                         
I think there's a path where the accumulator fix (this PR) lands first as a standalone simple option for those who prefer it, and dual-position glyphs come as a follow-up PR once the storage budget is worked out (possibly combined with #589's compression improvements).

Screenshot of the dual position glyph rendering:

<img width="984" height="1626" alt="image" src="https://github.com/user-attachments/assets/6d235e43-226b-4295-91fe-e2b0e099a839" />

znelson:
@ingpaschke Thanks for the details! Let me give this a few days on device to see if anything stands out 🙂

znelson:
After using this for a few days, I think it's a regression from what's in main.

> The fundamental problem is that we calculate sub-pixel positions, but with our pre-rendered pixel font that's antialiased for whole pixels, we can only render at whole pixel positions.

Yes, that's true. And we need to work within those constraints. 

The fundamental problem with this PR is that the kerning choice for any given pair of letters depends not only on the letters themselves, but also where those letters happen to fall in the accumulated X position across the line. 

This reverts exactly what #1413 was fixing. Glyphs must snap to a pixel, yes. But if we pick the kerning snap including accumulated fixed-point X position error across the line, we get inconsistencies in common kerning pairs.

Here are a number of examples to compare:

| Pair | master | This PR |
| -- | -- | -- |
| 'oo' 12px |  |   |
| 'rr' 12px |  |  |
| 'gg' 12px |  |  |
| 'dd' 12px |  |  |
| 'oo' 14px |  |  |
| 'oo' 14px |  |  |
| 'rr' 14px |  |  |
| 'ea' 'he' 14px |  |  |
| 'dd' 'bb' 14px |  |  |

@ingpaschke Can you please provide specific examples of what looks smoother in rendering with this PR versus master?

If it's about the right-edge alignment with justified text, maybe we'd be better with something like this patch: 
[justified-spacing.patch](https://github.com/user-attachments/files/26858091/justified-spacing.patch)

ingpaschke:
Based on your extensive analysis, I have to agree that this PR on its own is really a regression. That's why I went further and added the dual glyph rendering and tracking, which I still think produced superior results — not just consistent within pairs, but consistent across the whole line. The eye sees a uniform rhythm.

You can't really see the difference in screenshots from the device, because they don't show the greyscale antialising, so here are two photos.

Stock firmware: 

image

With FP4 accumulation, dual glyph and tracking:

image

But I understand that these changes are a larger restructuring than fits CrossPoint's current constraints. The dual-glyph approach naturally fits an SD-card-delivered model once user-font support lands — happy to contribute in that direction when the time comes.

znelson:
@ingpaschke Would you be willing to join the group discussing font changes in #1693? We definitely need people passionate about fonts and font rendering to drive the next wave of improvements.

znelson:
Closing in favor of other font discussions.

PR #1666, opened 2026-04-15
fix: footnote link text

steka:
## Summary

Previouls where ALL whitespace (and square brackes) removed from the footnote link text, however some link texts are multiworded, like "`turn to 252`" which were truncated into "`turnto252`", (an example from the first book "Flight from the Dark" of the Lone Wolf book series by Joe Dever, see [link](https://www.projectaon.org/en/Main/FlightFromTheDark))

* This change will only remove whitespaces from the beginning and end of the string
so "`   [  12  ]   `" will become "`12`" just like before, and "`   turn to 252   `" will become "`turn to 252`".

---

### AI Usage


CSCMe (approved):
Seems reasonable

znelson (approved):
LGTM

PR #1671, opened 2026-04-16
Make the keyboard a grid based entry instead of directional input

beef331:
## Summary
This PR changes the keyboard to what I find is a more intuitive format. Instead of pretending the buttons are directional inputs it assumes they're choosing one of three options. The 'volume' buttons are used for moving between pages of keys be them uppercase, lowercase, or symbols.

## Additional Context

Mostly making this PR to see if anyone has any interest in this. If there is I could make it an option for the standard keyboard or this.

To show case how it is to interact.

The default state is the lower case keymap.
screenshot-178549

The user hit 'volume down' which moved down to the uppercase keymap.
screenshot-184107

The user hit the 'confirm' button, choosing the first block.
screenshot-195108

The user hit the 'right button' choosing the last block.

screenshot-209564

---

### AI Usage

### DIFF
--- a/src/activities/util/KeyboardEntryActivity.cpp
+++ b/src/activities/util/KeyboardEntryActivity.cpp
@@ -6,19 +6,6 @@
 #include "components/UITheme.h"
 #include "fontIds.h"
 
-// Keyboard layouts - lowercase
-const char* const KeyboardEntryActivity::keyboard[NUM_ROWS] = {
-    "`1234567890-=", "qwertyuiop[]\\", "asdfghjkl;'", "zxcvbnm,./",
-    "^  _____<OK"  // ^ = shift, _ = space, < = backspace, OK = done
-};
-
-// Keyboard layouts - uppercase/symbols
-const char* const KeyboardEntryActivity::keyboardShift[NUM_ROWS] = {"~!@#$%^&*()_+", "QWERTYUIOP{}|", "ASDFGHJKL:\"",
-                                                                    "ZXCVBNM<>?", "SPECIAL ROW"};
-
-// Shift state strings
-const char* const KeyboardEntryActivity::shiftString[3] = {"shift", "SHIFT", "LOCK"};
-
 void KeyboardEntryActivity::onEnter() {
   Activity::onEnter();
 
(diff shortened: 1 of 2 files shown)


pablohc:
Interesting suggestion, @beef331. 
I'm currently working on a visual update to the current keyboard at https://github.com/crosspoint-reader/crosspoint-reader/pull/1644, but this would be useful as an additional mode, following @itsthisjustin's suggestion to simplify navigation using the arrow keys.
Do you have any screenshots about selection of symbols/numbers/special buttons like delete, space, etc., would look like?
I don't want to get ahead of myself, but to access the symbols, do you have to go to the third column > third column > third column, and that would bring up a new grid with all the available symbols?

image

Uri-Tauber:
@beef331 Really interesting and out-of-the-box idea. I think it might confuse about 95% of users, but it’s so cool that I’d suggest offering it as a secondary keyboard layout.

FireMotion:
I think #1644 also already put a lot of thought in redesigning the keyboard for better use, while keeping it a bit more traditional.

pablohc:
> I think #1644 also already put a lot of thought in redesigning the keyboard for better use, while keeping it a bit more traditional.

@FireMotion I still have a few improvements left to implement in my PR, but would you like to add this as an additional layout over my PR? Specifically, to the right of the "Shift" key is the Mode key, which toggles between "ABC" and "Symbol" with a single click. We could integrate it so that a long press activates this typing mode. But I'd like to try to understand a little more about how to access symbols or special functions.

steka:
I had a similar idea, but haven't started implementing it.
I think I will put it aside for now.

beef331:
@pablohc Space is done by holding right, backspace is done by holding left, symbols are just another page you access by hitting page down. So if you hit page down enough times you are greeted with
screenshot-139039

It's not presently utf8 aware, but with the current setup one just adds glyphs to https://github.com/crosspoint-reader/crosspoint-reader/pull/1671/changes#diff-7f1ddc2d54858309459df5a85e5a8ca6771230ad3982f07581f8c1606bf8ba50R22 where each entry is a row. It would not be overly difficult to support entire non latin alphabets once adding utf8 support.

In response for more features I see you PR has a cursor. My idea was to use holding volume up to move it left and holding volume down to move it right. Given the layout on a x4 it's not ideal, but I imagine its intuitive on a x3.

pablohc:
I really like @beef331! I'm going to try it out and give you some feedback.

itsthisjustin:
In principal I agree with a dpad style keyboard but I don't think this is the implementation we need right now. I'm going to close this as we just redid the keyboard to strong community response and this is not a priority at the moment. Please feel free to open this back up though. I'd encourage you to look at how TV remotes/game controllers do similar layouts

PR #1676, opened 2026-04-16
feat: sub-pixel text rendering with fixed-point positioning and dual-position glyphs

ingpaschke:
## Summary

* **What is the goal of this PR?** Improve text rendering quality through sub-pixel precision — smoother character spacing, better justified text, and fewer typographic artifacts at page boundaries.

This is all the rendering fixes I experimented with combined. It solves the perceived uneven glyph positioning problem and also adds significant improvements to the uniformness of justified text by leveraging the new subpixel positioning to help with text flow at the paragraph level, especially for justified and book type paragraph alignment. Supersedes #1662.

  - Replace integer glyph positioning with 12.4 fixed-point accumulation in `drawText`, `getTextAdvanceX`, and the layout engine, eliminating rounding drift that caused uneven character spacing at small font sizes.
  - Embed DEFLATE-compressed alternate glyph bitmaps rendered at 0.5px horizontal offset. The renderer selects between primary and alt based on the fractional cursor position. Enabled for Bookerly 12pt and 14pt. Adds `--dual-position` flag to `fontconvert.py`.
  - Paragraph-level tracking optimization for justified text: tries multiple letter-spacing values and re-breaks lines to minimize word-gap stretching. Per-line tracking capped at 0.5px/char.
  - Widow/orphan control prevents single lines at page boundaries.

## Additional Context

* The fixed-point accumulation and tracking improvements benefit all fonts — dual-position glyphs are an optional add-on that can be enabled per font via `--dual-position` in `fontconvert.py`. Fonts without dual-position data still get smoother spacing from the FP cursor.
* Flash usage increases from 89.0% to 94.4% (+364KB for dual-position glyph data in 8 Bookerly font variants). 16pt/18pt Bookerly and all other font families are not dual-position to stay within budget.
* Section cache version bumped to 20 (per-word tracking serialization) — cached books will re-index on first open.

---

### AI Usage


ingpaschke:
Re: the 43 font headers missing trailing fields — this is by design. In C++, aggregate initialization zero-fills any trailing members not explicitly listed, so the unchanged font headers get `nullptr` for all three alt fields at compile time. Regenerating them would produce identical binaries but touch 43 files for no functional change.

znelson:
I appreciate the spirit behind this change, but I don't think this is something we can take into CrossPoint at the moment. It's not sustainable to take changes for Bookerly that we can't take for other fonts. Longer-term, we need to get to a point where we can support user fonts, and this approach could become more viable then.

If you're interested in making fonts in CrossPoint better, please join the discussion at #1693.

PR #1677, opened 2026-04-16
feat: Add Latin Extended Additional (U+1E00-1E9F) to font Unicode coverage

scarolan:
## Summary

- Adds the **Latin Extended Additional** Unicode range (U+1E00-1E9F) to the default font conversion intervals in `fontconvert.py`
- Regenerates all 53 builtin font headers with the expanded character set
- Updates font IDs to reflect the new glyph data

This enables proper rendering of:
- **IAST/Sanskrit transliteration**: ṛ ṣ ṇ ṭ ḍ ṃ ḥ (e.g. Kṛṣṇa, Bhīṣma, Arjuna)
- **Welsh**: ẁ ẃ ŵ ẅ
- **Medievalist/scholarly texts**: various dot-below, dot-above diacritical characters
- **Romanized Arabic, Hebrew, and other transliteration** systems

Flash usage remains within the app partition limit (94.1%).

## Test plan
- [ ] Build succeeds (`pio run`)
- [ ] Flash to device and verify IAST characters render correctly (e.g. epub containing "Kṛṣṇa")
- [ ] Verify existing Latin, Cyrillic, and Vietnamese characters still render correctly
- [ ] Verify compression verification passes (`convert-builtin-fonts.sh`)


scarolan:
### Flash impact

| | Flash Used | % of partition |
|---|---|---|
| master | 5,809,363 bytes | 88.6% |
| This PR | 6,168,171 bytes | 94.1% |
| **Delta** | **+358,808 bytes (+351 KB)** | **+5.5%** |

RAM usage is unchanged (30.6%) since glyph data lives in flash. Headroom drops from ~744 KB to ~376 KB.

If the size increase is a concern, an alternative approach would be to use `--additional-intervals 0x1E00,0x1E9F` only for the reader fonts (Bookerly, Noto Sans, OpenDyslexic) and skip the UI fonts (Ubuntu) which don't need these characters. That would save ~20 KB.

znelson:
I'm definitely not against these additions, but I'll note we've held off on a few of the Unicode ranges in the past to keep some headroom for basic UI localization and other firmware tweaks. 

There are a bunch of ideas competing for the last 10% of the available flash space, and the most reasonable path forward is to make the font system more flexible. 

I'm going to close this PR, not because I don't think it's a valuable addition, but because we need to solve some architectural problems to get to a place where we can reasonably take all of the useful Unicode ranges. 

If you're interested in discussing fonts, please hop over to #1693 with us.

PR #1683, opened 2026-04-17
feat: add setting option to disable full refresh in reader

jonvex:
## Summary

I am not seeing any ghosting or smudging inside the reader. So I don't need the refresh and added an option in the settings to disable it.

## Additional Context

Can this harm the display in any way? That's the only potential concern I have.

---

### AI Usage

No


src/CrossPointSettings.cpp
```diff
@@ -288,6 +288,8 @@ int CrossPointSettings::getRefreshFrequency() const {
       return 15;
     case REFRESH_30:
       return 30;
+    case REFRESH_DISABLED:
+      return 9999;
```
znelson, line 292: Maybe:

```suggestion
      return std::numeric_limits<int>::max();
```

src/SettingsList.h
```diff
@@ -26,10 +26,10 @@ inline const std::vector<SettingInfo>& getSettingsList() {
       SettingInfo::Enum(StrId::STR_HIDE_BATTERY, &CrossPointSettings::hideBatteryPercentage,
                         {StrId::STR_NEVER, StrId::STR_IN_READER, StrId::STR_ALWAYS}, "hideBatteryPercentage",
                         StrId::STR_CAT_DISPLAY),
-      SettingInfo::Enum(
-          StrId::STR_REFRESH_FREQ, &CrossPointSettings::refreshFrequency,
-          {StrId::STR_PAGES_1, StrId::STR_PAGES_5, StrId::STR_PAGES_10, StrId::STR_PAGES_15, StrId::STR_PAGES_30},
-          "refreshFrequency", StrId::STR_CAT_DISPLAY),
+      SettingInfo::Enum(StrId::STR_REFRESH_FREQ, &CrossPointSettings::refreshFrequency,
+                        {StrId::STR_PAGES_1, StrId::STR_PAGES_5, StrId::STR_PAGES_10, StrId::STR_PAGES_15,
+                         StrId::STR_PAGES_30, StrId::STR_NEVER},
```
znelson, line 31: Missing a `STR_NEVER` localized string in this PR?

znelson: Ah, I see. `STR_NEVER` is already used for `hideBatteryPercentage`.

iandchasse:
Having this option may cause users to complain about the inevitable ghosting that might occur when reading for longer than X pages. Of course it’s worse in environments with more intense temperatures, and that’s unpredictable. It would be best to offer this option with a warning attached, maybe just a “(experimental)” tag on the string? I don’t know

itsthisjustin:
> Having this option may cause users to complain about the inevitable ghosting that might occur when reading for longer than X pages. Of course it’s worse in environments with more intense temperatures, and that’s unpredictable. It would be best to offer this option with a warning attached, maybe just a “(experimental)” tag on the string? I don’t know 

My position is...this is anti pattern to a reading device and eink standards. I don't think we should have experimental stuff in a release either.

src/CrossPointSettings.cpp
jonvex: I was concerned about overflow so I thought maybe just picking a big number would be more safe?

jonvex:
It was over 80 F (almost 30 C) yesterday and I read for over an hour with the refresh off and I didn't see any ghosting. 

My relevant configs are:
- Refresh Frequency: Never
- Sunlight fading fix: off
- Reader Font Family: Bookerly
- Reader Font Size: X Large
- Reader Line Spacing: Tight
- Reader Paragraph Alignment: Center
- Embedded Style: OFF
- Hyphenation: OFF
- Reading Orientation: Portrait
- Extra Paragraph Spacing: On
- Text Anti-Aliasing: On
- Images: Placeholder
My battery was pretty much full.

znelson:
I'm on the fence here. I hear that @jonvex can't see any regression with no full refresh, and I haven't tried it myself. And yet (I think) no commercial ereader ships with an option for zero full refresh cycle. I lean toward the side of "hard refresh every 30-ish pages is expected from this class of product, and not detrimental to the average user".

itsthisjustin:
> I'm on the fence here. I hear that @jonvex can't see any regression with no full refresh, and I haven't tried it myself. And yet (I think) no commercial ereader ships with an option for zero full refresh cycle. I lean toward the side of "hard refresh every 30-ish pages is expected from this class of product, and not detrimental to the average user". 

Yeah I'm in the camp of just because we can doesn't mean we should. I don't personally see the advantage here but we can poll community. In general I think we need to have more of an upvote downvote system for features like this so we aren't tunnel visioning what goes into the code base

znelson:
@jonvex thank you for the contribution. I don't think this is right for CrossPoint at the moment, because the experience improvement is relatively small versus the experimental nature. Please don't let this discourage you from participating and proposing other PRs. 🙂

PR #1686, opened 2026-04-17
fix: Switch to xpath map for paragraph level syncing in KOSync

itsthisjustin:
Switch KOReader sync progress mapping from chapter matching to XPath-based mapping.

- resolves KOReader positions using real XHTML ancestry paths
- supports paragraph-based upload mapping with text offsets
where needed
- passes the current paragraph index into sync so uploads map
back to KOReader more accurately

No HTTP client changes are included. No reader-state or resume-flow changes are included.


lib/Epub/Epub/parsers/ChapterHtmlSlimParser.cpp
```diff
@@ -1084,6 +1090,7 @@ void ChapterHtmlSlimParser::addLineToPage(std::shared_ptr<TextBlock> line) {
   }
 
   if (currentPageNextY + lineHeight > viewportHeight) {
+    paragraphIndexPerPage.push_back(xpathParagraphIndex);
```
znelson: I notice that all three `paragraphIndexPerPage.push_back()` calls happen immediately before a `completePageFn` callback to the Section; there's a 1:1 relationship between that callback and insertions into this new vector.

Instead of keeping a dynamically growing structure of paragraph indices here, one per page, can we pass the paragraph index directly into the callback? Then, in the `completePageFn` lambda in Section, store the paragraph indices in the same singular growing `lut` structure?

e.g. In Section.cpp:
```c++
// page lookup table structure
struct PageLUT {
  uint32_t fileOffset {0};
  uint16_t paragraphIndex {0};
};
std::vector<PageLUT> lut = {};

// ...

// completePageFn lambda:
[this, &lut](std::unique_ptr<Page> page, uint16_t paragraphIndex) {
  lut.emplace_back({onPageComplete(std::move(page)), paragraphIndex}); },
```

This avoids having an extra dynamically allocating vector fighting for contiguous heap space during chapter parsing.

lib/KOReaderSync/ChapterXPathResolver.cpp
```diff
@@ -0,0 +1,573 @@
...
+  for (const auto& segment : path) {
+    xpath += "/" + segment.name + "[" + std::to_string(segment.index) + "]";
+  }
+  if (charOffset > 0) {
+    xpath += "/text()." + std::to_string(charOffset);
+  }
+  return xpath;
+}
+
+size_t countUtf8Codepoints(const XML_Char* data, const int len) {
```
znelson, line 63: Might be able to simplify with `utf8NextCodepoint` in Utf8.h

src/activities/reader/KOReaderSyncActivity.h
```diff
@@ -58,6 +61,8 @@ class KOReaderSyncActivity final : public Activity {
   int currentSpineIndex;
   int currentPage;
   int totalPagesInSpine;
+  uint16_t currentParagraphIndex = 0;
+  bool hasCurrentParagraphIndex = false;
```
znelson: You could pass and store these as a single `std::optional<uint16_t>`

src/activities/reader/EpubReaderActivity.h
```diff
@@ -17,6 +17,7 @@ class EpubReaderActivity final : public Activity {
   int pagesUntilFullRefresh = 0;
   int cachedSpineIndex = 0;
   int cachedChapterTotalPageCount = 0;
+  bool pendingLastPageJump = false;
```
znelson: Can we make the generic, so it could potentially be used for a pending page jump to any page?

```
std::optional<uint16_t> pendingPageJump;
```

If non-empty and set to `UINT16_MAX`/`std::numeric_limits<uint16_t>::max()`/anything larger than the chapter page count, then jump to the last page?

lib/KOReaderSync/ProgressMapper.cpp
```diff
@@ -3,112 +3,228 @@
...
+    bytesWritten++;
+
+    if (c == '<') {
+      globalInTag = true;
+    } else if (c == '>') {
+      globalInTag = false;
+    } else if (!globalInTag) {
+      const bool startsCodepoint = (c & 0xC0) != 0x80;
+      if (startsCodepoint) {
+        totalVisChars++;
```
znelson: Not critical, but worth thinking about: I think there's still a mismatch here between visible characters and raw XHTML characters. Would an entity like `&amp;` be counted as five characters here?

znelson (commented):
I'm not a KoReader sync user, so I can't comment on or test functionality. Mentioned a couple of small opportunities to minimize dynamic allocations during section parsing and collapse int+bool data into a single optional.

The section parsing part matters most to me, because it affects everyone regardless of KoReader sync usage. The rest is entirely negotiable and fine for iteration.

lib/Epub/Epub/parsers/ChapterHtmlSlimParser.cpp
itsthisjustin: Good call. I removed the extra paragraphIndexPerPage vector from ChapterHtmlSlimParser and now pass the paragraph index directly through completePageFn, so Section stores it alongside the existing page LUT data during parsing. This keeps the same behavior, but avoids the extra dynamically growing parser-owned allocation you pointed out.

lib/KOReaderSync/ProgressMapper.cpp
itsthisjustin: Patched the fallback paragraph streamer so standard HTML entities are counted by their resolved visible text instead of raw source bytes

src/activities/reader/EpubReaderActivity.h
itsthisjustin: Refactored the internal jump state to a generic pendingPageJump optional so it can represent any deferred page jump.

src/activities/reader/KOReaderSyncActivity.h
itsthisjustin: Collapsed the paragraph index + bool pair into std::optional<uint16_t> so the state is passed and stored as a single value now.

lib/KOReaderSync/ChapterXPathResolver.cpp
itsthisjustin: Switched the UTF-8 codepoint counting over to utf8NextCodepoint and added the direct includes the file was relying on.

znelson (dismissed):
LGTM, needs a clang-format pass though

itsthisjustin:
> LGTM, needs a clang-format pass though

Drives me nuts that our ci just checks that but doesn't fix it lol

znelson (approved):

PR #1687, opened 2026-04-17
chore: Clarify X3 RTC in SCOPE.md

znelson:
## Summary

Clarify that X3 device has a reliable RTC chip, so clock feature can be in-scope depending on device.

---

### AI Usage

### DIFF
--- a/SCOPE.md
+++ b/SCOPE.md
@@ -27,7 +27,12 @@ usability over "swiss-army-knife" functionality.
 * **Language Support:** E.g. Support for multiple languages both in the reader and in the interfaces.
 * **Reference Tools:** E.g. Local dictionary lookup. Providing quick, offline definitions to enhance comprehension 
   without breaking focus.
-* **Clock Display (device dependent):** The X4 relies on the ESP32-C3's internal RTC, which drifts significantly during deep sleep. While NTP sync could partially correct this, CrossPoint does not connect to the internet on every boot. The X3 uses a dedicated DS3231 RTC, which maintains accurate time across sleep cycles and can be treated as a reliable wall clock.
+* **Clock Display (device dependent):** 
+
+| Device | Scope |
+| -- | -- |
+| X3 | The X3 uses a dedicated DS3231 RTC, which maintains accurate time across sleep cycles and can be treated as a reliable wall clock. |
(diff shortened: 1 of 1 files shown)


SCOPE.md
```diff
@@ -27,6 +27,7 @@ usability over "swiss-army-knife" functionality.
 * **Language Support:** E.g. Support for multiple languages both in the reader and in the interfaces.
 * **Reference Tools:** E.g. Local dictionary lookup. Providing quick, offline definitions to enhance comprehension 
   without breaking focus.
+* **Clock Display (device dependent):** The X4 relies on the ESP32-C3's internal RTC, which drifts significantly during deep sleep. While NTP sync could partially correct this, CrossPoint does not connect to the internet on every boot. The X3 uses a dedicated DS3231 RTC, which maintains accurate time across sleep cycles and can be treated as a reliable wall clock.
```
mcrosson: I'd probably turn this into a list by device or a table so it can grow more easily long-term.

mcrosson (commented):
Was the text's intent to leave the door open to an ntp discussion on the x4? The way you've written this, there is room to argue that an 'use ntp, must always sync for time' feature would be in-scope for the x4.

I should note: I'm slightly biased as I'd very much like a clock on my x4, even if only via ntp. That said, I'm also ok if that's a non-option for Crosspoint on the x4.

SCOPE.md
znelson: > Was the text's intent to leave the door open to an ntp discussion on the x4?

It wasn't initially, but I'm not completely against the idea with an appropriate opt-in UX.

How do you feel about amending this section to a table like:

| Device | Scope |
| -- | -- |
| X3 | The X3 uses a dedicated DS3231 RTC, which maintains accurate time across sleep cycles and can be treated as a reliable wall clock. |
| X4 | The X4 relies on the ESP32-C3's internal RTC, which drifts significantly during deep sleep. NTP sync could correct this, with an appropriate user experience around connecting to the internet on wake or on demand. This causes some tension with the **Active Connectivity** section below, so please open a discussion about this UX if it's a feature you would find useful. |

mcrosson: Aside from the squished device heading LGTM :+1:

znelson: Yeah, not sure what's up with the headings. Check out the 'View File' layout: https://github.com/znelson/crosspoint-reader/blob/329e4ff5892690fed9212d44b181cb5aadb2267b/SCOPE.md

mcrosson (approved):

mcrosson:
Approved. Looks good, TY.

PR #1689, opened 2026-04-17
revert: "refactor: replace picojpeg with JPEGDEC for cover art conversion"

znelson:
Reverts crosspoint-reader/crosspoint-reader#1517

While it's nice to get down to only one jpeg library, this needs more testing and polish to be ready. @itsthisjustin observed spurious rendering failures (empty covers) and ghosting.

### DIFF
--- a/lib/JpegToBmpConverter/JpegToBmpConverter.cpp
+++ b/lib/JpegToBmpConverter/JpegToBmpConverter.cpp
@@ -2,14 +2,21 @@
 
 #include <HalDisplay.h>
 #include <HalStorage.h>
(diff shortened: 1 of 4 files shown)


itsthisjustin:
@znelson I am not sure what is causing it. I put your branch on here and my cover started again. I put my fonts branch on which is merged with master, and its also fine there. I think we can close this out until I figure out what happened between commits that caused certain books fallback cover images to render blank. sorry for the missfire here

itsthisjustin:
Confirmed to not fix the image issues reported. They are progressive jpgs and are fixed using the existing epub optimizer.

PR #1714, opened 2026-04-20
feat: Add Ability to Pin Books to Homescreen

SiliconAves:
## Summary

Adds a "Pin" function to the currently unused Back button on the homescreen when a book is highlighted. 

The goal is to stop some books from dislodging from the homescreen when the user wants to read other content that can be finished in one sitting (like a short story on an article).

Currently, if the user wants to keep a certain book on the homescreen after reading something else, they need to go into the file browser and manually open it.

- Added variable to check for pinned status
- Added strings for "Pin" and "Unpin" function when a book is highlighted on the homescreen (pin option will not appear if a menu entry that isnt a book is highlighted)
- Updated Recent books behavior: pinned books are not removed from the recent books list
- Added an inline dot icon that appears before the title of a book when it's pinned on the homescreen (dot already present in system fonts, no new icons added)

## Additional Context

  specific areas to focus on).
 
 <del>When changing the theme from lyra extended to lyra, it doesnt seem like the pinned status is preserved correctly, especially considering lyra extended can have up to 3 pinned books, unlike regular lyra. </del>
 
<del>The classic theme seems to handle this better, by giving the option to unpin a book and letting another previously pinned book take its place. No dot was included for this version.</del>

This is now fixed. What caused the unwanted behaviors:
From the File Transfer and the Settings menu, the interaction for going back to the homescreen was mapped as "wasPressed" instead of "wasReleased" like the other menus. Since the pinning action is mapped to "wasReleased", the back interaction would "input leak" the wasPressed action to the homescreen, causing books to inadvertently unpin by triggering "wasReleased".
Now pinned books behave as expected.

---

### AI Usage


osteotek:
Does the Recent Books menu help you quickly return to the previous book? Additionally, there’s an option to display the last three books on the home screen (Lyra Extended theme).

SiliconAves:
> Does the Recent Books menu help you quickly return to the previous book? Additionally, there’s an option to display the last three books on the home screen (Lyra Extended theme).

I currently use the Lyra Extended theme.

The pinning feature is meant to streamline the workflow of users that read a **primary long-term book**, while occasionally opening multiple **short form epubs or test files for development purposes** .

The actions taken by the user for this UX issue to manifest:
- read portion of a long-term book 
- open 3+ short-term epub files (ex. short stories, converted articles, test files). **This pushes the long-term book off the homescreen.**

To reach the book again, the user needs:
- **5 clicks** to reach and enter the recent books menu when using Lyra Extended (reduced to 3 clicks for 1-Book-Lyra)
- **around 4 clicks** to reach and open the primary book (the user needs to skip over the short-term files before reaching  the long-term one).

With the ability to pin a book to the homescreen, these **7-9 clicks are reduced to one single click** (boot  > confirm to open book). This allows users to separate long-term reading material from temporary short-term reads (or files for testing purposes) without needing to redundantly navigate menus.

itsthisjustin:
Closing this due to adjacent nature of this being related to themes and as per @osteotek's comment around the recent books feature already existing. Feel free to get community validation though through a discussion and we can open this back up. It's a solid idea.

PR #1716, opened 2026-04-20
fix(sleep): correctly crop/scale sleep bitmap on X3 panel

giangthb:
## Summary

* **What is the goal of this PR?** Fix sleep screen bitmap not filling the panel correctly when the bitmap size does not match the screen, specifically when a sleep image is smaller than the screen in one or both dimensions (e.g. an X4-styled `sleep.bmp` at 480×800 rendered on an X3 panel at 528×792).

On X3 (528×792 portrait) an X4-styled sleep image (480×800) used to render at 1:1 in the top-left corner of the screen with ~48px blank on the right and ~72px blank on the bottom, even when `sleepScreenCoverMode=CROP`. Four interacting issues in the bitmap rendering path were responsible:

1. `SleepActivity::renderBitmapSleepScreen()` only entered the scale/crop branch when the bitmap was **strictly larger** than the screen in either dimension. Bitmaps smaller than the screen fell through to a center-only path that never computed a scale.

2. `GfxRenderer::drawBitmap()` only applied the computed fit scale when `fitScale < 1.0f` (downscale only). Upscale requests were silently dropped, so the branch in (1) — even when reached — produced a 1:1 render.

3. The per-pixel render loop iterated source pixels and mapped each to a single destination pixel. When scaling is applied, that leaves gap rows/cols between source rows/cols (an upscale artifact; no issue at 1:1 or downscale).

4. The 1-bit fast path at the top of `drawBitmap()` unconditionally routed uncropped 1-bit bitmaps through `drawBitmap1Bit()`, whose scaling logic only downscales. Undersized 1-bit sleep images would therefore still render at native size even when `SleepActivity` opted in to upscaling.

### Fixes

* Trigger the scale/crop branch in `SleepActivity` whenever the bitmap size differs from the screen size, not only when it exceeds it.
* Add an **opt-in `allowUpscale` flag** to `GfxRenderer::drawBitmap()` (defaults to `false`). With the default, the pre-existing downscale-only behavior is preserved for all other callers (`BmpViewerActivity`, `BaseTheme` cover thumbnails, Lyra themes), so undersized bitmaps continue to render at their native size within the target bounds. `SleepActivity` opts in, since its CONTAIN/CROP semantics deliberately want the sleep image to fill the screen.
* Bypass the 1-bit fast path when `allowUpscale=true` so 1-bit sleep bitmaps authored for a different panel size go through the span-aware 2-bit code and are scaled up to fill the screen. When `allowUpscale=false` (default) the fast path is unchanged — `drawBitmap1Bit` and all its existing callers keep their current behavior.
* Replace the 1:1 source-pixel loop with a **dest-span fill** (nearest-neighbor block) so every destination pixel is covered on upscale. Each source pixel now writes to the span
  `[floor(idx * scale), floor((idx + 1) * scale) - 1]`,
  which collapses to a single pixel when `scale ≤ 1.0`, so downscale and no-scale behavior are bit-for-bit identical to master.

### Why opt-in upscale rather than always upscale

Earlier revision of this PR made `drawBitmap()` upscale whenever `fitScale != 1.0f`. CodeRabbit correctly flagged that this silently changed behavior for callers like `BmpViewerActivity` that pass page/tile dimensions as bounds and relied on 1:1 rendering for undersized bitmaps. Moving the upscale decision to an explicit flag keeps the fix surgical: only `SleepActivity` gets the new behavior, all other callers — including the 1-bit fast path — are unchanged at the API default.

### Test case

X4-styled `sleep.bmp` (480×800) dropped onto an X3 device at 528×792 with `sleepScreenCoverMode=CROP`:

* **Before:** image rendered at 480×800 in the top-left of the screen with blank strips on the right and bottom; at 1:1 pixel mapping.
* **After:** image is scaled by `max(528/480, 792/800) ≈ 1.1` and cropped around the scale-locked axis, filling the full 528×792 panel, same visual model as `CROP` on larger-than-screen bitmaps. Tested with both 2-bit greyscale and 1-bit monochrome source BMPs.

Downscale (bitmap larger than screen) renders identically to master; other callers (`BmpViewer`, cover thumbnails) are unchanged at the API-default `allowUpscale = false`.

## Additional Context

* The new loop body is `O(src_w * src_h * scale²)`; scale is typically very close to 1 in practice, so the runtime cost is negligible. For 1:1 or downscale the behavior is bit-for-bit identical to the old loop since `dest_span` collapses to 1.
* API-level default is source-compatible — the new `allowUpscale` parameter has a default that matches the previous (downscale-only) behavior, so existing callers compile unchanged.
* Release-notes-worthy behavior change: users whose custom `/sleep/*.bmp` wallpapers were intentionally smaller than the panel and relied on native-size centering will now see them scaled up in CONTAIN/CROP modes. This is consistent with those modes' documented semantics, but is worth mentioning for anyone tracking 1:1-centered sleep image behavior.
* Originally discovered and fixed against a downstream X3 fork (`crosspoint-reader-x3-vi`) while shipping an X3 build using an X4-styled sleep image; cleaned up and cherry-picked here against upstream `master` with no fork-specific content.

---

### AI Usage

The diagnosis of the three underlying issues and the structure of the dest-span fix were proposed with the help of an AI assistant; the final code, validation, and testing on the target device were done by me. The `allowUpscale` opt-in refactor and the 1-bit fast-path bypass were prompted by CodeRabbit's review of earlier revisions and were written and verified by me.

### DIFF
--- a/lib/GfxRenderer/GfxRenderer.cpp
+++ b/lib/GfxRenderer/GfxRenderer.cpp
@@ -633,10 +633,13 @@ void GfxRenderer::drawIcon(const uint8_t bitmap[], const int x, const int y, con
 }
 
 void GfxRenderer::drawBitmap(const Bitmap& bitmap, const int x, const int y, const int maxWidth, const int maxHeight,
-                             const float cropX, const float cropY) const {
+                             const float cropX, const float cropY, const bool allowUpscale) const {
   if (fontCacheManager_ && fontCacheManager_->isScanning()) return;
-  // For 1-bit bitmaps, use optimized 1-bit rendering path (no crop support for 1-bit)
-  if (bitmap.is1Bit() && cropX == 0.0f && cropY == 0.0f) {
+  // For 1-bit bitmaps, use optimized 1-bit rendering path (no crop support for 1-bit).
+  // Skip the fast path when the caller opts in to upscaling, since drawBitmap1Bit() only
+  // downscales; falling through to the span-aware 2-bit path here lets 1-bit sleep images
+  // authored for a different panel size fill the screen.
+  if (bitmap.is1Bit() && cropX == 0.0f && cropY == 0.0f && !allowUpscale) {
     drawBitmap1Bit(bitmap, x, y, maxWidth, maxHeight);
     return;
   }
@@ -664,9 +667,15 @@ void GfxRenderer::drawBitmap(const Bitmap& bitmap, const int x, const int y, con
     hasTargetBounds = true;
   }
 
-  if (hasTargetBounds && fitScale < 1.0f) {
-    scale = fitScale;
-    isScaled = true;
+  // Downscale whenever the cropped bitmap overflows the target bounds (pre-existing behavior).
+  // Upscale is opt-in via `allowUpscale`: callers with CONTAIN/COVER semantics (SleepActivity)
+  // set the flag so an X4-sized sleep image on an X3 panel fills the screen in CROP mode, while
+  // callers like BmpViewer keep native-size rendering for undersized bitmaps.
+  if (hasTargetBounds) {
+    if (fitScale < 1.0f || (allowUpscale && fitScale > 1.0f)) {
+      scale = fitScale;
+      isScaled = true;
+    }
   }
   LOG_DBG("GFX", "Scaling by %f - %s", scale, isScaled ? "scaled" : "not scaled");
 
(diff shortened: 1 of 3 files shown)


giangthb:
Thanks for the review — good catch. You're right that letting `drawBitmap()` upscale for any `fitScale != 1.0` would silently change behavior for `BmpViewerActivity` (and other callers like `BaseTheme` covers and the Lyra themes) that pass page/tile dimensions as bounds and rely on 1:1 centering for undersized bitmaps.

I've gone with your second suggestion (renderer-side opt-in flag) in a force-push:

- Added `bool allowUpscale = false` as the last parameter of `GfxRenderer::drawBitmap()`.
- Default behavior is unchanged: only downscale is applied, undersized bitmaps still render at native size within the target bounds.
- `SleepActivity::renderBitmapSleepScreen()` is the only call site that opts in (passes `allowUpscale=true`), since its CONTAIN/CROP semantics deliberately want the sleep image to fill the screen.
- `BmpViewer`, `BaseTheme` cover thumbnails, and the Lyra themes are untouched and keep their pre-existing behavior.
- The dest-span fill in the render loop is kept; it's still bit-for-bit identical to master when `scale ≤ 1.0`.

Also called out in the updated PR description that this is a release-notes-worthy behavior change for users whose custom `/sleep/*.bmp` wallpapers are intentionally smaller than the panel — they'll now see them scaled up in CONTAIN/CROP, which is consistent with those modes' semantics but different from the previous center-at-native-size behavior.

Force-pushed as a single squashed commit on top of current `master`.

giangthb:
Good find — yeah, I missed that `drawBitmap1Bit()` is a separate fast path that never upscales, so a same-aspect undersized 1-bit sleep bitmap would still land at 1:1 despite `SleepActivity` setting `allowUpscale=true`.

Fixed as you suggested:

```cpp
if (bitmap.is1Bit() && cropX == 0.0f && cropY == 0.0f && !allowUpscale) {
  drawBitmap1Bit(bitmap, x, y, maxWidth, maxHeight);
  return;
}
```

When `allowUpscale=true`, 1-bit bitmaps now fall through to the span-aware 2-bit path and get the same dest-span fill as 2-bit/4-bit images. When `allowUpscale=false` (default, i.e. every caller except `SleepActivity`), the fast path is unchanged — no impact on `BmpViewer`, `BaseTheme` covers, or the Lyra themes.

Also updated the commit message/PR description to call this out as the fourth interacting issue so reviewers have the full context.

Force-pushed on top of `master`.

PR #1730, opened 2026-04-21
chore: git pre-commit hook for format fix

osteotek:
## Summary

* Adds pre-commit git hook that automatically runs ./bin/clang-format-fix

---

### AI Usage

### DIFF
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -1,0 +1,13 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+# Run formatter from repository root regardless of current directory.
+REPO_ROOT="$(git rev-parse --show-toplevel)"
+cd "${REPO_ROOT}"
+
+echo "Running clang-format fix before commit..."
+./bin/clang-format-fix
+
+# Ensure formatting changes are included in the pending commit.
+git add -u
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -13,23 +13,14 @@ while IFS= read -r -d '' file; do
   staged_files+=("${file}")
 done < <(git diff --cached --name-only -z --diff-filter=ACMR)
 
-# Skip formatter invocation when there are no modified C/C++ files.
-format_candidates="$(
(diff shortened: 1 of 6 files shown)


itsthisjustin:
Oh heck yeah haha

itsthisjustin:
You did this differently than I expected but I guess this actually makes more sense. This runs locally vs in our CI pipeline? I was originally thinking we'd just do this on a worker before a merge vs running it before I commit on my own machine.

znelson:
@itsthisjustin I know you want this. I want this. Go, or no-go (revise)?

znelson (approved):
I like it and I think it fixes the constant churn we see with PRs needing formatting.

How does one with an old repo clone add the pre-commit rule?

osteotek:
I can try to implement @itsthisjustin suggestion, with running in CI

@znelson in existing repo it's `git config core.hooksPath .githooks && chmod +x .githooks/pre-commit`

znelson:
Honest question from someone without a lot of experience in other projects. Is format-on-commit rule good enough? I lean toward yes. (And I think having a tool make changes after you push sounds kind of scary, tbh.)

osteotek:
> Honest question from someone without a lot of experience in other projects. Is format-on-commit rule good enough? I lean toward yes. (And I think having a tool make changes after you push sounds kind of scary, tbh.)

Should be enough. And it's easy to disable if commit hook doing something wrong

itsthisjustin:
I'm good with either setup but I find git commit hooks can sometimes cause weird issues depending on how you're comitting is all. Personal experience being unable to commit because some stupid linter failed.

itsthisjustin:
> > Honest question from someone without a lot of experience in other projects. Is format-on-commit rule good enough? I lean toward yes. (And I think having a tool make changes after you push sounds kind of scary, tbh.)
> 
> Should be enough. And it's easy to disable if commit hook doing something wrong

Lets just get this in then. We can always improve it later

itsthisjustin (approved):

znelson:
Very excited for this, thanks again @osteotek!

PR #1478, opened 2026-03-24
feat: Clippings - EPUB clippings (multi-page capture)

tnaftali:
## Why this relates to PR #1337 ([feat: epub bookmarks](https://github.com/crosspoint-reader/crosspoint-reader/pull/1337))

This branch extends the **bookmarks** reader work: same list patterns, `Page`/`TextBlock` text extraction, and menu integration. It is easiest to review **after** #1337 is familiar.

**Clippings-only diff vs the bookmarks branch** (for reviewers who want to skip re-reading bookmark commits):

`https://github.com/crosspoint-reader/crosspoint-reader/compare/vedi0boy:feature/bookmarks...tnaftali:feature/clippings-on-bookmarks`

(Compare across forks: bookmarks head → this branch.)

## Summary

- **Multi-page capture** while reading: save passages to `.crosspoint/clippings/` on the SD card.
- **Reader menu:** **Capture**, **Clippings** (list, full-text viewer, delete).
- **UX:** Status bar shows capture progress; **page back** finishes capture and saves **without** turning the page; **Confirm (menu)** or **short Back** **cancel** an in-progress capture; save/fail uses a centered popup like bookmarks.
- **List:** Ordered by position in the book; previews and metadata in subtitles.

## Screenshots

Reader menu

Clippings list

Capturing (status bar)

Clipping saved (popup)

Clipping detail (viewer)

Delete confirmation

## Documentation

- `USER_GUIDE.md` — **§7 Clippings (captures)**

## Testing

- [x] `pio run` (local)
- Capture / cancel / save / list / delete on device

PR #1402, opened 2026-03-15
feat: dictionary support

mcrosson:
## Summary

- General 'dictionary support' discussion has started in #1599 
  - Please use this discussion for bug reports, usage questions, etc. 
- User manual ; see `docs/dictionary.md` ([link](https://github.com/mcrosson/crosspoint-reader/blob/feat-dictionary/docs/dictionary.md))
  - This document contains more information than stated in this PR summary
- Developer documentation ; see `docs/dictionary-development.md` ([link](https://github.com/mcrosson/crosspoint-reader/blob/feat-dictionary/docs/dictionary-development.md))
  - This document contains more information than stated in this PR summary
- Compared to PR #857 
  - Full feature parity is complete
  - This was the starting point for this PR's work
- Compared to PR #1542
  - Full feature parity is complete
  - There may be performance differences but they are minor. See below discussion for additional detail.
- StarDict support: syn files, offset files, ifo metadata
  - Only dict/dz + idx files needed for dictionary use
- On-device dictionary pre-processing before use: decompression + offset file generation
  - Required if supplied dictionary is compressed (dz) or lacking offset files (oft)
- Offline pre-processing script (Python, run before transfer to device)
- Script that allows merging multiple StarDict dictionaries
  - This should be used if you need 'multiple active dictionaries'
- View dictionary ifo metadata on-device (parsed and raw views)
- Select a dictionary from a list that is built using `.dictionaries` or `dictionaries` folder on sd card
- Per-book dictionary override (reader menu)
- Word lookup: direct, case-normalized, stem variants, alt-form (.syn), fuzzy suggestions
- Phrase / multi-word lookup
- Lookup from definition (chained lookups without returning to reader)
- Lookup history with tunable max number of entries
- HTML rendering of dictionary data (expat-based, dedicated renderer)
- IPA pronunciation characters / info supported (see `docs/dictionary.md` for details)
- Test infrastructure: 19 dictionaries + 20 chapter test epub

## Additional Context

See the discussion linked above for more detail on progress, planned changes, etc

### General

- *This implementation is functionally complete*
- This branch is kept up to date with functional code. Major breakages should *not* happen.
- PR's to this PR are welcome
- Bug reports are welcome

### RAM / ROM Usage Differences

| Metric | crosspoint | feat-dictionary | Delta |
|--------|----------------|-----------------|-------|
| **RAM** | 95,212 B (29.1%) | 96,004 B (29.3%) | **+792 B (+0.2%)** |
[...]

PR #1742, opened 2026-04-22
feat: text clipping selection and saving in EPUB reader

cagliostro1991:
**What does this PR do?** This PR introduces text clipping (highlighting) functionality to the EPUB reader. The goal is to allow users to select a range of words on a page and save the selection to a Kindle-compatible `My Clippings.txt` file on the SD card.

**Summarize the main changes made:** This PR adds a new `ClipSelectionActivity` for word-by-word selection, a `ClippingsManager` for persisting clips, exposes word position data from `TextBlock`, and wires up the full flow from the reader menu through to file storage.

## Changes Made

• **ClipSelectionActivity:** New activity that renders a word-level cursor over the current page. The user navigates word-by-word (or line-by-line with long-press), sets a start mark with Confirm, then selects the end word to confirm the clip. Supports multi-page selection across up to 3 pages. Saves the framebuffer to heap to avoid re-rendering on each highlight update.

• **ClippingsManager:** New `saveClipping()` static method that appends the selected text to `/My Clippings.txt` on the SD card in Kindle-compatible format (title/author header, chapter + page location, quoted text, `==========` separator).

• **GfxRenderer / HalDisplay — windowed refresh:** Exposes `displayWindow(x, y, w, h)` (previously commented out as experimental) through the HAL layer, enabling partial e-ink refreshes aligned to 8-pixel boundaries.

• **TextBlock — word position exposure:** Added `getWordXpos()` and `getWordStyles()` accessors so `EpubReaderActivity` can reconstruct per-word bounding boxes for the selection UI without re-parsing layout.

• **EpubReaderActivity — SAVE_CLIPPING menu action:** Handles the new menu item by collecting word refs for up to 3 pages, launching `ClipSelectionActivity`, and calling `ClippingsManager::saveClipping()` on a non-cancelled result.

• **EpubReaderMenuActivity:** Added `SAVE_CLIPPING` menu item with `STR_SAVE_CLIPPING` i18n key.

• **i18n — STR_SAVE_CLIPPING:** Added `STR_SAVE_CLIPPING` translation key to all 24 supported language YAML files.

## Dependencies

• **Dependent PRs:** None.

• **External Dependencies:** None.

## Testing

• **New unit tests added:** No.

• **Existing unit tests updated:** No.


pablohc:
Amazing work @cagliostro1991! I just watched the demo video and I really like how it works. Just one question, since I haven't tried it on the device yet: Do the front buttons advance one word at a time, and the side buttons advance one line at a time?

cagliostro1991:
Thanks a lot @pablohc! Actually, it works a bit differently from what you described. I have to change it during dev tests. I plan to make it configurable if needed.
Now, both the side buttons (Up/Down) and the front buttons (Left/Right) move the cursor one word at a time with a single press. When you hold any of those buttons, the cursor jumps to the end of the line (forward) or the start of the line (backward). 
So it's tap = word, hold = line,  regardless of which button you use.
Selection memory has some limitations, and now it has only 3 screens for selection. Possibly make it configurable, but more requires some temporary caching in the file system.

pablohc:
Oh, the implementation is much better @cagliostro1991 I like it, and it doesn't force you to change to another buttons.

pablohc:
> Thanks a lot @pablohc! Actually, it works a bit differently from what you described. I have to change it during dev tests. I plan to make it configurable if needed. Now, both the side buttons (Up/Down) and the front buttons (Left/Right) move the cursor one word at a time with a single press. When you hold any of those buttons, the cursor jumps to the end of the line (forward) or the start of the line (backward). So it's tap = word, hold = line, regardless of which button you use. Selection memory has some limitations, and now it has only 3 screens for selection. Possibly make it configurable, but more requires some temporary caching in the file system.

I've been testing it and it looks great! Thanks so much @cagliostro1991 

Just a few things to note:
- The word selector has a cursor that's larger than the "selected word+the space bar", and it often overlaps the next word. > **Maybe limit it so that only the current word is highlighted, without any spaces?**
- I enable clipping selection mode, scroll forward two pages, and in the last paragraph of the third page, I confirm to start selecting by line. When it goes to the next page (outside of memory), it shows everything selected from the first page where the mode was enabled up to the current page. > **Maybe we shouldn't let things go any further so as not to cross the line?**
- When I select by holding down the button, it moves one line at a time, but the last word is highlighted first, and then the selection is completed with the preceding words on that same line. I know this is just a visual effect, but I'm mentioning it in case you want to reproduce it. 
- I had a crash; I’ll attach it below, but I’ll try to reproduce what I was doing when it happened. 
[crash_report.txt](https://github.com/user-attachments/files/27057313/crash_report.txt)

lizsugar:
First off, this is fantastic!!  I've been putting off starting a specific book because I know that I'll be highlighting passages in it and wanted a way to do that!  Thank you for this!

The format of My Clippings.txt is a bit different than that produced by my Kindle Keyboard:

**From my Kindle:**
```
Lift A Love Story (Anne Dey)
- Highlight on Page 26 | Loc. 385  | Added on Sunday, June 15, 2025, 06:40 PM

I don't mind messes
==========
```

**From my X4 testing this PR:**
```
Stargate Atlantis: Legacy 1: Homecoming / Graham, Jo, Scott, Melissa
Chapter: Chapter Twenty-six | Page 29

"“That’s a decoy, they’re trying to draw us out."

==========
```

- The author is represented different (parentheses vs. slash).
- The entire line about the location is just different.  I understand that Kindle "locations" and the date are basically not gonna happen on the x4 (well, actually diving into it, the [formula is known](https://wiki.mobileread.com/wiki/Page_numbers#Implementation) and locations could be implemented in a separate PR), but it should at least read like `- Highlight on Page 29, Chapter Twenty-Six` in the x4 example. 
- The clip itself has quotations around it that are not present in the text (you can see the one "fancy" quote that is present in the text though).

I bring these up because something here is causing a tool I use to parse these files to barf on the one generated by the x4 with this PR.  I imagine there's a lot of tools people use to parse My Clipings.txt, and anything kindle compatible would be expected to work in those tools.

Also, just a question:  Do you think it's possible to display which passages have been highlighted like the Kindle does, with an underline or something?  This would probably require locations to be implemented, but I'm just wondering aloud here.

cagliostro1991:
@pablohc @lizsugar Thanks for your review! I've incorporated your feedback into the latest updates. The file has been tested on [clippings.io](https://clippings.io) if you have time, plz re-test on your end as well. Also improved cursor movement handling and added crash logging to help diagnose any edge cases.

lizsugar:
Just tried this out and yes this looks and functions a lot better!  Thank you so much!  I have to modify my extraction script because it's looking for the datestamp that kindle puts into My Clippings, which I understand cannot be done on the x4 because there is no RTC.  That's just a limitation of the device and I can modify my script easily enough to account for that.  Thank you so much!

lizsugar:
I just wanted to add: underlining highlights in the book is just sublime. Absolutely fantastic! Thank you so much!

pablohc:
These are amazing improvements, this is a truly useful addition the underlining highlights. It looks and feels solid. Thanks, @cagliostro1991!
I have a few more suggestions, but I won't bother you with them. You have my approval once you've resolved the merge conflicts with the master branch.
I'm having some issues with my personal computer at the moment, but I'll try to get back in the swing of things this weekend and show you if I can implement them in a test branch. Excellent work!

lizsugar:
I've started seeing a weird behavior ever since flashing this PR yesterday: using the side buttons, sometimes when pressing the down/next button it will then act as if I pressed the up/previous button right after it. This happens in books and in the main menu too. I took a video that's too large to attach here, so I've hosted it elsewhere: 

https://files.lizsugar.me/PXL_20260502_060815297.TS.mp4

I'm currently running build

1.2.0-dev-detached-ea19e85

Again this only started happening yesterday after I flashed this feature. Let me know what you need me to do to test or get information! Thank you!

pablohc:
Hi @lizsugar, could you enable the serial port monitor to check if that pulse triggers it? Remember to compile in the Default environment so the logs are visible.
Yesterday I did some more in-depth testing and reviewed the code with Claude, and I noticed similar behavior when accessing the Clipping settings: releasing the Select button changes the first navigation option. I hope to finish a few more tests and share the details.

cagliostro1991:
@pablohc, @lizsugar,
Tracked down the control issue and reworked that part to make the behavior more predictable and easier to maintain. On top of that, I polished the translations and cleaned up a few minor things along the way.

From my side, it looks solid, and I would stop here. If possible, it would be great to include this in the upcoming release.

Please give it a test!

lizsugar:
I just tried the latest build, it's better, but the side buttons are still malfunctioning in the same way, but it's less frequent now.  It's super weird.  I won't be able to set up a dev environment and logging until monday or tuesday (USA central time) at the earliest.

lizsugar:
Alright, bad news for me!  I flashed release version 1.2.0 on my device and the button issue persisted.  I then flashed the English stock firmware and the _button issue persisted_.  This appears to be a weird intermittent hardware issue in my device.  I'm so sorry to hold this up without doing proper testing.  It only started after a recent flash of this PR, but I think the timing was just very unfortunately coincidental.  I do not believe there is a button issue in your code!  So, go ahead and disregard that and I'd love to see this move forward!

pablohc:
Hi @cagliostro1991, sorry for the delay in getting feedback to you.

The crash happened again the first time I opened the Reader Menu to select "Create Clipping" after flashing the firmware, but it didn't happen on subsequent attempts. I suppose the problem is caused by the settings being in a previous state without "Create Clipping" option, which triggers it.
[crash_report.txt](https://github.com/user-attachments/files/27435362/crash_report.txt)

One issue I've been having in my test branch too (https://github.com/pablohc/crosspoint-reader/tree/test/epub-text-clippings) is when I change certain text formatting settings, such as Hyphenation, Extra Paragraph spacing, or Embedded Style. Any annotation in the binary file becomes out of sync with the content, so I would suggest that if any of these formatting options are changed, the underline in the EPUB should be removed, while keeping the annotation in clipping.txt

Before change formatting:
[Call-Me-Ishmaelle_ch4_p1_0pct_90709.bmp](https://github.com/user-attachments/files/27435380/Call-Me-Ishmaelle_ch4_p1_0pct_90709.bmp)

After change formatting:
[Call-Me-Ishmaelle_ch4_p1_0pct_75955.bmp](https://github.com/user-attachments/files/27435389/Call-Me-Ishmaelle_ch4_p1_0pct_75955.bmp)

I also thought it would be useful for people who take frequent notes to have quick access via the power button, so that it triggers "Create Clipping" if that option is selected under Settings > Controls > Short press power button.

I've added this feature along with a few other improvements in my test branch, please feel free to check out in case you find anything useful there.

pablohc:
Hi @cagliostro1991, thank you so much for merging my changes into the PR!! ❤

I've edited the message so that it isn't interpreted as a flaw in the PR itself, but rather as an issue caused by the user performing unexpected actions, such as changing the formatting (font, size, hyphenation and paragraph spacing) of a book while reading it and highlighting sections.

Similarly, if this were to happen, the ultimate goal would be to obtain the clipping.txt file, and the current export format works perfectly well, so there would be no risk.

However, if that were a problem, a more granular, per-book settings option would solve it, provided the user keeps the current settings for that book between highlighting sessions.

itsthisjustin:
I THINK we have two PRs related to clippings? #1478

pablohc:
> I THINK we have two PRs related to clippings? #1478

Hi @itsthisjustin, although both use "clipping" as a common point, they are two different approaches:
#1478 basically acts as a "bookmark", allowing you to save the position of the entire page.
#1742 has a more specific purpose: it lets you "highlight" phrases and save the clipping as a .txt file.

@cagliostro1991 has done an incredible job here, and it is currently in a solid state.

itsthisjustin:
@cagliostro1991 This is a pretty big feature to add that will end up being out of scope. Considering Pablo has tested and used it I'm considering keeping into our new Phase 0 of legacy roadmap items but I'd need the conflicts solved (again). I know this is SUCH a pain but if you can merge that would be good. I will try to get some more eyes on this to make a decisions shortly

pablohc:
@cagliostro1991 If you need help, we can look at it together

iandchasse:
There might be some UI redundancy with #1402 when it comes to the text selection. Unsure if identifying this would help either PR before one is merged before the other.

pablohc (approved):
I tested it on the device after the last commit and it's working as expected. Thank you very much for updating the PR @cagliostro1991

itsthisjustin:
> There might be some UI redundancy with #1402 when it comes to the text selection. Unsure if identifying this would help either PR before one is merged before the other.

Important thing to consider. @mcrosson

cagliostro1991:
I looked at the overlap with #1402's `DictionaryWordSelectActivity` + `WordSelectNavigator`.

Both use word-cursor navigation over a page, but the outputs are completely different: this PR writes to `My Clippings.txt`, a Kindle-standard format parsed by Readwise, KOReader, and dozens of third-party tools. Dictionary lookup is self-contained. They're complementary features that happen to share a UI pattern, not competing ones.

If anything, `ClipSelectionActivity` could serve as a foundation for future text-selection features: different annotation types, underline styles, highlights, and so on. Dictionary word select stays focused on lookup. Two different directions built on the same idea.

pablohc:
It should also be noted that #1402 contains 15,000 lines of code and over 200 modified files.
Meanwhile, the current PR contains 1,600 lines, and of the 50 modified files, half are UI translations.

mcrosson:
> I looked at the overlap with #1402's `DictionaryWordSelectActivity` + `WordSelectNavigator`.
> 
> Both use word-cursor navigation over a page, but the outputs are completely different: this PR writes to `My Clippings.txt`, a Kindle-standard format parsed by Readwise, KOReader, and dozens of third-party tools. Dictionary lookup is self-contained. They're complementary features that happen to share a UI pattern, not competing ones.
> 
> If anything, `ClipSelectionActivity` could serve as a foundation for future text-selection features: different annotation types, underline styles, highlights, and so on. Dictionary word select stays focused on lookup. Two different directions built on the same idea.

I agree with complimentary. If you have a good, speedy, performant base for highlight of text that can do the following, #1402 can be adjusted once this is merged.

- single word nav + return of the word on select
- way to highlight multiple words, across line boundaries, and return the full set of words highlighted

dictionary needs are satisfied with these two. we dont need cross page boundaries but if they are available, we'd likely implement that too.

cagliostro1991:
Following up on the overlap discussion with #1402 — I've refactored `ClipSelectionActivity` into a reusable text-selection foundation (commit pushed to the branch).

**What changed**

- `WordRef` is now a standalone struct in its own header, usable independently of this activity
- Text processing logic (em-space stripping, hyphen merging, paragraph detection, anchor building) extracted into `ClipTextBuilder` — callable separately, independently testable
- `WordSelectResult` added for callers that only need the selected text without full clipping metadata
- `ClipSelectionActivity` accepts a `Config` struct:

```cpp
struct ClipSelectionConfig {
    enum class Mode { CLIPPING, WORD_SELECT };
    Mode            mode   = Mode::CLIPPING;  // default: existing behaviour
    ClipRenderConfig render;                  // visual style
};
```

**Flexible rendering via `WordStyle` bitmask**

Each visual state (cursor, selection) has its own combinable flags:

```cpp
struct ClipWordStyle {
    enum Flags : uint8_t {
        FILL      = 1 << 0,  // fillRectDither background
        INVERT    = 1 << 1,  // solid black + white text
        UNDERLINE = 1 << 2,  // line at font baseline
        BORDER    = 1 << 3,  // outline rect
    };
    uint8_t flags   = FILL;
    Color fillColor = Color::LightGray;
};
```

**For #1402 dictionary integration**

`@mcrosson` — based on your comment, here's what dictionary would look like:

```cpp
ClipSelectionConfig cfg;
cfg.mode = ClipSelectionConfig::Mode::WORD_SELECT;
cfg.render.cursor    = { ClipWordStyle::INVERT };
cfg.render.selection = { ClipWordStyle::FILL | ClipWordStyle::UNDERLINE, Color::LightGray };

startActivityForResult(
    std::make_unique<ClipSelectionActivity>(
        renderer, mappedInput, std::move(words),
        "", "", "", 0, fontId, section, startPage, mTop, mLeft, cfg),
    [](const ActivityResult& result) {
        if (result.isCancelled) return;
        const auto& sel = std::get<WordSelectResult>(result.data);
        // sel.text — selected words, sel.fromWordIdx / sel.toWordIdx
    });
```

Single-word: navigate to word → confirm twice (sets mark, immediately confirms one-word range).
Multi-word across lines: works as-is.

Default `Config{}` reproduces the original clipping behaviour exactly — no functional change to this PR.

jenningsb2:
Hey — first off, thanks for getting highlights/clippings off the ground here; the anchor-based selection is solid and I built directly on top of it.

I've been extending this on a branch and I'd like to suggest steering the feature in that direction, as I think it ends up a bit more elegant and more broadly useful. Short demo video: https://cleanshot.com/share/SNqSWm7V

The main differences:

- Stable IDs + full text per highlight. Each highlight gets a unique id (and stores its own text), which makes everything downstream trivial — no fuzzy text-matching to find or remove a highlight.
- A Highlights menu. Lists every highlight in the book with a text preview + chapter/page. Confirm jumps straight to that spot in the book; hold-Confirm deletes it (mirrors the existing Bookmarks UX).
- Export decoupled from creation. Instead of appending to My Clippings.txt on every highlight, creation just records the highlight, and there's an on-demand Export Highlights action that writes TXT and/or JSON (selectable in settings). The JSON makes it easy to pull highlights into Readwise/Obsidian/etc.
- Deletion that just works, since it's removeById rather than rewriting an append-only log.

I also tracked down a perf issue where creating a clipping took ~10s with custom SD-card fonts (glyph cache thrashing) and fixed it — it's effectively instant now.

Happy to open this as a PR (or against this branch) so it's easy to review side by side. Curious what you and others think about going this route.

lizsugar:
> Export decoupled from creation. Instead of appending to My Clippings.txt on every highlight, creation just records the highlight, and there's an on-demand Export Highlights action that writes TXT and/or JSON (selectable in settings). The JSON makes it easy to pull highlights into Readwise/Obsidian/etc.

I mean, it's done the way it is now, I understand, in order to maintain compatibility with Kindle clippings, a format and workflow that has not changed much for at least 16 years (that's when my Kindle Keyboard released). A lot of people with ereaders already have tooling and workflows around that format. Simply being able to pull the file without having to manually export every single time is part of the ease and usability that made that workflow so reliable for so many people. 

Fixing the slowness with sd fonts is A+ though!

jenningsb2:
@lizsugar Valid point! Im not married to the export format and flow that I landed on.

The core enhancements that my branch enables that I'd love to see supported in vanilla crosspoint are: 

- handle deleting highlights
- Ability to view book level highlights from the reader view (like with the new bookmark feature)
- ability to jump to specific highlights

---
I also added some double click shortcuts to enable jumping to the bottom and top of the page in the cursor mode which makes highlighting the right content faster.

jenningsb2:
@lizsugar —I think I've worked out a happy medium if you want to give my fork a spin! More context in the PR linked: https://github.com/jenningsb2/crosspoint-reader/pull/1

lizsugar:
Alright, I finally got platform.io to install on my windows machine (my linux one hated it even more ???) and built and flashed your latest (I've been otherwise relying on the CI/CD artifacts on this project PR).

I have not tested the dedicated export function, it's just.. not how I use this feature.  I'm glad it's there for people who want it.  

- "Append to My Clippings.txt" option: I toggled it on and it is appending perfectly to my existing file
- double taps for quicker navigation is amazing!!  This is probably as close as we can get to the old kindles that had the 5-way input
- sped up clipping with sd fonts is, it's everything to me.  I stopped reading with custom fonts for a while because it was so slow before
- the highlights menu is a wonderful little surprise!!  it even shows all my highlights from before this new version  **

** There is an issue with the jumping to highlights via the menu:  If you make a highlight with one font and then switch fonts, the jump points don't work right anymore.  It appears to be going to, for example, page 59, whereas after indexing the new font the highlight is _now_ on page 53.  I do believe this to be out of scope of this particular PR?  I think it's because CrossPoint simply doesn't do a total page count per book yet.  Kindle highlights mark their location using a global location and page ("Location 206-7, Page 14") whereas this PR does it by page within a chapter ("Page 101 | Part 2").

I think that simply requires a new location/page indexing system and might be out of scope here.  I think I've seen some PRs or discussions about making something similar though.  Mobilereads.com has documentation on the Kindle location system: https://wiki.mobileread.com/wiki/Page_numbers#Implementation

cagliostro1991:
@jenningsb2 Really like where you took this — stable highlight IDs, the Highlights menu with jump/delete, and that glyph-cache perf fix (10s → instant) are exactly the polish it needed. And @lizsugar's call on Kindle `My Clippings.txt` as an opt-in toggle alongside the clean TXT/JSON export feels right.

Let's converge into one PR instead of two parallel forks. Since your work builds on the selection here, opening your branch **against this one** (as you offered) stacks it as additive commits on top — one coherent review. I'll handle the latest `master` resync to clear the conflicts so your commits land clean, keeping #1742 as the single Phase 0 entry point @itsthisjustin mentioned.

Let's get it moving 🚀

cagliostro1991:
Master resync done ✅ — the branch is now on latest `master` (1.4.1), conflicts resolved, PR is mergeable again and CI is green.

A few notes on how this feature coexists with the new 1.4.x functionality, since a lot landed in the meantime:

- **Long-press Confirm** (new `longPressMenuFunction`: KOSync / bookmark / disabled from #1808) is untouched — clipping is triggered via the reader menu entry and the short power button option, so there's no gesture collision.
- **`shortPwrBtn` enum**: `SAVE_CLIPPING` is appended as the last option, so existing saved settings keep their indices.
- **Cache formats**: this branch never changed `section.bin` (was still v25 = base); master's bump to v27 merged in cleanly, and annotations live in their own file — no version collision.
- **NFC normalization** actually helps here: annotation anchors and rendered page words now go through the same normalization pipeline, so matching stays consistent.
- **Render path**: master's `renderContents` refactor (font prewarm + reworked grayscale flow) merged in; annotation underlines draw into the BW buffer after the main render. I'll verify on device that underlines survive the grayscale pass with `textAntiAliasing=on`.

@jenningsb2 the branch is clean now — feel free to open your PR against `feature/epub-text-clippings` so your commits stack on top as we discussed. One heads-up: master's new font prewarm (scan pass + glyph cache warm-up) may already cover your glyph-cache fix, so that part might shrink or drop entirely.

@itsthisjustin the condition from your earlier comment is met: conflicts are solved and the branch tracks latest master. Also, #1402 has since been closed, so the text-selection UI redundancy concern is off the table — the remaining overlap question is only convergence with #1478. Happy to align on the Phase 0 decision whenever you get more eyes on this.

itsthisjustin:
> @itsthisjustin the condition from your earlier comment is met: conflicts are solved and the branch tracks latest master. Also, #1402 has since been closed, so the text-selection UI redundancy concern is off the table — the remaining overlap question is only convergence with #1478. Happy to align on the Phase 0 decision whenever you get more eyes on this.

Absolutely hate to do this to you, but you need to rebase to develop instead of master :(

cagliostro1991:
@itsthisjustin no worries, done. Base switched to `develop`, branch synced, mergeable again, build is green (RAM 32.4%).

Some porting was needed since develop moved a lot under this feature. The `paragraphStart` flag from clippings is folded into the new flattened `TextBlock` layout (#2547), and since that makes the on-disk block differ from upstream v29 I bumped `SECTION_FILE_VERSION` to 30 so caches regenerate instead of colliding. Clip selection reads pages through `loadPage()` now that #2452 removed `loadPageFromSectionFile()`. Also picked up the `freeink-sdk` migration (#2449), submodule pinned to the same commit as develop.

Checked Selection Popup (#2358) too - it's the settings `OptionPopup`, no overlap with the text selection here.

itsthisjustin:
> @itsthisjustin no worries, done. Base switched to `develop`, branch synced, mergeable again, build is green (RAM 32.4%).
> 
> Some porting was needed since develop moved a lot under this feature. The `paragraphStart` flag from clippings is folded into the new flattened `TextBlock` layout (#2547), and since that makes the on-disk block differ from upstream v29 I bumped `SECTION_FILE_VERSION` to 30 so caches regenerate instead of colliding. Clip selection reads pages through `loadPage()` now that #2452 removed `loadPageFromSectionFile()`. Also picked up the `freeink-sdk` migration (#2449), submodule pinned to the same commit as develop.
> 
> Checked Selection Popup (#2358) too - it's the settings `OptionPopup`, no overlap with the text selection here.

Thanks so much. I will look into reviewing this as soon as I can.

jenningsb2:
Hey @cagliostro1991 I finally got around to pulling this in. Let me know if I need to make any changes. Hopefully, we can get this live!

cagliostro1991:
> Hey @cagliostro1991 I finally got around to pulling this in. Let me know if I need to make any changes. Hopefully, we can get this live!

Awesome, thanks for pulling it in! Just synced the branch with the latest `develop` and pushed - no conflicts, build's green. Let me know if anything needs adjusting on my end, happy to get this landed.

cagliostro1991:
@jenningsb2 any updates?

bloodysummers:
@cagliostro1991 I've been testing this for a few days on an X4, works well overall. A few things still open; I've left out what @jenningsb2's PR already covers (highlights menu, stable ids, deletion, decoupled export).

- **Clip selection ignores the reader orientation.** In landscape the side buttons still move by line and the front ones by word. `DictionaryWordSelectActivity` already solves this. Bind to `Btn::ScreenLeft/ScreenRight/ScreenUp/ScreenDown` and use `mapDirectionalLabels()` for the hints. Only the directional mode needs it, `CLIP_NAV_CONTINUOUS` drives `NavNext/NavPrevious`, which already flip.
- **No long-press shortcut.** Worth adding as a `longPressMenuFunction` option. Careful: it can't just be appended to the enum. `buildLongPressMenuValues()` hides the home-key-only `READER_MENU` by trimming the tail, and load clamps to the list size, so an appended value is unreachable on boards without a home key (the X4 among them). Inserting it before `READER_MENU` works.
- **Highlights disappear when the layout changes.** Anything that re-paginates (orientation, font size, spacing, margins) moves the passage to another page, and the `curPage < rec.sectionPage || curPage > rec.endSectionPage` gate drops the record before any anchor is tried. Everything below that line is already layout-independent, so deleting the gate fixes it. The same root cause has a second symptom in @jenningsb2's PR: the Highlights menu jumps with `ProgressChangeResult{sectionIdx, sectionPage}`, so after a re-pagination the entry is found but the jump lands on the wrong page. The stored pages are still worth keeping for the clippings file, just not as a constraint.
- **Anchor resolution can select the wrong span.** Start and end are searched independently (leftmost / rightmost), so a short passage occurring several times on a page gets underlined from the first occurrence to the last, swallowing everything between. A 2-word clipping of mine matched three phrases at once, and started one word early on another. Pairing by length fixed it: for each start match the end sits `wordCount` words later and the end anchor has to match exactly there, which rules out cross-occurrence pairs. The context words then work well as a tie-break. The `startIdx = 0 / endIdx = size()-1` fallbacks want the same bound, or they underline to the page edge whenever one anchor misses.
- **<sstream> costs ~167 KB of flash.** The four anchor finders build a `std::istringstream` per call, and this file is the only user of <sstream> in the firmware. Replacing them with a plain whitespace split measured flash 5,892,889 → 5,721,913 bytes (89.9% → 87.3% of app0) and static RAM 59,156 → 54,516. It matters more once the gate is gone, since the finders then run for every annotation on every render.
- **Highlight texts stay in DRAM all session (new in @jenningsb2's PR).** `load()` reads `clipText` into every record and the set stays resident for the underline pass, so a book with a few dozen highlights holds tens of KB against the ~50 KB of free heap a reading session leaves. They're only needed by the menu and by export, seeking past the text in `load()` and streaming it back when the menu opens keeps the format unchanged.
- **Placement could be exact instead of inferred.** The section cache already persists a `uint32_t visible-text` offset per page, and the reader uses it to keep the reading position across re-pagination (`rememberCurrentContentOffset()`). That's what Kindle's `Location` is, and why its page numbers are only approximate. Storing it in `AnnotationRecord` as the primary anchor, with text matching as the fallback for older records, would make placement exact, and would let the export carry a real Location field. (CrossInk does a weaker version: a `layoutSignature` next to stored word indices, falling back to searching the full clipped text.)

I have most of these on a branch already, device-tested on an X4: the orientation bindings, the long-press option, dropping the page gate and the `<sstream>` replacement. Removing the gate is what confirmed highlights survive both an orientation change and a font-size change, and it is also how I hit the wrong-span problem above, since the gate had been hiding it.
The length-pairing fix for that one is written and building but not verified on hardware yet. If any of it is useful I'm happy to open a PR against `feature/epub-text-clippings`, or rebased onto whatever base you land on after the resync, one concern per commit, so you can take only the parts you want.

liketheduck:
@cagliostro1991 , any progress recently? Would love to see this in the next release!

PR #1741, opened 2026-04-22
feat: Add royalty.dev funding integration

itsthisjustin:
Adds royalty.dev as a funding platform for contributors by adding custom funding link and badge to README

A maintainer would need to click the "Are you the maintainer?" button on https://app.royalty.dev/crosspoint-reader/crosspoint-reader to install the github app which pulls the PRs in and ranks them. 

(Disclosure, Royalty is a product I've recently built and am using for my own OSS and currently gating custom font builds and nightlys on crosspointreader.com)

This page currently can collect donations but without the integration it wont notify contributors and it wont rescore fresh PRs.

PR #1745, opened 2026-04-22
Revert "feat: Add royalty.dev funding integration"

znelson:
Reverts crosspoint-reader/crosspoint-reader#1741

I don't entirely understand Royalty.dev and want to vet this further.


Uri-Tauber:
@znelson This actually sounds like a really cool idea. 
I’d love to hear more about how the platform is built, though. If it’s Claude-coded, it might not be the best choice for handling credit card information.

PR #1747, opened 2026-04-23
feat: TXT reader options menu + byte-offset navigation

eunchurn:
## Summary

* **What is the goal of this PR?** Adds an in-reader options menu and faster, offset-based navigation to the TXT reader, bringing it close to feature parity with the EPUB reader.
  * **`perf: switch TXT reader to byte-offset navigation + add justified alignment`** — Replaces the up-front `buildPageIndex()` whole-file scan with byte-offset navigation. Opening a 40MB TXT no longer hangs the UI; the status bar shows an exact `%` (currentOffset / fileSize) and an N/M estimate seeded from the first rendered page. Previous-page nav uses a bounded history stack (cap 256) and falls back to a forward-walk from a guessed earlier offset when the stack is empty. `progress.bin` is now ~40 bytes (settings + current offset) instead of `numPages × 4 + header`. Also adds proper justified alignment for TXT (was silently treated as left-align): a parallel `vector<bool>` marks each line as a soft-wrap continuation vs. a paragraph-end line, and renderPage distributes `(contentWidth - textWidth)` across inter-character gaps via `drawText`'s `letterSpacing` on wrapped lines only.
  * **`feat: add reader options menu to TXT reader`** — Adds `TxtReaderMenuActivity` (mirroring `EpubReaderMenuActivity`'s style) opened on `Confirm`. Menu items:
    * **Reading orientation** — Portrait / Landscape CW / Inverted / Landscape CCW (persists to `SETTINGS.orientation`)
    * **Auto page turn** — OFF / 1 / 3 / 6 / 12 pages-per-minute. While active, the status bar title is overridden with the current rate.
    * **Long-press page jump** — OFF / 10 / 20 / 50 / 100. When enabled, navigation buttons (`Left`/`Right`/`PageBack`/`PageForward`) switch to release-based detection: short tap = 1 page, hold ≥ 1s = configured step. Uses `estBytesPerPage * step` for the offset estimate, then snaps to the nearest line start.
    * **Go to %** — Reuses `EpubReaderPercentSelectionActivity` for the slider UI; `jumpToPercent()` translates to a byte offset and snaps to a line start.
    * **Take screenshot** — `pendingScreenshot` flag handled at the end of `render()` via existing `ScreenshotUtil`.
    * **Go home** — Exit reader.
    * Adds `pageJumpOption` to `MenuResult` (EPUB ignores it).
    * New i18n key `STR_PAGE_JUMP_STEP`.

## Additional Context

* **Performance**: The byte-offset rewrite is the headline win — large TXT files (10MB+) now open instantly instead of blocking on a full scan. Backward navigation past the 256-entry history uses a coarse forward-walk windowed to ~2× the running per-page byte estimate, so deep back-nav stays bounded and lands within a page of the true previous page.
* **Memory**: All new menu state is on the stack or in small members on `TxtReaderActivity`; no new heap allocations in the hot path. The new `vector<bool>` `currentPageLineEndsParagraph` mirrors `currentPageLines` (already heap-allocated), so it scales with `linesPerPage`, not file size.
* **Auto-turn / page-jump options are reader-session ephemeral** (not written to global settings). Orientation is persisted because EPUB does the same, and we don't want it diverging across formats.
* **Power button** page-turn role (`SETTINGS.shortPwrBtn == PAGE_TURN`) is preserved while page-jump is enabled.
* **Areas to focus review on**:
  * The `loop()` branch that swaps press- vs. release-based nav based on `currentPageJumpOption` — want to make sure it doesn't regress quick-tap responsiveness for users who leave page-jump off.
  * `findBackwardPageStart()` — happy-path is fine, but the `estBytesPerPage == 0` first-frame edge case is worth a glance.
  * Justified alignment letterSpacing clamp to `int8_t ±127` — only a problem on pathologically short lines, but flagged in a comment.
* **Cache compatibility**: `progress.bin` schema changed (different magic + version). Existing TXT progress files will be ignored on first open after upgrade and the reader resets to offset 0, which is the same behavior as any other layout-setting change today.

---

### AI Usage

Did you use AI tools to help write this code? _**YES**_ — Claude was used to draft the menu activity, the long-press input branch, and percent/jump helpers; the byte-offset navigation and justified-layout design were human-driven, with AI assisting on cleanup. All changes were reviewed and verified locally on hardware before submission.


src/activities/reader/TxtReaderActivity.cpp
```diff
@@ -4,22 +4,124 @@
...
 #include <I18n.h>
+#include <Logging.h>
 #include <Serialization.h>
 #include <Utf8.h>
 
+#include <algorithm>
+
 #include "CrossPointSettings.h"
 #include "CrossPointState.h"
+#include "EpubReaderPercentSelectionActivity.h"
```
znelson, line 15: We should rename to just ReaderPercentSelectionActivity (but doesn't have to be in this change).

itsthisjustin:
This seems to be of high value to the ecosystem and imo just needs conflicts solved and technical validation. I won't approve yet because it'll just go stale but I'd like to surface this in Discord or Zulip to get some more eyes on it as it does add 1k LOC

PR #1760, opened 2026-04-24
fix: Exit reader activity for kosync and return when done

itsthisjustin:
KOReaderSyncActivity now saves progress directly to a binary file and navigates back to the reader using ActivityManager instead of returning results through callbacks. This simplifies the sync flow by removing the ActivityResult pattern and intermediate state management in EpubReaderActivity.

mbedTLS is very memory hungry. Exiting the reader activity clears sufficient RAM for TLS handshake activity during KOReader sync.


jpirnay (dismissed):
Lgtm

mcrosson:
I agree with code rabbit: can there be some kind of error/notice when the progress fails to save? Preferably something user facing.

mcrosson:
@itsthisjustin ; can you cleanup the conflicts on this PR?

mcrosson (approved):

wylanswets:
@mcrosson and @znelson I just tried the changes from this PR, and I found 2 issues:

1.  I can not sync with the book I was having trouble with using this branch.
2. It doesn't keep the orientation of the prior menu (my device is in landscape, this switches back to portrait)

I just re-tried the fix I implemented from https://github.com/crosspoint-reader/crosspoint-reader/pull/1855 and it keeps the orientation in place and it syncs.

wylanswets:
The results I'm getting between the two branches is a difference in approach. My branch fully removes the book from memory to free up a lot of ram while dealing with syncing. It doesn't seem to be removing the book from ram using this branch.

itsthisjustin:
> The results I'm getting between the two branches is a difference in approach. My branch fully removes the book from memory to free up a lot of ram while dealing with syncing. It doesn't seem to be removing the book from ram using this branch.

Is there any downside of removing the book from ram? Admittedly I don't use landscape so that's an interesting observation.

mcrosson:
> > The results I'm getting between the two branches is a difference in approach. My branch fully removes the book from memory to free up a lot of ram while dealing with syncing. It doesn't seem to be removing the book from ram using this branch.
> 
> Is there any downside of removing the book from ram? Admittedly I don't use landscape so that's an interesting observation.

So long as progress saves successfully it shouldn't be an issue. Itd be like going home and then re-opening the book.

wylanswets:
> > > The results I'm getting between the two branches is a difference in approach. My branch fully removes the book from memory to free up a lot of ram while dealing with syncing. It doesn't seem to be removing the book from ram using this branch.
> > 
> > 
> > Is there any downside of removing the book from ram? Admittedly I don't use landscape so that's an interesting observation.
> 
> So long as progress saves successfully it shouldn't be an issue. Itd be like going home and then re-opening the book.

Exactly the thinking, just reload the book once we need it again so that the ram can be freed for the syncing work. In my PR the position was pre-calculated and then the book closed so we could just compare local to remote. When making a decision that's whats applied upon re-opening the book.

mcrosson:
I've re-opened #1855 -- It's looking like both of these PRs do useful things with regard to kosync.

I think blending both together would be the right path forward for what these two PRs are trying to accomplish.

wylanswets:
@itsthisjustin 

> Admittedly I don't use landscape so that's an interesting observation.

I was only using landscape because the x4 I got has some broken columns I had to work around until my replacement arrived 😂

mcrosson:
@wylanswets just stated they will work to merge this PR with #1855 so we get both sets of ram gains in a single PR. We can leave this open for now but there should be a new PR that supersedes both at some point.

wylanswets:
@mcrosson and @itsthisjustin I have opened https://github.com/crosspoint-reader/crosspoint-reader/pull/1860 - can you please check it over?

mcrosson:
Superceded by #1860

PR #1768, opened 2026-04-27
fix: python requirements files

steka:
## Summary
The Python scripts in the current repo have many requirements that are not mentioned in any requirements.txt files, so I have therefore added them to the directories that need them.

Question:  Should these changes maybe moved into a "root" requirements.txt file?

---
### AI Usage

### DIFF
--- a/lib/EpdFont/scripts/requirements.txt
+++ b/lib/EpdFont/scripts/requirements.txt
@@ -1,2 +1,2 @@
-fonttools==4.62.1
-freetype-py==2.5.1
+fonttools>=4.62.1
+freetype-py>=2.5.1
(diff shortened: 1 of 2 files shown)


Uri-Tauber (dismissed):
Nice. However, do we need those specific versions, or can we use any version >= N?

steka:
@Uri-Tauber I just selected the latest version. I don't know the minimum version that works.

Uri-Tauber:
@steka I think it might be better to allow versions >= the current ones, so developers aren’t locked to a specific version and can benefit from future updates and fixes.

Uri-Tauber (approved):

PR #1771, opened 2026-04-27
feat: Add button lock feature with short power button press

marcinoktawian:
## Summary

* Implements a button lock feature that suspends input handling without powering off, allowing users to safely transport the device and resume reading immediately

PXL_20260428_090347769 MP
PXL_20260428_090429980 MP

* Added `buttonsLocked` and `buttonsLockedDisplayed` state flags to `CrossPointState` for persistent lock status
* Extended `SHORT_PWRBTN` enum in `CrossPointSettings` with new `LOCK_BUTTONS` option
* Visual indicator: popup displays "Buttons locked" when feature is active
* All buttons except power button are blocked when locked

## Additional Context

* Lock popup is drawn only once when state changes (not every frame) to avoid display refresh loops that were observed in testing
* Power button check happens before the lock block, ensuring unlock is always possible
* When locked, `activityManager.loop()` is skipped entirely, preventing activity processing while displaying the lock indicator

**Potential future enhancements:**
* Support for double-click on power button instead of short-press (discussed but deferred)

---

### AI Usage


marcinoktawian:
Error during turning off on reader book goes to the end

PR #1780, opened 2026-04-29
feat: self-heal from transient WiFi loss, add dBm indicator during WebServerActivity 

jeremydk:
## Summary

I was seeing hangs during File Transfer. Troubleshooting via serial showed it had to do with the SHUTTING_DOWN state.  Strengthened the WiFi State Machine a bit and added self-healing.  Additionally, I found it obnoxious to need to keep referring to serial for my wifi strength, so a dBm meter is added opposing the SSID when on the File Transfer page.

The dBm meter was at risk of causing rapid screen updates (if, say, we hovered right around a threshold), so I've implemented a basic hysteresis around this. RISING/FALLING would be more canonical variable names, but are reserved in the framework.

## Additional Context

My X4 has a terrible antenna, apparently, and I was running into this failure condition pretty regularly.  Wifi "bars" were chosen as a relatively pan-cultural glyph rather than relying on localization.

Tested on hardware at -83 dBm under sustained EPUB upload (60 books, ~30 MB). Transient losses up to ~14s now ride through; pre-fix the same losses required a power-cycle after as little as a 2s blip.

---

### AI Usage

Did you use AI tools to help write this code? Nope. Commits are hand written. Claude was used to build a local test harness for validation only during iteration.

### DIFF
--- a/src/activities/network/CrossPointWebServerActivity.h
+++ b/src/activities/network/CrossPointWebServerActivity.h
@@ -49,7 +49,11 @@ class CrossPointWebServerActivity final : public Activity {
   unsigned long firstDisconnectAt = 0;
   static constexpr unsigned long WIFI_ABANDON_MS = 5UL * 60UL * 1000UL;
 
+  // Cached signal-strength bracket (0..4) for the WiFi indicator.
+  int lastWifiBars = 0;
+
   void renderServerRunning() const;
+  void renderWifiIndicator(int subHeaderTop) const;
 
   void onNetworkModeSelected(NetworkMode mode);
   void onWifiSelectionComplete(bool connected);
--- a/src/activities/network/CrossPointWebServerActivity.h
+++ b/src/activities/network/CrossPointWebServerActivity.h
@@ -44,6 +44,11 @@ class CrossPointWebServerActivity final : public Activity {
   // Performance monitoring
   unsigned long lastHandleClientTime = 0;
 
+  // Sustained WiFi-loss tracking; abandon only after WIFI_ABANDON_MS.
+  int consecutiveDisconnects = 0;
+  unsigned long firstDisconnectAt = 0;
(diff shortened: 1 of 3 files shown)


jeremydk:
Sorry for the title noise, tried to retrigger and clear the formatter check to no avail offhand.  Since it's just the PR title check, I'll let that sit and await review/discussion.  Thanks folks, project has been wonderful so far.

src/activities/network/CrossPointWebServerActivity.h
```diff
@@ -44,7 +44,16 @@ class CrossPointWebServerActivity final : public Activity {
   // Performance monitoring
   unsigned long lastHandleClientTime = 0;
 
+  // Sustained WiFi-loss tracking; abandon only after WIFI_ABANDON_MS.
+  int consecutiveDisconnects = 0;
+  unsigned long firstDisconnectAt = 0;
+  static constexpr unsigned long WIFI_ABANDON_MS = 5UL * 60UL * 1000UL;
```
Uri-Tauber, line 50: Why wait 5 minutes before abandoning?

jeremydk: My thinking here is that sustained loss isn't necessarily a terminal state - it can be  - I'm guilty of walking off with the device still in Web Server mode, that will clearly drain battery.

But, since it isn't asserted to be a terminal state (i.e., wifi can still recover), this is just a balancing act of accidental battery drain versus retry time.

5 min seemed like a reasonable approach given the usual lockscreen timeouts.

Uri-Tauber (approved):
I didn't tested on my device, since I don't use the Wi-Fi functionality at all, but the code seems fine.

Uri-Tauber:
@crosspoint-reader/firmware-maintainers Could someone test this PR on their device, and verify it indeed working without any problems?

znelson (approved):

PR #1786, opened 2026-04-30
feat(update): SD-card firmware update + X3 bootloader compatibility

eunchurn:
## Summary

  Make firmware updates work reliably on every CrossPoint device — including X3 hardware shipped with the older OEM bootloader, and X4 silicon where the running ESP-IDF's `esp_image_verify` rejects our images during `esp_ota_set_boot_partition`. After this PR, the same `firmware.bin` boots regardless of how it was delivered (web flasher, SD card, or OTA), and OTA / SD updates use the same flash code path as the web flasher.

  1. **`scripts/patch_firmware_image.py` + `scripts/verify_firmware_image.py`** — post-build pipeline. The patch script clears `min_efuse_blk_rev_full` / `max_efuse_blk_rev_full` to `0` / `0xFFFF` inside `esp_app_desc_t`, recomputes the segment XOR checksum and SHA256 trailer; the verify script enforces those values shipped. Wired into `platformio.ini` `extra_scripts`.
  2. **`src/network/OtaBootSwitch.{h,cpp}`** — drops a fresh `esp_ota_select_entry_t` into the inactive otadata slot. CRC is over `ota_seq` only (matches `esp_flash_partitions.h` "CRC32 of ota_seq field only" and the existing web flasher behaviour).
  3. **`src/network/FirmwareFlasher.{h,cpp}`** — shared helper. Opens an SD-card path, validates `ESP_IMAGE_HEADER_MAGIC`, streams bytes into the next OTA app partition with **interleaved 64 KiB block-erase + 4 KiB writes** (so the e-ink progress bar advances 0→100% smoothly instead of stalling on a single up-front erase), then calls `OtaBootSwitch`.
  4. **`SdFirmwareUpdateActivity`** (new) — Settings → System → "SD Card Firmware Update" picks any `.bin` (via a new `FileBrowserActivity::Mode::PickFirmware`), confirms, flashes, auto-restarts. Reachable as a recovery boot mode (`UP + POWER` held at startup) for devices where USB enumeration is unreliable.
  5. **`OtaUpdater::installUpdate`** — replaces `esp_https_ota_*` with a two-phase flow:
     - **Phase 1** — `esp_http_client_perform` + event handler streams the GitHub release `firmware.bin` straight to `/.crosspoint/ota_firmware.bin` on the SD card. Same proven pattern `checkForUpdate` already uses; redirects + chunked encoding are handled by the HTTP client.
     - **Phase 2** — `firmware_flash::flashFromSdPath` flashes the cached SD file via the shared helper.

     A `Phase` enum drives the screen label (`Downloading…` vs `Updating…`), and a `lastError` string surfaces the failing step (`sd_open`, `http_status:404`, `flash:WRITE_FAIL`, …) on the Update Failed screen so users without serial output can still report what broke.
  6. **Auto-restart** after success on both SD and OTA paths, with a new `STR_RESTARTING_HINT` ("Restarting… If it does not restart, hold the power button for a few seconds.") replacing the old "Press and hold power button" message.

## Why

The X3 ships with an older second-stage bootloader that misreads two newer `esp_app_desc_t` fields and rejects the image with `Image requires efuse blk rev >= vXXX.YY` on boot. On X4 silicon, a separate bug in the running ESP-IDF's `esp_image_verify` rejects the same image during `esp_ota_set_boot_partition` — the reported "required" version differs build to build, suggesting the function reads from a build-specific offset inside `app_elf_sha256[]`.

The web flasher (raw bytes via esptool-js + direct otadata write) has been working on both chips, which proves the bootloader itself accepts the image. So this PR routes SD update and OTA through the same flow:

- **`patch_firmware_image.py`** keeps `firmware.bin` itself X3-friendly so the X3 bootloader accepts it without runtime patching.
- **`FirmwareFlasher`** + **`OtaBootSwitch`** sidestep `esp_ota_set_boot_partition`'s `esp_image_verify` so the X4-side validation no longer matters.

The intermediate SD copy on the OTA path is also a usability win: if anything dies mid-flash (power loss, brownout, magnetic-USB cable wobble), the cached file is still there for the user to retry through the SD update menu.

## How to Use

Background: Xteink has [disabled USB flashing](https://www.reddit.com/r/xteinkereader/comments/1szk0pq/flashing_of_firmwares_disabled/), but the OEM bootloader itself still allows freely flashing both the stock firmware and CrossPoint firmware from an SD card. Holding the left side button + power button at boot lets the OEM bootloader flash a new firmware from the SD card. However, CrossPoint firmware had no SD-card-based firmware update flow of its own, which is what this PR adds.

### Installing on a USB-flash-disabled device: X3 stock -> CrossPoint

1. Copy `firmware.bin` to the SD card root and rename it to `update.bin` (must be `1.2.0-ko.15` or newer. tested in Korean version).
2. Hold the left side button and long-press the power button to boot. The OEM bootloader will detect `update.bin` and flash it automatically.

### Rolling back to stock: CrossPoint -> X3 stock

1. Copy the X3 OEM firmware ([V5.4.3-X3-CH-PROD-0425_160909.bin](https://domestic-static-file.oss-cn-hangzhou.aliyuncs.com/admin_uploads/firmware/202604/27/751e134f-22b1-4a00-bbfa-0942593ef867/V5.4.3-X3-CH-PROD-0425_160909.bin)) anywhere on the SD card.
2. Go to Settings → System → SD Card Firmware Update and select the file.

## Additional Context

- **Flash size impact**: roughly +6 KiB binary text (helpers + activity), no measurable impact on free heap. The X3 patch script runs at build time only and adds nothing to the image.
- **Backwards compatibility**: existing OTA users see the same "Settings → Update" entry; the only user-visible change during a successful OTA is the new "Downloading…" label appearing before "Updating…", and the auto-restart at the end (previously the screen said "Press and hold power button"). On failure they now see a short error tag (`http_perform:ESP_ERR_TIMEOUT` etc.) instead of a generic "Update failed".
- **Recovery mode**: holding `UP + POWER` while booting now skips home/reader and lands directly in the SD picker. Useful on devices where USB enumeration is unreliable; otherwise reachable normally from Settings → System.
- **Tested on**: Xteink X4 (chip rev 1.30) — web flasher → boot OK → SD update with current `firmware.bin` → boot OK; OTA → download → flash → boot OK. X3 path was validated against the existing X3 fleet via the recovery boot mode (`UP + POWER`) and the patched `firmware.bin` flowing through the OEM bootloader.

---

### AI Usage

Did you use AI tools to help write this code? _**PARTIALLY**_ (Claude Code)


itsthisjustin:
If that sd path exists, once we are on crosspoint we can have OTA channels for releases. Thats FANTASTIC news if the locked devices support sd card flashing out of the box

Uri-Tauber:
@eunchurn, do you know if XTeink locked these new batches at the software level (by removing USB/JTAG code from the OEM firmware) or at the hardware level via eFuse locking (`DIS_USB_JTAG`).
If we flash CrossPoint via the SD card and the USB JTAG starts working again, we’ll know it was just software. Any insight on this from your testing?

itsthisjustin:
> @eunchurn, do you know if XTeink locked these new batches at the software level (by removing USB/JTAG code from the OEM firmware) or at the hardware level via eFuse locking (`DIS_USB_JTAG`). If we flash CrossPoint via the SD card and the USB JTAG starts working again, we’ll know it was just software. Any insight on this from your testing?

from the discord, after flashing stock english, it still doesn't enable usb debug

eunchurn:
> @eunchurn, do you know if XTeink locked these new batches at the software level (by removing USB/JTAG code from the OEM firmware) or at the hardware level via eFuse locking (`DIS_USB_JTAG`). If we flash CrossPoint via the SD card and the USB JTAG starts working again, we’ll know it was just software. Any insight on this from your testing?

The USB still does not work even after updating the X3 to CrossPoint firmware using an SD card. It seems to be a hardware-based issue rather than a software-based one. I plan to open up the device to check if it is eFuse locking (`DIS_USB_JTAG`).

eunchurn:
> If that sd path exists, once we are on crosspoint we can have OTA channels for releases. Thats FANTASTIC news if the locked devices support sd card flashing out of the box

On the X3, you can switch from the Store firmware to the CrossPoint firmware using the left side button and the power button, and then switch back from CrossPoint firmware to Store firmware via SD card flashing.

However, assuming that USB access is blocked on the X4, I haven’t been able to find a way to flash from the Store firmware to the CrossPoint firmware. I also couldn’t find any key combination for it.

Do you happen to know a method?

itsthisjustin:
> > If that sd path exists, once we are on crosspoint we can have OTA channels for releases. Thats FANTASTIC news if the locked devices support sd card flashing out of the box
> 
> On the X3, you can switch from the Store firmware to the CrossPoint firmware using the left side button and the power button, and then switch back from CrossPoint firmware to Store firmware via SD card flashing.
> 
> However, assuming that USB access is blocked on the X4, I haven’t been able to find a way to flash from the Store firmware to the CrossPoint firmware. I also couldn’t find any key combination for it.
> 
> Do you happen to know a method?

Nope but I'll spend the weekend finding one! This is amazing news

itsthisjustin:
If I were to map out how I'd guess this to work on x4. Look at the gpio mappings for each button on x3 and find which of those align to x4 because they are different between the two devices if I recall correctly.

zgredex:
Zrzut ekranu 2026-05-1 o 11 37 49

Quick and dirty LLM analysis. Can someone confirm it?
Zrzut ekranu 2026-05-1 o 11 54 27

eunchurn:
Partial finding from comparing stock V5.5.x X3 vs X4 firmwares — sharing in case it helps with diagnosis, not as a fix for the locked-X4 case.

The X4 stock firmware contains the same SD `update.bin` boot-OTA path as X3 (verified by strings, NVS keys, and webserver endpoints). The trigger is gated by a single NVS flag in `ns1`:

- `activate` (U8) — `1` on X3, `0` on X4 stock

On a `V5.5.3-X4-CH-PROD-0430_213333` unit where USB-JTAG was still reachable, patching that byte (CRC recomputed) at `0x9000` and writing it back was enough to make the X3 button combo (left side + power) load `/update.bin` from the SD card.

This does not help on the production X4 batches discussed in this thread — those have USB-JTAG hardware-disabled (likely `DIS_USB_JTAG` eFuse), so esptool can't reach the chip to patch NVS in the first place. So the firmware-level OTA path is intact; the actual blocker for those units is purely the closed USB.

Sharing only because it narrows the search: on locked X4 units we don't need to find new code paths inside the stock firmware to enable SD OTA — the path is already there. We need a way to flip `ns1/activate` to `1` (or write OTA data directly) without USB. Candidates worth checking:
- the running stock firmware's webserver (`/Put_sdInit`, `/Read_updataAddress`, etc. — does any endpoint write `activate`?)
- on-device button combo before NVS is consulted (i.e. a bootloader-level path, if the bootloader on these units has one)
- direct SPI flash access via test pads

itsthisjustin:
I got the ota flasher working on my x3. Your fix for the min_efuse_blk_rev_full is going in now to see if we can OTA locally the crosspoint firmware so there's an upgrade path for locked devices

itsthisjustin:
> I got the ota flasher working on my x3. Your fix for the min_efuse_blk_rev_full is going in now to see if we can OTA locally the crosspoint firmware so there's an upgrade path for locked devices

Tried this and OTA still fails with the same message.

itsthisjustin:
@eunchurn you gave me a huge lead on the ota update failures so I was able to patch that in and fix the update progress bar too. If you can pull from master now and remove anything related to that max_efuse_blk_rev I think that will make this easier to review. We definitely want to pull this in eventually as SD flashing is awesome. If you don't think my fix fully does what your code does, please let me know.

eunchurn:
> @eunchurn you gave me a huge lead on the ota update failures so I was able to patch that in and fix the update progress bar too. If you can pull from master now and remove anything related to that max_efuse_blk_rev I think that will make this easier to review. We definitely want to pull this in eventually as SD flashing is awesome. If you don't think my fix fully does what your code does, please let me know.

@itsthisjustin Pulled from master. As suggested:

- Reverted `OtaUpdater` / `OtaUpdateActivity` to the simple `esp_https_ota_*` flow from #1805 — the SD-cache hop in our version was a workaround for the same `bootloader_common_check_efuse_blk_validity` issue your runtime wrap now solves.
- Dropped `scripts/patch_firmware_image.py` / `scripts/verify_firmware_image.py` and their `post:` hooks in `platformio.ini`.

What's left is just the SD firmware update feature itself.

A few reasons I think SD-based firmware update is worth keeping alongside network OTA:

- **Version rollback.** Network OTA always pulls the latest GitHub release; there's no in-firmware way to revert to an older build when a release regresses.
- **Forks.** Network OTA is hard-wired to `crosspoint-reader/crosspoint-reader/releases/latest`, so downstream forks (e.g. the Korean localization fork I maintain at `crosspoint-reader-ko/crosspoint-reader-ko`) have no in-firmware install path. Your OTA Unlocker / serial works, but SD flashing covers fork users without leaving the device UI.
- **Offline installs.** First-time setup, factory provisioning, no-Wi-Fi environments.
- **Independent of `api.github.com`** for users behind rate limits, regional blocks, or temporary outages.

SD is a manual escape hatch alongside network OTA, not a replacement.

itsthisjustin:
> > @eunchurn you gave me a huge lead on the ota update failures so I was able to patch that in and fix the update progress bar too. If you can pull from master now and remove anything related to that max_efuse_blk_rev I think that will make this easier to review. We definitely want to pull this in eventually as SD flashing is awesome. If you don't think my fix fully does what your code does, please let me know.
> 
> @itsthisjustin Pulled from master. As suggested:
> 
> - Reverted `OtaUpdater` / `OtaUpdateActivity` to the simple `esp_https_ota_*` flow from #1805 — the SD-cache hop in our version was a workaround for the same `bootloader_common_check_efuse_blk_validity` issue your runtime wrap now solves.
> - Dropped `scripts/patch_firmware_image.py` / `scripts/verify_firmware_image.py` and their `post:` hooks in `platformio.ini`.
> 
> What's left is just the SD firmware update feature itself.
> 
> A few reasons I think SD-based firmware update is worth keeping alongside network OTA:
> 
> - **Version rollback.** Network OTA always pulls the latest GitHub release; there's no in-firmware way to revert to an older build when a release regresses.
> - **Forks.** Network OTA is hard-wired to `crosspoint-reader/crosspoint-reader/releases/latest`, so downstream forks (e.g. the Korean localization fork I maintain at `crosspoint-reader-ko/crosspoint-reader-ko`) have no in-firmware install path. Your OTA Unlocker / serial works, but SD flashing covers fork users without leaving the device UI.
> - **Offline installs.** First-time setup, factory provisioning, no-Wi-Fi environments.
> - **Independent of `api.github.com`** for users behind rate limits, regional blocks, or temporary outages.
> 
> SD is a manual escape hatch alongside network OTA, not a replacement.
> 

Amazing. I'll test tomorrow and I think we should get this merged. Big value to the user

eunchurn:
@itsthisjustin Devices that have USB disabled on the X4 would ideally be flashed to the CrossPoint firmware using the OTA Unlocker you provided, and then use the SD card flasher to install other firmware. This seems to be the only viable way to use the Korean fork version of the CrossPoint Reader.

Alternatively, it would be even better if you could include "CrossPoint Reader KO" among the firmware options in the "Pick a release" section of the OTA Unlocker. Another option would be to allow users to upload or select custom firmware files directly.

Going one step further — what if firmware updates were redesigned around the SD card flasher, splitting OTA into two distinct phases?

**Download Phase.** OTA fetches the `.bin` and writes it to the SD card. No flashing yet; the running firmware is untouched. If WiFi drops, only the partial SD copy is affected.

**Update Phase.** The SD card flasher takes over and writes the `.bin` from SD to the OTA partition. This is the same code path users already hit when they hand-place a firmware file on the card.

Benefits of separating the two phases:

- **Network resilience.** A mid-download disconnect just leaves a partial file on SD; the OTA partition is never half-written. Today's streaming `esp_https_ota_*` path can land the device in that half-written state when WiFi flaps.
- **Single flashing code path.** Online (OTA) and offline (manual SD copy) updates converge on the same raw-write helper. One place to harden, one place a failure can surface.
- **Retryable without re-downloading.** If the Update Phase fails, the user can rerun it against the same SD `.bin` — no need to refetch over the network.
- **Naturally absorbs the "USB disabled" and "custom firmware" cases.** Users who can't run OTA at all, or who want to flash something other than the official release, already use the SD path; this design just makes the official OTA path go through the same machinery.

The Korean fork has been shipping this two-phase flow since 1.2.0-ko.15 (`fix(ota): download firmware to SD then flash via shared raw-write helper`), and in practice it's been noticeably more forgiving of flaky networks than the streaming path.

itsthisjustin:
> @itsthisjustin Devices that have USB disabled on the X4 would ideally be flashed to the CrossPoint firmware using the OTA Unlocker you provided, and then use the SD card flasher to install other firmware. This seems to be the only viable way to use the Korean fork version of the CrossPoint Reader.
> 
> Alternatively, it would be even better if you could include "CrossPoint Reader KO" among the firmware options in the "Pick a release" section of the OTA Unlocker. Another option would be to allow users to upload or select custom firmware files directly.
> 
> Going one step further — what if firmware updates were redesigned around the SD card flasher, splitting OTA into two distinct phases?
> 
> **Download Phase.** OTA fetches the `.bin` and writes it to the SD card. No flashing yet; the running firmware is untouched. If WiFi drops, only the partial SD copy is affected.
> 
> **Update Phase.** The SD card flasher takes over and writes the `.bin` from SD to the OTA partition. This is the same code path users already hit when they hand-place a firmware file on the card.
> 
> Benefits of separating the two phases:
> 
> - **Network resilience.** A mid-download disconnect just leaves a partial file on SD; the OTA partition is never half-written. Today's streaming `esp_https_ota_*` path can land the device in that half-written state when WiFi flaps.
> - **Single flashing code path.** Online (OTA) and offline (manual SD copy) updates converge on the same raw-write helper. One place to harden, one place a failure can surface.
> - **Retryable without re-downloading.** If the Update Phase fails, the user can rerun it against the same SD `.bin` — no need to refetch over the network.
> - **Naturally absorbs the "USB disabled" and "custom firmware" cases.** Users who can't run OTA at all, or who want to flash something other than the official release, already use the SD path; this design just makes the official OTA path go through the same machinery.
> 
> The Korean fork has been shipping this two-phase flow since 1.2.0-ko.15 (`fix(ota): download firmware to SD then flash via shared raw-write helper`), and in practice it's been noticeably more forgiving of flaky networks than the streaming path.

Can you add a catalog endpoint somewhere that follows this spec? https://crossink.uxj.io/catalog

If you give me that link I will add you to the dropdown. Already getting requests from users. I have a hidden debug switch for upload bin but I'm afraid to allow that because if they upload anything that doesn't support ota then they are stuck forever

eunchurn:
> Can you add a catalog endpoint somewhere that follows this spec? https://crossink.uxj.io/catalog
> 
> If you give me that link I will add you to the dropdown. Already getting requests from users. I have a hidden debug switch for upload bin but I'm afraid to allow that because if they upload anything that doesn't support ota then they are stuck forever

@itsthisjustin Here is link https://crosspoint.idlerecord.com/catalog

itsthisjustin:
> > Can you add a catalog endpoint somewhere that follows this spec? [crossink.uxj.io/catalog](https://crossink.uxj.io/catalog)
> > If you give me that link I will add you to the dropdown. Already getting requests from users. I have a hidden debug switch for upload bin but I'm afraid to allow that because if they upload anything that doesn't support ota then they are stuck forever
> 
> @itsthisjustin Here is link [crosspoint.idlerecord.com/catalog](https://crosspoint.idlerecord.com/catalog)

Rad. I'll get this implemented in a few minutes

eunchurn:
> > > Can you add a catalog endpoint somewhere that follows this spec? [crossink.uxj.io/catalog](https://crossink.uxj.io/catalog)
> > > If you give me that link I will add you to the dropdown. Already getting requests from users. I have a hidden debug switch for upload bin but I'm afraid to allow that because if they upload anything that doesn't support ota then they are stuck forever
> > 
> > 
> > @itsthisjustin Here is link [crosspoint.idlerecord.com/catalog](https://crosspoint.idlerecord.com/catalog)
> 
> Rad. I'll get this implemented in a few minutes

@itsthisjustin Thank you for applying this so quickly. The Korean fork firmware follows upstream for most features. However, OTA has been modified to use a download-then-flash process.

I’ve tested this on three devices: one X3 with USB lock enabled, one X3 with the stock (initial) firmware, and one X4 with working USB. For the X4, OTA Unlocker is currently the only viable method, so I really appreciate you creating it.

Also, I would appreciate it if you could consider my earlier suggestion to split the OTA phase into two stages.

itsthisjustin:
> > > > Can you add a catalog endpoint somewhere that follows this spec? [crossink.uxj.io/catalog](https://crossink.uxj.io/catalog)
> > > > If you give me that link I will add you to the dropdown. Already getting requests from users. I have a hidden debug switch for upload bin but I'm afraid to allow that because if they upload anything that doesn't support ota then they are stuck forever
> > > 
> > > 
> > > @itsthisjustin Here is link [crosspoint.idlerecord.com/catalog](https://crosspoint.idlerecord.com/catalog)
> > 
> > 
> > Rad. I'll get this implemented in a few minutes
> 
> @itsthisjustin Thank you for applying this so quickly. The Korean fork firmware follows upstream for most features. However, OTA has been modified to use a download-then-flash process.
> 
> I’ve tested this on three devices: one X3 with USB lock enabled, one X3 with the stock (initial) firmware, and one X4 with working USB. For the X4, OTA Unlocker is currently the only viable method, so I really appreciate you creating it.
> 
> Also, I would appreciate it if you could consider my earlier suggestion to split the OTA phase into two stages.

I think it's a good idea. Check out https://github.com/crosspoint-reader/crosspoint-reader/pull/1810 and see if it maybe makes sense to propose some stuff there?

itsthisjustin:
> Also, I would appreciate it if you could consider my earlier suggestion to split the OTA phase into two stages.

@znelson can you chime in here? I'd like to get this sd firmware path in asap as it provides a recovery path for locked devices

lib/I18n/translations/english.yaml
```diff
@@ -168,6 +168,7 @@ STR_NO_UPDATE: "No update available"
 STR_UPDATE_FAILED: "Update failed"
 STR_UPDATE_COMPLETE: "Update complete"
 STR_POWER_ON_HINT: "Press and hold power button to turn back on"
+STR_RESTARTING_HINT: "Restarting... If it does not restart, hold the power button for a few seconds."
```
znelson:
```suggestion
STR_RESTARTING_HINT: "Restarting... If device does not restart, hold the power button for a few seconds."
```

src/activities/settings/SdFirmwareUpdateActivity.cpp
```diff
@@ -0,0 +1,239 @@
...
+  HalFile file;
+  if (!Storage.openFileForRead("FW", firmwarePath.c_str(), file) || !file) {
+    errorMessage = tr(STR_FIRMWARE_FILE_OPEN_FAILED);
+    return false;
+  }
+  firmwareSize = file.fileSize();
+  file.close();
+
+  // Probe the OTA partition first so the integrity check can also enforce a fits-partition bound.
+  if (!Update.begin(firmwareSize)) {
```
znelson: Instead of Arudino `Update.start` to get the partition size, then immediately aborting, does the following ESP SDK function work?

```
const esp_partition_t* dest = esp_ota_get_next_update_partition(nullptr);
const size_t partitionLimit = dest->size;
```

src/network/FirmwareFlasher.cpp
```diff
@@ -0,0 +1,305 @@
...
+  // Resolve destination first so we can size-check during validation. The full image-integrity
+  // pass below verifies header, segment table, XOR checksum and SHA256 trailer end-to-end before
+  // we touch otadata, so a truncated/corrupted .bin can never become the next boot target.
+  const esp_partition_t* dest = esp_ota_get_next_update_partition(nullptr);
+  if (!dest) {
+    LOG_ERR("FLASH", "no next-update partition");
+    return Result::NO_PARTITION;
+  }
+
+  const Result validateRes = validateImageFile(sdPath, dest->size);
```
znelson: Do we end up validating twice, if we already validated in `SdFirmwareUpdateActivity`?

znelson (commented):
This looks like a great improvement. Couple of small tweaks / questions in comments.

lib/I18n/translations/english.yaml
eunchurn: applied as suggested.

src/network/FirmwareFlasher.cpp
eunchurn: Added an alreadyValidated flag on flashFromSdPath so callers can opt out, but kept it false for the SD path: CodeRabbit pointed out that SD is removable, so the validate -> flash gap is a real TOCTOU window and we want the flasher to re-check. The flag is there for a future caller (e.g. OTA staged to SD with end-to-end hashing) where the second pass really is redundant.

src/activities/settings/SdFirmwareUpdateActivity.cpp
eunchurn: Dropped the `Update.begin/abort` round-trip. `esp_ota_get_next_update_partition(nullptr)->size` has no side effects. Added an explicit size-bound check since we no longer rely on `Update.begin`.

znelson (approved):
This looks good to me. Great work!

@itsthisjustin wdyt?

itsthisjustin:
> This looks good to me. Great work!
> 
> @itsthisjustin wdyt?

yep lets get it in. It's also a fantastic backup path for locked devices. @eunchurn I need your OTA update url for the ko fork to support in the unlocker fyi. Please ping me here or elsewhere with the url so i can add it.

itsthisjustin (approved):

eunchurn:
> > This looks good to me. Great work!
> > @itsthisjustin wdyt?
> 
> yep lets get it in. It's also a fantastic backup path for locked devices. @eunchurn I need your OTA update url for the ko fork to support in the unlocker fyi. Please ping me here or elsewhere with the url so i can add it.

@itsthisjustin thank you justin. for OTA unlocker I gave you this URL <https://crosspoint.idlerecord.com/catalog>. For OTA link for Korean fork is <https://api.github.com/repos/crosspoint-reader-ko/crosspoint-reader-ko/releases/latest>. only use `"name": "firmware.bin"`. can i catch you?

Uri-Tauber:
@eunchurn @itsthisjustin thanks for your work here!
Can someone update the user guide to explain how to use this method?

eunchurn:
> @eunchurn @itsthisjustin thanks for your work here! Can someone update the user guide to explain how to use this method?

I forgot updating `USER_GUIDE.md`. Here is updated USER_GUIDE.md and patch file.

[USER_GUIDE.md](https://github.com/user-attachments/files/27409714/USER_GUIDE.md)
[pr-1786-review-responses.patch](https://github.com/user-attachments/files/27409725/pr-1786-review-responses.patch)

Uri-Tauber:
@@eunchurn can you open new PR?

eunchurn:
> @@eunchurn can you open new PR?
OK I will

eunchurn:
> > This looks good to me. Great work!
> > @itsthisjustin wdyt?
> 
> yep lets get it in. It's also a fantastic backup path for locked devices. @eunchurn I need your OTA update url for the ko fork to support in the unlocker fyi. Please ping me here or elsewhere with the url so i can add it.

@itsthisjustin I realized the KO fork uses a different partition layout.

Devices running the upstream/master nightly firmware have a 6.25 MB OTA partition. The Korean fork exceeds that size because it includes the KoPub Batang font, Korean i18n, and other additions, so we increased the OTA partition to 6.625 MB.

Since the SD-card OTA updater writes against the existing partition table on the device, trying to flash a KO binary larger than 6.25 MB on a stock/master layout fails with Firmware too large for partition.

That also means devices first unlocked/flashed with the standard CrossPoint firmware currently can’t OTA-update to the KO fork directly.

Would it be possible to share the Unlocker source? I think we may need a KO-specific unlocker with the updated partition table.

If possible, I’d also like to fork it for the Korean fork. Is there already a public repo for the Unlocker, and would it be okay for me to fork and distribute a KO-specific version for Korean users?

itsthisjustin:
Source is available! https://github.com/SoFriendly/xteink-unlocker/

itsthisjustin:
> Source is available! https://github.com/SoFriendly/xteink-unlocker/

The issue isn't just crosspoint, the stock ota was failing for ko for the same reason too

eunchurn:
> > Source is available! https://github.com/SoFriendly/xteink-unlocker/
> 
> The issue isn't just crosspoint, the stock ota was failing for ko for the same reason too

That’s strange. In the case of the X3, I was able to successfully flash the KO fork version from the initial store firmware by placing the firmware on the SD card. However, after flashing CrossPoint 1.2.0 and then trying to flash the KO fork version again, it failed due to the same partition-related issue mentioned above. I’ll pull the repo and test it myself. Thank you

znelson:
FWIW: CrossPoint has been conservative about changing partition layout in order to remain compatible with flashing the stock firmware.

PR #1707, opened 2026-04-19
feat: complete Hebrew UI translation (all 296 string keys)

dcherrera:
Hey @Uri-Tauber — you mentioned in #1700 that the i18n strings haven't been translated yet, so here's a complete Hebrew translation covering all 296 `STR_*` keys.

This builds on top of your `split/rtl-2-reader` branch.

> **AI Disclosure:** This PR was developed with the assistance of Claude (Anthropic).

---

## Summary

Replaces the partial 35-key Hebrew stub with a complete translation covering:
- All menus, settings, and reader UI strings
- Network/WiFi screens (connect, scan, hotspot, OPDS, Calibre)
- Status bar, progress, and navigation labels
- Error messages and prompts
- KOReader sync strings
- Button labels and system messages

## Dependencies

- Depends on #1700 (RTL reader support) being merged first

## Notes

- The Hebrew font glyphs needed to render these strings in the UI will be submitted as a separate PR
- Language name displays as "עברית" in the language selector (requires Hebrew glyphs in the Ubuntu UI font to be visible)

## Test plan
- [x] Generated with `gen_i18n.py` — 23 languages, 296 keys confirmed
- [x] Build succeeds
- [x] Tested Hebrew UI on device with Hebrew font glyphs present

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1797, opened 2026-05-01
chore: Removed unused loc CHARACTER_SETS data

znelson:
## Summary

Noticed while digging into other localization cleanup that CHARACTER_SETS is unused. Removing this clears 2,536 bytes of flash.

---

### AI Usage

### DIFF
--- a/lib/I18n/I18n.cpp
+++ b/lib/I18n/I18n.cpp
@@ -108,12 +108,3 @@ void I18n::loadSettings() {
   }
 }
 
-// Generate character set for a specific language
-const char* I18n::getCharacterSet(Language lang) {
-  const auto langIndex = static_cast<size_t>(lang);
-  if (langIndex >= static_cast<size_t>(Language::_COUNT)) {
-    lang = Language::EN;  // Fallback to first language
-  }
-
-  return CHARACTER_SETS[static_cast<size_t>(lang)];
-}
--- a/lib/I18n/I18n.cpp
+++ b/lib/I18n/I18n.cpp
@@ -107,4 +107,3 @@ void I18n::loadSettings() {
     }
   }
 }
-
--- a/lib/I18n/I18n.h
+++ b/lib/I18n/I18n.h
@@ -28,10 +28,6 @@ class I18n {
   void saveSettings();
   void loadSettings();
 
-  // Get all unique characters used in a specific language
-  // Returns a sorted string of unique characters
-  static const char* getCharacterSet(Language lang);
-
  private:
   I18n() : _language(Language::EN) {}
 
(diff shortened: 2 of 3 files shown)


Uri-Tauber:
@znelson I believe I am responsible for creating `compute_character_set`. The logic behind it was that at some point in the future, we would want to load the UI font from the SD card, and in such a case, it would be much more efficient to know **exactly** which glyphs each language requires. This would be super efficient for CJK, as it would allow loading a few dozen glyphs into RAM instead of tens of thousands....

znelson:
@Uri-Tauber We only know which glyphs each language requires for localized strings, not for reading. 

Is the idea that we'd offload the UI font to the SD card? My feeling is that's a step too far and we probably ought to at least include a fallback font with the glyphs necessary for the UI, but probably at least one full font so reading works without jumping through any hoops with the SD card. 

If we include a font for the minimal set of glyphs for the OS, we can compute and flash that at build time.

I'm sort of reluctant to keep speculative data taking up space, when we could add that data later in exactly the format that serves the need.

mcrosson:
I think it wise to do the following

- Stabilize / finish sd font support ( #1693 )
- Look at the other languages that weve been asked to get sorted ( #1757 )
- Loop around to UI / translations / etc in terms of font

I have a feeling the sd fonts and language support will expose a bit more of how we want to approach a cleanup of the included fonts.

IMO, we'd have a set of baseline fonts, one for each alphabet (latin / cryllic / greek / hebrew / arabic / chinese / japanese / korean / thai / mala / etc) and use those on a 'per build' basis. For UI fall back reasons (english being the default translation) we can include a sub-set of latin chars for just the ui with non-latin alphabet fonts.

If possible, we could shift these to the spiffs at flash time so we keep the core firmware trim while providing a 'fundamental font' as a starting point.

On top of that we will be distributing cpfont files via the sd font changes and can easily put forth additional fonts users may want to download and add to the sd card when performing an initial flash -- since we require internet for the initial flash, having a user pull a font for their sd card, at that time, isnt a huge added burden. We also require the sd card for crosspoint to function so we can easily 'require' a font on the sd card if the 'minimal defaults' are insufficient for reading purposes.

znelson:
@mcrosson I can't tell if you're saying "keep this data" or "delete it for now" 🙂

mcrosson:
I'm saying 'leave it be for the moment and loop around soon'. The sd font and language stuff will churn the code some and will unlock added translation options.

When folk start talking 'lets add a non-latin alphabet translation' (hebrew is already on my radar as @Uri-Tauber has it basically done in #1707 ). I think thats when we should discuss more broadly and generally so we can get a win across all languages/alphabets.

znelson:
@Uri-Tauber @mcrosson acknowledged. We should put it on someone's to-do list though, because it can be difficult to come across truly unused things (without a bit of luck), when you're building the next iteration of something. 🙂

mcrosson:
i think i have it on my short list ; ill double check and add so we dont lose sight of UI language 'things'.

PR #1800, opened 2026-05-01
refactor: Account for deduplicated string storage in gen_i18n stats

jpirnay:
This PR applies the single deduplication stats fix. It updates `scripts/gen_i18n.py` to compare the current offset-table layout against a deduplicated pointer-model estimate, adds a per-language unique string size column, and preserves intentional empty translations while using English fallback for missing entries.

### DIFF
--- a/lib/I18n/I18n.cpp
+++ b/lib/I18n/I18n.cpp
@@ -27,7 +27,12 @@ const char* I18n::get(StrId id) const {
 
   // Use generated helper function - no hardcoded switch needed!
   const LangStrings lang = getLanguageStrings(_language);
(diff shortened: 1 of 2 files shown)


znelson:
@jpirnay Please take a look at #1802. We probably still need a tweak to how we report stats during the Python script run. My other PR realizes the savings by omitting English duplicates from flash.

znelson:
I updated the python script statistics in https://github.com/crosspoint-reader/crosspoint-reader/pull/1802/changes/5e949987ff496fedd9ac32dec57cea03103e4160. I think we can close this one now and move forward with #1802.

PR #1801, opened 2026-05-01
fix: prevent chapter skipping after taking a screenshot

WuTofu:
## Summary

Fix the bug where the chapter skip action could still trigger after taking a screenshot.

  - .Updated `main.cpp` so the screenshot-button combo now fully blocks the rest of the main loop until both buttons are released, preventing sleep or other button actions from firing.
  - Removed the redundant screenshot-specific skip guard in `EpubReaderActivity.cpp` and let the new main-loop handling prevent the issue at the source.

---

### AI Usage

### DIFF
--- a/src/activities/reader/EpubReaderActivity.cpp
+++ b/src/activities/reader/EpubReaderActivity.cpp
@@ -203,12 +203,6 @@ void EpubReaderActivity::loop() {
   }
 
   const bool skipChapter = !fromTilt && SETTINGS.longPressChapterSkip && mappedInput.getHeldTime() > skipChapterMs;
-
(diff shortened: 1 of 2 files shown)


Uri-Tauber:
@WuToFu This issue was already addressed in #1625.
You’re likely seeing it on an older firmware version. Please compile and flash the latest commit, then test again to confirm whether the bug still occurs.

WuTofu:
> You’re likely seeing it on an older firmware version.

I was on v1.2.0. Just compiled and flashed the latest commit, and can confirm this bug is gone.

Thanks!

PR #1803, opened 2026-05-01
feat: enhance long press action to delete both files and directories

WuTofu:
## Summary

Enhance file browser long-press behavior so the delete action now works for both files and directories.

Updated `FileBrowserActivity.cpp` to support for directory deletion on long-press

---

### AI Usage

### DIFF
--- a/src/activities/home/FileBrowserActivity.cpp
+++ b/src/activities/home/FileBrowserActivity.cpp
@@ -139,17 +139,20 @@ void FileBrowserActivity::loop() {
       return;
     }
 
-    if (mode == Mode::Books && mappedInput.getHeldTime() >= GO_HOME_MS && !isDirectory) {
-      // --- LONG PRESS ACTION: DELETE FILE ---
+    if (mode == Mode::Books && mappedInput.getHeldTime() >= GO_HOME_MS) {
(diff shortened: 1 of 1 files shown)


WuTofu:
@znelson, @Uri-Tauber, @pablohc, sorry to ping you guys. If your hands are too full for this pr, I'd understand.

Since we already allow users to delete files in FileBrowser, this pr adds deleting directory capability on top of that. So this is just a small change, I'd appreciate it if you could take the time to review it.

Uri-Tauber (approved):
Looks solid. Two notes:

- Deleting a single epub removes its cache folder too — would be nice to handle this recursively for directories, but not a blocker.
- In `removeDir()`, `char name[128]` should probably be bumped to `500`, same as `FileBrowserActivity.cpp:31`.

Nice work!

pablohc (approved):
Thank you so much for this feature, I've been needing it for a while! ;)

WuTofu:
> Deleting a single epub removes its cache folder too — would be nice to handle this recursively for directories, but not a blocker.

Ohh! I was trying to delete my screenshot folders when I made this change. Somehow it never occurred to me that we should handle this recursively.

I will create another pr for that soon

PR #1806, opened 2026-05-02
feat: render <sup> tags as ^ prefix in EPUB parser

### LINKED ISSUE #1696: Superscript is rendered as normal text
### Affected Version

1.2.0

### Describe the Bug

In What If? by Randall Munroe, Numbers like 10¹⁵ are rendered as 1015; see attached screenshot. I checked in KOReader and there it is rendered correctly, so I assume the epub is not at fault here.

I found this PR: https://github.com/crosspoint-reader/crosspoint-reader/pull/463 but apparently it didn't help.(Or was it merged after 1.2.0?)

### Steps to Reproduce

Open a book with superscript text.

### Expected Behavior

Superscript should be rendered correctly, not as normal text.

Image

fain182:
## Summary

Improve readability of EPUB content that uses the `<sup>` tag, which is currently ignored causing superscript text to merge with the surrounding text (e.g. `1015` instead of `10^15`). 
Fixes #1696.
 In `ChapterHtmlSlimParser::startElement()`, added handling for the `<sup>` tag: when the opening tag is encountered, a `^` character is injected into the text stream, maintaining visual continuity with the preceding word via `nextWordContinues = true`.

## Alternatives Considered

* **No handling (status quo):** 
text inside `<sup>` is emitted inline without any separator, making the content unreadable (e.g. `1015`).
* **Reduced font size without vertical offset:** 
recognise `<sup>` and select a smaller font size, but the text would still sit on the same baseline — visually incorrect and potentially misleading.
* **Full implementation (reduced font + raised baseline):** 
would require changes to the layout engine to handle mixed-baseline lines, per-word-token overhead, and a non-trivial refactor.
* **Square brackets `[N]`:** 
considered for footnote references (Markdown/Wikipedia convention), rejected because `^` is more universal and works for both footnotes and mathematical expressions. Footnote links are already visually distinguished by their underline styling.

## Additional Context

* No changes to the layout engine, CSS parser, or font system — zero RAM impact.
* Text inside `<sup>` is already handled by the existing character data flow; no `endElement` handler is needed.

---

### AI Usage

---
_Generated by [Claude Code](https://claude.ai/code/session_01SKExZQDqyed89pkESmybCr)_

### DIFF
--- a/lib/Epub/Epub/parsers/ChapterHtmlSlimParser.cpp
+++ b/lib/Epub/Epub/parsers/ChapterHtmlSlimParser.cpp
@@ -655,6 +655,10 @@ void XMLCALL ChapterHtmlSlimParser::startElement(void* userData, const XML_Char*
     }
     self->inlineStyleStack.push_back(entry);
     self->updateEffectiveInlineStyle();
+  } else if (strcmp(name, "sup") == 0) {
+    self->flushPartWordBuffer();
+    self->nextWordContinues = true;
+    self->characterData(userData, "^", 1);
   } else if (strcmp(name, "span") == 0 || !isHeaderOrBlock(name)) {
     // Handle span and other inline elements for CSS styling
     if (cssStyle.hasFontWeight() || cssStyle.hasFontStyle() || cssStyle.hasTextDecoration()) {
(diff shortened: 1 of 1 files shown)


fnordsh:
Thank you for taking the time to implement this.
While this workaround is certainly better than nothing, as the author of the original issue I would argue that this does not fix it, because superscript is still not rendered correctly, as I think it should be.

znelson:
Closing based on feedback from @fnordsh

PR #1815, opened 2026-05-03
feat: Track reading time

Charlie-83:
## Summary

Add a minimal time tracking feature. Record the time spent reading each book and display this in the Epub menu. Future work can add to this to show total time across all books etc.
`progress.bin` has an additional 4 bytes representing the time spent reading the book in milliseconds. The epub reader menu shows this time in the top row.

## Additional Context

There has been a lot of discussion and PRs around time tracking/reading statistics (#226). Until a more comprehensive solution is merged, I just wanted something simple to start tracking the time: I can manually transfer this data over to any new system.

I don't know if my approach in this PR is the one the project will eventually want to take but I offer it as a starting point from which further contributions can build.

The first question to answer is: Should this kind of "reading statistics" data be storing in the book cache (as is the case in this PR) or should more files be added outside of the cache directory. If the "in-cache" option is taken, the general attitude/advice around the cache would need to change since people will no longer want to delete it to fix issues. 

---

### AI Usage


itsthisjustin:
This is a super interesting case study in the way this community should operate. @Charlie-83 I love your approach to a discussion first mentality...but I don't know if that should live in an open PR especially while there are other competing PRs with the same goal? I'm going to close this for now in the name of cleanup but could you open a discussion and reference this PR there?

PR #1816, opened 2026-05-03
fix: support legacy XTC file headers where pageTableOffset=48

uxjulia:
## Summary

Restore compatibility with older XTC files that use the legacy 48-byte header layout, so they open correctly instead of failing header/page-table validation.

- Updates the XTC parser to accept legacy files where pageTableOffset starts at 0x30 instead of the newer full header size.
- Avoids treating legacy header bytes as a valid chapterOffset, so older files are not misclassified as having chapters.
- Switches XTC file size and seek operations to 64-bit-safe wrapper calls in HalStorage, which matches the XTC format’s 64-bit offset fields and keeps page/chapter lookups from narrowing offsets.
- Keeps the existing bounds checks and improves related logging when XTC page loads fail.

## Additional Context

- This is a narrow compatibility fix for XTC handling. The functional change is in the XTC parser; the HalStorage changes are supporting wrapper methods needed by that parser update.
- Risk should be low outside XTC reading, since the HAL changes only expose existing underlying file APIs rather than changing storage behavior.

---

### AI Usage

### DIFF
--- a/src/activities/reader/XtcReaderActivity.cpp
+++ b/src/activities/reader/XtcReaderActivity.cpp
@@ -179,7 +179,6 @@ void XtcReaderActivity::renderPage() {
   // Load page data
   size_t bytesRead = xtc->loadPage(currentPage, pageBuffer, pageBufferSize);
   if (bytesRead == 0) {
-    // LOG_ERR("XTR", "Failed to load page %lu", currentPage);
     LOG_ERR("XTR", "Failed to load page %lu: bufferSize=%lu bitDepth=%u error=%s", currentPage, pageBufferSize,
             bitDepth, xtc::errorToString(xtc->getLastError()));
     free(pageBuffer);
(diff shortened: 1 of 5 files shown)


itsthisjustin:
@znelson julia flagged that my fix for xtc files boot looping the device accidentally removed support for older xtc file formats so I asked her to pr her patch for this

src/activities/reader/XtcReaderActivity.cpp
```diff
@@ -179,7 +179,9 @@ void XtcReaderActivity::renderPage() {
   // Load page data
   size_t bytesRead = xtc->loadPage(currentPage, pageBuffer, pageBufferSize);
   if (bytesRead == 0) {
-    LOG_ERR("XTR", "Failed to load page %lu", currentPage);
+    // LOG_ERR("XTR", "Failed to load page %lu", currentPage);
```
znelson: Nit: you can remove this old commented LOG

znelson (dismissed):

znelson (approved):

PR #1819, opened 2026-05-04
refactor: Use std::size instead of sizeof/sizeof

znelson:
## Summary

Simplify code calculating compile-time array sizes with `sizeof(array)/sizeof(element)` to use `std::size`.

---

### AI Usage


lib/Epub/Epub/parsers/ChapterHtmlSlimParser.cpp
```diff
@@ -14,36 +14,23 @@
...
+constexpr const char* IMAGE_TAGS[] = {"img"};
+constexpr const char* SKIP_TAGS[] = {"head"};
 
 bool isWhitespace(const char c) { return c == ' ' || c == '\r' || c == '\n' || c == '\t'; }
 
-// given the start and end of a tag, check to see if it matches a known tag
-bool matches(const char* tag_name, const char* possible_tags[], const int possible_tag_count) {
-  for (int i = 0; i < possible_tag_count; i++) {
+template <size_t N>
+bool matches(const char* tag_name, const char* const (&possible_tags)[N]) {
```
Uri-Tauber: Nice cleanup! Worth adding a small comment near the template though - this pattern creates one function copy per array size, which can bloat flash on embedded systems. Fine here since the body is tiny (the PR add ~130 bytes flash), but good to flag so nobody blindly copies it for a heavier function.

Uri-Tauber (dismissed):

lib/Epub/Epub/parsers/ChapterHtmlSlimParser.cpp
znelson: Good point about the template bloat. Instead of some special case and a comment here, I just moved to (compile-time) `std::size` throughout.

Uri-Tauber (approved):

PR #1820, opened 2026-05-04
feat: Send optional document metadata with KOSync progress uploads

nperez0111:
## Summary

- Implements the metadata extension from [KOReader PR #15306](https://github.com/koreader/koreader/pull/15306)
- When enabled via Settings > KOReader Sync > "Send Document Metadata", progress PUT requests include a `metadata` JSON object with `filename`, `title`, and `authors` fields
- Opt-in setting (default off) matching upstream KOReader behavior; the official sync server ignores this field, but custom servers (e.g. book-stats trackers) can use it

## Changes

- **`KOReaderSyncClient.h`**: Added `KOReaderMetadata` struct and optional metadata field on `KOReaderProgress`
- **`KOReaderSyncClient.cpp`**: Serializes metadata as nested JSON object in PUT body when present
- **`KOReaderCredentialStore`**: New `sendMetadata` bool setting with getter/setter
- **`KOReaderSyncActivity.cpp`**: Populates metadata from epub path, title, and author when setting is enabled
- **`JsonSettingsIO.cpp`**: Persists `sendMetadata` to `koreader.json`
- **`KOReaderSettingsActivity.cpp`**: New "Send Document Metadata" toggle (On/Off) in device settings menu
- **`SettingsList.h`**: Exposes setting in web UI
- **`english.yaml`**: Added `STR_SEND_METADATA`, `STR_ON`, `STR_OFF` i18n strings

## JSON payload example (when enabled)

```json
{
  "document": "abc123...",
  "metadata": {
    "filename": "my_book.epub",
    "title": "The Great Novel",
    "authors": "Jane Author"
  },
  "progress": "/body/DocFragment[3]/body/p[12]/text().45",
  "percentage": 0.42,
  "device": "CrossPoint",
  "device_id": "crosspoint-reader"
}
```

## Testing

- Build compiles cleanly with 0 errors/warnings
- RAM impact: negligible (metadata strings only allocated during upload, freed immediately after)
- Backward compatible: existing `koreader.json` configs load fine (defaults to `false`)


itsthisjustin:
Help me understand the gain from this and why it wouldn't just be default to on?

nperez0111:
@itsthisjustin, the gain from this is that right now is that the KOSync synchronization is based on a privacy preserving hash of the document. This enables users to confidently use shared implementations like https://github.com/koreader/koreader-sync-server#privacy-and-security but, this comes at the cost of 3rd-party integrations.

The reason that it is off by default, is that some people prefer not to share their book details with others. Others are fine with it if they get a benefit out of it (such as being able to keep progress tracked in a book tracking application like Goodreads).

I would be personally okay with making this default on, as I would be implementing a 3rd-party book tracking platform https://bookhive.buzz to this. So if that is more palatable, I'm happy to make changes to the PR. Thanks for reaching out!

itsthisjustin (approved):

PR #1824, opened 2026-05-04
docs: expand first use of OPDS acronym and provide a wikipedia link

sizezero:
No description needed beyond title.


sizezero:
I updated the title with a "docs: " prefix. I'm not sure why the test is still failing.

znelson (approved):

PR #1832, opened 2026-05-05
refactor: Added utils for non-throwing memory allocation and scoped cleanup

znelson:
## Summary

Pared down version of #1418.

Following up on b5df6cb2b5f4415ec7ffa49b06c853a78b741f68. Added lib/Memory/Memory.h with:
- `makeUniqueNoThrow<T>` a `nothrow` wrapper for `std::make_unique` that return `nullptr` on OOM instead of calling `abort()` (the behavior of bare `new` with `-fno-exceptions`)
- `ScopedCleanup` a helper to call a cleanup lambda on scope exit.

These utilities help to write code that handles OOM scenarios gracefully, and consistently cleans up resources on scope exit.

JpegToBmpConverter.cpp has been converted to use these utilities. Other files can be converted later. 

This will simplify some of the SD card font resource management in #1327.

---

### AI Usage


Uri-Tauber:
@ngxson, could you please take a look at this PR? It seems to be in your area of expertise.

itsthisjustin (approved):

PR #1836, opened 2026-05-05
chore: Update SDK to fork in CrossPoint org

znelson:
## Summary

Update the Open X4 SDK to point to the CrossPoint organization fork. This allows us to take SDK changes without being blocked on the upstream repository.

---

### AI Usage

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
 [submodule "open-x4-sdk"]
 	path = open-x4-sdk
-	url = https://github.com/open-x4-epaper/community-sdk.git
+	url = https://github.com/crosspoint-reader/community-sdk.git


Uri-Tauber:
What difference does it make?

mcrosson:
> What difference does it make?

We don't have any kind of ability to push changes, etc to the original repo location. If we don't shift we are 'stuck' without changes to the sdk.

mcrosson (approved):

znelson:
Exactly, the only difference is for timely turnaround for device support changes. Ideally, any change we make to the crosspoint-reader org fork of the open-x4-sdk should move directly upstream to the original fork, but that process may be more delayed.

PR #1837, opened 2026-05-06
refactor: Simplify XtcReaderActivity with detectPageTurn

znelson:
## Summary

Simplify duplicated code in `XtcReaderActivity` to use `ReaderUtils::detectPageTurn`. This implementation is now shared with `EpubReaderActivity` and `TxtReaderActivity`. Also deduplicated the chapter skip time constant.

---

### AI Usage


Uri-Tauber (approved):
Haven't tested on my device, since I don't use XTC files, but the code looks great.

PR #1838, opened 2026-05-06
refactor: Simplify isReaderActivity bookkeeping

znelson:
## Summary

Before, any sub-activity of a reader activity needed to override `isReaderActivity` to maintain correct bookkeeping through `ActivityManager::isReaderActivity`. We could easily miss this in any new sub-activities. Instead, simplify so each reader activity correctly reports and then `ActivityManager` checks for any reader activity in its stack.

---

### AI Usage

### DIFF
--- a/src/activities/ActivityManager.cpp
+++ b/src/activities/ActivityManager.cpp
@@ -2,6 +2,8 @@
 
 #include <HalPowerManager.h>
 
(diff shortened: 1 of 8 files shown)


Uri-Tauber (approved):
Haven't tested on my device, but looks good.

PR #1392, opened 2026-03-14
feat: SD Card Fonts: Config, compilation, CI, and distribution (WiFi + manual)

adriancaruana:
### What is the goal of this PR?

This PR completes a series of work to add SD card font support to Crosspoint. It builds on https://github.com/crosspoint-reader/crosspoint-reader/pull/1327 (back-end firmware support for SD card fonts, under review) and https://github.com/crosspoint-reader/crosspoint-reader/pull/1056 (font rendering optimisations to support SD card fonts).

The main features of the PR include:
- A comprehensive build script `lib/EpdFont/scripts/build-sd-fonts.py` and config `lib/EpdFont/scripts/sd-fonts.yaml` for compiling SD card fonts
- A decent set of new fonts to expand language/script coverage
- An action for adding compiled fonts to the repository's assets upon release.
- UI and Activities for managing SD card fonts, selecting fonts, and downloading them from device via WiFi. (Fonts are stored in `.crosspoint/fonts/`).

Instructions if you want to test the WiFi download path: You need to add the following build flag:
```
-DFONT_MANIFEST_URL=\"https://github.com/adriancaruana/crosspoint-reader/releases/download/sd-fonts/fonts.json\"
```

Performance:
1. Advance table for CJK layout measurement — SD card fonts with >512 unique codepoints per paragraph (CJK text) no longer
  stall for 30+ seconds during section indexing

Bug fixes:
1. CJK hyphenation fix — CJK text no longer gets hyphens inserted at every line break

### Summary of font support before/after
  Before (Flash-only fonts)                                                                                                                                                     

  - 3 reading font families: Bookerly (proprietary), Noto Sans, OpenDyslexic                                                                                                    
  - 1 UI font: Ubuntu (10pt, 12pt)                                                                                                                                              
  - 4 styles per family: Regular, Bold, Italic, Bold Italic
  - Scripts: Latin only (Noto Sans had basic Latin subset baked in)
  - Sizes: 4 per family (8–18pt range)
  - ~80 global EpdFont objects compiled into flash

  After (Flash + SD card downloadable fonts)

  - 22 downloadable font families from SD card, across 6 categories:
    - 8 Serif: Literata, Source Serif 4, Noto Serif Extended, Merriweather, Lora, Gentium Book Plus, IBM Plex Serif, Bitter
    - 5 Sans-serif: Noto Sans Extended, Inter, Source Sans 3, IBM Plex Sans, Alegreya
    - 2 Monospace: IBM Plex Mono, Source Code Pro
[...]

PR #1839, opened 2026-05-06
docs: RFC for multi-language plug-in font architecture (v3)

ChiaChe726:
## Status

**Draft / Request for Comments. No code ships in this PR — only design documents.** The goal is to socialise the proposed architecture before any implementation lands so we can incorporate maintainer feedback at design time rather than after the fact.

## TL;DR

CrossPoint currently supports 23 European languages plus Chinese as a UI translation, but reading actual Chinese ebooks doesn't work — the reader fonts (NotoSerif/NotoSans/OpenDyslexic 12–18 pt) ship as Flash arrays sized for European-language coverage (~1070 glyphs each), with no CJK glyphs.

This RFC proposes **Font System v3** — a re-architecture that:

1. **Pages glyph metadata to SD** ([ADR 0001](../blob/feature/font-system-v3/docs/adr/0001-cpf-paged-metadata.md)) — cuts per-CJK-font resident cost from ~125 KB to ~1 KB. Backward compatible via a `cpf2::Header.flags` bit; existing `.cpf` files keep loading.

2. **Replaces \`EpdFontFamily\` with \`FontStack\`** ([ADR 0002](../blob/feature/font-system-v3/docs/adr/0002-font-fallback-chain.md)) — an ordered tier list with missing-glyph fallback, mirroring CSS3 \`unicode-range\` semantics. Lets users stack \`Latin + CJK + Symbols + Emoji\` fonts without per-language firmware builds.

3. **Adds a FreeRTOS prefetcher** that warms 5–10 pages of glyphs in the background, hitting the user's first-page-latency target of < 150 ms p95.

4. **Auto-discovers \`/fonts/*.cpf\`** at boot via a new \`FontRegistry\`, with NVS-persisted role-to-stack mapping and a Settings UI for reordering.

5. **Defers complex-script work** ([ADR 0003](../blob/feature/font-system-v3/docs/adr/0003-defer-harfbuzz-bidi.md)) — HarfBuzz cursive shaping, bidi, Indic conjuncts, Thai word-segmenting, runtime vector rasterisation — explicitly out of scope, with documented re-entry triggers (mostly tied to a future ESP32-S3 + PSRAM hardware refresh).

The full plan, memory budget, phased roadmap, and verification gates are in [docs/architecture/font-system-v3.md](../blob/feature/font-system-v3/docs/architecture/font-system-v3.md).

## Why this is needed

Concrete numbers from the four classical Chinese novels in our local test corpus (三国演义, 水浒传, 红楼梦, 西游记):

| Book              | Unique CJK |
| ---               | ---        |
| 三国演义           | 4121       |
| 水浒传             | 4251       |
| 红楼梦             | 4817       |
| 西游记             | 4531       |
| **Combined unique** | **5920** |
| **+ SC→TC variants** | **7830** |

At 16 B per `EpdGlyph` row, fully-resident metadata for full-coverage CJK is **~125 KB per font style**, before any bitmap data. Four styles (R/B/I/BI) won't fit in the C3's 380 KB total RAM after subtracting the 48 KB framebuffer, FreeRTOS overhead, and activity workspace (observed min-free during EPUB indexing: 22 KB).

Existing in-flight work (`feature/sd-font-loading` and successors on a forked branch) has SD-streamed bitmap data (working) but left glyph metadata fully resident. v3 makes the multi-font fallback case fit by paging the metadata too.

## Why these design choices, not LVGL Tiny TTF or KOReader's stack

Both the LVGL Tiny TTF integration and KOReader's FreeType + HarfBuzz approach are the canonical answers to "all languages from one font file" elsewhere in the embedded / e-reader space. **They both assume the entire TTF fits in RAM**: a subset NotoSansCJK is 1.9 MB, full is 9 MB. On the C3 (380 KB total RAM, no PSRAM) neither fits. Without a streaming TTF parser — which would need to be written from scratch since neither stb_truetype nor FreeType supports section-on-demand reads — that path is not viable on this chip.

The architectural exit ramp is preserved: \`IDecompressor\` (already in upstream at \`lib/EpdFont/IDecompressor.h\`) abstracts the glyph-data source. A future \`TtfDecompressor\` for a PSRAM-equipped hardware variant would plug in alongside \`FontDecompressor\` and \`SdDecompressor\` without renderer changes.

## Why open this RFC before code lands

Each subsequent PR (Phases 1 through 5 in the architecture doc) modifies a small, focused area of the codebase (~300–700 LOC each). Reviewing them individually without shared context on the overall direction is harder for both sides. This RFC sets the context once.

If maintainers see fundamental issues with the direction, catching them now is much cheaper than after Phase 4 lands.

## Asks of the reviewer

- Does the architecture make sense as the next step beyond the existing `.cpf v2` SD-streaming work?
- Is the deferral list in [ADR 0003](../blob/feature/font-system-v3/docs/adr/0003-defer-harfbuzz-bidi.md) what you expect, or is there scope you'd like added/removed?
- Are there specific phases you'd like to see split or combined?
- Any preference on ADR location / format (currently `docs/adr/`)?

I'll mark each follow-up PR as a draft against the same branch until this RFC's main points are resolved. No code merging is requested in this PR.

## Files in this PR

- `docs/architecture/font-system-v3.md` (324 lines)
- `docs/adr/0001-cpf-paged-metadata.md` (103 lines)
- `docs/adr/0002-font-fallback-chain.md` (172 lines)
- `docs/adr/0003-defer-harfbuzz-bidi.md` (92 lines)
- `README.md` — adds "Font system roadmap" section under Internals, linking the four documents

Total: ~700 lines of markdown.

## Test plan

- [x] Markdown links are valid (manually verified — relative paths against the branch tree)
- [x] All four Chinese-corpus benchmark numbers in the appendix were computed against the actual book files (`三国演义.txt`, `水浒传.txt`, `红楼梦.txt`, `西游记.txt`)
- [N/A] No code changes; `pio run` / `pio check` parity with master is by definition

### DIFF
--- a/README.md
+++ b/README.md
@@ -170,6 +170,23 @@ file will use a new cache directory, resetting the reading progress.
 
 For more details on the internal file structures, see the [file formats document](./docs/file-formats.md).
 
+### Font system roadmap
+
+The current reader fonts (NotoSerif, NotoSans, OpenDyslexic at 12–18 pt) ship as `static const` Flash arrays sized
+for European-language coverage (~1070 glyphs each). This works for the 23 currently-supported languages but does not
+scale to CJK reading (~5000+ glyphs), Hebrew/Arabic, Indic, Thai, etc.
+
+A draft architecture (Font System v3) is in development to support **plug-in fonts via SD card** with multi-script
+fallback chains, while staying within the C3's 380 KB RAM budget (no PSRAM). Design documents:
+
+* [Architecture overview](./docs/architecture/font-system-v3.md) — long-form RFC
+* [ADR 0001 — `.cpf` paged glyph metadata](./docs/adr/0001-cpf-paged-metadata.md) — how we fit ~5000 CJK glyphs in 1 KB resident
+* [ADR 0002 — Multi-font fallback chain (`FontStack`)](./docs/adr/0002-font-fallback-chain.md) — the CSS-style fallback model
+* [ADR 0003 — Defer complex-script shaping](./docs/adr/0003-defer-harfbuzz-bidi.md) — why HarfBuzz / bidi / Indic / vector
+  rasterisation are out of scope for v3 and what triggers their re-entry
+
+These documents are RFCs, not implemented yet. Comments and pushback welcome on the corresponding GitHub discussion.
+
 ## Contributing
 
 Contributions are very welcome!
--- a/docs/adr/0001-cpf-paged-metadata.md
+++ b/docs/adr/0001-cpf-paged-metadata.md
@@ -1,0 +1,103 @@
+# ADR 0001 — `.cpf` Paged Glyph Metadata
+
+| Field        | Value                                                |
+| ---          | ---                                                  |
+| Status       | Proposed                                             |
+| Date         | 2026-05-06                                           |
+| Decision by  | TBD (RFC pending upstream review)                    |
+| Supersedes   | —                                                    |
+| Superseded by| —                                                    |
+
+## Context
+
+CrossPoint Reader currently supports 23 European languages plus Chinese as a UI translation. The reader-side fonts (`NotoSerif`, `NotoSans`, `OpenDyslexic` at 12–18 pt) ship as `static const` Flash arrays of ~1070 glyphs covering Latin, Cyrillic, and common symbols. This works because the worst-case 23-language coverage stays within ~1000 glyph metadata entries (16 B each = ~17 KB resident per font), which fits comfortably in Flash.
+
+**CJK changes the regime, not just the language.** Reading common Traditional Chinese requires ~5000 glyph metadata entries; covering the four classical novels in the project's test corpus (三国演义, 水浒传, 红楼梦, 西游记) requires **7830 unique CJK codepoints** including Simplified-to-Traditional variants. At 16 B per `EpdGlyph` row this is ~125 KB resident **per font style**, before any bitmap data. Instantiate four styles (Regular/Bold/Italic/BoldItalic) and the metadata alone overshoots the ESP32-C3's 380 KB total RAM after subtracting the 48 KB framebuffer, FreeRTOS overhead, and activity workspace.
+
+The companion exploration in `lib/EpdFont/SdFontLoader.{h,cpp}` already streams *bitmap* data from SD via a 32 KB LRU group cache, but **leaves glyph metadata fully resident** (`SdFont::epdGlyphs` vector). That single decision caps practical multi-language fallback chains at "one CJK font in RAM at a time", precluding the goal of dropping any TTF on SD and getting all-language coverage.
+
+## Decision
+
+Add a backward-compatible cpf v2.1 extension that **pages glyph metadata to SD** behind a small in-memory page index, keeping resident metadata cost at ~1 KB per font regardless of glyph count.
+
+### Format change
+
+The existing `cpf2::Header.flags` field (`lib/EpdFont/CpfV2Format.h:82`) reserves **bit 1 = "paged metadata"**. When set, the file layout becomes:
+
+```
+[Header 64 B]
+[Interval Table  12 B × intervalCount]    ← stays fully resident (~600 B for CJK)
+[GlyphPageIndex  4 B × ⌈glyphCount/256⌉]   ← stays fully resident (~80 B for 5000 CJK glyphs)
+[Glyph Pages     4 KB × pageCount]         ← demand-loaded into a metadata page LRU
+[Group Table     16 B × groupCount]
+[Compressed Glyph Groups]
+```
+
+Each glyph page holds 256 `cpf2::Glyph` rows (= 4 KB = exactly one SD sector × 8). Lookup of glyph index `idx` becomes:
+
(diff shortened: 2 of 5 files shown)


mcrosson:
Work for other languages has already begun. See #1693 and #1757 for additional details.

ChiaChe726:
Hi @mcrosson, thank you for organizing this thread—I apologize for skipping directly to PR #1839 without checking here first.
                                                                                                  
 I'd like to help with the Chinese side. I have a local test corpus of four classical Chinese novels (~7800 unique codepoints) I can use to test once #1392 lands, and I can put together a shortlist of CJK fonts that convert cleanly through the cpfont tool.                                                                                           

Should I close #1839 for now and wait until CJK is on the roadmap, or keep it as a draft? Happy to follow whatever pace works for the team.

IMG_1816

mcrosson:
We will have a need for fallback and I've got some notes that say 'figure out UI font stuff'. What you have here is likely going to be a solid starting point for other work and/or highly insightful for how we can manage things as we work through fonts, languges, etc.

If you're open to working with us to get Chinese integreated into the firmware, we'd <3 the help. I just don't know when we will be ready to start that work.

Currently we are working to get the sd fonts v1 work into a public beta and once that's sorted we are hoping to work with @Uri-Tauber to get rtl and hebrew support added. Once these are addressed I had planned to loop around and start asking folk to update their language support PRs so we can start bringing them online.

For this PR in particular: can I ask how you're handling UI font integration for Chinese? That's one of my bigger open ended items at the moment.

ChiaChe726:
@mcrosson Thank you. I'd be happy to help when you're ready to work on the Chinese font. Let me first explain what we're currently doing. The following explains how we're handling the font aspect of the UI, which is different from the reader-font v3 plugin stacking mentioned in this RFC.

How the Chinese UI currently runs on the fork:

We didn't create a separate font path for the UI. Our approach is a bit rudimentary but effective:

1. We pre-rasterized a subset of the CJK font and merged it into the existing UI EpdFontFamily (the sizes of NotoSans 8/10/12 and NotoSerif UI). The subset includes characters appearing in the translated strings of chinese_tw.yaml / chinese_cn.yaml, plus common punctuation and phonetic symbols. Each UI font adds approximately 1.5–2k codepoints, which is still within budget when packed into Flash.

2. The UITheme and the caller of tr() remain unchanged. Because it's an in-place expansion of the original UI font family, each `renderer.drawText(FONT_UI_*, …, tr(STR_…))` will directly render Chinese characters, eliminating the need to introduce a fallback chain concept at the UI layer—that's cleaner to leave to the v3 reader font.

3. Build-time language filter. A flag in `platformio.ini` will completely remove unnecessary languages ​​from the build, so the Flash cost for the CJK subset is only incurred by the CJK build; the default build is unaffected.

4. Reader path normalization (SC→TC) is performed during parsing (not UI). This allows TC-only builds to render Traditional Chinese characters in Simplified Chinese EPUBs. It's char-level, table-driven, and runs only once during the entire book's parsing, not in the render loop.

5. The O(N²) time complexity for line breaks in TXT readers needs to be adjusted. The original naive line breaks would get stuck with CJK fonts—because each character in CJK is a word breakpoint. We changed it to performing a binary search on the cumulative width. Anyone who tries to incorporate CJK into the reader path will immediately run into this problem, so let me warn you.

The reason this is a stopgap measure and not the RFC's solution is that it only works if the UI string set is finite and enumerable. Once users are reading CJK books, it's impossible to pre-burn all codepoints into Flash—that's what SD-streamed .cpf v2 (paged metadata + bitmap LRU) and the FontStack fallback chain in the RFC need to address. So if your order is "Hebrew/RTL first, then look at other languages ​​later," the above UI font merging method should be sufficient to ship the Chinese UI earlier; it can then be cleanly replaced once the v3 reader font is released.

By the way, something unrelated:

I also added a dark mode display setting and refactored it to use the theme layer (so that each theme can define its own dark

variants in the future, without needing global color inversion). It's currently on a feature branch I forked. If needed, I can rebase it to the upstream master and create a separate

PR, separate from this RFC. It's fine if you're not interested; I just don't want to cram irrelevant changes into the font discussion.

When you start working, let me know which contribution method is more convenient for you—a complete PR, or just submitting the conversion script and glyph subset for your team to handle. Either is fine.

https://github.com/user-attachments/assets/eab0a6d5-55c6-4a88-af5c-d5941fdf376c

mcrosson:
I'd like to discuss your work further. Have you posted about any of this in the #1757 discussion? I think discussing this further in the language thread would be a better spot so otthers can more easily keep an eye on how we are progressing.

mcrosson:
We are slowly getting a chat/forum setup going for Crosspoint and I was wondering if you'd be willing to join us over there in the Language channel so we can more easily discuss Chinese language support and how to get it into Crosspoint long-term.

You can find the language discussion area [here](https://crosspoint.zulipchat.com/#narrow/channel/596641-55-Crosspoint-Languages/topic/Pinned.20Content/with/594687770)

Uri-Tauber:
Closing as stale.
Several CJK-related fixes and improvements have landed in CrossPoint.

PR #1844, opened 2026-05-07
refactor: Use fixed-size integers for BookMetadataCache data

znelson:
## Summary

Serialization for `BookMetadataCache` data should specify the exact integer size it expects. `size_t` varies depending on the platform. @uxjulia's crosspoint-simulator needs to patch these values to work correctly on 64-bit systems.

https://github.com/uxjulia/crosspoint-simulator/blob/9a90b6eef00cb910fae3768c575f33a50fcb75d2/run_simulator.py#L57-L68

---

### AI Usage

### DIFF
--- a/lib/Epub/Epub/BookMetadataCache.h
+++ b/lib/Epub/Epub/BookMetadataCache.h
@@ -18,11 +18,11 @@ class BookMetadataCache {
 
   struct SpineEntry {
     std::string href;
-    size_t cumulativeSize;
+    uint32_t cumulativeSize;
     int16_t tocIndex;
 
     SpineEntry() : cumulativeSize(0), tocIndex(-1) {}
-    SpineEntry(std::string href, const size_t cumulativeSize, const int16_t tocIndex)
+    SpineEntry(std::string href, const uint32_t cumulativeSize, const int16_t tocIndex)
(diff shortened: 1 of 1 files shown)


uxjulia:
This all looks good and matches the simulator issue I had to patch locally. My only concern is that this changes the `book.bin` layout, but `BOOK_CACHE_VERSION` is unchanged. I think the version should be bumped so older caches are rejected and rebuilt instead of being read with the new field widths.

znelson:
@uxjulia Good catch on thinking about cache versioning, that's the right instinct! In this case though, the on-disk layout is actually unchanged. `writePod`/`readPod` serialize based on `sizeof(T)`, and on ESP32-C3 (our only firmware target), `sizeof(size_t) == sizeof(uint32_t) == 4`. So existing `book.bin` files on SD cards are already byte-for-byte compatible with the new types.

This change makes the serialization contract explicit rather than relying on `size_t` happening to be the right width, so we don't have to patch 64-bit targets. 🙂

uxjulia:
Well then! Ignore that thought haha I've just had so many issues with caching I always assume the worst.

uxjulia (approved):

Uri-Tauber (approved):
LGTM

PR #1847, opened 2026-05-07
refactor: Deduplicate Roundraff battery drawing

znelson:
## Summary

Deduplicated battery drawing code for Roundraff theme with other themes. 

Now `BaseTheme` provides non-virtual `drawBatteryLeft` and `drawBatteryRight` methods, which use a shared static `drawBatteryOutline` implementation, and defer to a virtual `fillBatteryIcon` implementation. In Classic and Roundraff `fillBatteryIcon` uses a solid fill, while Lyra and Lyra Extended use a segmented fill. 

The charging indicator inside the battery was missing for Roundraff, but is now consistent with other themes because of the shared drawing implementation.

---

### AI Usage

### DIFF
--- a/src/components/themes/BaseTheme.cpp
+++ b/src/components/themes/BaseTheme.cpp
@@ -20,37 +20,6 @@ constexpr int homeMenuMargin = 20;
 constexpr int homeMarginTop = 30;
 constexpr int subtitleY = 738;
 
-// Helper: draw battery icon at given position
-void drawBatteryIcon(const GfxRenderer& renderer, int x, int y, int battWidth, int rectHeight, uint16_t percentage) {
-  // Draw battery outline (shared code)
-  BaseTheme::drawBatteryOutline(renderer, x, y, battWidth, rectHeight);
-
-  const bool charging = gpio.isUsbConnected();
-
-  // The +1 is to round up, so that we always fill at least one pixel
-  const int maxFillWidth = battWidth - 5;
-  const int fillHeight = rectHeight - 4;
-  if (maxFillWidth <= 0 || fillHeight <= 0) {
-    return;
-  }
-  int filledWidth = percentage * maxFillWidth / 100 + 1;
-  if (filledWidth > maxFillWidth) {
-    filledWidth = maxFillWidth;
-  }
-
-  // When charging, ensure minimum fill so lightning bolt is fully visible
-  constexpr int minFillForBolt = 8;
-  if (charging && filledWidth < minFillForBolt) {
-    filledWidth = std::min(minFillForBolt, maxFillWidth);
-  }
-
-  renderer.fillRect(x + 2, y + 2, filledWidth, fillHeight);
-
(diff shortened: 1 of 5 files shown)


src/components/themes/BaseTheme.cpp
```diff
@@ -79,18 +48,46 @@ void BaseTheme::drawBatteryLightningBolt(const GfxRenderer& renderer, int boltX,
...
   // Left aligned: icon on left, percentage on right (reader mode)
   const uint16_t percentage = powerManager.getBatteryPercentage();
   const int y = rect.y + 6;
 
   if (showPercentage) {
     const auto percentageText = std::to_string(percentage) + "%";
-    renderer.drawText(SMALL_FONT_ID, rect.x + BaseTheme::batteryPercentSpacing + BaseMetrics::values.batteryWidth,
-                      rect.y, percentageText.c_str());
+    renderer.drawText(SMALL_FONT_ID, rect.x + batteryPercentSpacing + rect.width, rect.y, percentageText.c_str());
   }
```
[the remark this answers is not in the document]
znelson: All callers already clear the battery area before invoking these methods — drawBatteryLeft is only called from drawStatusBar, which runs after a full clearScreen(), and every drawHeader implementation clears the battery region itself before calling drawBatteryRight. The per-text fillRect was redundant.

pablohc (dismissed):
LGTM

src/components/themes/BaseTheme.h
```diff
@@ -153,15 +154,14 @@ class BaseTheme {
...
-  virtual void drawStatusBar(GfxRenderer& renderer, const float bookProgress, const int currentPage,
-                             const int pageCount, std::string title, const int paddingBottom = 0,
-                             const int textYOffset = 0) const;
-  virtual void drawHelpText(const GfxRenderer& renderer, Rect rect, const char* label) const;
-  virtual void drawTextField(const GfxRenderer& renderer, Rect rect, const int textWidth, bool cursorMode = false,
-                             int contentStartX = 0, int contentWidth = 0) const;
-  virtual void drawKeyboardKey(const GfxRenderer& renderer, Rect rect, const char* label, const bool isSelected,
-                               const char* secondaryLabel = nullptr, KeyboardKeyType keyType = KeyboardKeyType::Normal,
-                               bool inactiveSelection = false) const;
+  void drawStatusBar(GfxRenderer& renderer, const float bookProgress, const int currentPage, const int pageCount,
```
Uri-Tauber, line 157: Overall, good work here. I can also confirm I tested on my device and there isn't any regresion. 
One question: Did you deliberately removed the `virtual` on those functions? It will collide with my work in #1851.

znelson: I did, just to eliminate virtual method dispatch in places where we don't ever override the method. It's fine to add them back for the cases in #1851 where you really do need to override the behavior.

Uri-Tauber (approved):

PR #1849, opened 2026-05-07
feat: Status bar for XTC files 

### LINKED ISSUE #1848: [Feature Request] Status bar for XTC files
Add a status bar for XTC files similar to that for epubs - using the XTC headers for chapters if available to populate the relevant fields.

People are baking in status bars using the epub-to-xtc converters but that doesn't allow for displaying dynamic information like remaining battery.

leecming82:
## Summary
Add ability to show a status bar for 1-bit XTC files  (closes #1848) 

Overlays a status bar (similar style to that for epubs) over the image. Defaults to hidden, users can set to show and either top or bottom of the screen within "Customize Status Bar" reader settings. 

I've leaned toward a simple overlay approach rather than trying anything clever e.g. resizing the original image or allowing for shifting the image around. I think it's more straight forward for the user to create a buffer zone when generating the XTC files. 

For 2-bit image files, it'd require more work to handle the multiple render passes so I'm holding off on that. 

Sample from a Japanese text XTC
image

### AI Usage

YES - Codex


src/activities/reader/XtcReaderActivity.cpp
```diff
@@ -131,6 +133,101 @@ void XtcReaderActivity::render(RenderLock&&) {
...
+  if (chapterIt == chapters.end() || chapterIt->endPage < chapterIt->startPage) {
+    return StatusBarPageInfo{bookPage, bookPageCount};
+  }
+
+  return StatusBarPageInfo{static_cast<int>(currentPage - chapterIt->startPage) + 1,
+                           static_cast<int>(chapterIt->endPage - chapterIt->startPage) + 1};
+}
+
+void XtcReaderActivity::renderStatusBarOverlay(const StatusBarOverlayPosition position) const {
+  if (!xtc) {
```
znelson: Probably an unnecessary guard.

src/activities/reader/XtcReaderActivity.cpp
```diff
@@ -131,6 +133,101 @@ void XtcReaderActivity::render(RenderLock&&) {
...
+  if (SETTINGS.statusBarTitle == CrossPointSettings::STATUS_BAR_TITLE::BOOK_TITLE) {
+    return xtc->getTitle();
+  }
+
+  if (SETTINGS.statusBarTitle != CrossPointSettings::STATUS_BAR_TITLE::CHAPTER_TITLE || !xtc->hasChapters()) {
+    return "";
+  }
+
+  const auto& chapters = xtc->getChapters();
+  const auto chapterIt = std::find_if(chapters.begin(), chapters.end(), [this](const xtc::ChapterInfo& chapter) {
```
znelson: Perhaps `StatusBarPageInfo` should just include the chapter title as a member, so you don't need to run two `std::find_if` linear searches over the chapters (one in `getStatusBarTitle`, another in `getStatusBarPageInfo`) to draw the status bar?

src/activities/reader/XtcReaderActivity.cpp
```diff
@@ -131,6 +133,101 @@ void XtcReaderActivity::render(RenderLock&&) {
...
+  if (statusBarHeight <= 0) {
+    return;
+  }
+
+  int orientedMarginTop, orientedMarginRight, orientedMarginBottom, orientedMarginLeft;
+  renderer.getOrientedViewableTRBL(&orientedMarginTop, &orientedMarginRight, &orientedMarginBottom,
+                                   &orientedMarginLeft);
+  (void)orientedMarginTop;
+  (void)orientedMarginRight;
+  (void)orientedMarginLeft;
```
znelson: Do you get warnings without these? `BaseTheme::fillPopupProgress` does the same `renderer.getOrientedViewableTRBL` call and never references `orientedMarginTop` after, so maybe you can drop these?

znelson (dismissed):

src/activities/reader/XtcReaderActivity.cpp
leecming82: Removed

src/activities/reader/XtcReaderActivity.cpp
leecming82: Deduped the two functions into a single one

src/activities/reader/XtcReaderActivity.cpp
leecming82: Removed

leecming82:
@znelson the three comments made sense, fixed. Let me know

src/activities/reader/XtcReaderActivity.cpp
```diff
@@ -131,6 +133,77 @@ void XtcReaderActivity::render(RenderLock&&) {
...
+    return StatusBarInfo{bookPage, bookPageCount, bookTitle};
+  }
+
+  std::string title = bookTitle;
+  if (SETTINGS.statusBarTitle == CrossPointSettings::STATUS_BAR_TITLE::CHAPTER_TITLE) {
+    title = chapterIt->name.empty() ? tr(STR_UNNAMED) : chapterIt->name;
+  }
+
+  return StatusBarInfo{static_cast<int>(currentPage - chapterIt->startPage) + 1,
+                       static_cast<int>(chapterIt->endPage - chapterIt->startPage) + 1, title};
```
znelson: Small nit: this makes a copy of `bookTitle` to `title`, then makes a copy into `StatusBarInfo`. If you make `bookTitle` non-const, you can just overwrite it with chapter title in the chapter block, then move into the `StatusBarInfo`, and save a couple of potential copies.


```suggestion
  std::string title =
      xtc && SETTINGS.statusBarTitle == CrossPointSettings::STATUS_BAR_TITLE::BOOK_TITLE ? xtc->getTitle() : "";

  if (!xtc || !xtc->hasChapters()) {
    return StatusBarInfo{bookPage, bookPageCount, std::move(title)};
  }

  const auto& chapters = xtc->getChapters();
  const auto chapterIt = std::find_if(chapters.begin(), chapters.end(), [this](const xtc::ChapterInfo& chapter) {
    return currentPage >= chapter.startPage && currentPage <= chapter.endPage;
  });

  if (chapterIt == chapters.end() || chapterIt->endPage < chapterIt->startPage) {
    return StatusBarInfo{bookPage, bookPageCount, std::move(title)};
  }

  if (SETTINGS.statusBarTitle == CrossPointSettings::STATUS_BAR_TITLE::CHAPTER_TITLE) {
    title = chapterIt->name.empty() ? tr(STR_UNNAMED) : chapterIt->name;
  }

  return StatusBarInfo{static_cast<int>(currentPage - chapterIt->startPage) + 1,
                       static_cast<int>(chapterIt->endPage - chapterIt->startPage) + 1, std::move(title)};
```

znelson (dismissed):

znelson:
There's also some unnecessary checks of `xtc` in `XtcReaderActivity::getStatusBarInfo` you could consider removing. As far as I can tell, we never render status bar unless `xtc` is valid, so repeated checking just leads the next dev to wonder how careful they need to be and write overly defensive code. We can always follow up on these small nits in another PR though, too. No stress.

leecming82:
@znelson thx, your suggested changes made sense, have updated to reflect them

znelson (approved):

znelson:
Thanks @leecming82!

zgredex:
@leecming82 @znelson nice one. My #1614 will bring this into 2 bit. Got it working together with conflict resolution this one introduced.

PR #1850, opened 2026-05-07
Create SECURITY.md

onsali:
## Summary

Add a `SECURITY.md` policy so security issues can be reported responsibly and privately.
Added `SECURITY.md` at the repository root with:
    - supported version guidance
    - vulnerability reporting instructions via GitHub Security Advisories
## Additional Context

  specific areas to focus on).

This PR is intended to establish a secure vulnerability disclosure channel for CrossPoint before sensitive details are shared publicly. I have identified some security issues and am not disclosing details via PR to avoid increasing risk before a coordinated process is in place.

---

### AI Usage

No

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,32 @@
+# Security Policy
+## Supported Versions
+
+Security fixes are provided for the latest released version of CrossPoint Reader.
+
+| Version | Supported          |
+| ------- | ------------------ |
+| latest  | :white_check_mark: |
+| older   | :x:                |
+
+## Reporting a Vulnerability
+
+Please **do not report security vulnerabilities in public GitHub issues, discussions, or pull requests**.
+
+Instead, report vulnerabilities privately using **GitHub Private Vulnerability Reporting / Security Advisories** 
+- Go to the repository’s **Security** tab and choose **Report a vulnerability**.
+
+Please include:
+
+- A clear description of the issue
+- Steps to reproduce
+- Affected version(s) / commit(s)
+- Any suggested remediation (if known)
+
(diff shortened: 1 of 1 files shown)


Uri-Tauber:
I find it hard to believe we would have security issues.
Can you give concrete example this would be useful?

onsali:
Good question. This PR is preventive process hygiene, not a proof of concept for exploitable issues. Those come in after a vulnerability disclosure process is in place. For instance, I cannot share how one would be able to remotely takeover, and corrupt an XTE-ink device running Crosspoint in a public issue or pull request. 

Without SECURITY.md or a private reporting path, researchers either disclose publicly or have no clear channel. I myself run this firmware on my device, and would like to contribute to its security.

This policy creates a safe route for coordinated disclosure so maintainers can patch before details are public.

Uri-Tauber:
I honestly don’t think this sort of infrastructure is relevant to Crosspoint.

At the end of the day, it’s just an e-reader, not a smartphone or laptop. People generally don’t store sensitive information on their e-readers.

Not to mention, we don’t encrypt the SD card, and Wi-Fi passwords are stored in plain text anyway. So if someone really wanted to steal your credit card details, they could just take the actual device.

If you’ve done work that improves security without harming the UX at all (even a 10ms delay probably isn’t worth it), feel free to open a public PR.

mcrosson:
I agree with @Uri-Tauber ; unless we suddenly start encrypting things, this is a moot point and not in scope.

If youd like to work on lock screen, encrypted storage, etc; please do. At that time we can revisit a document like this.

PR #1851, opened 2026-05-07
fix: two roundedraff bugs

### LINKED ISSUE #1843: Keyboard issue with RoundedRaff
### Affected Version

1.2.0

### Describe the Bug

So, I had a weird issue, I connected to a new wifi, but when it came to input the password it froze. When I try reverting to the normal Lyra, no issue, the keyboard popped up, but when using the RoundedRaff theme, stuck. After connecting, from Lyra, reverted to RoundedRaff, and since the password was already saved from just before, no issue. Tried with other options that needed keyboard (new opds, ...) same, froze, but no issue with other theme

https://github.com/user-attachments/assets/de054ca5-f336-4f02-8a8b-e930638c3bad

### Steps to Reproduce

1. select RoundedRaff as theme
2. pick any field that'd need something to be input (opds, sync, wifi...)
3. expect the keyboard to popup but.... just freezes.

### Expected Behavior

After selecting a field that'd need an input, the keyboard would popup and let enter the data

```shell

```

Uri-Tauber:
## Summary

* **What is the goal of this PR?** fix two bugs in roundedraff theme.

fix #1843

The two new functions, drawTextField and drawKeyboardKey, were added just to keep the UI styling consistent with the rest of the theme — the fix itself works independently of them.

DSCF3965

---

### AI Usage

### DIFF
--- a/src/components/themes/roundedraff/RoundedRaffTheme.cpp
+++ b/src/components/themes/roundedraff/RoundedRaffTheme.cpp
@@ -282,6 +282,77 @@ void RoundedRaffTheme::drawButtonMenu(GfxRenderer& renderer, Rect rect, int butt
   drawScrollBar(renderer, rect, buttonCount, pageStartIndex, pageItems);
 }
 
+void RoundedRaffTheme::drawTextField(const GfxRenderer& renderer, Rect rect, const int textWidth, bool cursorMode,
+                                     int contentStartX, int contentWidth) const {
+  const auto& metrics = UITheme::getInstance().getMetrics();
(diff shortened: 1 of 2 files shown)


znelson (approved):
I agree with the coderabbit concern about hardcoding `keyRadius` instead of using `keyboardKeyCornerRadius`. This looks fine for now and it sounds like you have a follow-up PR to refine this further.

znelson:
@Uri-Tauber should we go ahead with this?

Uri-Tauber:
I think so.

PR #1857, opened 2026-05-07
feat: Add button lock feature with short power button press

marcinoktawian:
## Summary

* Implements a button lock feature that suspends input handling without powering off, allowing users to safely transport the device and resume reading immediately.
image
image

* Added `buttonsLocked` and `buttonsLockedDisplayed` state flags to `CrossPointState` for persistent lock status
* Extended `SHORT_PWRBTN` enum in `CrossPointSettings` with new `LOCK_BUTTONS` option
* Visual indicator: popup displays "Buttons locked" when feature is active
* All buttons except power button are blocked when locked
* To unlock buttons user needs to use key combination [POWER + BACK]

## Additional Context

* Lock popup is drawn only once when state changes (not every frame) to avoid display refresh loops that were observed in testing
* Power button check happens before the lock block, ensuring unlock is always possible
* When locked, `activityManager.loop()` is skipped entirely, preventing activity processing while displaying the lock indicator

**Potential future enhancements:**
* Support for double-click on power button instead of short-press (discussed but deferred)

---

### AI Usage


pablohc:
I love it! Thank you so much for this, several of us had been asking for it in this discussion.: https://github.com/crosspoint-reader/crosspoint-reader/discussions/1471

Just a suggestion that would add a little more security. 
Since the button that unlocks the device is the Power button itself, which is usually exposed, would it be possible to use a key combination to exit this mode, or perhaps a specific sequence?

In other words, lock buttons with a short press of the Power button and exit only by:

- Hold [ BACK ] + Press [ POWER ] (Currently, the device restarts; its behavior needs to be modified.)

- Press [ UP ] > Press [ LEFT ] > Press [ POWER ]

hisnameisjimmy:
I would be a perfect guinea pig for this issue, since it seems to occur quite frequently to me. I am very excited for this, thank you for working on it!

marcinoktawian:
During updating the feature I've found the bug with long press power off 
https://github.com/crosspoint-reader/crosspoint-reader/issues/1881
I'm holding this feature till fix this issue

pablohc:
> During updating the feature I've found the bug with long press power off #1881 I'm holding this feature till fix this issue

Hello @marcinoktawian Sorry, I thought that combination might cause problems, which is why I suggested an alternative.
Maybe using [ DOWN ] + [ BACK ] ?
I'll try running some tests tonight and let you know if there are any updates.
Thanks anyway for taking my suggestion! ;)

marcinoktawian:
No worries @pablohc,
I rather prefer to think that this combination found the problem rather than caused it :)

I will work with a different combination until the issue is fixed, but I agree that due to the localization of the buttons, a combination using one front button and one side button will probably be the best approach.

The [DOWN] button is located both on the side and on the front, so it can be accidentally triggered when pressing all front buttons together. Because of that, I think [POWER]+ [BACK] or [POWER] + [SELECT] would be the best combinations.

marcinoktawian:
Buttons are unlock with combination [POWER] + [BACK]
I've added that information to popup, but in future popup should accept multiline message
image

pablohc:
Hi @marcinoktawian, thank you very much for the update. I'll take a look at it tonight and give you my feedback.

insunaa:
I like the idea, but can we have an option for the key lock to also be the Back+Power combo? I much prefer using the power button as my page turner, but would still like to take advantage of this feature

pablohc:
Hi @marcinoktawian, I've just tested it, and although the combination of the back and power buttons no longer turn off, I think the power button issue is still present.
If, for some reason, you hold down any button other than BACK and then give the POWER button a short press, the device turns off. When it turns back on, the buttons would be unlocked again, which defeats the purpose.

marcinoktawian:
> Hi @marcinoktawian, I've just tested it, and although the combination of the back and power buttons no longer turn off, I think the power button issue is still present. If, for some reason, you hold down any button other than BACK and then give the POWER button a short press, the device turns off. When it turns back on, the buttons would be unlocked again, which defeats the purpose.

For this problem, I’ve created a separate PR because the issue already existed and was not introduced by this feature — it was discovered during its implementation
My PRs to fix that:
https://github.com/crosspoint-reader/community-sdk/pull/3
https://github.com/crosspoint-reader/crosspoint-reader/pull/1890

marcinoktawian:
> I like the idea, but can we have an option for the key lock to also be the Back+Power combo? I much prefer using the power button as my page turner, but would still like to take advantage of this feature

Sure, I will consider adding it as a separate feature toggle.

pablohc:
> > I like the idea, but can we have an option for the key lock to also be the Back+Power combo? I much prefer using the power button as my page turner, but would still like to take advantage of this feature
> 
> Sure, I will consider adding it as a separate feature toggle.

Hi @marcinoktawian, I have something in mind. If you're interested in the idea, I'll reach out to you later to see if it meets your needs.

Joseph-DiGiovanni:
I think something more similar to #1933 should be considered. Why keep the device on for safe transport?

marcinoktawian:
> I think something more similar to #1933 should be considered. Why keep the device on for safe transport?

Hi @Joseph-DiGiovanni, this feature is meant for quickly getting back to your book without spending time turning the device on again. It’s more like a temporary lock while actively reading for a few seconds or minutes, not a transport/security lock to prevent the device from waking up in a bag or pocket for hours.

For example, I like reading on public transport. I start reading at the bus stop, press the block buttons, get on the bus, and continue reading immediately. Then I block the device again, walk 100 meters to catch another bus or tram, and instantly resume reading without waiting for the device to turn on.

marcinoktawian:
> > > I like the idea, but can we have an option for the key lock to also be the Back+Power combo? I much prefer using the power button as my page turner, but would still like to take advantage of this feature
> > 
> > 
> > Sure, I will consider adding it as a separate feature toggle.
> 
> Hi @marcinoktawian, I have something in mind. If you're interested in the idea, I'll reach out to you later to see if it meets your needs.

Hi @pablohc, @insunaa 
I implemented the requested behavior where the device is locked using Power + Back and unlocked using Power + Back as well, instead of relying on the current short-press power button action.

However, there is a problem I currently do not know how to solve cleanly. Before the lock is triggered, the normal actions assigned to those buttons are still executed. For example, pressing Back may navigate to the previous view and refresh the screen before the buttons become locked.

At the moment, I do not have a good idea how to bypass this behavior.

If anyone has ideas or wants to experiment with it, I pushed the feature toggle implementation to this branch:
[crosspoint-reader test-branch](https://github.com/marcinoktawian/crosspoint-reader/tree/test-branch?utm_source=chatgpt.com)

pablohc:
> > > > I like the idea, but can we have an option for the key lock to also be the Back+Power combo? I much prefer using the power button as my page turner, but would still like to take advantage of this feature
> > > 
> > > 
> > > Sure, I will consider adding it as a separate feature toggle.
> > 
> > 
> > Hi @marcinoktawian, I have something in mind. If you're interested in the idea, I'll reach out to you later to see if it meets your needs.
> 
> Hi @pablohc, @insunaa I implemented the requested behavior where the device is locked using Power + Back and unlocked using Power + Back as well, instead of relying on the current short-press power button action.
> 
> However, there is a problem I currently do not know how to solve cleanly. Before the lock is triggered, the normal actions assigned to those buttons are still executed. For example, pressing Back may navigate to the previous view and refresh the screen before the buttons become locked.
> 
> At the moment, I do not have a good idea how to bypass this behavior.
> 
> If anyone has ideas or wants to experiment with it, I pushed the feature toggle implementation to this branch: [crosspoint-reader test-branch](https://github.com/marcinoktawian/crosspoint-reader/tree/test-branch?utm_source=chatgpt.com)

Thank you so much for your work, @marcinoktawian. Let me test it a little longer and we'll see if we can find a solution.

Joseph-DiGiovanni:
> It’s more like a temporary lock while actively reading for a few seconds or minutes, not a transport/security lock to prevent the device from waking up in a bag or pocket for hours.

My suggested solution solves both issues as well as basic tamper resistance from kids, etc. Wake time including loading the previous page is less than 4 seconds.

Frankly I think button combos to lock/unlock inputs for a brief pause is overengineering the solution when sleep/wake isn't much slower.

itsthisjustin:
Closing this as not in scope but great idea in conjuction with the lock screen request. Can revist later

PR #1862, opened 2026-05-07
feat: add numeric chapter prefix to reader menus

ffrafat:
## Summary

The goal of this PR is to improve the usability and accessibility of the chapter selection menu, particularly for users reading content in languages or fonts that are not fully supported by the current system fonts.

- Added a numeric prefix (`X.`) to all TOC entries in the **EPUB** chapter selection activity.
- Added the same numeric prefix to the **XTC** (pre-rendered) chapter selection activity.
- Implemented consistent text truncation to ensure the prefix remains visible even for very long chapter titles.

## Additional Context

Currently, if a chapter title contains characters that the current font cannot render (e.g., specific UTF-8 symbols or unsupported scripts), the entry appears completely blank in the menu. This makes navigation impossible for affected books.

By adding a numeric prefix in a standard supported font, users can always identify the chapter index even if the rest of the title is unreadable. This is a low-risk, high-impact UX improvement that makes the reader more robust for international users.

---

### AI Usage

### DIFF
--- a/src/activities/reader/EpubReaderChapterSelectionActivity.cpp
+++ b/src/activities/reader/EpubReaderChapterSelectionActivity.cpp
@@ -124,8 +124,14 @@ void EpubReaderChapterSelectionActivity::render(RenderLock&&) {
 
     // Indent per TOC level while keeping content within the gutter-safe region.
     const int indentSize = contentX + 20 + (item.level - 1) * 15;
+
+    char prefix[32];
+    snprintf(prefix, sizeof(prefix), "%d. ", itemIndex + 1);
+    std::string fullTitle = prefix;
+    fullTitle += item.title;
+
     const std::string chapterName =
-        renderer.truncatedText(UI_10_FONT_ID, item.title.c_str(), contentWidth - 40 - indentSize);
+        renderer.truncatedText(UI_10_FONT_ID, fullTitle.c_str(), contentWidth - 40 - (indentSize - contentX));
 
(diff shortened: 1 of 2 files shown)


Uri-Tauber:
Given that we’ve almost finished custom font support, I don’t think this will be necessary.

Also, I agree with CodeRabbit — having “Chapter X:” could get quite annoying when reading in languages other than English. Maybe just using “X.” would be better?

ffrafat:
> Given that we’ve almost finished custom font support, I don’t think this will be necessary.
> 
> Also, I agree with CodeRabbit — having “Chapter X:” could get quite annoying when reading in languages other than English. Maybe just using “X.” would be better?

@Uri-Tauber Thanks for your insights. Well, I am a Bengali speaker. The issue is that even if I convert my Bengali fonts to bin they are still broken and don't work properly. Because unlike English language and various Roman scripts, Bengali and other indic languages have diacritics. Those are not properly in place in any of the firmwares supporting custom fonts. It needs specialized mapping probably. For example this is the actual Bengali Text: সেই যথার্থ মানুষ যে জীবনের পরিবর্তন দেখেছে এবং পরিবর্তনের সাথে নিজেও পরিবর্তিত হয়েছে। and here how converted font renders. Anyway, I think I will remove the string Chapters and will just keep the number. Thus even the users who have proper custom font won't be bored and they can see the chapter index.
image

itsthisjustin:
Closing this for now as we should consolidate this into proper fixes inside of a new font system

PR #1882, opened 2026-05-08
feat: X3 clock display 

leecming82:
## Summary
- Adds a clock option for the reader status bar
- Continues/supersedes #1612 (OP hasn't responded in a month) with minor additional fixes to account for the XTC status bar, and to show the option only for clock capable devices (X3 currently)

---

### AI Usage
Did you use AI tools to help write this code? YES - Codex


leecming82:
@mcrosson mind reviewing this?

And to your last question in the original PR about using NTP only (wihout DS3231), you'd need to constantly poll wifi between sleeps which a) the network may not be available and b) is energy expensive

PR #1875, opened 2026-05-08
refactor: unify book cache clearing for epub, txt, and xtc files

WuTofu:
## Summary

* What is the goal of this PR?  
  Implements unified book cache invalidation method for text, epub, and xtc files content paths to ensure cache is cleared whenever files are uploaded, renamed, moved, or deleted.

* What changes are included?  
  - Added `Txt::clearCache()` support in `Txt.cpp` / `Txt.h`
  - Added `clearBookCache()` helper in `BookCacheUtils.cpp` / `BookCacheUtils.h`
  - Replaced EPUB-specific cache clearing calls with the generic helper in:
    - OpdsBookBrowserActivity.cpp
    - CrossPointWebServer.cpp
    - WebDAVHandler.cpp
  - Extended settings cache cleanup in ClearCacheActivity.cpp to handle `txt_` cache directories as well

## Additional Context

I will work on `FileBrowserActivity.cpp` as part of #1803's follow-up after this pr.

---

### AI Usage

While CrossPoint doesn't have restrictions on AI tools in contributing, please be transparent about their usage as it 
helps set the right context for reviewers.

Did you use AI tools to help write this code? _**PARTIALLY**_

PR #1892, opened 2026-05-09
fix: clear cache when deleting folders in FileBrowserActivity

WuTofu:
## Summary

  Fix folder deletion in `FileBrowserActivity` so cache metadata is cleared for all files inside a deleted directory, not just for individual EPUB files.

  - Added `fileNameBuffer` allocation to safely read directory entries via `FsFile::getName`.
  - Replaced old per-file metadata clearing with `removeDirFile()` to recursively delete directories and clear book cache for files encountered during deletion and avoided double traversal of directories by combining metadata cleanup and removal in one pass.

## Additional Context

This PR uses a helper function that is introduced in #1875, so we won't only clear epubs' cache, but also the cache of txt, xtc files.

---

### AI Usage

### DIFF
--- a/lib/Txt/Txt.cpp
+++ b/lib/Txt/Txt.cpp
@@ -155,6 +155,21 @@ bool Txt::generateCoverBmp() const {
   return false;
 }
 
+bool Txt::clearCache() const {
+  if (!Storage.exists(cachePath.c_str())) {
+    LOG_DBG("TXT", "Cache does not exist, no action needed");
+    return true;
+  }
+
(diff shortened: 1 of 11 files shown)


itsthisjustin:
@WuTofu does https://github.com/crosspoint-reader/crosspoint-reader/pull/1875 have to be merged first for this to be functional?

WuTofu:
> @WuTofu does #1875 have to be merged first for this to be functional?

That's right 😊

itsthisjustin (approved):

PR #1910, opened 2026-05-10
fix: build-script bug fixes for fontconvert{,_sdcard}.py

jpirnay:
Fontconvert_sdcard.py:

* Move face.set_char_size() before any load_glyph() call. The validation loop runs load_glyph() with FT_LOAD_RENDER, which renders at the *active* size — calling it before set_char_size() wastes work at the default size and triggers Invalid_Size_Handle on some fonts.

* Walk bitmap.buffer using bitmap.pitch with negative-pitch handling. The previous linear iteration assumed pitch == width and a top-down layout, which silently corrupts output for any padded or bottom-up bitmap FreeType returns.

* Fix operator-precedence bug in 2-bit tail-padding: px << (4 - … % 4) * 2 evaluated as (px << (4 - … % 4)) * 2 due to << binding tighter than *. Add the missing outer parens.

* Drop SMP codepoints (> U+FFFF) from kern and ligature tables before packing — the binary format uses uint16/uint32 codepoint fields and would raise struct.error otherwise.

* Skip GPOS kern subtables that aren't lookup type 2 (PairPos). _extract_pairpos_subtable assumes Type-2 layout; cursive attachment and other types reachable through the kern feature crash inside it.

Fontconvert.py:

* Force UTF-8 stdout so `python fontconvert.py … > foo.h` on Windows doesn't emit UTF-16 / replacement characters in the generated header.

## Summary

## Additional Context

  specific areas to focus on).

---

### AI Usage

### DIFF
--- a/lib/EpdFont/scripts/fontconvert.py
+++ b/lib/EpdFont/scripts/fontconvert.py
@@ -8,6 +8,12 @@ import argparse
 from collections import namedtuple
 from fontTools.ttLib import TTFont
 
+# Force UTF-8 stdout so that `python fontconvert.py … > foo.h` on Windows
+# (default cp1252) doesn't emit UTF-16 LE / replacement chars in the generated
+# header. Wrapped in a hasattr guard so it's a no-op on older Pythons.
+if hasattr(sys.stdout, 'reconfigure'):
+    sys.stdout.reconfigure(encoding='utf-8')
+
 # Originally from https://github.com/vroland/epdiy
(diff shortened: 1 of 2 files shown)


mcrosson:
@znelson can you review this PR

znelson (approved):

znelson:
This change blew up the font rendering time by several orders of magnitude. Working on figuring out a fix...

PR #1911, opened 2026-05-10
fix: capture instantiateVariableFont return value

jpirnay:
instantiateVariableFont() returns a *new* TTFont with the requested axes pinned; the existing call discards that return and proceeds to font.save() the original (still-variable) font. The "static instance" written to disk is therefore identical to the source variable font and the requested axis values are silently ignored.

For sd-fonts.yaml entries that use the `variable: {wght: 400}` form (e.g. Bitter, Lora) this means every weight ends up rasterised from the same default-weight glyphs — bold and regular look the same.

Capture the return value, and pass two diagnostically useful kwargs while we're touching the call:

* updateFontNames=True  — rewrite the name table so the saved TTF reports its actual weight/style instead of retaining the source variable-font names.
* optimize=False        — skip the gvar interpolation optimisation;
                          fully pinning every axis drops gvar anyway,
                          so the work would be wasted.

The atomic-write scaffolding around the call is left untouched.

## Summary

## Additional Context

  specific areas to focus on).

---

### AI Usage

### DIFF
--- a/lib/EpdFont/scripts/build-sd-fonts.py
+++ b/lib/EpdFont/scripts/build-sd-fonts.py
@@ -87,7 +87,18 @@ def extract_static_instance(source_path: Path, axes: dict, family_name: str, sty
     tmp_path = Path(tmp_name)
     font = TTFont(str(source_path))
     try:
-        instantiateVariableFont(font, axes)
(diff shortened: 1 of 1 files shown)


mcrosson:

znelson (approved):

PR #1912, opened 2026-05-10
feat: closest-pt size selection instead of ordinal slot

jpirnay:
Add SdCardFontFamilyInfo::pickClosestSize(targetPtSize) and have the manager and SdCardFontSystem drive size selection from a target point size derived from the user's font-size enum (SMALL=12, MEDIUM=14, LARGE=16, EXTRA_LARGE=18) rather than indexing the family's sorted size list by enum ordinal.

The ordinal-slot approach mis-selects whenever a family doesn't ship the canonical {12,14,16,18} set: a family with only [10,14,18] would map SMALL/MEDIUM/LARGE/EXTRA_LARGE to 10/14/18/18 — fine for SMALL but arbitrary for the rest. Closest-pt always picks the on-disk file nearest to the user-intended point size, with a deterministic smaller-pt tie-break.

No change for canonical-sized families.

## Summary

## Additional Context

  specific areas to focus on).

---

### AI Usage

### DIFF
--- a/src/SdCardFontSystem.cpp
+++ b/src/SdCardFontSystem.cpp
@@ -5,10 +5,15 @@
 
 #include "CrossPointSettings.h"
 
-static uint8_t fontSizeEnumFromSettings() {
+// Map fontSize enum (SMALL=0, MEDIUM=1, LARGE=2, EXTRA_LARGE=3) to the point
+// sizes shipped with the built-in fonts. Used to drive closest-pt selection
+// in the SD card font registry (see SdCardFontFamilyInfo::pickClosestSize).
+static constexpr uint8_t FONT_SIZE_TO_PT[CrossPointSettings::FONT_SIZE_COUNT] = {12, 14, 16, 18};
+
+static uint8_t targetPtSizeFromSettings() {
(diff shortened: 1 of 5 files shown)


mcrosson:

src/SdCardFontSystem.cpp
```diff
@@ -5,10 +5,15 @@
 
 #include "CrossPointSettings.h"
 
-static uint8_t fontSizeEnumFromSettings() {
+// Map fontSize enum (SMALL=0, MEDIUM=1, LARGE=2, EXTRA_LARGE=3) to the point
+// sizes shipped with the built-in fonts. Used to drive closest-pt selection
+// in the SD card font registry (see SdCardFontFamilyInfo::pickClosestSize).
+static constexpr uint8_t FONT_SIZE_TO_PT[CrossPointSettings::FONT_SIZE_COUNT] = {12, 14, 16, 18};
```
znelson, line 11: Eventually we should extract this from the .cpfont binary, not hardcoded here. This is now coupling the S/M/L/XL sizes in the settings to the 12/14/16/18 sizes in the font conversion script.

znelson (approved):

PR #1913, opened 2026-05-10
feat: cap compressed group size at 64 KB

jpirnay:
Add GROUP_MAX_UNCOMPRESSED_BYTES=65536 on top of the existing script- based grouping. When adding the next glyph would push a group past the cap, the group is closed and a new one started with the same script ID.

Without the cap, the size of a script group is bounded only by however many glyphs the font supplies in that block. Dense scripts (CJK, Vietnamese precomposed, user-supplied fonts with large Unicode blocks) can produce a single group whose uncompressed size exceeds what fits in the embedded decompressor's transient malloc on the ESP32-C3, which manifests as a runtime allocation failure instead of a build error.

The 64 KB ceiling is large enough to hold any single built-in script group with headroom and small enough to be a comfortable transient allocation on-device.

The check uses byte-aligned size (4-pixel-aligned row stride × height), which is what the decompressor actually consumes — not the packed on-disk length. A defensive guard rejects any single glyph whose own size exceeds the cap with a clear error pointing at the offending codepoint.

## Summary

## Additional Context

  specific areas to focus on).

---

### AI Usage

### DIFF
--- a/lib/EpdFont/scripts/fontconvert.py
+++ b/lib/EpdFont/scripts/fontconvert.py
@@ -792,6 +792,15 @@ if compress:
     # are grouped together for efficient LRU caching on the embedded target.
     # Since glyphs are in codepoint order, glyphs in the same Unicode block
     # are contiguous in the array and form natural groups.
+    #
+    # On top of script boundaries, a hard size cap (GROUP_MAX_UNCOMPRESSED_BYTES)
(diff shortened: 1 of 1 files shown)


mcrosson:
@znelson Can you review this PR

znelson (approved):

PR #1914, opened 2026-05-10
feat: bounded per-codepoint miss diagnostics

jpirnay:
Add a small per-style ring of codepoints already logged as missing for this style since the last cache clear. prewarmStyle() and buildAdvanceTable() count every miss as before but now LOG_DBG only the first MAX_REPORTED_MISSES (32) distinct codepoints per style.

Currently both paths silently drop missing codepoints — useful for the fast path but unhelpful for diagnosing why an SD font looks wrong on a particular document. Logging every miss the first time around is fine, but layout pagination calls prewarm/buildAdvanceTable repeatedly with the same text, so naive logging spams the log with the same handful of codepoints at every pass.

The ring is reset on freeStyleAll() (font teardown) and on clearCache() (section/chapter boundary, where a missing-glyph diagnostic from the next section is genuinely new information). Codepoints that fall off the ring still resolve correctly via the replacement glyph.

## Summary

## Additional Context

  specific areas to focus on).

---

### AI Usage

### DIFF
--- a/lib/EpdFont/SdCardFont.cpp
+++ b/lib/EpdFont/SdCardFont.cpp
@@ -84,7 +84,25 @@ void SdCardFont::freeStyleAll(PerStyle& s) {
   delete[] s.fullIntervals;
   s.fullIntervals = nullptr;
   freeStyleKernLigatureData(s);
+  s.reportedMissCount = 0;
   s.present = false;
+}
+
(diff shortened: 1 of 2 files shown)


mcrosson:

itsthisjustin:
@jeremydk please review this as it's beyond my understanding of the system and you're better suited for this. I trust @jpirnay though in general and assume it adds value

PR #1916, opened 2026-05-10
feat: add latin-cyrillic / greek-letters / musical-symbols presets

jpirnay:
Add three new INTERVAL_PRESETS entries to fontconvert_sdcard.py and re-point existing sd-fonts.yaml entries at them, plus add a 10pt size to every family.

* latin-cyrillic   — combined Latin (incl. punctuation) + extended
                     Cyrillic in one preset, for fonts that ship both
                     scripts. Picks up the three Cyrillic extended
                     blocks (1C80, 2DE0, A640 ranges) on top of plain
                     "cyrillic", since Latin+Cyrillic fonts generally
                     also cover those.
* greek-letters    — modern Greek letters only (0x0370-0x03FF), drops
                     the polytonic block that "greek" includes. Useful
                     for fonts that don't ship polytonic glyphs, where
                     requesting "greek" produces empty slots in
                     0x1F00-0x1FFF.
* musical-symbols  — card suits + Western musical notation (0x2660-
                     0x266F + 0x1D100-0x1D1FF). Small, common in body
                     text (e.g. ♪ ♫ ♬), enabled across all readable
                     families.

Existing presets are unchanged. Existing family entries in sd-fonts.yaml are updated to use the new presets and to add a 10pt size; the previous {12,14,16,18} stays.

Behaviour deltas vs prior config:
- Polytonic Greek (0x1F00-0x1FFF) is no longer requested for fonts that switched from "greek" to "greek-letters". For families that ship modern Greek but no polytonic glyphs (most), this trims empty glyph slots; for those that do ship polytonic, callers can still request "greek" explicitly.
- Cyrillic extended blocks are now requested via "latin-cyrillic" for Latin+Cyrillic families that previously requested "latin-ext,cyrillic".
- Musical-symbols block adds ~20 glyphs to the BMP block plus the SMP notation block. The on-disk format renders SMP cps fine; kerning / ligature data for SMP cps is filtered out by the existing pipeline.
- 10pt size adds one extra .cpfont per family on disk.

## Summary

## Additional Context

  specific areas to focus on).

---

### AI Usage


mcrosson:
We need to be careful about adding a whole bunch of code points/glyphs. There is a ram cost to this and I'd like to be very mindful of this fact before we add more.

I'd like to defer this awhile as there are plenty of other things that have surfaced with fonts that need our attention.

znelson:
I don't understand why this change adds a new, inaccessible font size to the generation process?

insunaa:
Closed on purpose? I was really looking forward to this

jpirnay:
Won't be integrated soon according to the reviewer feedback - so no use in letting stuff lying around - if there's a clearer need then we can revisit it

pablohc:
I think the Hail Mary book would have benefited from this PR trend that’s been pretty popular lately...
Thank you for your work, as always @jpirnay 
I understand that the point of using 10pt was to take up as little flash storage as possible? right?
Maybe using 12pt would be better received by others mantainers?

nameless-madman:
hi, jpirnay. why did you delete your crosspoint++ repo? i like it very much and i use your weather app often

jpirnay:
> hi, jpirnay. why did you delete your crosspoint++ repo? i like it very much and i use your weather app often

I made an honest mistake by reusing the code here, providing ample reference in the PRs and the release notes but losing the github commit author information in the progress. But then all hell broke lose, ending up in insults, harassment and plain lies without even caring for feedback. Fine, i don't need to feed the trolls nor do I need to deal with idiots. So I took everything down. Good riddance.

PR #1917, opened 2026-05-10
feat: add new font families to sd-fonts.yaml

jpirnay:
## Summary
Extend the SD-card font catalogue with 10 families across serif and sans-serif scripts. All entries use stable upstream sources (Google Fonts,  nicoverbruggen/ebook-fonts).

| Category | Family | Source | Variable? |
|---|---|---|---|
| Serif | Gelasio | google/fonts | ✓ wght |
| Serif | Readerly | nicoverbruggen/ebook-fonts | – |
| Serif | Charis | nicoverbruggen/ebook-fonts | – |
| Serif | Sourcerer | nicoverbruggen/ebook-fonts | – |
| Serif | Spectral | google/fonts | – |
| Serif | Vollkorn | google/fonts | ✓ wght |
| Sans | Arimo | google/fonts | ✓ wght |

## Dependency
Eight of the ten families use the `latin-cyrillic` and `musical-symbols` presets introduced in #N (`sdfont-yaml-presets`). This PR will fail to build until that one merges.

### DIFF
--- a/lib/EpdFont/scripts/fontconvert_sdcard.py
+++ b/lib/EpdFont/scripts/fontconvert_sdcard.py
@@ -43,7 +43,24 @@ INTERVAL_PRESETS = {
     "latin-ext":   [(0x0020, 0x007E), (0x0080, 0x00FF), (0x0100, 0x024F),
                     (0x1E00, 0x1EFF), (0x2000, 0x206F), (0xFB00, 0xFB06)],
     "greek":       [(0x0370, 0x03FF), (0x1F00, 0x1FFF)],
+    # Greek monotonic letters only — drops the polytonic block (0x1F00-0x1FFF)
+    # that "greek" includes. Useful for fonts that ship modern Greek but no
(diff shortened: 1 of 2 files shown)


WuTofu:
@jpirnay, I noticed that you only picked simplified chinese fonts of NotoSansCJK and NotoSerifCJK, which may be an issue for CJK readers.
Some CJK characters share the same unicode, but the writings are different. So we might also need to generate fonts for traditional chinese, japanese, and korean.

Also do we want to include CJK font support in this release? I have concerns that including CJK fonts in the manifest at this moment may give users the wrong impression.

mcrosson:
- CJK support is not available in crosspoint. Bluntly: until there is official cjk support, we will not ship CJK fonts. Same goes for hebrew, arabic, thai, etc etc etc.
  - Unless we support a language, will not be shipping a font via our official channels.
  - See #1757 for info on various language support, discussions and progress.
- The modified fonts need to clearly labeled as modified forms of the original fonts and make it clear these are not the original fonts as published.

PR #1919, opened 2026-05-10
feat: add font family deletion functionality

WuTofu:
## Summary

  1. Adds font family deletion support to the font download activity so users can remove installed font families directly from the download list.
  2. Address an issue where a font family manually deleted in the file browser by user will show as "update" in the `FontDownloadActivity`
 

  * Updated button hint label to show `Delete` when deletion is available.
  * Added confirmation before deleting a selected installed font family.
  * Refreshed font registry in `fetchAndParseManifest` (for the second goal above)

---

### AI Usage


WuTofu:
@mcrosson, @pablohc, could you review this pr if you have a sec?

mcrosson:
Should we take this opportunity to rename the menu entry from 'Download fonts' to 'Manage fonts'?

Everything else looks good and my tests pass.

lukestein:
> Should we take this opportunity to rename the menu entry from 'Download fonts' to 'Manage fonts'?

@mcrosson that seems sensible to me. However—and this is maybe really for @WuTofu —I think that change would make most sense if we didn't need to go through the wifi connection before getting the ability to delete a font…

mcrosson:
> > Should we take this opportunity to rename the menu entry from 'Download fonts' to 'Manage fonts'?
> 
> @mcrosson that seems sensible to me. However—and this is maybe really for @WuTofu —I think that change would make most sense if we didn't need to go through the wifi connection before getting the ability to delete a font…

I agree re the 'not having to pull a manifest to delete a font' However, I dont think this is the right PR for that functionality.

I think a different PR that's more of a cleanup pass to make it a bit more useful/flexible would be better.

pablohc:
> @mcrosson, @pablohc, could you review this pr if you have a sec?

Hi @WuTofu, I've been testing it and it works as expected.

I agree with @lukestein’s comment, but I think the problem lies at the source and isn’t specific to this PR.
In a separate PR, we could add access to the cached list of fonts, where you can view the downloaded fonts or delete them. An option to refresh the list in case you want to download or update any of them, requiring Wi-Fi only at that specific step.

pablohc (dismissed):
LGTM

WuTofu:
> I think that change would make most sense if we didn't need to go through the wifi connection before getting the ability to delete a font…

Can't agree more lol. I even grew inpatient when I tested this. Waiting for wifi to connect and download the manifest is kind of painful when you have to test it several times.

> Should we take this opportunity to rename the menu entry from 'Download fonts' to 'Manage fonts'?

@mcrosson @pablohc I have replaced "Download Fonts" with "Manage fonts". With this (pr) change, I also find "Font Download" weird, so I replaced it with "Font Browser".

image

@steka, could you review the swedish translations? I will blindly follow coderabbitai's suggestion for swedish translations here. Could you open a pr if you think the translation is off?

mcrosson (approved):
LGTM 👍

pablohc (approved):
LGTM

steka:
> @steka, could you review the swedish translations? I will blindly follow coderabbitai's suggestion for swedish translations here. Could you open a pr if you think the translation is off?

Sure, and they looks good to me (TGTM)!

PR #1925, opened 2026-05-10
fix: jump page on hold in font family and language selection

zgredex:
## Summary
- Holding the navigation button in **Settings → Reader → Font family** now advances the selection by a full visible page instead of one item at a time.
- Same fix applied to **Settings → Language**, which had the same one-item-only behavior.
- Mirrors the pattern already used in font download (ceb3fed) and chapter selection screens.

## Test plan
- [ ] Settings → Reader → Font family: tap moves by one item; hold jumps a page (wraps at ends)
- [ ] Settings → Language: tap moves by one item; hold jumps a page (wraps at ends)

### DIFF
--- a/src/activities/settings/FontSelectionActivity.cpp
+++ b/src/activities/settings/FontSelectionActivity.cpp
@@ -60,13 +60,26 @@ void FontSelectionActivity::loop() {
     return;
   }
 
-  buttonNavigator_.onNextRelease([this] {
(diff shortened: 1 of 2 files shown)


zgredex:
Same kind of nitpick as my last fix. With all fonts downloaded the list can exceed one page. This brings font selector and language selector behavior in line with the rest of the system

Uri-Tauber (approved):
The code seems fine. Also, tested on my device, and seems to work as expected.

mcrosson (approved):
LGTM

PR #1929, opened 2026-05-10
fix: handle fallbacks for advance table and prewarm

### LINKED ISSUE #1928: Fix: Advance table and prewarm caches do not account for font style fallbacks
### Affected Version

master

### Describe the Bug

SD font style fallback is only applied during actual rendering, not during cache preparation. 

If an EPUB requests bold/italic text but the selected SD font only contains regular, render-time fallback correctly draws with regular, but prewarm() and buildAdvanceTable() previously used the requested style mask literally and skipped unavailable styles. This causes bitmap prewarm to miss the real fallback glyphs, leading to slow on-demand SD/cache misses during page render; for the advance table, missing fallback-style entries can return zero advances once the fast path is active, causing incorrect text measurement and wrapping. 

Cache preparation should resolve requested styles to the actual available fallback styles before warming metrics or bitmaps.

(@mcrosson per discussion)

### Steps to Reproduce

1. Generate and use a regular-only set of cpfonts
2. Open a bold heavy epub
3. See massive slow down

### Expected Behavior

No change in perf

```shell

```

leecming82:
## Summary
- Fixes #1928 by having the prewarm and advance table functions resolve fallback styles
---

### AI Usage

### DIFF
--- a/lib/EpdFont/SdCardFont.cpp
+++ b/lib/EpdFont/SdCardFont.cpp
@@ -1198,9 +1198,13 @@ bool SdCardFont::hasStyle(uint8_t style) const { return styles_[style & (MAX_STY
 
 uint8_t SdCardFont::resolveStyle(uint8_t style) const {
   static const uint8_t kFallbacks[MAX_STYLES][MAX_STYLES] = {
+      // REGULAR: REGULAR -> BOLD -> ITALIC -> BOLD_ITALIC
       {EpdFontFamily::REGULAR, EpdFontFamily::BOLD, EpdFontFamily::ITALIC, EpdFontFamily::BOLD_ITALIC},
+      // BOLD: BOLD -> REGULAR -> BOLD_ITALIC -> ITALIC
       {EpdFontFamily::BOLD, EpdFontFamily::REGULAR, EpdFontFamily::BOLD_ITALIC, EpdFontFamily::ITALIC},
+      // ITALIC: ITALIC -> REGULAR -> BOLD_ITALIC -> BOLD
       {EpdFontFamily::ITALIC, EpdFontFamily::REGULAR, EpdFontFamily::BOLD_ITALIC, EpdFontFamily::BOLD},
+      // BOLD_ITALIC: BOLD_ITALIC -> BOLD -> ITALIC -> REGULAR
       {EpdFontFamily::BOLD_ITALIC, EpdFontFamily::BOLD, EpdFontFamily::ITALIC, EpdFontFamily::REGULAR},
   };
 
--- a/lib/EpdFont/SdCardFont.cpp
+++ b/lib/EpdFont/SdCardFont.cpp
@@ -1197,7 +1197,12 @@ bool SdCardFont::hasStyle(uint8_t style) const { return styles_[style & (MAX_STY
 uint8_t SdCardFont::resolveStyle(uint8_t style) const {
   // Keep this fallback order aligned with EpdFontFamily::getFont() and the
   // renderer's SD advance-table fast path.
-  static const uint8_t kFallbacks[MAX_STYLES][MAX_STYLES] = {
+  static const uint8_t kFallbacks[4][4] = {
+      {EpdFontFamily::REGULAR, EpdFontFamily::BOLD, EpdFontFamily::ITALIC, EpdFontFamily::BOLD_ITALIC},
+      {EpdFontFamily::BOLD, EpdFontFamily::BOLD_ITALIC, EpdFontFamily::REGULAR, EpdFontFamily::ITALIC},
+      {EpdFontFamily::ITALIC, EpdFontFamily::REGULAR, EpdFontFamily::BOLD, EpdFontFamily::BOLD_ITALIC},
+      {EpdFontFamily::BOLD_ITALIC, EpdFontFamily::BOLD, EpdFontFamily::ITALIC, EpdFontFamily::REGULAR},
+  };
       {0, 1, 2, 3},  // regular
       {1, 3, 0, 2},  // bold
       {2, 0, 1, 3},  // italic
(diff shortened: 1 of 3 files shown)


mcrosson:
@znelson / @itsthisjustin can you take a look at this. Y'all are more familiar with the cache code than I am.

This looks good to me but I'd like your input.

lib/EpdFont/SdCardFont.cpp
```diff
@@ -1190,6 +1194,33 @@ EpdFont* SdCardFont::getEpdFont(uint8_t style) {
 
 bool SdCardFont::hasStyle(uint8_t style) const { return styles_[style & (MAX_STYLES - 1)].present; }
 
+uint8_t SdCardFont::resolveStyle(uint8_t style) const {
+  // Keep this fallback order aligned with EpdFontFamily::getFont() and the
+  // renderer's SD advance-table fast path.
+  static const uint8_t kFallbacks[MAX_STYLES][MAX_STYLES] = {
```
Uri-Tauber, line 1200: Would be easier to understand and maintain with the enum:

```suggestion
  static const uint8_t kFallbacks[4][4] = {
      {EpdFontFamily::REGULAR, EpdFontFamily::BOLD, EpdFontFamily::ITALIC, EpdFontFamily::BOLD_ITALIC},
      {EpdFontFamily::BOLD, EpdFontFamily::BOLD_ITALIC, EpdFontFamily::REGULAR, EpdFontFamily::ITALIC},
      {EpdFontFamily::ITALIC, EpdFontFamily::REGULAR, EpdFontFamily::BOLD, EpdFontFamily::BOLD_ITALIC},
      {EpdFontFamily::BOLD_ITALIC, EpdFontFamily::BOLD, EpdFontFamily::ITALIC, EpdFontFamily::REGULAR},
  };
```

itsthisjustin:
Valid fix. The cache was definitely built in mind assuming all styles

itsthisjustin (dismissed):

lib/EpdFont/SdCardFont.cpp
```diff
@@ -1190,6 +1196,33 @@ EpdFont* SdCardFont::getEpdFont(uint8_t style) {
 
 bool SdCardFont::hasStyle(uint8_t style) const { return styles_[style & (MAX_STYLES - 1)].present; }
 
+uint8_t SdCardFont::resolveStyle(uint8_t style) const {
+  // Keep this fallback order aligned with EpdFontFamily::getFont() and the
```
znelson: This is a materially different fallback behavior than `EpdFontFamily::getFont`, which does:
- `REGULAR`: no fallback
- `BOLD`: `REGULAR`
- `ITALIC`: `REGULAR`
- `BOLD_ITALIC`: `BOLD`, `ITALIC`, `REGULAR`

Not that this implementation is necessarily wrong, but the comment is misleading.

lib/EpdFont/SdCardFont.cpp
znelson: ~~I'm not sure I understand the feedback. Suggesting we replace `MAX_STYLES` with `4`?~~ 

Disregard, the GitHub presentation of your comment against the newer diff was confusing. Your suggestion was already applied.

lib/EpdFont/SdCardFont.cpp
```diff
@@ -1190,6 +1196,33 @@ EpdFont* SdCardFont::getEpdFont(uint8_t style) {
...
+  // renderer's SD advance-table fast path.
+  static const uint8_t kFallbacks[MAX_STYLES][MAX_STYLES] = {
+      {EpdFontFamily::REGULAR, EpdFontFamily::BOLD, EpdFontFamily::ITALIC, EpdFontFamily::BOLD_ITALIC},
+      {EpdFontFamily::BOLD, EpdFontFamily::BOLD_ITALIC, EpdFontFamily::REGULAR, EpdFontFamily::ITALIC},
+      {EpdFontFamily::ITALIC, EpdFontFamily::REGULAR, EpdFontFamily::BOLD, EpdFontFamily::BOLD_ITALIC},
```
znelson: Why `BOLD` -> `BOLD_ITALIC` as a fallback, but `ITALIC` -> `REGULAR`?

Can you explain how you came up with this ordering?

Uri-Tauber: I believe it’s coming from my #1921.

My thinking was that `BOLD_ITALIC → ITALIC → REGULAR` was the most logical order, and the same applied to italic. However, I ran into some issues with italic words, so I moved `REGULAR` to the front. I didn’t update the bold logic since it wasn’t affected by those issues. I agree that both should have a similar pattern.

lib/EpdFont/SdCardFont.cpp
leecming82: My bad, removed the comment

lib/EpdFont/SdCardFont.cpp
leecming82: Yup, it was just taken from #1921. 

How about BOLD->REGULAR->BOLD_ITALIC->ITALIC
and  ITALIC->REGULAR->BOLD_ITALIC->BOLD as per my new commit with the argument that neutral is a better fallback than forcing BOLD or ITALIC onto a request that did not ask for either

znelson (dismissed):

lib/EpdFont/SdCardFont.cpp
```diff
@@ -1190,6 +1196,35 @@ EpdFont* SdCardFont::getEpdFont(uint8_t style) {
...
+      {EpdFontFamily::ITALIC, EpdFontFamily::REGULAR, EpdFontFamily::BOLD_ITALIC, EpdFontFamily::BOLD},
+      // BOLD_ITALIC: BOLD_ITALIC -> BOLD -> ITALIC -> REGULAR
+      {EpdFontFamily::BOLD_ITALIC, EpdFontFamily::BOLD, EpdFontFamily::ITALIC, EpdFontFamily::REGULAR},
+  };
+
+  const uint8_t styleBits = style & (MAX_STYLES - 1);
+  for (uint8_t candidate : kFallbacks[styleBits]) {
+    if (styles_[candidate].present) return candidate;
+  }
+  return 0;
```
Uri-Tauber: would be more clear
```suggestion
  return EpdFontFamily::REGULAR;
```

Uri-Tauber (approved):

PR #1934, opened 2026-05-10
fix: show convert icon usage without optional deps

sabraman:
## Summary

* Let `scripts/convert_icon.py` print its usage message before importing optional image conversion dependencies.
* Correct the usage text to reference `scripts/convert_icon.py` instead of the non-existent `convert_image.py`.

## Additional Context

Running `python3 scripts/convert_icon.py` without arguments should show usage, but it could fail first if `cairosvg` was not installed. Moving the image library imports into the conversion paths keeps the basic CLI help available before installing script dependencies.

Verified with:

```sh
python3 scripts/convert_icon.py
python3 -m py_compile scripts/convert_icon.py
```

### DIFF
--- a/scripts/convert_icon.py
+++ b/scripts/convert_icon.py
@@ -1,18 +1,20 @@
-import sys
 import os
-from PIL import Image
-import cairosvg
 import io
+import sys
 
 threshold = 128
 
 def svg_to_png_bytes(svg_path, width, height):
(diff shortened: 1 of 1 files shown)


sabraman:
Closing in favor of #1937, which consolidates this script help/startup fix with the remaining related help-path fixes.

PR #1940, opened 2026-05-11
docs: update delete endpoint reference

sabraman:
## Summary

Update the `/delete` endpoint reference to match the current web server handler.

## Details

The endpoint docs still described a `type` form field and the old `Deleted successfully` response. The handler now accepts either `path` for a single item or `paths` as a JSON array for multi-delete, infers file versus folder from the SD card entry, and returns `All items deleted successfully` when all deletes complete.

This updates the curl examples, parameter table, success response, and error response list to match the implemented behavior.

## Validation

- Compared the documented parameters and response strings with `CrossPointWebServer::handleDelete()`
- `git diff --check`

### DIFF
--- a/docs/webserver-endpoints.md
+++ b/docs/webserver-endpoints.md
@@ -204,40 +204,41 @@ Folder created: NewFolder
 
 ### POST `/delete` - Delete File or Folder
 
-Deletes a file or folder from the SD card.
+Deletes one or more files or empty folders from the SD card.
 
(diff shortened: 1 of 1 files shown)


itsthisjustin (approved):

itsthisjustin:
Approved on basis of scope and value. @znelson if you validate it, we can get it merged.

znelson (approved):

PR #1959, opened 2026-05-12
fix: prune books missing form sd card in recent books list

mcrosson:
## Summary

When entering the recent books list or opening a new book, prune any books no longer on the sd card from the recent books list. This frees slots so the queue/list of recent books gets updated cleanly. Prior to this change there were times the book list wouldn't fill to '10' and it was due to books being removed from sd card.

## Additional Context

If a user does delete a book by accident, they can restore it prior to opening another book or opening the recent list to keep the accidentally deleted book in place on the recent book list.

---

### AI Usage

Did you use AI tools to help write this code? **Yes, claude**

### DIFF
--- a/src/activities/home/HomeActivity.cpp
+++ b/src/activities/home/HomeActivity.cpp
@@ -43,7 +43,7 @@ void HomeActivity::loadRecentBooks(int maxBooks) {
     }
 
     // Skip if file no longer exists
-    if (!Storage.exists(book.path.c_str())) {
+    if (RecentBooksStore::isMissing(book)) {
       continue;
     }
 
--- a/src/activities/home/RecentBooksActivity.cpp
+++ b/src/activities/home/RecentBooksActivity.cpp
@@ -15,10 +15,7 @@ namespace {
 constexpr unsigned long GO_HOME_MS = 1000;
 }  // namespace
 
-void RecentBooksActivity::loadRecentBooks() {
-  const auto& books = RECENT_BOOKS.getBooks();
-  recentBooks = books;
-}
+void RecentBooksActivity::loadRecentBooks() { recentBooks = RECENT_BOOKS.getBooks(); }
 
 void RecentBooksActivity::onEnter() {
   Activity::onEnter();
--- a/src/activities/home/RecentBooksActivity.cpp
+++ b/src/activities/home/RecentBooksActivity.cpp
@@ -16,21 +16,18 @@ constexpr unsigned long GO_HOME_MS = 1000;
 }  // namespace
 
 void RecentBooksActivity::loadRecentBooks() {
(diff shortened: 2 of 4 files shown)


src/activities/home/RecentBooksActivity.cpp
```diff
@@ -16,22 +16,19 @@ constexpr unsigned long GO_HOME_MS = 1000;
...
-  recentBooks.reserve(books.size());
-
-  for (const auto& book : books) {
-    // Skip if file no longer exists
-    if (!Storage.exists(book.path.c_str())) {
-      continue;
-    }
-    recentBooks.push_back(book);
-  }
+  recentBooks = books;
```
znelson: You don't need `books` local. Just:

```
recentBooks = RECENT_BOOKS.getBooks();
```

znelson (dismissed):
`HomeActivity::loadRecentBooks` is doing its own independent prune. Consider refactoring that to share code with this new impl.

src/activities/home/RecentBooksActivity.cpp
mcrosson: Addressed in latest commit

src/activities/home/HomeActivity.cpp
```diff
@@ -43,7 +43,7 @@ void HomeActivity::loadRecentBooks(int maxBooks) {
     }
 
     // Skip if file no longer exists
-    if (!Storage.exists(book.path.c_str())) {
```
znelson, line 46: It's fine to iterate on this in another change, just curious for your thoughts: why not have the RecentBooksStore prune itself here, instead of just checking and skipping missing books on-the-fly?

znelson (approved):

src/activities/home/HomeActivity.cpp
mcrosson: I had Claude set this up so if someone deletes an epub by mistake and immediately restores it, the recent books list wont get changed. The more full rationale it spat out when asked:

```
 1. Add an explanatory comment at the isMissing check

 File: src/activities/home/HomeActivity.cpp — loadRecentBooks (:45-48 (src/activities/home/HomeActivity.cpp))

 Replace the existing one-line comment:

     // Skip if file no longer exists
     if (RecentBooksStore::isMissing(book)) {
       continue;
     }

 with:

     // Display-only filter: a temporarily-missing file (SD swapped, deleted-then-restored)
     // should silently reappear here without losing its recent-list entry. The persistent
     // store is pruned only at two commit points: RecentBooksStore::addBook() and
     // RecentBooksActivity::onEnter().
     if (RecentBooksStore::isMissing(book)) {
       continue;
     }

```

Uri-Tauber (approved):

PR #1961, opened 2026-05-13
refactor: Eliminated relative path includes

znelson:
## Summary

Relative includes can hide inappropriate dependency relationships. In this case, I found that lib/KOReaderSync/KOReaderCredentialStore.cpp was dependent on src/JsonSettingsIO.h -- a lib -> app dependency, the opposite direction dependencies should flow in this project.

This change replaces all relative includes with root-relative includes, and corrects the KOReaderCredentialStore dependency by moving its JSON settings serialization local to the KOReaderSync library.

---

### AI Usage


lib/KOReaderSync/KOReaderCredentialStore.cpp
```diff
@@ -6,7 +6,7 @@
 #include <ObfuscationUtils.h>
 #include <Serialization.h>
 
-#include "../../src/JsonSettingsIO.h"
```
znelson, line 9: This code in lib/ should not be using JsonSettingsIO.h from the app code.

Uri-Tauber (approved):
Nice

PR #1965, opened 2026-05-13
fix: several QoL updates for SD font's UI

WuTofu:
## Summary

Improve the UI based on feedback from someone on discord

> Downloading ALL fonts feature.
> 1.1 Disable sleep when downloading, in my case went directly to sleep just right after downloading.
> 1.2 It would be great to have and overall progress indicator as we only have the indication of each font family
> 1.3 Any cancel or pause function might come in handy in case battery is running out and then resume or retry with pending fonts

  -  Now the UI can show overall progress across every file being downloaded in the batch, not just progress inside the current family.
  - Extended `HttpDownloader::downloadToFile()` to accept a cancel flag and abort the download.
  - Rendered a cancel button in the font download UI while a download is in progress.
  - `preventAutoSleep()` in `FontDownloadActivity.h` now returns true for `state_ == COMPLETE` and `state_ == ERROR` in addition to `LOADING_MANIFEST` and `DOWNLOADING`

## Additional Context

Not very satisfied with how `HttpDownloader.cpp` is right now, might try to refactor it after v1.3.0

---

### AI Usage

### DIFF
--- a/src/activities/browser/OpdsBookBrowserActivity.cpp
+++ b/src/activities/browser/OpdsBookBrowserActivity.cpp
@@ -274,7 +274,7 @@ void OpdsBookBrowserActivity::downloadBook(const OpdsEntry& book) {
         downloadTotal = total;
         requestUpdate(true);
       },
-      server.username, server.password);
+      nullptr, server.username, server.password);
 
   if (result == HttpDownloader::OK) {
     Epub(filename, "/.crosspoint").clearCache();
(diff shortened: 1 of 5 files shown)


uxjulia:
I was just thinking how this needed a cancel button. I accidentally clicked on "Download all" and the only way I could bail was by forcing a reset.

pablohc (approved):
Thank you very much for this improvement @WuTofu

znelson (approved):

PR #1973, opened 2026-05-14
fix: Prepare SD card font caches from txt reader

znelson:
## Summary

SD card font fixes:
- `TxtReaderActivity` needs to call `renderer.ensureSdCardFontReady` to build the advance lookup table to support rendering with SD card fonts. This revealed that `TxtReaderActivity` was inconsistently performing layout with `getTextWidth`, when the renderer actually uses `getTextAdvanceX`, which can lead to minor inconsistencies in alignment. 
- Avoid allocating one big `allText` string in `ParsedText::layoutAndExtractLines`. Instead, pass the vector of word strings directly to `SdCardFont::buildAdvanceTable`, where the algorithm just needs to iterate codepoints anyway.

---

### AI Usage


Uri-Tauber:
@znelson Can you address the nitpick comment of CodeRabbit? I think it had a point.

Uri-Tauber (commented):
Overall looks good.
Did you benchmarked this?

znelson:
I haven't done any benchmarking beyond confirming that .txt files open in a reasonable amount of time with an SD card font, rather than crashing.

Uri-Tauber:
Aha. I didn't understand you worked on this bug. OK, I had the impression justin did something about this.

Uri-Tauber (dismissed):

znelson:
@Uri-Tauber understandable confusion -- Justin pushed the fix to this PR branch. Then the CodeRabbit concern you mentioned earlier came from Justin's commit, which was partially due to my original refactoring to pass a vector directly.

mcrosson (approved):
Tested with 2 txt files, both well over 20k. One took 'long' (less than a minute) and the other was near instant. This definitely solves the problem it aims to solve.

PR #1974, opened 2026-05-14
refactor: extract Typesetter from ChapterHtmlSlimParser

znelson:
## Summary

This is the first of several changes to isolate the formatted text rendering engine from Epub, to be reused by Txt and Markdown readers.

This change extracts page layout logic from `ChapterHtmlSlimParser` into a new `Typesetter` class (`lib/Typesetter/`). The parser retains HTML/XML parsing and style resolution, while `Typesetter` now owns all page construction state and exposes four operations:
- `submitParagraph` for laying out a complete text block
- `partialFlush` for early-flushing oversized blocks (the 750+ word path)
- `submitImage` for placing images with page-break handling
- `finish` for flushing the final in-progress page. 

There are no behavioral changes; this is a pure mechanical extraction.

At the moment, `Typesetter` depends on types in `lib/Epub/` (`Page`, `ParsedText`, `ImageBlock`, `FootnoteEntry`), which creates a bidirectional library reference. Those types are self-contained layout primitives with no back-dependency on the parser, so the cycle is structural rather than logical. Extracting them into `Typesetter` will come in future changes. Eventually `Typesetter` will sit _below_ `Epub`.

---

### AI Usage

### DIFF
--- a/lib/Epub/Epub/parsers/ChapterHtmlSlimParser.cpp
+++ b/lib/Epub/Epub/parsers/ChapterHtmlSlimParser.cpp
@@ -128,7 +128,7 @@ void ChapterHtmlSlimParser::startNewTextBlock(const BlockStyle& blockStyle) {
       currentTextBlock->setBlockStyle(style.getCombinedBlockStyle(blockStyle, BlockStyle::CombineAxis::Vertical));
 
       if (!pendingAnchorId.empty()) {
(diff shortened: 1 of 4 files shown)


znelson:
#1971 paints a more complete picture of the refactoring this is building toward.

Uri-Tauber:
I haven't looked at the code, but I like the direction.
Can it wait until #1700 is merged?

znelson:
@Uri-Tauber Ok, can you update that PR so it's not sitting on conflicts?

PR #1975, opened 2026-05-14
chore: Removed unused icon header files

znelson:
## Summary

Deleted two unused header files containing binary icon data.

---

### AI Usage


Uri-Tauber (approved):
Very good.
I wonder if we can manually draw the icons (like the battery, for example), but that's another PR.

PR #1977, opened 2026-05-14
fix: take orientation into account for border generation in ScreenshotUtil

Disasm:
## Summary

Previously `ScreenshotUtil` used physical display size to draw a border around the screen contents. Because of this, in landscape orientation the border was shown as a broken square. This PR changes border drawing to use logical screen size instead of a physical display size to take orientation into account.

## Additional Context

* Tested on X4 in all 4 reading orientations. Behavior is now correct, however it doesn't look perfect on my X4: the border is much closer to the physical top side of the display than to the other sides. This might be related to assembly variation during manufacturing, but it might as well be related to the way a eink controller is connected to the display (controller supports bigger display sizes, so an offset may be present).

---

### AI Usage

### DIFF
--- a/src/util/ScreenshotUtil.cpp
+++ b/src/util/ScreenshotUtil.cpp
@@ -88,7 +88,12 @@ void ScreenshotUtil::takeScreenshot(GfxRenderer& renderer) {
 
   // Display a border around the screen to indicate a screenshot was taken
   if (renderer.storeBwBuffer()) {
-    renderer.drawRect(6, 6, renderer.getDisplayHeight() - 12, renderer.getDisplayWidth() - 12, 2, true);
+    int marginTop, marginRight, marginBottom, marginLeft;
+    renderer.getOrientedViewableTRBL(&marginTop, &marginRight, &marginBottom, &marginLeft);
+    int width = renderer.getScreenWidth() - marginLeft - marginRight - 1;
+    int height = renderer.getScreenHeight() - marginTop - marginBottom - 1;
+    // Add extra margin to the border to make it more visible
+    renderer.drawRect(marginLeft + 1, marginTop + 1, width - 2, height - 2, 2, true);
     renderer.displayBuffer();
     delay(1000);
     renderer.restoreBwBuffer();


lukestein:
Just a note @Disasm on your observation about screen assembly and full bleed… there's a discussion thread where folks have documented the degree of variation in their usable margins. Right now I believe we continue to rely on Dave's original sensible defaults, but they def don't result in perfect behavior on every physical device

Disasm:
Oh nice, thanks for pointing this out!

Disasm:
Added a call to `getOrientedViewableTRBL` to compensate for this.

itsthisjustin:
This is super minimal risk and LOC so I'm inclined to approve but @Disasm is it ready for merge?

Disasm:
@itsthisjustin yes, it's ready for merge. I'm just waiting for someone to review the PR.

itsthisjustin (approved):

itsthisjustin:
@lukestein are you able to approve and merge if it's good to go? I approve based on scope and risk but I have not reviewed the implementation as this is an area of the code I'm unfamiliar with

lukestein (approved):

lukestein:
> @lukestein are you able to approve and merge if it's good to go? I approve based on scope and risk but I have not reviewed the implementation as this is an area of the code I'm unfamiliar with

Code seems reasonable to me (though code review is not my expertise here).

I do not have the ability to test on x3, but **this tests well for me in all four screen orientations, and also respects the custom screen margins** I have set in my `GfxRenderer.h` (so that Crosspoint will respect my screen which is installed slightly shifted from what crosspoint expects).

lukestein:
> Oh nice, thanks for pointing this out!

btw @Disasm (or anyone else): You can set your own screen margins in a custom build by editing e.g.,
```
  static constexpr int VIEWABLE_MARGIN_TOP = 7;
  static constexpr int VIEWABLE_MARGIN_RIGHT = 3;
  static constexpr int VIEWABLE_MARGIN_BOTTOM = 4;
  static constexpr int VIEWABLE_MARGIN_LEFT = 3;
```

in `GfxRenderer.h`.

PR #1981, opened 2026-05-14
fix: navigate to TOC anchor when selecting sub-chapters

Disasm:
Chapter selection previously only used the spine index, causing navigation to always land on page 0 of the spine item. Sub-chapters that share a spine file but differ by anchor (e.g. `chapter.xhtml#sec2`) were silently ignored. Now the TOC anchor is passed through `ChapterResult` and applied via the existing `pendingAnchor` mechanism.

## Summary

* This PR implements navigation to sub-chapters which didn't work correctly previously. If a sub-chapter of the current top-level chapter was selected, nothing happened. If a sub-chapter of another top-level chapter was selected, reader switched to the beginning of the top-level chapter.
* In addition, chapters now always start from a new page. This fixes anchor to page calculation for the cases when the actual chapter content doesn't fit on the page where the corresponding ToC anchor was found.

## Additional Context

* I might misuse `pendingAnchor` here which was previously used for footnote navigation, please double check. I'm open to suggestions for improvements.
* Note that the chapter selected by default when `EpubReaderChapterSelectionActivity` opens is still wrong. I'm going to fix this separately. This PR addresses only navigation to the selected chapter.
* I tested this PR on my X4 and verified that navigation to a different sub-chapter works correctly, both inside and outside the current spine.
* Some of the changes were borrowed from https://github.com/crosspoint-reader/crosspoint-reader/pull/1455
---

### AI Usage

### DIFF
--- a/src/activities/reader/EpubReaderChapterSelectionActivity.cpp
+++ b/src/activities/reader/EpubReaderChapterSelectionActivity.cpp
@@ -48,14 +48,14 @@ void EpubReaderChapterSelectionActivity::loop() {
   const int totalItems = getTotalItems();
 
   if (mappedInput.wasReleased(MappedInputManager::Button::Confirm)) {
-    const auto newSpineIndex = epub->getSpineIndexForTocIndex(selectorIndex);
-    if (newSpineIndex == -1) {
+    const auto tocItem = epub->getTocItem(selectorIndex);
+    if (tocItem.spineIndex == -1) {
       ActivityResult result;
       result.isCancelled = true;
       setResult(std::move(result));
       finish();
     } else {
-      setResult(ChapterResult{newSpineIndex});
+      setResult(ChapterResult{tocItem.spineIndex, tocItem.anchor});
       finish();
     }
   } else if (mappedInput.wasReleased(MappedInputManager::Button::Back)) {
--- a/lib/Epub/Epub/Section.cpp
+++ b/lib/Epub/Epub/Section.cpp
@@ -10,7 +10,7 @@
 #include "parsers/ChapterHtmlSlimParser.h"
 
 namespace {
-constexpr uint8_t SECTION_FILE_VERSION = 23;
+constexpr uint8_t SECTION_FILE_VERSION = 24;
 constexpr uint32_t HEADER_SIZE = sizeof(uint8_t) + sizeof(int) + sizeof(float) + sizeof(bool) + sizeof(uint8_t) +
                                  sizeof(uint16_t) + sizeof(uint16_t) + sizeof(uint16_t) + sizeof(bool) + sizeof(bool) +
                                  sizeof(uint8_t) + sizeof(bool) + sizeof(uint32_t) + sizeof(uint32_t) +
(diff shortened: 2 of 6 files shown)


Uri-Tauber:
Please compare with #1455

Disasm:
Yeah, I didn't notice this PR at first. My PR is a small subset of #1455, so I'm going wait until it's merged and then close mine or update it if needed.

znelson:
#1455 is months old, I’m moving on. Feel free to go ahead with this PR.

Uri-Tauber:
since @znelson closed his #1455, I really wants to get this one reviewed and merged.

Disasm:
I tested this change recently with updated `master` and noticed that sometimes navigation jumps to a page right before anchor, which is not how I'd expect it to work. However, since current version of navigation if far less perfect, I'd propose to integrate this PR as it is. I hope I'll have time to improve the implementation with follow-up PRs.

znelson:
You're more than welcome to borrow the changes from #1455 if they meet the needs more completely! Either revive and update that PR, or just directly integrate the changes into yours here. None of the code is precious; it's all free to use however you want 🙂

itsthisjustin:
@Uri-Tauber can we get this merged?

Uri-Tauber:
@itsthisjustin I would prefer @Disasm would fix the bug he mentioned two messages earlier, rather than having follow-up PR with a fix.

lukestein:
Testing on device and seems to work great _except_ that—consistent with @Disasm's note—if the target appears on the first line of a page, the navigation seems to jump to the preceding page.

If the fix here can be identified, I would like to see it get implemented before we merge. But if it's a tricky bug, I think it's worth considering merging and fixing afterwards since multiple [sub]chapters in a single spine has been a consistent pain point in my core reading experience.

Disasm:
Ok, now it seems to work correctly. Ready for review.

lukestein:
Thanks for this @Disasm. I'm going to continue with some more on-device testing but this fix seems to be working 🎉.

One thing I'll note in case other reviewers or community members want to weigh in: This PR introduces a page break above each subchapter anchor. I think I am okay with that in order to have stable targeting of the TOC links, but it does alter the reading experience for certain books. And I could imagine not everyone loving the forced page breaks. Not saying we want to take it out, but food for thought.

Uri-Tauber (approved):

Uri-Tauber:
I haven't tested the latest changes on my device, but the code looks clean, and I trust @lukestein to do the testing here.

src/activities/reader/EpubReaderChapterSelectionActivity.cpp
```diff
@@ -28,18 +44,18 @@ void EpubReaderChapterSelectionActivity::onEnter() {
...
-    if (newSpineIndex == -1) {
+    const auto tocItem = epub->getTocItem(selectorIndex);
+    if (tocItem.spineIndex == -1) {
       ActivityResult result;
       result.isCancelled = true;
       setResult(std::move(result));
       finish();
     } else {
-      setResult(ChapterResult{newSpineIndex});
+      setResult(ChapterResult{tocItem.spineIndex, tocItem.anchor});
```
znelson, line 58: One `move` opportunity here (need to remove `const` on `tocItem` above):
```suggestion
      setResult(ChapterResult{tocItem.spineIndex, std::move(tocItem.anchor)});
```

znelson (approved):
LGTM, the parser fix looks correct and this is a real improvement. A few areas to keep an eye out for (feel free to draw inspiration from, or just outright copy, #1455):

**Long-press chapter skip jumps by spine, not TOC entry: on a multi-TOC spine, long-press skips all sub-chapters at once.**

> 1455's `EpubReaderActivity::loop()` walks TOC entries for chapter skip. Same-spine sibling chapters resolve via `section->getPageForTocIndex()`, cross-spine via `pendingTocIndex`. Includes guards for orphan spines (`curTocIndex < 0`) and end-of-book clamping.
>
> 1981 doesn't touch this code path. Long-press still skips by spine boundary, so on a book that packs N TOC chapters into one spine file, long-press jumps over all N of them at once.

**Status bar shows the wrong sub-chapter title: falls back to the first TOC entry for the spine, so sub-chapters 2..N show the parent chapter's name.**

> 1455's `renderStatusBar()` uses `section->getTocIndexForPage(section->currentPage)` so the title updates as the reader pages through sub-chapters within a multi-TOC spine.
>
> 1981 doesn't touch this. Status bar still falls back to `epub->getTocIndexForSpineIndex(currentSpineIndex)`, which returns the first TOC entry for the spine, rendering incorrectly for sub-chapters 2..N of a multi-TOC spine.

**`Section::tocBoundaries` + per-page TOC lookup**

> Both of the above features depend on `Section::getTocIndexForPage()` / `getPageForTocIndex()`, backed by an in-memory `tocBoundaries` vector built from the anchor map on section load. 1981 doesn't add this infrastructure at all: Section.h is untouched, and Section.cpp only adds the `tocAnchors` collection passed to the parser, not the disk-scan-and-build of `tocBoundaries` (1455's `buildTocBoundaries` covers both the create-time in-memory path and the load-time disk path).
>
> So even if someone wanted to wire status-bar/long-press in a follow-up, they'd need that lookup structure first.

**Default-selected chapter in the selector**

_Already acknowledged in the PR description, just highlighting here for completeness:_
> 1455 passes a `tocIdx` argument to `EpubReaderChapterSelectionActivity` so the selector opens on the actual current sub-chapter. 1981 still passes only `spineIdx`.

lukestein (approved):
Works well in my testing. I'll leave the merging to @znelson since he has some open comments and expertise on the code

Disasm:
@znelson valid comments, I'll try do address at least some of them in the following PRs. This PR is just for navigation by selecting a chapter from a menu. My goal was to make this fix as small as possible to simplify the review process and increase probability of integration.

PR #1986, opened 2026-05-14
refactor: Named constants for Section cache offsets

znelson:
## Summary

Added named `constexpr` constants for all of the Section header byte offset values.

---

### AI Usage


Uri-Tauber (approved):
I really like this. this will definitely simplify maintenance. One nitpick: can we swap uint32_t for uint16_t? A 64KB limit should be enough.

Uri-Tauber:
This is super important. Why did you closed it?

PR #1990, opened 2026-05-15
fix: Add documentation for USB-locked Xteink devices

itsthisjustin:
Document the Xteink Unlocker tool requirement for third-party purchased xteink units that ship with USB flashing locked. Include warnings about bricking risks when flashing unsupported firmwares (e.g. Papyrix) on locked devices, as they may permanently lock the device with no recovery path.


znelson (approved):

mcrosson (approved):
Paging @znelson

PR #2000, opened 2026-05-16
feat: Add memory stats to Home screen

leecming82:
## Summary
Show heap memory stats on the Home Screen status bar - SD fonts and other new features have added memory pressure, would be useful to be able to eyeball memory usage.

Defaults to Hidden, can be set in Settings->Display

image

### AI Usage


Uri-Tauber:
Hi @leecming82,

First of all, congrats on the #2000 PR!

Just a thought — this might be a bit misleading or not very useful. I’ve used the debugging monitor quite a bit, and I’ve noticed there doesn’t seem to be any relationship between the memory state on the home page and the reader; they appear to be completely separate. All the memory get released when exiting the reader - so the numbers are relevant only to the home screen itself...

So I’m wondering how the stats here would actually be useful in practice.

leecming82:
Thx for taking a look. I did think about placing the stats in the reader menu so that's an alternative.

I don't think it's right to say all memory is released when exiting the reader though - my screenshot shows 56K free heap but only 25K max allocatable as a block so there's at the very least some heap fragmentation going on

leecming82:
Nice to have but not a pressing concern so closing

Issue #1954, opened 2026-05-12
 Landscape CCW: page-turn order setting does not affect the physical right-edge buttons


cmorganenbw:
### Affected Version

1.2.0

### Describe the Bug

 When reading in Landscape CCW, the bottom/front buttons become the physical buttons on the right edge of the device.

  I hold the reader in my right hand in this orientation and use those right-edge buttons for page turning. I expected Settings → Controls → Side Button Layout (reader) to let me invert the page-
  turn direction for that physical button position.

  However, changing the setting between:

  - Prev/Next
  - Next/Prev

  does not change the behavior of those buttons.

  After looking at the code, this appears to be because the setting only swaps the volume side buttons via PageBack / PageForward. The front Left / Right buttons are still mapped directly as:

  - Left = previous page
  - Right = next page

So the issue is not that the setting is entirely broken. It works for the actual volume side buttons, but it does not cover the buttons that become the right-edge controls in Landscape CCW.

### Steps to Reproduce

Settings: Controls: Side Button Layout (reader): Prev/Next or Next/Prev 

### Expected behavior:

  In Landscape CCW, there should be a way to invert the page-turn direction of the physical right-edge buttons used for reading.

###  Actual behavior:

  Side Button Layout (reader) only affects the volume side buttons, so the physical right-edge buttons in Landscape CCW keep the same page-turn order regardless of the setting.

### Patch
I don't have a full dev environment setup but the patch would look something like:

```diff
diff --git i/src/activities/reader/ReaderUtils.h w/src/activities/reader/ReaderUtils.h
index 1de6174..400ad89 100644
--- i/src/activities/reader/ReaderUtils.h
+++ w/src/activities/reader/ReaderUtils.h
@@ -41,16 +41,22 @@ inline PageTurnResult detectPageTurn(const MappedInputManager& input) {
   const bool usePress = SETTINGS.longPressButtonBehavior == SETTINGS.OFF;
   const bool tiltNext = SETTINGS.tiltPageTurn && halTiltSensor.wasTiltedForward();
   const bool tiltPrev = SETTINGS.tiltPageTurn && halTiltSensor.wasTiltedBack();
+  const bool swapLandscapeCcwFrontButtons = SETTINGS.orientation == CrossPointSettings::LANDSCAPE_CCW &&
+                                            SETTINGS.sideButtonLayout == CrossPointSettings::NEXT_PREV;
+  const auto frontPrevButton =
[...]

PR #2023, opened 2026-05-17
feat: Make page turn naturally follow orientation

jacksongoode:
## Summary

Addresses https://github.com/crosspoint-reader/crosspoint-reader/issues/1954 to allow the prev/next to follow the orientation of the device. That is button presses on Inverted & Landscape CCW will follow the proper directionality pressing right/downwards will move to the next page, while left/upwards will move to the previous page.

A simple logical abstraction that swaps the next/prev assignment based on orientation.

## Additional Context

Currently this is just implemented by default. ~I don't *think* it needs to be a toggle in settings but I could add this as well.~ A toggle setting to enable this behavior was added under `Side Button Layout` in `Controls`.

Tested on my device and appears to work!

---

### AI Usage


Uri-Tauber (dismissed):
Very neat.

Uri-Tauber:
@jacksongoode Please run clang-format.

Uri-Tauber (dismissed):

uxjulia:
@jacksongoode - My concern with not allowing this to be a setting is that today we allow people to map their side buttons to be either Next/Prev or Prev/Next, so this PR assumes Up = Prev and Down = Next, ignoring there are people who may think the opposite (problematic in "Inverted" mode). I think for the X4 specifically, in Landscape CCW the majority of people will accept that Left/Up = Back, Right/Down = Next. On the X3 though, this assumption breaks down because the position of the Next/Prev buttons on either side of the device turn them into Up/Down buttons when in landscape... which to different people can mean either Next/Prev or Prev/Next, just like with the X4's side buttons.

tdlr; Next/Prev is currently customizable but this PR assumes only one "correct" setting.

jacksongoode:
Thanks for the response. That makes sense, I suppose the hope was that a user's preference in directionality would be relatively preserved on orientation change. But it does change the current expectation for using the front facing buttons while the device in inverted or landscape CCW.

But now I see the concern around the X3... I was imagining that this only affects the front buttons of either device. I'm not sure what the X3's button set does (are each side's rockers redundant - do they both turn the page and the front buttons activate the menu?). If you have a suggestion I can implement in the X3 case.

uxjulia:
> Thanks for the response. That makes sense, I suppose the hope was that a user's preference in directionality would be relatively preserved on orientation change. But it does change the current expectation for using the front facing buttons while the device in inverted or landscape CCW.
> 
> But now I see the concern around the X3... I was imagining that this only affects the front buttons of either device. I'm not sure what the X3's button set does (are each side's rockers redundant - do they both turn the page and the front buttons activate the menu?). If you have a suggestion I can implement in the X3 case.

Even if it were only applied to the front buttons, Landscape CW people who have the mental model where Up = increment the page and down = decrement the page are overlooked in this PR.

I think a better implementation is to avoid assuming any directional mapping at all and just "swap" the buttons to do the opposite when the setting is on. This way, regardless of what a user's mental model is and what orientation they've set things at, they can turn it on/off in whichever orientation they want and the code just swaps the buttons.

And to answer your question about the buttons on the X3, they do similar things, except take the top rocker button the X4 and put it on the left side of the device. Image for reference:
X3 Buttons

jacksongoode:
Hmm, I see. In that case it makes sense to add a follow to have buttons "follow orientation", below the side button layout setting in "Controls"? I'll add this.

Edit: Please consider the i18n are all machine translations so proper review is needed.

uxjulia:
I tested this on my device and just a few notes:

1. The setting is called "Buttons match orientation" but it only applies to the front buttons. This should be made clearer somehow if this won't extend functionality to the side buttons.
2. When you open the settings from the reader, the buttons are still in their original orientation (i.e. when in Landscape CCW > in-reader options > Up/Down are in their original orientation. I think these should also match orientation.

Edit: 
3. When in "Inverted" orientation, the labels are still upside down. To make the buttons truly match orientation, these labels should flip as well.

jacksongoode:
@uxjulia Thanks for the comment, I'll make the label change. But I want to make sure the string is good, would something like "Orient front buttons" make more sense?

uxjulia:
I don't want to say that Crossink did it correctly or anything, but I chose "Orientation Aware" for my implementation there. Can go with that or something similar.

jacksongoode:
@uxjulia @Uri-Tauber Any further suggestions before merge?

Uri-Tauber (approved):

uxjulia (approved):

PR #2026, opened 2026-05-17
feat: auto KOSync on book open/close

### LINKED ISSUE #914: Feat: Dedicated options to push/pull sync
It would be great to have an option in the reading menu to "Push progress and Sleep" to be used at the end of a reading session.

togotago:
## Summary

* **What is the goal of this PR?** Automatic KOReader progress sync on book open and close events. Pushes reading position to the KOSync server when closing a book (or going to sleep), and optionally pulls the latest position when opening a book. Keeps progress in sync across devices without manual sync. Resolves #914.

**Three-state setting in KOSync Settings:**

| Mode | Behavior |
|---|---|
| **Off** | No auto sync (default, current behavior) |
| **On Close** | Pushes progress on book close/sleep only. **Recommended.** |
| **On Open + Close** | Full two-way sync. Adds ~2-4s delay on book open. |

**Implementation details:**

- Sync-on-open runs in ReaderActivity before loadEpub() where full ~115KB heap is available for TLS
- Sync-on-close precomputes payload while epub is alive, executes after epub.reset() frees heap
- Uses ProgressMapper + section LUT refinement for sub-page position accuracy (same path as manual sync)
- Direction-aware: only pushes when local is ahead, only pulls when remote is ahead
- WiFi credentials loaded at boot so sync works after wake-from-deep-sleep
- 7 skip rules: no creds, no saved WiFi, session <30s, cooldown (2min, independent for open/close), position unchanged, battery <15%, exponential backoff
- WiFi timeout 5s for fast failure when away from home
- UI feedback: "Syncing..." / "Synced" / "Sync: no WiFi" / "Sync failed"
- Debug log at /.crosspoint/autosync.log on SD card

**Files changed:**

- `lib/KOReaderSync/AutoKOSync.h` + `.cpp` (new, ~500 lines) - self-contained sync library
- `src/main.cpp` - load WiFi credentials at boot
- `src/activities/reader/ReaderActivity.cpp/.h` - sync-on-open before loadEpub
- `src/activities/reader/EpubReaderActivity.cpp/.h` - apply remote on open, close payload + sync
- `src/activities/settings/KOReaderSettingsActivity.cpp` - 6th menu item
- `src/CrossPointSettings.h` - autoKOSync field
- `src/SettingsList.h` - web settings entry
- `lib/KOReaderSync/KOReaderSyncClient.cpp` - HTTP timeout 15s to 5s
- `lib/I18n/translations/english.yaml` - 8 new i18n strings

## Additional Context

- **Battery impact:** Estimated well under 1% per day based on ESP32-C3 datasheet current draw (~130mA WiFi TX, ~2-4s per sync cycle) and skip rules. Not empirically measured on device.
- **HTTP timeout change:** This reduces the HTTP timeout in KOReaderSyncClient.cpp from 15s to 5s. KOSync payloads are small JSON and all successful requests in testing completed under 4s. The 15s timeout was causing 30-40s delays before sleep on marginal networks. Happy to revert if preferred.
- **Testing:** Developed and tested on CrossInk/X4 hardware over multiple days, syncing between an X4 and a Kindle running KOReader via sync.koreader.rocks. Ported to CrossPoint and verified clean build against master. Also submitted as uxjulia/CrossInk#162.

---

### AI Usage


itsthisjustin:
Closing this in favor of #1090 which doesn't run a sync concurrently with other ui operations

togotago:
@itsthisjustin I see the other PR is marked as stale. Open to suggestions on how this implementation could be improved. Have been running this on a patched version of crossink the last days and it has been very stable when looking at the logs, even when using custom SD fonts.

itsthisjustin:
It's pretty much impossible for any PR here to not go stale haha. My concern with this approach is that it's doesn't give the user a ton of control and could lead to memory issues eventually. I'd prefer something that runs less intrusively but my preference is not what should be a deciding factor here

PR #2032, opened 2026-05-17
feat: port crossink 'read book move' feature to crosspoint

mcrosson:
## Summary

Adds an option under `System` in the settings to move read books to a `/read` folder. This will fire when the 'End of book' pops up on screen. Defaults to off.

## Additional Context

This is a minimal port of the feature and intentionally avoids the manual setting of a book as read from the reader menu. The reading stats present in Crossink collide with this feature lightly and I felt it best to keep the port to a bare minimum.

---

### AI Usage

Did you use AI tools to help write this code? *YES, Claude*


itsthisjustin:
Should we go ahead and add the string translations for the other languages?

uxjulia:
I think the only small thing is that in this implementation, when the book is moved to the "Read" folder, it's removed from the "Recent Books" list. Crossink updates the path by calling `RECENT_BOOKS.updatePath(...)` in `EpubReaderActivity.ccp` to the new relocated path so that the book isn't removed from the recent books list. Other than that this looks good to me.

mcrosson:
> I think the only small thing is that in this implementation, when the book is moved to the "Read" folder, it's removed from the "Recent Books" list. Crossink updates the path by calling `RECENT_BOOKS.updatePath(...)` in `EpubReaderActivity.ccp` to the new relocated path so that the book isn't removed from the recent books list. Other than that this looks good to me.

This was intentional. When I've finished reading a book and its been moved to the 'read' folder, I want it off my recents list.

I could make this part tunable if there is enough 'could go either way' here. If we had a way to remove books from the recents list I'd be more inclined to keep the crossink behavior but I'm not sure we want that additional work in this PR. I am open to it though.

LMK how y'all would like to proceed.

uxjulia:
I'd be more inclined to keep existing behavior where users see all their recent books and open another PR to allow removing recent books from the list. Otherwise, this PR needs to change the "Recent Books" label to "Recent Unread Books" to make it clear that the "Recent Books" no longer actually lists all recent books. Silently removing books from the "Recent Books" list will make people think it's a bug.

itsthisjustin:
> I'd be more inclined to keep existing behavior where users see all their recent books and open another PR to allow removing recent books from the list. Otherwise, this PR needs to change the "Recent Books" label to "Recent Unread Books" to make it clear that the "Recent Books" no longer actually lists all recent books. Silently removing books from the "Recent Books" list will make people think it's a bug.

Unfortunately I agree. I'd also want it gone from the list but it will look like a bug

mcrosson:
Ill make it consistent with crossink, and cook 2 new prs: one to allow 'hold confirm to remove from recent list' (with confirm prompt) and another to add a setting that removes read books from the recent list.

mcrosson:
Changes made & pushed.

itsthisjustin:
Can you have Claude update all the strings for the other languages?

uxjulia (approved):

mcrosson:
> Can you have Claude update all the strings for the other languages?

Done

itsthisjustin (approved):

PR #2033, opened 2026-05-18
fix: wire through silent restart clear resume state with sdk

jeremydk:
## Summary

After a silent reboot, there was a small window where the esp32 would listen for button presses but the full refresh would hold the event loop. This gave a UX experience where the silent reboot had completed to the home screen, a user taps select (at any time during the process), and they find themselves unexpectedly in a book.

## Additional Context

This must land after https://github.com/crosspoint-reader/community-sdk/pull/11 and will need the submodule SHA changes included in. 

---

### AI Usage

Did you use AI tools to help write this code? partially

### DIFF
--- a/lib/hal/HalDisplay.cpp
+++ b/lib/hal/HalDisplay.cpp
@@ -18,14 +18,18 @@ void HalDisplay::begin(bool seamless) {
 
   einkDisplay.begin();
 
+  if (seamless) {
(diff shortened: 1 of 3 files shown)


jeremydk:
Validated on x3 and x4 with fading fix off and on, _and_ with the new page as sleep screen off and on.

itsthisjustin (approved):

mcrosson (approved):

PR #2037, opened 2026-05-18
feat: serial test harness for on-device validation

jeremydk:
Lets a host Python script drive a connected X3/X4 over USB CDC: inject button events, observe activity transitions, and assert on screen state without touching the hardware. Pytest-runnable so device behaviour can join the rest of the development workflow.

Firmware side (all gated behind ENABLE_SERIAL_LOG so release/slim builds are unaffected):

- HalGPIO::simulateButton(idx, down) overlays a virtual button-down on top of the real InputManager state. update() OR's physical and virtual, so wasPressed/wasReleased edges fire for every activity exactly as if the user had pressed the button. simulateButtonTap(idx, ms) schedules an auto-release for one-shot taps without paying serial RTT on the release edge.

- main.cpp CMD parser extended:
    CMD:BTN PRESS|RELEASE|TAP|HOLD <name>   button injection
    CMD:STATE                               re-emit current activity
    CMD:HOME                                activityManager.goHome()
  CMD:SCREENSHOT is unchanged.

- Activity::onEnter emits '[STATE] activity=<name>'. Canonical activity- transition signal the host syncs on.

- ActivityManager render task emits '[REFRESH] done' after every frame push. Lets tests wait for the screen to settle rather than guessing with sleeps.

- HomeActivity::render emits '[HOME] menu=<symbols> selector=<n>' with stable, locale-independent symbols (RecentBook<i>, FileBrowser, Recents, OpdsBrowser, FileTransfer, Settings). Adapts to per-device config without hard-coded indices.

Host side:

- scripts/devharness/Device opens the USB CDC port once and runs a reader thread that captures every line into a bounded deque. Forces DTR/RTS off on open; the macOS default pulses the C3 USB peripheral and triggers USB_UART_CHIP_RESET.

  Methods: press, release, tap, hold, request_state, go_home, home_menu, navigate_home_to, wait_for, wait_for_activity, wait_for_refresh, current_activity. _write snapshots a send cursor before writing so wait_for can scan lines emitted in response to the send even if the firmware beats the host back into the Condition.

- test/device/ pytest fixtures (session-scoped Device, per-test home that resets via go_home), plus test_smoke (primitives round-trip) and test_navigation (each Home menu destination drives to its matching activity).

- scripts/requirements.txt += pytest.

Run: uv run pytest test/device. 12 passing live on X3 in ~33s.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,5 @@ build
 lib/EpdFont/scripts/downloaded_fonts/
 lib/EpdFont/scripts/instanced_fonts/
 lib/EpdFont/scripts/output/
+# Local serial harness log capture; written by scripts/devharness when debugging.
+harness.log
--- a/lib/hal/HalGPIO.cpp
+++ b/lib/hal/HalGPIO.cpp
@@ -207,10 +207,68 @@ void HalGPIO::update() {
   const bool connected = isUsbConnected();
   usbStateChanged = (connected != lastUsbConnected);
   lastUsbConnected = connected;
+
+#ifdef ENABLE_SERIAL_LOG
+  const uint32_t now = millis();
+  for (uint8_t i = 0; i < 7; i++) {
+    // Consume any scheduled auto-release whose deadline has passed.
+    if (_simulatedReleaseAt[i] != 0 && now >= _simulatedReleaseAt[i]) {
+      _simulatedDown[i] = false;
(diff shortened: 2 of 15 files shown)


jeremydk:
sat without review for 2 weeks, even though this has been pivotal into every single investigation i've been asked to do.

PR #2039, opened 2026-05-18
feat(serial): Heapdump investigation tools via serial test harness

jeremydk:
## Summary

Leverage the serial test harness and add live heapdumping capabilities.  The combination of these two allows us to repeat through a series of known actions, sample the heap, and troubleshoot regions.

## Additional Context

DRAFT, rebase after the serial harness if accepted, and then move to ready for review. Depends on https://github.com/crosspoint-reader/crosspoint-reader/pull/2037

---

### AI Usage


jeremydk:
see #2037

PR #2053, opened 2026-05-18
fix: harden web file operations

sabraman:
## Summary

This hardens web-facing file operations so failed transfers do not destroy existing files and protected paths stay hidden from WebDAV/browser access.

## Changes

- Normalize trailing `.` and `..` path components correctly.
- Apply normalized/protected path checks across web download, upload, create, rename, move, delete, and WebSocket upload paths.
- Write WebDAV `PUT`, `COPY`, and `MOVE` replacements through temporary files with destination backup/restore handling.
- Write browser uploads, WebSocket uploads, and font uploads through temporary files before replacing existing targets.
- Preserve existing OPDS/HTTP download targets until the download body and size checks complete.
- Block WebDAV `PROPFIND` access to protected paths and log rollback failures that leave backups recoverable.

## Why

Several file-transfer paths removed or overwrote the destination before the replacement data was known-good. If a client disconnected, the SD card write failed, validation failed, or the final rename failed, an existing book/font/download could be lost or hidden behind a temporary filename.

## Validation

- `./bin/clang-format-fix`
- `git diff --check`
- `pio run`
- `pio check --fail-on-defect low --fail-on-defect medium --fail-on-defect high`
- `coderabbit review --agent -t committed --base origin/master` returned 0 findings before the final WebDAV safety fix
- `coderabbit review --agent -t uncommitted` returned 0 findings for the final WebDAV safety fix


sabraman:
Extracted the duplicated temp-path, directory check, backup/commit, and rollback logging helpers into FsHelpers, then updated the web upload, WebSocket upload, font upload, WebDAV, and HTTP download paths to use the shared helpers. The HTTP paths keep their previous decimal timestamp suffixes; web/WebDAV paths keep their previous hex suffixes.

sabraman:
Addressed the latest review gaps: /api/files now normalizes and rejects protected paths before listing, hidden/protected name checks are case-insensitive, valid names containing '..' are allowed while exact '.'/'..' remain rejected, WebDAV COPY fails on read errors before promotion, and backup cleanup failures are logged through the shared FsHelpers helper used by both commitTempFile and WebDAV replacement paths.

jeremydk:
Two things to fix before shipping (and handling merges):

1. commitTempFile should remove its own temp on failure. Two return false paths leave the temp on disk; the four current callers all Storage.remove(tempPath) after, but the contract isn't in FsHelpers.h. The next caller who skips that line leaks a temp across reboots. Clean up inside the helper (it owns the path) or document it.

2. Backups shouldn't land at hidden paths. makeTempPath puts both the upload temp and the backup-of-original at .<prefix>-…, which the file browser and PROPFIND hide. If commitTempFile's target rename fails and the restore rename also fails (the LOG_ERR branch), the original is at a path the user can't see. Use a visible prefix for the backup (e.g. <target>.bak-<ts>); temp can stay hidden.

Nit: the two scripts/jpegdec_patches/*.patch edits strip the --  mail signature git format-patch emits. Revert those, or split a separate PR that exempts *.patch from the formatter.

itsthisjustin:
@sabraman can you solve conflicts here?

sabraman:
All conflicts have been resolved and the PR branch has been updated with the latest develop. Verified locally with `pio run -e default`.

PR #2062, opened 2026-05-19
fix: validate OPF cover items as images

lpla:
## Summary
- only accept meta name="cover" manifest targets when their media type is image/*
- prevent XHTML wrapper entries from blocking EPUB3 cover-image / guide fallback
- align guide fallback image selection with current converters by limiting candidates to JPG/PNG

## Why
Some EPUBs map meta name="cover" to an XHTML wrapper page instead of the actual image, which caused Crosspoint to select an unsupported cover target and fail to generate cover.bmp and thumbnails.

## Testing
- Code has been tested on X4, fixing some books from Apple Books I exported.

---

### AI Usage


lpla:
PR tested with Apple Books .epub folders compressed back to .ePub files work now.

lib/Epub/Epub.cpp
```diff
@@ -105,7 +105,7 @@ bool Epub::parseContentOpf(BookMetadataCache::BookMetadata& bookMetadata) {
           if (endPos != std::string::npos) {
             const auto ref = std::string_view{coverPageHtml}.substr(pos, endPos - pos);
             // Check if it's an image file
-            if (FsHelpers::hasPngExtension(ref) || FsHelpers::hasJpgExtension(ref) || FsHelpers::hasGifExtension(ref)) {
+            if (FsHelpers::hasPngExtension(ref) || FsHelpers::hasJpgExtension(ref)) {
```
Uri-Tauber, line 109: Why did you removed the `.gif` check?

lib/Epub/Epub/parsers/ContentOpfParser.cpp
```diff
@@ -10,7 +10,27 @@
...
+  constexpr size_t prefixLen = sizeof(MEDIA_TYPE_IMAGE_PREFIX) - 1;
+  if (mediaType.size() < prefixLen) {
+    return false;
+  }
+
+  for (size_t i = 0; i < prefixLen; ++i) {
+    char c = mediaType[i];
+    if (c >= 'A' && c <= 'Z') {
+      c = static_cast<char>(c - 'A' + 'a');
+    }
```
Uri-Tauber: Probably better to use `std::tolower`

lib/Epub/Epub.cpp
lpla: Good question. GIF is intentionally excluded in this fallback because the downstream cover outputs only have JPG/PNG converter paths today: `generateCoverBmp()` handles JPG/PNG and rejects unsupported formats, and `generateThumbBmp()` does the same. If the guide XHTML contains an earlier `.gif`, accepting it can set `coverItemHref` to a format we cannot convert and prevent a later supported JPG/PNG reference from being selected. I pushed `3ef72f8` with an inline comment at the filter point so that rationale is visible in the code.

lib/Epub/Epub/parsers/ContentOpfParser.cpp
lpla: Agreed, thanks. I pushed `3ef72f8` changing this helper to `std::tolower(static_cast<unsigned char>(...))`, so the `image/` media-type prefix check stays case-insensitive without the hand-rolled ASCII branch.

Uri-Tauber (approved):
LGTM

itsthisjustin (approved):

PR #2075, opened 2026-05-20
refactor: move HttpDownloader onto esp_http_client

jeremydk:
Based on the learnings from https://github.com/crosspoint-reader/crosspoint-reader/pull/2074 , I wanted to bring the same buffer savings to the rest of our HTTP Client stack.  That being said, HttpDownloader (fonts/OPDS) used the Arduino HTTPClient.

HttpDownloader was the last consumer of the Arduino HTTPClient + NetworkClientSecure stack. OtaUpdater already runs on esp_http_client, so this drops the parallel HTTP/TLS implementation. It also fixes a class of OPDS/font download failures: HTTPClient's setTimeout is uint16 and truncates, and its short per-read deadline killed slow or chunked responses (the -11 / incomplete-data errors).

What changed:

- Rewrote fetchUrl/downloadToFile around esp_http_client with a streaming open() -> fetch_headers() -> read() loop, manual redirect following, and is_complete_data_received() as the completeness gate. Body bytes go straight to the sink (OPDS parser stream, std::string, or file), so nothing buffers the payload.

- HTTPS is now verified against the CA bundle instead of NetworkClientSecure::setInsecure(). esp-tls is built with CONFIG_ESP_TLS_INSECURE off, so an unverified handshake can't be set up anyway; the model is public servers over verified https and local servers over plain http (transport is chosen from the URL scheme). 

** Self-signed https servers are no longer supported, by design. **

- timeout_ms is 60s; esp_http_client's timeout is uint32, so unlike HTTPClient it doesn't silently truncate.

- HTTP buffers are 4096 (rx) / 1024 (tx). 4096 holds real OPDS server headers; the GitHub release CDN sends more and logs a non-fatal truncation warning, but the headers we read (Location, Content-Length) come first and survive.

- Removed the now-unused UrlUtils::isHttpsUrl and a stale HTTPClient comment in FontDownloadActivity.

Validated on device: OPDS browse and a 3.4 MB book download over verified https, GitHub font downloads (crc-checked), redirect handling matching curl, and slow/erroring servers surfaced correctly.

### DIFF
--- a/src/activities/settings/FontDownloadActivity.h
+++ b/src/activities/settings/FontDownloadActivity.h
@@ -36,9 +36,9 @@ class FontDownloadActivity : public Activity {
   void render(RenderLock&&) override;
   bool preventAutoSleep() override {
     return state_ == LOADING_MANIFEST || state_ == DOWNLOADING ||
-           // This is added because HTTPClient is a synchronous/blocking function,
-           // and blocks the main loop until the download is complete.
-           // So `activityManager.preventAutoSleep()` is never called during downloading
+           // The download is synchronous and blocks the main loop until it
+           // completes, so activityManager.preventAutoSleep() is never polled
+           // during downloading.
            state_ == COMPLETE || state_ == ERROR;
   }
   bool skipLoopDelay() override { return true; }
--- a/src/network/HttpDownloader.cpp
+++ b/src/network/HttpDownloader.cpp
@@ -1,204 +1,201 @@
 #include "HttpDownloader.h"
 
-#include <HTTPClient.h>
+#include <Arduino.h>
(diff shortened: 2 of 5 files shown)


jeremydk:
Tested against https OPDS with no auth. I don't have a local opds set up -- if anyone running calibre/etc wants to give this a shot, I'd appreciate it.

WuTofu:
> I don't have a local opds set up -- if anyone running calibre/etc wants to give this a shot, I'd appreciate it.

I can connect to my opds server via http with auth and download books with no issues👍

znelson (approved):

itsthisjustin (approved):

PR #2082, opened 2026-05-20
feat: Compress BW buffer to avoid AA fallback (save ~0.5s in page display time)

leecming82:
## Summary
RLE-compress the BW buffer to prevent OOM issues forcing the AA code from using the BW buffer fallback (which adds ~0.5 seconds in page display time) 

## Additional Context
When AA is turned on, page renders start with the BW pass, saving it to a buffer, running the grayscale AA passes, then restoring the BW buffer. With SD fonts, it's possible that "big" fonts combined with languages like CJK with many distinct glyphs will lead to the per-page glyph cache crowding out the BW buffer from memory forcing the AA code to use a slower fallback approach.

To prevent that, RLE-compress the BW buffer so less memory is needed. The compression/decompression adds <10ms of CPU time but saves potentially 0.5s in avoiding the AA pass doing a more complicated BW restore path. 

## Technical Details
`GfxRenderer::storeBwBuffer()` now tries a bounded RLE-compressed backup before falling back to the previous raw chunk backup:

1. Render and display the BW page.
2. Try to compress the BW framebuffer into a fixed 32KB RLE buffer.
3. If compression succeeds, preserve that backup through the AA passes.
4. Restore by decompressing into `frameBuffer`, then run the same grayscale cleanup path.
5. If compression allocation or size cap fails, try the previous raw chunk backup.
6. If raw backup also fails, keep the existing BW re-render fallback.

---

### AI Usage

### DIFF
--- a/lib/GfxRenderer/GfxRenderer.cpp
+++ b/lib/GfxRenderer/GfxRenderer.cpp
@@ -13,6 +13,7 @@
 namespace {
 
 const char* resolveVisualText(const char* text, std::string& visualBuffer, int paragraphLevel);
+constexpr uint32_t COMPRESSED_BW_BACKUP_CAP = 32 * 1024;
 
 /**
  * Resolves the requested style to the best available style in the given SD card font.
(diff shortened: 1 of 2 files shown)


Uri-Tauber:
> To prevent that, RLE-compress the BW buffer so less memory is needed. The compression/decompression adds <10ms of CPU time but saves **potentially** 0.5s in avoiding the AA pass doing a more complicated BW restore path.

Can you provide exact measurements?

leecming82:
This is of my heavily modified branch that supports Japanese but the delta should still apply

| Metric | X3 Pre | X3 Post |
|---|---:|---:|
| AA fallback rows | 5 | 0 |
| Compressed backup rows | 0 | 24 |
| BW framebuffer bytes | 52,272 | 52,272 |
| Compressed BW median bytes | n/a | 22,515 |
| Compressed BW max bytes | n/a | 24,629 |
| BW store | 3 ms | 10 ms |
| BW re-render fallback | 461 ms fallback-only | 0 ms |
| Gray cleanup after fallback | 113 ms fallback-only | 0 ms |
| Total page render | 2224 ms fallback-only / 1716 ms all-normal | 1705 ms |

itsthisjustin:
I'm doing a bunch of work to consolidate x3 luts/waveforms into the same system that x4 uses so lets hold off on this for now

PR #2085, opened 2026-05-20
chore: Replace JPEGDEC patch script with in-org fork

znelson:
## Summary

Instead of applying patches to JPEGDEC code after pulling locally, pull from an [in-org fork of JPEGDEC](https://github.com/crosspoint-reader/JPEGDEC) with our patches applied in a branch.

---

### AI Usage


Uri-Tauber:
I don't really mind either way. I do think we obliged to contribute bug-fixes upstream, so having fork would be good idea.

Uri-Tauber:
@znelson You created the PR branches in the main repo - not your fork.

znelson:
> You created the PR branches in the main repo - not your fork.

@Uri-Tauber Yeah, intentional. My fork is pretty divergent at this point. But if you (or others) prefer I can definitely keep an upstream branch in my fork and make PR branches there.

znelson:
> I do think we obliged to contribute bug-fixes upstream, so having fork would be good idea.

Yeah, I think a fork helps in two aspects:
- Easier to contribute changes back upstream
- Easier to see what exactly the code we're using looks like, on-the-fly patches are kind of haphazard

Uri-Tauber (approved):
Approving, but I would like to get @jeremydk input here

Uri-Tauber:
> @Uri-Tauber Yeah, intentional. My fork is pretty divergent at this point. But if you (or others) prefer I can definitely keep an upstream branch in my fork and make PR branches there.

I think it would be better - keeping the main repo clean.

znelson:
> I think it would be better - keeping the main repo clean.

Noted, will do. ✅

jeremydk:
Frankly i'm cool either way here, I just _really dislike_ the pattern of a buildscript using an inline edit against a global package store.  Moving to the git format-patch strategy is a way to solve that, as is a fork.  Happy with either.

Issue #1966, opened 2026-05-13
"Short Power Button Click" set to "Sleep" causes device to wake with short button press


Joseph-DiGiovanni:
### Affected Version

1.2.0-dev-master-71518b3

### Describe the Bug

As the title says, when the short power button click action is set to "sleep" the device will wake with a single short button press rather than needing to be held a half second.

### Steps to Reproduce

1. Go to Settings > Controls
2. Set "Short Power Button Click" to "Sleep"
3. Put device to sleep
4. Quickly press the power button once

### Expected Behavior

Power button should need to be held for about half a second to wake the device.

### Relevant Log Output/Screenshots

```shell

```

uxjulia:
I'm pretty sure this is intended behavior.

Joseph-DiGiovanni:
I highly doubt it, or it would be called something like "Power". The user guide only says this: "Sleep- A short press puts the device into sleep mode". Nothing about waking. Even if it were intentional, it is extremely unintuitive.

Took me quite a while to realize why my device was so sensitive to accidental wakes in my pocket. I only enabled the setting thinking I could more quickly put the device to sleep as I put it away.

If a quick (or perhaps longer) wake were needed, I would suggest a separate option to set the power button wake duration independently.

uxjulia:
I do think it's intentional.

See this issue that states short-press to wake isn't working: https://github.com/crosspoint-reader/crosspoint-reader/issues/288
And the fix to make it work: https://github.com/crosspoint-reader/crosspoint-reader/pull/482

Maybe a PR to change the label to "Sleep/Wake" or "Power" as you suggested would make it more obvious

Joseph-DiGiovanni:
I don't think referencing an issue/fix for a completely different option that no longer exists is enough to say. Both the current name, guide, and setting this is an option under seem to imply it remaps the power button while the device is on (Can't page turn or refresh the screen while asleep for example). If a short power button wake is needed I'd recommend a separate PR for an independent wake duration.

PR #2104, opened 2026-05-21
feat: Independent power button sleep/wake duration

Joseph-DiGiovanni:
The existing setting for mapping the short power button press to the sleep function did not make it clear it also wakes the device with a short press. I added independent control of the power button wake and sleep hold duration to correct this, as well as add an even longer wake hold duration for those finding their device turning on unexpectedly in bags or pockets.

Addresses #1966 but should also be a good solution for people requesting a lock screen due to accidental wakes.

---


itsthisjustin:
@jeremydk can you take a look at this and either approve or close?

Joseph-DiGiovanni:
Sorry, botched the first rebase attempt and open-x4-sdk changes got thrown in there. Fixed it.

jeremydk:
Looks good in shape. Happy to ship once a couple of things are tied up.

1. Don't silently regress shortPwrBtn = SLEEP users. Today that setting gives them tap-to-sleep and tap-to-wake; after this PR they keep tap-to-sleep but their wake jumps to a 400 ms hold because the new powerButtonWakeDuration defaults to NORMAL. Either migrate at settings load time (if shortPwrBtn == SLEEP and the new field is at its default, promote it to WAKE_SHORT) or have getPowerButtonWakeDuration() fall back to the old shortPwrBtn derived value when the user hasn't explicitly set the new one.

2. Delete verifyPowerButtonDuration() in src/main.cpp. It has no callers (HalGPIO::verifyPowerButtonWakeup is the live path after this I believe), the PR currently edits it in lockstep with the live call, which is the kind of thing that quietly drifts on the next refactor.

Thanks!

Joseph-DiGiovanni:
> Delete verifyPowerButtonDuration() in src/main.cpp

Good catch. Done.

> Don't silently regress shortPwrBtn = SLEEP users

I disagree its a regression. The original short wake behavior was not documented or even implied by the option name itself. I didn't even correlate the short wake to the setting for a while, hence my bug report for it (#1966). Frankly the result is easy to notice if you're expecting a short wake and the new setting is easy to find.

ruffson:
> > Delete verifyPowerButtonDuration() in src/main.cpp
> 
> Good catch. Done.
> 
> > Don't silently regress shortPwrBtn = SLEEP users
> 
> I disagree its a regression. The original short wake behavior was not documented or even implied by the option name itself. I didn't even correlate the short wake to the setting for a while, hence my bug report for it (#1966). Frankly the result is easy to notice if you're expecting a short wake and the new setting is easy to find.

Lets not get hung up on how to migrate the current settings, I agree that the current state is confusing but it is also best to not change that behavior for all users after the update.

This is a very much desired feature. Please merge it :pray: .

itsthisjustin:
@Joseph-DiGiovanni can you solve conflicts after rebasing to develop?

PR #2112, opened 2026-05-22
feat: add Norwegian hyphenation (Bokmål and Nynorsk)

tomlarse:
## Summary

* **What is the goal of this PR?** Add Norwegian hyphenation support, covering
  both Bokmål and Nynorsk.
  * Add Norwegian trie via `scripts/update_hyphenation.sh` (added `process no`),
    generated from hypher's `no.bin`.
  * Register a single `norwegianHyphenator` in `LanguageRegistry.cpp`.
  * Normalize the `dc:language` tags `nb`, `nn`, `nob`, `nno`, and `nor` to `no`
    in `Hyphenator.cpp` so every Norwegian variant resolves to the one trie.

## Additional Context

* **Why one trie for both standards?** typst/hypher ships only a single `no`
  trie — there is no separate `nb`/`nn`. It is built from `hyph-no.tex`, which
  is the combined Norwegian pattern set (~27k patterns) covering both Bokmål and
  Nynorsk, so a single trie hyphenates both correctly.
* **Memory:** the trie is ~156 KB `constexpr` in flash (2nd-largest after
  German). Zero resting DRAM cost — patterns decode into DRAM only while a
  Norwegian book is open.
* **Tests:** no eval test data added, following the Finnish PR (#2084) precedent.
  The eval harness needs a corpus-derived test file; pyphen does support `nb`/`nn`,
  so a test set can be generated from a Norwegian EPUB later if desired.
* **Verification:** builds clean with `-Wall -Wextra -pedantic`; existing language
  eval scores unchanged. Confirmed all six tags (`no`/`nb`/`nn`/`nob`/`nno`/`nor`)
  produce identical correct splits (`bok=måls=ord=boka`, `grunn=lovs=dag`,
  `ny=norsk=bru=kar`), while an English control mangles the same word.

---

### AI Usage


tomlarse:
Addressed the review feedback in b88aad6, along with the clang-format CI failure (the new alias array needed reformatting).

1. **Rename `Iso639Mapping`** — Done. Renamed the struct to `LanguageTagAlias` and the array to `kLanguageTagAliases` (field `iso639_1` → `canonicalTag`), and updated the doc comment to reflect that it normalizes general language-tag aliases (including the BCP-47/639-1 Norwegian tags `nb`/`nn`), not just ISO 639-2.

2. **`wget --fail` in `update_hyphenation.sh`** — The bare `wget` predates this PR and applies to all nine existing languages, so I've left it out of scope here to keep the change focused on adding Norwegian. Happy to do it as a separate PR if maintainers would like.

3. **`std::array` size deduction via CTAD / `std::to_array`** — The manually-sized `std::array` is the existing pattern across the whole registry rather than something this PR introduces, so it seems better left as a separate cleanup that can be applied consistently.

takatsa-ship-it:
It would be nice for me too if Hungarian ebook hyphenation could be solved. But I don't think the solution is to build the hyphenation dictionary of smaller languages ​​into the firmware, as there is not enough space for them. It would be much easier to expand the "Optimize Epub" option of the Crosspoint Upload menu with automatic soft hyphenation of the epub you want to upload. After all, Crosspoint recognizes soft hyphens and uses them correctly.
Instead of explaining, check out this little app that automatically performs soft hyphenation for the selected language. Something like this should be built into the Upload menu, it would be a great help to many of us.
https://epub-for-crosspoint--maartenvan5.replit.app/

itsthisjustin:
I agree with @takatsa-ship-it and we need to rethink hyphenation across the whole firmware. I'm going to close this for now in favor of eventually taking this work (hopefully @tomlarse you'll be here to help!) and integrating it into a better long term solution

takatsa-ship-it:
Hi Justin,
It's great that you're addressing this issue. Many users will be very happy
with this option. I'm just afraid to mention that soft hyphenation could be
a general solution for all languages (including English, French, German,
etc.) and this method would free up a lot of space in the firmware that
could be used for other things.
Thanks again for addressing this request.
Best regards,
Augustine

Justin Mitchell ***@***.***> ezt írta (időpont: 2026. máj.
25., H, 22:40):

PR #2113, opened 2026-05-22
feat(i18n): add Norwegian Bokmål translation

tomlarse:
## Summary

* **What is the goal of this PR?** Add a complete Norwegian Bokmål (`nb`) UI
  translation.
  * New `lib/I18n/translations/norwegian.yaml` with all 381 `STR_*` keys
    translated (`_language_name: "Norsk bokmål"`, `_language_code: "NB"`,
    `_order: "22"`).

## Additional Context

* **No code changes needed.** The language picker is data-driven — the generator
  builds the `Language` enum and string tables from the YAML at build time, so a
  new translation file is all that's required. Verified `python scripts/gen_i18n.py`
  reports 381/381 keys with 0 missing.
* **Fonts:** Norwegian's `æ ø å` are already covered by the bundled fonts (used
  by Danish and Swedish), so no font work is required.
* **Translation notes (Bokmål-specific choices):**
  * "Read folder" is left untranslated (`Read-mappen`) because it refers to the
    fixed `/read` directory on the SD card, consistent with all other locales.
  * On-screen keyboard key captions (`[***]`, `[abc]`, `ABC`, `URL`, `DEL`) are
    kept literal so they match the rendered keys.
  * Format specifiers (`%zu`, `%d`, `%.2f%%`, `%s`) and `«»` arrows preserved.
* This is the UI translation only; Norwegian hyphenation is a separate PR (#2112).

---

### AI Usage


tomlarse:
Both flagged items check out as non-issues for this file:

1. **Root keys indented** — They aren't; all `_`/`STR_` keys are at column 0 in `norwegian.yaml` (matching the other locale files). There's no leading indentation to remove.

2. **`_order` as a quoted string** — This matches the existing convention in all 22 other translation files (`_order: "N"`), and the generator sorts numerically rather than lexicographically — it casts with `int()` in `scripts/gen_i18n.py` (`key=lambda item: int(item[0])`). So the `"10" < "2"` concern doesn't apply here. Switching only this file to an unquoted integer would diverge from the established format without changing behaviour.

itsthisjustin (approved):

itsthisjustin:
@tomlarse I'm not sure why the tests are failing without digging in but can you check those?

tomlarse:
Fixed. valencian.yaml was merged after I opened this PR and took `_order: "22"`, so the generator failed on the duplicate. Merged master and bumped Norwegian to `26`. I also re-synced the file against current english.yaml since quite a few keys changed in the meantime: translated the 16 new ones (bookmarks, sleep timer etc.) and removed the ones that are gone from English. CI should be green now.

Uri-Tauber (approved):

Issue #1706, opened 2026-04-19
Horizontal rules not rendered


STrRedWolf:
### Affected Version

v1.2.0

### Describe the Bug

ePubs that use a horizontal rule element have said element not render in CrossPoint.  It's just missing.

### Steps to Reproduce

1. Download this file: [ePubTest v2.zip](https://github.com/user-attachments/files/26872989/ePubTest.v2.zip)
2. Rename it to "ePubTest v2.epub"
3. Upload it to CrossPoint
4. View the file.

### Expected Behavior

Horizontal rule between first to paragraphs shows.

<img width="417" height="433" alt="Image" src="https://github.com/user-attachments/assets/3ef278ee-b55f-4a58-922b-117963a676cc" />

### Relevant Log Output/Screenshots

```shell

```

STrRedWolf:
Note, there is a font related bug here but that is already being worked on.  ePub file is submitted for testing against CrossPoint.

WuTofu:
```html
<hr style="margin-left:0%; margin-right:0%; width:100%; border-color:#000000; background-color:#000000; border-width:0.74pt 0 0 0; border-style:solid;"/>
```

I recalled that I have seem @uxjulia added [`<hr>` support](https://www.reddit.com/r/XTEINK/comments/1t5906f/comment/ok8m1a4/) to crossink.

Hey @uxjulia, would you be open to opening a pr? Or I can open a pr, of course, the credit goes to you.

Uri-Tauber:
Solved in #2117.

PR #2117, opened 2026-05-22
feat: ports hr tag rendering from crossink

uxjulia:
## **What is the goal of this PR?**  
  Add EPUB `<hr>` tag support so horizontal rules render as visible separators instead of being ignored.

## **What changes are included?**
  * Adds a `PageHorizontalRule` page element with serialization, deserialization, and rendering support.
  * Updates the EPUB HTML parser to emit a centered horizontal separator when it sees `<hr>`.
  * Bumps the `section.bin` cache version so existing cached EPUB pages are rebuilt.
  * Updates `docs/file-formats.md` for the new serialized page element.

## Additional Context

* The rule is stored as a lightweight page element: x/y position, width, and thickness.
* Existing EPUB cache data is invalidated by the section format version bump, which avoids trying to read old cached pages with the new element format.
* The version in `file-formats.md` was out of date (8). This updates it to the proper `24` which is what is the next version bump in `Section.cpp`

---
Example of how the `<hr>` tag renders

hr example

---

### AI Usage

### DIFF
--- a/docs/file-formats.md
+++ b/docs/file-formats.md
@@ -61,29 +61,29 @@ struct TocEntry {
 struct BookBin {
     // Header
     u8 version [[comment("Format version"), color("FFD93D")]];
-    
+
     // Version validation
     if (version != EXPECTED_VERSION) {
         std::error(std::format("Unsupported version: {} (expected {})", version, EXPECTED_VERSION));
     }
-    
+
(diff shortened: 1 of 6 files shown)


mcrosson (approved):
LGTM

itsthisjustin (approved):

STrRedWolf:
This should fix #1706 .

Eloren1:
Beautiful

PR #2118, opened 2026-05-23
feat: add Explore Mode for navigating without losing position

ZephByte:

To allow users to quickly navigate back to where they left off from, before they navigated to a chapter or percentage. Helpful when quickly looking at a map, glossary, etc in a different part of the book without losing your spot. Similar to Kindle's "Back to page..." and "Stay Here" feature.

- **Settings → Reader**: New "Explore Mode" toggle.
- **Reader menu**: When active and a return point exists, "Back to [Chapter]" and "Stay Here" appear at the top of the menu.
- **Status bar**: Inverted and prefixed with `Exploring - [Chapter]` while exploring.
- **Return point persistence**: Saved to `.crosspoint/epub_{hash}/returnPoint.bin` so it survives sleep/reboot.
- **Orientation change**: Automatically returns to the saved position before reflowing, since page numbers can't be remapped across layouts.
- **Disabling Explore Mode**: Clears the return point and locks in the current page so no stale state remains.

## Additional Context

* Turning the explore mode setting off and losing your `returnPoint` feels bad. I would rather it returns them to home first, then removes the `returnPoint`, but I am unsure how or if I even can do that from the settings menu.
* ~~I included commit [41dcea4](https://github.com/crosspoint-reader/crosspoint-reader/commit/41dcea43568dfbe2b954f67c85684c35471d13ae) that alters the "Select Chapter" logic to navigate to page 1 of the chapter even if you are currently in the same chapter. I was wanting to be able to quick jump to the start of the chapter in "Explore Mode" and realized I couldn't with the current "Select Chapter" logic. This may be dropped if it is determined to be out of scope of this change or if the original logic is preferred.~~ Scratch that, #1981 by Disasm addressed this and merged in!
* This change caused crosspoint to crash frequently with `assert failed: xTaskPriorityDisinherit`. It felt like a race condition with SD Card. It's been a minute since I have programmed anything embedded system, so I gave Claude the logs and my suspicions to debug. Commits [9a382db](https://github.com/crosspoint-reader/crosspoint-reader/pull/2118/changes/9a382dbe0867290dc7127823f26a9e2653572485) and 
[248b05b](https://github.com/crosspoint-reader/crosspoint-reader/pull/2118/changes/248b05ba202afc5b9b80c699356e0566539aa7f5) shows the fix Claude made in `HalStorage.cpp`. This seems like a change that could possibly have a reaching impact, so I would appreciate extra scrutiny there as well as any thoughts you more experienced devs may have. The crashes seem to have largely disappeared after this change.
[crash_report.txt](https://github.com/user-attachments/files/28170974/crash_report.txt) · [crash_report-2.txt](https://github.com/user-attachments/files/28170975/crash_report-2.txt) · [crash_report-3.txt](https://github.com/user-attachments/files/28170976/crash_report-3.txt)

### Screenshots

| | |
|---|---|
| Red-Rising_ch8_p1_2pct_574037 | screenshot-749414 |

### TODO
- [ ] Documentation, especially for new returnPoint.bin file
- [x] Footnote back press can be stale when you "Go to..." fixed with chapter jump, not yet fixed for percent jump
- [x] AI added a `pendingSyncSaveError` pop-up on failed return point retrieval. This is misleading since this is for KOReader sync, and isn't need. We should just log instead
- [x] Allow return point to persist rotation. I assumed it would be tricky and put it off. I think it might be easier than I was guessing
- [x] `returnPoint` validation. What if the book changed from the outside? or other edge cases
- [x] Add return point after nav to new bookmarks (yay!)
- [ ] Double check for conflicts with #1981 since it touched similar code and could introduce edge cases. I've already rebased and handled conflicts, but I want to look closer.

---

### AI Usage


pablohc:
Thanks!!!! I'll give it a try and let you know what I think! ;)

ZephByte:
Probably should have done this first... but I just opened a [Feature Proposal thread](https://github.com/crosspoint-reader/crosspoint-reader/discussions/2164) in the project discussions.

Uri-Tauber:
@ZephByte Yesterday Bookmarks support was added to CrossPoint. I **think** you can refactor your work here to use bookmarks under the hood (meaning Explore Mode can programmatically drop a hidden, temporary "system bookmark" when a navigation event occurs). 

That said, looking at the project new scope at #2149, this specific feature is likely going to fall outside of the core scope for the near future. As a result, even with a refactor, this PR might not be merged for a couple of months. Let me know what you think.

ZephByte:
> @ZephByte Yesterday Bookmarks support was added to CrossPoint. I **think** you can refactor your work here to use bookmarks under the hood (meaning Explore Mode can programmatically drop a hidden, temporary "system bookmark" when a navigation event occurs).

Oh good idea. I did notice the bookmark addition, but it never occurred to me to repurpose it for the return point. That would probably help cut down on new code additions and avoid the added `returnPoint.bin` file. I'll check it out!

> That said, looking at the project new scope at #2149, this specific feature is likely going to fall outside of the core scope for the near future. As a result, even with a refactor, this PR might not be merged for a couple of months. Let me know what you think.

Probably. This change was mostly a nice-to-have for myself. I figured I'd push it to a PR in case anyone else found it useful. I agree, it's probably more important to focus on optimization and cleanup for now. I'll likely be keeping this branch up to date and fiddling with it for personal use in the meantime. If it eventually gets merged, great! If not, I won't lose any sleep.

Uri-Tauber:
@ZephByte Awesome. Thanks for your work here.

PR #2119, opened 2026-05-23
feat: add Bluetooth file transfer

sabraman:
## Summary

Adds a Bluetooth Transfer mode under File Transfer for direct reader-to-host transfers without starting Wi-Fi or WebDAV. This is aimed at small operational workflows where Bluetooth is the lower-friction path: pulling a crash report, dropping an EPUB or BMP onto the device, and staging firmware from a nearby host.

## Scope

- Add a Bluetooth Transfer activity and File Transfer menu entry.
- Advertise a CrossPoint-specific BLE transfer service with a visible six-digit first-use code.
- Support EPUB uploads into `/Books` and BMP uploads into `/Pictures`.
- Support firmware `.bin` upload, SD staging, image validation, on-device confirmation, OTA flashing, and restart.
- Support authenticated crash-report download from `/crash_report.txt`.
- Add trusted-host auth using host id plus HMAC over the device nonce, saved only after a successful authenticated upload.
- Add a Python BLE transfer CLI for books, BMPs, firmware OTA, and crash-report download.
- Document the protocol, capabilities, browser companion URL, and CrossPoint-specific limits.

## Explicit Non-goals

CrossPoint does not gain package/plugin support here. `SCOPE.md` and `GOVERNANCE.md` keep CrossPoint focused on reader-owned workflows rather than a package/plugin platform, and Marginalia exists for that broader model.

The BLE service therefore does not expose package uploads, package-state diagnostics, arbitrary SD browsing, arbitrary path reads, or arbitrary path writes. These are not deferred CrossPoint BLE work.

## Browser Companion

The reader advertises `https://ble.xteink.lol/` and protocol capability fields so the shared browser companion can hide unsupported operations. The matching companion PR is marginalia-os/ble-xteink#6.

## Validation

- `python3 -m py_compile scripts/ble_transfer.py`
- `git diff --check upstream/master..HEAD`
- `./bin/clang-format-fix`
- `pio run`
- `pio check -e default`

## Notes

The local SDK submodule had to be initialized with `git submodule update --init --recursive open-x4-sdk` before PlatformIO could build. No `open-x4-sdk` code changes were needed.


Uri-Tauber:
@sabraman I have to admit I’m not particularly in favor of this feature. CrossPoint already provides several ways to transfer files to the device — Web Server, WebDAV, and OPDS - not to mention simply removing the SD card.

In my opinion, we should probably draw the line somewhere, especially considering BLE’s very limited transfer speed, which makes this use case feel somewhat unnecessary to me.

That said, this is only my personal opinion. If others believe the feature adds value, I’m happy to open a discussion here.

We also already had a previous PR for Bluetooth file transfer: #483.

rhythmerc:
> @sabraman I have to admit I’m not particularly in favor of this feature. CrossPoint already provides several ways to transfer files to the device — Web Server, WebDAV, and OPDS - not to mention simply removing the SD card.
> 
> In my opinion, we should probably draw the line somewhere, especially considering BLE’s very limited transfer speed, which makes this use case feel somewhat unnecessary to me.
> 
> That said, this is only my personal opinion. If others believe the feature adds value, I’m happy to open a discussion here.
> 
> We also already had a previous PR for Bluetooth file transfer: #483.

Just wanted to put support in for a discussion here! Low-friction article sending from phones is a big use case desire for me, and I felt the PR you linked presented exactly that.

sabraman:
@Uri-Tauber I understand the concern about adding another transfer path.

For me the main value is not raw throughput. My XTEINK very often fails to connect to my home and work Wi-Fi and eventually times out. The hotspot path is more reliable, but then I have to move my phone/laptop onto the reader’s network, which also means losing normal internet access on that device while doing the transfer.

BLE avoids that setup step. For small local operations, that matters more than peak speed: sending one EPUB, dropping a BMP, or fetching a crash report is quicker and less fragile for me over BLE than trying to get the Wi-Fi path working first.

I also measured OTA locally, and with the current implementation it is only about 15 seconds slower than doing the same update over the serial path, so the speed tradeoff is not as bad as I expected for maintenance tasks.

About #483: I think this PR is a different pass at the same idea. It includes the browser companion flow, trusted-host pairing, crash report download, and OTA staging, so it is more focused on device maintenance and small direct transfers than only phone-to-reader file sending.

itsthisjustin:
I'm going to close this as out of scope for NOW but I do think we can eventually get this merged in a future phase. Incredibly cool work here but out of scope for current roadmap

PR #2122, opened 2026-05-23
fix(epub): correct footnote page assignment when hyphenation is active

borisfaure:
Hyphenation inserts remainder tokens into ParsedText::words[], inflating the post-layout word count relative to the pre-layout index stored in pendingFootnotes. This caused footnotes whose anchor fell near a hyphenated word to be assigned to the wrong (earlier) page.

Track which words are hyphenation remainders via wordIsHyphenRemainder[]. extractLine() now computes a pre-hyphenation word count per line and passes it to addLineToPage(), which uses it to accumulate wordsExtractedInBlock. Both wordIndex and the accumulator now count in the same coordinate system, fixing the page assignment.

Bump SECTION_FILE_VERSION 23 → 24 to invalidate stale caches.

## Summary

* **What is the goal of this PR?** Fix a bug about the footnotes not being accessible in some cases when hyphenation is on.

## Additional Context

bug footnotes
In this image, there are 2 footnotes: 22 and 23. When going to the footnotes pages, only 23 is listed.

---

### AI Usage

Did you use AI tools to help write this code? _**< YES  >**_

### DIFF
--- a/lib/Epub/Epub/ParsedText.cpp
+++ b/lib/Epub/Epub/ParsedText.cpp
@@ -138,6 +138,7 @@ void ParsedText::addWord(std::string word, const EpdFontFamily::Style fontStyle,
     wordStyles.push_back(baseStyle);
     wordContinues.push_back(attachToPrevious);
     wordIsFocusSuffix.push_back(false);
+    wordIsHyphenRemainder.push_back(false);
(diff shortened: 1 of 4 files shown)


itsthisjustin:
Closing this in favor of #2508

PR #2125, opened 2026-05-24
fix: bound network fetches

sabraman:
## Summary
- add connect/read timeouts to shared HTTP fetch and download paths
- cap streamed feed fetches at 1 MiB to avoid unbounded reads
- cap KOReader sync response buffering at 16 KiB

## Validation
- git diff --check
- pio run


jeremydk:
The httpdownloader half of this has been superseded by upstream -- can you extract out the koreader sync changes? those would still be a solid addition.

itsthisjustin:
Closing in favor of existing upstream work that should harden this. Thanks for your input on this though @sabraman

PR #2126, opened 2026-05-24
fix: verify HTTPS downloads

sabraman:
## Summary
- replace unverified HTTPS clients in shared downloads with the ESP certificate bundle
- keep plain HTTP behavior unchanged
- leave OTA and KOReader sync on their existing certificate-bundle paths

## Validation
- git diff --check
- pio run


jeremydk:
Have you done any validation of the remaining heap after this? HTTPS has proven to be surprisingly expensive on these devices.

sabraman:
Added a follow-up commit (`db31853`) that logs HTTPS heap before/after `HttpDownloader` requests and refuses HTTPS fetch/download when free heap is below the existing 55KB TLS threshold used by KOReader sync.

I also validated this on an ESP32-C3/X4 over serial with an HTTPS OPDS fetch:

```text
[HTTP] HTTPS fetch heap before request: free=107584 min=105580 maxAlloc=102388
[HTTP] HTTPS fetch heap after request:  free=55240  min=46028  maxAlloc=49140
[HTTP] Fetch success
```

So the concern was valid: certificate-verified HTTPS did not crash in this test, but the handshake/request path pushed min-free down to ~46KB and left max alloc at ~49KB afterward. The new guard gives us a clean failure path before starting TLS when the device is already below the safer threshold.

itsthisjustin:
@sabraman we need conflicts fixed and @jeremydk I'm going to leave this up to you to decide to merge or not

itsthisjustin:
Closing for now as we are deciding not to bundle certs into the firmware. Let me know if I'm off base here though and there's a solid reason to keep this in after the wolfssl stuff was merged.

PR #2127, opened 2026-05-24
fix: restart web server after Wi-Fi reconnect

sabraman:
## Summary
- stop the STA web server when Wi-Fi drops so stale HTTP/WebSocket/UDP sockets are closed
- keep monitoring Wi-Fi while the server is stopped during reconnect
- recreate the web server and refresh the displayed IP after Wi-Fi returns

## Validation
- git diff --check
- pio run
- coderabbit review --agent -t uncommitted

### DIFF
--- a/src/activities/network/CrossPointWebServerActivity.cpp
+++ b/src/activities/network/CrossPointWebServerActivity.cpp
@@ -274,7 +274,7 @@ void CrossPointWebServerActivity::loop() {
     }
 
     // STA mode: Monitor WiFi connection health
-    if (!isApMode && webServer && webServer->isRunning()) {
+    if (!isApMode && webServer) {
       static unsigned long lastWifiCheck = 0;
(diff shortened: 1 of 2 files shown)


jeremydk:
This one isn't your fault, per se, but this code is going to make us hit it a lot more. startWebServer() uses bare new CrossPointWebServer(), which aborts on OOM under -fno-exceptions instead of returning null. The PR runs this allocation on every Wi-Fi reconnect rather than once at activity entry, so a heap constrained reconnect turns into a hard crash. Use makeUniqueNoThrow and null check; the existing onGoHome() error path is already the right recovery.
  
state stays SERVER_RUNNING while the server is actually stopped during a Wi-Fi drop. "Is the server up?" now has to be answered by webServer->isRunning(), not state. Either add a WAITING_FOR_STA state or comment the field so the next reader doesn't trust the name.

I think this also may do weird things with momentary blips during upload, I need to think through the state machine here more.

Issue #1952, opened 2026-05-12
Feature Request: More Customisable Sleep Timer


GrimblyEnthusiast:
I'd love it if I could set a longer time for when it goes to sleep, and if theres an option to turn off auto sleep. I'm not sure if booting actually takes that much extra battery, but I use the x4 inbetween chores and helping customers etc, and it's annoying to have to boot again every time, even if it is just really short.

Thank you!

Uri-Tauber:
Closing as solved in #2206.

PR #2128, opened 2026-05-24
feat: Disable auto sleep and add longer timeouts

Joseph-DiGiovanni:
## Summary

Allow disabling auto sleep. As requested in #1952

## Additional Context

I'll be honest, I thought this was a silly feature request at first. What if you forget to put the device to sleep and its dead next time you go to use it? I thought about just adding a super long option like 12 hours to play it safe, but after seeing how little the battery drained while left awake for hours, it would probably take 2+ days to die even while left on.

So, I say let people do as they please and disable it if they really want to.

---

### AI Usage


src/SettingsList.h
```diff
@@ -179,7 +179,7 @@ inline std::vector<SettingInfo> getSettingsList(const SdCardFontRegistry* regist
         // --- System ---
         SettingInfo::Enum(StrId::STR_TIME_TO_SLEEP, &CrossPointSettings::sleepTimeout,
                           {StrId::STR_MIN_1, StrId::STR_MIN_3, StrId::STR_MIN_5, StrId::STR_MIN_10, StrId::STR_MIN_15,
-                           StrId::STR_MIN_30},
+                           StrId::STR_MIN_30, StrId::STR_NEVER},
                           "sleepTimeout", StrId::STR_CAT_SYSTEM),
```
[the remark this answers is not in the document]
Joseph-DiGiovanni: This is intended, and not even part of this PR

USER_GUIDE.md
```diff
@@ -192,7 +192,7 @@ The Settings screen allows you to configure the device's behavior. There are a f
 
 #### 3.6.4 System
 
-- **Time to Sleep**: Set the duration of inactivity before the device automatically goes to sleep; options are 1, 5, 10 (default), 15 or 30 minutes.
+- **Time to Sleep**: Set the duration of inactivity before the device automatically goes to sleep; options are 1, 3, 5, 10 (default), 15, 30 minutes, or never.
```
lpla: Why 3?

lpla (changes requested):

USER_GUIDE.md
Joseph-DiGiovanni: A previous PR added 3 minutes, but never updated the user guide.

lpla (approved):

Eloren1:
Maybe add  2hrs, 6hrs and 12hrs as well? If 3 mins added between 1m and 5m, then there should be something like this as well because otherwise it's too extreme and risky

Joseph-DiGiovanni:
> Maybe add 2hrs, 6hrs and 12hrs as well?

I've already mentioned why I didn't add those in the PR summary. I don't mind adding them personally I just wanted to limit the scope of the PR to leave the debate of what/how many longer options to add for another day.

Joseph-DiGiovanni:
I had to resolve PR conflicts anyway so I went ahead and added 1, 2 and 4 hour timeouts. I leaned towards relatively shorter times as they're more likely to be used and conserve battery.

Eloren1 (commented):
Now with multiple hour options it looks nice!

Joseph-DiGiovanni:
Closed due to conflicts with #2206

PR #2130, opened 2026-05-24
perf: Speed up wifi reconnect

jpirnay:
## Summary

* **What is the goal of this PR?** Reduce the time spent connecting to a known
  Wi-Fi network from ~1200 ms to ~100 ms by caching per-credential
  BSSID/channel and IP configuration on first successful connect, and reusing
  them on subsequent attempts to skip channel scanning and DHCP.
  * Extend `WifiCredential` with `bssid[6]`, `channel`, `ip[4]`, `gateway[4]`,
    `mask[4]`, `dns[4]` (zero values mean "no cache, do a full scan +
    DHCP"). `WifiCredentialStore` gains `updateConnectionCache` /
    `clearConnectionCache`. Persisted to `wifi.json` alongside existing
    credential fields; existing files without these fields load unchanged.
  * `WiFi.begin(ssid, pwd, channel, bssid, true)` is used when a hint is
    cached, skipping the multi-channel scan inside the SDK.
  * `WiFi.config(ip, gw, mask, dns)` is applied before `WiFi.begin()` when a
    cached IP is available. Cache is keyed to the BSSID we're using, so a
    router subnet change is detected via BSSID mismatch + fallback.
  * Mesh / roam safety net: if the hint attempt hard-fails or doesn't
    complete in 3 s (`HINT_ATTEMPT_TIMEOUT_MS`), one silent retry with full
    scan + DHCP before surfacing failure. The stale cache is cleared so the
    next session starts fresh; on retry success the new ground truth
    overwrites it.
  * Drop the `delay(100)` previously inserted after `WiFi.disconnect(true,
    true)` — it was padding for an already-synchronous call.
  * Skip `WiFi.disconnect(true, true)` (which erases NVS) when WiFi status is
    already in an idle/uninitialised state — pure cost when there is nothing
    to tear down.
  * Add WiFi event handlers (`STA_CONNECTED` / `STA_GOT_IP`) plus per-attempt
    timing logs so the breakdown (pre-begin / association / DHCP) is visible
    in `LOG_DBG` output.

## Additional Context

Measured on an ESP32-C3 device against a stable home AP (RSSI ~ -70 dBm):

| Phase                | Before  | After  |
|----------------------|---------|--------|
| Pre-begin overhead   | ~106 ms | ~1 ms  |
| Channel scan + auth  | ~500 ms | ~75 ms |
| DHCP                 | ~1000 ms| ~1 ms  |
| **Total**            | **~1200 ms** | **~100 ms** |

The single biggest win is skipping DHCP via cached static IP; the BSSID hint
is essential to keep the static-IP path safe (cache invalidation depends on
detecting BSSID mismatch rather than a network-reachability probe, which
would cost most of the saved time).

Things worth a careful look:

* **Cache invalidation correctness.** The static IP is applied only when a
  BSSID hint is also present, and the hint-attempt timeout (3 s) plus silent
  retry without hint+IP handles the case where the cache is stale (mesh
  roam, channel change, ISP subnet change). The pathological case is "AP
  keeps the same BSSID but the router moves the subnet" — extremely rare in
  practice. If it happens, the user can hit "Forget network" to wipe the
  cache.
* **Per-credential JSON layout.** New `bssid` (12-char lowercase hex),
  `channel`, `ip`/`gw`/`mask`/`dns` (dotted-quad strings), and `ts` keys.
  Loader tolerates missing fields, rejects malformed dotted-quads
  (trailing-character check via `%n`), and otherwise ignores garbage.
* **NVS skip safety.** `prepareForConnect()` now skips `WiFi.disconnect(true, true)` (which erases the SDK-managed NVS copy of the AP config) when `WiFi.status()` indicates there's nothing to tear down. Upstream calls it unconditionally on every attempt. `WiFi.persistent(false)` is still called before each `WiFi.begin()`, and the SSID/password are always passed explicitly, so stale NVS state shouldn't influence which AP is chosen. If reviewers know of an IDF/SDK path that depends on a fresh NVS erase per connect, please flag it.

The added `LOG_DBG` lines (`WiFi.begin -> ... staticIp=...`,
`EVT associated`, `EVT got_ip`, `Connected to ... in NNN ms`) are gated by
`LOG_LEVEL=2` and suppressed in `gh_release` builds. They make this code
self-diagnosing for future regressions; happy to strip them if you prefer.

---

### AI Usage

### DIFF
--- a/src/JsonSettingsIO.cpp
+++ b/src/JsonSettingsIO.cpp
@@ -263,6 +263,29 @@ bool JsonSettingsIO::saveWifi(const WifiCredentialStore& store, const char* path
     JsonObject obj = arr.add<JsonObject>();
     obj["ssid"] = cred.ssid;
     obj["password_obf"] = obfuscation::obfuscateToBase64(cred.password);
+    bool hasHint = cred.channel != 0;
+    for (int i = 0; i < 6 && !hasHint; i++) {
+      if (cred.bssid[i] != 0) hasHint = true;
+    }
+    if (hasHint) {
+      char bssidHex[13];
+      snprintf(bssidHex, sizeof(bssidHex), "%02x%02x%02x%02x%02x%02x", cred.bssid[0], cred.bssid[1], cred.bssid[2],
+               cred.bssid[3], cred.bssid[4], cred.bssid[5]);
+      obj["bssid"] = bssidHex;
+      obj["channel"] = cred.channel;
+    }
+    if (cred.ip[0] != 0) {
+      char buf[16];
+      snprintf(buf, sizeof(buf), "%d.%d.%d.%d", cred.ip[0], cred.ip[1], cred.ip[2], cred.ip[3]);
+      obj["ip"] = buf;
+      snprintf(buf, sizeof(buf), "%d.%d.%d.%d", cred.gateway[0], cred.gateway[1], cred.gateway[2], cred.gateway[3]);
+      obj["gw"] = buf;
+      snprintf(buf, sizeof(buf), "%d.%d.%d.%d", cred.mask[0], cred.mask[1], cred.mask[2], cred.mask[3]);
+      obj["mask"] = buf;
+      snprintf(buf, sizeof(buf), "%d.%d.%d.%d", cred.dns[0], cred.dns[1], cred.dns[2], cred.dns[3]);
+      obj["dns"] = buf;
+      obj["ts"] = cred.cacheTimestamp;
+    }
   }
 
   String json;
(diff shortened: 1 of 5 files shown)


jeremydk:
I am not certain that skipping dhcp to save a second is going to be worth the potential, challenging to track, network issues and bugs this is capable of causing on real world networks.

mcrosson:
I agree with @jeremydk this is going to create a nightmare of network issues on a lot of consumer hardware. Especially after any restarts of a consumer grade dhcp endpoint. On top of that some of us run checks for ip/arp oddities and this would definitely trip some of those sensors.

All this to save maybe 2s? This is not worth the support cost, pain of troubleshooting and more.

DHCP has timeout rules for good reason, ignoring them and going off into 'wildly unsupported, novel approaches' land is *not* a good call.

PR #2133, opened 2026-05-24
feat: new options for Refresh Frequency

matteoscopel:
## Summary

This PR adds the following options to the "Refresh Frequency" setting:

- 100 pages
- Never

The first two options increase the number of pages before a full refresh is made; the latter prevents a refresh from happening at all.

## Additional Context

After noticing that ghosting was not significant even after 30 page turns on my X4, I've started pushing that number and seeing if any artifact would arise. At 100 page turns, the quality is not noticeably impacted:

I've kept experimenting, reaching without particular trouble much higher numbers (i.e. 300).
I think the main benefits one gets from increasing the pages before CrossPoint does a full refreshes are:

- less power used
- less interruptions in the reading experience

Ghosting can, however, become an issue. Moreover, if a lot of writes are done to the screen without a full refresh, when the screen is turned off it tends to turn grey. This is visible with Quick Resume enabled on timeout.

Additionally, doing some very superficial math with the help of a chatbot (so take that with a pinch of salt), considering that a full refresh should consume 7-8 mJ per cm2, we get that refresh intervals higher than 100 tend to have a reduced energy saving impact compared to the difference between the energy saved between an interval of 30  pages and one of 100.

Taking this all into account, I've considered it safe to add the intermediate interval value of 50 and a maximum of 100, and no more than that.

This PR also adds the possibility of disabling the refresh altogether, leaving it up to the user to eventually trigger it manually. As raised on Discord by _Mein Name ist Hase_ (who has also contributed to this PR), this can be useful both for those who are bothered by the refresh but it can also bypass an issue on the X3 that causes it not to refresh when it should. 

---

### AI Usage


Uri-Tauber (dismissed):
LGTM. Haven't tested on my device.
I would suggest to remove one of the frequencies (maybe 50 pages) to avoid cluttering the toggle, but not a blocker.

Uri-Tauber:
@matteoscopel When refresh is set to off, do we still get full refresh between chapters?

matteoscopel:
@Uri-Tauber I have removed the option to refresh every 50 pages. I think a 100 pages interval makes sense, as it allows one to still have a refresh once in a while to avoid eventual ghosting.

And no, when refresh is set to off, we do not get a full refresh between chapters. Should that happen? Right now, I'm not getting full refreshes between chapters even on 1.3.0 before this PR.

uxjulia:
Tested on device, didn't get up all the way to 100, but it definitely wasn't refreshing at 30 pages when I had it set to "Never". LGTM.

uxjulia (approved):

jeremydk (changes requested):
I'm really not sure this is a change we should be taking. This is a fundamental limitation of an EPD and outside of ghosting, I think we owe it to our users to opt for safety.

itsthisjustin:
I don't really know if I see the value in this feature. What are we gaining here?

matteoscopel:
Thank you all for the feedback!

@itsthisjustin We gain:

- less or no refreshes: some people can find it pleasant to read with less interruptions
- potentially longer battery life, as a full refresh is more energy-intensive than a partial one

which I realize, in the light of what I'm going to say next, is not much:

@jeremydk I absolutely see what you mean, one wouldn't want to risk causing damage to the HW of people and it's wise to privilege safety over features.

I'll close this PR, as I feel I can't guarantee that it won't break someone's screen :)

Issue #2047, opened 2026-05-18
Crash report - Panic reason: assert failed: xTaskPriorityDisinherit tasks.c:5156 (pxTCB == pxCurrentTCBs[ 0 ])


dpanych:
### Affected Version

1.3.0

### Describe the Bug

I am testing out different version of the Bible, specially table of contents/chapter formatting; one of the times when I opened the epub, I got this crash report/panic. Not exactly sure what caused it. Is anyone able to dissect the crash_report.txt and see what happened exactly?

### Steps to Reproduce

Unable to reproduce.

### Expected Behavior

[crash_report.txt](https://github.com/user-attachments/files/27967716/crash_report.txt)

### Relevant Log Output/Screenshots

```shell
CrossPoint version: 1.3.0

Panic reason: assert failed: xTaskPriorityDisinherit tasks.c:5156 (pxTCB == pxCurrentTCBs[ 0 ])

Last logs:
[18] [INF] [HW] Using cached device type: X3
[19] [INF] [GYR] QMI8658 IMU found at 0x6B
[20] [INF] [GYR] QMI8658 gyro initialized and put to sleep
[21] [INF] [MAIN] Hardware detect: X3
[22] [INF] [HW] Using cached device type: X3
[24] [INF] [GYR] QMI8658 IMU found at 0x6B
[25] [INF] [GYR] QMI8658 gyro initialized and put to sleep
[26] [INF] [MAIN] Hardware detect: X3


Stack memory:
0x3FCB14C0: 0x3FCA8000 0x3FCA1B68 0x3C575224 0x4038E022 0x00000602 0x00000001 0x3FC94B20 0x3FC94240 
0x3FCB14E0: 0x00000602 0x00000001 0x3FC94B20 0x36353135 0x3FCB1500 0x3FC94228 0x3C575224 0x3FC94728 
0x3FCB1500: 0x3C1B5122 0x3FC94238 0x3FCB14EC 0x3FC9423C 0x3C1B53A0 0x3FC94240 0x00000000 0x00000000 
0x3FCB1520: 0x3F000000 0x65737361 0x66207472 0x656C6961 0x78203A64 0x6B736154 0x6F697250 0x79746972 
0x3FCB1540: 0x69736944 0x7265686E 0x74207469 0x736B7361 0x353A632E 0x20363531 0x54787028 0x3D204243 
[...]

PR #2135, opened 2026-05-25
fix: serialize SdFat FsFile close through HalStorage mutex

jeremydk:
SdFat's SdSpiCard tracks SPI bus state with an unsynchronized m_spiActive bool. When two tasks call into SdFat concurrently they can confuse that state machine, ending with one task calling SPIClass::endTransaction() against a paramLock the other task holds. That trips FreeRTOS's xTaskPriorityDisinherit assert (tasks.c:5156, pxTCB == pxCurrentTCBs[0]) and panics the system.

HalStorage already serialized every explicit method call via storageMutex, but HalFile's destructor was `= default`, which let the underlying SdFat FsFile destructor run close() outside any lock (DESTRUCTOR_CLOSES_FILE=1). Any task that destructed a HalFile while another task was mid-SD-op would race the unsynchronized state.

Move the locking discipline into HalFile::Impl::~Impl: an explicit close() under StorageLock, then the FsFile member destructor's redundant close() is a no-op. HalFile's special members can stay = default.

Switch storageMutex to xSemaphoreCreateRecursiveMutex so openFileForRead and openFileForWrite can hold the lock while assigning to a HalFile& out-param whose prior Impl needs locked teardown. Priority inheritance still applies to recursive mutexes.

Also documented the no-bypass rule in CLAUDE.md: never call SdFat / SdSpiCard / FsBaseFile / SDCardManager directly, never define HAL_STORAGE_IMPL outside HalStorage.cpp.

Addresses my admittedly synthetic repro for #2047

### DIFF
--- a/.skills/SKILL.md
+++ b/.skills/SKILL.md
@@ -161,6 +161,12 @@ if (Storage.openFileForRead("MODULE", "/path/to/file.bin", file)) {
 
 **Usage**: See example above. Uses `FsFile` (SdFat), NOT Arduino `File`. Do NOT add `file.close()` for local variables (see DESTRUCTOR_CLOSES_FILE above).
 
+**SdFat is not thread-safe; all SD access MUST go through HalStorage**:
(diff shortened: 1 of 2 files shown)


uxjulia (approved):

itsthisjustin:
Does this fix the crash report txt thing I reported by any chance?

jeremydk:
> Does this fix the crash report txt thing I reported by any chance?

I'm *pretty sure* it's the same underlying crash when SPI is under load.  Can't say for sure because this has been a pain to hit organically.

itsthisjustin (approved):

PR #2145, opened 2026-05-25
fix: replace deprecated WebSockets flush call in ESP32 disconnect path

lpla:
## Summary

  Keep Arduino-ESP32 3.x builds warning-clean without changing the current WebSockets disconnect behavior used by CrossPoint.

  Add a small pre-build patch script that rewrites the WebSockets 2.7.3 ESP32 disconnect path from `client->tcp->flush()` to `client->tcp->clear()`, and register that script in PlatformIO `extra_scripts`.

## Additional Context

WebSockets 2.7.3 still calls `flush()` in its client disconnect path. On Arduino-ESP32 3.x, `NetworkClient::flush()` is deprecated with the message `Use clear() instead.` The current implementation delegates `flush()` to `clear()`, so this change preserves the effective behavior while removing the deprecation warning.

The patch is intentionally narrow. It only changes the `NETWORK_ESP32` path, keeps the existing `flush()` path for other network types, and fails fast if the upstream WebSockets source no longer matches the expected pattern. That avoids silently patching an unexpected upstream revision.

This also follows the same local-libdep patch approach already used elsewhere in the project for JPEGDEC.

## Verification

Built locally with `pio run` on macOS using the default environment and confirmed the WebSockets deprecation warning no longer appears.

---

### AI Usage

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -53,6 +53,7 @@ extra_scripts =
   pre:scripts/gen_i18n.py
   pre:scripts/git_branch.py
   pre:scripts/patch_jpegdec.py
+  pre:scripts/patch_websockets.py
 
 ; Libraries
 lib_deps =
--- a/scripts/patch_websockets.py
+++ b/scripts/patch_websockets.py
@@ -1,0 +1,48 @@
+"""
+PlatformIO pre-build script: patch WebSockets for Arduino-ESP32 3.x.
+
+WebSockets 2.7.3 still calls WiFiClient::flush() in the client disconnect
+path. On Arduino-ESP32 3.x, WiFiClient is a typedef for NetworkClient and
+NetworkClient::flush() is deprecated in favor of clear(). The implementation
+currently delegates flush() to clear(), so this patch preserves behavior while
+removing the compiler warning.
+"""
+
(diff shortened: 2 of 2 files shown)


lpla:
Bots got confused because I pushed the wrong branch. I just force-pushed the right one.

jeremydk (changes requested):
This patch will silence the warning, but the call accomplishes nothing observable on either side of the patch. 

Upstream WebSockets already reached this conclusion on the server side: WebSocketsServer.cpp:549 excludes NETWORK_ESP32 from the flush() entirely, and NetworkClientSecure::flush() is {}.

The right fix to match that precedent and address the warning is one extra && != NETWORK_ESP32 term on the existing #if, or just accept a warning line with a no-op, avoiding the new file, patch script, pinned-version coupling, and client/server asymmetry.

If we do take this, IMO, we'd either want to take this as a community fork, or via a git-file patch. I'm really not crazy about the inline string patching.

Or just roll with a warning message during build that doesn't impact us.

lpla:
Thanks for the review. I agree with this assessment. This was more a warning OCD silencer more than a useful patch, honestly. But we can just live with the warning.

I’m going to close this PR.

PR #2149, opened 2026-05-25
chore: Clarify project scope and development priorities

itsthisjustin:
Expand scope to support multiple e-ink devices beyond X4. Add guiding principle to fill gaps in stock firmware rather than duplicate functionality. Establish current focus on memory/flash optimization and code cleanup. Temporarily freeze new themes and network connectors to consolidate codebase before multi-device expansion.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -3,16 +3,34 @@
 * **What is the goal of this PR?** (e.g., Implements the new feature for file uploading.)
 * **What changes are included?**
 
+## Scope Check
+
+CrossPoint is intentionally narrow. See [SCOPE.md](../blob/master/SCOPE.md) and [ROADMAP.md](../blob/master/ROADMAP.md).
+Please confirm:
+
+- [ ] I have read SCOPE.md and ROADMAP.md.
+- [ ] This PR is **not** a new built-in theme (themes are temporarily closed pending the move to SD-loaded themes).
+- [ ] This PR is **not** a new external network connector (sync engine, cloud storage, remote file access, etc.).
+- [ ] This PR is **not** an interactive app, writing tool, RSS/news/browser, media playback, or PDF feature.
+- [ ] The stock firmware does not already handle this well, **and** no other popular CrossPoint fork already does
+      (or, if one does, I explain why CrossPoint still needs it below).
+- [ ] If this PR touches `open-x4-sdk/`, `lib/hal/`, the bootloader, OTA, or recovery code, I have coordinated with
+      the relevant maintainer (currently @jeremydk for SDK abstraction and bootloader work).
+
+**If this PR was opened against the previous (broader) scope and was already in flight under Phase 0, link the
+relevant Discussion or issue so reviewers can see the history.**
+
 ## Additional Context
 
-* Add any other information that might be helpful for the reviewer (e.g., performance implications, potential risks, 
+* Add any other information that might be helpful for the reviewer (e.g., performance implications, potential risks,
   specific areas to focus on).
+* Memory / flash impact, if known.
 
 ---
 
 ### AI Usage
 
-While CrossPoint doesn't have restrictions on AI tools in contributing, please be transparent about their usage as it 
+While CrossPoint doesn't have restrictions on AI tools in contributing, please be transparent about their usage as it
 helps set the right context for reviewers.
 
 Did you use AI tools to help write this code? _**< YES | PARTIALLY | NO >**_
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -14,8 +14,8 @@ Please confirm:
 - [ ] This PR is **not** an interactive app, writing tool, RSS/news/browser, media playback, or PDF feature.
 - [ ] The stock firmware does not already handle this well, **and** no other popular CrossPoint fork already does
       (or, if one does, I explain why CrossPoint still needs it below).
-- [ ] If this PR touches `open-x4-sdk/`, `lib/hal/`, the bootloader, OTA, or recovery code, I have coordinated with
-      the relevant maintainer (currently @jeremydk for SDK abstraction and bootloader work).
+- [ ] If this PR touches `freeink-sdk/`, `lib/hal/`, the bootloader, OTA, or recovery code, I have coordinated with
+      the relevant maintainer.
 
 **If this PR was opened against the previous (broader) scope and was already in flight under Phase 0, link the
 relevant Discussion or issue so reviewers can see the history.**
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -17,12 +17,14 @@ hanging, and so we enter the stricter phases with a clean slate.
 
 * **RTL support PRs** that are currently open. Reviewing, iterating, and merging the in-flight right-to-left work.
 * **Dictionary PR** that is currently open. Reviewing and merging the offline dictionary lookup work.
+* **Bookmarks** feature. Finishing the in-flight bookmarks work so users have first-class navigation markers in
+  EPUBs.
 * **Transparent sleep screens** (potential). If a clean implementation lands during this phase, it is in. If it
   stalls or balloons in scope, it moves out and is not picked back up under the stricter phases.
 
 **Exit criteria for Phase 0:**
 
-* The RTL and dictionary PRs are either merged or explicitly closed with a reason.
+* The RTL, dictionary, and bookmarks PRs are either merged or explicitly closed with a reason.
 * Transparent sleep screens are either merged or shelved.
 * No other "legacy scope" features are accepted during this phase. New work that does not fit the post-Phase-0 scope
   should wait.
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -45,6 +45,11 @@ work has room to breathe.
 * Flash footprint reduction (dead code, redundant strings, oversized tables).
 * Refactors that tighten the HAL / SDK boundary in preparation for non-Xteink ESP32 devices.
 * Moving themes off-firmware to SD-loaded assets (see SCOPE.md Section 6).
+* **Moving hyphenation files off-firmware.** Hyphenation rules vary per language and the files are large (German
+  alone is ~200KB). Today these eat flash budget that should be available for the reader core. The plan is to build
+  a downloader analogous to the existing font downloader and store the dictionaries on SD / SPIFFS, loading on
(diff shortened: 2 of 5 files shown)


ROADMAP.md
```diff
@@ -0,0 +1,107 @@
...
+
+**Goal:** Reduce memory and flash usage and clean up the codebase so that future device support and reading-quality
+work has room to breathe.
+
+**Focus areas:**
+
+* DRAM and heap fragmentation reduction across the reader core.
+* Flash footprint reduction (dead code, redundant strings, oversized tables).
+* Refactors that tighten the HAL / SDK boundary in preparation for non-Xteink ESP32 devices.
+* Moving themes off-firmware to SD-loaded assets (see SCOPE.md Section 6).
```
Uri-Tauber: I don't really mind this, but I think moving the hyphenation files somewhere else will be much more beneficial.

ROADMAP.md
```diff
@@ -0,0 +1,107 @@
...
+Phases are sequential. We do not start the next phase until the prior one is wrapped or explicitly carried over.
+
+---
+
+## Phase 0 - Close Out Legacy Scope Items
+
+**Goal:** Land the work that was already in motion under the prior, broader scope so contributors are not left
+hanging, and so we enter the stricter phases with a clean slate.
+
+**In Phase 0:**
```
Uri-Tauber: You forgot the bookmarks feature. We're really close to merge it.

SCOPE.md
```diff
@@ -1,62 +1,193 @@
...
+adjacent variants). Every accepted change should make that goal easier, not harder. Device-specific code should live
+behind the HAL / SDK boundary so the reader core stays portable.
+
+## 2. Guiding Principle: Fill Gaps the Stock Firmware Leaves
+
+CrossPoint exists to do the things the stock firmware does poorly or not at all. New work is evaluated against that
+delineator:
+
+* **Does the stock firmware already do this well?** If yes, we will not duplicate it.
+* **Is another popular CrossPoint fork already solving this well?** If yes, we generally defer to that fork rather than
```
Uri-Tauber, line 24: I'm not sure I understand this: won't it be better to migrate the feature to crosspoint **to avoid** fragmenting the user experience?

itsthisjustin: Depends on what it is I guess. I think support for CJK is an important thing to think about here. Why support Chinese when there's a crosspoint fork already and stock handles it well?

ROADMAP.md
itsthisjustin: Valid, will add

SCOPE.md
Uri-Tauber: Yeah. You're right about this. I thought about the reading stats from CrossInk.

ROADMAP.md
itsthisjustin: I'm not accutely aware of how hyphenation works. Can you explain what you mean a bit more?

SCOPE.md
itsthisjustin: I think that's a good example too actually. My general thought there is, that's awesome, let Julia be the fork for that. For folks that don't want that, they can stick to Crosspoint

ROADMAP.md
Uri-Tauber: Well, in a nutshell: in languages with very long words, you often want to split words across lines so you don’t end up with only a few words per line. You *can* just break at the line boundary, but that usually looks pretty ugly. Hyphenation files define the “correct” places where words can be split, and those rules vary by language.

German is the record holder here - I think its hyphenation file is around 200 KB.

If we can create something like the font downloader we already have, but for Hyphenation files, and then move them to SPIFF, it will probably be the best long-term move.

jeremydk:
This makes a lot of sense to me in general.

My thinking here:  We've got extremely limited ram, storage, and compute.  We've seen problems across the fork-ecosystem around updating, people getting stranded on older builds with broken updates, and frankly, we've been adding features (since all the forks have them!) rather than a focus on a stable core.

I, personally, think, that a focus on a light, fast, great ereader core with a focus on speed and correctness.  We _cannot_ add every feature being requested. Flash space, heap fragmentation, etc all cause issues, and every feature has been stepping on one another.

If we pull scope to this - I think, after some bumps, it helps the entire ecosystem;  We'd be lending the stronger, lighter core to each fork, and they'd be able to add /their/ custom feature set over the top.   I can very much see CJK focused forks, manga/image focused forks, etc.  I just don't think _we_ can fit it all in.

In the back of my head, one of my biggest concerns as we tackle CJK is the necessity of native readers.  As I was digging around to see what Korean (specifically) would take to implement, I needed to iterate back and forth with a native reader because I fundamentally didn't understand layout and hyphenation rules of a language I don't speak.

Along with this reorientation, I'm hoping that we can use this to remove some of the engineering chokepoints when it comes to OSS development -- the common files/functions that nearly every PR is merge-conflicting on.  We've been unable to really take a refactor/rearchitecture across any of these planes *due* to the tight coupling everywhere and the firehose of public PRs.

ROADMAP.md
itsthisjustin: Yes we should do that. I'm going to add that to the scope

Uri-Tauber:
Ha! 
I thought we merged this weeks ago!

@itsthisjustin Can you update this to reflect the recent changes (bookmarks and RTL support got merged. maybe other things), so we can merge this now.

itsthisjustin:
> Ha! 
> I thought we merged this weeks ago!
> 
> @itsthisjustin Can you update this to reflect the recent changes (bookmarks and RTL support got merged. maybe other things), so we can merge this now.

Yep need to it here and on website. Thought about it today

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -3,16 +3,34 @@
...
+Please confirm:
+
+- [ ] I have read SCOPE.md and ROADMAP.md.
+- [ ] This PR is **not** a new built-in theme (themes are temporarily closed pending the move to SD-loaded themes).
+- [ ] This PR is **not** a new external network connector (sync engine, cloud storage, remote file access, etc.).
+- [ ] This PR is **not** an interactive app, writing tool, RSS/news/browser, media playback, or PDF feature.
+- [ ] The stock firmware does not already handle this well, **and** no other popular CrossPoint fork already does
+      (or, if one does, I explain why CrossPoint still needs it below).
+- [ ] If this PR touches `open-x4-sdk/`, `lib/hal/`, the bootloader, OTA, or recovery code, I have coordinated with
+      the relevant maintainer (currently @jeremydk for SDK abstraction and bootloader work).
```
Uri-Tauber: Need to update those lines.

ROADMAP.md
```diff
@@ -0,0 +1,88 @@
...
+* Active connectivity features (RSS, news, browsers).
+* PDF rendering as a first-class format.
+
+See [SCOPE.md](SCOPE.md) for the full rationale.
+
+---
+
+## How This Roadmap Changes
+
+* Phase boundaries are decided by maintainers, not by individual PRs.
```
Uri-Tauber, line 86: Whats that means?

Uri-Tauber:
Overall, very good. Approving.

ROADMAP.md
itsthisjustin: Means we decide what goes into each phase, having a PR out doesn't change that

Uri-Tauber (approved):

SCOPE.md
brianhuster: > Why support Chinese when there's a crosspoint fork already and stock handles it well?

Does that mean Crosspoint doesn't want to support CJK? I am concered about this, because I don't want to switch between firmwares just for reading different books in different languages, especially as Crosspoint doesn't support extensions yet

Uri-Tauber: I believe CJK support works very well already.

The idea isn't to trim features—our focus is simply on different things. If you submit a PR with a concrete CJK fix, we'll be happy to accept it. But if, for example, adding CJK UI strings costs 50 KB of firmware space, that's a trade-off worth discussing beforehand...

brianhuster: > I believe CJK support works very well already.

In my opinion, 2 things missing from CJK support are
- Fallback to SD fonts for CJK characters in file names, table of contents https://github.com/crosspoint-reader/crosspoint-reader/pull/2521: As built-in fonts doesn't support CJK (and probably never support), CJK file names and table of contents (TOC) are just blank. Even though renaming files to English/pinyin/romaji/romaja is trivial, editing TOC is much harder, I don't think most users can do that. So right now TOC is just unusable when reading CJK file names
- Ruby characters support: [I thought Crosspoint supports a simple from of `ruby` handling](https://github.com/crosspoint-reader/crosspoint-reader/discussions/1757#discussioncomment-17559632), but I was wrong. It can only render ruby characters in a bit readable form if EPUB creators use `<rp>` tag. But even `<rp>` tag is discouraged by [EPUB 3 standard](https://www.w3.org/TR/epub-33/#sec-xhtml-deviations-rp) which seems to believe all EPUB3 readers already have proper ruby support. As a result, many modern books don't use `<rp>` tag, and Crosspoint can't render them. Even rendering ruby characters inline inside parenthesis like Crosspoint does when there is `<rp>（</rp> ... <rp>）</rp>` could still make the text look messy and unreadable, e.g 漢（かん）字（じ）は難（むず）かしいですが、面（おも）白（しろ）いです, hence I think Crosspoint should support proper ruby handling, which only increases 4kB in flash usage

Uri-Tauber: Those two points seems valid to me! I will try to get #2521 in before release 1.5.0.
Can you explain more what are ruby characters, and what the problem with displaying them? I have absolutely zero knowledge in CJK typography.

brianhuster: I think this article explains well what ruby characters are, it includes the code to display them, and how they look visually on a screen https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-ruby-element

If there are anything you want to know more, please ask me in https://github.com/crosspoint-reader/crosspoint-reader/pull/2665 (sorry for not including link to that PR in my previous comment)

szetszho: My two cents on CJK support: I completely understand the language barrier developers face when maintaining language support in an open-source project, especially with workarounds and forks already trying to fill that gap.

That said, I'd still support adding basic CJK support (or other languages, if there are contributors for them) to Crosspoint. I'm a bilingual user myself and consume content in both languages, so I understand that most users come to Crosspoint as native English speakers, but there's also a sizable CJK-bilingual user base out there. Crosspoint's feature set is already one of the most well-received in this space — I think CJK users would really appreciate the chance to use it too.

itsthisjustin:
Github isn't letting me reply to these comments so I'm going to try over email. This scope language was left over from the prior update and was conflicting with another stance I stock I wrote. I've since updated it directly on the develop branch if you can take a look at that

PR #2151, opened 2026-05-26
refactor: replace bare new with nothrow across 38 sites

jeremydk:
Effectively constrains our current codebase to the standards laid out in the claude.md for agent use + guidelines in code comments for humans.

Mechanically refactored.

With -fno-exceptions on ESP32, bare `new` calls abort() on OOM instead of returning nullptr, so any null-check after `ptr.reset(new T(...))` was dead code and any OOM produced a mystery watchdog reset with no log line, or could raise an abort() screen.

Convert every fallible allocation to either makeUniqueNoThrow<T>(...) or `new (std::nothrow) T(...)` with a guarded null-check that logs `[TAG] OOM: <type>` and degrades on the relevant failure path rather than the current behavior: abort():

- Deserialize (Page, PageLine, PageImage, ImageBlock, TextBlock) now propagates nullptr; Page::deserialize null-checks the children.
- ChapterHtmlSlimParser's six Page/ParsedText allocs early-return on OOM and unify the previously inconsistent nothrow/bare-new mix.
- Epub::load, Xtc::load, and the Section/Epub/Xtc/Txt loaders in EpubReaderActivity/ReaderActivity return false/nullptr cleanly.
- ImageDecoderFactory returns nullptr on decoder alloc failure; image rendering already handles that by skipping the image.
- JPEGDEC/PNGdec file-handle callbacks construct HalFile with makeUniqueNoThrow then .release() ownership across the void* boundary.
- Bitmap.cpp + PngToBmpConverter's per-image ditherer allocs gain a full cleanup cascade with a new BmpReaderError::OomDitherer.
- BitmapHelpers.h's three ditherer classes use std::unique_ptr<int16_t[]> with inline `new (std::nothrow)` (the header is included across libs, so it can't depend on lib/Memory). Adds isValid() so callers can detect partial-construction OOM; destructor is safe regardless.
- Activities allocating WebServer/DNSServer/WebSocketsServer/ CrossPointWebServer on OOM call onGoHome() to exit cleanly (which takes the existing silentRestart path).
- WebDAVHandler is the one ownership-transfer site: WebServer continues without the WebDAV endpoints if alloc fails, instead of aborting.

Build: pio run -e default + -e gh_release both green. DRAM usage unchanged (51.35%), Flash ~88 bytes smaller (cold paths only).

Validated via some navigation on my x4; no OOM lines thrown, but also, we haven't had a reliable heap crash repro.

Did you use AI tools to help write this code? partial, claude, me, comby and ast-grep

### DIFF
--- a/lib/Epub/Epub.cpp
+++ b/lib/Epub/Epub.cpp
@@ -4,6 +4,7 @@
 #include <HalStorage.h>
 #include <JpegToBmpConverter.h>
 #include <Logging.h>
+#include <Memory.h>
 #include <PngToBmpConverter.h>
 #include <ZipFile.h>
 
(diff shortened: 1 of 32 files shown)


jeremydk:
Threw this back into draft.  This is exposing great data, but it's exposing other crashes at different points (now that I can see them), so I'm just pulling the thread to see where it goes.

jeremydk:
Project for a later date.

PR #2157, opened 2026-05-26
fix: handle SD font advance cache overflow for CJK

WuSandWitch:
# Summary

- SD card font's persistent advance cache silently dropped codepoints past its 768-entry cap, making `getAdvance()` return 0 for the missing chars. The renderer fast-path summed these zeros into wildly wrong word widths, which fed Knuth-Plass line-breaking and broke CJK layouts after ~2 chapters of reading.
- Add a per-call overflow table for cap-exceeding codepoints, plus a last-resort single-glyph SD read as a safety net. Persistent cache size unchanged — Latin-script books pay no baseline memory cost.

## Background

I started reading Chinese EPUBs after 1.3.0 shipped SD card font support — loaded a CJK font via the SD-card font flow and immediately ran into irregular line-break breakage that got worse the further I read:

- single characters orphaned on their own line
- characters overflowing the right margin
- characters overlapping each other

Symptoms became frequent (almost every line) by chapter 2-3. Triggered specifically by SD card fonts; built-in fonts unaffected.

## Root cause

`SdCardFont::buildAdvanceTable()` maintains a persistent `advanceTable_` per style, capped at `ADVANCE_CACHE_LIMIT = 768`. Two compounding gates silently dropped codepoints once full:

1. `fetchAdvancesForCodepoints()` did `if (advanceTableSize_[si] >= ADVANCE_CACHE_LIMIT) continue;` — never even attempted SD reads for new codepoints once the cap was hit.
2. `mergeIntoAdvanceTable()` tail-truncates merges, so any high codepoints that did make it through got dropped during merge.

The cap was Latin-sized — typical Chinese novels exceed 768 unique codepoints within the first two chapters. Once exceeded:

`getAdvance(cp)` → `0` (silent miss) → `getTextAdvanceX(word)` sums to underestimated word width → `computeLineBreaks()` runs Knuth-Plass DP on wrong widths → orphans / overflow / overlaps.

## Fix

Three-tier lookup. Persistent cache size unchanged so Latin-script books incur no baseline memory regression:

| Tier | Source | Lifetime |
|------|--------|----------|
| 1 | `advanceTable_` (existing, 768 cap) | font lifetime |
| 2 | `overflowAdvance_` (new) | one `fetchAdvancesForCodepoints` call (per-paragraph), replaced on each call |
| 3 | `readSingleAdvanceFromSd()` (new) | none — single SD read on demand |

`fetchAdvancesForCodepoints()` no longer skips when the cap is hit; it batch-reads everything needed and splits the result between persistent (fills first, preserving the original "early codepoints stay cached" intent) and overflow (per-call leftover). `getAdvance()` chains the three tiers. `hasAdvanceTable()` also considers overflow so the renderer fast-path stays enabled if persistent allocation ever fails.

`getAdvance()` stays `const` — the Tier 3 SD read opens a file through `HalStorage` but does not mutate `SdCardFont` state.

## Memory impact

- **Persistent:** unchanged, ≤24 KB across all four styles (lazy-allocated per style).
- **Overflow:** exact-sized per call; typically <500 B per style for a Chinese paragraph after the persistent cache fills. Bounded by `MAX_UNIQUE_CODEPOINTS = 4096` worst case.
- **Tier 3:** zero — single SD read, no caching.

## Test plan

- Host gtest suites pass (`ctest --test-dir build/test --output-on-failure`) — 79/79.
- `pio run -e default` builds clean.
- Read affected Chinese EPUB on real hardware (X4) with a CJK SD font through past chapter 3:
  - Right-margin overflow gone.
  - Overlapping characters gone.
  - Orphan characters still present — separate issue, see below.

## Out of scope (follow-up work for separate PRs)

Noticed during testing but unrelated to this advance-table bug:

1. **Orphan characters in CJK layout** — single Chinese chars still occasionally get pushed to the start of a line. Likely a `ParsedText::computeLineBreaks()` issue around CJK break rules — Chinese has no inter-word spaces and almost every char is a legal break point, so Knuth-Plass needs different penalty weights / forbidden-break-before rules than the Latin defaults. Independent of advance widths (orphans persist even with correct widths).
2. **UI text doesn't pick up the SD card font** — chapter-select dialog, status bar title, and home screen book title / author still render through the built-in font path. Chinese titles in those surfaces show as missing glyphs even when a CJK SD font is loaded for the reader. These code paths don't go through the SD font registry and would need a separate plumbing change to route UI strings through the active SD font.

### DIFF
--- a/lib/EpdFont/SdCardFont.cpp
+++ b/lib/EpdFont/SdCardFont.cpp
@@ -965,6 +965,9 @@ void SdCardFont::clearPersistentCache() {
     delete[] advanceTable_[i];
     advanceTable_[i] = nullptr;
     advanceTableSize_[i] = 0;
+    delete[] overflowAdvance_[i];
+    overflowAdvance_[i] = nullptr;
+    overflowAdvanceSize_[i] = 0;
   }
 }
 
@@ -1021,19 +1024,10 @@ void SdCardFont::mergeIntoAdvanceTable(uint8_t styleIdx, const AdvanceEntry* sor
   advanceTableSize_[styleIdx] = k;
 }
 
-bool SdCardFont::hasAdvanceTable() const {
-  for (uint8_t i = 0; i < MAX_STYLES; i++) {
-    if (advanceTable_[i]) return true;
-  }
-  return false;
-}
-
-uint16_t SdCardFont::getAdvance(uint32_t codepoint, uint8_t style) const {
-  style &= (MAX_STYLES - 1);
-  if (!advanceTable_[style]) return 0;
-  const AdvanceEntry* table = advanceTable_[style];
-  const uint32_t size = advanceTableSize_[style];
-  // Binary search sorted by codepoint
+bool SdCardFont::overflowAdvanceLookup(uint8_t styleIdx, uint32_t codepoint, uint16_t* outAdvance) const {
+  const AdvanceEntry* table = overflowAdvance_[styleIdx];
+  const uint32_t size = overflowAdvanceSize_[styleIdx];
+  if (!table || size == 0) return false;
   uint32_t lo = 0, hi = size;
   while (lo < hi) {
     uint32_t mid = lo + (hi - lo) / 2;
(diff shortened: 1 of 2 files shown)


WuSandWitch:
### Quick follow-up on the "Out of scope" section above:
Digging into this more, I realized 1.3.0 doesn't actually claim CJK support — the SD-card font flow lets users load a CJK font, but the rendering pipeline and UI surfaces weren't designed around CJK. So strictly speaking what this PR fixes is "SD font advance table breaks under high unique-codepoint density", not "CJK reading".

That said, there's already a population of Chinese-reading users in practice (myself included) using the SD-card font flow as a workaround, and most of us can navigate an English UI fine. So I think the patches in this PR could plausibly be the foundation for a minimum-viable CJK support tier without the project taking on full localization scope. The minimum useful surface is just the three places where user-provided content shows up:

  1. Reader body text renders correctly — this PR.
  2. Chapter select dialog — show Chinese chapter titles.
  3. Home screen / Browse Files — show Chinese book titles and authors.

Those three together cover the end-to-end reading flow: pick book (see title/author) → open → pick chapter → read. Settings, system menus, status bar, etc. can stay English — Chinese users or maybe CJK users can read English UI; what we actually need is for our content to display correctly.

Happy to take a stab at (2) and (3) as separate PRs after this lands, if you'd be open to scoping CJK support that way.

brianhuster:
Could you switch the target branch to `develop` please?

Uri-Tauber:
@brianhuster Can you check if it's still relevant?
Several CJK fixes have landed over the past two months.

brianhuster:
I think it could be closed for now. I made an epub file with more than 2000 unique Kanji but don't see any overlapping

Uri-Tauber:
Closing as stale.

PR #2161, opened 2026-05-26
fix: BOOK_CACHE_VERSION jump

Uri-Tauber:
## Summary

## Additional Context

  specific areas to focus on).

---

### AI Usage

### DIFF
--- a/lib/Epub/Epub/BookMetadataCache.cpp
+++ b/lib/Epub/Epub/BookMetadataCache.cpp
@@ -9,7 +9,7 @@
 #include "FsHelpers.h"
 
 namespace {
-constexpr uint8_t BOOK_CACHE_VERSION = 9;
+constexpr uint8_t BOOK_CACHE_VERSION = 6;
 constexpr char bookBinFile[] = "/book.bin";
 constexpr char tmpSpineBinFile[] = "/spine.bin.tmp";
 constexpr char tmpTocBinFile[] = "/toc.bin.tmp";


lib/Epub/Epub/BookMetadataCache.cpp
```diff
@@ -9,7 +9,7 @@
 #include "FsHelpers.h"
 
 namespace {
-constexpr uint8_t BOOK_CACHE_VERSION = 9;
+constexpr uint8_t BOOK_CACHE_VERSION = 6;
```
znelson, line 12: I actually don't think anything in BookMetadataCache was affected by the sup/sub change. Should we just go back to version 5?

Uri-Tauber: I **Think** it's needed because `enum Style` got two new fields.

znelson: Worth double checking. Bumping the book metadata cache isn't the worst thing, but I think it doesn't need to be invalidated and will just cause devices to rebuild the same book cache.

We do need to invalidate the CSS cache, because now we're looking for new style properties. c5e861d71c14c90419676049ffd63d7e8586fc23 bumped `CSS_CACHE_VERSION` from 4 to 5.

We also need to invalidate the section caches, because lines of text will potentially lay out differently. I believe the CSS cache change automatically invalidates the section cache. So not necessary to bump `SECTION_FILE_VERSION`.

znelson (approved):

PR #2162, opened 2026-05-26
perf: optimise `normalisePath`

Uri-Tauber:
## Summary

**perf: optimise `normalisePath` to eliminate intermediate allocations**

Replaces character-by-character string building with index-based parsing using `string_view`, avoiding a heap allocation per path segment. Also pre-reserves vector capacity (8 slots) and the final result string to reduce reallocations.

**Changes:**
- `vector<string>` → `vector<string_view>`: components are now zero-copy views into the input string
- Single loop handles both `/` delimiters and end-of-string, removing the post-loop special case
- Result string capacity pre-calculated before assembly

---

### AI Usage

### DIFF
--- a/lib/FsHelpers/FsHelpers.cpp
+++ b/lib/FsHelpers/FsHelpers.cpp
@@ -8,12 +8,14 @@
 namespace FsHelpers {
 
 std::string normalisePath(const std::string& path) {
-  std::vector<std::string> components;
-  std::string component;
+  std::vector<std::string_view> components;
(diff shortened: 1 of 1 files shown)


itsthisjustin:
Without taking the time to trace the path here, what does this impact on device? Like does it improve epub reading speed or what metric

Uri-Tauber:
@itsthisjustin Just eliminate few heap allocations. It's part of series of PRs I want to introduce to solve the heap fragmentation. One byte at a time.

itsthisjustin (dismissed):

PR #2166, opened 2026-05-27
feat: add setting to disable screenshot shortcut (#controls)

danielweis:
## Summary

* **What is the goal of this PR?** Adds a user-facing toggle to disable the Power + Down screenshot shortcut, for users who accidentally trigger it while reading.
  * `CrossPointSettings`: new `screenshotShortcutEnabled` field (`uint8_t`, default `1`)
  * `SettingsList`: toggle entry under the **Controls** category (`"screenshotShortcutEnabled"` key, available in both device UI and web settings API)
  * `main.cpp`: screenshot combo detection and the "don't sleep while combo is pressed" guard are both gated on the setting
  * `lib/I18n/translations/*.yaml`: new `STR_SCREENSHOT_SHORTCUT: "Screenshot Shortcut"` string (I wrote the English one and asked Claude to generate the other languages)

## Additional Context

* Defaults to enabled (`1`), so existing devices are unaffected — no migration needed.
* The sleep-guard path (`if (gpio.isPressed(BTN_DOWN)) return;`) is also gated, so disabling the shortcut restores normal Power-button sleep behaviour when Down is held simultaneously.

---

### AI Usage


danielweis:
@itsthisjustin gentle bump on this PR and my other one - apologies for the ping but I'm not sure what the right etiquette is. Happy to take feedback to contribute, thanks!

itsthisjustin:
> @itsthisjustin gentle bump on this PR and my other one - apologies for the ping but I'm not sure what the right etiquette is. Happy to take feedback to contribute, thanks!

No appreciate the bump! I had not seen this. Request review is the best way to get attention here. Let me circulate this with the team

danielweis:
Thanks - I didn't see a way to request a review but I'll try to figure that out for my other PR.

Uri-Tauber (approved):
The code is minimal and clean. LGTM.

One question, though: do we want a dedicated toggle for each half-action in the firmware?
This feels more like a UX consideration, so I'll defer to @uxjulia.

uxjulia:
> The code is minimal and clean. LGTM.
> 
> 
> 
> One question, though: do we want a dedicated toggle for each half-action in the firmware?
> 
> This feels more like a UX consideration, so I'll defer to @uxjulia.
> 
> 

This is exactly why I have reservations around adding very specific options for toggling things that seem really niche. I don't know how often people are accidentally pressing both the power button and the volume down button at the same time, it seems like an infrequent edge case to me. I'd personally want more evidence that this is happening to a larger number of users before adding this. Before you know it we'll have requests to toggle _everything_.

danielweis:
Thanks for the thoughtful comments!

@uxjulia perhaps a better way to orient this is "should this feature exist at all"? If it's considered so niche, can we instead remove it?

Issue #2170, opened 2026-05-27
Crash when opening an epub "createSpineEntry called but not in build mode"


vedi0boy:
### Affected Version

master

### Describe the Bug

I get the following crash when trying to open the [attached epub](https://github.com/user-attachments/files/28309001/Bible.zip):

This is a large epub file but I have been able to open it before. This only started this morning when I pulled the last 25 commits (between 38210be and bbb3e06) into my branch and now I can't load it anymore.

### Steps to Reproduce

1. Open the attached epub, either from recents or from files
2. Takes a few seconds with nothing happening. Buttons are unresponsive
3. X4 crashes

### Expected Behavior

Load the epub as before.

### Relevant Log Output/Screenshots

```shell
CrossPoint version: 1.3.0-dev-feature/bookmarks-962d881

Panic reason: abort() was called at PC 0x4216f8e9 on core 0

Last logs:
[22945] [DBG] [BMC] createSpineEntry called but not in build mode
[22945] [DBG] [BMC] createSpineEntry called but not in build mode
[22945] [DBG] [BMC] createSpineEntry called but not in build mode
[22945] [DBG] [BMC] createSpineEntry called but not in build mode
[22946] [DBG] [BMC] createSpineEntry called but not in build mode
[22946] [DBG] [BMC] createSpineEntry called but not in build mode
[22947] [DBG] [BMC] createSpineEntry called but not in build mode
[22947] [DBG] [BMC] createSpineEntry called but not in build mode
[22948] [DBG] [BMC] createSpineEntry called but not in build mode
[22948] [DBG] [COF] Entering guide state.
[22963] [DBG] [COF] Found cover reference in guide: OEBPS/cover.xhtml
[22964] [DBG] [ZIP] Decompressed 39424 bytes into 523917 bytes
[22964] [DBG] [EBP] Successfully parsed content.opf
[1] [DBG] [UI] Using Lyra theme
[275] [INF] [HW] Using cached device type: X4
[275] [INF] [MAIN] Hardware detect: X4


Stack memory:
0x3FCB1430: 0x00000001 0x00000000 0x3FCB1448 0x4038DED4 0x00000000 0x00000000 0x3FCA0030 0x3FC94210 
0x3FCB1450: 0x3FCB145C 0x3FC9422C 0x3FCB1448 0x36313234 0x39653866 0x00000000 0x726F6261 0x20292874 
[...]

PR #2172, opened 2026-05-27
fix: Fix OOM Crashes on Large EPUBs during CSS Discovery #2170

Uri-Tauber:
## Summary

**The Issue**
Opening large EPUBs with a stale CSS cache caused memory exhaustion and application crashes (`abort()`) due to two compounding inefficiencies:

* **Unnecessary Spine Processing:** When parsing `content.opf` solely to rebuild the CSS cache, the system was unnecessarily processing spine entries. This resulted in wasted CPU cycles and significant memory drain from building in-memory item indices, resolving lookups, and allocating strings for thousands of items.
* **Heavy Heap Allocations in ZIP Scanning:** The `discoverCssFilesFromZip()` method used `loadAllFileStatSlims()`, which loaded every ZIP central directory entry into heap memory at once. For large books with thousands of files, this allocated 200KB+ of RAM, ultimately triggering an out-of-memory abort.

**The Fix**
This PR optimizes the CSS parsing path to be entirely memory-efficient:

* **Bypass OPF Spine Processing:** Added a `skipSpine` parameter to `parseContentOpf()`. When scanning for CSS, the method now passes a `nullptr` cache pointer to the parser, instructing it to completely skip spine work (avoiding temp files, indices, and string allocations).
* **Zero-Allocation ZIP Streaming:** Implemented a new `streamFilePaths()` method in the `ZipFile` class. This replaces the heap-heavy directory loading by streaming ZIP central directory entries sequentially, utilizing only a lightweight 256-byte stack buffer.

---

### AI Usage

### DIFF
--- a/lib/ZipFile/ZipFile.h
+++ b/lib/ZipFile/ZipFile.h
@@ -77,4 +77,50 @@ class ZipFile {
       callback(std::string_view{entry.first});
     }
   }
+
+  // Stream file paths from the ZIP central directory without caching all entries in memory.
+  // Calls callback(std::string_view) for each file path found. Returns false on open/read failure.
+  template <typename F>
+  bool streamFilePaths(F&& callback) {
+    if (!isOpen()) {
+      if (!open()) return false;
+      bool result = streamFilePathsImpl(std::forward<F>(callback));
+      close();
+      return result;
+    }
+    return streamFilePathsImpl(std::forward<F>(callback));
+  }
+
+ private:
+  template <typename F>
+  bool streamFilePathsImpl(F&& callback) {
+    if (!loadZipDetails()) return false;
+
+    file.seek(zipDetails.centralDirOffset);
+
+    uint32_t sig;
+    char itemName[256];
+
+    while (file.available()) {
+      if (file.read(&sig, 4) != 4 || sig != 0x02014b50) break;
+
+      file.seekCur(22);
+      uint16_t nameLen, m, k;
+      file.read(&nameLen, 2);
+      file.read(&m, 2);
+      file.read(&k, 2);
+      file.seekCur(12);
+
+      if (nameLen < sizeof(itemName)) {
+        file.read(itemName, nameLen);
+        itemName[nameLen] = '\0';
+        callback(std::string_view{itemName, nameLen});
+      } else {
+        file.seekCur(nameLen);
+      }
+
+      file.seekCur(m + k);
+    }
+    return true;
+  }
 };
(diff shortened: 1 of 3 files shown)


Uri-Tauber:
Fixes #2170. Tested with the problematic epub, and it was loaded successfully.
I did noted in this epub each page turn would take ~2 seconds. Not sure why, but will investigate.

vedi0boy:
> Fixes #2170. Tested with the problematic epub, and it was loaded successfully. I did noted in this epub each page turn would take ~2 seconds. Not sure why, but will investigate.

It doesn't do that for me. However I know the first few pages are split into several chapters, so maybe it's indexing on each page turn. Once I am in a chapter, the page turns are responsive for me and work as before. Thanks for your work!

lib/ZipFile/ZipFile.h
```diff
@@ -77,4 +77,50 @@ class ZipFile {
...
+
+    file.seek(zipDetails.centralDirOffset);
+
+    uint32_t sig;
+    char itemName[256];
+
+    while (file.available()) {
+      if (file.read(&sig, 4) != 4 || sig != 0x02014b50) break;
+
+      file.seekCur(22);
```
znelson: Should this be 
```suggestion
      file.seekCur(24);
```

In `ZipFile::loadAllFileStatSlims`, `ZipFile::loadFileStatSlim`, and ZipFile::fillUncompressedSizes it seeks forward `6`, then reads `2` for method, then forward `8`, then reads `8` for compressed and uncompressed size. 6+2+8+8 = 24.

Uri-Tauber: Hmmm. it seems you right. Interesting the reader didn't crash or produce garbage to the serial monitor...

znelson (approved):
My understanding is that this only happens when we have a valid book/spine cache, but not a valid CSS cache, which would happen precisely now after we've recently bumped the CSS cache version. And it's a persistent crash, not a transient one, meaning trying again is likely to hit the same thing repeatedly, unless you manually clear the book's cache from the SD card (since you can't get to the book menu inside the reader activity to clear the cache).

This looks OK for a quick fix of the immediate crash happening on this book, but I think there's a larger problem here that this is pointing to. ContentOpfParser is one monolithic state machine that does several different jobs in one pass: manifest cataloguing, spine building, CSS collection, TOC extraction, cover discovery, guide parsing. Its callers want different subsets of that work depending on context, but the parser has no notion of mode. Its callers want different subsets of that work depending on context, but the parser has no notion of mode.

In another change we could add some sort of explicit ParseMode flags passed to ContentOpfParser to specify the subset of work it should perform. And/or persist CSS file hrefs inside of BookMetadataCache, so we don't need to dive into ContentOpfParser again when the CSS cache rules change.

Uri-Tauber:
suppressed by #2213

PR #2179, opened 2026-05-28
feat(reader): Fast Mode: ~6x faster perceived page turns

jeremydk:
Adds an opt-in Fast Mode to the EPUB reader. BW + 64×64 blue-noise ordered dither replaces the 4-level grayscale post-pass. Perceived page turn drops from ~77 ms to ~13 ms on X4. The legacy 4-level grayscale path stays in place behind a new Text Anti-Aliasing toggle (default on).

Two perf wins ride along:

- Next-page prerender (Fast Mode only). Idle time rasterizes the next page's BW content into the framebuffer; the next forward turn skips the SD load + BW raster and goes straight to flush.
- Dedicated input-poll task at 200 Hz. Press/release edges land in atomic bitmasks that the main loop reads non-destructively, so presses survive long render passes that previously dropped
them.

Settings

Two new toggles in Settings → Reader, mutually exclusive in the menu (toggling one clears the other; the renderer can't usefully run both):

- textAntiAliasing — 4-level grayscale text path. Default on.
- fastMode — BW + blue-noise dither.

AI Usage: Partial


Uri-Tauber:
I tested this on my device with **Fast Mode ON**. Visually, the page turns don't feel noticeably faster yet, so I grabbed some logs from the debugging monitor to compare `master` against this PR.

There are two notable takeaways from the data:

1. **The Good:** `Page render` is consistently **~65–70ms faster** on this PR, which is very impressive.
2. **The Concern:** The time from `clearScreen` to `displayBuffer` has spiked significantly, which I'n not sure what that's means.

Here is the log breakdown:

## Page 6

**`master`**

```bash
[16:53:16]   Writing frame buffer to BW RAM (48000 bytes)...
[16:53:16]   BW RAM write complete (12 ms)
[16:53:16]   Powering on display 0xDF (fast refresh)...
[16:53:16]   Waiting for display refresh...
[16:53:17]   Wait complete: fast (639 ms)
[16:53:17]   Writing frame buffer to RED RAM (48000 bytes)...
[16:53:17]   RED RAM write complete (12 ms)
[16:53:17] [DBG] [ERS] Page render: prewarm=32ms bw_render=48ms display=665ms total=745ms
[16:53:17] [DBG] [ERS] Rendered page in 747ms
[16:53:17] [DBG] [ERS] Progress saved: spine=5 page=6
[16:53:18] [DBG] [FCM] prewarmCache(SD): 1 glyph(s) not found (styleMask=0x05)
[16:53:18] [DBG] [GFX] Time = 118 ms from clearScreen to displayBuffer

```

**This PR**

```bash
[16:57:20]   Writing frame buffer to BW RAM (48000 bytes)...
[16:57:20]   BW RAM write complete (13 ms)
[16:57:20]   Powering on display 0xDF (fast refresh)...
[16:57:20]   Waiting for display refresh...
[16:57:21]   Wait complete: fast (639 ms)
[16:57:21]   Writing frame buffer to RED RAM (48000 bytes)...
[16:57:21]   RED RAM write complete (12 ms)
[16:57:21] [DBG] [ERS] Page render: prewarm=0ms bw_render=16ms display=664ms total=680ms
[16:57:21] [DBG] [ERS] Displayed pre-rendered page in 681ms
[16:57:21] [DBG] [FCM] prewarmCache(SD): 1 glyph(s) not found (styleMask=0x05)
[16:57:21] [DBG] [ERS] Pre-rendered page 7/51
[16:57:21] [DBG] [ERS] Progress saved: spine=5 page=6
[16:57:23] [DBG] [GFX] Time = 1662 ms from clearScreen to displayBuffer

```

---

## Page 5

**`master`**

```bash
[16:53:14]   Writing frame buffer to BW RAM (48000 bytes)...
[16:53:14]   BW RAM write complete (13 ms)
[16:53:14]   Powering on display 0xDF (fast refresh)...
[16:53:14]   Waiting for display refresh...
[16:53:15]   Wait complete: fast (638 ms)
[16:53:15]   Writing frame buffer to RED RAM (48000 bytes)...
[16:53:15]   RED RAM write complete (12 ms)
[16:53:15] [DBG] [ERS] Page render: prewarm=33ms bw_render=54ms display=664ms total=751ms
[16:53:15] [DBG] [ERS] Rendered page in 753ms
[16:53:15] [DBG] [ERS] Progress saved: spine=5 page=5
[16:53:16] [DBG] [FCM] prewarmCache(SD): 1 glyph(s) not found (styleMask=0x05)
[16:53:16] [DBG] [GFX] Time = 112 ms from clearScreen to displayBuffer

```

**This PR**

```bash
[16:57:11]   Writing frame buffer to BW RAM (48000 bytes)...
[16:57:11]   BW RAM write complete (13 ms)
[16:57:11]   Powering on display 0xDF (fast refresh)...
[16:57:11]   Waiting for display refresh...
[16:57:12]   Wait complete: fast (638 ms)
[16:57:12]   Writing frame buffer to RED RAM (48000 bytes)...
[16:57:12]   RED RAM write complete (12 ms)
[16:57:12] [DBG] [ERS] Page render: prewarm=0ms bw_render=16ms display=664ms total=680ms
[16:57:12] [DBG] [ERS] Displayed pre-rendered page in 681ms
[16:57:12] [DBG] [FCM] prewarmCache(SD): 1 glyph(s) not found (styleMask=0x05)
[16:57:12] [DBG] [ERS] Pre-rendered page 6/51
[16:57:12] [DBG] [ERS] Progress saved: spine=5 page=5
[16:57:14] [DBG] [PWR] Going to low-power mode
[16:57:15] [INF] [MEM] Free: 128108 bytes, Total: 228072 bytes, Min Free: 110896 bytes, MaxAlloc: 114676 bytes
[16:57:20] [DBG] [PWR] Restoring normal CPU frequency
[16:57:20] [DBG] [GFX] Time = 8330 ms from clearScreen to displayBuffer

```

uxjulia:
From a UX standpoint, I have concerns about the setting just being called "Fast Mode" since this isn't very clear what this will do. It'd be a little more self-explanatory if the setting was something like:

Text Anti-aliasing: ON / OFF / Fast Mode
Images: Display / Placeholder / Suppress / Fast Mode

Not sure if it's possible to split this feature out like that, but if not, toggling one of those to "Fast Mode" could auto-toggle the other. Changing the settings to something like that is an easy way to tell the user "Fast Mode" is going to affect how text anti-aliasing and your images are rendered.

Other than that, I tested this on device and it appears to be working as intended.

pablohc:
> Text Anti-aliasing: ON / OFF / Fast 

Maybe it's time to call a spade a spade. Something like "Quality Mode: Low / High / Fast"
The term "Anti-aliasing" loses some of its meaning when it's set to "OFF" or "Fast Mode"

itsthisjustin:
I'm personally in favor of simply removing AA as a setting and calling this Reader Mode or something. I'd rather not have two settings and this is so much better than current AA

@uri-tauber are you reading RTL? Due to the way this works for precaching pages going left takes slightly more time than going right

Uri-Tauber:
@itsthisjustin Actually I haven't tested yet with RTL text. I read today a few chapters from normal english book. The measurements I took are from there.
If you could do your own measurements, it would be beneficial.

jeremydk:
pageturn_bench

jeremydk:
These metrics contain the display refresh timing itself; I need to rewire the sub-refresh timing to get the perceived latencies, but the numbers are clear.

edit: What I mean by this is that the next page is visible before the display refresh timer is complete. _that_ is the time which is most dramatically improved.

uxjulia:
I think the path of least resistance right now without having to redefine what things mean (which could happen down the line if needed) is to put this setting in existing settings that users already understand conceptually, and since it's mutually exclusive to AA, it should live inside that setting:

Text Anti-aliasing: ON / OFF / Fast Mode
Images: Display / Placeholder / Suppress / Fast Mode

If "Fast Mode" = your text and your images will render differently, then toggling one automatically toggles the other.

jeremydk:
> I think the path of least resistance right now without having to redefine what things mean (which could happen down the line if needed) is to put this setting in existing settings that users already understand conceptually, and since it's mutually exclusive to AA, it should live inside that setting:
> 
> Text Anti-aliasing: ON / OFF / Fast Mode Images: Display / Placeholder / Suppress / Fast Mode
> 
> If "Fast Mode" = your text and your images will render differently, then toggling one automatically toggles the other.

I don't agree with splitting fast mode between text and images, it would be an obnoxious amount of rework with a much worse outcome for the user.

uxjulia:
> > I think the path of least resistance right now without having to redefine what things mean (which could happen down the line if needed) is to put this setting in existing settings that users already understand conceptually, and since it's mutually exclusive to AA, it should live inside that setting:
> > Text Anti-aliasing: ON / OFF / Fast Mode Images: Display / Placeholder / Suppress / Fast Mode
> > If "Fast Mode" = your text and your images will render differently, then toggling one automatically toggles the other.
> 
> I don't agree with splitting fast mode between text and images, it would be an obnoxious amount of rework with a much worse outcome for the user.

I wasn't telling you to split what the code does. My proposal was to let the settings itself explain what the code does to the user. The idea is to have the UI itself be the documentation/readme. If they choose "Fast Mode" for images, that will automatically set text anti-aliasing to also be "Fast Mode", since "Fast Mode" will affect both text anti-aliasing and image rendering (and yes, I understand under the hood that's not all it does, but user-facing, those are the major things people will notice). It's the same concept you had when you had Text Anti-Aliasing automatically set to "OFF" when Fast Mode = "ON". It's not some huge lift for what you've already implemented. It's a change to how the settings are toggled.

itsthisjustin:
So understanding your goal Julia my counter to this is that text is a first class citizen on these devices not images. Imo we should have EPUB Rendering as a setting and have be AA, B&W, and Fast since this only effects epubs anyway (I believe) and doesn't impact xtc which would have images so that images setting isn't really accurate

uxjulia:
> So understanding your goal Julia my counter to this is that text is a first class citizen on these devices not images. Imo we should have EPUB Rendering as a setting and have be AA, B&W, and Fast since this only effects epubs anyway (I believe) and doesn't impact xtc which would have images so that images setting isn't really accurate

Sounds good to me.

EPUB Rendering: Anti-Aliased | Black & White | Fast 

@jeremydk - Do you want to make this update so we can get this merged or would you prefer someone else move this forward on your behalf?

jeremydk:
I've decided to implement this in a different way that perhaps all contributors can agree adds speed without sending me back to go collect data over serial using a harness that I couldn't get merged anyway.

PR #2180, opened 2026-05-28
chore: fix clang-tidy warnings for cppcheck replacement

Uri-Tauber:
## Summary

I explored replacing `cppcheck` with `clang-tidy`.

Although `clang-tidy` generated many false positives and low-value warnings, it also highlighted a few real issues, primarily around memory ownership and allocation.

This PR fixes those findings, mainly by replacing several raw allocation usages with smart pointers where appropriate.

---

### AI Usage

Did you use AI tools to help write this code? _**< PARTIALLY >**_ Claude used to help review the changes and check for possible regressions.


Uri-Tauber:
suppress arithOperationsOnVoidPointer warnings - this is a known cppcheck false positive (according to Claude) where it misidentifies std::unique_ptr<T[]>::get() as returning void* (it actually returns T*). There's no real void pointer arithmetic in the code.

@coderabbitai can you confirm?

lpla:
Hey @Uri-Tauber, had a quick pass here too. The RAII direction for the Home/TXT/XTC buffers looks good 👍

Main thing I noticed: `lib/GfxRenderer/BitmapHelpers.h:24-211` uses `std::make_unique`, which can abort on OOM in our `-fno-exceptions` build. We’d want `makeUniqueNoThrow` + null handling there. I’d also avoid the repo-wide cppcheck suppression in `platformio.ini:16-17`.

This one conflicts with current `develop` too. Want a hand rebasing or splitting out the safe bits?

Uri-Tauber:
@lpla I completely forgot about this one. 
Feel free to play around with it.

lpla:
I’ve fast-forwarded the existing `chore/clang-tidy` branch to `290200d82`.

I just:

- Converted the Home recent-cover snapshot buffer to checked RAII using `makeUniqueNoThrow`.
- Converted TXT page buffers to checked RAII and removed manual cleanup paths.
- Converted XTC page buffers to checked RAII while preserving the existing 1-bit and 2-bit plane layout.
- Marked `HalFile` move operations `noexcept`.
- Added the complete `RecentBook` type to `HomeActivity.h`; this fixes a portability error exposed by Apple libc++ when destroying `std::vector<RecentBook>`.
- Replaced the OTA `sscanf` comparison with an allocation-free semantic-version parser.
- Added host tests for normal releases, `v` prefixes, `1.6.0-rc+hash`, the project’s `1.6.0rc` form, development suffixes, malformed/overflowing versions, and stable-versus-RC ordering.
- Added rejection tests for incomplete suffixes such as `1.2.3-` and `1.2.3+`.
- Addressed the CI formatting and cppcheck portability findings without adding blanket suppressions.


I did not retain the old `BitmapHelpers` ditherer rewrite. It used `std::make_unique`, which is unsafe as a recoverable OOM strategy in this firmware because it is built with `-fno-exceptions`; allocation failure would abort rather than return a null pointer. I think the narrower dither work in #2332 is the better place to handle that path with `makeUniqueNoThrow`, but that's your call.

I also left out the old global cppcheck suppressions. The actual warnings have been fixed at their call sites instead.

Before approval, I’d still like hardware confirmation for:

1. Repeated Home -> Recents navigation with recent-book covers, checking that the cover snapshot is restored and released correctly.
2. TXT books with long lines and repeated page navigation.
3. Both 1-bit and 2-bit XTC files, including grayscale rendering. This is also worth checking alongside #3173.
4. OTA comparisons around RC builds: older RC, same RC, stable release superseding the matching RC, and no-update cases.
5. Serial heap logs before and after repeated entry/exit from these activities, preferably on an ESP32-C3 device.

Uri-Tauber:
@lpla I dont think `lib/Version/SemanticVersion.h` belong in this PR.

See my comment and Justin's repley in https://github.com/crosspoint-reader/crosspoint-reader/pull/2931#issuecomment-5231997650

lpla:
Agreed. I had pulled the semantic-version work into this branch while investigating the existing OTA comparison behavior.

I just removed `lib/Version/SemanticVersion.h`, the `OtaUpdater` changes, and the semantic-version tests from the resulting PR diff.

I also finished the remaining buffer review:

- Home, TXT, and XTC use a checked non-zeroing RAII allocation only where the complete allocation is guaranteed to be overwritten before reading. This avoids a redundant zero-fill; it does not reduce the allocation size.
- The XTC page buffer is reused across normal page turns to avoid repeated 48–96 KB allocation/free cycles.
- Because the reader remains alive underneath the chapter picker, that buffer is released before opening the picker and reacquired when rendering resumes.
- The chapter-picker activity itself now also uses checked nothrow allocation.

itsthisjustin:
@Uri-Tauber do you want this merged for 1.6?

Uri-Tauber:
@itsthisjustin No. Too much changes in important areas, and I haven't reviewed the changes @lpla did.

PR #2184, opened 2026-05-28
feat: Add support for Azerbaijani language

kheyyam:


Uri-Tauber:
@kheyyam Have you tested it on an actual device? Photo would be appreciated. 
CrossPoint (still) doesn't support RTL scripts and Arabic/Farsi. Does Azerbaijani have similar alphabet, or does it use latin letters?

kheyyam:
@Uri-Tauber 
1. **Testing on a device:** Unfortunately, I haven't been able to test it on a physical device just yet. My unit is currently running `crosspet` firmware, and I've run into an issue trying to flash the `crosspoint` firmware back onto it to test. Because of this, I can't provide a photo right now, though the YAML syntax itself is fully validated.

2. **Azerbaijani Alphabet:** Modern Azerbaijani uses a **Latin-based script** and is written completely **Left-to-Right (LTR)**, exactly like English, German, or Turkish. It does not use Arabic/Farsi script or require any RTL support. It does feature a few standard modified Latin characters (such as `ə`, `ç`, `ğ`, `ı`, `ö`, `ş`, `ü`), but if the system handles standard Latin fonts and basic diacritics, it will display perfectly.

Let me know if you need any adjustments to the file!

Uri-Tauber:
@kheyyam Thanks for your detailed response.

I would prefer to accept only PRs which were tested on actual hardware. Even if the firmware compiles, there probably still some adjustments needed - short some labels. Change translation so they convey the meaning better... things that harder to fix just by staring at the YAML file.

Another question is how much users will benefit from adding this translation? If even you not going to use it, we might be adding 5-6kb with zero value whatsoever. I you know of another Azerbaijani user who use CrossPoint, I would suggest asking them to run your build and verify the translation.

muhas:
Сharacters _ə Ç ə_ are missing, maybe some others
photo_2026-05-31_13-48-53

kheyyam:
@muhas Thanks for testing this and sharing the photo! 

Looking  at image the issue isn't with the translation file itself, but rather with the **system fonts** used by CrossPoint. 
The translation file is correctly encoded in UTF-8, which is why standard modified Latin characters shared with Turkish (like `ş`, `ğ`, `ö`, `ı`) are rendering flawlessly across the UI (for example, the `ı` in *faylları* and the `ö` in *göstrmk*). 
To fix this completely, the CrossPoint firmware fonts would need to be updated to a font family that includes full Azerbaijani Latin script support.

Issue #2176, opened 2026-05-27
Bookmarks always show 0% progress


leecming82:
### Affected Version

Master

### Describe the Bug

I think there may be an issue with the casting of the book percentage progress shown with bookmarks where the float gets cast as an int and always shows as 0%

https://github.com/crosspoint-reader/crosspoint-reader/blob/02573dc592b0f494ef96753201b21ab094040172/src/activities/reader/EpubReaderBookmarksActivity.cpp#L189

### Steps to Reproduce

Add bookmark,
View bookmark

### Expected Behavior

Show the correct %

### Relevant Log Output/Screenshots

```shell

```

vedi0boy:
Nice catch! I must have forgotten to pay attention to this in the last changes I made to optimize the bookmarks activity. I have opened a PR to address this and the page numbers to start from 1, not 0: #2188

Uri-Tauber:
@leecming82 Can you test #2188 to verify it solved this bug?

Uri-Tauber:
Closing as solved.

PR #2188, opened 2026-05-28
fix: bookmark percentage always 0% and page number starts at 0

vedi0boy:
## Summary

* Fix the issue described in #2176 where bookmark percentages are always 0%
* Also fixed the page number rendering to +1 because they are 0 index based, but should be rendered starting from 1

---

### AI Usage

### DIFF
--- a/src/activities/reader/EpubReaderBookmarksActivity.cpp
+++ b/src/activities/reader/EpubReaderBookmarksActivity.cpp
@@ -186,7 +186,7 @@ void EpubReaderBookmarksActivity::render(RenderLock&&) {
     auto bookmark = bookmarks.at(confirmingDelete >= DELETE_MODE_DISPLAY ? selectorIndex : index);
     auto tocIndex = epub->getTocIndexForSpineIndex(bookmark.computedSpineIndex);
     auto tocTitle = (tocIndex >= 0) ? (epub->getTocItem(tocIndex)).title : tr(STR_UNNAMED);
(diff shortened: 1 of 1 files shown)


Uri-Tauber (dismissed):
LGTM

src/activities/reader/EpubReaderBookmarksActivity.cpp
```diff
@@ -186,7 +186,8 @@ void EpubReaderBookmarksActivity::render(RenderLock&&) {
     auto bookmark = bookmarks.at(confirmingDelete >= DELETE_MODE_DISPLAY ? selectorIndex : index);
     auto tocIndex = epub->getTocIndexForSpineIndex(bookmark.computedSpineIndex);
     auto tocTitle = (tocIndex >= 0) ? (epub->getTocItem(tocIndex)).title : tr(STR_UNNAMED);
-    return std::to_string((int)bookmark.percentage) + "% - " + std::to_string(bookmark.computedChapterProgress) + "/" +
+    return std::to_string((int)(bookmark.percentage * 100)) + "% - " +
```
leecming82: How about `std::clamp(bookmark.percentage, 0.0f, 1.0f) * 100.0f + 0.5f`

At minimum, you should do +0.5f so that it rounds to the closest integer instead of always rounding down 

Otherwise LGTM

vedi0boy: Good idea! Applied.

Uri-Tauber (approved):

itsthisjustin (approved):

PR #2194, opened 2026-05-28
feat: add Spacing Modifier Letters range to font presets

### LINKED ISSUE #1980: Feature Request: Modifier Letter Prime
I believe UTF support is on the list of upcoming features. In the meantime, can we have support for "Modifier Letter Prime"? Also known as `ʹ`?

https://www.compart.com/en/unicode/U+02B9

Thank you!

itsthisjustin:
Adds Unicode range 0x02B0-0x02FF (Spacing Modifier Letters) to both 'latin_ext' and 'reading' font presets to support additional phonetic and modifier characters. Closes #1980


Uri-Tauber (commented):
Approving, but I think @mcrosson should also give it a look.

Uri-Tauber (approved):

PR #2210, opened 2026-05-29
feat: Update italian.yaml

alan0ford:
## Summary
* **What is the goal of this PR?** Improve the Italian (IT) localization.
  * Optimized several strings that were too long for the UI.
  * Reviewed and refined the newly added strings.

## Additional Context
* Changes are limited to the Italian language file; no functional/code changes.

---
### AI Usage


matteoscopel:
The changes here just undo a previous PR to fix the Italian translation and contain translations that no native speaker would consider better.

For example, you replace "tenere premuto" with "tenere spinto", which is absolutely unusual in this context. Or you replace "sfondo" with "wallpaper" (why, when every single OS uses the word "sfondo" for this?). Or you replace "compatto" with "stretto", which is again unusual in the context of typography.

Frankly, this PR just feels machine translated without much consideration for the context of the translation.

alan0ford:
I can partly agree with the criticism, but not with the claim that it's a machine translation — I gave it quite a lot of thought...

It really bothers me to see text that's truncated and partly unreadable; I prefer a translation that conveys the function even if it's not entirely orthodox. Unfortunately, some English words can only be translated correctly with much longer roundabout phrasings.

We're not like the French, who have to translate everything literally — "wallpaper" is perfectly understandable to any Italian, whereas "sfondo" makes me think of something else (but that might just be me).

For the rest, mine is meant to be just a simple contribution; I fully defer to your judgment (if need be, I'll compile my own version). Ciao. :-)

Uri-Tauber:
It seems this PR and #2207 competing here. 
@alan0ford @matteoscopel Can you agree what's better to merge?

alan0ford:
I'm sorry, I'm leaving

PR #2211, opened 2026-05-29
fix: KOReader sync drift when syncing at chapter start 

uxjulia:
## Summary

  Fix KOReader progress sync positions that land a few pages too far forward when the synced position is at the start of a chapter.

  * Detect KOReader chapter-start XPath formats like `/body/DocFragment[13].0` and pin them to the start of the resolved spine item instead of falling back to percentage-based byte mapping.
  * Track whether XPath parsing actually resolved an intra-chapter position, so `0.0` can mean “real chapter start” instead of “parser failed.”

## Additional Context

* The main issue was that chapter-start sync paths could look like valid DocFragment positions without a paragraph/text anchor. Those were previously treated as unresolved and mapped by percentage, which can be a few paragraphs ahead on small screens.
* The fix is intentionally narrow and only changes KOReader progress mapping/refinement behavior.

---

### AI Usage

### DIFF
--- a/lib/KOReaderSync/ProgressMapper.cpp
+++ b/lib/KOReaderSync/ProgressMapper.cpp
@@ -56,6 +56,64 @@ int parseTextNodeIndex(const std::string& xpath) {
   return val > 0 ? val : 1;
 }
 
+bool isChapterStartXPath(const std::string& xpath) {
+  if (xpath.find("/p[") != std::string::npos || xpath.find("/li[") != std::string::npos) {
+    return false;
+  }
+
+  static constexpr char kDocFragment[] = "/body/DocFragment[";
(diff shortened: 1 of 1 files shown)


itsthisjustin (approved):

Uri-Tauber:
I don't use Kosync, so I can't test this, but I will look at the code tomorrow.
Is this also relevant to the off-by-one-page bug with bookmarks?

Uri-Tauber:
Completely forgot about this review.
Sorry, @uxjulia. 

If this PR is still relevant, can you reopen it?

uxjulia:
Oh shoot, for some reason I thought this was merged already so I deleted my branch.

Edit: opened new PR #2245

Issue #2124, opened 2026-05-23
Ignored .css stylesheet in EPUB


vladimirwutang:
Hi everyone! When opening my EPUB books, Crosspoint completely ignores the stylesheet because it is located at a specific path: **OEBPS\css\main.css**. Is it possible to fix this in future versions by adding a feature to scan the entire EPUB structure for *.css files? For instance, my Kindle Paperwhite opens these EPUB files absolutely correctly.

vladimirwutang:
How the same EPUB looks on both readers
<img width="1280" height="1234" alt="Image" src="https://github.com/user-attachments/assets/ed454340-3dd6-4e79-9ca1-988c43a4fd42" />

[The Doomed City.zip](https://github.com/user-attachments/files/28184680/The.Doomed.City.zip)

latonis:
Looking at the epub you linked, the stylesheet `OEBPS/css/main.css` is only linked in the `book0.html` but not in the OPF manifest. The firmware doesn't parse CSS files unless they are present in the manifest, so I believe that is why. A behavior would need to be implemented to parse CSS files if they're listed in the HTML but not in the manifest

latonis:
I am working on a PR to address this :) there are a few ways we could go:
- scan zip for any css files and parse
- scan zip for any css files mentioned in html files (another pass of the html files?)
- keep the original behavior and acknowledge this epub isn't following spec (seems like most other e-readers have a fallback to allow for this)

PR #2213, opened 2026-05-30
fix: avoid zip-wide css scan for large epubs

uxjulia:
## Summary

  * This PR ports over Crossink's handling of large EPUBs that helps prevent crashes during open after the book metadata cache is built.

  * Removes the post-indexing ZIP-wide CSS discovery pass that built an in-memory map of every ZIP entry.
  * Reuses `content.opf` parsing to collect declared CSS files without writing spine entries again.
  * Temporarily releases the loaded book metadata cache while rebuilding CSS for cached books.
  * Parses CSS before reloading `book.bin` after a fresh cache build, leaving more heap available during CSS rule parsing.

## Additional Context

* User reported their EPUB opening fine on Crossink but would crash on Crosspoint. Verified this claim on my own devices.
* The crash this addresses happened after `book.bin` was successfully built, when CSS discovery allocated a large `unordered_map` for ~3k EPUB ZIP entries.
* Tradeoff: CSS files not declared in `content.opf` are no longer discovered by scanning the full ZIP. This avoids the high-risk memory allocation but improperly formatted EPUBs (ones that don't declare their CSS styles in `content.opf` will render without styling and fallback to inline styles.
* User provided epub that was crashing prior to this change: https://www.mediafire.com/file/g57ea4mj13iunvh/Quang+%C3%82m+Chi+Ngo%E1%BA%A1i+-+Nh%C4%A9+C%C4%83n.epub/file

---

### AI Usage

### DIFF
--- a/lib/Epub/Epub.cpp
+++ b/lib/Epub/Epub.cpp
@@ -459,6 +459,7 @@ bool Epub::load(const bool buildIfMissing, const bool skipLoadingCss) {
 
   if (!skipLoadingCss) {
     // Parse CSS before reloading book.bin to leave more heap for CSS rule-table growth.
+    bookMetadataCache.reset();
     parseCssFiles();
     Storage.removeDir((cachePath + "/sections").c_str());
   }
--- a/lib/Epub/Epub.cpp
+++ b/lib/Epub/Epub.cpp
@@ -256,6 +256,30 @@ bool Epub::parseTocNavFile() const {
   return true;
 }
 
+void Epub::discoverCssFilesFromZip() {
+  const std::string& opfDir = contentBasePath;
+  ZipFile zf(filepath);
+
+  if (!zf.enumerateFilePaths([&](std::string_view filePath) {
+        if (!opfDir.empty() && filePath.find(opfDir) != 0) {
+          return;
+        }
+
+        if (!FsHelpers::hasCssExtension(filePath)) {
+          return;
+        }
+
+        if (std::find(cssFiles.begin(), cssFiles.end(), filePath) != cssFiles.end()) {
+          return;
+        }
+
+        LOG_DBG("EBP", "Discovered CSS file via ZIP enumeration: %.*s", (int)filePath.size(), filePath.data());
+        cssFiles.push_back(std::string{filePath});
+      })) {
+    LOG_ERR("EBP", "Failed to enumerate ZIP file paths for CSS discovery");
+  }
+}
+
 void Epub::parseCssFiles() const {
   // Maximum CSS file size we'll attempt to parse (uncompressed)
   // Larger files risk memory exhaustion on ESP32
@@ -356,6 +380,8 @@ bool Epub::load(const bool buildIfMissing, const bool skipLoadingCss) {
         if (!parseContentOpf(cachedMetadata, /*writeSpineEntries=*/false)) {
           LOG_ERR("EBP", "Could not parse content.opf from cached bookMetadata for CSS files");
           // continue anyway - book will work without CSS and we'll still load any inline style CSS
+        } else {
+          discoverCssFilesFromZip();
         }
         bookMetadataCache.reset();
         parseCssFiles();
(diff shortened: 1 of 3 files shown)


uxjulia:
Just realized this is trying to do the same thing as @Uri-Tauber's #2172 but slightly differently. @Uri-Tauber can you review and close if it duplicates your work or grab whatever is different that might be useful?

muhas:
> Tradeoff: CSS files not declared in content.opf

 #2124 
If I understand correctly, this opens this issue again?

Uri-Tauber:
@uxjulia I will look at the code tomorrow. Can you meanwhile check your PR haven't introduce regression compared to https://github.com/crosspoint-reader/crosspoint-reader/pull/2148 - which cause this crash in the first place, and my PR try to solve it while still scanning the entire epub for .css files missing from the OPF.


EDIT: Never mind. I didn't read the entire description:

> Tradeoff: CSS files not declared in content.opf are no longer discovered by scanning the full ZIP. This avoids the high-risk memory allocation but improperly formatted EPUBs (ones that don't declare their CSS styles in content.opf will render without styling and fallback to inline styles.

uxjulia:
> @uxjulia I will look at the code tomorrow. Can you meanwhile check your PR haven't introduce regression compared to #2148 - which cause this crash in the first place, and my PR try to solve it while still scanning the entire epub for .css files missing from the OPF.
> 
> EDIT: Never mind. I didn't read the entire description:
> 
> > Tradeoff: CSS files not declared in content.opf are no longer discovered by scanning the full ZIP. This avoids the high-risk memory allocation but improperly formatted EPUBs (ones that don't declare their CSS styles in content.opf will render without styling and fallback to inline styles.

To not re-introduce the issue found in #2124, I ended up having to implement essentially what you did in your PR so the two overlap considerably. The difference now lies in the fact this PR also drops `bookMetadataCache` before `parseCssFiles()` and then reloads it afterwards to lower peak ram. That behavior also now happens during first book open and for already-cached books. If you think that's something worth keeping around, we can add it to your PR and close this one.

Uri-Tauber:
@uxjulia  Looking from above, I must say your PR have cleaner design. If you could solve the CodeRabbit warnings, I will close mine.

@znelson What do you think?

lib/Epub/Epub.cpp
```diff
@@ -256,35 +257,27 @@ bool Epub::parseTocNavFile() const {
 }
 
 void Epub::discoverCssFilesFromZip() {
-  if (!bookMetadataCache || !bookMetadataCache->isLoaded()) {
-    LOG_ERR("EBP", "Cannot discover CSS from ZIP because book metadata cache is not loaded");
-    return;
-  }
-
+  const std::string& opfDir = contentBasePath;
```
Uri-Tauber, line 260: I **think** `contentBasePath` includes the filename (e.g. OEBPS/content.opf). Strip after the last slash to get just the directory (OEBPS/). 
```suggestion
  const size_t lastSlash = contentBasePath.find_last_of('/');
  const std::string opfDir = (lastSlash != std::string::npos) ? contentBasePath.substr(0, lastSlash + 1) : "";
```

uxjulia: Hey, thanks for reviewing! So `contentBasePath` is already the OPF directory at this point. It’s set earlier in [lib/Epub/Epub.cpp:54](/Users/home/code/crosspoint-reader/lib/Epub/Epub.cpp:54) via `contentOpfFilePath.substr(0, contentOpfFilePath.find_last_of('/') + 1)`, so `OEBPS/content.opf` becomes `OEBPS/`, and `discoverCssFilesFromZip()` just reuses that cached directory, so we don't need to do it again here

Uri-Tauber: OK. Thanks for the correction. 
If you can try both epubs from #2124 and #2170 to make sure everything works correctly, It would be great.

uxjulia: Confirmed the epub in #2124 opened with CSS styles rendering:

image

And confirmed the epub in #2170 opens without crashing:

image

Uri-Tauber: Awesome.

Uri-Tauber (approved):

itsthisjustin (approved):

PR #2220, opened 2026-05-31
fix: WebDAV hidden files 

muhas:
## Summary

WebDAV ignores the system settings for showing hidden files
This PR fixes the issue 

---

### AI Usage

### DIFF
--- a/src/network/WebDAVHandler.cpp
+++ b/src/network/WebDAVHandler.cpp
@@ -5,6 +5,7 @@
 #include <Logging.h>
 #include <esp_task_wdt.h>
 
+#include "SettingsList.h"
 #include "util/BookCacheUtils.h"
 
 namespace {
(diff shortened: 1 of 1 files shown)


muhas:
This two-line PR has opened up an abyss of weirdness for me regarding how CrossPoint handles files.

Browser files on the device, web, and WebDAV all behave differently with directories starting with dots, files within those directories, and files starting with a dot.For example, you can create a .DS_Store file using web/WebDAV, but you can't delete a directory containing one; meanwhile, filebrowser on the device will let you delete a file/directory starting with a dot, but it won't show the .DS_Store file.
You can't delete a directory containing files via web, but you can via WebDAV/filebrowser.
In web, you can delete/create files inside a hidden directory, but you can't in WebDAV.
Ideally, all files operations needs to be brought into some kind of more uniform state, but I definitely won't be able to handle that 😵‍💫

I think I'll close the PR for now

PR #2221, opened 2026-05-31
feat: Add Simplified Chinese and Traditional Chinese translation

benbenzi:
Add Simplified Chinese and Traditional Chinese translationHi! I have added the localization files for Simplified Chinese and Traditional Chinese.

**Note on Font Support:**
Please note that this PR only includes the `.yaml` language files. For the Chinese characters to render correctly on the device, the system will need corresponding CJK (Chinese, Japanese, Korean) font support. 

Since the project already supports Noto fonts, I highly recommend using an open-source font like **Noto Sans CJK** (also known as Source Han Sans / 思源黑体) if Chinese font support hasn't been implemented yet.

Let me know if any changes are needed. Thanks for the great project!


Uri-Tauber:
Converting this PR to draft, since we don't have enough space to add the glyphs for full Chinese range.

Feel free to check
https://github.com/icannotttt/crosspoint-chinesetype
https://github.com/aBER0724/crosspoint-reader-cjk

PR #2226, opened 2026-06-01
Fix ghosting on pages following images in grayscale

itsthisjustin:
Skip rendering images that don't intersect the active grayscale band to avoid redundant processing. Force next page after an image to use HALF refresh mode to clear residual gray charge that causes ghosting when using fast diff refresh.


jeremydk:
Review?

PR #2228, opened 2026-06-01
feat: Add SD-card theme system with custom theme support

itsthisjustin:
Introduces downloadable theme packages stored on SD card under /.themes/ or /themes/. Themes are defined via theme.json with device-specific overrides, metrics, and component styling. Includes manifest generation script for hosted themes and documentation for theme creation. 

This branch is intentionally broad because SD themes touch theme discovery, theme selection, home rendering, shared Lyra drawing, downloads, settings, file browsing, recent books, and reader chrome. The table below records the largest changed areas and whether they are core to the branch or good cleanup candidates.

| Area | Approx. changed lines | Logic added or changed | Necessity | Cleanup / merge note |
| --- | ---: | --- | --- | --- |
| `src/components/themes/SdCardThemeRegistry.cpp` | 948 | Parses `theme.json`, validates ids and asset paths, applies device overrides, reads screen/widget specs, maps font/icon/action tokens, discovers installed themes from visible and hidden SD roots. | Core. This is the firmware boundary between theme JSON and runtime theme structs. | Could be split by parser domain later, but reducing it now mostly moves code around. Keep parser strict so bad themes fall back safely. |
| `src/components/themes/lyra/LyraTheme.cpp` | 917 | Makes Lyra accept optional SD theme specs for header, list, tab bar, button hints, home menu, custom icons, cover strip recents, and themed battery/status drawing while preserving built-in Lyra defaults. | Core. The branch deliberately keeps SD themes visually close to Lyra instead of creating a separate renderer that drifts. | This is the main integration point with future touch work. It routes layout through shared metrics and `UITheme::uiScale()` so touch can override density without rewriting SD theme rendering. |
| `docs/theme-creation.md` | 881 | Documents the theme schema, layout slots, widgets, metrics, components, reader chrome, icon assets, packaging, and compatibility expectations. | Useful, but not runtime-critical. | Best candidate to move to `crosspoint-tools` or a creator-facing docs area if firmware PR size matters more than keeping schema docs with parser code. |
| `src/activities/settings/ThemeDownloadActivity.cpp` | 584 | Adds in-device theme manifest fetching, install/update/delete flow, CRC validation, progress UI, cancellation, and theme list rendering. | Core if themes are installable from firmware. Not needed for manual-only SD themes. | Duplicates concepts from `FontDownloadActivity`. Later cleanup could extract a shared downloadable-package activity/helper for fonts and themes. |
| `src/activities/home/ThemeHomeRenderer.cpp` | 561 | Renders SD theme home screens from layout slots and widget specs: header title, clock, battery, recents, featured book, cover grid, launcher menu/grid/tabs, and themed button hints. | Core. This is what removes manual home positioning from individual themes. | Keep separate from `HomeActivity` so touch can swap input/layout behavior more cleanly and home activity stays mostly orchestration. |
| `src/activities/home/HomeActivity.cpp` | 316 | Converts home actions to theme-driven action entries, supports split-axis/carousel navigation, requests only needed recent covers, and delegates SD screen rendering to `ThemeHomeRenderer`. | Core. Needed for arbitrary widgets opening arbitrary activities and for carousel/tabs navigation. | Touch merge should reuse action entries instead of re-deriving menu indexes. The navigation mode enum keeps button/touch behavior explicit. |
| `src/components/UITheme.cpp` | 300 | Loads active SD theme, applies resolution/density scaling, owns current SD theme state, exposes screen specs, exposes cover thumbnail needs, and falls back to built-in themes. | Core. This centralizes runtime theme state and fallback. | Built for touch by separating resolution scale from density scale. Button devices use density `1.0`; touch branch can override `uiScale()` from board profile. |
| `src/components/themes/BaseTheme.cpp` | 223 | Adds themed reader chrome and battery rendering support while preserving base behavior. | Core for reader chrome theming. | Keep reader chrome support in base/theme layer so reader activities do not need SD-specific drawing branches. |
| `src/components/themes/BaseTheme.h` | 182 | Adds theme spec structs for recents, lists, menus, button hints, tabs, headers, battery bars, reader chrome, icon maps, and metrics scaling guard. | Core. These are the runtime data contracts parsed from JSON. | Some structs could move closer to `ThemeLayout.h`, but keeping shared component specs near `BaseTheme` makes Lyra/Base rendering dependencies explicit. |
| `src/activities/home/RecentBooksActivity.cpp` | 178 | Lets the recent-books screen use an SD layout and optional cover grid, while falling back to the standard Lyra list. | Core because themes also affect Recent Books. | Existing list remains default. Grid is theme-selected, not hardcoded globally. |
| `src/components/themes/ThemeLayout.h` | 170 | Defines row/column layout nodes, slot output, home widget types, screen widget types, actions, and navigation modes. | Core. This is the firmware-facing theme layout contract. | Mirrors the SDK layout direction so future touch scaling/layout work can converge instead of diverge. |
| `src/activities/settings/SettingsActivity.cpp` | 162 | Adds theme management entry, dynamic theme selection, and optional SD layout slots for settings chrome/list/buttons. | Core because themes affect settings. | Keep rendering path fallback-first so malformed themes cannot trap users in settings. |
| `src/ThemeInstaller.cpp` | 128 | Validates theme ids and relative paths, creates theme directories, installs files, deletes themes, clears active theme if deleted. | Core for safe downloads and manual installs. | Similar to `FontInstaller`; could be generalized later with package-root/path policy. |
| `scripts/generate-theme-icons.py` | 117 | Generates 1-bit BMP icon assets for hosted themes. | Tooling, not runtime-critical. | Good candidate to move to `crosspoint-tools`; theme builder already references this workflow. |
| `src/components/themes/ThemeLayout.cpp` | 110 | Computes row/column layout slots from fixed/flex/token sizes and normalizes header slots. | Core. This is the small runtime layout engine used by themed screens. | This is intentionally close to SDK row/column semantics to reduce future touch-branch conflict. |
| `src/activities/settings/ThemeDownloadActivity.h` | 93 | Declares downloader state, manifest data, progress state, and UI helpers. | Core if firmware installs themes. | Could shrink only if shared with font downloader. |
| `src/activities/home/RecentBookCoverPainter.cpp` | 91 | Shared cover/placeholder/title painter for home and recent-book grids. | Core for consistent cover behavior across widgets/screens. | Good deduplication: prevents home grid and Recent Books grid from generating different placeholder/cover behavior. |
| `test/theme_layout/ThemeLayoutTest.cpp` | 85 | Unit tests for layout slot emission and file-browser layout behavior. | Necessary coverage for the new layout contract. | Small and worth keeping; catches malformed slot math before firmware flashing. |
| `src/SettingsList.h` | 71 | Adds dynamic SD theme choices after built-in themes while keeping built-in indexes stable. | Core for selecting themes from settings and web settings. | Keeps old built-in enum values stable; SD themes are appended dynamically. |
| `scripts/generate-theme-manifest.py` | 66 | Creates hosted theme manifest with file sizes and CRC32 values. | Tooling, not runtime-critical. | Strong candidate to move to `crosspoint-tools`, since hosted packages and deployment live there. |

## Touch Branch Compatibility

The branch was shaped to merge with the touch work instead of competing with it:

- Theme geometry flows through `ThemeMetrics`, `scaleThemeMetrics()`, and `UITheme::uiScale()`. Button devices currently return `1.0`; the touch branch can provide device density without rewriting SD theme JSON.
- Resolution scaling and density scaling are separated. Resolution scales everything to fit a panel; density only enlarges eligible chrome so touch targets can grow without blowing up fixed cover/status layouts.
- Home actions are now semantic entries (`ThemeHomeActionEntry`) instead of hardcoded menu indexes. That gives touch and button navigation one shared activity/action model.
- Screen layout is expressed as row/column/flex/fixed/token slots, matching the SDK direction and reducing reliance on manual pixel positioning.
- SD themes remain fallback-first. If a screen layout is missing or invalid, firmware falls back to built-in Lyra behavior instead of rendering a blank or trapping the user.


itsthisjustin:
@WuTofu @jeremydk raw github used now, no changes to the downloader code, I did leave in the bitmap fix as it's essential to making the themes work. Left fonts in, left old themes in, only added 2 new themes for testing of the system. Should be a very minimal approach. Not a priority for review

sd-themes/themes.json
```diff
@@ -0,0 +1,124 @@
+{
+  "version": 1,
+  "baseUrl": "https://raw.githubusercontent.com/crosspoint-reader/crosspoint-reader/feat-sd-theme-system/sd-themes",
```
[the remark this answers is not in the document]
itsthisjustin: This will be changed prior to a merge. We just need it testable on this branch. I do not want to deal with full gh releases for themes as we should be able to update them on master to patch OTA essentially. Once we move to gitflow we will leave master as safe path and create feature/release branches which eliminates the need to do versioning of themes as well. We can add a release version compatibility layer to the theme download system if we need to release updates to themes while still supporting older builds but the destructive path here is really low as the built in theme is always available and there is no force update mechanism for ota-ing themes

docs/theme-creation.md
```diff
@@ -0,0 +1,499 @@
...
+
+## Package manifest
+
+After adding or changing hosted themes, regenerate `sd-themes/themes.json`:
+
+```bash
+python3 scripts/generate-theme-manifest.py \
+  --root sd-themes \
+  --base-url https://raw.githubusercontent.com/crosspoint-reader/crosspoint-reader/feat-sd-theme-system/sd-themes \
+  --output sd-themes/themes.json
```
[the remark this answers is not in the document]
itsthisjustin: We will fix this before merge. This allows testing on this branch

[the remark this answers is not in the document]
WuTofu: @itsthisjustin  So, we are going to use `master` for non-dev now? I'm not sure if that's the best decision.
And could you help me understand why you wouldn't have a setup like we did for sd font? That way, the beta image you provide on your website, pr testers, and the official release could just share the same link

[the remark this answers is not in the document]
itsthisjustin: Building a gh release every time we want to bump a pixel in a single theme doesn't make a ton of sense to me. Open to discussion. I found the font release system super arduous and complex when it's just a manifest file.

Imo proper git strategy would be master is used for active release including building intermediary releases for patches and emergency fixes. We'd then have feat branches that merge into a release branch for testing instead of master and then when we cut a release that release branch would merge into master. This let's us cut releases in between major versions by merging to master while still having upcoming releases on their own branches that don't keep master from having intermediator patches built.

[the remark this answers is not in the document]
WuTofu: Thanks for your explanation, that works for me

Uri-Tauber:
Oh. you solved the conflicts. 
Superb.

Just clang-format and cppcheck.

Uri-Tauber:
@itsthisjustin I will try to find time to do a proper review in the next couple of days.
Meanwhile, you can review my https://github.com/crosspoint-reader/crosspoint-reader/pull/2358, should play nicely with your work here.

itsthisjustin:
Ah yes I need to look at that. It's a great candidate to not be in the firmware though and just be a freeink component we put props in. There's already one super similar to it

itsthisjustin:
> @itsthisjustin I will try to find time to do a proper review in the next couple of days. Meanwhile, you can review my #2358, should play nicely with your work here.

Hold off for now. I'm doing a pretty big refactor and may just open up a fresh PR. We'll see how clean I can get this in the end. Theming opens up a lot of potential for streaming sd fonts which helps with localization in the UI. I'm going to broaden this scope a bit to help with other areas and base it on freeinks grid system instead of manual pixel calculations so it will work on our new touch screen devices

Uri-Tauber:
> Hold off for now. I'm doing a pretty big refactor and may just open up a fresh PR. We'll see how clean I can get this in the end. Theming opens up a lot of potential for streaming sd fonts which helps with localization in the UI. I'm going to broaden this scope a bit to help with other areas and base it on freeinks grid system instead of manual pixel calculations so it will work on our new touch screen devices

Sure. Ping me when its ready.

Do you want to open right now PR just for switching to freeink-sdk?

itsthisjustin:
> > Hold off for now. I'm doing a pretty big refactor and may just open up a fresh PR. We'll see how clean I can get this in the end. Theming opens up a lot of potential for streaming sd fonts which helps with localization in the UI. I'm going to broaden this scope a bit to help with other areas and base it on freeinks grid system instead of manual pixel calculations so it will work on our new touch screen devices
> 
> Sure. Ping me when its ready.
> 
> Do you want to open right now PR just for switching to freeink-sdk?

Yeah I was debating doing a PR that actually adds some meaningful reason to do that though? I can also just open a PR that's literally just that

Uri-Tauber:
> Yeah I was debating doing a PR that actually adds some meaningful reason to do that though? I can also just open a PR that's literally just that

I'm in favor of exclusive PR.
Will let us the chance to catch problems without mixing this with other stuff.

src/activities/settings/ThemeDownloadActivity.h
```diff
@@ -8,8 +8,7 @@
 #include "util/ButtonNavigator.h"
 
 #define THEMES_MANIFEST_VERSION 1
-#define THEME_ROOT_URL \
-  "https://raw.githubusercontent.com/crosspoint-reader/crosspoint-reader/feat-sd-theme-system/sd-themes"
+#define THEME_ROOT_URL "https://crosspointreader.com/themes"
```
WuTofu: Please stop smuggling your own domain

WuTofu (changes requested):

src/activities/settings/ThemeDownloadActivity.h
Uri-Tauber: @WuTofu Can you explain me what so bothers you about Justin's domain? 
I'm not on Discord, so forgive me if you already explained that 100 times before.

WuTofu: And this: https://github.com/crosspoint-reader/crosspoint-tools/blob/c7ba3e836030543abf28b8e22da7cbab2e2d8779/public/themes/themes.json#L3

Uri-Tauber: @WuTofu Care to share any explanation?

itsthisjustin: So just to chime in on this. Github is awful for us on these devices. Any github url we use goes through multiple redirects and forces TLS which causes downloads to go out of memory. Right now users can't even do basic OTA updates without first restarting their device and hoping for the best to the heap gods. Crosspointreader.com is the official domain for this project, not a fan site. Instead of fighting against github constantly it's smarter for us to just skip all that headache and host things in a place where we can control ssl and keep redirects to a minimum. The domain also allows us to change the final location of a url after firmware has already shipped in case we need to patch something that would require everyone to update their firmware.

WuTofu: @Uri-Tauber  sorry for the late reply. Here is my reply to Diirge's comment, which is also the explanation you are asking for

> Any github url we use goes through multiple redirects and forces TLS which causes downloads to go out of memory.

Then why not just use `crosspoint-reader.github.io`? It could minimize the redirects to zero.

> Crosspointreader.com is the official domain for this project, not a fan site. 

If this domain is solely owned and managed by you with no transparency, it functions as a private site that happens to be hosted by the current project owner (and who says you will always be?)
Furthermore, hardcoding a privately owned domain into the firmware creates a single point of failure and leaves users stranded if the domain registration lapses, changes hands, or there is a shift in project ownership down the road.

> The domain also allows us to change the final location of a url

Oh, so we now want redirection again?

itsthisjustin: I think the github pages url actually adds even MORE redirects on the backend. It's why crosspet can't currently ota at all

itsthisjustin: What does solely owned by you with no transparency mean? The website is open source and domains are owned by individual people at the end of the day. I can't exactly change how that works.

Uri-Tauber: Maybe proper solution would be to centralize all hard-coded URLs in one header, so it can be easier to swap them in the future if the need arise.

@WuTofu if other individual will have edit permissions for the website (Julia maybe...), would that satisfy you?

itsthisjustin: To be clear, outside of dns changes everyone has edit access to the website as it's in the crosspoint-tools repo in this org

WuTofu: At this point, I have lost interest in continuing this conversation. And it is not like I can do anything to stop you from doing any misconduct anyway. You do you, boo.

Uri-Tauber: @WuTofu Well, I can't speak in @itsthisjustin name, but I myself do want to hear you out, and to try to find solution which will satisfy everyone. 
You're welcome to e-mail me if you want to continue this discussion privately.

itsthisjustin: Not sure I'd consider moving downloads off github because github causes out of memory issues as misconduct....

Not trying to be a brat here but I'd love a solution to the problem besides just rejecting my solution. Github is bad for on device downloads on esp32. What's your alternative fix?

itsthisjustin: I believe the SSL issues will now be fixed by my addition of #2475

itsthisjustin:
Going to refactor this again now that touch is down from 10k lines to 2k

itsthisjustin:
Closing as I'm going to rework this after the touch branch gets in

PR #2229, opened 2026-06-01
feat: text highlighting with JSON storage and web Markdown export

jonash54:
## Summary

  Add text highlighting to the EPUB reader. Readers can select a span of text on the
  device, the selection is saved to the SD card as JSON, and highlights can later be
  exported as Markdown from the web interface — sorted by their position in the book.

  * **Reader selection mode** (`EpubReaderActivity`): a new "Highlight text" entry in
    the reader menu starts a word-by-word cursor. The first Confirm sets the start, the
    second saves the highlight. Selections may span page breaks within a chapter.
    Back cancels.
  * **Re-display**: stored highlights are drawn as underlines on a best-effort basis,
    only when the cached page geometry still matches the values captured at creation
    time (so a font/margin/orientation change never produces misplaced marks — the
    stored text stays correct regardless).
  * **Storage**: one append-only JSON file per book under `/.crosspoint/highlights/`,
    following the existing bookmark pattern (`HighlightEntry`, `HighlightUtil`,
    `JsonSettingsIO::save/loadHighlights`).
  * **Web export**: a new `/highlights` page plus `/api/highlights` (list books) and
    `/api/highlights/file` (raw JSON) endpoints. The page sorts entries by chapter then
    in-book percent and builds a Markdown (or raw JSON) download **in the browser**, so
    no Markdown-generation code lives on the device. Linked from all existing nav bars.
  * One small library getter added (`TextBlock::getWordXpos()`/`getWordStyles()`) to
    expose per-word x positions for hit-testing. No cache format change, so no
    `section.bin` version bump.

## Additional Context

* **Memory**: no new long-lived allocations. The selectable-word list is transient
  (only while selecting) and `reserve()`d; highlight JSON is loaded briefly into a
  vector like bookmarks. Overlays use plain BW line/rect draws so they survive the
  differential e-ink refresh without grayscale handling. Build: RAM 30.9%, Flash 90.4%.
* **Scope note**: `SCOPE.md` excludes "Complex Annotation: No typed out notes." This
  feature is highlighting only — no typed notes — so it should fit, but I'd appreciate
  a maintainer sanity-check here.
* **Areas to focus review on**: the cross-page selection collection logic in
  `finalizeHighlight()`, and the geometry-match gate in `drawStoredHighlights()`.
* **Known limitation (by design for now)**: selections can span pages but not chapter
  (spine) boundaries.
* **Verification**: `pio run -e default` (SUCCESS), `pio check -e default` (no defects),
  `clang-format` clean, and boots cleanly on real X4 hardware (no panic, heap healthy).
  On-device highlight interaction and the web export flow were exercised manually.

---

### AI Usage


iandchasse:
Some overlap with #1742, that implements the standard clippings format while this is JSON. It may be best to stick to the known format that kindle, koreader, etc already use/parse. I'm curious if there is any difference in the way that text selection/highlighting is approached between these two PRs, there may be room for improvement

jonash54:
Thanks for your answer! I didn't know #1742 existed when I started this. Looks like you've put a lot more into it (the windowed refresh, the annotation redisplay). If you'd rather go that direction I'm fine dropping mine.

On format: I changed the redisplay to match highlights by their saved text instead of stored coordinates, so they survive font/orientation changes, and added a Kindle-style `My Clippings.txt` export. Kept it small on purpose.

iandchasse:
> Thanks for your answer! I didn't know #1742 existed when I started this. Looks like you've put a lot more into it (the windowed refresh, the annotation redisplay). If you'd rather go that direction I'm fine dropping mine.
> 
That PR isn't mine, it's @cagliostro1991 's. I'm just pointing it out as a follower of both, I'm sure there's potential for parts of both implementation wise. I really love the idea of highlighting passages either way, would make for a good "quotes" collection to keep on the device. I've not used the my clippings feature anywhere, but it seemed like a good legacy approach

jonash54:
>  I really love the idea of highlighting passages either way, would make for a good "quotes" collection to keep on the device. 

Ah yes, i agree. I keep finding myself wanting to highlight things for looking them up later quickly. 

Also i wanted to be able to fully export all my highlights to a single file per book or of the full collection with notes on where they are from, so i built that, but it needs some improvements.

PR #2235, opened 2026-06-01
Update websocket compile orange warning flush (there is a websocket function that is going to be deprecated)

mfiordi:
## Summary
* **What is the goal of this PR?** Silence the compile-time `-Wdeprecated-declarations` warning from the `links2004/WebSockets @ 2.7.3` dependency. On recent arduino-esp32 cores, `WebSocketsClient::clientDisconnect` calls `NetworkClient::flush()`, which is now marked `[[deprecated("Use clear() instead.")]]`, so every release build prints a warning.
* **What changes are included?** A pre-build patch that swaps the deprecated `flush()` for `clear()` in the WebSockets libdep, applied via a hook following the existing `patch_jpegdec.py` convention. Specifically: `scripts/websockets_patches/0001-use-clear-instead-of-deprecated-flush.patch` (the one-line `flush()` → `clear()` change), `scripts/patch_websockets.py` (idempotent pre-build apply using GNU `patch`), and a `pre:scripts/patch_websockets.py` registration in `platformio.ini`.

## Additional Context

* The file lives in gitignored `.pio/libdeps/` (re-downloaded every build), so it's patched at build time rather than edited directly — same approach the repo already uses for JPEGDEC. The script uses GNU `patch` instead of `git apply` because the libdep sits inside the project's own git work tree, where `git apply` resolves against the enclosing repo and silently no-ops (JPEGDEC dodges this only because it's its own `.git` checkout). Risk is very low: the change is effectively compile-time only — CrossPoint uses `WebSocketsServer`, never `WebSocketsClient`, and `clear()`/`flush()` are behavior-equivalent before `stop()`. Verified with a full `pio run -e gh_release`: 0 warnings, `Applied WebSockets patch` printed, and the unpatched source still reproduces the warning when compiled directly. Suggested focus for review: the idempotency logic in `patch_websockets.py`.

---

### AI Usage

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -53,6 +53,7 @@ extra_scripts =
   pre:scripts/gen_i18n.py
   pre:scripts/git_branch.py
   pre:scripts/patch_jpegdec.py
+  pre:scripts/patch_websockets.py
   post:scripts/register_unit_tests_target.py
 
 ; Libraries
--- a/scripts/patch_websockets.py
+++ b/scripts/patch_websockets.py
@@ -1,0 +1,103 @@
+"""
+PlatformIO pre-build script: apply CrossPoint's WebSockets patches via
+`git apply`.
+
+The upstream links2004/WebSockets 2.7.3 pin calls the deprecated
+`NetworkClient::flush()` in `WebSocketsClient::clientDisconnect`. Recent
+arduino-esp32 cores tag that method `[[deprecated("Use clear() instead.")]]`,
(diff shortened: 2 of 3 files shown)


Uri-Tauber:
@mfiordi Sorry for ignoring your PR.
I'm afraid we have much more PRs than time to review them. Since your work here doesn't add concrete improvement, I advocate to close this.
Sure, the warning is annoying, but patching upstream library can break somewhere down the road...

Issue #1674, opened 2026-04-16
long press to move to previous chapter should actually move to start of current


mfiordi:
### Affected Version

1.2.0

### Describe the Bug

when I long press to go to previous chapter, I get to the start of previous chapter; actually, usual ebook readers (kindle, kobo) would first move to start of current chapter, and then another click would go to the start of the previous.

### Steps to Reproduce

1. open a book
2. go to chapter 2+, not first page
3. long press for previous chapter
4. i get to previous chapter instead of first page of current chapter 

### Expected Behavior

at first, show the first page of CURRENT chapter, unless you are already on it; second long press, would go to first page of previous chapter 

### Relevant Log Output/Screenshots

```shell

```

Uri-Tauber:
Closing as solved.

PR #2243, opened 2026-06-02
fix: long-press back should move to the start of chapter

Uri-Tauber:
## Summary

* **What is the goal of this PR?** fix issue #1674

## Additional Context

`EpubReaderActivity.cpp:340-357` — In the `CHAPTER_SKIP` long-press handler, when pressing "back" (`!nextTriggered`): if the current section exists and we're not already on its first page (`currentPage > 0`), reset `currentPage` to 0. Only when we're already on page 0 (or going forward) does it behave as before — skipping to the adjacent chapter.

---

### AI Usage

### DIFF
--- a/src/activities/reader/EpubReaderActivity.cpp
+++ b/src/activities/reader/EpubReaderActivity.cpp
@@ -338,17 +338,20 @@ void EpubReaderActivity::loop() {
   }
 
   if (longPress && SETTINGS.longPressButtonBehavior == SETTINGS.CHAPTER_SKIP) {
+    if (!nextTriggered && section && section->currentPage > 0) {
+      section->currentPage = 0;
+      requestUpdate();
+      return;
+    }
+
     // We don't want to delete the section mid-render, so grab the semaphore
     {
       RenderLock lock(*this);
       nextPageNumber = 0;
       if (nextTriggered) {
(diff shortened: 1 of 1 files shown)


uxjulia:
This is definitely a nitpick, but moving the special-case handling to the top would be easier to read. Other than that, tested it on device and it works great! Thank you for this, this always bugged me haha

```
  if (longPress && SETTINGS.longPressButtonBehavior == SETTINGS.CHAPTER_SKIP) {
    if (!nextTriggered && section && section->currentPage > 0) {
      section->currentPage = 0;
      requestUpdate();
      return;
    }

    // We don't want to delete the section mid-render, so grab the semaphore
    {
      RenderLock lock(*this);
      nextPageNumber = 0;
      if (nextTriggered) {
        currentSpineIndex++;
      } else if (currentSpineIndex > 0) {
        currentSpineIndex--;
      }
      section.reset();
    }
    requestUpdate();
    return;
  }
```

uxjulia (dismissed):

Uri-Tauber:
> This is definitely a nitpick, but moving the special-case handling to the top would be easier to read.

I'm too lazy to do it right now. Feel free to edit the PR.

znelson (approved):

PR #2247, opened 2026-06-02
fix: resolve KOSync XPath for div-only chapters

davidstosik:
## Summary

* Fixes KOReader Sync uploads for EPUB chapters whose visible text is stored in `<div>` elements rather than `<p>`/`<li>`.
* CrossPoint previously could fall back to a chapter-level locator like `/body/DocFragment[9]/body`, causing other KOReader-compatible clients to jump to the beginning of the chapter when applying CrossPoint progress.
* The progress resolver now counts and resolves non-whitespace visible text anywhere inside `<body>`, so div-only chapters can produce precise element/text XPath locators.
* Adds focused host-side regression tests for `ChapterXPathResolver::findXPathForProgress`.

## Additional Context

* Verified on hardware with the KOReader Sync server:
  * Before: `/body/DocFragment[9]/body`
  * After: `/body/DocFragment[9]/body/div[59]/text()[1].63`
* Host tests run:
  * `ctest --test-dir build/test --output-on-failure -j`
* Firmware build/upload was tested locally on device.

---

### AI Usage


davidstosik:
@itsthisjustin
Hi!
Do you need anything from me to get this PR through the finish line?

itsthisjustin:
> @itsthisjustin Hi! Do you need anything from me to get this PR through the finish line?

I guess there's some conflict in the test now? Can you solve that? @uxjulia can you also take a look at this as you might be closer to the kosync stuff than I am at this point due to user feedback

uxjulia:
@davidstosik - Can you rebase this on the latest `develop` branch and make sure your changes don't conflict with any code in this area since June 2? There's been a lot of updates in this area since then.

Uri-Tauber:
I think #2805 solved this in a different way.
@davidstosik Could you try latest `develop` and see if your fix is still required?

PR #2249, opened 2026-06-02
fix(epub): decode percent-encoded internal asset paths so assets render correctly

uxjulia:
## Summary

  * Fix EPUBs where internal file references are written in URL-style escaped form, like spaces appearing as `%20`, so the reader can find the right files instead of treating those references as missing.
  * Added a small shared helper that converts those escaped EPUB-internal paths back into their normal filenames before we try to look them up.
  * Applied that cleanup step across the EPUB parsing flow wherever we resolve internal references, including cover images, manifest items, TOC links, spine entries, and inline HTML images.
  * Kept the change narrowly focused on EPUB-internal asset resolution rather than changing broader URL or networking behavior.

## Additional Context

* The user-facing bug here is that some books package their internal filenames in an escaped form, so a file like `Chapter 1.xhtml` may be referenced more like `Chapter%201.xhtml`. The reader was treating that escaped text as the literal filename, which means otherwise-valid books could lose images, covers, or chapter targets because the lookup no longer matched the real file inside the EPUB.
* Risk is intentionally low. The helper only rewrites valid `%XX` escape sequences and leaves malformed input alone, so it should improve compatibility with escaped filenames without broadening the parser’s behavior in unrelated cases.

## Local Testing Performed
* This was tested on my device with the user-provided optimized epub that was not rendering images within the text prior to this fix (cover image and chapter headers were rendering fine): 
[orv_main_baseline.epub.zip](https://github.com/user-attachments/files/28529545/orv_main_baseline.epub.zip)
* This was also tested by the user with a local build and the affected epub and confirmed to be working

## Steps for Testing
* Try to open the affected epub (linked above) or any epub that has similar percent-encoding on a build prior to this fix.
* Apply this fix, clear book cache, and re-open the affected book.
* Images should render properly.
---

### AI Usage

### DIFF
--- a/lib/Epub/Epub/parsers/ContentOpfParser.cpp
+++ b/lib/Epub/Epub/parsers/ContentOpfParser.cpp
@@ -186,7 +186,7 @@ void XMLCALL ContentOpfParser::startElement(void* userData, const XML_Char* name
       if (strcmp(atts[i], "id") == 0) {
         itemId = atts[i + 1];
       } else if (strcmp(atts[i], "href") == 0) {
-        href = FsHelpers::normalisePath(self->baseContentPath + atts[i + 1]);
+        href = FsHelpers::normalisePath(FsHelpers::decodeUriEscapes(self->baseContentPath + atts[i + 1]));
       } else if (strcmp(atts[i], "media-type") == 0) {
         mediaType = atts[i + 1];
       } else if (strcmp(atts[i], "properties") == 0) {
@@ -315,7 +315,7 @@ void XMLCALL ContentOpfParser::startElement(void* userData, const XML_Char* name
       if (strcmp(atts[i], "type") == 0) {
         type = atts[i + 1];
       } else if (strcmp(atts[i], "href") == 0) {
-        guideHref = FsHelpers::normalisePath(self->baseContentPath + atts[i + 1]);
+        guideHref = FsHelpers::normalisePath(FsHelpers::decodeUriEscapes(self->baseContentPath + atts[i + 1]));
       }
     }
     if (!guideHref.empty()) {
--- a/lib/Epub/Epub.cpp
+++ b/lib/Epub/Epub.cpp
@@ -118,7 +118,7 @@ bool Epub::parseContentOpf(BookMetadataCache::BookMetadata& bookMetadata, const 
       }
 
       if (!imageRef.empty()) {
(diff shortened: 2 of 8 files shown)


Uri-Tauber (dismissed):
LGTM.
@uxjulia do you have an epub I can test this with?

uxjulia:
> LGTM. @uxjulia do you have an epub I can test this with?

Yep! It's linked in the description of this PR, just remove the `.zip` extension after downloading

Uri-Tauber:
> Yep! It's linked in the description of this PR, just remove the .zip extension after downloading

Oppps. My bad. Completely missed that.

uxjulia:
Addressed some coderabbit comments

Uri-Tauber:
Tested on my device with the provided epub, and seems to work well.
Good job.

Uri-Tauber (approved):

lib/Epub/Epub/parsers/ContentOpfParser.cpp
```diff
@@ -186,7 +186,7 @@ void XMLCALL ContentOpfParser::startElement(void* userData, const XML_Char* name
       if (strcmp(atts[i], "id") == 0) {
         itemId = atts[i + 1];
       } else if (strcmp(atts[i], "href") == 0) {
-        href = FsHelpers::normalisePath(self->baseContentPath + atts[i + 1]);
+        href = FsHelpers::normalisePath(FsHelpers::decodeUriEscapes(self->baseContentPath + atts[i + 1]));
```
znelson, line 189: Now that spine entries are stored decoded, `Epub::resolveHrefToSpineIndex` at lib/Epub/Epub.cpp:900 does an exact-string compare between the (still-encoded) raw href from `FootnoteEntry` and the (now-decoded) spine entries. So likely footnote/cross-chapter links in EPUBs with encoded `<a href="...">` will fail to resolve. A minimal fix is to decode href at the top of `resolveHrefToSpineIndex` before the # split. Alternatively, decode at ChapterHtmlSlimParser.cpp:727 where `currentFootnote.href` is captured.

znelson (approved):

lib/Epub/Epub/parsers/ContentOpfParser.cpp
uxjulia: Good catch. Since this PR is approved, I’ll handle this as a small follow-up PR.

Issue #2199, opened 2026-05-29
screenshot doesn't work


Evgen135:
### Affected Version

Version 1.3.0

### Describe the Bug

When using the "take a screenshot" function, or when pressing the power button and volume down button simultaneously, some books do not generate a screenshot, and no folder with the book’s title appears in the screenshots/ folder.

All books are in Russian. I have tried changing and shortening the book titles. I have also tried changing the titles to English. I updated the metadata through Calibre. Nothing helped.

Here is an example of a book where I cannot take a screenshot:

[Ирвин Ялом - Вглядываясь в солнце.zip](https://github.com/user-attachments/files/28388281/-.zip)

Here is a book where the screenshot works:

[Станислав Лем - Эдем (полный перевод).zip](https://github.com/user-attachments/files/28388287/-.zip)

For the book that works correctly, a folder named "Эдем-(полный-перевод)" is created. If I delete that folder and change the book’s title, when I take a screenshot again, a folder with the old title is created.

So, it seems that the title is taken from some internal book data that I cannot access.

### Steps to Reproduce

1. Unzip the book from the archive—it's in .epab format.
2. DownIoad it to your device.
З. Ореп the book.
4. Таке а screenshot Taking а screenshot won•t capture the screen; it wilI simply ореп the current раде.
5. Check the Screenshots folder оп уоиг device.
б. There won't be а screenshot. and по folder with the booes пате will be created.

### Expected Behavior

I expected a folder with the book's name to be created in the screenshots folder. A screenshot would be placed in it.

### Relevant Log Output/Screenshots

```shell

```

bizzon2:
I have the same problem. First, I discovered that screenshots weren't being taken on one book, and today, on another. There's a problem, and it needs to be addressed.

serialx:
Possible duplicate of #2103

PR #2254, opened 2026-06-03
fix: keep FAT32 path sanitizer from splitting UTF-8 characters

### LINKED ISSUE #2103: Screenshots fail (perhaps due to cyrilic book title?)
### Affected Version

Version 1.3.0

### Describe the Bug

Since the bookmark feature isn't implemented yet, I'm using the screenshot feature.

I was reading a book today and decided to take a screenshot. I discovered that the screenshot wasn't created. I tried it with other books, and it did.

I've attached the book in the archive. The book is in Russian.

[Инчаспе Джесси Богиня глюкозы Нормализуйте уровень сахара в крови, чтобы изменить свою жизнь.zip](https://github.com/user-attachments/files/28114978/default.zip)

### Steps to Reproduce

1. Unzip the book from the archive—it's in .epab format.
2. Download it to your device.
3. Open the book.
4. Take a screenshot. Taking a screenshot won't capture the screen; it will simply open the current page.
5. Check the Screenshots folder on your device.
6. There won't be a screenshot, and no folder with the book's name will be created.

### Expected Behavior

I expected a folder with the book's name to be created in the screenshots folder. A screenshot would be placed in it.

```shell

```

SurprisedDuck:
## Problem

Fixes #2103 (and likely the long-title cases in #2199): taking a screenshot of a book with a long non-ASCII title (e.g. Cyrillic) silently fails — no `screenshots/<title>/` folder is created.

From the reporter's log in #2103:

```
[SCR] Directory /screenshots/Богиня-глюкозы.-Нормализуйте-уров� does not exist, creating it
[SCR] Failed to create directory /screenshots/Богиня-глюкозы.-Нормализуйте-уров�
[SCR] Failed to save screenshot
```

Note the trailing `�` — the directory name ends in an **incomplete UTF-8 sequence**.

## Root cause

[`FsHelpers::sanitizePathComponentForFat32`](https://github.com/crosspoint-reader/crosspoint-reader/blob/master/lib/FsHelpers/FsHelpers.cpp) copies the input byte-by-byte up to `maxLen - 1` and then null-terminates, with no regard for UTF-8 character boundaries. `ScreenshotUtil::buildFilename` calls it with a 64-byte buffer:

```cpp
char sanitizedTitle[64];
FsHelpers::sanitizePathComponentForFat32(info.title, sanitizedTitle, sizeof(sanitizedTitle));
```

A Cyrillic title (2 bytes/char) that reaches the 63-byte cap mid-character is left with a dangling lead byte. FAT32 long filenames are stored as UTF-16, so SdFat (`USE_UTF8_LONG_NAMES`) cannot convert the invalid sequence and directory creation fails — so the screenshot is never saved.

This is exactly why **short** non-ASCII titles work but **long** ones don't (as both reporters observed): it depends purely on whether byte 63 lands inside a multi-byte character.

## Fix

After truncation, if the first un-copied byte is a UTF-8 continuation byte (`10xxxxxx`), drop the partial trailing character so the result is always valid UTF-8. ASCII-only titles and titles that fit are unaffected.

## Verification

Verified on host with the function copied verbatim, using the real failing title:

| | output | valid UTF-8? |
|---|---|---|
| before | `Богиня-глюкозы.-Нормализуйте-уров�` (63 B) | ❌ (matches the log) |
| after  | `Богиня-глюкозы.-Нормализуйте-уров` (62 B)  | ✅ |

Also exercised 2- and 3-byte characters truncated at **every** boundary offset, exact-fit, ASCII mid-word cuts, and `maxLen` of 0/1 — all produce valid UTF-8, and existing behaviour (e.g. space → `-`) is preserved.

🔲 Not yet tested on hardware — would appreciate a maintainer confirming a screenshot of the EPUBs attached to #2103 / #2199 now works.

Supported by Claude Opus 4.8


lib/FsHelpers/FsHelpers.cpp
```diff
@@ -154,6 +154,21 @@ void sanitizePathComponentForFat32(const char* input, char* output, size_t maxLe
...
+  // copy; if it is a continuation byte (10xxxxxx) the last copied character
+  // is incomplete and must be removed.
+  if (input[i] != '\0' && (static_cast<unsigned char>(input[i]) & 0xC0) == 0x80) {
+    while (i > 0 && (static_cast<unsigned char>(output[i - 1]) & 0xC0) == 0x80) {
+      i--;
+    }
+    if (i > 0) {
+      i--;
+    }
+  }
```
znelson: Would we be able to replace this with just a call to Utf8.h utility?

```suggestion
  i = utf8SafeTruncateBuffer(output, static_cast<int>(i));
```

Uri-Tauber:
Tested with both epubs from #2103 and #2199, and the screenshot still failing on both.
Seems it wasn't the problem.

I cant see now the log `[SCR] Failed to create directory X`, only `[SCR] Failed to save screenshot`. No new directory was created.

Still, thx for trying.

SurprisedDuck:
@Uri-Tauber you're right — thanks for retesting, that was the missing piece. The dangling `�` you still saw pointed at a second truncation I'd missed.

**Root cause:** the title is byte-truncated *before* it ever reaches the sanitizer. `EpubReaderActivity::getScreenshotInfo()` does `snprintf(info.title /* char[64] */, "%s", title)`, which cuts a long Cyrillic title at a byte boundary and leaves a dangling UTF-8 lead byte followed by `\0`. My original fix only handled truncation happening *inside* `sanitizePathComponentForFat32` — it keyed off the next input byte being a continuation byte. On a pre-truncated string the copy loop stops on `\0`, so that check never fired and the partial byte passed straight through → SdFat (`USE_UTF8_LONG_NAMES`) rejected the name → no folder, "Failed to save screenshot".

**Fix (`caffc62`):** the sanitizer now inspects its own output tail — it walks back over the trailing continuation bytes to the lead byte and keeps the character only if the whole sequence is present, otherwise drops it. That covers both the in-function and the pre-truncated case, and leaves complete strings (including 4-byte glyphs) untouched.

I reproduced both reported titles (#2103 "Богиня глюкозы…", #2199 "Вглядываясь в солнце…") on the host: the pre-truncation step does produce invalid UTF-8, and after sanitizing the output is valid UTF-8 (`Вглядываясь-в-солнце.-Жизнь-без-ст`). Added a host gtest suite under `test/fs_helpers/` covering both. Would appreciate one more on-device check when you get a chance 🙏

Uri-Tauber:
@SurprisedDuck Do you actually have (or intend to buy) Xteink reader?

SurprisedDuck:
I have one, and used it up to 1.2 of crosspoint-reader.
Now its at a friends until he finishes his book.
You can leave this for now and I'll come back to it when I have the reader again.

Uri-Tauber:
@SurprisedDuck Ok. Great.
Will wait to hear from you.

SurprisedDuck:
Rebased on latest `master` and pushed an update that gets the new `test/fs_helpers` target green:

- **unit-tests**: that target is the first to compile `FsHelpers.cpp` on the host, which surfaced a latent missing `#include <cstdint>` (used by `hexValue`/`decodeUriEscapes` — it only built on-device because the Arduino core pulls `<cstdint>` in transitively). Added the explicit include.
- **clang-format**: reflowed the one-line `ReservedCharactersReplaced` test.
- **@znelson's suggestion**: done — the hand-rolled UTF-8 tail walk in `sanitizePathComponentForFat32` is now a single call to `utf8SafeTruncateBuffer()` from `Utf8.h`, and `Utf8.cpp` is linked into the test target. Behaviour is identical for all realistic inputs; the full host suite (85 tests, incl. the 6 `SanitizeFat32` cases) passes locally. (Only theoretical difference: a buffer that is *entirely* one incomplete multi-byte sequence is kept rather than emptied — not reachable for a 64-byte screenshot title.)

On the **on-device check**: I still can't run it yet — the reader is at a friend's until he finishes his book. I'll verify screenshot saving on hardware (the #2103/#2199 Cyrillic titles) and report back here once I have it again, so I'd suggest leaving this as draft until then.

Uri-Tauber:
Closing as stale.
@SurprisedDuck When you get your device, open new PR.

PR #2263, opened 2026-06-04
perf: Minimize string allocations in CSS parsing

znelson:
## Summary

Use `std::string_view` and case-insensitive comparisons to avoid string allocations during CSS parsing. 

**Hot path:** `resolveStyle` (called per HTML start tag during chapter rendering) now does zero heap allocations. Previously it allocated a normalized tag string, a vector of class strings, and a composite key per class. For a chapter with ~2000 tags × 2 classes each, that's ~12 000 small short-lived allocations eliminated per page render — primarily a heap-fragmentation win on the ESP32-C3's ~380KB RAM.

**Cold path:** CSS load no longer allocates per-rule selector vectors or per-token strings; `splitOnChar`/`splitWhitespace` are gone, replaced with callback-based tokenization (`forEachDelimitedToken`).

**Behavioral notes:**
- The selector `unordered_map` now uses an ASCII-case-insensitive hash/equal. Selectors are stored with their original case rather than pre-lowercased; the observable lookup result is unchanged.
- `stripTrailingImportant` is now case-insensitive (per CSS spec; previously matched only lowercase `!important`).

**Cache compatibility:** `CSS_CACHE_VERSION` unchanged. Old caches (lowercase selectors) load correctly under the new lookup; new caches will contain verbatim-case selectors — both forms work.

---

### AI Usage

### DIFF
--- a/lib/Epub/Epub/css/CssParser.cpp
+++ b/lib/Epub/Epub/css/CssParser.cpp
@@ -92,6 +92,13 @@ void forEachDelimitedToken(std::string_view s, Pred isDelimiter, F&& fn) {
   }
 }
 
+// FNV-1a (64-bit) per Fowler/Noll/Vo. `fnv1aMix` is the per-byte mix step;
+// callers apply any byte-level transform (e.g. asciiToLower) before mixing.
+constexpr size_t FNV_OFFSET_BASIS = 14695981039346656037ULL;
+constexpr size_t FNV_PRIME = 1099511628211ULL;
+
+constexpr size_t fnv1aMix(size_t hash, unsigned char byte) { return (hash ^ byte) * FNV_PRIME; }
+
 // Parse the entirety of s as a number into `out`. Accepts an optional leading
 // '+' (which std::from_chars rejects by spec) so callers can pass CSS-style
 // signed numbers without manual trimming. Returns false on empty input, a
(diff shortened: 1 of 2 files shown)


Uri-Tauber:
Awesome.
Will review later.

lib/Epub/Epub/css/CssParser.cpp
```diff
@@ -50,7 +50,79 @@ constexpr size_t MIN_FREE_HEAP_FOR_CSS = 48 * 1024;
...
+      }
+      start = i + 1;
+    }
+  }
+}
+
+// FNV-1a (64-bit) per Fowler/Noll/Vo. `fnv1aMix` is the per-byte mix step;
+// callers apply any byte-level transform (e.g. asciiToLower) before mixing.
+constexpr size_t FNV_OFFSET_BASIS = 14695981039346656037ULL;
+constexpr size_t FNV_PRIME = 1099511628211ULL;
```
Uri-Tauber: ESP32 is 32-bit, so `size_t` is 32 bits. this assignment will truncates the FNV consts.
use uint64_t and update `fnv1aMix` accordingly.

lib/Epub/Epub/css/CssParser.cpp
```diff
@@ -260,124 +297,121 @@ bool CssParser::tryInterpretLength(const std::string& val, CssLength& out) {
...
-  std::string propValueBuf;
 
   size_t start = 0;
   for (size_t i = 0; i <= declBlock.size(); ++i) {
     if (i == declBlock.size() || declBlock[i] == ';') {
       if (i > start) {
-        const size_t len = i - start;
-        std::string decl = declBlock.substr(start, len);
+        const std::string_view decl = declBlock.substr(start, i - start);
         if (!decl.empty()) {
```
Uri-Tauber: Since  `i > start`  is required to enter this block,  `i - start`  is strictly positive and  `decl`  can never be empty.

lib/Epub/Epub/css/CssParser.cpp
znelson: Excellent catch, thank you

Uri-Tauber (approved):

itsthisjustin (approved):

PR #2268, opened 2026-06-05
chore: add t5s3 fork

Uri-Tauber:
## Summary

* **What is the goal of this PR?** Reference new Crosspoint fork, teagerting Lilygo T5 S3
---

### AI Usage

### DIFF
--- a/README.md
+++ b/README.md
@@ -254,7 +254,7 @@ One of the best things about open source is that anyone can take the code in a d
 
 - [crosspoint-reader-papers3](https://github.com/juicecultus/crosspoint-reader-papers3) — Crosspoint port for M5Stack Paper S3. 
 
-- [t5s3-reader](https://github.com/ShallowGreen123/t5s3-reader) — Crosspoint port for LilyGo T5 ePaper S3 / T5S3 4.7-inch e-paper device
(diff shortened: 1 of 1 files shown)


itsthisjustin:
So my preference is to wait a bit on this and actually swap in the freeink.org sdk here which adds native support for this device to crosspoint vs adding in forks. Open to suggestion though. In general we need to abstract away from just xteink devices officially vs just having more and more forks

znelson (approved):

Uri-Tauber:
> So my preference is to wait a bit on this and actually swap in the freeink.org sdk here which adds native support for this device to crosspoint vs adding in forks. Open to suggestion though. In general we need to abstract away from just xteink devices officially vs just having more and more forks

No arguments here. Exactly what I said [here](https://crosspoint.zulipchat.com/#narrow/channel/597154-Other-Hardware/topic/Abstracting.20CrossPoint). However, until we can ship functional and stable S3 release, I do think we should highlight the forks already doing so. Letting the community know about them.
Plus, I'm counting on porting upstream parts of this and the M5stack forks.

itsthisjustin (approved):

PR #2277, opened 2026-06-06
fix(epub): NFC-normalize EPUB text so NFD diacritics render correctly

jetaudio:
## Problem

Some EPUBs store text in **NFD** form (a base letter followed by separate combining diacritical marks) instead of precomposed **NFC**. This is common for Vietnamese and is used for many EPUB `<h1>` chapter headings. The device fonts carry no combining-mark positioning, so NFD text renders with the accents detached/misplaced and titles lose their diacritics entirely.

Example: the `toc.ncx` chapter titles of some Vietnamese EPUBs render broken on device.

## Fix

Add `utf8ComposeNfc()` — a canonical-composition pass for the Latin/Vietnamese range backed by a generated composition table (`lib/Utf8/Utf8ComposeTable.h`, 847 pairs) — and apply it where text reaches the screen:

- **`Epub::parseContentOpf`** — composes the OPF book title at index time.
- **`Epub::getTocItem`** — composes *by value on read*, so already-cached books (whose `book.bin` may still hold NFD titles) display correctly **without re-indexing**.
- **`EpubReaderActivity::renderStatusBar`** — the book-title status-bar line.
- **`ParsedText::addWord`** — the single funnel every laid-out word passes through, fixing in-page chapter headings and body text.

`addWord` runs the pass once per word at layout time (the result is cached in the section file), backed by a cheap fast path in `utf8ComposeNfc` that returns mark-free text untouched without walking codepoints or allocating.

`SECTION_FILE_VERSION` is bumped `25 -> 26` to invalidate NFD section caches so already-indexed books re-layout with composed text.

## Cost / impact

- Self-contained in the existing `lib/Utf8` module; no new dependencies.
- Fast path makes it a near-no-op for ASCII / already-NFC text (the vast majority of words).
- Builds clean for `default` (ESP32-C3): RAM 30.9%, Flash 78.9%.

## Notes

This change was developed with AI assistance (Claude); authored and reviewed by me.


lib/Epub/Epub/Section.cpp
```diff
@@ -10,7 +10,9 @@
 #include "parsers/ChapterHtmlSlimParser.h"
 
 namespace {
-constexpr uint8_t SECTION_FILE_VERSION = 25;
+// v26: words are NFC-composed at layout time (ParsedText::addWord) so NFD Vietnamese
+// chapter headings/body render with precomposed glyphs. Bump invalidates NFD caches.
```
Uri-Tauber: No need really to over detail.
```suggestion
```

lib/Utf8/Utf8ComposeTable.h
```diff
@@ -0,0 +1,229 @@
...
+    {0x0456, 0x0308, 0x0457}, {0x0474, 0x030F, 0x0476}, {0x0475, 0x030F, 0x0477}, {0x04D8, 0x0308, 0x04DA},
+    {0x04D9, 0x0308, 0x04DB}, {0x04E8, 0x0308, 0x04EA}, {0x04E9, 0x0308, 0x04EB}, {0x1E36, 0x0304, 0x1E38},
+    {0x1E37, 0x0304, 0x1E39}, {0x1E5A, 0x0304, 0x1E5C}, {0x1E5B, 0x0304, 0x1E5D}, {0x1E60, 0x0323, 0x1E68},
+    {0x1E61, 0x0323, 0x1E69}, {0x1E62, 0x0307, 0x1E68}, {0x1E63, 0x0307, 0x1E69}, {0x1EA0, 0x0302, 0x1EAC},
+    {0x1EA0, 0x0306, 0x1EB6}, {0x1EA1, 0x0302, 0x1EAD}, {0x1EA1, 0x0306, 0x1EB7}, {0x1EB8, 0x0302, 0x1EC6},
+    {0x1EB9, 0x0302, 0x1EC7}, {0x1ECC, 0x0302, 0x1ED8}, {0x1ECC, 0x031B, 0x1EE2}, {0x1ECD, 0x0302, 0x1ED9},
+    {0x1ECD, 0x031B, 0x1EE3}, {0x1ECE, 0x031B, 0x1EDE}, {0x1ECF, 0x031B, 0x1EDF}, {0x1EE4, 0x031B, 0x1EF0},
+    {0x1EE5, 0x031B, 0x1EF1}, {0x1EE6, 0x031B, 0x1EEC}, {0x1EE7, 0x031B, 0x1EED},
+};
+static constexpr int kUtf8ComposeTableSize = 847;
```
Uri-Tauber: Would be more maintainable
```suggestion
static constexpr int kUtf8ComposeTableSize =
    sizeof(kUtf8ComposeTable) / sizeof(kUtf8ComposeTable[0]);
```

src/activities/reader/EpubReaderActivity.cpp
```diff
@@ -1095,7 +1096,7 @@ void EpubReaderActivity::renderStatusBar() const {
     }
 
   } else if (SETTINGS.statusBarTitle == CrossPointSettings::STATUS_BAR_TITLE::BOOK_TITLE) {
-    title = epub->getTitle();
+    title = utf8ComposeNfc(epub->getTitle());  // NFC so NFD titles render correctly
```
Uri-Tauber: Correct me if I'm wrong here: Doesn't `bookMetadata.title` already normalized at parse time in Epub.cpp?

lib/Epub/Epub.cpp
```diff
@@ -815,7 +817,11 @@ BookMetadataCache::TocEntry Epub::getTocItem(const int tocIndex) const {
     return {};
   }
 
-  return bookMetadataCache->getTocEntry(tocIndex);
+  // Normalize the chapter title to NFC here so already-cached books (whose
+  // book.bin may hold NFD titles) also display correctly without re-indexing.
+  auto entry = bookMetadataCache->getTocEntry(tocIndex);
+  entry.title = utf8ComposeNfc(entry.title);
+  return entry;
```
Uri-Tauber: Do we really need this? It add extra heap allocations, and the book will probably need reindexing anyway because `SECTION_FILE_VERSION` was changed.

Uri-Tauber:
Apart from those comments, the PR is very good. The scoping is beautiful. 

Two more points:
a. Can you take some screenshots with your phone of before/after?
b. Not a blocker, but would be useful to add some tests here.

Uri-Tauber:
@jetaudio Do you think you will have time to fix and solve conflicts soon?

jetaudio:
@Uri-Tauber U
Sorry for late reply. I'm solving the conflict right now.

Uri-Tauber (dismissed):
Very Good.
Still need to test on device, though.

jetaudio:
|BEFORE|AFTER|
|---|---|
|before|after|



Add before and after screenshots

lib/Utf8/Utf8ComposeTable.h
```diff
@@ -0,0 +1,229 @@
...
+#include <cstdint>
+
+struct Utf8ComposeEntry {
+  uint16_t base;
+  uint16_t mark;
+  uint16_t composed;
+};
+
+// Sorted by (base, mark) for binary search.
+static const Utf8ComposeEntry kUtf8ComposeTable[] = {
```
uxjulia: Nit: per the repo’s CLAUDE.md “constexpr first” rule for lookup tables, should this be static constexpr instead of static const? The table is compile-time data, and the size constant below is already constexpr.

```suggestion
static constexpr Utf8ComposeEntry kUtf8ComposeTable[] = {
```

Ariel14ss14:
@Uri-Tauber can you give me a review so i chack on my device the hebrew nikud

Uri-Tauber:
@Ariel14ss14 This PR probably won't help with Hebrew niqqud by itself.

As I understand it (correct me if I'm wrong, @jetaudio), the issue being fixed here is that some EPUBs store Vietnamese text in NFD form (base letter + vowel marks) even though equivalent precomposed NFC characters exist. The firmware doesn't know how to position vowel marks correctly. This PR converts those NFD sequences into their NFC equivalents before rendering.
Hebrew niqqud is different. Hebrew letters with niqqud do not have standard precomposed Unicode NFC characters. As a result, this PR doesn't solve Hebrew niqqud rendering issues by itself.

My idea for Hebrew was to create custom glyphs in the font file (mapped into an unused codepoint range) and then apply a similar composition step that converts common Hebrew letter+niqqud combinations into those custom glyphs. But that would be a separate work from this PR.

Ariel14ss14:
Oh now i understand

jetaudio:
> @Ariel14ss14 This PR probably won't help with Hebrew niqqud by itself.
> 
> 
> 
> As I understand it (correct me if I'm wrong, @jetaudio), the issue being fixed here is that some EPUBs store Vietnamese text in NFD form (base letter + vowel marks) even though equivalent precomposed NFC characters exist. The firmware doesn't know how to position vowel marks correctly. This PR converts those NFD sequences into their NFC equivalents before rendering.
> 
> Hebrew niqqud is different. Hebrew letters with niqqud do not have standard precomposed Unicode NFC characters. As a result, this PR doesn't solve Hebrew niqqud rendering issues by itself.
> 
> 
> 
> My idea for Hebrew was to create custom glyphs in the font file (mapped into an unused codepoint range) and then apply a similar composition step that converts common Hebrew letter+niqqud combinations into those custom glyphs. But that would be a separate work from this PR.

That's right. This PR is Vietnamese only

uxjulia (dismissed):

PR #2288, opened 2026-06-08
feat: Support for Korean line breaks and glyph spacing

itsthisjustin:
This branch improves Korean/CJK EPUB rendering by adding targeted line-breaking and font metric fixes in the reader layout engine. Korean, Chinese, Japanese, and fullwidth text can now wrap at character boundaries with basic punctuation guards, while English and other space-delimited languages continue using the existing word and hyphenation behavior. It also tightens SD-card font selection and glyph measurement so multilingual fonts are measured more consistently, with memory optimizations for larger EPUB/ font workloads.

 - Added CJK-aware character wrapping for Hangul, CJK ideographs, kana, and fullwidth text.

 - Added punctuation rules to avoid bad breaks around opening/closing CJK punctuation.

 - Disabled kerning around CJK codepoints to prevent compressed or nested glyph placement.

 - Improved SD-card font glyph advance fallback when cached metrics are missing.

 - Centralized SD-card reader font-size selection using closest 12/14/16/18 pt matches.

 - Optimized SD-card font interval storage for BMP-only fonts.

 - Reduced EPUB ZIP extraction peak memory by streaming compressed reads.


Uri-Tauber:
Do we plan to support CJK?
I was under the impression that "If another popular fork solves it well, we defer." to quote [the roadmap](https://crosspointreader.com/roadmap).

itsthisjustin:
> Do we plan to support CJK? I was under the impression that "If another popular fork solves it well, we defer." to quote [the roadmap](https://crosspointreader.com/roadmap).

Stock doesn't handle korean well as far as I know but could be wrong. Ko crosspoint is a 16mb binary and previously wasn't able to be OTA flashed on locked devices. This isn't a must have but we had a lot requests for it and it's a pretty minimal implementation that fixes a few of other font/memory things in the process

Uri-Tauber:
> Stock doesn't handle korean well as far as I know but could be wrong. Ko crosspoint is a 16mb binary and previously wasn't able to be OTA flashed on locked devices. 

From what I can see, from https://github.com/crosspoint-reader-ko/crosspoint-reader-ko/releases/tag/1.2.0-ko.15 onwards, crosspoint-reader-ko support locked devices.

I don't have something specific against this, just worry it will cause flood of other CJK-related requests.

itsthisjustin:
> > Stock doesn't handle korean well as far as I know but could be wrong. Ko crosspoint is a 16mb binary and previously wasn't able to be OTA flashed on locked devices.
> 
> From what I can see, from [crosspoint-reader-ko/crosspoint-reader-ko@`1.2.0-ko.15` (release)](https://github.com/crosspoint-reader-ko/crosspoint-reader-ko/releases/tag/1.2.0-ko.15) onwards, crosspoint-reader-ko support locked devices.
> 
> I don't have something specific against this, just worry it will cause flood of other CJK-related requests.

Now that we know x4 supports sd flashing, it may be a non-issue. Agree with flood of requests but also think for ~500 loc it rounds out our codebase. I'm open to discussion on this. The zip decompression fix and the bmp optimization I think are worth adding regardless and past that it's just some kerning rules for CJK glyphs

Uri-Tauber:
> Now that we know x4 supports sd flashing, it may be a non-issue. Agree with flood of requests but also think for ~500 loc it rounds out our codebase. I'm open to discussion on this. The zip decompression fix and the bmp optimization I think are worth adding regardless and past that it's just some kerning rules for CJK glyphs

Sure. Can you create seperate PR for the two fixes. 
Not going to insist on this, although I would love to hear what other maintainers think about this.

itsthisjustin:
Update after speaking with our korean users

The KO fork has apparently fallen behind CrossPoint and the maintainer has gotten busy and keeping in sync has become a large chore. So I think this does add a lot of value here as a baseline before taking the compatibility any futher

Uri-Tauber:
OK.
Good point.

Uri-Tauber:
Can you solve the cppcheck errors?

Also, can some Korean speakers try this to make sure everything works as intended, before we start to review?

itsthisjustin:
> Can you solve the cppcheck errors?
> 
> Also, can some Korean speakers try this to make sure everything works as intended, before we start to review?

This has been tested by a korean user and validated to be working. I worked 1:1 with them for hours iterating on this until it was ready

Uri-Tauber:
great.
I hope I can review it today.

lib/Epub/Epub/ParsedText.cpp
```diff
@@ -24,6 +24,8 @@ constexpr size_t RTL_PARAGRAPH_PROBE_WORDS = 3;
 // Per-word: scan enough chars to see through leading neutrals (quotes, numbers)
 // before giving up. 64 is a hedge for pathological cases like long numeric tokens.
 constexpr int RTL_PER_WORD_PROBE_DEPTH = 64;
+constexpr size_t MIN_JUSTIFY_GAPS = 2;
```
Uri-Tauber: Nitpick: any reason why `2`?
The old code (line 702) used `actualGapCount >= 1`, so we changing the behaviour here for lines with only two words.

lib/Epub/Epub/Section.cpp
```diff
@@ -10,7 +10,7 @@
 #include "parsers/ChapterHtmlSlimParser.h"
 
 namespace {
-constexpr uint8_t SECTION_FILE_VERSION = 25;
+constexpr uint8_t SECTION_FILE_VERSION = 30;
```
Uri-Tauber:
```suggestion
constexpr uint8_t SECTION_FILE_VERSION = 26;
```

itsthisjustin: Yeah can do. It was from me testing my own bumps and I forgot to reset this back down

lib/EpdFont/SdCardFont.cpp
```diff
@@ -516,59 +519,94 @@ bool SdCardFont::load(const char* path) {
...
+        freeAll();
+        return false;
+      }
       for (uint32_t j = 0; j < s.header.intervalCount; ++j) {
-        const auto& iv = s.fullIntervals[j];
-        if (iv.first > iv.last) {
-          LOG_ERR("SDCF", "Style %u: invalid interval %u (first 0x%lX > last 0x%lX)", i, j,
-                  static_cast<unsigned long>(iv.first), static_cast<unsigned long>(iv.last));
-          file.close();
+        if (file.read(reinterpret_cast<uint8_t*>(&iv), sizeof(iv)) != sizeof(iv)) {
```
Uri-Tauber, line 589: Not sure about this one, but I think every interval is read from the SD card twice - first time in line 544, second here. Do we need this? maybe we can cache the value (assuming it will not exhaust the RAM...)?
What do you think?

Uri-Tauber:
Overall, solid code. 
I want to test on my X4, to make sure no regressions for Hebrew, or parsing time. Will do tomorrow.

itsthisjustin:
> Overall, solid code. I want to test on my X4, to make sure no regressions for Hebrew, or parsing time. Will do tomorrow.

ngl I think it's even faster haha. I only saw improvements on this branch

lib/EpdFont/SdCardFont.cpp
itsthisjustin: Yeah it does read them twice, but that's on purpose. We don't know if we can use the 6-byte table until we've looked at every interval, so the first pass is just validate + figure that out, then the second pass actually fills the array. Caching them instead would mean holding up to 48KB in RAM at once which is the exact thing this change was trying to get rid of. The re-read is cheap anyway, it's sequential so SdFat serves it from its buffer, and it only happens once when the font loads. So a few ms instead of a big RAM spike.

Uri-Tauber: ok. thanks for clarifing this.

Uri-Tauber:
Tested on my device and noticed an alignment bug. In the book I'm currently reading, some lines are correctly justified while others are right-aligned. Not sure what's causing it.

[master.bmp](https://github.com/user-attachments/files/28832247/master.bmp)
[pr-2288.bmp](https://github.com/user-attachments/files/28832255/pr-2288.bmp)

itsthisjustin:
> Tested on my device and noticed an alignment bug. In the book I'm currently reading, some lines are correctly justified while others are right-aligned. Not sure what's causing it.
> 
> [master.bmp](https://github.com/user-attachments/files/28832247/master.bmp)
> [pr-2288.bmp](https://github.com/user-attachments/files/28832255/pr-2288.bmp)
> 

What alignment settings do you use?

Uri-Tauber:
> What alignment settings do you use?

Book's style + Embedded style OFF.
Haven't tried with other settings. 

It also happens with english books, but less frequently. (probably because the hyphenation)

Uri-Tauber:
Did you manage to reproduce the bug?

itsthisjustin:
@Uri-Tauber can you try this again on latest commit?

Uri-Tauber:
Solved the bug!
Thx!

itsthisjustin:
> Solved the bug!
> Thx!

Awesome. Let's get this in prior to release? It prevents crashes on cjk epubs

Uri-Tauber:
Absolutely. Can you solve the conflicts?

Uri-Tauber (approved):

uxjulia (approved):

PR #2290, opened 2026-06-08
feat: add drag-and-drop to upload modal

krunkosaurus:
Adds drag-and-drop to the file manager's upload modal — you can drop files straight onto the modal instead of clicking through the file picker.

Dropped files are handed to the existing file input, so nothing downstream changes: the **Optimize EPUB** conversion and its advanced settings still work exactly as before, along with multi-file/batch uploads and the WebSocket transfer (with the HTTP fallback). Click-to-browse is untouched.

Upload modal with drop zone

The change is contained to `FilesPage.html`: a dashed drop zone around the file input plus the drag listeners. It highlights on drag-over (a depth counter avoids flicker over child elements), replaces the current selection like the native picker does, and ignores drops while an upload is already in progress. No server or protocol changes.

Tested on an X4: dropping a single EPUB, a non-EPUB, and several files at once all populate the picker correctly, the Optimize EPUB panel still appears for EPUBs, drag-over highlighting behaves, and uploads complete.


Uri-Tauber (dismissed):
Very Nice!
One nitpick: The hint says "click to browse" but clicking the padding area or the hint text itself won't open the file picker. Only clicking directly on "Choose Files" works.

I haven't tested on my device - just reviewing the code.

krunkosaurus:
Thanks for the review, @Uri-Tauber! 🙏

Addressed the nitpick in `b096cb9` (plus a follow-up for consistency):

- **Click-to-browse now works across the whole drop zone**, not just the native "Choose Files" button. Added a click handler on `.drop-zone` (and `cursor: pointer`); the hint's `pointer-events: none` makes its clicks retarget to the zone, and clicks on the native button are skipped so the picker never opens twice.
- **Made the drop zone inert during in-flight uploads** (per CodeRabbit): the click handler and the dragover hover highlight now bail out while `isUploadInProgress`, matching the existing drop-handler guard and the disabled upload/close buttons. Extracted a shared `uploadBusy()` helper to keep the checks consistent.

Note: I reviewed the logic but haven't flashed it to a device yet. Mind giving it a re-approval when you get a chance? The earlier approval was auto-dismissed by the new commits.

Uri-Tauber:
@krunkosaurus Thanks for addressing my feedback.
Can you flash on your device and verify it's working as expected?

krunkosaurus:
https://github.com/user-attachments/assets/9702c8dd-64bb-4dc9-a7a6-83649a6daa7e

Confirmed all good. Even can drop more than one file. Video attached

Uri-Tauber (dismissed):

src/network/html/FilesPage.html
```diff
@@ -3023,6 +3047,68 @@ <h3>📂 Move File</h3>
...
+      // button and the drop handler's guard below.
+      const uploadBusy = () =>
+        typeof isUploadInProgress !== 'undefined' && isUploadInProgress;
+
+      // "click to browse" should work anywhere in the zone, not just on the
+      // native button. The hint has pointer-events:none so its clicks retarget
+      // to the zone; clicks on the native button retarget to fileInput (shadow
+      // DOM) and are skipped here so the picker isn't opened twice.
+      dropZone.addEventListener('click', function(e) {
+        if (uploadBusy()) return;
```
uxjulia, line 3074: This guard only covers clicks handled by the drop-zone wrapper. A direct click on the native file input can still open the picker during an active upload (I was able to do this in my local testing), then `validateFile()` can re-enable the upload button. We should also protect the actual upload path, e.g. disable `fileInput` while `isUploadInProgress` and add an early `if (isUploadInProgress) return;` in `uploadFile()`.

uxjulia (approved):

Uri-Tauber (approved):

PR #2294, opened 2026-06-08
chore: Initial multi-core compatibility

Uri-Tauber:
## Summary

* **What is the goal of this PR?** This is the first step towards ESP32-S3 support (M5stack paper, Lilygo T5, de-link, Morphy M3...). I focused here solely on support for multi-core devices. 

##   Changes Made 

  Problem 1: `requestedUpdate` data race

  - `ActivityManager.h`: Changed `bool requestedUpdate` to `std::atomic<bool> requestedUpdate{false}`
  - `ActivityManager.cpp`: Changed `if (requestedUpdate) { requestedUpdate = false; ...}` to `if (requestedUpdate.exchange(false)) { ... }` — a single atomic
  read-modify-write that's safe across cores

  Problem 2: Critical sections pass `nullptr` (`no-op` on dual-core)

  - `ActivityManager.cpp`: Added static `portMUX_TYPE activityManagerSpinlock = portMUX_INITIALIZER_UNLOCKED`;
  - Both `taskENTER_CRITICAL(nullptr)` / `taskEXIT_CRITICAL(nullptr)` call sites now use `&activityManagerSpinlock` — this provides actual cross-core mutual
  exclusion on ESP32-S3 (spinlock) while remaining backward-compatible on ESP32-C3 (just disables interrupts)

  Problem 3: No core affinity

  - `ActivityManager.cpp`: Changed `xTaskCreate` to `xTaskCreatePinnedToCore(..., 0)` — pins render task to core 0 (`PRO_CPU`). On ESP32-S3 Arduino, the main loop runs
   on core 1 by default, so the two tasks get clean separation without contention

  Problem 4: Unprotected singletons

  - `CrossPointSettings.h/.cpp`: Added mutable `std::mutex _mutex` + `getMutex()` accessor. Protected `saveToFile()`, `loadFromFile()`, and `loadFromBinaryFile()` with
  `std::lock_guard`
  - `CrossPointState.h/.cpp`: Same pattern — mutex guards around `saveToFile()`, `loadFromFile()`, and `loadFromBinaryFile()`

---

### AI Usage

### DIFF
--- a/src/CrossPointSettings.cpp
+++ b/src/CrossPointSettings.cpp
@@ -6,6 +6,7 @@
 #include <Serialization.h>
 
 #include <cstring>
+#include <mutex>
 #include <string>
 
 #include "I18nKeys.h"
@@ -96,6 +97,7 @@ uint8_t CrossPointSettings::sleepTimeoutEnumToMinutes(const uint8_t legacyValue)
 }
 
 bool CrossPointSettings::saveToFile() const {
+  std::lock_guard<std::mutex> lock(_mutex);
   Storage.mkdir("/.crosspoint");
   return JsonSettingsIO::saveSettings(*this, SETTINGS_FILE_JSON);
 }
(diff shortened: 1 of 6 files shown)


Uri-Tauber:
@itsthisjustin Moving the discussion here.

I can't say I read the entire SDK, so there is a slim chance it is handled there. 

The point of this PR is to add several mutexes and lock to handle race conditions between cores - something which we didn't need to worry about with the ESP32-C3, because it have only one core, but can cause very-hard-to-debug problems with two cores.

itsthisjustin:
> @itsthisjustin Moving the discussion here.
> 
> I can't say I read the entire SDK, so there is a slim chance it is handled there.
> 
> The point of this PR is to add several mutexes and lock to handle race conditions between cores - something which we didn't need to worry about with the ESP32-C3, because it have only one core, but can cause very-hard-to-debug problems with two cores.

Ok lets get it in. I already merged a pixel thing I was doing in the other pr and we can merge this and i'll just redo my pr

itsthisjustin (approved):

PR #2304, opened 2026-06-09
fix: Restore first-line paragraph indentation

### LINKED ISSUE #2241: No first-line indentation of paragraphs
### Affected Version

1.3.0

### Describe the Bug

After updating from 1.1.0 to 1.3.0 the first line of a paragraph was no longer indented (when extra paragraph spacing is turned off). This is contrary to the manual, which says: 'Extra Paragraph Spacing: [...] "OFF" - Paragraphs will not have vertical space added, but will have first-line indentation). It would greatly improve the reading experience if this property is restored. 

[no-par-indent.bmp](https://github.com/user-attachments/files/28500876/no-par-indent.bmp)

### Steps to Reproduce

1. Go to Settings > Reader
2. Select 'Embedded Style': OFF
3. Select 'Extra Paragraph Spacing': OFF
4. Inspect book text

### Expected Behavior

If extra paragraph spacing is turned off, the first line of each paragraph should be indented with a small amount of white space (see manual)

```shell

```

Uri-Tauber:
## Summary

* **What is the goal of this PR?** Should fix #2241
 
### Problem

Users upgrading from v1.1.0 to v1.3.0 reported that first-line paragraph indentation disappeared when both "Embedded Style" and "Extra Paragraph Spacing" were set to OFF — contrary to documented behavior, which states that disabling extra paragraph spacing should produce first-line indentation instead.

The root cause was a combination of two issues introduced in v1.3.0:

1. `resolveFirstLineIndent()` only returned a non-zero indent when CSS `text-indent` was explicitly defined in the stylesheet. With "Embedded Style" OFF, no CSS `text-indent` existed, so it always returned `0`.
2. The fallback path in `applyParagraphIndent()` tried to compensate by prepending a literal Em Space character (`U+2003`, `\xe2\x80\x83`) directly to the first word string. This caused downstream formatting bugs — the pixel layout pipeline still believed the first-line indent was `0`, so line start coordinates were placed flush with the left margin, and the hidden character interfered with line splitting and alignment.

### Fix

**Removed** `ParsedText::applyParagraphIndent()` entirely — the "hidden character hack" approach is gone.

**Updated** `ParsedText::resolveFirstLineIndent()` to accept `GfxRenderer` and `fontId` parameters so it can measure real pixel widths. It now handles all three cases cleanly:

- **CSS `text-indent` defined** → returns the explicit CSS value (existing behavior preserved).
- **No CSS `text-indent`, `extraParagraphSpacing` OFF** → queries `GfxRenderer::getTextAdvanceX()` for the pixel width of the Em Space glyph in the current font, and returns that as the indent. This feeds directly into the standard layout pipeline's X-coordinate offset.
- **`extraParagraphSpacing` ON, or non-natural alignment** → returns `0` (no indent).

All three `resolveFirstLineIndent` call sites (`computeLineBreaks`, `computeHyphenatedLineBreaks`, `extractLine`) are updated to pass through `renderer` and `fontId`.

---

### AI Usage

Did you use AI tools to help write this code? _**< YEP >**_

### DIFF
--- a/lib/Epub/Epub/ParsedText.cpp
+++ b/lib/Epub/Epub/ParsedText.cpp
@@ -261,10 +261,19 @@ void ParsedText::addWord(std::string word, const EpdFontFamily::Style fontStyle,
   }
 }
 
-int ParsedText::resolveFirstLineIndent(const bool isFirstLine) const {
-  if (isFirstLine && blockStyle.textIndentDefined && (blockStyle.textIndent < 0 || !extraParagraphSpacing) &&
-      isNaturalAlign) {
-    return blockStyle.textIndent;
+int ParsedText::resolveFirstLineIndent(const bool isFirstLine, const GfxRenderer& renderer, const int fontId) const {
+  if (!isFirstLine || !isNaturalAlign) {
+    return 0;
+  }
+  if (blockStyle.textIndentDefined) {
+    if (blockStyle.textIndent < 0 || !extraParagraphSpacing) {
+      return blockStyle.textIndent;
+    }
+    return 0;
+  }
+  if (!extraParagraphSpacing) {
+    // Fallback: Use EmSpace width for visual indent when CSS text-indent is not defined
+    return renderer.getTextAdvanceX(fontId, "\xe2\x80\x83", EpdFontFamily::REGULAR);
   }
   return 0;
 }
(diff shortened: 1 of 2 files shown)


Uri-Tauber:
Correct fix in #2320

PR #2305, opened 2026-06-09
fix: prevent progress.bin corruption from interrupted writes (#2275)

### LINKED ISSUE #2275: Book cache not updating and can not be cleared
### Affected Version

1.3.0

### Describe the Bug

I've had this happen a few times while reading a book. The book cache does not update, so the book progress does not get updated. Trying to `Delete Book Cache` on the individual book quietly fails. Going to `Settings->System->Clear Reading Cache` will clean all books and error on the book that's not updating.

### Steps to Reproduce

I can not reliably reproduce getting into this state. When the SD card is in this state the following behavior can be observed.
1. Open book and note what page you are on
2. Go forward or back a page (or more)
3. Exit book to main screen or sleep reader
4. Return to book from main screen or wake from sleep
5. Book will open at previous safe point and not where you changed to in step 2.

Failure to clear book cache individually.
1. Open book
2. Press `select` button
3. Select `Delete Book Cache`
4. There will be no error, but safe point will still not update.

Failure to clear book cache from Settings.
1. From `main page` select `Settings`
2. Navigate to `System` 
3. Select `Clear Reading Cache`
4. At least 1 book will error in the resulting page.

### Expected Behavior

- Reading progress would continue to work.
- Clearing the book cache should always work to fix this.

### Work around / Fix

1. Power off reader
2. Remove SD card and insert into a computer (instructions for Linux)
3. Manually remove cache files
4. umount partition and fsck disk
5. Remove SD card from computer and reinsert into device
6. Wake up device and all should be working as expected

See example below.

### Example SD Card Repair

```shell
andy@tora:~$ lsblk
NAME        MAJ:MIN RM   SIZE RO TYPE MOUNTPOINTS
sda           8:0    1  29.1G  0 disk 
└─sda1        8:1    1  29.1G  0 part /media/andy/D823-C2D3
nvme0n1     259:0    0 953.9G  0 disk 
├─nvme0n1p1 259:1    0   512M  0 part /boot/efi
├─nvme0n1p2 259:2    0 714.3G  0 part /home
├─nvme0n1p3 259:3    0   977M  0 part [SWAP]
└─nvme0n1p4 259:4    0 238.1G  0 part /
andy@tora:~$ cd /media/andy/D823-C2D3/.crosspoint/epub_1214177769/
andy@tora:/media/andy/D823-C2D3/.crosspoint/epub_1214177769$ ls -l
total 112
-rw-r--r-- 1 andy andy  1597 Dec 31 16:00 book.bin
-rw-r--r-- 1 andy andy  1624 Dec 31 16:00 css_rules.cache
-rw-r--r-- 1 andy andy   904 Dec 31 16:00 img_2_0.jpg
-rw-r--r-- 1 andy andy   744 Dec 31 16:00 img_2_0.pxc
-rw-r--r-- 1 andy andy     6 Dec 31 16:00 progress.bin
drwxr-xr-x 2 andy andy 16384 Dec 31 16:00 sections
-rw-r--r-- 1 andy andy  5486 Dec 31 16:00 thumb_226.bmp
andy@tora:/media/andy/D823-C2D3/.crosspoint/epub_1214177769$ rm -rf *^C
andy@tora:/media/andy/D823-C2D3/.crosspoint/epub_1214177769$ cd
andy@tora:~$ sudo umount /dev/sda1
andy@tora:~$ sudo fsck /dev/sda1
fsck from util-linux 2.39.3
fsck.fat 4.2 (2021-01-31)
There are differences between boot sector and its backup.
This is mostly harmless. Differences: (offset:original/backup)
  65:00/01
1) Copy original to backup
2) Copy backup to original
3) No action
[123?q]? 1
/.crosspoint/epub_1214177769/PROGRESS.BIN
  Contains a free cluster (17048). Assuming EOF.
/.crosspoint/epub_1214177769/PROGRESS.BIN
  File size is 6 bytes, cluster chain length is 0 bytes.
  Truncating file to 0 bytes.
Reclaimed 1 unused cluster (16384 bytes).
Free cluster summary wrong (1889612 vs. really 1889613)
1) Correct
2) Don't correct
[12?q]? 1

*** Filesystem was changed ***
The changes have not yet been written, you can still choose to leave the
filesystem unmodified:
1) Write changes
2) Leave filesystem unchanged
[12?q]? 1
/dev/sda1: 122 files, 17587/1907200 clusters
```

peuic:
## Summary

  Fixes #2275. A book could get stuck reopening on an old page, with progress no
  longer saving and neither "Delete Book Cache" nor "Clear Reading Cache" able to
  fix it. Root cause: `progress.bin` was written truncate-in-place, so an
  interrupted write (power loss, or a crash mid-SPI during sleep) left it with a
  broken FAT cluster chain that the firmware could neither rewrite nor delete —
  recovery required `fsck`/manual deletion on a host PC.

  Confirmed in the SDK: `SDCardManager::openFileForWrite` opens with
  `O_RDWR | O_CREAT | O_TRUNC`, so the canonical file is zeroed before the few
  progress bytes are rewritten — exactly the window that corrupts the FAT chain.

  * New shared helper `ProgressFile::writeAtomic()`
    (`src/activities/reader/ProgressFile.h`): writes progress to
    `progress.bin.tmp`, flushes and closes it, then `remove`s the old
    `progress.bin` and `rename`s the temp into place. An interrupted write now
    only ever damages the throwaway temp; the canonical file is never torn.
  * All three readers route their progress saves through the helper:
    EPUB (`EpubReaderUtils.h`), `TxtReaderActivity`, `XtcReaderActivity` — they
    all shared the identical vulnerable pattern.
  * Minor: `EpubReaderUtils::saveProgress` now takes `const Epub&` (clears a
    cppcheck `constParameterReference` finding).

## Additional Context

* **Crash-safe, not metadata-atomic.** On FAT the replace is `remove` + `rename`
  (two directory ops; SdFat's `rename` won't overwrite, hence remove-first). A
  crash between them leaves *neither* file, which reads as "no saved progress" on
  next launch — a harmless reset to an old page, never a corrupt/unclearable file.
  The guarantee is that `progress.bin` is never half-written.
* **Prevents, does not repair.** This stops new corruption on healthy cards. It
  cannot fix an already-corrupted `progress.bin` (removing it may itself fail at
  the FAT level) — those still need `fsck`/manual deletion, as in the issue's
  workaround.
* **Known follow-up (out of scope here):** a crash *while writing the temp* can
  leave an orphan `progress.bin.tmp`. It's harmless and self-healing (the next
  save overwrites it, and it never blocks reading progress), but a boot-time
  orphan-`.tmp` cleanup would be a tidy follow-up.
* **Focus areas for review:** the close-before-rename ordering in
  `ProgressFile.h` and the remove-before-rename rationale.

## Verification

* `./bin/clang-format-fix` — clean
* `pio check --fail-on-defect low --fail-on-defect medium --fail-on-defect high` — no defects
* `pio run` — SUCCESS (RAM 30.9%, Flash 78.8%; footprint essentially unchanged)
* Tested on a **Xteink X4** device: open book, turn pages, sleep/exit, reopen —
  progress now restores to the navigated page across all three readers
  (EPUB / TXT / XTC).

---

### AI Usage

### DIFF
--- a/src/activities/reader/TxtReaderActivity.cpp
+++ b/src/activities/reader/TxtReaderActivity.cpp
@@ -427,7 +427,9 @@ void TxtReaderActivity::saveProgress() const {
   data[1] = (currentPage >> 8) & 0xFF;
   data[2] = 0;
   data[3] = 0;
-  ProgressFile::writeAtomic(txt->getCachePath(), data, sizeof(data));
+  if (!ProgressFile::writeAtomic(txt->getCachePath(), data, sizeof(data))) {
+    LOG_ERR("TRS", "Failed to save progress: page %d", currentPage);
+  }
 }
 
 void TxtReaderActivity::loadProgress() {
--- a/src/activities/reader/TxtReaderActivity.cpp
+++ b/src/activities/reader/TxtReaderActivity.cpp
@@ -11,6 +11,7 @@
 #include "CrossPointSettings.h"
 #include "CrossPointState.h"
 #include "MappedInputManager.h"
+#include "ProgressFile.h"
 #include "ReaderUtils.h"
 #include "RecentBooksStore.h"
 #include "components/UITheme.h"
@@ -421,15 +422,12 @@ void TxtReaderActivity::renderStatusBar() const {
 }
 
 void TxtReaderActivity::saveProgress() const {
-  HalFile f;
-  if (Storage.openFileForWrite("TRS", txt->getCachePath() + "/progress.bin", f)) {
-    uint8_t data[4];
-    data[0] = currentPage & 0xFF;
-    data[1] = (currentPage >> 8) & 0xFF;
-    data[2] = 0;
-    data[3] = 0;
-    f.write(data, 4);
-  }
+  uint8_t data[4];
+  data[0] = currentPage & 0xFF;
+  data[1] = (currentPage >> 8) & 0xFF;
+  data[2] = 0;
+  data[3] = 0;
+  ProgressFile::writeAtomic(txt->getCachePath(), data, sizeof(data));
 }
 
 void TxtReaderActivity::loadProgress() {
--- a/src/activities/reader/XtcReaderActivity.cpp
+++ b/src/activities/reader/XtcReaderActivity.cpp
@@ -381,7 +381,9 @@ void XtcReaderActivity::saveProgress() const {
   data[1] = (currentPage >> 8) & 0xFF;
   data[2] = (currentPage >> 16) & 0xFF;
   data[3] = (currentPage >> 24) & 0xFF;
-  ProgressFile::writeAtomic(xtc->getCachePath(), data, sizeof(data));
+  if (!ProgressFile::writeAtomic(xtc->getCachePath(), data, sizeof(data))) {
+    LOG_ERR("XTR", "Failed to save progress: page %lu", currentPage);
+  }
 }
 
 void XtcReaderActivity::loadProgress() {
--- a/src/activities/reader/XtcReaderActivity.cpp
+++ b/src/activities/reader/XtcReaderActivity.cpp
@@ -17,6 +17,7 @@
 #include "CrossPointSettings.h"
 #include "CrossPointState.h"
 #include "MappedInputManager.h"
+#include "ProgressFile.h"
 #include "ReaderUtils.h"
 #include "RecentBooksStore.h"
 #include "XtcReaderChapterSelectionActivity.h"
@@ -375,16 +376,12 @@ void XtcReaderActivity::renderPage() {
 }
 
 void XtcReaderActivity::saveProgress() const {
-  HalFile f;
-  if (Storage.openFileForWrite("XTR", xtc->getCachePath() + "/progress.bin", f)) {
-    uint8_t data[4];
-    data[0] = currentPage & 0xFF;
-    data[1] = (currentPage >> 8) & 0xFF;
-    data[2] = (currentPage >> 16) & 0xFF;
-    data[3] = (currentPage >> 24) & 0xFF;
-    f.write(data, 4);
-    f.close();
-  }
+  uint8_t data[4];
+  data[0] = currentPage & 0xFF;
+  data[1] = (currentPage >> 8) & 0xFF;
+  data[2] = (currentPage >> 16) & 0xFF;
+  data[3] = (currentPage >> 24) & 0xFF;
+  ProgressFile::writeAtomic(xtc->getCachePath(), data, sizeof(data));
 }
 
 void XtcReaderActivity::loadProgress() {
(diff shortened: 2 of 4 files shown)


uxjulia:
I'm good with this approach. Structurally, it's similar to the EPUB progress-save hardening I added in my CrossInk fork: write the new progress to a temp file first, close it, then swap it into place so `progress.bin` is never rewritten in-place.

CrossInk has a couple extra hardening pieces on top: it keeps the previous `progress.bin` as a `.bak` fallback, checks `sync()`/`close()`, and can restore the backup if the final rename fails. I don't think that needs to block this PR, since this already removes the dangerous truncate-in-place write path. The backup/restore behavior can be a follow-up improvement PR if we want to make the recovery path safer.

uxjulia (dismissed):

src/activities/reader/TxtReaderActivity.cpp
```diff
@@ -421,15 +422,12 @@ void TxtReaderActivity::renderStatusBar() const {
...
-    data[2] = 0;
-    data[3] = 0;
-    f.write(data, 4);
-  }
+  uint8_t data[4];
+  data[0] = currentPage & 0xFF;
+  data[1] = (currentPage >> 8) & 0xFF;
+  data[2] = 0;
+  data[3] = 0;
+  ProgressFile::writeAtomic(txt->getCachePath(), data, sizeof(data));
```
Uri-Tauber: the caller ignore the return value of `ProgressFile::writeAtomic`. 
Not a blocker, but `EpubReaderUtils::saveProgress` correctly check the return value, and it would be a good practice to treats failure here too.

src/activities/reader/XtcReaderActivity.cpp
```diff
@@ -375,16 +376,12 @@ void XtcReaderActivity::renderPage() {
...
-    data[3] = (currentPage >> 24) & 0xFF;
-    f.write(data, 4);
-    f.close();
-  }
+  uint8_t data[4];
+  data[0] = currentPage & 0xFF;
+  data[1] = (currentPage >> 8) & 0xFF;
+  data[2] = (currentPage >> 16) & 0xFF;
+  data[3] = (currentPage >> 24) & 0xFF;
+  ProgressFile::writeAtomic(xtc->getCachePath(), data, sizeof(data));
```
Uri-Tauber: Same

Uri-Tauber (dismissed):
Approve. The code is clean and DRY.
Two nitpicks

src/activities/reader/TxtReaderActivity.cpp
peuic: thanks for the review! i worked on both cases!

Uri-Tauber (approved):

uxjulia (approved):

PR #2320, opened 2026-06-10
fix: Restore first-line paragraph indentation

### LINKED ISSUE #2241: No first-line indentation of paragraphs

Uri-Tauber:
## Summary

* **What is the goal of this PR?** Fix #2241

## Additional Context

- Root cause: The paragraph indentation relied on prepending an Em Space character (U+2003) to the first word. Some SD card fonts don't include this glyph, so its advance width is 0, resulting in no visible indent.

- Fix: Replace the character-insertion approach with a pixel-based indent computed from getSpaceWidth * 3 (~1em), which works for any font since every font has a regular space glyph.

[before.bmp](https://github.com/user-attachments/files/28807502/before.bmp)

[after.bmp](https://github.com/user-attachments/files/28807505/after.bmp)

---

### AI Usage

### DIFF
--- a/lib/Epub/Epub/ParsedText.cpp
+++ b/lib/Epub/Epub/ParsedText.cpp
@@ -261,10 +261,18 @@ void ParsedText::addWord(std::string word, const EpdFontFamily::Style fontStyle,
   }
 }
 
-int ParsedText::resolveFirstLineIndent(const bool isFirstLine) const {
-  if (isFirstLine && blockStyle.textIndentDefined && (blockStyle.textIndent < 0 || !extraParagraphSpacing) &&
-      isNaturalAlign) {
-    return blockStyle.textIndent;
+int ParsedText::resolveFirstLineIndent(const bool isFirstLine, const GfxRenderer& renderer, const int fontId) const {
+  if (!isFirstLine || !isNaturalAlign) {
+    return 0;
+  }
+  if (blockStyle.textIndentDefined) {
+    if (blockStyle.textIndent < 0 || !extraParagraphSpacing) {
+      return blockStyle.textIndent;
+    }
+    return 0;
+  }
+  if (!extraParagraphSpacing) {
+    return renderer.getSpaceWidth(fontId, EpdFontFamily::REGULAR) * 3;
   }
   return 0;
 }
@@ -292,9 +300,6 @@ void ParsedText::layoutAndExtractLines(const GfxRenderer& renderer, const int fo
   isNaturalAlign =
       blockStyle.alignment == CssTextAlign::Justify ||
       (blockStyle.isRtl ? blockStyle.alignment == CssTextAlign::Right : blockStyle.alignment == CssTextAlign::Left);
-
-  // Apply fixed transforms before any per-line layout work.
-  applyParagraphIndent();
 
   // Ensure SD card font glyph metrics are loaded before measuring word widths.
   // For flash-based fonts isSdCardFont() returns false and this block is skipped
@@ -356,7 +361,7 @@ std::vector<size_t> ParsedText::computeLineBreaks(const GfxRenderer& renderer, c
     return {};
   }
 
-  const int firstLineIndent = resolveFirstLineIndent(true);
+  const int firstLineIndent = resolveFirstLineIndent(true, renderer, fontId);
 
   // Ensure any word that would overflow even as the first entry on a line is split using fallback hyphenation.
   for (size_t i = 0; i < wordWidths.size(); ++i) {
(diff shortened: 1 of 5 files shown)


itsthisjustin:
Couldn't we solve this with proper font fallback support instead?

Uri-Tauber:
Do you mean falling back to specific glyphs from built-in fonts when they're missing from the custom font?
Do we already support something like that? It feels quite RAM-intensive to me, though we'd need to test it to know for sure.

Either way, I see the current approach of injecting an em-space glyph as more of a hack than proper indentation. My implementation is cleaner and more robust. You can see that just by looking at how much code it removes.

uxjulia:
> Couldn't we solve this with proper font fallback support instead?

Yes. We could update the font builder to automatically inject the selected glyph for the user (using some standard fallback fonts for certain unicode ranges) instead of relying on the user to provide a font that has those glyphs. If they DO provide one that has those glyphs, our fallback font doesn't need to do any patching.

Based on the user's selected unicode coverage, we'd patch their provided font with those intervals using our pre-determined set of fallback font families that we know contain those intervals. Only downside is our standard set of fallback families might look very different from their chosen font (different x-height or widths might make the injected glyphs look out of place, but it's better than being completely missing). 

I can see what I can whip up in the font-builder. In the meantime, we can move forward with this fix for people who already have fonts and are running into this issue or wait... not sure what you guys want to do.

znelson (commented):
I wonder if we should just fall back to the 3x space width when em-space doesn't exist, since the em-space font metric is technically more correct when it _is_ part of the font?

Uri-Tauber:
@znelson Wouldn't this complicate the logic unnecessarily? I'm not a fan of injecting new characters—it feels a bit fragile to me.
An em space is equivalent to 3–4 regular spaces. Using three regular spaces should be indistinguishable in most fonts.

uxjulia:
Whoops, this PR got auto-closed when Justin merged in my PR for the website's font-builder. However, I think the proper fix is to have the SD card fonts include the em dash glyph (and some other common punctuation similar to em dashes) by default to mirror what Crosspoint's built-in fonts already do, which my website PR did for user-built fonts. The only remaining gap now is the downloadable SD card fonts that don't already have those glyphs, which can follow the same pattern as the website's SD font builder which is to automatically inject it when a font is missing that glyph.

This removes responsibility from the firmware and handles it within the generated font.

Uri-Tauber:
Well, I still think my PR is cleaner implementation, but I'm obviously biased here.
My main concern is those users who already downloaded the fonts without em-space. (rough estimation would be 20-30k). We need to have a way to solve this bug without making them re-download everything.

Doing on-device fallback seems like a ram-killer to me. If someone wants to try this direction and do some benchmarking - would be appreciated.

uxjulia:
> Well, I still think my PR is cleaner implementation, but I'm obviously biased here.
> 
> My main concern is those users who already downloaded the fonts without em-space. (rough estimation would be 20-30k). We need to have a way to solve this bug without making them re-download everything.
> 
> 
> 
> Doing on-device fallback seems like a ram-killer to me. If someone wants to try this direction and do some benchmarking - would be appreciated. 

We definitely don't want to do on-device fallback (and my fix doesn't do that). And yes, I agree with your concern regarding the people who have already downloaded fonts (12 of the 20 don't include em dash) but an update to the downloadable fonts is just an in-app update for them through `Manage Fonts`. That would leave the people who created their own fonts, and it's easy enough that the solution is for them to re-build their font with the web font-builder. 

Moving forward, everyone that builds new fonts or downloads them through the firmware will then have all punctuation glyphs included and this patch will eventually become obsolete. 

Also, I should add that it is NOT standard practice to be using em dashes for paragraph indentations, that's not their purpose. So this fix is geared towards a very small subset of epubs that are improperly formatted + users who created fonts without the em dash. Just something to keep in mind. 

Edit: I conflated em dashes with em spaces. My bad. Although uncommon, em spaces are sometimes used to indent paragraphs for whatever reason.

Uri-Tauber:
> Moving forward, everyone that builds new fonts or downloads them through the firmware will then have all punctuation glyphs included and this patch will eventually become obsolete.

This is where I disagree with you. I think think fix make the code **more correct**, regardless to the question if it solve the bug or not.

uxjulia:
> > Moving forward, everyone that builds new fonts or downloads them through the firmware will then have all punctuation glyphs included and this patch will eventually become obsolete.
> 
> This is where I disagree with you. I think think fix make the code **more correct**, regardless to the question if it solve the bug or not.

Definitely not saying it's not correct, just saying if we patch all areas where user fonts can somehow get missing punctuation glyphs: built-in, downloaded, user-generated, then eventually they should never hit this code. However, if users can obtain `.cpfont` files from other sources (which is definitely a possibility in the future as people fork the font-builder) and those sources do things differently, then your fix would certainly address those cases.

uxjulia (approved):

itsthisjustin:
@Uri-Tauber how will this impact other devices with different resolutions or displays?

Uri-Tauber:
@itsthisjustin Irrelevant. We use `space` (U+0020) as a reference point, so it will differ between different font sizes.

itsthisjustin (approved):

PR #2324, opened 2026-06-11
fix: Hanging indent causes overlapping words

znelson:
## Summary

Hanging-indent paragraphs (CSS `text-indent: <negative>`) rendered with overlapping text. Small regression from the RTL refactor in #1700.

**Example**

"The Story of B" uses `p.hanging` for dialogue:

```css
p.hanging {
  margin-left: 2.5em;
  text-indent: -1em;  /* first line sticks out 1em to the left of the rest */
}
```

```html
<p class="hanging">ME: Hello.</p>
<p class="hanging">HIM: Fr. Osborne?</p>
```

Bug: `E:` is drawn on top of `H`, `M:` on top of `F`. The first word's x-position clamps to 0, but the cursor for subsequent words advances from the un-clamped (negative) base — collapsing the gap.

**Root cause**

#1700 consolidated the per-word positioning loops in `extractLine` and added a defensive clamp at each push site:

```cpp
lineXPos.push_back(static_cast<int16_t>(xpos < 0 ? 0 : xpos));
```

For RTL the clamp absorbs real underflow when `xpos -= wordWidth` runs past the left edge. The same clamp got applied to the LTR branch, where the *only* way `xpos` starts negative is `firstLineIndent < 0`, i.e. a legitimate hanging indent.

**Fix**

Drop the clamp at all three push sites (LTR, RTL, BiDi-reorder). Negative `xpos` is a valid relative coordinate that `BlockStyle::leftInset()` already absorbs when the line is positioned on screen.

**Screenshots**

| Before | After |
| -- | -- |
|  |  |

---

### AI Usage

### DIFF
--- a/lib/Epub/Epub/ParsedText.cpp
+++ b/lib/Epub/Epub/ParsedText.cpp
@@ -799,7 +799,7 @@ void ParsedText::extractLine(const size_t breakIndex, const int pageWidth, const
     }
 
     for (size_t wordIdx = 0; wordIdx < reorderedWidthsScratch.size(); wordIdx++) {
-      lineXPos.push_back(static_cast<int16_t>(xpos < 0 ? 0 : xpos));
+      lineXPos.push_back(static_cast<int16_t>(xpos));
       xpos += reorderedWidthsScratch[wordIdx];
 
       const bool nextIsContinuation =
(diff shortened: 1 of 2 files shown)


itsthisjustin:
Huh wonder why I've never seen this. Thanks for the fix Zach. @Uri-Tauber does this negatively impact rtl at all?

itsthisjustin (approved):

Uri-Tauber:
@itsthisjustin Need to check.

Uri-Tauber (approved):
LGTM. Tested on device and working fine.

PR #2331, opened 2026-06-12
fix: use Noto Sans as punctuation fallback when generating downloadable SD fonts

uxjulia:
## Summary

  * Ensure downloadable SD-card fonts include common punctuation/symbol glyphs even when the source font does not provide them by bundling in the missing glyphs during font generation.
  * Adds per-style fallback font arguments to `fontconvert_sdcard.py`.
  * Uses the fallback font during glyph availability checks and rasterization when the primary font is missing a codepoint.
  * Wires `build-sd-fonts.py` to pass the built-in Noto Sans Regular font as the fallback for generated SD fonts.

## Additional Context

* This keeps fallback resolution in the font generation step, so the device still reads normal `.cpfont` glyph data at runtime rather than doing an extra fallback lookup.
* Generated `.cpfont` files may grow where missing punctuation/symbols are now included from Noto Sans.
* Validation run: `python3 -m py_compile lib/EpdFont/scripts/build-sd-fonts.py lib/EpdFont/scripts/fontconvert_sdcard.py`.
---

### AI Usage

### DIFF
--- a/lib/EpdFont/scripts/build-sd-fonts.py
+++ b/lib/EpdFont/scripts/build-sd-fonts.py
@@ -46,6 +46,7 @@ DEFAULT_CONFIG = SCRIPT_DIR / "sd-fonts.yaml"
 DEFAULT_OUTPUT = SCRIPT_DIR / "output"
 DOWNLOAD_DIR = SCRIPT_DIR / "downloaded_fonts"
 INSTANCE_DIR = SCRIPT_DIR / "instanced_fonts"
+DEFAULT_FALLBACK_FONT = EPDFONTS_DIR / "builtinFonts/source/NotoSans/NotoSans-Regular.ttf"
 
 
 def download_font(url: str, dest: Path) -> Path:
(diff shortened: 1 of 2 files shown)


itsthisjustin (approved):

lib/EpdFont/scripts/build-sd-fonts.py
```diff
@@ -186,12 +187,14 @@ def build_family(
         # Multi-style mode
         for style_name, font_path in resolved_styles.items():
             cmd.extend([f"--{style_name}", str(font_path)])
+            cmd.extend([f"--fallback-{style_name}", str(DEFAULT_FALLBACK_FONT)])
```
Uri-Tauber, line 190: Do we want to use `NotoSans-Regular.ttf` for **every** style? Wouldn't it be better to use `NotoSans-Bold.ttf` etc...?

uxjulia: This is a fair question. This fallback is mostly for things like em dashes, smart quotes, ellipses, bullets, and spacing/formatting punctuation. A lot of that set is either visually very subtle at our render sizes or literally spacing/control behavior, so matching bold/italic weight exactly adds complexity for limited practical benefit and would also mean carrying additional fallback faces through the generation pipeline.

itsthisjustin: I'm in favor of less but functional here. Let the symbol show and if it doesn't happen to be bold, it's better than what exists today without bloating

Uri-Tauber: okay

Uri-Tauber (approved):

PR #2335, opened 2026-06-12
fix: Switch XTC cover BMP output from 1-bit to 2-bit grayscale

itsthisjustin:
Replace the 1-bit BMP header generation with a new 2-bit grayscale format that includes a 4-color palette (black, dark gray, light gray, white). This matches the layout used by cover converters and writes the complete header as a single buffer.

Previously XTC covers were generating 1 bit b&w images even on 2bit files. This is a bit of a minimal fix as the larger fix involved like 6 files and would have been much harder to review.

### DIFF
--- a/lib/Xtc/Xtc.cpp
+++ b/lib/Xtc/Xtc.cpp
@@ -173,37 +173,53 @@ bool Xtc::generateCoverBmp() const {
     return false;
   }
 
-  // Write 1-bit BMP header (top-down row order)
-  BmpHeader bmpHeader;
-  createBmpHeader(&bmpHeader, pageInfo.width, pageInfo.height, BmpRowOrder::TopDown);
-  coverBmp.write(reinterpret_cast<const uint8_t*>(&bmpHeader), sizeof(bmpHeader));
-
-  const uint32_t rowSize = ((pageInfo.width + 31) / 32) * 4;
-
-  // Write bitmap data
-  // BMP requires 4-byte row alignment
-  const size_t dstRowSize = (pageInfo.width + 7) / 8;  // 1-bit destination row size
-
   if (bitDepth == 2) {
-    // XTH 2-bit mode: Two bit planes, column-major order
+    // XTH 2-bit mode: preserve all 4 gray levels in a 2-bit BMP so the sleep
+    // screen's grayscale pass can render them (a 1-bit cover would silently
+    // disable it). Source is two bit planes, column-major order:
     // - Columns scanned right to left (x = width-1 down to 0)
     // - 8 vertical pixels per byte (MSB = topmost pixel in group)
     // - First plane: Bit1, Second plane: Bit2
-    // - Pixel value = (bit1 << 1) | bit2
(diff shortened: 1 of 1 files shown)


itsthisjustin:
@fain182 There's a pretty big loc count difference between our two PRs. Can you compare them?

fain182:
Thanks for flagging #2335 — I hadn't seen it, and it turns out we reached the same conclusions independently: same diagnosis, same {3, 1, 2, 0} XTH-to-palette table, same {255, 85, 170, 0} gray order for the thumbnail. Good confirmation that the level mapping is right, since an automated reviewer here pushed hard for {3, 2, 1, 0}.

One thing your version does better than mine: allocating the row buffer at the full padded row size and zeroing it, so each row goes out in a single write() with the padding included. Mine keeps a separate padding write, which is noisier for no benefit. I'll adopt that.

What this PR adds on top of yours:

- **Cache invalidation.** generateCoverBmp() returns early when cover.bmp exists, so on any device that has already opened the book the old 1-bit cover keeps being served and the fix is invisible. I hit exactly this while testing on hardware. Here the cover cache file is named after its depth (cover.bmp / cover_2bit.bmp), following the cover.bmp / cover_crop.bmp convention in Epub.
- **Thumbnail copy path** The scale >= 1.0f shortcut byte-copies the cover into the thumbnail cache, so once the cover is 2-bit, small pages produce 2-bit thumbnails against the documented 1-bit behaviour. The scale is clamped instead, and every thumbnail goes through the regular path.
- **Plane indexing** loadPage() reads two planes of ceil(width * height / 8) bytes, so column bits are packed continuously; the cover, thumbnail and reader all indexed them as if each column were padded to whole bytes. For a height that is not a multiple of 8 that reads bytes the parser nevehe reader, dropped pixels in the thumbnail. Allthree now use the packed offset (identical arithmetic for 8-aligned heights, so rendering is unchanged in practice).
- **Write failures** Every cover BMP write is checks removed rather than cached as valid.
- **BMP header** createBmp2BitHeader() sits next to the existing createBmpHeader() and shares its field-filling code, instead of a 70-byte array with patched offsets.

Your call which one goes in... if you wanted something small, go with yours. Worth taking the cache-name change with it though, otherwise devices that already opened the book keep serving the old 1-bit cover.

itsthisjustin:
I'll let @Uri-Tauber chime in here

lpla:
yo Justin, ty for putting this together 🙌 the level mapping is solid and this was super useful as an independent cross-check.

For the RC I'm leaning #3173 because it also handles existing `cover.bmp` caches, keeps thumbnails 1-bit, fixes packed-plane indexing, and removes partial cache files when a write fails.

pls leave this open for a sec as fallback while #3173 lands its tiny padded-row/single-write follow-up and gets one last pass. Once that's settled, we can close this as superseded. No approval/closure action from me yet.

Uri-Tauber:
I'm leaning towards this one. Mainly because I dont see the invalidation as such a big deal.
I didnt reviewed the code of either PR.

itsthisjustin:
@Uri-Tauber one of these should be merged before 1.6

PR #2341, opened 2026-06-14
fix: swap reader menu navigation direction in CCW/inverted (#2321)

### LINKED ISSUE #2321: Reader menu up/down does not work in CCW mode
### Affected Version

1.4.0

### Describe the Bug

When CCW is selected, pressing the up and down buttons just jumps between "Select Chapter" and "Delete Book Cache". The other items can not be selected.
Sometimes times the bug only shows up when you cyle through all the items once. 

Going to the settings and setting the orientation back to portrait fixes the problem temporarily.

### Steps to Reproduce

Select CCW.
Open a book.
Open the reader menu.
Press up/down

### Expected Behavior

Pressing the up/down button goes to next menu item. 

```shell

```

TheCyberRonin:
## Summary

Fixes #2321 — in `LANDSCAPE_CCW` and `INVERTED` orientations (with **Front buttons follow orientation** enabled), the reader menu's navigation buttons acted **opposite to their on-screen labels**. The hint labels were already swapped to match the rotated screen, but list/menu navigation was not, so pressing the button labeled "up" moved the selection down (and the list appeared to only jump between the first and last item).

## Root cause

The "control axis is flipped" condition (`frontButtonFollowOrientation && (orientation == INVERTED || orientation == LANDSCAPE_CCW)`) was already applied to:
- button hint **labels** in `MappedInputManager::mapLabels`, and
- front Left/Right **page-turn** direction in `ReaderUtils::detectPageTurn`,

but **not** to list navigation, which flows through `ButtonNavigator`'s fixed `getNextButtons()` / `getPreviousButtons()` sets. So labels swapped while behaviour didn't.

## Change

- Extract the condition into a single predicate `CrossPointSettings::isNavDirectionSwappedByOrientation()`.
- Apply it at the navigation consumption site: `ButtonNavigator::getNextButtons()` / `getPreviousButtons()` now swap their direction sets when it returns true (mirroring how `detectPageTurn` already swaps at its call site).
- Refactor the two existing duplicated conditions (`mapLabels`, `detectPageTurn`) to use the shared predicate — behaviour-neutral.

This centrally fixes every `ButtonNavigator`-driven list (reader menu, chapter select, bookmarks, home, settings, OPDS, recent books).

## Testing

- `pio run` builds clean (no warnings); RAM 30.9%, well within budget. Logic-only change, no measurable RAM/flash cost.
- `clang-format` (v21, matching CI) reports no changes.
- **Verified on-device** in `LANDSCAPE_CCW` with *Front buttons follow orientation* enabled: navigation now matches the labels and traverses the full menu. Portrait / Landscape CW unchanged; page-turning unaffected.

### DIFF
--- a/src/activities/reader/ReaderUtils.h
+++ b/src/activities/reader/ReaderUtils.h
@@ -43,9 +43,7 @@ inline PageTurnResult detectPageTurn(const MappedInputManager& input) {
   const bool usePress = SETTINGS.longPressButtonBehavior == SETTINGS.OFF;
   const bool tiltNext = SETTINGS.tiltPageTurn && halTiltSensor.wasTiltedForward();
   const bool tiltPrev = SETTINGS.tiltPageTurn && halTiltSensor.wasTiltedBack();
-  const bool swapFront =
-      SETTINGS.frontButtonFollowOrientation && (SETTINGS.orientation == CrossPointSettings::INVERTED ||
-                                                SETTINGS.orientation == CrossPointSettings::LANDSCAPE_CCW);
+  const bool swapFront = SETTINGS.isNavDirectionSwappedByOrientation();
   const auto prevButton = swapFront ? MappedInputManager::Button::Right : MappedInputManager::Button::Left;
   const auto nextButton = swapFront ? MappedInputManager::Button::Left : MappedInputManager::Button::Right;
   const bool prev =
--- a/src/activities/reader/ReaderUtils.h
+++ b/src/activities/reader/ReaderUtils.h
@@ -43,9 +43,8 @@ inline PageTurnResult detectPageTurn(const MappedInputManager& input) {
   const bool usePress = SETTINGS.longPressButtonBehavior == SETTINGS.OFF;
   const bool tiltNext = SETTINGS.tiltPageTurn && halTiltSensor.wasTiltedForward();
   const bool tiltPrev = SETTINGS.tiltPageTurn && halTiltSensor.wasTiltedBack();
-  const bool swapFront = input.isNavDirectionSwapped();
-  const auto prevButton = swapFront ? MappedInputManager::Button::Right : MappedInputManager::Button::Left;
-  const auto nextButton = swapFront ? MappedInputManager::Button::Left : MappedInputManager::Button::Right;
+  const auto prevButton = MappedInputManager::Button::NavPrevious;
+  const auto nextButton = MappedInputManager::Button::NavNext;
   const bool prev =
       tiltPrev ||
       (usePress ? (input.wasPressed(MappedInputManager::Button::PageBack) || input.wasPressed(prevButton))
--- a/src/activities/reader/ReaderUtils.h
+++ b/src/activities/reader/ReaderUtils.h
@@ -43,7 +43,7 @@ inline PageTurnResult detectPageTurn(const MappedInputManager& input) {
   const bool usePress = SETTINGS.longPressButtonBehavior == SETTINGS.OFF;
   const bool tiltNext = SETTINGS.tiltPageTurn && halTiltSensor.wasTiltedForward();
   const bool tiltPrev = SETTINGS.tiltPageTurn && halTiltSensor.wasTiltedBack();
-  const bool swapFront = SETTINGS.isNavDirectionSwappedByOrientation();
+  const bool swapFront = input.isNavDirectionSwapped();
   const auto prevButton = swapFront ? MappedInputManager::Button::Right : MappedInputManager::Button::Left;
   const auto nextButton = swapFront ? MappedInputManager::Button::Left : MappedInputManager::Button::Right;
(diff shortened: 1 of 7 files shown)


Uri-Tauber:
Hey, thanks for this — extracting `isNavDirectionSwappedByOrientation()` removes the three-way duplication cleanly.

One thing I’d reconsider is where the swap logic lives. `ButtonNavigator::getNextButtons()` is now reaching into `CrossPointSettings`, but that logic feels like it belongs with the rest of the logical-to-physical button mapping in `MappedInputManager`. `mapButton()` is already the single place where logical buttons are translated to physical GPIOs, and `mapLabels()` handles the orientation swap for labels there for the same reason.

Navigation is the same kind of concern. If you add `NavNext` and `NavPrevious` to `Button`, handle the orientation swap inside `mapButton()`, and have `ButtonNavigator::getNextButtons()` return `{NavNext}` unconditionally, then `ButtonNavigator` can stay settings-free and the orientation rule lives in one place. That also makes it easier for future callers to use the same behavior without needing to know about the swap.

TheCyberRonin:
I have no problem with changing that up, I'll adjust the PR this evening!

TheCyberRonin:
Reworked as suggested — the orientation swap now lives in the logical→physical mapping layer:

- Added `NavNext` / `NavPrevious` to `MappedInputManager::Button`.
- `mapButton()` resolves them to the side+front buttons and applies the direction swap there, alongside `mapLabels()`.
- `ButtonNavigator::getNextButtons()` / `getPreviousButtons()` now return `{NavNext}` / `{NavPrevious}` unconditionally, so `ButtonNavigator` is settings-free and back to header-only (no `.cpp` / `CrossPointSettings` dependency). The three-way duplication stays collapsed into one predicate.

While testing this on-device I hit a bug the original approach also had: the swap **bled into the portrait main menu** (and settings), flipping up/down where it shouldn't.

Root cause: the predicate keyed on the persisted `SETTINGS.orientation`, but that's just the *reader's* saved preference — it stays `LANDSCAPE_CCW` even while portrait UI is on screen. The reader rotates the renderer on enter and restores portrait on exit (`EpubReaderActivity.cpp:176`), and only reader-family activities ever rotate, so `SETTINGS.orientation` is the wrong signal outside the reader.

Fix: `isNavDirectionSwapped()` now keys on the **live renderer orientation** instead:

```cpp
bool MappedInputManager::isNavDirectionSwapped() const {
  const auto orientation = renderer.getOrientation();
  return SETTINGS.frontButtonFollowOrientation &&
         (orientation == GfxRenderer::PortraitInverted ||
          orientation == GfxRenderer::LandscapeCounterClockwise);
}
```

So the swap applies only when the screen is *actually* rendered rotated — the reader and its modal menus (exactly #2321) — and never in portrait UI like home/settings. It also corrects the same latent issue in the hint labels (`mapLabels`) on those screens. The cost is `MappedInputManager` holding a `const GfxRenderer&` so the mapping layer can see live orientation (rationale commented at the member).

A nice side effect: because the decision is derived from the live orientation rather than hardcoded to the reader, any future screen that renders rotated (e.g. a rotatable main menu) gets correct button-direction and label swapping for free — no per-activity flags to remember to set.

Verified on-device: portrait UI no longer swaps; reader menu navigation matches labels and traverses the full list in `LANDSCAPE_CCW` / `INVERTED`; page-turning and Portrait / Landscape CW unchanged.

src/activities/reader/ReaderUtils.h
```diff
@@ -43,9 +43,7 @@ inline PageTurnResult detectPageTurn(const MappedInputManager& input) {
   const bool usePress = SETTINGS.longPressButtonBehavior == SETTINGS.OFF;
   const bool tiltNext = SETTINGS.tiltPageTurn && halTiltSensor.wasTiltedForward();
   const bool tiltPrev = SETTINGS.tiltPageTurn && halTiltSensor.wasTiltedBack();
-  const bool swapFront =
-      SETTINGS.frontButtonFollowOrientation && (SETTINGS.orientation == CrossPointSettings::INVERTED ||
-                                                SETTINGS.orientation == CrossPointSettings::LANDSCAPE_CCW);
+  const bool swapFront = input.isNavDirectionSwapped();
   const auto prevButton = swapFront ? MappedInputManager::Button::Right : MappedInputManager::Button::Left;
   const auto nextButton = swapFront ? MappedInputManager::Button::Left : MappedInputManager::Button::Right;
```
Uri-Tauber: Can be simplified now to 
```suggestion
  const auto prevButton = MappedInputManager::Button::NavPrevious;
  const auto nextButton = MappedInputManager::Button::NavNext;
```

Uri-Tauber (dismissed):

Uri-Tauber (approved):

itsthisjustin (approved):

PR #2346, opened 2026-06-14
feat: file browser context menu (long-press) with Rename and Delete

jetaudio:
## Summary

Adds a **context menu to the file browser**, opened with a long-press of the
**Open** button on any entry. The menu currently offers two actions — **Rename**
and **Delete** — and is designed so further per-file actions can be added with a
single line.

This replaces the current behaviour where a long-press of Open goes straight to
the delete confirmation (only that dialog stands between a long press and an
irreversible delete). The long-press now surfaces an explicit menu, which is both
safer and more discoverable, and makes room for Rename.

Everything is rendered as **overlays composited over the file list**, so you can
still see the entry you are acting on behind the dialog.

> **Refreshed 2026-08-12:** rebased onto current `develop` (was conflicting against
> the old `master` base) and retargeted at `develop`. The branch is now a single
> commit on top of `develop`, the incidental translation-file churn from the old
> merge commit is gone, and the flash figure below has been re-measured.

## Motivation

- A long-press that deletes directly is easy to trigger by accident and offers
  no room to grow. A context menu is the conventional, extensible home for
  per-item actions.
- Renaming a book on-device previously required a computer / Calibre. Renaming
  is the most-requested companion to delete, so it ships in the same menu.

## Scope Check

- [x] The stock firmware does not already handle this well — it has long-press delete and no on-device rename.

### Opening the menu
- Highlight an entry, then **hold Open**. The menu appears the instant the 1 s
  long-press threshold is reached — you do **not** have to release the button
  first. The release that follows is swallowed so it cannot also activate the
  first menu item.
- A normal (short) press of Open still opens the file / enters the directory
  exactly as before.
- In the menu: **Up/Down** move the selection, **Select** activates, **Back**
  dismisses.

### Rename
- Opens the on-screen keyboard **prefilled with the editable base name only**.
  The extension is preserved automatically and re-appended on confirm, so users
  can't accidentally strip `.epub`/`.txt`/`.xtc` and make the book disappear
  from the (extension-filtered) browser.
- Directories are renamed in full (no extension). A leading dot on a hidden file
  is not treated as an extension separator.
- The old entry's reading cache is cleared before the rename. The cache is keyed
  by a hash of the file path, so a rename would otherwise orphan it on the SD
  card. After the rename the list refreshes and the selection follows the entry.
- Empty / unchanged names are ignored.

### Delete
- Selecting Delete shows a confirmation dialog (`Delete? <name>`) before anything
  is removed; it reuses the existing recursive `removeDirFile()` so deleting a
  directory still clears the whole subtree (and each book's cache) in one pass.
- On confirm, the file list **repaints immediately** — before the potentially
  slow cache clear — so the dialog never lingers on screen during the delete.

### `ContextMenuActivity` (new — `src/activities/util/`)
A small, reusable modal list rendered as a centred overlay box (white fill,
black border, thin halo). It takes a title and a list of `{StrId label, int
action}` items, returns the chosen action via `MenuResult`, and cancels on Back.
Navigation uses `ButtonNavigator`; the Up/Down hints are hidden when there is
only one item. It deliberately does not `clearScreen()` — it composites over the
caller's last frame.

### `ConfirmationActivity` — opt-in overlay mode
A new `overlay` constructor flag (default **false**) turns the full-screen prompt
into a centred dialog box composited over the caller's screen. In overlay mode:
- Cancel / Confirm sit on the **first two hint slots** (the Back and Confirm
  buttons); the Left/Right buttons are intentionally inert.
- A `lockNextConfirmRelease` guard swallows a Confirm release carried over from
  the Confirm-driven menu that launched the dialog, so it can't auto-confirm.

The default (full-screen) path is unchanged, so the existing callers — SD
firmware update, font download, and "remove from recents" — keep their current
appearance and their Left = Cancel / Right = Confirm controls.

### `FileBrowserActivity`
- Long-press detection moved from "on release, if held >= 1 s" to "while held,
  the moment held >= 1 s", guarded by a `contextMenuArmed` flag so it fires once
  per hold and the trailing release doesn't also open the entry.
- New `showContextMenu()` / `promptRename()` helpers; the delete path was
  factored into `confirmAndDelete()`.

## Overlay rendering / e-ink notes

Both overlays draw on top of the file list without clearing it. Because the file
browser's own bottom chrome (button hints + the path line) would otherwise show
through alongside the dialog's own hints, each overlay first wipes that bottom
strip. Between the menu closing and the confirmation opening, the file list is
repainted (`requestUpdateAndWait`) so the menu box doesn't bleed under the
confirmation — they share a single framebuffer in single-buffer mode.

## Internationalisation

Adds one new key, `STR_RENAME`, translated in **27 of the 31 languages**. The four
languages added to the repo after this work started (Arabic, Bosnian, Indonesian,
Norwegian) fall back to English until a native speaker fills them in. All other
user-facing strings reuse existing keys (`STR_DELETE`, `STR_CANCEL`,
`STR_CONFIRM`, `STR_BACK`, `STR_SELECT`, …).

## Backwards compatibility

- No public API removed. `ConfirmationActivity`'s new `overlay` parameter
  defaults to `false`; all three existing callers are untouched.
- Behavioural change: a long-press of Open now opens a menu instead of going
  straight to the delete confirmation. Deleting is one extra button press
  (Open-hold → Delete → Confirm), which is the intended safety/usability
  improvement.

## Testing

- [x] `clang-format`
- [x] `python -m platformio run -e default` on the rebased branch — succeeds
- [x] Exercised end-to-end in a host emulator (compiles the real firmware against
      Arduino/HAL shims, pixel-identical framebuffer) before the rebase:
      long-press timing, menu navigation, Rename (extension preserved, list
      refresh, cache cleared), Delete (confirm/cancel, immediate dialog
      dismissal), and the overlay hint layout
- [ ] Re-run on hardware since the rebase onto `develop`

## Build impact (`default` env, re-measured on this rebase)

| | `develop` | this branch | delta |
|---|---|---|---|
| Flash | 5,583,111 B | 5,587,095 B | **+3,984 B** |
| RAM | 50,668 B | 50,668 B | unchanged |

---

### AI Usage

This feature was designed and implemented by Claude through Claude Code, based on
the author's requirements and iterative testing feedback. The rebase onto
`develop` was likewise done with Claude Code, reviewed and built locally by the
contributor.


Uri-Tauber:
Hmm. I don't think renaming from the file explorer on the device would be a useful addition right now. The keyboard navigation is too clumsy for anyone to use when there's other option. Renaming via the web server is much more intuitive.

I think it's best to keep this a draft for now. This feature will make a lot more sense when we officially support touchscreen devices

jetaudio:
> Hmm. I don't think renaming from the file explorer on the device would be a useful addition right now. The keyboard navigation is too clumsy for anyone to use when there's other option. Renaming via the web server is much more intuitive.
> 
> 
> 
> I think it's best to keep this a draft for now. This feature will make a lot more sense when we officially support touchscreen devices

Fair enough. I will keep this for future touch devices.

Uri-Tauber:
Drafting this for now.

PR #2349, opened 2026-06-14
feat: add live font preview pane to font selection screen

PaulDelestrac:
## Summary
* Adds a live font preview pane to the font selection screen so users can see how a font looks before committing to it.
* Changes
  * A preview pane occupying the top 30% of the font selection screen, rendering sample pangram text in the previewed font
  * A two-step confirm flow: first press (enter button) previews the font, second press selects it
  * Back restores the original font settings, so browsing has no side effects
  * Layout dimensions cached in `onEnter()` to avoid redundant recalculation between `loop()` and `render()`

## Additional Context
* Preview sample text is hardcoded English; didn't want to use AI for translation as I would not be able to verify the output in most languages...
* The preview pane reduces visible list height; this is compensated by passing the reserved height into `getNumberOfItemsPerPage`

---
### AI Usage
AI use for assisting in coding and in writing the PR description.


Uri-Tauber:
Very nice!

[screenshot-42636.bmp](https://github.com/user-attachments/files/28975538/screenshot-42636.bmp)

One thing: `constexpr const char* PREVIEW_TEXT` is too long for the preview pane. Also, I can imagine non-english speakers would prefer a sentence in their own language. So, it probably best to move it under I18N rather than hardcode it.
For example, the Hebrew equivalent of "The quick brown fox jumps over the lazy dog." is "דג סקרן שט בים מאוכזב ולפתע מצא חברה"

PaulDelestrac:
> Very nice!
> 
> [screenshot-42636.bmp](https://github.com/user-attachments/files/28975538/screenshot-42636.bmp)
> 
> One thing: `constexpr const char* PREVIEW_TEXT` is too long for the preview pane. Also, I can imagine non-english speakers would prefer a sentence in their own language. So, it probably best to move it under I18N rather than hardcode it. For example, the Hebrew equivalent of "The quick brown fox jumps over the lazy dog." is "דג סקרן שט בים מאוכזב ולפתע מצא חברה"

Thanks @Uri-Tauber! I'll work on it!
I'll add your Hebrew sentence and I can also add a French equivalent. I'll leave the rest to the trusted team of [translators](https://github.com/crosspoint-reader/crosspoint-reader/blob/master/docs/translators.md).

iandchasse:
A whole bunch of examples here: https://en.wikipedia.org/wiki/Pangram for future reference. Any chance the size could be toggled for preview as well somehow?

src/activities/settings/FontSelectionActivity.cpp
```diff
@@ -99,39 +146,87 @@ void FontSelectionActivity::handleSelection() {
   finish();
 }
 
+int FontSelectionActivity::getFontIdForPreview(int index) const {
+  if (index < 0 || index >= static_cast<int>(fonts_.size())) return 0;
+  return SETTINGS.getReaderFontId();
+}
```
Uri-Tauber: index is validated but never used

src/activities/settings/FontSelectionActivity.cpp
```diff
@@ -99,39 +146,87 @@ void FontSelectionActivity::handleSelection() {
...
+
+void FontSelectionActivity::renderPreviewPane(int top, int height, int fontId, bool available,
```
Uri-Tauber: in the call site you send `previewFontId, previewFontId != 0`, which translate here to `int fontId, bool available`, but the only use of `available` is in line 170: `if (!available || fontId == 0) return;` which can be translated to `(fontId == 0 || fontId == 0)`.
Basically `available` is redundant.

src/activities/settings/FontSelectionActivity.cpp
```diff
@@ -1,26 +1,61 @@
...
 #include "CrossPointSettings.h"
 #include "MappedInputManager.h"
+#include "SdCardFontSystem.h"
 #include "components/UITheme.h"
 #include "fontIds.h"
 
+namespace {
+constexpr const char* ELLIPSIS_UTF8 = "\xe2\x80\xa6";
+constexpr int PREVIEW_PADDING = 12;
+constexpr int PREVIEW_HEIGHT_PERCENT = 30;
```
Uri-Tauber: Maybe move those two to `ThemeMetrics`?

Uri-Tauber (changes requested):
Overall, very good code. Few nitpicks to fix.

Uri-Tauber:
@iandchasse I think this should be a different PR.

BTW, good to see you back here.

PaulDelestrac:
@Uri-Tauber thanks for the review and the comments! 🙏 

@iandchasse Thanks for the suggestion, I added pangram for all languages in i18n! 😄

Uri-Tauber (approved):

itsthisjustin:
So does this only work on installed fonts? Not in the download screen?

Uri-Tauber:
@itsthisjustin Yes.

itsthisjustin (approved):

PR #2353, opened 2026-06-15
feat: add double click functionality to power button

jmt-gh:
## Summary
- **What is the goal of this PR?** Implements a new feature to allow double clicking of the power button to perform an action
  - Adds a new "Double Power Button Click" setting in Settings → Controls, independent of the existing single short-press action. Options: Ignore (default), Sleep, Refresh Screen, Orientation change.
  - Adds power-button double-press gesture detection in HalGPIO: raw releases are resolved into one-shot single/double-press events using a 300 ms window, exposed via wasPowerSinglePressed() / wasPowerDoublePressed() (forwarded through MappedInputManager).
  - Defers the single short-press action when a double-press action is configured: the existing single-press triggers (force refresh, power-as-page-turn, footnotes) now fire off the resolved single-press event, so the first tap waits out the double-press window instead of firing immediately. When double-press is set to Ignore, detection is disabled and single-press fires instantly on release — behaviour is unchanged.
  - Dispatches the double-press action: Sleep and Refresh globally in the main loop; Orientation change in the EPUB reader (reusing the existing applyOrientation re-layout path).

Split into 3 reviewable commits: (1) detection primitive, (2) the setting + i18n, (3) wiring/dispatch.

## Additional Context

- Memory: Cost is one uint8_t setting plus a few bool/unsigned long fields in the HalGPIO singleton. No heap allocation, no new tasks, no framebuffer impact. Build verified: RAM 30.9%, Flash 79.4%.
- Persistence: Uses the generic JSON settings path (JsonSettingsIO iterates the settings list by key and clamps enum values), so no manual save/load code was needed. The binary loader is migration-only and predates this field.
- Areas to focus review on:
  - The gesture state machine in HalGPIO::update() — particularly the deferral logic and the one-shot semantics of the flags.
  - Latency tradeoff: when a double-press action is enabled, single short-press actions are intentionally delayed by ~300 ms. This only affects users who opt into a double-press action.
  - The screenshot-combo guard: power releases while DOWN is held are ignored so the Power+Down screenshot gesture isn't misread as a tap. Edge cases around release ordering are possible but mirror existing screenshot-avoidance behaviour.
- Known interaction: if Short Power Button Click is set to Sleep, the device sleeps on the first tap (10 ms threshold), so a double-press can't form — double-press is intended to pair with a non-Sleep single-press setting.
- Scope note: orientation-change-on-double-press is EPUB-reader-only, matching the existing precedent (the long-press orientation feature and Txt/Xtc readers behave the same way).
- Testing: Compiles cleanly (pio run) and passes clang-format-21. Validated on X4 hardware as well.
---
### AI Usage


itsthisjustin:
@jmt-gh can you solve conflicts and coderabbit?

PR #2357, opened 2026-06-16
fix: enable Shift key for URL keyboard input (#2178)

### LINKED ISSUE #2178: Shift key, inside of the OPDS doesnt work
### Affected Version

1.3.0

### Describe the Bug

Shift key in anything that has the url shortcuts in the keyboard doesn't function

### Steps to Reproduce

1. go to settings
2. OPDS Servers/KOreader Sync 
3. choose to input a url

### Expected Behavior

The new keyboard shift key doesnt work when entering URLs

```shell

```

nnnkit:
## Summary

Fixes #2178.

When entering a URL (OPDS server address or KOReader Sync URL), the on-screen keyboard's **Shift key did nothing**, so it was impossible to type uppercase letters — exactly the problem the reporter hit when their OPDS endpoint contained capital letters.

## Root cause

In `KeyboardEntryActivity`, Shift was hard-disabled for `InputType::Url`:

- `handleKeyPress()` returned early for `inputType == InputType::Url`, so `shiftState` never toggled.
- The bottom Shift key was themed `Disabled` for URL input.
- The long-press "secondary char" fallback (`getAlternativeChar()`) is limited to the number row for URL input, so it could not produce uppercase letters on the letter rows either.

The net effect: **no way to enter capital letters in a URL field.**

## Fix

Enable Shift for the URL letter layout, while keeping it disabled where it has no meaning:

- the symbol layout (`symMode`), and
- the URL-snippet panel (`urlMode`, the `https://` / `.com` / `:8080` … shortcuts).

`getSelectedChar()` already selects from `urlLayout`, whose `secondary` entries are the uppercase letters, so once `shiftState` can toggle, uppercase input works with no further changes. The bottom key now renders enabled and shows its `shift` / `SHIFT` state, consistent with the regular text keyboard.

## Behavior after the change

- URL input, normal letter layout: **Shift toggles lower/upper case** (caps-lock style, same as text input).
- URL input, `#@!` symbol layout: Shift stays disabled (unchanged).
- URL input, `URL` snippet panel: Shift stays disabled (unchanged).
- Non-URL `Text` / `Password` input: unchanged.

## Verification

- 🔲 On-device check (flagged for a human tester): in Settings → OPDS Servers / KOReader Sync → enter a URL, press Shift, confirm letters are inserted uppercase and the Shift label toggles `shift` ↔ `SHIFT`; confirm the symbol and URL-snippet sub-modes still show Shift disabled.
- Change is confined to two boolean conditions in `KeyboardEntryActivity.cpp`; no new allocations, no layout/struct changes. clang-format clean (lines ≤120 cols). Full build / format are covered by CI.

### DIFF
--- a/src/activities/util/KeyboardEntryActivity.cpp
+++ b/src/activities/util/KeyboardEntryActivity.cpp
@@ -99,7 +99,9 @@ bool KeyboardEntryActivity::handleKeyPress() {
       case SpecialKeyType::Shift:
         delPressCount = 0;
         hintVisible = false;
-        if (urlMode || inputType == InputType::Url) return true;
+        // Shift is meaningless in the URL-snippet panel and the symbol layout, but
+        // must work for the URL letter layout so uppercase letters can be entered (#2178).
+        if (urlMode) return true;
         if (symMode) return true;
         shiftState = (shiftState + 1) % 2;
         return true;
@@ -675,8 +677,8 @@ void KeyboardEntryActivity::render(RenderLock&&) {
     const char* label;
   };
   const BottomKeyInfo bottomKeys[BOTTOM_KEY_COUNT] = {
-      {(symMode || urlMode || inputType == InputType::Url) ? KeyboardKeyType::Disabled : KeyboardKeyType::Shift,
-       (symMode || urlMode || inputType == InputType::Url) ? shiftString[0] : shiftString[shiftState]},
+      {(symMode || urlMode) ? KeyboardKeyType::Disabled : KeyboardKeyType::Shift,
+       (symMode || urlMode) ? shiftString[0] : shiftString[shiftState]},
       {KeyboardKeyType::Mode, urlMode ? "abc" : (symMode ? "abc" : "#@!")},
       {inputType == InputType::Url ? KeyboardKeyType::Mode : KeyboardKeyType::Space,
        inputType == InputType::Url ? "URL" : nullptr},


Uri-Tauber:
@nnnkit Hey. Thanks for the contribution. Have you tested this on your device?

nnnkit:
I was not able to test it on real device. My X4's screen is broken at the moment and haven't received another one yet , so I can't flash and try it directly. Will test and update when I get my new device.

Here's the verification I did do instead:

- The only behavioral change is that Shift is no longer force-disabled for `InputType::Url`. `getSelectedChar()` already returns the layout's uppercase secondary when shiftState > 0, and urlLayout's letter keys already define those uppercase secondaries {'q','Q'}, {'w','W'}. So letting the Shift toggle through is all that's needed for uppercase input     
no other paths change.
- The change is scoped to URL input only; Text and Password input are untouched. Shift stays disabled in the symbol layout (symMode) and the URL-snippet panel (urlMode), where it has no meaning.
- I exercised the keyboard's selection logic off-device against the real urlLayout table: with Shift on, typing now yields uppercase (e.g. GIT) and toggles back to lowercase, whereas before the fix it stayed lowercase.

nnnkit:
ok so i finally got to test this on actual hardware (X4).

shift works in the URL field now - you hit it and the keyboard flips to uppercase and types caps, hit it again its back to lowercase. the #@! and URL snippet screens still keep shift disabled which is what we want. so yeah this fixes #2178.

image

one thing worth mentioning - my screen is cracked rn so i couldnt actually look at the device. what i did was read the framebuffer over usb with the CMD:SCREENSHOT command thats already in the firmware and mirror it to my laptop in the browser (roughly 1 fps but good enough). it pulls from RAM so the broken panel doesnt matter, worked fine for checking the fix.

image

Uri-Tauber:
Thanks. I hope you get your new device soon.
I saw on reddit people explaining how to replace broken screen - probably will be much cheaper. Maybe you should try this.

Uri-Tauber (approved):

uxjulia (approved):

itsthisjustin:
Why do we need uppercase letters in the url field?

NewbyPengu:
@nnnkit Thank you so much for the fix! @itsthisjustin it's needed when entering an OPD code from apps like Kavita that generate a random password.

PR #1842, opened 2026-05-06
feat: home cover settings

pablohc:
## Summary

EPUBs without pre-processing can take tens of seconds to generate a cover (up to 60s). This PR adds **three configurable cover modes** (`Settings → Display → Home Screen Cover`) with a **5-option navigable popup**, **per-book overrides** via the Reader Menu, and **theme-aware selection highlights** so the UI never blocks regardless of whether books have been optimized.

## Display Settings — Home Screen Cover popup

Selecting "Home Screen Cover" in Settings opens a navigable popup with 5 options:

| Option | Description |
|---|---|
| **Enabled** *(default)* | Show original cover in new books |
| **Regenerate All** | Enable and generate original covers |
| **Timeout (10s)** | Show original cover with time limit |
| **Disabled** | Show alternative cover in new books |
| **Delete All** | Disable and show alternative covers |

- Up/Down navigation with wrap-around
- [BACK] closes popup without changes, [SELECT] applies selected option

### Regenerate All

Enables covers for all recent books and forces regeneration on next Home visit. The most recently opened book regenerates immediately; others regenerate on demand when opened.

image

### Delete All

Disables covers for all recent books and deletes all `thumb_*.bmp` files. Sleep screen covers (`cover.bmp`, `cover_crop.bmp`) and XTC covers are preserved.

image

## Reader Menu — per-book cover control (overrides global):

- *Enabled/Disabled* global mode → toggle "Enable cover" / "Disable cover" per book
- *Timeout* global mode → "Generate cover" force-renders with no deadline

image

## Classic theme — cover card improvements:

|Cover generated successfully|Cover disabled (alternative procedural cover)|
|-|-|
|image|image|

## Quick test checklist

- [ ] **Popup navigation** — Up/Down wraps around, [BACK] closes, [SELECT] applies
- [ ] **Enabled** — show original cover in new books
- [ ] **Regenerate All** — enable and generate original covers for all books
- [ ] **Timeout** — show original cover with time limit, fallback if exceeded
- [ ] **Disabled** — show alternative cover in new books
- [ ] **Delete All** — disable and show alternative covers for all books
- [ ] **Per-book toggle** — disable/enable cover from Reader Menu works
- [ ] **Theme consistency** — popup highlight matches theme style in Classic, Lyra, and RoundedRaff

<details>
<summary>Technical Implementation</summary>

[...]

PR #2358, opened 2026-06-16
feat: Selection Popup

Uri-Tauber:
## Summary

Replaces the existing "cycle through options" interaction pattern with a dedicated OptionPopup dialog that lets users select from a visible list of choices.

Binary (2-option) settings continue to cycle as before.
Settings with 3+ options now open the popup, pre-selected on the current value.

Credit for the idea goes to @pablohc, who introduced it in #1842. This PR extends that concept to the entire system.

### What's added:

- New `OptionPopup` component (`src/components/OptionPopup.h`) — a self-contained, reusable popup with input handling, rendering, and a selection callback.
- `drawOptionPopup` rendering method added to `BaseTheme`, with full theme metric support (padding, font, corner radius, row highlight style, separator).
Theme values configured for all three themes: Base, Lyra, and RoundedRaff. Should be minimal adjustments for @itsthisjustin costum themes PR.

### Screenshots:
lyra_popup
roundedraff_popup

---

### AI Usage


itsthisjustin:
This is also much better ux for touch devices.

itsthisjustin:
@Uri-Tauber is this hooked up to work for confirmation screens like clearing book cache? I want to merge this now so my touch branch can use it

Uri-Tauber:
@itsthisjustin No, but it's a good idea. 
I was planning to refactor this to use `contextMenu` from freeInk SDK, like you suggested. I hope I will get to this in the next couple of days.

itsthisjustin:
@Uri-Tauber I'm just going to merge this now, pull it into my branch, and i'll make the pop up work on the screens I'm missing if you're cool with that? It makes sense for the touch branch to get this where needed and i'd rather work with this than a parallel path

Uri-Tauber:
@itsthisjustin OK. sure.

PR #2360, opened 2026-06-17
feat: Hidden wifi ssid support

HgGamer:
## Summary

    Enables connecting to **hidden (non-broadcast) Wi-Fi networks**. Previously the Wi-Fi setup screen could only join networks returned by the scan, and empty-SSID (hidden) results are deliberately skipped
  — so there was no way to reach a hidden AP.

    * Added an **"Add hidden network..."** entry at the bottom of the Wi-Fi network list (also shown when the scan finds nothing or fails). Selecting it prompts for the **SSID** via the keyboard (32-char
  max), then the **password** (an empty password joins open hidden APs), then connects through the existing connection flow.
    * If credentials are already saved for the entered SSID, the saved password is reused and the network connects directly.
    * `src/activities/network/WifiSelectionActivity.{h,cpp}`: new synthetic `isHiddenPlaceholder` list entry, new `HIDDEN_SSID_ENTRY` state and `promptHiddenSsid()`, extracted `promptPasswordEntry()`
  (reused by both scanned and hidden networks), and a `realNetworkCount` so the synthetic entry doesn't inflate the "%zu networks found" header count.
    * `lib/I18n/translations/english.yaml`: new strings `STR_ADD_HIDDEN_NETWORK` and `STR_ENTER_WIFI_SSID` (other languages fall back to English; generated i18n tables regenerate at build time).

  ## Additional Context

  * **No new resource cost.** Hidden APs already work at the connection layer (`WiFi.begin(ssid, password)`); the only missing piece was SSID entry. The change reuses the existing `networks` vector (one
  extra element per list build), the existing `KeyboardEntryActivity`, and the existing connection path — no new heap allocations, buffers, or tasks. 
  * **Areas to focus on for review:** the two-step keyboard flow (SSID → password) is driven through `HIDDEN_SSID_ENTRY` in `loop()` rather than nesting `startActivityForResult` callbacks, mirroring the
  existing password-entry transition; and `render()` suppresses drawing during that state, same as `PASSWORD_ENTRY`.
  * **Verified:** `pio run` builds clean, `clang-format` applied, flashed to an Xteink X4 and tested it.

  ---

  ### AI Usage


HgGamer:
@itsthisjustin  why did you close my pull request?

itsthisjustin:
did not mean to! Mistake when merging branches! Going to fix manually now. So sorry.

itsthisjustin:
@HgGamer can you just use AI to add the other language strings for this? Then we can get this merged in

HgGamer:
@itsthisjustin  resolved the merge conflict, and added the 2 new keys to all languages with AI translation ( wasn't sure if AI translation is better than no translation)

itsthisjustin:
> @itsthisjustin  resolved the merge conflict, and added the 2 new keys to all languages with AI translation ( wasn't sure if AI translation is better than no translation)

Imo it's better to get it in now so that it's much easier for less technical folks to update them later

itsthisjustin:
Can you make sure this works fine with the newly merged auto connect to last x wifi networks feature? @HgGamer ? 

I just put that into develop last night and I want to make sure hidden networks are included in that list

HgGamer:
i can only test hidden ssid on monday

HgGamer:
@itsthisjustin  tested it, works fine

itsthisjustin (approved):

PR #2372, opened 2026-06-18
fix: several bookmarks UX improvments

Uri-Tauber:
## Summary

This PR enhances the EPUB reader's bookmark system with two complementary improvements: a per-page bookmark indicator icon and toggle behavior on the existing long-press action.

---

### What Changed

**Bookmark Toggle (was: add-only)**

The long-press Confirm action now toggles bookmarks rather than always adding. `addBookmark()` checks whether a bookmark with the same xpath already exists in the in-memory cache:
- If found → removes it and shows "Bookmark removed."
- If not found → adds it and shows "Bookmark added."

A new `STR_BOOKMARK_REMOVED` translation string was added to support the removal message.

**Bookmark Icon Indicator**

A `BookmarkIcon` is now drawn at the top-right corner of the page whenever the current page has a bookmark. `updateBookmarkFlag()` is called at render time to determine whether the current page is bookmarked.

**In-Memory Bookmark Cache**

Bookmarks are now loaded into `cachedBookmarks` on `onEnter()` rather than being re-read from disk on every toggle. All subsequent add/remove operations work against this cache and flush to disk, avoiding redundant file reads on each bookmark action.

**Faster bookmarks list**

Previously, calculating "page X/Y" for each entry required decompressing the entire spine item. We now persist `si`/`pc`/`pp` (spine index, page count, and page progress) in the bookmark JSON when saving, and restore them when loading. This avoids the expensive `toCrossPoint()` loop in `onEnter()`, significantly reducing the cost of initializing the bookmarks list.

---

### Files Changed

- `EpubReaderActivity.cpp` — `addBookmark()` toggle logic, `updateBookmarkFlag()` (new), icon rendering in `renderContents()`, cache initialization in `onEnter()`
- `EpubReaderActivity.h` — new fields: `currentPageBookmarked`, `bookmarkRemoved`, `cachedBookmarks`; new method declaration `updateBookmarkFlag()`

---

### AI Usage


Uri-Tauber:
@Eloren1 @vedi0boy Want to take a look?

vedi0boy:
Just a couple of concerns from me:

It seems like the flag update for whether the page is bookmarked is done on every render. Might be a good idea to only do this on page turn, and toggle it efficiently when a bookmark is added or removed.

Also, I'm not sure it's a good idea to load in all the bookmarks and keep them all I'm memory at once. Seems to me like we should be loading them on indexing and then cache the page numbers that have bookmarks for the indexed portion. This would precomputed the calculation above to determine if a page was bookmarked and also use less memory.

Ot course, I could be overly cautious here and if there is no cause for concern, just let me know, but I want to avoid potential issues if there are pages with large chapters and we hit the memory limit.

Eloren1:
Yes, it does look nice!

image

Eloren1:
Although for now it ignores orientation

image
image

Eloren1:
@coderabbitai is there a function to detect orientation?

Uri-Tauber:
> It seems like the flag update for whether the page is bookmarked is done on every render. Might be a good idea to only do this on page turn, and toggle it efficiently when a bookmark is added or removed.

`render()` only fires on `requestUpdate()`, which happens on page turns, bookmark toggles, and menu exits. And with the in-memory vector, `updateBookmarkFlag()` is now just handful of float subtractions. Zero I/O, no ZIP decompression. Moving it to only fire on page turn would save nothing measurable.

> Also, I'm not sure it's a good idea to load in all the bookmarks and keep them all I'm memory at once. Seems to me like we should be loading them on indexing and then cache the page numbers that have bookmarks for the indexed portion. This would precomputed the calculation above to determine if a page was bookmarked and also use less memory.

My calculations ( correct me if I'm wrong here):
- Each BookmarkEntry is roughly: XPath (~50 bytes) + summary (~72 bytes) + float + 3×uint16 = ~130 bytes.
- 16 bookmarks (our reserved capacity) consume only ~2 KB.
- Even 50 bookmarks would be around ~6.5 KB.

One advantage of storing bookmark percentages rather than page numbers is that page numbers can become stale when the user changes orientation, font size, or other layout settings.

That said, I agree that if someone goes overboard and creates hundreds of bookmarks, it could theoretically lead to OOM crash.

Uri-Tauber:
removed "page X/Y" label in bookmarks list.

Each bookmark entry required ZIP decompression (~500ms each) in `EpubReaderBookmarksActivity::onEnter()` — all for the same spine item, just to compute "page X/Y" labels for the list display.

Different approach would be to cache `computedSpineIndex`/`computedChapterProgress`/`computedChapterPageCount` in the JSON at bookmark-creation time.
Tell me what do you prefer.

Uri-Tauber:
After playing around a bit I decided to go with caching in bookmark JSON. We avoid the heavy decompression while keeping "page X/Y" for each entry.

Uri-Tauber:
If anyone else want to make the bookmark icon orientation-aware - go ahead. I will probably have time for this in two days.

vedi0boy:
@Uri-Tauber sweet! Thanks for clearing my concern 

Yea, what I meant by page number is that during indexing, the page number can be precalculated because from what I understand, the indexing is done for the users orientation and font size, so it would make sense to calculate and cache the bookmarked page numbers then.

Anyway, no need to optimize this if it's not an issue. Thanks for working on this! LGTM

Uri-Tauber:
@itsthisjustin @uxjulia @znelson When you have time to review...

itsthisjustin:
Somehow bookmarks was also on by default for me on new builds. Is that turned off now with these fixes?

Uri-Tauber:
@itsthisjustin Nope. In my eyes, it better UX this way. 
A lot of users asked for bookmarks.

Open to other opinions.

itsthisjustin:
I find that I accidentally create bookmarks constantly now when trying to lock the device. I was more advocating for short power button to default to nothing so the setting is in user control

Uri-Tauber:
Short power button shouldn't create new bookmark. I think you mean long select button.

itsthisjustin:
> Short power button shouldn't create new bookmark. I think you mean long select button.

Ah valid point. On the touch devices I've been testing on we have limited buttons so confirm and power are the same. That explains the ux for me. We can get this merged now

itsthisjustin (dismissed):

uxjulia:
Found a weird bug:
- Set reader margins to 5
- Add a bookmark
- Change reader margins to 30
- Bookmark icon disappears, but the bookmarked page still remains in the bookmarks list.

Poor UX:
- As of this PR, we allow users to set the `Long-press Action Menu` to `Disabled`, `KOSync`, and `Bookmark` (This PR changes this default to `Bookmark`, which IMO should be opt-in since there's more than 1 option for the long-press shortcut now so they should choose). If the user sets this to anything other than `Bookmark` they can no longer add bookmarks from inside a book. An in-reader option to `Toggle Bookmark` should be added.

Minor nitpicks, non-blocking, but something to keep on our radar for the enhancements that come later:

- Bookmark was only added to long-press menu action, I think it should also be added as an option for short power button click. This can be a PR to clean up the shortcuts so both long-press menu and short-press power have the same options.
- The bookmark icon overlaps text at pretty much any margin setting. This just looks bad.

itsthisjustin:
> Found a weird bug:
> - Set reader margins to 5
> - Add a bookmark
> - Change reader margins to 30
> - Bookmark icon disappears, but the bookmarked page still remains in the bookmarks list.
> 
> Poor UX:
> - As of this PR, we allow users to set the `Long-press Action Menu` to `Disabled`, `KOSync`, and `Bookmark` (This PR changes this default to `Bookmark`, which IMO should be opt-in since there's more than 1 option for the long-press shortcut now so they should choose). If the user sets this to anything other than `Bookmark` they can no longer add bookmarks from inside a book. An in-reader option to `Toggle Bookmark` should be added.
> 
> Minor nitpicks, non-blocking, but something to keep on our radar for the enhancements that come later:
> 
> - Bookmark was only added to long-press menu action, I think it should also be added as an option for short power button click. This can be a PR to clean up the shortcuts so both long-press menu and short-press power have the same options.
> - The bookmark icon overlaps text at pretty much any margin setting. This just looks bad.
> 

Agree with all of these. I was going to work on the icon overlap myself but tbh I never thought of a good alternative. The bookmarks feature being on by default I feel is a must to change

Uri-Tauber:
@itsthisjustin Can you work on those and push here?
I have a lot of homework to finish today.

itsthisjustin:
> @itsthisjustin Can you work on those and push here?
> I have a lot of homework to finish today.

Will see what I can pull off if @uxjulia is also unavailable. Whole house is sick today on my end

uxjulia:
> > Found a weird bug:
> > 
> > * Set reader margins to 5
> > * Add a bookmark
> > * Change reader margins to 30
> > * Bookmark icon disappears, but the bookmarked page still remains in the bookmarks list.
> > 
> > Poor UX:
> > 
> > * As of this PR, we allow users to set the `Long-press Action Menu` to `Disabled`, `KOSync`, and `Bookmark` (This PR changes this default to `Bookmark`, which IMO should be opt-in since there's more than 1 option for the long-press shortcut now so they should choose). If the user sets this to anything other than `Bookmark` they can no longer add bookmarks from inside a book. An in-reader option to `Toggle Bookmark` should be added.
> > 
> > Minor nitpicks, non-blocking, but something to keep on our radar for the enhancements that come later:
> > 
> > * Bookmark was only added to long-press menu action, I think it should also be added as an option for short power button click. This can be a PR to clean up the shortcuts so both long-press menu and short-press power have the same options.
> > * The bookmark icon overlaps text at pretty much any margin setting. This just looks bad.
> 
> Agree with all of these. I was going to work on the icon overlap myself but tbh I never thought of a good alternative. The bookmarks feature being on by default I feel is a must to change

I do like the Kindle-style bookmark in the upper corner, but Kindles have a pretty large minimum margin so they can get away with it. The status bar would be the path of least resistance that wouldn't require re-drawing the layout to make the icon fit.

itsthisjustin:
> I do like the Kindle-style bookmark in the upper corner, but Kindles have a pretty large minimum margin so they can get away with it. The status bar would be the path of least resistance that wouldn't require re-drawing the layout to make the icon fit.

Oh yeah great idea

Eloren1:
What if I don't use status bar indicators?

itsthisjustin:
> What if I don't use status bar indicators?

Even better. Gives you control over if you can see it or not

uxjulia:
My changes:  
* Adds a bookmark icon in the bottom status bar icon and renders it in the left-most position when the current page is bookmarked.
* Adds a reader menu `Toggle Bookmark` action that adds or removes the current page bookmark depending on current state.
* Only shows the `Bookmarks` menu item when there are bookmarks to display.
* No longer sets `Bookmark` as a long-press menu action by default
* Changed the "Hold Confirm to Delete" help-text in the bookmarks list to "Hold Open to Delete", since that was the visible button label.
* Added the rest of the i18n translations.

New bookmark icon in status bar:
Alice&#39;s-Adventures-in-Wonderland-and-Through-the-Looking-Glass_ch13_p5_19pct_19034 Medium

lib/I18n/translations/hebrew.yaml
```diff
@@ -359,8 +360,7 @@ STR_FRONT_BTN_FOLLOW_ORIENTATION: "התאמת לחצנים קדמיים לכיו
 STR_REMOVE_READ_FROM_RECENTS: "הסר ספרים שנקראו מרשימת האחרונים"
 STR_MOVE_FINISHED_TO_READ: "העבר ספרים שהסתיימו לתיקיית 'נקראו'"
 STR_DISABLED: "מבוטל"
-STR_HOLD_CONFIRM_TO_DELETE: "החזק לחוץ על אישור כדי למחוק"
-STR_BOOKMARK_INSTRUCTIONS: "החזק לחוץ על כפתור אישור בזמן הקריאה כדי להוסיף סימנייה"
+STR_HOLD_OPEN_TO_DELETE: "החזק לחוץ על פתח כדי למחוק"
```
Uri-Tauber:
```suggestion
STR_HOLD_OPEN_TO_DELETE: "לחיצה ארוכה על 'פתח' כדי למחוק"
```

lib/I18n/translations/hebrew.yaml
```diff
@@ -188,6 +188,7 @@ STR_HOME: "מסך הבית »"
 STR_SELECT: "בחר"
 STR_SELECTED: "נבחר"
 STR_TOGGLE: "בחר"
+STR_TOGGLE_BOOKMARK: "החלף סימנייה"
```
Uri-Tauber:
```suggestion
STR_TOGGLE_BOOKMARK: "הוסף/הסר סימנייה"
```

uxjulia (approved):

itsthisjustin:
> My changes:  
> * Adds a bookmark icon in the bottom status bar icon and renders it in the left-most position when the current page is bookmarked.
> * Adds a reader menu `Toggle Bookmark` action that adds or removes the current page bookmark depending on current state.
> * Only shows the `Bookmarks` menu item when there are bookmarks to display.
> * No longer sets `Bookmark` as a long-press menu action by default
> * Changed the "Hold Confirm to Delete" help-text in the bookmarks list to "Hold Open to Delete", since that was the visible button label.
> * Added the rest of the i18n translations.
> 
> New bookmark icon in status bar:
> Alice&#39;s-Adventures-in-Wonderland-and-Through-the-Looking-Glass_ch13_p5_19pct_19034 Medium
> 
> 

Amazing

vedi0boy:
Just my opinion, I'm conflicted on hiding the bookmark menu when there are no bookmarks.

When there are no bookmarks, the menu will display a message explaining to the user how to add bookmarks. I feel that it is more user friendly this way as it helps the user see what features are available to them and how to use them. Many users may not even know that the feature is available to them unless they read the manual (unlikely) or they accidentally add a bookmark (also unlikely). Just my thoughts.

uxjulia:
> Just my opinion, I'm conflicted on hiding the bookmark menu when there are no bookmarks.
> 
> When there are no bookmarks, the menu will display a message explaining to the user how to add bookmarks. I feel that it is more user friendly this way as it helps the user see what features are available to them and how to use them. Many users may not even know that the feature is available to them unless they read the manual (unlikely) or they accidentally add a bookmark (also unlikely). Just my thoughts.

There is a new "Toggle Bookmark" action in the reader menu. An empty "Bookmarks" would've just told them to use the toggle bookmarks menu item, so that would've been redundant. 

A persistent "Toggle Bookmarks" menu item shown with no bookmarks:
toggle bookmark

"Bookmarks" shows up after the user adds at least one bookmark in the book:
bookmarks

vedi0boy:
> > Just my opinion, I'm conflicted on hiding the bookmark menu when there are no bookmarks.
> > 
> > When there are no bookmarks, the menu will display a message explaining to the user how to add bookmarks. I feel that it is more user friendly this way as it helps the user see what features are available to them and how to use them. Many users may not even know that the feature is available to them unless they read the manual (unlikely) or they accidentally add a bookmark (also unlikely). Just my thoughts.
> 
> There is a new "Toggle Bookmark" action in the reader menu. An empty "Bookmarks" would've just told them to use the toggle bookmarks menu item, so that would've been redundant. 
> 
> A persistent "Toggle Bookmarks" menu item shown with no bookmarks:
> toggle bookmark
> 
> "Bookmarks" shows up after the user adds at least one bookmark in the book:
> bookmarks
> 

Ohh cool. My bad, I'll refrain from leaving comments without actually testing this 😅

itsthisjustin (approved):

JINL020:
Hellooo, and thank you @Uri-Tauber for implementing this amazing feature!

I have a small UI suggestion regarding the bookmark icon in the status bar. Currently, when a page is bookmarked, the bookmark icon appears before the battery indicator, causing the battery indicator to shift to the right. Would it make sense to place the bookmark icon after the battery indicator (or before the page count) instead? That would keep the battery indicator in a fixed position and make the status bar feel a bit more stable.😗

Uri-Tauber:
@JINL020 The credit goes to `@vedi0boy`, I just made a few quirks here.

I will try to look into your suggestion tomorrow.

itsthisjustin:
> Hellooo, and thank you @Uri-Tauber for implementing this amazing feature!
> 
> I have a small UI suggestion regarding the bookmark icon in the status bar. Currently, when a page is bookmarked, the bookmark icon appears before the battery indicator, causing the battery indicator to shift to the right. Would it make sense to place the bookmark icon after the battery indicator (or before the page count) instead? That would keep the battery indicator in a fixed position and make the status bar feel a bit more stable.😗 

Totally agree

JINL020:
> @JINL020 The credit goes to `@vedi0boy`, I just made a few quirks here.
> 
> I will try to look into your suggestion tomorrow.

Upps, shout out to @vedi0boy and everyone who worked on it 👏🏼

vedi0boy:
> > @JINL020 The credit goes to `@vedi0boy`, I just made a few quirks here.
> > 
> > I will try to look into your suggestion tomorrow.
> 
> Upps, shout out to @vedi0boy and everyone who worked on it 👏🏼 

Not just me. I wasn't the first to make a bookmark feature so I ended up collabing with the creator of the first bookmark MR to make it better.

PR #2379, opened 2026-06-19
fix: automatically connect to wifi for clock sync

uxjulia:
## Summary

  * Make manual clock sync work even when the device is not already connected to Wi-Fi, so the user can start the sync flow directly from settings instead of being blocked by connection state.
  * `ClockSyncActivity` now launches the normal Wi-Fi selection flow before syncing when the device is offline.
  * After Wi-Fi selection succeeds, clock sync resumes automatically and performs the existing forced NTP sync.
  * If clock sync had to bring Wi-Fi up, the activity disconnects and uses the existing silent restart cleanup path to avoid leaving the device in a fragmented post-Wi-Fi heap state.
  * The completion UI now only advertises Back, matching the updated input handling.

## Additional Context

* Existing behavior is unchanged when Wi-Fi is already connected: the activity syncs immediately.
* Cancelling Wi-Fi selection exits the clock sync flow.

---

### AI Usage

### DIFF
--- a/src/activities/settings/ClockSyncActivity.cpp
+++ b/src/activities/settings/ClockSyncActivity.cpp
@@ -10,6 +10,8 @@
 
 #include "CrossPointSettings.h"
 #include "MappedInputManager.h"
+#include "SilentRestart.h"
+#include "activities/network/WifiSelectionActivity.h"
 #include "components/UITheme.h"
 #include "fontIds.h"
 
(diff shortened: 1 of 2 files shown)


itsthisjustin:
Thanks for doing this!

itsthisjustin (approved):

Uri-Tauber (approved):

PR #2214, opened 2026-05-30
feat: Smooth progressive JPEG cover upscales in BMP conversion

lpla:
## Summary

* **What is the goal of this PR?** Improve the visual quality of progressive JPEG covers when they are upscaled for BMP generation on supported display sizes (tested on X4), especially for sleep covers and home thumbnails where JPEGDEC decodes progressive images on a reduced 1/8 grid.
* **What changes are included?**
  * Add a progressive-only smoothing path in `JpegToBmpConverter` before dithering.
  * Use a small line-buffered bilinear pass for progressive JPEG upscales.
  * Keep the existing scaling path for non-progressive JPEGs and other cases.
  * Store the smoothing rows in one contiguous allocation sized at `3 * outWidth` bytes.

## Additional Context

* JPEGDEC decodes progressive JPEGs on a reduced 1/8 grid in this path, so the old upscale behavior could make covers look visibly blocky.
* This PR only changes progressive upscale cases. Non-progressive JPEG covers continue on the existing path.
* The new path keeps memory bounded and local to the converter. For progressive upscales, it uses a single temporary smoothing buffer of `3 * outWidth` bytes.
* Compatibility note: this is not X4-hardcoded. The cover target comes from runtime display dimensions (`display.getDisplayHeight()` / `display.getDisplayWidth()`); thumbnail dimensions are supplied by the caller. The 480 px / 150 px figures below are X4 test cases, not fixed limits.
* In the tested cases, that was:
  * `1440` bytes for a `480 px` sleep cover
  * `450` bytes for a `150 px` thumbnail
* Relative to the old scaler-local working set, this is smaller for progressive upscales:
  * `480 px` output: `3840` bytes before, `1440` bytes now
  * `150 px` output: `1200` bytes before, `450` bytes now
* The master comparison did not show a measurable whole-system heap regression during the tested flow. The observed low-water marks were effectively unchanged, and the lower memory warnings seen during monitoring appeared earlier in broader EPUB/cache work rather than inside this smoothing path.
* Measured impact from the tested branch:
  * Static RAM unchanged from master
  * Flash about `1 KB` higher than master
  * No meaningful conversion-time penalty in the tested cover flow
* Validation performed:
  * `git diff --check`
  * `pio run`
  * `pio check`
  * `pio run -t upload`
  * On-device checks with a progressive JPEG cover in home thumbnail and sleep cover views
  * On-device check with a normal non-progressive JPEG cover to confirm the old path remains unchanged

---

### AI Usage

[...]

PR #2380, opened 2026-06-19
feat: smooth progressive JPEG cover BMP upscales

gershalfred:
## Summary
- smooth progressive JPEG EPUB covers in the BMP generation path
- keep the existing progressive decode handling, but avoid blocky 1/8-grid upscales
- leave non-progressive and non-upscale cases on the existing path

## Why
Progressive JPEG covers can look noticeably pixelated on Xteink/CrossPoint-derived firmware when cover BMPs are generated for library/home/sleep surfaces. The decode path already recognizes progressive JPEGs, but the BMP converter was still enlarging that reduced decode grid without a smoothing pass.

## What changed
- adds a progressive-only bilinear smoothing path in `lib/JpegToBmpConverter/JpegToBmpConverter.cpp`
- keeps memory bounded to small line buffers
- gates smoothing so it only activates when both decoded axes are being upscaled

## Verification
- independent review passed for correctness / memory-safety / regression risk
- targeted file compile passed for the modified converter in the sibling fork implementation
- this repo patch is based on the existing upstream PR direction and applied cleanly here

## Notes
I was not able to produce a full local PlatformIO firmware build in this environment because repository/tooling setup blocked before patch-level verification (`open-x4-sdk/libs/hardware/BatteryMonitor` symlink/package issue during `python3 -m platformio run -e default`).

### DIFF
--- a/lib/JpegToBmpConverter/JpegToBmpConverter.cpp
+++ b/lib/JpegToBmpConverter/JpegToBmpConverter.cpp
@@ -164,6 +164,7 @@ namespace {
 constexpr int MAX_MCU_HEIGHT = 16;
 constexpr size_t JPEG_DECODER_SIZE = 20 * 1024;
 constexpr size_t MIN_FREE_HEAP = JPEG_DECODER_SIZE + 32 * 1024;
+constexpr uint32_t FP_ONE = 1UL << 16;
 
 // Static file pointer for JPEGDEC open callback.
 // Safe in single-threaded embedded context; never accessed concurrently.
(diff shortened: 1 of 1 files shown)


uxjulia:
Tested on device. Works well. @gershalfred also shared the same code change for Crossink which I'll be adding to the next release.

uxjulia (approved):

lpla:
Duplicated? https://github.com/crosspoint-reader/crosspoint-reader/pull/2214

uxjulia:
> Duplicated? https://github.com/crosspoint-reader/crosspoint-reader/pull/2214

Didn't notice this! Looks like this one is identical to yours byte for byte. Will change my approval to yours

gershalfred:
Thanks for flagging. I hadn’t seen #2214 when I opened this. Looks like we arrived at essentially the same patch independently. Happy to have this closed in favor of the earlier PR if that’s easier for review/merge flow.

lpla:
@gershalfred That would be great, as my PR has been tested with several books for weeks on my daily device.

Issue #2336, opened 2026-06-13
Broken links on footnotes


javiims:
### Affected Version

1.4.0 RC2 (1.3.0-dev-master-6f05e401a)

### Describe the Bug

When opening a footnote link, it forwards you to the beginning of the section fo the footnotes, not to the footnote itself

### Steps to Reproduce

Go to the page with one footnote (*)
Open footnote

### Expected Behavior

It should forwards to the footnote itself, not to the beginnning of the section

### Relevant Log Output/Screenshots

```shell

```

Uri-Tauber:
I just tested the footnote behavior, and it's working fine. There may be an issue with your specific book. Did the links in this book work correctly in version 1.3?

javiims:
They did not work on version 1.3, but they do work on any other device or ebook reader

javiims:
[Jugarse la piel.txt](https://github.com/user-attachments/files/28949593/Jugarse.la.piel.txt)
I attach an example. Change file extension to epub, and go to page 327

adequacy1:
Also affected 1.3.0-dev+55a5691, same behaviour with jumping to the beginnning of the section.

adequacy1:
I discovered that in version 1.2.0, footnotes worked correctly in those same ePub books. Very strange. Bug is somewhere near 2ea042a68a54c1dfe13120feeabd67cfe862e711 commit I think.

Uri-Tauber:
I believe the bug with @javiims's book was the footnotes links were too long, and silently got tracuated. 
@javiims Can you try to flash this to check if it solve the problem?

[firmware.bin.zip](https://github.com/user-attachments/files/29067504/firmware.bin.zip)

javiims:
> I believe the bug with [@javiims](https://github.com/javiims)'s book was the footnotes links were too long, and silently got tracuated. [@javiims](https://github.com/javiims) Can you try to flash this to check if it solve the problem?
> 
> [firmware.bin.zip](https://github.com/user-attachments/files/29067504/firmware.bin.zip)

I flashed and tried, but still not working

Uri-Tauber:
Hmmm. I will dig further.

adequacy1:
> Hmmm. I will dig further.

Two epub files affected, zipped:

[epub1.zip](https://github.com/user-attachments/files/29077075/epub1.zip)
[epub2.zip](https://github.com/user-attachments/files/29077076/epub2.zip)

[...]

PR #2382, opened 2026-06-20
fix(epub): flush displaced anchor before overwrite (#2336)

kygia:
## Summary

Fixes footnote links landing in start section instead of the specific foot note target.

ChapterHtmlSlimParser.cpp: call flushPendingAnchor() before overwriting pendingAnchorId. First id got lost due to consecutive non-block elements carry ids, the first id was lost and the reader had no page to jump to, so it defaulted to page 0.

## Additional Context

Tested with the epubs attached to to #2336. Need to clear .crosspoint/ cache after flashing so the anchor map gets rebuilt with the fix.

  specific areas to focus on).

---

### AI Usage

Partially assisted with writing documentation.


Uri-Tauber (approved):
Thanks for this.

javiims:
Thansk so much for the fix!

uxjulia (approved):

PR #2385, opened 2026-06-20
fix: prevent out-of-memory crash when opening directories with many files

10ne1:
## Summary

I noticed that x3 crashes when opening directories with many files due to exhausting the heap memory, so I created this patch to avoid crashing by truncating the list of files. This is just a reasonable immediate fix for the crash, not a complete / permanent solution to the problem.

## Additional Context

Please see the commit message for the full context and also potential follow-up work to either increase the truncation limit or eliminate it entirely, for convenience I'm pasting it here:

```
fix: prevent crash when opening dirs with many files

Opening a directory with thousands of files crashes the device with an
abort() panic (PC resolves to __cxxabiv1::__terminate) because large
dirs exhaust the heap in FileBrowserActivity::loadFiles() which builds
an unreserved std::vector<std::string> until the "next" alloc fails.

The decoded stack trace for the OOM is:
  operator new(unsigned int) -> std::bad_alloc()
    -> __cxa_allocate_exception() -> __terminate() -> abort()
  make_unique<HalFile::Impl>(FsFile&&)   HalStorage.cpp:64
  HalFile::openNextFile()                HalStorage.cpp:169
  FileBrowserActivity::loadFiles()       FileBrowserActivity.cpp:39
  FileBrowserActivity::loop()            FileBrowserActivity.cpp:270
  ActivityManager::loop()                ActivityManager.cpp:66
  loop()                                 main.cpp:583

So the HalStorage open paths allocate HalFile::Impl via std::make_unique
which is a bare new. With -fno-exceptions, failed new raises a bad_alloc
which routes to __terminate -> abort() instead of returning null.

The above is a violation of the resource protocol rule 9 (see CLAUDE.md)
and it also violates rule 7 because it doesn't reserve the vector size
and causes reallocation churn.

The fix has 3 steps:

  1. HalStorage: allocate with makeUniqueNoThrow in the open paths and
     degrade gracefully (empty HalFile/false) on OOM instead of crashing.

  2. FileBrowserActivity: reserve the list to cut reallocation churn.

  3. FileBrowserActivity: stop adding entries once the free heap drops
     below a safety floor, logging the truncation. This bounds RAM use
     so the throwing vector can no longer abort.

The heap limit at point 3 is set to a conservative estimate of 24 kb,
based on my observed system usage with a minimum free remaining heap
dip to ~16 kb with an ~17kb largest continuous available block. It's
enough to hold / truncate to 1300 items in memory without crashing.

Obviously further work can be done on this: the std string vec can be
optimized to be more space efficient, or, even better, an implementation
which avoids keeping all filenames in heap memory can replace the naive
string vec, since we never show all files at the same time.

Verified on X3: large directories now open with the file list truncated
and a logged warning instead of crashing and rebooting.

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>
Signed-off-by: Adrian Ratiu <adrian.ratiu@collabora.com>
```

## AI Usage

This PR was written with the help of Claude Code, but was **not** vibe-coded. I'm not a fan of vibe-coding. I used AI to better understand the codebase myself (asked it a lot of questions), review both my code and the code generated by AI, used it especially to detect bugs and corner-cases and suggest fixes, however I very carefully reviewed & edited each code/comment lines, the commit messages and so on, until I was satisfied with the result.

### DIFF
--- a/lib/hal/HalStorage.cpp
+++ b/lib/hal/HalStorage.cpp
@@ -2,6 +2,7 @@
 
 #include <FS.h>  // need to be included before SdFat.h for compatibility with FS.h's File class
 #include <Logging.h>
+#include <Memory.h>
 #include <SDCardManager.h>
 
 #include <cassert>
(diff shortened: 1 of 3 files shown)


Uri-Tauber:
Hey @10ne1 Sorry for missing this one.
The code looks clean, and I trust you it's working. 
Two things before merge:
a. Could you update USER_GUIDE to reflect this, so users will understand why their 2,000-files folder is only half shown.
b. Maybe it would be better to have popup message, in addition to the debugging info. Not sure if it worth the ~30 new strings...

BTW, how much are we talking about? 200 files? 500?

PR #2393, opened 2026-06-22
feat: render grayscale epub images without text aa

### LINKED ISSUE #2348: Image display in black and white or grayscale
When the font smoothing function is enabled, grayscale images in the text are displayed correctly. When the function is disabled, the image appears almost unrecognizable, rendered only in black and white. Is it possible to apply this function selectively to images, so that they are displayed correctly in grayscale even when font smoothing is turned off?

uxjulia:
## Summary

  * Keep EPUB images rendering in grayscale even when text anti-aliasing is disabled.
  * Split grayscale rendering decisions between text and images, so image pages still run grayscale passes while plain text pages with AA off stay on the BW-only path.
  * Add `Page::renderImages()` to rerender only image elements during grayscale passes when text AA is off.
  * Preserve the existing full-page render path when text AA is enabled.
  * Fixes #2348 

## Additional Context

* This was ported from CrossInk's existing implementation of this feature.
* This avoids treating “text AA off” as “no grayscale needed”, image pages still need grayscale to preserve image quality.
* When text AA is off, the grayscale pass rerenders only images instead of the whole page, so text remains BW.
* I verified on device with an image EPUB with Text AA on/off and confirmed working.
---

### AI Usage

### DIFF
--- a/src/activities/reader/EpubReaderActivity.cpp
+++ b/src/activities/reader/EpubReaderActivity.cpp
@@ -968,20 +968,27 @@ void EpubReaderActivity::renderContents(std::unique_ptr<Page> page, const int or
                                         const int orientedMarginRight, const int orientedMarginBottom,
                                         const int orientedMarginLeft) {
   const auto t0 = millis();
+  const int fontId = SETTINGS.getReaderFontId();
 
   // Font prewarm: scan pass accumulates text, then prewarm, then real render
   auto* fcm = renderer.getFontCacheManager();
   auto scope = fcm->createPrewarmScope();
-  page->render(renderer, SETTINGS.getReaderFontId(), orientedMarginLeft, orientedMarginTop);  // scan pass
+  page->render(renderer, fontId, orientedMarginLeft, orientedMarginTop);  // scan pass
   scope.endScanAndPrewarm();
   const auto tPrewarm = millis();
 
   const bool pageHasImages = page->hasImages();
-  const bool needsImageGrayscale = pageHasImages;
   const bool needsTextGrayscale = SETTINGS.textAntiAliasing;
-  const bool needsAnyGrayscale = needsTextGrayscale || needsImageGrayscale;
+  const bool needsAnyGrayscale = needsTextGrayscale || pageHasImages;
+  auto renderGrayscalePass = [&]() {
+    if (needsTextGrayscale) {
+      page->render(renderer, fontId, orientedMarginLeft, orientedMarginTop);
+    } else {
+      page->renderImages(renderer, fontId, orientedMarginLeft, orientedMarginTop);
+    }
+  };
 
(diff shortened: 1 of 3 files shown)


src/activities/reader/EpubReaderActivity.cpp
```diff
@@ -1036,7 +1038,11 @@ void EpubReaderActivity::renderContents(std::unique_ptr<Page> page, const int or
         const int rows = (gh - y < STRIP_ROWS) ? (gh - y) : STRIP_ROWS;
         renderer.beginStripTarget(scratch.get(), y, rows);
         renderer.clearScreen(0x00);
-        page->render(renderer, SETTINGS.getReaderFontId(), orientedMarginLeft, orientedMarginTop);
+        if (needsTextGrayscale) {
+          page->render(renderer, SETTINGS.getReaderFontId(), orientedMarginLeft, orientedMarginTop);
+        } else {
+          page->renderImages(renderer, SETTINGS.getReaderFontId(), orientedMarginLeft, orientedMarginTop);
+        }
```
Uri-Tauber: This block of code is repeated four times. Can be extracted to a lambda function. Something like
```
auto renderGrayscalePass = [&]() {
  if (needsTextGrayscale)
    page->render(renderer, fontId, orientedMarginLeft, orientedMarginTop);
  else
    page->renderImages(renderer, fontId, orientedMarginLeft, orientedMarginTop);
};
```

src/activities/reader/EpubReaderActivity.cpp
```diff
@@ -976,14 +976,16 @@ void EpubReaderActivity::renderContents(std::unique_ptr<Page> page, const int or
   scope.endScanAndPrewarm();
   const auto tPrewarm = millis();
 
-  // Force special handling for pages with images when anti-aliasing is on
-  bool imagePageWithAA = page->hasImages() && SETTINGS.textAntiAliasing;
+  const bool pageHasImages = page->hasImages();
+  const bool needsImageGrayscale = pageHasImages;
+  const bool needsTextGrayscale = SETTINGS.textAntiAliasing;
+  const bool needsAnyGrayscale = needsTextGrayscale || needsImageGrayscale;
```
Uri-Tauber: Nitpick: `needsImageGrayscale` is redundant. Probably leftover from previous iteration.

```suggestion
  const bool pageHasImages = page->hasImages();
  const bool needsTextGrayscale = SETTINGS.textAntiAliasing;
  const bool needsAnyGrayscale = needsTextGrayscale || pageHasImages;
```

Uri-Tauber (changes requested):
Very Good. Thanks for that @uxjulia!
Two Nitpicks.

Uri-Tauber (approved):

itsthisjustin (approved):

Issue #2381, opened 2026-06-20
Sleep on footnotes page will not allow back button to return to main chapter.


eduardo-svg:
### Affected Version

1.3.0

### Describe the Bug

After sleeping and waking the device whole reading a footnote link, the back button does not return me to the original footnote location as normal.

### Steps to Reproduce

1. Follow footnote from main EPUB text
2. Sleep device
3. Wake device
4. Press back button

### Expected Behavior

Pressing the back button should return the reader to the original hyperlink location.

### Relevant Log Output/Screenshots

```shell

```

ZephByte:
I think this is the current behavior, but anyone can correct me if I'm wrong since I'm new here. This was something I noticed while working on [return button code for fun](https://github.com/crosspoint-reader/crosspoint-reader/pull/2118#issuecomment-4566926304). Right now the footnote return point is basically just pushed onto a stack that is lost when you sleep. I had two fixes in mind:

**1.** If we sleep while the user is still on a footnote, silently auto-return to the original page before powering off. They wake up at their pre-footnote spot and can re-open the footnote if they want. Less code, but it does take an action without the user knowing.
**2.** Use something like my `returnPoint.bin` file or a custom behind-the-scenes bookmark to persist the footnote stack so it survives sleep/reboot. Now the back button works as expected on wake with no surprise side effects.

I like option 2 the most because it matches what the user would expect, but option 1 is less code and a safer addition. I might try to whip up a branch for option 1 tomorrow to see what it looks like. I'm also planning on doing more work on my return point branch, but it doesn't sound like that will be merged anytime soon, or ever (it is a little janky and out of the current project scope).

ZephByte:
The more I thought about it, the more I realized how simple option 1 would be, so I opened #2394. It silently returns to the pre-footnote position on sleep (or any exit) while a footnote is open. Not sure if this is the right call or if the current behavior is preferred, but here it is!

eduardo-svg:
@ZephByte For my part, option 1 is a fine solution. I don't think I could tell you at any given time whether I was on the original page or the footnotes the last time I slept my X4. As long as I don't end up having to navigate back to the original page manually I'm happy.

PR #2394, opened 2026-06-22
feat: footnote returns to original position on deep sleep to avoid losing reading position

ZephByte:
## Summary

  Possible fix for #2381

  If the reader leaves the book while still viewing a footnote, we silently return them to their pre-footnote reading position. This way the true reading position isn't lost, and they can jump back into the footnote again if they wish.

  The current behavior keeps the footnote return point only in RAM, so leaving the reader (most commonly via deep sleep) loses it and the user resumes stranded on the footnote, possibly far from where they navigated from.

## Additional Context
The simpler of two ideas I had for alleviating some of the pain mentioned in #2381. The fix lives in `onExit()`, so deep sleep is the main case but it also covers the other ways out of the reader while a footnote is open (long-press to the file browser, or the reader menu). It restores the original position from before the first footnote jump.

---

### AI Usage

### DIFF
--- a/USER_GUIDE.md
+++ b/USER_GUIDE.md
@@ -527,6 +527,8 @@ On the **Xteink X3**, the gyroscope can be used to turn pages by tilting the dev
 
 When reading an EPUB that contains footnotes, you can navigate to the footnote text by selecting the footnote reference in the book. From the footnote, you can return to your original reading position.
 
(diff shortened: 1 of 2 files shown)


Uri-Tauber:
Neat!
Can you update `USER_GUIDE.md` to reflect this behaviour?

ZephByte:
> Neat! Can you update `USER_GUIDE.md` to reflect this behaviour?

Done!

Uri-Tauber (approved):

PR #1541, opened 2026-03-30
feat: improve epub support of text-decoration (underline, overline, linethrough)

JiangoJ:
## Summary
* Related to https://github.com/crosspoint-reader/crosspoint-reader/issues/1398
* Adding support for `text-decorator` fields `overline` and `linethrough`
* Line decorators will now stretch across multiple words
* Doubled the width of the line

## Additional Context
- Bumped `SECTION_FILE_VERSION` and `CSS_CACHE_VERSION` 
- I utilized `EpdFontFamily::Style` in conjunction with `CssTextDecoration` as an enum bitmask for text decorators underline, overline, linethrough. The goal is to not create additional size in the per-word format state that is loaded/cached. The bitmask pattern should support future decoration types added. 
- Screenshot of epub mentioned in issue, also with more explicit test cases in `test_text_decoration.epub`.  
    - [text_decoration](https://github.com/user-attachments/files/26416923/feat_text_decoration_2.bmp)
- Specific cascade / inheritance logic follows (https://www.w3.org/TR/css-cascade-5/#cascade-sort, https://www.w3.org/TR/css-text-decor-3/#line-decoration) 
    - Line decorators are propagated to children regardless of conflicting styling
    - user-defined text-decorator overrides user-agent styling
    - InternalLinks still maintain underline despite ^ rules since there aren't any other visual indicators
---

[text_decoration](https://github.com/user-attachments/files/26416923/feat_text_decoration_2.bmp)

### AI Usage
Did you use AI tools to help write this code? 
No code was generated with AI, only used for Q&A

Issue #1398, opened 2026-03-15
Doesn't handle all text-decoration (only underline)


steka:
### Affected Version

v1.1.1

### Describe the Bug

Currently only underline is shown in the reader, and:
* with only one pixel thickness, may I suggest to increase that to two pixels
* it is also shown word by word, with no underline between words, other readers show the underline as a continuous line between the words.
* missing is overline and line-through decorations.

Screenshot:
<img width="240" height="400" alt="Image" src="https://github.com/user-attachments/assets/c8fa9fd3-c183-4f67-a1e0-3eb6ea3ea38d" />

### Steps to Reproduce

View [attached epub](https://github.com/user-attachments/files/26003932/text-decorations.zip) in the reader.

### Expected Behavior

Expected screenshot:

<img width="240" height="400" alt="Image" src="https://github.com/user-attachments/assets/c3b87d32-31da-40b0-85b2-4a4a2e14a440" />

### Relevant Log Output/Screenshots

```shell

```

jdk2pq:
@steka Thanks for writing this up! I would call this more of a "feature request" than a bug report, as this would extend the current handling of `text-decoration` to support additional values. It was intentionally not implemented to cover all cases (at least at first).

I know this would make our handling of `text-decoration` more complete according to CSS, but are overlines or strikethroughs common in EPUBs? From a very quick search, it looks like overlines only became supported in EPUB 3, which is far less common than EPUB 2. I suppose strikethrough may come up more frequently than overline, but even still, are there examples of this kind of styling in EPUBs?

steka:
Yeah sure, it might be more of a feature request.
I noticed it when reading Andy Weir's "**Project Hail Mary**" where strikethroughs where used in Eridian clock/counting system. It uses the symbols: “ℓ”, “I”, “V”, “λ”, “+”, and “V”, where the last "V" should have a strikethrough.

JiangoJ:
Hey! Am a new contributor here; saw this as a `good-first-issue` so submitted a small pr addressing just the basics of "straight-line" related `text-decoration` (overline, linethrough, and underline).

Uri-Tauber:
Closing as solved.

PR #2397, opened 2026-06-22
feat(epub): improve text-decoration support

lpla:
## Summary

  Revives and adapts #1541 on top of current `develop`, adding EPUB support for combined `text-decoration` line values. Related to #1398.

  - Supports underline and line-through text decorations.
  - Supports the corresponding HTML tags, including `<u>`, `<ins>`, `<del>`, `<s>`, and `<strike>`.
  - Preserves existing underline behavior for internal EPUB links.
  - Keeps decoration state in the existing per-word style bitmask, avoiding a larger cached per-word style record.
  - Preserves current `develop` text layout behavior, including RTL/CJK handling, Focus Reading, superscript/subscript, anchors, and footnotes.
  - Adds a focused text-decoration test EPUB adapted from #1541.
  - Updates section cache format documentation for the new serialized style bit.

## Additional Context

* This remains intentionally close to the already-reviewed but now-conflicting #1541 by @JiangoJ.
* `CSS_CACHE_VERSION` is bumped to `7`.
* `SECTION_FILE_VERSION` is bumped to `28` because section caches serialize `wordStyles`.
* CSS parsing matches whole `text-decoration` tokens, so malformed values like `notunderline` do not accidentally enable underline.
* Decoration rendering uses two fixed local trackers for underline and line-through spans; no heap-owned rendering buffer is added.
* Overline support was removed following maintainer feedback to keep the parsing and rendering scope focused on higher-value decorations for constrained hardware.

Verification performed against the merge result with current `origin/develop`:

* `pio run`
* `pio check --fail-on-defect low --fail-on-defect medium --fail-on-defect high`
* 85 native tests via CMake/CTest
* `./bin/clang-format-fix`
* `git diff --check`
* EPUB ZIP integrity check

The earlier version was tested on-device by a maintainer and reported working overall. The revised test fixture has not been rendered on hardware.

---

### AI Usage


Uri-Tauber:
Thanks @lpla.
Just add @JiangoJ as co-author. 

I'm hoping to release crosspoint 1.4 today, so we can start with this immediately afterwards.

lpla:
@Uri-Tauber Co-author added!

lpla:
@Uri-Tauber I see that 1.4 and 1.4.1 have been released! I see no conflicts with `master` branch, but I also see that you are working now in several different branches, like `master`, `release/1.5` and `develop`. Which one should be the one to be pointing this PR to?

Uri-Tauber:
@lpla `develop` is the working branch now.

I apologize for not getting to this yet. I'm very busy right now. I hope to review this soon.

lpla:
@Uri-Tauber no problem. I just changed the target branch.

Uri-Tauber:
I tested the PR on my device, and overall working well.
I do have one technical doubt: do we need to support **overline**? I never saw epub with such decoration, and it will mean a bit longer parsing time without any gain. (It maybe useful for math equations, but, in my epubs, equations are usually embedded pngs.)

line-through is very welcome, and so the fix for underline.

lpla:
I'd say these are the three text decorations available in CSS in epub2 standard. I guess that `overline` is the least frequent of all, but still, possible. For the sake of completeness of the standard, I would keep it.

Uri-Tauber:
> I'd say these are the three text decorations available in CSS in epub2 standard. I guess that overline is the least frequent of all, but still, possible. For the sake of completeness of the standard, I would keep it.

I disagree with you, but I'm not going to insist on this.
@uxjulia want to be the arbitrator here?

uxjulia:
> > I'd say these are the three text decorations available in CSS in epub2 standard. I guess that overline is the least frequent of all, but still, possible. For the sake of completeness of the standard, I would keep it.
> 
> I disagree with you, but I'm not going to insist on it. @uxjulia wants to be the arbitrator here?

I'm going to have to agree with @Uri-Tauber on this one simply because of the hardware constraint. If we didn't have such a limited device to work with, it'd be great to align with EPUB standards, so I get your point @lpla, but the more parsing the device has to do the longer indexing eventually takes (currently dealing with all the overhead in Crossink so I know this problem well) so we really need to pick high-value rendering options and pass on the lower value ones. If it becomes a high demand ask at a later time, then this can be revisited but I'd err on the side of caution for now simply because there are so many other things people want improved in terms of rendering.

lpla:
Addressed the feedback from @Uri-Tauber and @uxjulia in `4faafd26`: removed overline parsing and rendering, removed its serialized style bit and documentation, and adapted the EPUB fixture to cover underline and line-through combinations.

Revalidated against the merge result with current `develop`: firmware build passed, cppcheck found no defects, all 85 native tests passed, and formatting/diff/EPUB integrity checks passed. The renderer still uses a fixed local tracker with no new heap allocation; the tracker array is now reduced from three entries to two.

Uri-Tauber (approved):

PR #2401, opened 2026-06-23
chore: Replace product link with affiliate tracking link

Uri-Tauber:
## Summary

Update product link to affiliate URL
Replace direct Xteink product link with affiliate tracking link (go.sjv.io) in README.md.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,6 +8,8 @@ CrossPoint is open-source e-reader firmware - community-built, fully hackable, f
 
 ![CrossPoint Reader running on Xteink device](./docs/images/cover.jpg)
 
+> If you're planning to buy an Xteink device, consider purchasing an **X3/X4 Developer Edition** through https://crosspointreader.com. CrossPoint receives a small share of each sale, helping fund development costs.
+
 ## What can CrossPoint do?
 
 - **Reader engine**: EPUB 2/3 rendering with embedded-style option, image handling, hyphenation, kerning, chapter navigation, footnotes, bookmarks, go-to-percent, auto page turn, orientation control, focus reading, KOReader progress sync and more. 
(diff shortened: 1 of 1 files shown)


itsthisjustin:
I don't actually think this will work as I had to verify the domain and whitelist it for the tracking to work

Uri-Tauber:
> I don't actually think this will work as I had to verify the domain and whitelist it for the tracking to work

@itsthisjustin Why do you need a whitelist at all? Wouldn't it be more practical if the link could be accessed from anywhere?

itsthisjustin:
> > I don't actually think this will work as I had to verify the domain and whitelist it for the tracking to work
> 
> @itsthisjustin Why do you need a whitelist at all? Wouldn't it be more practical if the link could be accessed from anywhere?

Limitation of their tracking platform, not something I control

Uri-Tauber:
> Limitation of their tracking platform, not something I control

OK. Just tried to manually enter the URL, and it worked.

itsthisjustin:
> > Limitation of their tracking platform, not something I control
> 
> OK. Just tried to manually enter the URL, and it worked.

Haha yes the link will work we may just not get credit for the sale is all

Uri-Tauber:
> Haha yes the link will work we may just not get credit for the sale is all

Ok. so maybe add note saying "if you planning to buy Xteink device, you can support us by going to  ..."

What do you think?

itsthisjustin:
> > Haha yes the link will work we may just not get credit for the sale is all
> 
> Ok. so maybe add note saying "if you planning to buy Xteink device, you can support us by going to ..."
> 
> What do you think?

Yep I really like that

Uri-Tauber:
> Yep I really like that

Any thoughts?

PR #2408, opened 2026-06-24
feat: option to accept self-signed / untrusted certificates

### LINKED ISSUE #2400: Feature request: Trust self-signed certificates for local KOReader Sync servers
### Feature request

Please add an option to use self-hosted KOReader Sync servers over HTTPS when they use a self-signed certificate or a local CA.

### Context

I run KOReader Sync locally on my server. In that setup, HTTPS is often provided with a self-signed certificate or a private local CA.

Currently, CrossPoint appears to reject those certificates and the KOReader Sync authentication fails with a network/authentication error, even though the same server works from other KOReader clients when the certificate is trusted or bypassed.

### Current behavior

When configuring KOReader Sync with a local HTTPS endpoint using a self-signed certificate, authentication fails.
Example: https://192.168.1.x:7200

### Proposed behavior

In the KOReader Sync settings, add an explicit per-server option such as:
- "Trust this self-signed certificate"
- or "Allow self-signed certificate for this server"

This should apply only to the configured KOReader Sync server, not globally to all HTTPS requests.

HgGamer:
## Summary

Fixes https://github.com/crosspoint-reader/crosspoint-reader/issues/2400
* Adds an "Accept Untrusted Cert" toggle to the KOReader Sync settings menu so the reader can sync against a self-hosted sync server that uses a self-signed (or otherwise untrusted) TLS certificate. When enabled, the sync client skips certificate verification for all KOReader sync requests (auth, get progress, update progress).

Default is off — the existing verified-TLS behavior (CA bundle) is unchanged, so the public sync.koreader.rocks server keeps working exactly as before.

lib/KOReaderSync/KOReaderCredentialStore.{h,cpp} — new allowUntrustedCert flag with getter/setter.
lib/KOReaderSync/KOReaderJsonIO.cpp — persist allowUntrustedCert in /.crosspoint/koreader.json (defaults to false when absent, so existing files keep the secure behavior).
lib/KOReaderSync/KOReaderSyncClient.cpp
New performInsecure() helper: a WiFiClientSecure + HTTPClient request used by all three endpoints when the toggle is on. Same KOSync headers (x-auth-user, x-auth-key) and Basic-auth fallback as the verified path.
Extracted parseProgress() so both the verified and insecure paths share one JSON parser (no duplication).
createClient() (verified path) is untouched in behavior; it only attaches the CA bundle and is never reached when the toggle is on.
src/activities/settings/KOReaderSettingsActivity.cpp — new menu row "Accept Untrusted Cert" (Yes/No toggle), inserted before "Authenticate".
lib/I18n/translations/english.yaml — new STR_ALLOW_UNTRUSTED_CERT string. Other languages fall back to English (per the existing i18n convention).
## Additional Context

No new persistent allocations. performInsecure() builds a WiFiClientSecure on the stack only for the duration of one request. TLS memory is dominated by mbedTLS content buffers, which are identical to the existing esp-tls path, so heap cost is comparable.
The existing MIN_HEAP_FOR_TLS (55 KB) guard still gates both paths.
On device, free heap at auth time was observed at ~88 KB; the handshake completed with ample margin.

mbedTLS (the SDK's TLS library) rejects certificates with an empty Subject/Issuer DN at parse time — before verification — so setInsecure() cannot rescue such a cert. A self-signed server cert must carry a non-empty CN (e.g. generated with -subj "/CN=<host>" -addext "subjectAltName=IP:<host>"). This is a server-side requirement, not a reader limitation.

---

### AI Usage


0xFanko:
Hi ! I noticed this PR is currently blocked by merge conflicts with develop and it seems this may have gone unnoticed.

Would it be possible to revisit the conflicts and see whether this can be moved forward ? 

Thanks!

HgGamer:
develop already calls `http.setInsecure()` on every KOSync request (SecureHttpClient/wolfSSL has no CA bundle wired up 
`freeink-sdk/libs/network/SecureNet/include/SecureHttpClient.h:50`, `KOReaderSyncClient.cpp:85,124,156,270`). Self-signed servers already work, so the PR's

PR #2410, opened 2026-06-24
fix: correct font size selection

### LINKED ISSUE #2407: 1.4.0 RC3 doesn't map smaller font sizes correctly
### Affected Version

1.4.0 RC3

### Describe the Bug

When using the Font Builder on the CrossPoint website, I built my fonts at sizes 10/12/14/16 instead of the default (12/14/16/18) and did all of my reading on the Small setting. Version 1.3.0 rendered the text correctly at 10pt size, but version 1.4.0 RC3 renders Small as 12pt, despite checking and re-building the .cpfonts.

Upon re-flashing CrossPoint 1.3.0, everything works correctly again.

### Steps to Reproduce

1. Use the Font Builder and create a font with smaller sizes than default.
2. Transfer the font to the XTEink reader.
3. Set Reader Font Size to Small.

### Expected Behavior

I expect Small to map to my chosen font size

Image

.

```shell

```

Uri-Tauber:
## Summary: Fix #2407

The bug was introduced in commit 22f357506 which replaced ordinal (index-based) font size selection with a closest-to-hardcoded-target approach. The hardcoded targets {12, 14, 16, 18} match the default Font Builder sizes, but when a user builds fonts at custom sizes (10/12/14/16), "Small" targets 12pt and picks the wrong file.

The fix uses ordinal selection when the family has ≥4 sizes (covering all enum slots), and falls back to closest-match only for families with fewer files (like CJK packs that ship 2-3 sizes).

---

## Testing
Compiled Bookerly with 
```bash
python3 fontconvert_sdcard.py --intervals latin-ext,punctuation,reading,symbols --sizes 8,10,12,14 --regular /home/urit/Desktop/fonts/Bookerly/Bookerly-Regular.ttf --bold /home/urit/Desktop/fonts/Bookerly/Bookerly-Bold.ttf  --name Bookerly --italic /home/urit/Desktop/fonts/Bookerly/Bookerly-RegularItalic.ttf --bolditalic /home/urit/Desktop/fonts/Bookerly/Bookerly-BoldItalic.ttf --output-dir Bookerly/
```

Current master: 
The-Colour-of-Magic_ch7_p6_5pct_27930.bmp

With the fix:
The-Colour-of-Magic_ch7_p3_5pct_18338.bmp

### AI Usage

### DIFF
--- a/lib/EpdFont/SdCardFontRegistry.cpp
+++ b/lib/EpdFont/SdCardFontRegistry.cpp
@@ -19,6 +19,26 @@ const SdCardFontFileInfo* SdCardFontFamilyInfo::findClosestReaderSize(const uint
                                                                       const uint8_t style) const {
   if (files.empty()) return nullptr;
 
+  // Collect sizes matching the requested style, sorted ascending.
+  std::vector<uint8_t> sizes;
+  for (const auto& f : files) {
+    if (f.style != style) continue;
+    sizes.push_back(f.pointSize);
+  }
+  if (sizes.empty()) return nullptr;
+  std::sort(sizes.begin(), sizes.end());
+
+  // When the family provides at least 4 sizes, use ordinal (index-based)
+  // selection so custom-built font sets (e.g. 10/12/14/16) map SMALL to
+  // the smallest file, not to a hardcoded 12pt target.
+  if (sizes.size() >= 4) {
+    uint8_t idx = fontSizeEnum;
+    if (idx >= sizes.size()) idx = sizes.size() - 1;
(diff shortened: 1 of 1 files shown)


Uri-Tauber:
One (hopefully) last fix before the release!

itsthisjustin:
Huh interesting. I didn't really expect that change to cause an issue as I viewed the original method as an actual bug itself. We probably just shouldn't do any matching up of sizes and should instead just read them and rank them per font but this fix is fine. Sorry about the downstream issue here.

itsthisjustin (approved):

PR #2412, opened 2026-06-24
fix(reader): stabilize X3 EPUB grayscale page turns

gershalfred:
## Summary
- use the grayscale-aware base pass for ordinary X3 EPUB grayscale turns
- run grayscale preconditioning immediately after periodic HALF refresh cleanup turns

## Problem
On X3 EPUB page turns with text anti-aliasing enabled, the new page can appear too dark first and then become slightly lighter when the grayscale overlay completes.

## Root cause
`src/activities/reader/EpubReaderActivity.cpp` still uses the normal reader refresh cycle for text-only grayscale EPUB turns, even though `displayGrayscaleBase(...)` and `preconditionGrayscale(...)` already exist lower in the display stack.

## Change
Wire those existing display primitives into the EPUB reader path:
- cleanup turns: `HALF_REFRESH` then `preconditionGrayscale()`
- ordinary grayscale turns: `displayGrayscaleBase(HalDisplay::FAST_REFRESH)`

This keeps the change narrow to the reader logic because CrossPoint `master` already has the HAL / renderer support.

### Checked locally
- `git diff --check` ✅
- `pio run -e default` is currently blocked in this checkout by a pre-existing PlatformIO/sdk layout problem: `open-x4-sdk/libs/hardware/BatteryMonitor` is missing, so Library Manager aborts before compile starts

### Hardware verification requested
On an X3 with EPUB text anti-aliasing enabled:
1. open a text-heavy EPUB
2. turn to a fresh page
3. verify the first visible frame is closer to the final anti-aliased result instead of flashing too dark and then softening a moment later
4. verify periodic cleanup turns still clear ghosting correctly

### DIFF
--- a/src/activities/reader/EpubReaderActivity.cpp
+++ b/src/activities/reader/EpubReaderActivity.cpp
@@ -1065,6 +1065,21 @@ void EpubReaderActivity::renderContents(std::unique_ptr<Page> page, const int or
     // HALF ghost-cleanup path, which drives every pixel to its target
     // regardless of residue.
     pagesUntilFullRefresh = 1;
+  } else if (needsAnyGrayscale) {
+    if (pagesUntilFullRefresh <= 1) {
+      // Cleanup turns still need the stronger HALF pass, but on X3 the
+      // grayscale overlay behaves better if we immediately run the OEM settle
+      // step before writing the gray planes.
+      renderer.displayBuffer(HalDisplay::HALF_REFRESH);
+      renderer.preconditionGrayscale();
+      pagesUntilFullRefresh = SETTINGS.getRefreshFrequency();
+    } else {
+      // Use the grayscale-aware base waveform so the first visible pass is
+      // closer to the final anti-aliased result instead of flashing overly
+      // dark text first and softening a moment later.
+      renderer.displayGrayscaleBase(HalDisplay::FAST_REFRESH);
+      pagesUntilFullRefresh--;
+    }
   } else {
     ReaderUtils::displayWithRefreshCycle(renderer, pagesUntilFullRefresh);
   }


gershalfred:
Quick follow-up with current status and a tighter repro/validation frame.

I re-checked the current upstream state and this PR still appears to cover a gap that is not on `master` yet: the EPUB reader path still does not have the grayscale-only branch that uses `displayGrayscaleBase(HalDisplay::FAST_REFRESH)` for ordinary text-AA page turns or `preconditionGrayscale()` immediately after cleanup turns.

User-visible symptom this is targeting:

1. On X3, open a text-heavy EPUB with text anti-aliasing enabled.
2. Turn to a new text-only page.
3. The first visible frame appears a touch too dark.
4. A fraction of a second later, when the grayscale AA pass lands, the glyph edges soften and the text looks slightly lighter/fainter.

Why this patch is narrow:

- it only changes the no-image grayscale EPUB refresh branch;
- image/uncached-image handling remains on the existing cleanup path;
- CrossPoint already has the lower-level grayscale primitives, so this PR only wires the existing primitives into the EPUB reader decision point.

Current automated validation looks good from the PR page:

- `clang-format` passed
- `cppcheck` passed
- `build` passed
- `unit-tests` passed
- CodeRabbit generated no actionable comments

The one thing I cannot personally validate from here is hardware behavior on an X3. If a maintainer has an X3 handy, the verification is simply whether the first visible text frame is closer to the final anti-aliased appearance instead of visibly dark-first / lighter-after.

Happy to adjust the approach if you prefer a different refresh cadence, but this still looks like the smallest reader-level fix for the observed page-turn contrast shift.

itsthisjustin:
@gershalfred I'm not entirely sure I understand what this does. The current behavior is somewhat intended as it loads bw quickly, and then does a gray pass over the text to AA it

gershalfred:
Thanks — that matches my understanding of the intended pipeline.

This PR is not trying to remove the fast BW-first behavior or the later gray AA pass. The issue it is targeting is narrower: on X3 EPUB text-only turns, the first BW-visible frame can look noticeably darker/heavier than the final AA result, so the user sees a quick dark-text flash and then the text softens/lightens when the gray pass lands.

The idea here is to keep the same two-stage user experience, but make the first visible pass closer to the final grayscale/AA appearance by using the existing grayscale-aware base path for the no-image grayscale EPUB branch.

So, intended behavior preserved:
- fast initial visible page turn
- gray AA pass still follows
- cleanup / image-heavy paths stay on the existing stronger refresh behavior

What changes:
- for ordinary text-only grayscale EPUB turns, the base pass is less visually different from the final AA result
- after periodic cleanup turns, `preconditionGrayscale()` runs before writing the gray planes

If the current dark-then-soften transition is considered acceptable/intentional as-is, I’m fine closing this. But if the goal is to keep the fast turn while reducing that visible contrast shift, this seemed like the smallest reader-level hook since the lower-level grayscale primitives already exist.

itsthisjustin:
Can you compare this to #2611

gershalfred:
Thanks — yes, that matches the behavior I was trying to isolate: the visible change appears to come from the later grayscale/text anti-aliasing pass, rather than from removing the fast BW-first path.

I compared this PR with #2611. #2611 changes the same `EpubReaderActivity.cpp` area substantially for async/tiled refresh, heap pressure, and page-turn latency, but its current diff does not appear to carry over the specific grayscale-aware base/precondition behavior proposed here.

So I see #2412 as a narrow X3 EPUB text-AA presentation fix:

- preserve the fast initial BW frame;
- make that first frame closer to the eventual grayscale-AA result instead of briefly showing noticeably heavier/darker text;
- keep the stronger grayscale preconditioning on cleanup turns.

I do not want to duplicate or conflict with the broader #2611 work. If #2611 is the preferred base, I can rebase this narrow behavior onto it, or drop this PR if the async/tiled path already resolves the same intermediate-frame effect on X3. Which direction would you prefer?

itsthisjustin:
Hate to add more work to your plate but maybe just look at the logs on both branches and see what the page turn latency is between the two approaches?

gershalfred:
Good suggestion. I checked the local artifacts and GitHub branches, but I do not have captured X3 serial logs from the same device/book on both branches, so I cannot honestly claim a latency comparison yet.

The existing `ERS` logs expose the useful phase timings:

- `prewarm`
- `bw_render`
- `display`
- grayscale LSB/MSB work
- grayscale display/cleanup
- total render time

#2611 adds the async-refresh path and its own timing breakdown, while #2412 changes the grayscale-aware base/precondition behavior. The clean comparison should be on the same X3, same cached text-heavy EPUB, with text anti-aliasing enabled, using a fixed set of page turns on each branch. We should record both total render time and time-to-first-visible BW frame, since the visual issue here is specifically the transition between the fast BW frame and the later AA pass.

I agree that this is the right next validation before deciding whether #2412 should be rebased onto #2611 or closed. I do not want to claim that the narrow grayscale change is necessary until we have those measurements.

PR #2417, opened 2026-06-24
fix: Optimize Bookmark Rendering by Removing XPath Lookup

### LINKED ISSUE #2414: Adding a bookmark significantly slows page turn speed
https://imgur.com/a/uOHeZs8

### Affected Version

1.4.0, official release

### Describe the Bug

Updated to CrossPoint 1.4.0 official release on XTeink X4.

Noticed it on a book which is bit on the larger side since it’s a combination of two books, as well as the chapters/sections being large too (400+ pages for the chapter I initially tried it in). 

I tried this on a couple other books as well.  
For one with half as many total pages but similarly large chapters (150+ pages), the same issue seems to occur.

Then one with more total pages but short chapters (40-50 pages), it was much less noticeable but there was still some lag compared to baseline (same chapter without adding a bookmark). In the second example, I also tried it on a longer chapter in the same book (200+ pages) and the lag got significantly worse compared to the short chapter, so the issue seems to be related to chapter size.

### Steps to Reproduce

1. Update to 1.4.0 official full release.
2. Open up book with large chapter without any bookmarks active. 
3. Flip through pages and gain a sense of baseline page turn speed.
4. Add bookmark (long press or through menu).
5. Flip through pages. If reproduced, you should notice significantly lag between button click and page turn. 

### Expected Behavior

No lag should be present over baseline page turn speed. 

```shell

```

Uri-Tauber:
## Summary

fix #2414 

### Root Cause

`updateBookmarkFlag()` was introduced in commit 1db1442 and is executed on every render (every page turn). The function calls:

`ProgressMapper::toSavedProgress()`
→ `ChapterXPathResolver::findXPathForProgress()`

This path decompresses the current EPUB section content twice:

1. To count visible characters.
2. To resolve the corresponding XPath.

For larger sections (e.g. ~133 KB decompressed content), this adds approximately **1 second of I/O overhead per page turn**, with the cost increasing as chapter size grows.

### Fix

`updateBookmarkFlag()` only needs to determine whether a bookmark falls within the currently displayed page range.

The required information is already available during rendering:

* `currentPage`
* `section->pageCount`
* `currentSpineIndex`

Instead of converting the current location to a saved progress object (and resolving an XPath), the implementation now computes the current page's progress range directly and compares bookmark percentages against that range.

This is effectively the same percentage-based matching logic already used as a fallback in `bookmarkMatchesProgress()` when XPath matching is unavailable.

---

### AI Usage

### DIFF
--- a/src/activities/reader/EpubReaderActivity.cpp
+++ b/src/activities/reader/EpubReaderActivity.cpp
@@ -1350,11 +1350,15 @@ void EpubReaderActivity::updateBookmarkFlag() {
     currentPageBookmarked = false;
     return;
   }
-  SavedProgressPosition progress = ProgressMapper::toSavedProgress(epub, getCurrentPosition());
   const ProgressRange pageRange =
       getPageProgressRange(epub, currentSpineIndex, section->currentPage, section->pageCount);
   currentPageBookmarked = std::any_of(cachedBookmarks.begin(), cachedBookmarks.end(), [&](const BookmarkEntry& b) {
-    return bookmarkMatchesProgress(b, progress, pageRange);
+    if (b.computedSpineIndex == currentSpineIndex && b.computedChapterPageCount == section->pageCount &&
+        b.computedChapterProgress == section->currentPage) {
+      return true;
+    }
+    const float bp = std::clamp(b.percentage, 0.0f, 1.0f);
+    return bp + bookmarkProgressEpsilon >= pageRange.start && bp - bookmarkProgressEpsilon <= pageRange.end;
   });
 }
 


Uri-Tauber:
@uxjulia @itsthisjustin Important fix here.

Maybe we should do release 1.4.1 in a day or two, to see if there are more substantial fixes we should ship immediately.

itsthisjustin:
> @uxjulia @itsthisjustin Important fix here.
> 
> Maybe we should do release 1.4.1 in a day or two, to see if there are more substantial fixes we should ship immediately.

I'm good with that if we use it as a test bed for new branching strategy :)

uxjulia (dismissed):
Tested on device with an SD card font. Created 6 bookmarks throughout and paged through. No lag perceived in page turns.

uxjulia (approved):

PR #2420, opened 2026-06-24
fix: small translation changes for german

dasrecht:
small translation changes in german

also added STR_LONG_PRESS_MENU

## Summary

* **What is the goal of this PR?** Small translation changes to german to make it a bit more uniform.

---

### AI Usage


dasrecht:
Was also debating if we should do a more uniform rewrite of the german `ß` to `ss` as it's not used uniformly in german speaking countries (e.g. Germany uses it, Switzerland and Lichtenstein woudln't)

Uri-Tauber (approved):

PR #2429, opened 2026-06-25
feat: preview image files inline in web file browser

fain182:

Let users view image files directly in the web file browser. Currently clicking an image (e.g. a screenshot) forces a download; this lets you preview it inline instead.

  * Clicking an image file (`png`/`jpg`/`jpeg`/`bmp`/`gif`/`webp`) opens an in-page preview modal; non-image files keep the existing download-on-click behavior.
  * Image files now get a distinct 🖼️ icon in the file list (alongside the existing 📗 for EPUBs and 📄 for other files).
  * The modal includes a **Download** button to still save the file

## Screenshots
Screenshot From 2026-06-25 21-10-03

## Additional Context

* The preview uses an `` pointed at the existing `/download?path=...` endpoint (browsers render it inline regardless of the `Content-Disposition` header).

---

### AI Usage

### DIFF
--- a/src/network/html/FilesPage.html
+++ b/src/network/html/FilesPage.html
@@ -2040,7 +2040,7 @@
           if (fileIsImage) {
             fileTableContent += `<a href="#" onclick="openImagePreview('${filePath.replaceAll("'", "\\'")}', '${file.name.replaceAll("'", "\\'")}'); return false;" class="file-link">${escapeHtml(file.name)}</a>`;
           } else {
-            fileTableContent += `<a rel="noopener noreferrer" target="_blank" href="/download?path=${encodeURIComponent(filePath)}" class="file-link">${escapeHtml(file.name)}</a>`;
+            fileTableContent += `<a rel="noopener noreferrer" target="_blank" href="${downloadUrl(filePath)}" class="file-link">${escapeHtml(file.name)}</a>`;
           }
           fileTableContent += '</td>';
           fileTableContent += file.isEpub ? '<td><span class="epub-badge">EPUB</span></td>' : `<td>${escapeHtml(file.name.split('.').pop().toUpperCase())}</td>`;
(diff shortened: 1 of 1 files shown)


pmiossec:
Such a proposition will imply to have a previous/next buttons 😉

fain182:
Good idea! Let's treat this as a first step and add the prev/next buttons in a follow-up PR. This already improves the UX quite a bit compared to having to download images to see them. 😉

src/network/html/FilesPage.html
```diff
@@ -2002,6 +2044,29 @@ <h3>📂 Move File</h3>
...
+    document.getElementById('imagePreviewName').textContent = name;
+    img.src = url;
+    img.alt = name;
+    document.getElementById('imagePreviewDownload').href = url;
+    document.getElementById('imagePreviewModal').classList.add('open');
+  }
+
+  function closeImagePreview() {
+    document.getElementById('imagePreviewModal').classList.remove('open');
+    document.getElementById('imagePreviewImg').src = '';
```
pmiossec, line 2083: Not: Could be removed ?
Or if you want to keep it, for consistency, unset all the properties set on opening...

pmiossec (approved):

src/network/html/FilesPage.html
pmiossec: Maybe create a function taking  parameters that set/clear everything...

fain182: Good catch — I kept the reset on purpose and added a comment explaining why (57335577).

This modal is different from the others because it holds an : leaving the src set keeps the decoded bitmap in memory while the modal is closed, and on reopening with another file the old image would flash until the new one is downloaded from the device. A leftover text input value (rename/move/etc.) has neither cost, so those close handlers just toggle the class.

I left the other fields (name, alt, href) untouched for the same reason the other modals do: they aren't visible while closed and are overwritten on open.

pmiossec: >This modal is different from the others because it holds an : leaving the src set keeps the decoded bitmap in memory while the modal is closed, and on reopening with another file the old image would flash until the new one is downloaded from the device.

I suspected something like that. Thanks for the clarification.

fain182:
@itsthisjustin do you need that i rebase this on develop? or any other doubt? :slightly_smiling_face:

itsthisjustin:
> @itsthisjustin do you need that i rebase this on develop? or any other doubt? :slightly_smiling_face: 

Yes please rebase to develop

fain182:
@itsthisjustin done — rebased onto `develop` and updated the PR base accordingly. CI is green. 🙂

itsthisjustin (approved):

PR #2430, opened 2026-06-25
feat: move file manager breadcrumb into contents card header

fain182:

Improve the File Manager web UI header, which was visually confusing: the breadcrumb sat inline next to the "File Manager" heading where it read as a continuation of the title and was easy to overlook.

###What changes are included?

  * Move the breadcrumb into the contents card header, in place of the generic "Contents" label — so the path describes what the listing below it shows, and sits on its own background (the card) instead of glued to the title.
  * Drop the redundant "Contents" heading and its now-unused CSS.
  * Label home as "🏠 Home" (was a bare house emoji) and use a "›" chevron separator.
  * Make the separator readable: it used `--border-color` (near-invisible) — now `--label-color`.

## Screenshots

**Before**
Screenshot From 2026-06-25 21-36-19

**After**
Screenshot From 2026-06-25 21-46-06

---

### AI Usage

### DIFF
--- a/src/CrossPointSettings.h
+++ b/src/CrossPointSettings.h
@@ -65,6 +65,8 @@ class CrossPointSettings {
     XTC_STATUS_BAR_MODE_COUNT
   };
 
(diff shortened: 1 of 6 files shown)


Uri-Tauber (approved):
LGTM

PR #2432, opened 2026-06-25
perf: use one-shot inflate for whole-file ZIP reads

hooligan333:
## Summary
* **Goal:** Cut peak RAM during EPUB content reads on the no-PSRAM ESP32-C3.
* **Changes:** `ZipFile::readFileToMemory` now inits the inflater in one-shot mode (`init(false)`) for the DEFLATED branch — the whole entry is decompressed into a single contiguous buffer in one `read()`, so back-references resolve against that buffer and the 32KB streaming ring buffer is unnecessary. `readFileToStream` keeps `init(true)` (needs the window for cross-chunk back-refs).

## Additional Context
Removes a 32KB malloc + 32KB memset on every OPF/container/XHTML read. Pure peak-RAM/fragmentation relief; behaviour unchanged.

### AI Usage

### DIFF
--- a/lib/ZipFile/ZipFile.cpp
+++ b/lib/ZipFile/ZipFile.cpp
@@ -410,7 +410,10 @@ uint8_t* ZipFile::readFileToMemory(const char* filename, size_t* size, const boo
     ctx.readBuf = fileReadBuffer;
     ctx.readBufSize = 1024;
 
-    if (!ctx.reader.init(true)) {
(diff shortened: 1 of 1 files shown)


itsthisjustin:
Can you compare this method against #2563? I experimented with this same thing and had it backed by book cache at one point in time but I think this latest PR is more efficient for single spine or massive chapter books. I'd be interested for you to compare your results against mine and see which is winning?

itsthisjustin:
Closing as this will be handled by #2563

PR #2437, opened 2026-06-25
perf: cache section LUT offset to speed up page loads

hooligan333:
## Summary
* **Goal:** Remove redundant per-page-load header I/O.
* **Changes:** `loadPageFromSectionFile` re-seeked the header and re-read the invariant page-LUT offset on every page load, though the `Section` lives for the whole chapter. Cache `lutOffset` when the file is loaded or built (it sits immediately after `pageCount` in the header) and use it directly, with a header-read fallback.

## Additional Context
Removes a seek + small read per page load. Fallback preserves exact prior behaviour if a Section wasn't populated through the load/build paths.

### AI Usage

### DIFF
--- a/lib/Epub/Epub/Section.cpp
+++ b/lib/Epub/Epub/Section.cpp
@@ -127,6 +127,9 @@ bool Section::loadSectionFile(const int fontId, const float lineCompression, con
   }
 
   serialization::readPod(file, pageCount);
+  // lutOffset is written immediately after pageCount in the header; cache it here so
+  // loadPageFromSectionFile can skip the per-load header seek + read.
+  serialization::readPod(file, lutOffset);
   // Explicit close() required: member variable persists beyond function scope
   file.close();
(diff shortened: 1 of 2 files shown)


Uri-Tauber:
Hey @hooligan333, I see you've opened quite a few PRs.

Before we can review them properly, I need to see reproducible benchmarking results. For example, if a PR claims to speed up page loads, by how much? Does it have any impact on memory usage?

The easiest way to do this is to connect your X4 to your PC, open the debugging monitor script, and capture the same metrics on both the current develop branch and your optimized branch.

If the results show a measurable improvement, please post them here along with clear reproduction steps. For an example of the kind of benchmark I'm looking for, see my comment here: https://github.com/crosspoint-reader/crosspoint-reader/pull/2270#issuecomment-4630656326 .

P.S. Thanks for splitting the changes into several small PRs instead of bundling everything into one. It makes reviewing them much easier.

PR #2441, opened 2026-06-25
perf: cache cumulative spine sizes in RAM

hooligan333:
## Summary

* **Goal:** Make the repeated cumulative-spine-size reads O(1) RAM lookups instead of SD seeks.
* **Changes:** `BookMetadataCache::load()` reads the per-spine cumulative sizes into a `std::vector<uint32_t>` in one sequential pass (spine entries are stored contiguously in index order right after the LUTs), served via a new `getCumulativeSize()`. `Epub::getCumulativeSpineItemSize` is repointed at it, so every consumer benefits.

## Additional Context

`calculateProgress` hits this ~3x per render and `jumpToPercent` loops it over the whole spine; each call previously did 2 seeks + a heap-allocating `SpineEntry` read (throwaway href string) off `book.bin`. ~4 bytes/spine item (<1 KB typical), and **no `book.bin` format change**. Reviewer focus: the `load()` population reads spine entries sequentially starting at `lutOffset + lutSize`.

---

### AI Usage

### DIFF
--- a/lib/Epub/Epub.cpp
+++ b/lib/Epub/Epub.cpp
@@ -790,7 +790,12 @@ int Epub::getSpineItemsCount() const {
   return bookMetadataCache->getSpineCount();
 }
 
-size_t Epub::getCumulativeSpineItemSize(const int spineIndex) const { return getSpineItem(spineIndex).cumulativeSize; }
+size_t Epub::getCumulativeSpineItemSize(const int spineIndex) const {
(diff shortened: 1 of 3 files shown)


Uri-Tauber (approved):
LGTM

itsthisjustin:
@Uri-Tauber I think this needed to be moved to develop not master

Uri-Tauber:
@itsthisjustin Good catch.
I didnt noticed it was still pointing at master.

PR #2444, opened 2026-06-26
fix: keep status bar indicators stable when toggling bookmarks

uxjulia:
## Summary

  * Keep the reader status bar’s battery and left-side clock visually stable when a bookmark is toggled on the current page.
  * Moves the bookmark status icon layout after the battery and clock layout in `BaseTheme::drawStatusBar`.
  * Places the bookmark icon to the right of the battery, or to the right of both battery and left-positioned X3 clock when that clock is enabled.
  * Preserves the existing right-side clock behavior.

## Additional Context

* This is a layout-only change; it does not change bookmark storage, status bar settings, or battery/clock rendering logic.
* Practical behavior: the conditional bookmark icon is now the item that appears/disappears at the end of the left cluster, instead of pushing stable indicators around.

Disorientation_ch7_p1_1pct_32939 Medium
Disorientation_ch7_p1_1pct_242877 Medium
Disorientation_ch7_p1_1pct_279405 Medium
Disorientation_ch7_p1_1pct_309007 Medium

---

### AI Usage

### DIFF
--- a/src/components/themes/BaseTheme.cpp
+++ b/src/components/themes/BaseTheme.cpp
@@ -803,13 +803,6 @@ void BaseTheme::drawStatusBar(GfxRenderer& renderer, const float bookProgress, c
     renderer.fillRect(barMarginLeft, progressBarY, barWidth, barHeight, true);
   }
 
-  // Draw Bookmark
-  if (showStatusBarTextLane && isPageBookmarked) {
-    const int bookmarkY = textY + 5;
-    drawBookmarkStatusIcon(renderer, leftClusterX, bookmarkY);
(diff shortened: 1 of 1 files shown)


Uri-Tauber (approved):
LGTM

Uri-Tauber:
Very neat!

pmiossec:
@uxjulia @Uri-Tauber First, sorry for the late comment, I just saw this PR once merged:

I think it makes more sens to put it in the right side of the status bar, next to the page and progress. That will make this side "book related" and the left side more "system related".

Side note: clock make more sense displayed on the left (for the same "system related" reason). I know that it will simplify the code to remove the option to change the side of the clock if you decide to move bookmark to the right. I let you think, about it...

Issue #1984, opened 2026-05-14
Feature Request: Search book for text


cmrivera0991-ctrl:
Would be great to have the ability to search for a specific few words. I often read / listen to books in multiple formats ( physical book , audiobook, ebook) and depending on the book it's sometimes difficult to find the same location. With koreader you can search specific text, wonder if that could be possible to implement. I know with the button layout it may be difficult, but could implement a number pad text like back in the old texting days to minimize navigation

PR #2451, opened 2026-06-27
feat: in-book search

chongfun:
## Summary

  - Implement in-book search (https://github.com/crosspoint-reader/crosspoint-reader/discussions/456, https://github.com/crosspoint-reader/crosspoint-reader/issues/1984)

  - In-book search is implemented for EPUBs as a forward scan over compact text records stored beside the rendered pages in each section cache.
    - queries are limited to 64 UTF-8 bytes
    - search starts at the current rendered page and moves forward
    - after reaching the end of the spine, it continues through later spines
    - after reaching the end of the book, it wraps once and stops at the page the search was initiated from
    - the first matching page is returned immediately
    - repeating the same query from the page returned by search starts at the next page, and the wrap stops before that originating page, so "find next" advances to a different match or reports no matches rather than re-returning it
    - while searching, the status screen shows an approximate percentage of how much of the scan has completed, measured from where the search began (so it rises from 0% to 100% over the whole scan even when the search starts mid-book)

search_demo

I've tested searching with the X4 emulator and flashed the build onto my X3 successfully.

## Additional Context
See [In-Book Search Architecture](https://github.com/chongfun/crosspoint-reader/blob/d552131b725e9dfe40242128fe67b51817988b26/docs/search-architecture.md) for more context, but I've copied the trade-offs section here:

- A cold search can be slow. Reaching an uncached spine requires normal EPUB layout and may write images and section data before scanning can continue.
- Cancellation is handled between page scans. An individual uncached-section layout remains a blocking unit of work.
- The cache uses more SD space: roughly the rendered text size plus 8 bytes per page.
- Matches are page-level. Repeating a query skips the rest of the current page, so multiple occurrences on one page are not individually navigable.
- There is no match highlighting or result list.
- Case-insensitive matching is ASCII-only. Non-ASCII case variants must match exactly.
- Search text is reconstructed from rendered word tokens with single spaces, so it can differ from the EPUB source in spacing and in words split by layout-time hyphenation. Matching ignores ASCII spaces and hyphens and carries match state across adjacent same-spine pages to absorb these — including hyphenation and phrases split across a page boundary (see Matching algorithm). Other punctuation-glyph differences (curly vs straight quotes, em dash, the ellipsis character vs three dots) are not normalized and can still cause a miss, and a match split across a chapter (spine) boundary is not joined.
- Search results depend on the current layout settings. Font, viewport, orientation, margins, paragraph settings, hyphenation, embedded CSS, image mode, or Focus Reading changes can invalidate and rebuild section caches.
- The X3 display path runs at 16 MHz rather than the X4's 40 MHz SPI rate. Search paints the status screen when entering, when changing state, and when the progress percentage advances a whole repaint step, but it does not refresh the e-ink panel for every page scanned; page matching remains an SD/CPU operation. Progress is interpolated within the current spine so it advances per page (not only at chapter boundaries), and the repaint step bounds total e-ink refreshes regardless of how many spines the book has.

---

### AI Usage


pmiossec:
I would put it below "Toggle bookmark" as for search one more click is not that important contrary to bookmark where you want a quick access...

Uri-Tauber:
@chongfun Hey man, thanks for the contribution! It looks like you put a lot of effort into this.

Right now, though, the UX for typing text using the navigation buttons is just too slow to be practical. However, we're working on porting CrossPoint to touch e-reader, and once that's officially merged, your work here will be much more meaningful.

Also, @itsthisjustin is working on Bluetooth keyboard support, which would also make more sense for this feature.

So I think we should hold off on this for now and revisit it after one of those two features has been merged.

Uri-Tauber:
@chongfun You can draft it, instead of closing.

PR #2452, opened 2026-06-27
feat: Lazy incremental EPUB section indexing

### LINKED ISSUE #1067: Auto Indexing Extremely Large Sections
I have read quite a few books that don't have chapters, the last couple books I read I think it split them into sections of about 200 pages on the X4, which is great. The latest book I opened was Guards! Guards! by Terry Pratchett (from Kobo then DRM stripped) and it made the entire main section of the book into one section at a little less than 1,000 pages.

There are a couple disadvantages to this

- navigation is slightly less convenient, you basically have to use "go to %" no matter what
- indexing time is greatly increased with such a large section

I know this is a bit niche, but I would appreciate it if it could be adjusted to automatically create multiple smaller sections instead of using 300+ page sections!

itsthisjustin:
## Summary

Reworks how EPUB sections are indexed so the reader shows page 1 almost immediately and lays out the rest of a chapter in the background, instead of blocking the UI while it builds the whole section up front. Adds a per-interval "uniform advance" to the SD-card font format (cpfont v5) so CJK text measures with zero per-glyph SD reads. 

## Why

Indexing was previously done one whole spine item at a time, synchronously, on the render task. That is fine for an average chapter, but it falls apart at the extremes:

- A book whose front matter is dozens of tiny one-page spine items blocked on (and re-drew an INDEXING popup for) every page turn. - A book that is one giant chapter (a whole novel in a single spine item, hundreds of pages) froze up to a minute laying out the entire thing before showing page 1. - For SD fonts, CJK measurement hammered a 768-entry per-style advance cache that thrashes badly on large-vocabulary text.

### Incremental / lazy section build

- The HTML parser (`ChapterHtmlSlimParser`) is now resumable: `beginParse()` / `parseStep()` / `finishParse()` / `abortParse()`. `parseAndBuildPages()` is now a thin one-shot wrapper over them. - `Section` gains an incremental build API: `startBuild()`, `buildSomeMore(maxPages)`, `finalizeBuild()`, `abandonBuild()`, `loadPageDuringBuild()`, `isBuilding()`, `isBuildComplete()`. `createSectionFile()` is now the one-shot wrapper over these. - The reader builds only enough to show the landing page, renders it, then continues building in `loop()` off the page-turn critical path. Pages already laid out turn instantly; turning past the build watermark catches up on demand. - Partial reads use the same `O_RDWR` `.bin` handle (seek-read-restore) via the in-RAM page table, so no second file handle and no separate index are needed.

### Giant single-spine handling

- The background build keeps only a small window of pages ahead of the reader (`BUILD_WINDOW_AHEAD`), then stops and lets the device sleep, resuming as the reader advances. A chapter smaller than the window still builds fully and caches as before, so normal books are unaffected. - Status bar shows a byte-based estimate ("page X of ~Y") while a giant spine is still building

### Caching

- The unzipped section HTML is cached per book (keyed on the book, not on render settings) and is promoted to the cache as soon as the inflate succeeds. This means even a window-only giant spine, whose `.bin` never finalizes, still caches its HTML, so re-opening skips the multi-second zip inflation.

### CJK font advances (cpfont v5)

- The cpfont format gains a per-Unicode-interval uniform advance: if every glyph in an interval shares one advance, it is stored once. A CJK block is fixed-width, so it collapses to a single value; the reader resolves those advances with no per-glyph SD reads and no entry in the 768-cap advance table (the cap thrash is gone for CJK). Proportional fonts (e.g. Latin) have no dominant advance, so they keep the existing per-glyph path unchanged. - Generator (`fontconvert_sdcard.py`) computes and writes it; reader (`SdCardFont`) loads a parallel per-style array, allocated only when some interval is uniform. - Validated end to end on Noto Sans SC: the Han block is 100% uniform (one advance across all glyphs); ASCII correctly non-uniform.

## Known limitations / follow-ups

The one downside to the approach I'm using for the single spine books is that the total pages number isn't 100% accurate on immediate load and gets more accurate over time as you load pages. imo it's a worthy trade off but up for discussion. I compare the pages I laid out to the total size of the file and guess at the number of pages. I was off by about 4% off using this math and as you turn pages it'll get more accurate until its the same.


uxjulia:
SD card fonts don't seem to work anymore on this branch. The font preview stopped working too.

itsthisjustin:
> SD card fonts don't seem to work anymore on this branch. The font preview stopped working too.

That's odd. The sd fonts loaded for me. However that's because the sd fonts got bumped and we'd need to regenerate them. I can pull the changes to the sd font out and make it it's own branch though

itsthisjustin:
Reverted font stuff @uxjulia

Uri-Tauber:
Tested on my device with few books, and seems to work well.
I hope I will have the time to review today.

src/activities/reader/EpubReaderActivity.cpp
```diff
@@ -977,36 +1079,28 @@ void EpubReaderActivity::render(RenderLock&& lock) {
   }
 }
 
-void EpubReaderActivity::silentIndexNextChapterIfNeeded(const uint16_t viewportWidth, const uint16_t viewportHeight) {
```
Uri-Tauber, line 981: It seems a bit unrelated to remove this behaviour here.
I can understand the logic here, but this is a behaviour regression: next chapter will now trigger a fresh popup-visible indexing right at the chapter turn, where before it was invisible.

itsthisjustin: But won't ever do this because we always are indexing the next chapter

Uri-Tauber: So I'm missing something here: with the lazy indexing, when we reach spine end, we immediately start the next one? or only when the user enter the next spine?

itsthisjustin: It's unrelated to spine end. We index new pages ahead of your current page as you page through the book

Uri-Tauber: OK. thanks for clerifying.

Uri-Tauber:
@itsthisjustin I suspect there is some overlap between here and #2486.
Need to think how to cleanly adjust it to your work here.

itsthisjustin:
> @itsthisjustin I suspect there is some overlap between here and #2486.
> Need to think how to cleanly adjust it to your work here.

So I thought about that but my understanding was this was related to the reader menu and toc nav which shouldn't be impacted here

Uri-Tauber (approved):

uxjulia:
The only nit I have with this is that re-entering a book mid-chapter requires a re-index everytime. And the further you're into a large chapter, the longer it takes. Example, you're on page 100/178 of Small Gods by Terry Pratchett. The `section.bin` doesn't get created until the entire section has been indexed (towards the end), so leaving the book and re-entering can cause a larger delay than what users are used to. With small chapter sections, there's barely a delay (but that's not what this PR targets anyways). Books with huge sections (which this is trying to help), there's a noticeable delay. So initial start to reading is faster than usual, but subsequent entries into the book can cause a delay. Something to keep in mind as a potential complaint. This doesn't affect most books, but it will affect the problem books that this PR is trying to solve.

Uri-Tauber:
@uxjulia Nice catch! 
I didn't noticed that.

itsthisjustin:
> @uxjulia Nice catch! 
> I didn't noticed that.

Ah yes I did notice this and had planned a fix for it but forgot to. I know what to do. Let me update

itsthisjustin:
@uxjulia @Uri-Tauber try the single spin resume now :)

Go 100 pages deep into a book and then back out and go back in

src/activities/reader/EpubReaderBookmarksActivity.cpp
```diff
@@ -108,7 +108,15 @@ void EpubReaderBookmarksActivity::loop() {
       return;
     }
     auto bookmark = bookmarks.at(selectorIndex);
-    CrossPointPosition pos = ProgressMapper::toCrossPoint(epub, {bookmark.xpath, bookmark.percentage}, renderer);
+    CrossPointPosition pos{};
+    if (bookmark.computedChapterPageCount > 0 && bookmark.computedChapterProgress < bookmark.computedChapterPageCount &&
```
Uri-Tauber, line 114: @uxjulia Should we also check here if the cached pagination still matches current render settings?

Uri-Tauber:
@uxjulia Addressed the points you mentioned in the chat.
Want to check I didn't missed anything?

uxjulia:
@Uri-Tauber - I found a bug when navigating backwards into the previous chapter and pushed the fix. I think this looks good now.

uxjulia (approved):

PR #2462, opened 2026-06-28
fix: random spaces between words

itsthisjustin:
Fixes random words having 5-10 spaces between them on lines when not using hyphenation. 
Untitled design - A4 (Landscape) - Vivaldi 2026-06-28 at 4 55 58 AM


Uri-Tauber:
I'm not sure I understand what was the problem this fix.
The left image shows justification to both sides, while the right image justify to left.

uxjulia:
Yeah, not sure the right side image is any better for justified text?

itsthisjustin:
Lol no way! You'd rather have random spaces between words than properly justified text? The right image is correct justification given word lengths. I think this example maybe looks bad because the lines end in long words so less words fit.

Uri-Tauber:
@itsthisjustin This **is** properly justified text, man!

You can change it in settings->reader->reader paragraph alignment.

Uri-Tauber:
What just happened?

PR #2465, opened 2026-06-28
feat(epub): bilingual view mode toggle for translated books

puritysb:
## Summary

* **Goal:** Add an in-reader **Bilingual View Mode** toggle for translated EPUBs.
* **Changes:** In-reader toggle between **Both / Original-only / Translation-only** for EPUBs whose paragraphs carry `class="cp-original"` or `class="cp-translation"`.
  * `ChapterHtmlSlimParser`: token-exact class match; sets `CssDisplay::None` on the complementary half before the existing skip path.
  * Section cache bumped **v27 → v28** — one `u8 bilingualViewMode` field added to the header (between `focusReadingEnabled` and the `pageCount` placeholder) so mode changes invalidate the per-section cache automatically. All other fields unchanged from v27.
  * Reader: `cycleBilingualMode()` drops the current section and re-renders; a transient popup names the new mode. Default binding is long-press Confirm (configurable in Settings → Controls → Long-press Menu). Exposed under Settings → Reader.
  * `docs/bilingual-epub.md` documents the marker contract; `scripts/generate_bilingual_test_epub.py` produces a 2.4 KB sample for smoke testing.

## Additional Context

* The marker classes (`cp-original` / `cp-translation`) are an **opt-in contract** a translation pipeline emits; EPUBs without them are completely unaffected — no behavior change, only a 1-byte header bump in the section cache.
* Self-contained to the EPUB / reader / cache layers — no new heap allocations in the render path, no struct layout changes beyond the documented header field.
* ✅ Tested on-device (X4) with a marked-up bilingual EPUB: all three modes (Both / Original-only / Translation-only) cycle correctly and the section cache re-renders on mode change. `pio run` (default env) builds clean.

---

### AI Usage

Did you use AI tools to help write this code? **YES** — authored with Claude Code, reviewed and tested locally.


puritysb:
Superseded by #2498, which reimplements this with a hybrid xml:lang + cp-* marker resolver (works with standard producers like bookfere without a conversion pass) and a self-contained, producer-neutral doc set. Closing to keep review in a single thread.

PR #2461, opened 2026-06-28
feat(reader): Fast (blue-noise dithered) grayscale render mode

rhythmerc:
## Summary

Adds a new **Settings → Reader → Grayscale Render Mode: Quality / Fast** option for the text/EPUB readers.

- **Quality** (default): the existing 4-level grayscale path. Unchanged.
- **Fast**: renders anti-aliased text and images in a **single BW pass** using a 64×64 blue-noise ordered dither, instead of the LSB/MSB grayscale planes + grayscale waveform. Much faster perceived page turns, especially on image-heavy pages.

The setting is intentionally **independent of the existing Text Anti-Aliasing toggle**, because it also governs image rendering (images get the smooth treatment regardless of the text-AA toggle).

### Image quality
Images are dithered **straight from the 8-bit gray at decode time** (a single ordered pattern), rather than blue-noising an already-Bayer-quantized value — stacking two ordered patterns looked muddy. Fast mode writes its own pixel cache (`.pxcd`) keyed separately from the Quality 4-level cache (`.pxc`), so the two never collide and neither rebuilds when you toggle the setting.

## Credit

The blue-noise dithering idea — specifically the **64×64 void-and-cluster blue-noise threshold table** and the **per-pixel ink-threshold** approach — is drawn from @jeremydk's excellent PR #2179 (*"Fast Mode: ~6x faster perceived page turns"*).

This PR deliberately takes **only** the dithering core. It does **not** include the other two optimizations bundled in #2179 (next-page prerender and the dedicated 200 Hz input-poll task) — those are orthogonal and can be considered separately.

## Measurements (X4)

End-to-end `Rendered page in …`:

| Page | Quality | Fast | Speedup |
|------|--------:|-----:|--------:|
| Dune cover (full-page image) | 3330 ms | 600 ms | ~5.6× |
| Dune ch.1 p.2 (text) | 765 ms | 584 ms | ~1.3× |

<details><summary>Raw serial logs</summary>

**Dune cover page**
```
Quality - [115537] [DBG] [GFX] Time = 939 ms from clearScreen to displayBuffer
        - [117953] [DBG] [ERS] Page render (tiled): prewarm=0ms bw_render=238ms display=1121ms gray_lsb=946ms gray_msb=949ms gray_display=63ms cleanup=12ms total=3329ms
        - [117954] [DBG] [ERS] Rendered page in 3330ms

Fast    - [264728] [DBG] [GFX] Time = 185 ms from clearScreen to displayBuffer
[...]

PR #2466, opened 2026-06-28
feat(fonts): crispier UI fonts using native monochrome rasterizer

fain182:
  Make the built-in 1-bit UI fonts render crisply at small sizes. The UI fonts
  (Ubuntu 10/12, Noto Sans 8) looked heavy and slightly fuzzy: stems were
  over-inked and edges uneven. The cause was the conversion pipeline, not the
  typeface — `fontconvert.py` rendered each glyph as antialiased greyscale and
  then thresholded *any* coverage of roughly 12%+ to black (the "treat any 2+ as
  black" path). That fattened stems and produced irregular edges once collapsed
  to 1 bit.

###What changes are included?

  * `fontconvert.py`: the 1-bit path now uses FreeType's **native monochrome
    rasterizer** (`FT_LOAD_TARGET_MONO`) — hinted and drop-out controlled — and
    repacks that bitmap directly, instead of greyscale-then-threshold. The 2-bit
    (reading font) path is unchanged.
  * Regenerated the five 1-bit headers: `ubuntu_10/12` regular+bold,
    `notosans_8_regular`.

#### Before

IMG_8894

#### After 
IMG_8895

## Additional Context

* **Glyphs stay 1 bit/pixel** — this is *monochrome* rasterization, not a switch
  to a monospace font. Advance widths come from `linearHoriAdvance` and are
  unchanged, so **text layout / words-per-line are identical**; only the inked
  pixels differ.
* **Smaller, not larger**: bitmap data shrinks ~7–10% per font (e.g.
  `ubuntu_12_regular` 25,740 → 23,938 bytes) because the old threshold over-inked.
* **No regression to reading fonts**: the 2-bit path output is byte-identical
  (verified by regenerating a Noto face with the patched script).
* **Verification**: `pio run` builds clean (RAM 30.9%, Flash 79.6%); `pio check`
  reports no defects; flashed to the Xteink X4 and confirmed crisper UI text. No
  cache changes — UI fonts aren't cached and reading fonts are untouched, so no
  `.crosspoint/` invalidation is required.
* **Minor note for reviewers**: regenerating the headers today also merges a few
  adjacent Unicode intervals (current fontTools/FreeType vs. the toolchain that
  produced the committed files). It's an equivalent representation, unrelated to
  the monochrome change — it appears with the unpatched script too.

---

### AI Usage


rhythmerc (approved):
Fantastic find and a super-minimal diff to go with it.

If you have any interest, I think packing 1bpp + 2bpp representations into the reader fonts would be a huge improvement over the current no-AA path in the reader (FreeType's monochrome render path is very clearly superior to our threshold downscale), and I think it would also represent an improvement over [this PR](https://github.com/crosspoint-reader/crosspoint-reader/pull/2461)'s attempt at dithered AA.

Could be a worthwhile follow-up!

itsthisjustin:
This looks rad

itsthisjustin:
@fain182 can you make develop the main branch and merge it into this so this is up to date?

fain182:
Thanks! I rebased it @itsthisjustin

jpirnay:
A word of caution: this works well for fonts like the one crosspoint-reader is using where stems are comfortably ≥2px and the font is well-hinted. It's the wrong call for a thin face like Inter, where the sub-pixel stem width has to be biased back to a solid pixel. Having said that , no need to change something, but worth revisiting, when a different font choice might be made in the future.

rmmh:
I looked into this the other day and came to a similar conclusion. Very few fonts are truly optimized for 1-bit rendering at these point sizes now.

Here are some samples to compare the choices a bit more. You'll want to view them zoomed in without blur (NEAREST):

inter
ubuntu
notoserif
notosans

For comparison, a few Windows fonts:
arial
tahoma
georgia

Ubuntu clearly has a manually-hinted bitmap font at these low sizes, so it looks excellent in the mono regime.

Script here: 
[render_font_samples.py](https://github.com/user-attachments/files/29494541/render_font_samples.py)

itsthisjustin:
Closing this as for now we don't have a good way to get crisp UI fonts across the board without causing super thin small fonts that folks have complained look poor. This is a super solid implementation though and I suspect we will loop back to this eventually.

PR #2470, opened 2026-06-28
fix(icons): align home menu icons with their labels

fain182:
## Summary

  Make the home menu icons line up with their labels. They sat noticeably
  below the text, and the offset could not be tuned away — moving the icon by
  1px often did nothing, then jumped by 8px.

  * `GfxRenderer::drawIcon` now plots the icon **pixel-by-pixel via `drawPixel`**
    instead of `display.drawImageTransparent`. The blit copies the bitmap into
    the panel framebuffer one byte at a time, so the icon's X write is
    byte-aligned — and because the UI is drawn rotated, that byte alignment
    lands on the **vertical** axis, snapping the icon to an 8px grid. Per-pixel
    plotting is pixel-precise and reproduces the same on-screen orientation.
  * With pixel-precise placement, align the menu icon's top with the label's
    text cell (`iconY = textY`), which centers it on the capital letters.
  * `drawIcon` now takes a single `size` instead of `(width, height)`: every
    icon is square, so two dimensions were redundant.

### Before
IMG_8894

### After
IMG_8898

## Additional Context

* **Why per-pixel is fine here**: `drawIcon` is only called by the UI themes
  (Lyra / Lyra3Covers / RoundedRaff), which all render in forced Portrait, so
  the fixed orientation mapping is correct for every caller. It runs ~5 times
  per menu/list render (not a per-frame hot path — the e-ink refresh dominates),
  at ~1024 `drawPixel` calls per 32×32 icon.
* **Verified on hardware** (Xteink X4): home menu icons aligned, file-browser
  list icons and icon orientation unchanged.
* No RAM/flash impact of note; no cache or asset changes.

---

### AI Usage

### DIFF
--- a/lib/GfxRenderer/GfxRenderer.cpp
+++ b/lib/GfxRenderer/GfxRenderer.cpp
@@ -1054,8 +1054,24 @@ void GfxRenderer::drawImage(const uint8_t bitmap[], const int x, const int y, co
   display.drawImage(bitmap, rotatedX, rotatedY, width, height);
 }
 
-void GfxRenderer::drawIcon(const uint8_t bitmap[], const int x, const int y, const int width, const int height) const {
-  display.drawImageTransparent(bitmap, y, getScreenWidth() - width - x, height, width);
+void GfxRenderer::drawIcon(const uint8_t bitmap[], const int x, const int y, const int size) const {
+  // Plot the icon pixel-by-pixel through drawPixel (which applies the orientation
+  // transform) instead of the byte-aligned framebuffer blit. The blit snaps the
+  // icon's position to 8px (one byte) along the rotated axis, which prevents it
+  // from aligning with adjacent text; per-pixel plotting is pixel-precise.
+  // Icons are square and 1bpp (MSB-first, bit==0 = ink). The (size-1-row, col)
+  // mapping reproduces the Portrait orientation the blit produced; drawIcon is
+  // only called by the UI themes, which all render in forced Portrait.
+  const int rowBytes = (size + 7) / 8;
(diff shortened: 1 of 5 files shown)


itsthisjustin:
so I THINK I have this done already in another branch that I'm hoping to get merged soon which changes how all the icons are drawn to get away from all the weird manual positioning. If you can give it a shot, check out #2472

fain182:
Thanks for the pointer — I took a look at #2472.

You're right that it reworks the home menu icons: the new drawIcon(freeink::Icon&, …) overload draws per-pixel via drawPixel and the Lyra menu/list call sites now center via opticalCenterY. So for those specific icons there's clear overlap with this PR.

One thing worth flagging though: #2472 keeps the legacy drawIcon(bitmap, x, y, width, height) overload unchanged — it still goes through display.drawImageTransparent, so the 8px byte-alignment snap (the actual root cause here) is still present. It's just no longer hit by the menu icons, but it's still used by icon_bookmark_16 (BaseTheme) and the CoverIcon calls.

This PR instead fixes that root overload directly, so it also covers the bookmark/cover icons, and it's a small, self-contained change that's green now — whereas #2472 is a much larger effort (touch + S3 + freeink auto-scaling) on a longer timeline.

Happy to go whichever way works best for you:
- keep this as a focused bridge fix until #2472 lands, or
- close it in favor of #2472 if that's landing soon (in which case it might still be worth fixing the legacy overload there so bookmark/cover icons don't keep the snap to the 8px grid).

itsthisjustin:
Lets get this merged and then i'll rework my branch based off these changes

itsthisjustin (approved):

PR #2471, opened 2026-06-29
fix: x4 sleep/boot ghosting

itsthisjustin:
I'd say this is up for discussion as it DOES cause flashing on boot and sleep but it's what the stock firmware does and it will prevent ghosting users are seeing on sleep screens. I imagine if the stock firmware does it, we probably should too

### DIFF
--- a/src/activities/boot_sleep/SleepActivity.cpp
+++ b/src/activities/boot_sleep/SleepActivity.cpp
@@ -163,7 +163,7 @@ void SleepActivity::renderDefaultSleepScreen() const {
     renderer.invertScreen();
   }
 
-  renderer.displayBuffer(HalDisplay::HALF_REFRESH);
+  renderer.displayBuffer(HalDisplay::FULL_REFRESH);
 }
 
 void SleepActivity::renderBitmapSleepScreen(const Bitmap& bitmap) const {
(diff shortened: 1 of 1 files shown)


Uri-Tauber (approved):
I didn't experienced those ghosting, but the change looks minimal and reasonable.

itsthisjustin:
> I didn't experienced those ghosting, but the change looks minimal and reasonable.

I think it really depends on your cover and grays

Uri-Tauber:
@itsthisjustin Do you have a specific book causing ghosting.

itsthisjustin:
> @itsthisjustin Do you have a specific book causing ghosting.

No but reports on reddit. A lot of folks using manga covers

InkyDevWork:
Hi, I don't know if this is the right place but I tried v 1.4.0 and the flashing has rendered the "Quick Resume" sleep screen option effectively useless and ended up downgrading - perhaps the scope of this fix can be limited for those who have the "Cover" sleep screen settings?

itsthisjustin:
> Hi, I don't know if this is the right place but I tried v 1.4.0 and the flashing has rendered the "Quick Resume" sleep screen option effectively useless and ended up downgrading - perhaps the scope of this fix can be limited for those who have the "Cover" sleep screen settings?

@Uri-Tauber I also think this fix somehow caused the quick resume and that cover thing to regress. I suspect its simply a race conditions with setting the refresh vs setting the "cover". I noticed it today with quick resume while testing. I can fix. 

(Edit just noticed this was present on 1.4.0? so maybe not related tbh. However I did see the quick resume bug today where it sometimes just saves a white screen)

uxjulia:
> > Hi, I don't know if this is the right place but I tried v 1.4.0 and the flashing has rendered the "Quick Resume" sleep screen option effectively useless and ended up downgrading - perhaps the scope of this fix can be limited for those who have the "Cover" sleep screen settings?
> 
> @Uri-Tauber I also think this fix somehow caused the quick resume and that cover thing to regress. I suspect its simply a race conditions with setting the refresh vs setting the "cover". I noticed it today with quick resume while testing. I can fix.
> 
> (Edit just noticed this was present on 1.4.0? so maybe not related tbh. However I did see the quick resume bug today where it sometimes just saves a white screen)

I don't think the flashing was in 1.4.0... isn't it only in a nightly?

itsthisjustin:
> > > Hi, I don't know if this is the right place but I tried v 1.4.0 and the flashing has rendered the "Quick Resume" sleep screen option effectively useless and ended up downgrading - perhaps the scope of this fix can be limited for those who have the "Cover" sleep screen settings?
> > 
> > @Uri-Tauber I also think this fix somehow caused the quick resume and that cover thing to regress. I suspect its simply a race conditions with setting the refresh vs setting the "cover". I noticed it today with quick resume while testing. I can fix.
> > 
> > (Edit just noticed this was present on 1.4.0? so maybe not related tbh. However I did see the quick resume bug today where it sometimes just saves a white screen)
> 
> I don't think the flashing was in 1.4.0... isn't it only in a nightly?

Yep hence my confusion haha

PR #2472, opened 2026-06-29
feat: Adds support for touch screen devices, ESP32-S3, and auto-scaling UI elements

itsthisjustin:
Uses the freeinksdk to automatically scale icons and ui elements and adds touch screen gestures for back, page turns, and tap zones for all UI elements.


Uri-Tauber:
That's quite a big PR, man.
I really would prefer if you could split it to several smaller PRs - each one targeting different aspect. Will be much easier to review.

Also, check https://github.com/crosspoint-reader/crosspoint-reader/pull/2294, should we merge it to `develop` or into `feat-touch`?

itsthisjustin:
> That's quite a big PR, man. I really would prefer if you could split it to several smaller PRs - each one targeting different aspect. Will be much easier to review.
> 
> Also, check #2294, should we merge it to `develop` or into `feat-touch`?

I'll think on how this can be done. Adapting the firmware for more devices isn't a small change though. Every interaction needs touch added and this branch is insanely stable. It works on existing xteink issues without any regressions and I use it daily across multiple touch devices. 

On your PR, could you help me understand what it's doing that isn't magically handled by the SDK already? We can definitely merge it first and then I can rework this branch against that update.

PR #2475, opened 2026-06-29
fix: oom exceptions for OPDS, KOSync, and OTA via wolfssl

### LINKED ISSUE #1681: OPDS: „Failed to fetch feed“ for some feeds from calibre web automated instance
### Affected Version

1.2.0

### Describe the Bug

I have an instance of calibre web automated (CWA) hosted on my Synology NAS and I’m trying to access my library using OPDS. This works on my koreader and other devices but on my X4 with crosspoint it only partially works. 

The initial connection works, but I can’t open many of the feeds. For example, I always get „Error: failed to fetch feed“ when trying to open „Recently added Books“. 
„Alphabetical Books“ opens fine at first but then only some letters actually display the list of books, others return the same error as above. 

I first thought it might have to do with the number of books, since it seems I can’t open any letter with more than 8 books, but then I have also hit the same error with a letter for which I have only two books. 

I am happy to provide more details if there is anything I can test. 

### Steps to Reproduce

1. Use „OPDS Browser“ to access CWA library. 
2. browse through library
3. hit error „Failed to fetch feed“

### Expected Behavior

Feeds should be opened as on other devices. 

```shell

```

### LINKED ISSUE #2018: X3 OTA fails with `ESP_ERR_NO_MEM` at `esp_https_ota_begin`
### Affected Version

v1.3.0

## Symptom

On X3 hardware, OTA from CrossPoint 1.3.0 → any newer release reliably fails
within ~50 ms of the firmware GET starting. The X4 succeeds with the same
firmware payload.

Device-side serial output during a failing X3 attempt:

```
[206855]   X3_OEM_TRIGGER=0x12
[207363] [DBG] [OTA] Checking for update (current: 1.3.0-dev-master-93e81daf)
[208047] [DBG] [OTA] Found update: tag=99.9.9 size=5804912
[208047] [DBG] [OTA] Firmware URL: https://unlocker.crosspointreader.com/firmware/firmware.bin
[208893] [DBG] [OTA] New update available, starting download...
[208911] [DBG] [OTA] Update progress: 0 / 5804912
E (216176) esp_https_ota: Couldn't allocate memory to upgrade data buffer
[210210] [DBG] [OTA] HTTP OTA Begin Failed: ESP_ERR_NO_MEM
[210210] [DBG] [OTA] Update failed: 5
[211924] [INF] [MEM] Free: 91448 bytes, Total: 228152 bytes, Min Free: 10084 bytes, MaxAlloc: 40948 bytes
```

`MaxAlloc: 40948` mbedTLS + esp_http_client + the OTA
upgrade buffer collectively need more than ~40 KB of contiguous heap during
`esp_https_ota_begin`, and the X3 simply doesn't have it at that moment.

X4 has more headroom (more total heap, less fragmentation by the time OTA
runs) and passes the same call without trouble.

## Root cause

`esp_https_ota_begin` calls into mbedTLS to set up the SSL context for the
HTTPS download. mbedTLS's default in/out buffers (16 KB each), plus the
HTTP client buffers (`buffer_size`/`buffer_size_tx` are 8192 in
`OtaUpdater.cpp`), plus the OTA upgrade buffer, plus whatever's already
allocated to the renderer / Wi-Fi stack / SD card cache, exceed the largest
free contiguous block on X3.

This is not unique to any specific OTA payload — the failure happens before
any firmware bytes are read.

## What I did on the unlocker side

Flipped the spoofed manifest's `browser_download_url` from `https://` to
`http://`:

```diff
- let download_url = "https://unlocker.crosspointreader.com/firmware/firmware.bin";
+ let download_url = "http://unlocker.crosspointreader.com/firmware/firmware.bin";
```

The unlocker already serves both HTTP (:80) and HTTPS (:443) on the bridge
interface. Routing the firmware over HTTP skips mbedTLS entirely and frees
the ~25–40 KB of contiguous heap that was making `esp_https_ota_begin` fail
on X3.

This depends on `CONFIG_OTA_ALLOW_HTTP=y`, which is the default in
arduino-esp32's bundled sdkconfig. If a future CrossPoint build flips it
off, OTA will refuse the http:// URL.

### Steps to Reproduce

Run the unlocker on an x3 device running 1.3.0

### Expected Behavior

It doesn't error out in seconds

```shell

```

itsthisjustin:
Another upside to this is that TLS 1.3 issues should be solved now for KoSync servers


src/network/HttpDownloader.cpp
```diff
@@ -38,6 +46,189 @@ bool isRedirect(int status) {
...
+      LOG_ERR("HTTP", "wolfSSL bad URL: %s", url.c_str());
+      return HttpDownloader::HTTP_ERROR;
+    }
+
+    WiFiClient plainClient;
+    freeink::SecureClient secureClient;
+    Client* client = nullptr;
+    if (parsed.scheme == "https") {
+      secureClient.setInsecure();
+      client = &secureClient;
```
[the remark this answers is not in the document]
itsthisjustin: We purposefully use http when we can to avoid memory pressure

src/activities/browser/OpdsBookBrowserActivity.cpp
```diff
@@ -273,19 +285,30 @@ void OpdsBookBrowserActivity::downloadBook(const OpdsEntry& book) {
...
+  int lastRenderedPercent = -1;
+  unsigned long lastProgressUpdateMs = 0;
   const auto result = HttpDownloader::downloadToFile(
       downloadUrl, filename,
-      [this](const size_t downloaded, const size_t total) {
+      [this, &lastRenderedPercent, &lastProgressUpdateMs](const size_t downloaded, const size_t total) {
         downloadProgress = downloaded;
         downloadTotal = total;
-        requestUpdate(true);
+        const int percent = total > 0 ? static_cast<int>(downloaded * 100 / total) : 0;
```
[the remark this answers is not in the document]
itsthisjustin: I'm not concerned about folks trying to download that large of a file on device

itsthisjustin:
@rhythmerc if you want to chime in on this PR just about your own experience using this package. Also I fixed one more thing in the SDK related to this if you want to personally bump to main

rhythmerc:
Happy to pitch in! Two-fold thoughts here:

1. The stock Arduino SSL stack is leaky/fraggy (as evidenced by the silent reboots we've got peppered across our wifi-related activities). WolfSSL is significantly better at restoring the heap, which allowed my fork to drop the silent reboots.

2. My fork has a lower RAM floor than stock (primarily on account of the SD card font system being extended for UI use). WolfSSL can operate with a slimmer maxAlloc pool than the Arduino SSL stack, which I needed to get downloader activities working again with the lower floor.

Haven't run into any significant issues since migrating over. It's a stable stack!

src/network/HttpDownloader.cpp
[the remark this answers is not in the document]
itsthisjustin: A correct fix requires WolfSSL to have a trust anchor, and the repo ships no certs

finlay-mcaree:
Just tested this branch (on an X3) and while it doesn't `abort()` any more, it still doesn't render large OPDS xml (I've got a 525kb one I've been testing with).

My own solve for this was writing the xml to a local temp file, then chunk reading from there, which _does_ work for the 525kb xml. I don't know how that approach would work with these changes (I've been working off master), but if you want to take a look it's [here](https://github.com/finlay-mcaree/crosspoint-reader-opds-oom/commit/f45d782aef91af202c7a600a520b346dea00b2b2).

Here's the error from this branch for the large xml:
```
E (116781) esp-aes: Failed to allocate memory
E (116781) esp-tls-mbedtls: read error :-0x0001
E (116782) transport_base: esp_tls_conn_read error, errno=Success
E (116811) esp-aes: Failed to allocate memory
E (116812) esp-tls-mbedtls: read error :-0x0001
E (116812) transport_base: esp_tls_conn_read error, errno=Success
[09:09:23] [ERR] [HTTP] read error after 163196 bytes
```

itsthisjustin:
> Just tested this branch (on an X3) and while it doesn't `abort()` any more, it still doesn't render large OPDS xml (I've got a 525kb one I've been testing with).
> 
> My own solve for this was writing the xml to a local temp file, then chunk reading from there, which _does_ work for the 525kb xml. I don't know how that approach would work with these changes (I've been working off master), but if you want to take a look it's [here](https://github.com/finlay-mcaree/crosspoint-reader-opds-oom/commit/f45d782aef91af202c7a600a520b346dea00b2b2).
> 
> Here's the error from this branch for the large xml:
> ```
> E (116781) esp-aes: Failed to allocate memory
> E (116781) esp-tls-mbedtls: read error :-0x0001
> E (116782) transport_base: esp_tls_conn_read error, errno=Success
> E (116811) esp-aes: Failed to allocate memory
> E (116812) esp-tls-mbedtls: read error :-0x0001
> E (116812) transport_base: esp_tls_conn_read error, errno=Success
> [09:09:23] [ERR] [HTTP] read error after 163196 bytes
> ```

Any way I can test this easily? I was testing with mayberry which has thousands of books but I do limit each page to 50 at a time

finlay-mcaree:
Hmm, so I put a dummy file [on github.io](https://finlay-mcaree.github.io/opds-test.xml) to replicate the behaviour (the original file is self-hosted on my LAN). 

Strangely, I can load that XML using this branch, if extremely slowly (sometimes minutes), but when I host on my own LAN (nginx 1.31.2 w/lets encrypt), I get the read error I mentioned above. Exactly the same file, but one host works and one does not. Changing the LAN file to <100kb seemed to allow it to load.

itsthisjustin:
> Hmm, so I put a dummy file [on github.io](https://finlay-mcaree.github.io/opds-test.xml) to replicate the behaviour (the original file is self-hosted on my LAN).
> 
> Strangely, I can load that XML using this branch, if extremely slowly (sometimes minutes), but when I host on my own LAN (nginx 1.31.2 w/lets encrypt), I get the read error I mentioned above. Exactly the same file, but one host works and one does not. Changing the LAN file to <100kb seemed to allow it to load.

Ok thanks. I'll look into this

itsthisjustin:
@finlay-mcaree give this a shot now after the updates I made. I also moved kosync and our OTA updater to the new path and it's a night and day difference

finlay-mcaree:
Yeah it's significantly faster for the GitHub.io link, thanks!

My self hosted (LE-secured) domains are failing ssl connects now though, not even getting to the xml load. I'll try to investigate more later.

itsthisjustin:
> Yeah it's significantly faster for the GitHub.io link, thanks!
> 
> My self hosted (LE-secured) domains are failing ssl connects now though, not even getting to the xml load. I'll try to investigate more later. 

Any chance you can email me access to that directly? I'd love to fix but hard to debug this way

itsthisjustin:
@finlay-mcaree give it a shot now

finlay-mcaree:
Synced and tried latest, but same errors - sorry I couldn't post this morning, was just checking quickly.

The two domains have different errors from wolfssl, but they're both LE certs (I added some extra logging around it)

```
[SecureClient] wolfSSL_connect failed: -112 (mp_exptmod error state)
[SecureClient]   host=host1 port=443
[SecureClient]   last send alert: level=2 desc=51
[SecureClient]   last recv alert: level=-1 desc=-1
[SecureClient]   free heap=67560 bytes
[16:22:48] [ERR] [HTTP] wolfSSL connect failed: host1:443

[SecureClient] wolfSSL_connect failed: -313 (received alert fatal error)
[SecureClient]   host=host2 port=443
[SecureClient]   last send alert: level=-1 desc=-1
[SecureClient]   last recv alert: level=2 desc=40
[SecureClient]   free heap=73008 bytes
[16:23:04] [ERR] [HTTP] wolfSSL connect failed: host2:443
```

-112 [might need FP_MAX_BITS set](https://github.com/wolfSSL/wolfssl/issues/2924#issuecomment-620030245)
-313 [might be](https://github.com/wolfSSL/wolfssl/tree/9e71da21ac417299d6e16b67a15e130aab855383/wolfcrypt/src/port/Espressif/esp_crt_bundle#tls-13-issues-with-howsmysslcom) because the server is only TLS 1.2, not TLS 1.3 - my guess would be if I upgraded the cert I would hit the -112 error instead.

I don't have a mechanism to give you access to either of the servers, apologies.

itsthisjustin:
> Synced and tried latest, but same errors - sorry I couldn't post this morning, was just checking quickly.
> 
> The two domains have different errors from wolfssl, but they're both LE certs (I added some extra logging around it)
> 
> ```
> [SecureClient] wolfSSL_connect failed: -112 (mp_exptmod error state)
> [SecureClient]   host=host1 port=443
> [SecureClient]   last send alert: level=2 desc=51
> [SecureClient]   last recv alert: level=-1 desc=-1
> [SecureClient]   free heap=67560 bytes
> [16:22:48] [ERR] [HTTP] wolfSSL connect failed: host1:443
> 
> [SecureClient] wolfSSL_connect failed: -313 (received alert fatal error)
> [SecureClient]   host=host2 port=443
> [SecureClient]   last send alert: level=-1 desc=-1
> [SecureClient]   last recv alert: level=2 desc=40
> [SecureClient]   free heap=73008 bytes
> [16:23:04] [ERR] [HTTP] wolfSSL connect failed: host2:443
> ```
> 
> -112 [might need FP_MAX_BITS set](https://github.com/wolfSSL/wolfssl/issues/2924#issuecomment-620030245) -313 [might be](https://github.com/wolfSSL/wolfssl/tree/9e71da21ac417299d6e16b67a15e130aab855383/wolfcrypt/src/port/Espressif/esp_crt_bundle#tls-13-issues-with-howsmysslcom) because the server is only TLS 1.2, not TLS 1.3 - my guess would be if I upgraded the cert I would hit the -112 error instead.
> 
> I don't have a mechanism to give you access to either of the servers, apologies.

Thanks for the log. Helpful. I already allowed it to downgrade to 1.2 so I'm suprised by that but looking now...

itsthisjustin:
ok @finlay-mcaree try again. Thanks for working with me on this

itsthisjustin:
@rhythmerc you'll probably want to pull this into folio as it SHOULD fix tls 1.2 issues

finlay-mcaree:
Yesss that fixed the -112/mp_exptmod error.

The other server is still giving the -313 error however - let me know if there's anything else I can do to debug (cert info etc).

itsthisjustin:
@finlay-mcaree tried another thing here to gracefully fall back if first ack fails

finlay-mcaree:
It works!

```
[SecureClient] wolfSSL_connect failed (auto): -313
[SecureClient] retrying with TLS 1.2-only handshake
[19:46:10] [DBG] [OPDS] Feed truncated to fit memory
```

Thanks so much!

itsthisjustin:
> It works!
> 
> ```
> [SecureClient] wolfSSL_connect failed (auto): -313
> [SecureClient] retrying with TLS 1.2-only handshake
> [19:46:10] [DBG] [OPDS] Feed truncated to fit memory
> ```
> 
> Thanks so much!

Heck yeah!

src/network/HttpDownloader.cpp
```diff
@@ -7,24 +7,38 @@
...
+#include <cstdlib>
 #include <cstring>
 #include <functional>
 #include <string>
 
+#if defined(FREEINK_NET_WOLFSSL)
+#include <SecureClient.h>
+#include <WiFiClient.h>
+
+extern "C" void wolfSSL_Arduino_Serial_Print(const char* const) {}
```
Uri-Tauber: This stub silently discards all wolfSSL diagnostics. would be useful to route them through logging.

```suggestion
extern "C" void wolfSSL_Arduino_Serial_Print(const char* const msg) { LOG_DBG("WOLFSSL", "%s", msg); }
```

src/network/HttpDownloader.cpp
```diff
@@ -38,6 +52,292 @@ bool isRedirect(int status) {
...
+    // Strip any chunk extension (";name=value") before parsing the hex size.
+    const size_t ext = line.find(';');
+    const unsigned long chunkSize =
+        strtoul((ext == std::string::npos ? line : line.substr(0, ext)).c_str(), nullptr, 16);
+    if (chunkSize == 0) {
+      // Last chunk: consume optional trailers up to the terminating blank line.
+      while (readLine(client, line, millis() + HTTP_TIMEOUT_MS) && !line.empty()) {
+      }
+      return HttpDownloader::OK;
+    }
```
Uri-Tauber: Two issues Claude flagged in the chunk-size path, both per-chunk:

1. `std::string line` is constructed every iteration and `line.substr(0, ext)` allocates per chunk. Hoisting `line` out lets `clear()` retain capacity, and `strtoul` already stops at the first non-hex char, so the `substr` is unnecessary.
2. A malformed size line makes `strtoul` return 0, which is indistinguishable from the terminal chunk — a corrupted stream returns a truncated body as `OK`.

```suggestion
  std::string line;
  line.reserve(64);
  for (;;) {
    if (sink.cancelFlag && *sink.cancelFlag) return HttpDownloader::ABORTED;
    if (!readLine(client, line, millis() + HTTP_TIMEOUT_MS)) {
      LOG_ERR("HTTP", "wolfSSL chunk size read failed after %zu bytes", sink.downloaded);
      return HttpDownloader::HTTP_ERROR;
    }
    // strtoul stops at the first non-hex char, so any ";ext" suffix is ignored.
    char* end = nullptr;
    const unsigned long chunkSize = strtoul(line.c_str(), &end, 16);
    if (end == line.c_str()) {
      LOG_ERR("HTTP", "wolfSSL malformed chunk size");
      return HttpDownloader::HTTP_ERROR;
    }
    if (chunkSize == 0) {
      // Last chunk: consume optional trailers up to the terminating blank line.
      while (readLine(client, line, millis() + HTTP_TIMEOUT_MS) && !line.empty()) {
      }
      return HttpDownloader::OK;
    }
```

src/network/HttpDownloader.cpp
```diff
@@ -38,6 +52,292 @@ bool isRedirect(int status) {
...
+    if (!parseUrl(url, parsed)) {
+      LOG_ERR("HTTP", "wolfSSL bad URL: %s", url.c_str());
+      return HttpDownloader::HTTP_ERROR;
+    }
+
+    WiFiClient plainClient;
+    freeink::SecureClient secureClient;
+    Client* client = nullptr;
+    if (parsed.scheme == "https") {
+      secureClient.setInsecure();
```
Uri-Tauber: Are we disabling TLS verification on purpose?

src/network/HttpDownloader.cpp
```diff
@@ -38,6 +52,292 @@ bool isRedirect(int status) {
...
+      std::transform(name.begin(), name.end(), name.begin(),
+                     [](unsigned char c) { return static_cast<char>(tolower(c)); });
+      if (name == "content-length") contentLength = static_cast<size_t>(strtoul(value.c_str(), nullptr, 10));
+      if (name == "location") location = value;
+      if (name == "transfer-encoding") {
+        while (!value.empty() && isspace(static_cast<unsigned char>(value.back()))) value.pop_back();
+        std::transform(value.begin(), value.end(), value.begin(),
+                       [](unsigned char c) { return static_cast<char>(tolower(c)); });
+        transferEncoding = value;
+      }
```
Uri-Tauber: Easy to make allocation-free — lowercase the name in place and read the value out of the same buffer:

```suggestion
    while (readLine(*client, line, headerDeadline)) {
      if (line.empty()) break;
      const size_t colon = line.find(':');
      if (colon == std::string::npos) continue;
      // Lowercase the name in place and read the value from the same buffer —
      // avoids two substr() heap allocations per header line.
      std::transform(line.begin(), line.begin() + colon, line.begin(),
                     [](unsigned char c) { return static_cast<char>(tolower(c)); });
      const char* value = line.c_str() + colon + 1;
      while (*value == ' ') ++value;
      if (line.compare(0, colon, "content-length") == 0) {
        contentLength = static_cast<size_t>(strtoul(value, nullptr, 10));
      } else if (line.compare(0, colon, "location") == 0) {
        location = value;
      } else if (line.compare(0, colon, "transfer-encoding") == 0) {
        transferEncoding = value;
        while (!transferEncoding.empty() && isspace(static_cast<unsigned char>(transferEncoding.back())))
          transferEncoding.pop_back();
        std::transform(transferEncoding.begin(), transferEncoding.end(), transferEncoding.begin(),
                       [](unsigned char c) { return static_cast<char>(tolower(c)); });
      }
```

src/network/HttpDownloader.cpp
```diff
@@ -7,24 +7,38 @@
...
-// survive. Smaller keeps contiguous heap free while WiFi and TLS are up. TX
-// only carries our GET; the body streams in READ_CHUNK pieces.
-constexpr int HTTP_RX_BUF = 4096;
-constexpr int HTTP_TX_BUF = 1024;
+#if !defined(FREEINK_NET_WOLFSSL)
+// RX holds the response headers. Smaller buffers leave enough contiguous heap
+// for mbedTLS on redirect-heavy OPDS feeds while still preserving the headers
+// we read directly (Location, Content-Length).
+constexpr int HTTP_RX_BUF = 2048;
+constexpr int HTTP_TX_BUF = 512;
```
Uri-Tauber, line 26: Have you verified OPDS servers still works with the trancuated buffer?

src/network/HttpDownloader.cpp
itsthisjustin: Yes because we don't have certs in the firmware and this lets us connect without worrying about parsing it all and comparing

src/network/HttpDownloader.cpp
itsthisjustin: Yes. See testing comments above from me and another user. Mayberry also works very well.

src/network/OtaUpdater.cpp
```diff
@@ -116,72 +109,67 @@ OtaUpdater::OtaUpdaterError OtaUpdater::installUpdate(ProgressCallback onProgres
...
   esp_wifi_set_ps(WIFI_PS_MIN_MODEM);
 
-  if (esp_err != ESP_OK) {
-    LOG_ERR("OTA", "esp_https_ota_perform Failed: %s", esp_err_to_name(esp_err));
-    esp_https_ota_finish(ota_handle);
+  if (!fetchOk || !flashOk) {
+    LOG_ERR("OTA", "Firmware download failed (flashOk=%d)", flashOk ? 1 : 0);
+    esp_ota_abort(otaHandle);
     return HTTP_ERROR;
   }
```
Uri-Tauber, line 162: `esp_ota_write` failure surfaces as `HTTP_ERROR`, so a flash fault gets shown/handled as a network error. Distinguish them:

```suggestion
  if (!fetchOk || !flashOk) {
    LOG_ERR("OTA", "Firmware install failed (%s)", flashOk ? "download" : "flash write");
    esp_ota_abort(otaHandle);
    return flashOk ? HTTP_ERROR : INTERNAL_UPDATE_ERROR;
  }
```

src/activities/browser/OpdsBookBrowserActivity.cpp
```diff
@@ -216,21 +219,28 @@ void OpdsBookBrowserActivity::fetchFeed(const std::string& path) {
...
 
+  entries.reserve(entries.size() + (prevUrl.empty() ? 0 : 1) + (nextUrl.empty() ? 0 : 1));
   if (!prevUrl.empty()) {
     entries.insert(entries.begin(), OpdsEntry{OpdsEntryType::NAVIGATION, tr(STR_PREV_PAGE), "", prevUrl, ""});
   }
   if (!nextUrl.empty()) {
     entries.push_back(OpdsEntry{OpdsEntryType::NAVIGATION, tr(STR_NEXT_PAGE), "", nextUrl, ""});
   }
+  if (feedTruncated) {
+    LOG_DBG("OPDS", "Feed truncated to fit memory");
```
Uri-Tauber: Change to LOG_INF, so it would be visible in GH release:
```suggestion
    LOG_INF("OPDS", "Feed truncated to fit memory");
```

Also, might be a good idea to add UI popup.

Uri-Tauber:
Conflict...

Uri-Tauber:
I noticed a lot of the new additions in `HttpDownloader.cpp` are already implemented in `freeink-sdk/libs/network/SecureNet/include/SecureHttpClient.h`.
For example, `parseUrl` (lines 63–81 in `HttpDownloader.cpp`) is almost exact copy of `parseUrl` (lines 159–178 in `SecureHttpClient.h`)

Other suspects for deduplication: `formatAuthority`, `readLine`, `readBodyExact`, `readChunkedBody`. I probably missed others.

It's not a blocker, but would be a good idea to reconstruct around `SecureHttpClient.h` and save a lot of deduplication work in the future.

EDIT: It will require some work on `SecureHttpClient.h`, so it can be used to stream chunks. Maybe extract shared helper?

src/network/HttpDownloader.cpp
```diff
@@ -7,24 +7,38 @@
 #include <esp_crt_bundle.h>
 #include <esp_http_client.h>
```
Uri-Tauber: I think those two can be safely moved into `#if !defined(FREEINK_NET_WOLFSSL)` to save compile time.

xlab:
As discussed in https://github.com/crosspoint-reader/crosspoint-reader/pull/2561 I will benchmark this pr on large downloads from OPDS server. Note that in my tests switching between HTTPS and HTTP didn't make great difference, the RTT ping to the server combined with receive window was the bounding factor.

That being said, I doubt that this PR can cross 300 KiB/s line while using a remote server, not something from local network. But we'll see.

itsthisjustin:
> I noticed a lot of the new additions in `HttpDownloader.cpp` are already implemented in `freeink-sdk/libs/network/SecureNet/include/SecureHttpClient.h`. For example, `parseUrl` (lines 63–81 in `HttpDownloader.cpp`) is almost exact copy of `parseUrl` (lines 159–178 in `SecureHttpClient.h`)
> 
> Other suspects for deduplication: `formatAuthority`, `readLine`, `readBodyExact`, `readChunkedBody`. I probably missed others.
> 
> It's not a blocker, but would be a good idea to reconstruct around `SecureHttpClient.h` and save a lot of deduplication work in the future.
> 
> EDIT: It will require some work on `SecureHttpClient.h`, so it can be used to stream chunks. Maybe extract shared helper?

Did this.  Good call out

Uri-Tauber (approved):
Awesome. Approving now.
Before we merge, I would like you to re-test it on real hardware - just a sanity check nothing got broken during the refactor.

Ah. cppcheck complaint.

itsthisjustin:
> Awesome. Approving now. Before we merge, I would like you to re-test it on real hardware - just a sanity check nothing got broken during the refactor.
> 
> Ah. cppcheck complaint.

Yep tested downloading files and navigating in and out of large OPDS folders. All working. Pushed one more fix that increases chunk streaming size for larger files. All working

Uri-Tauber:
Great! 
Can you solve the cppcheck error?

PR #2477, opened 2026-06-29
feat: dictionary support

rmmh:
## Summary

* **What is the goal of this PR?** Implements on-device Stardict dictionary support (as discussed in #1599)

This adds a core engine that can load StarDict dictionaries, preprocess them for on-device usage (decompressing the definitions and adding additional indices), and supports looking up words in a chain, word history, multiple dictionaries with per-book options, and more.

An example of an open dictionary page, looking up another word:
image

## Additional Context

ROM: +138.46 KB (Code: +75.55 KB, Data: +62.91 KB) | DRAM: +328 B (BSS: +272 B, Data: +56 B)

This is a continuation of #1402, but reshaped with these specific changes:
- rebased onto master (~1.4.1) and squashed commits into subsystem-based ones that each compile and pass tests, while attempting to maintain accurate authorship metadata. Hopefully this makes review much easier. (@WuTofu -- yours are not currently split out, I am happy to split out your changes if you prefer)
- does not require decompressing the entire dictionary on-device, saving many minutes of time during preparation.
- uses a much more efficient .idx indexing structure
- extended IPA font to support more common & useful glyphs for dictionaries, like greek letters used in etymologies, and a variety of additional symbols (specifically covering GCIDE), and renamed it to DICT_FONT for accuracy.
- added option for short power button press -> look up word
- refactored the tests to use CMake/Googletest consistently
- removed large checked in binaries, including all the synthetic dictionary data (~30MB)-- these are regenerated on-demand during tests
- refactored and optimized miscellaneous code
- tweaked word select nav logic to try to maintain x-position when traversing vertically

- probably other things I forgot

This mostly works (tested on-device), but still has rough edges and OOMs with unfortunate frequency with my test dictionary ([GNU Collaborative International Dictionary of English](http://build.koreader.rocks/download/dict/gcide.tar.gz) / https://gcide.gnu.org.ua/ ). I'm not sure how solid it needs to be to be merged. Perhaps some sort of abort handler that pops the activity stack on OOM (and resets the device?) would be sufficient.

Open questions:
- ~~should we support both /.dictionaries/ and /dictionaries/ for dictionary placements?~~
~~- is it worth having the ability to override dictionary font size, given that we only ship the dictionary font at one point size?~~
- is it worth having separate dictionary font family+font size settings overrides at all?
- should we have the ability to browse headwords directly (supporting to the previous/next word as well as up/down)?
- is there any legitimate reason to have the current dictionary in dictionary.bin (a raw bit of text) and not the settings.json?

TODO:
- ~~validate that we actually need to fully decompress .dict.gz -- this is by far the slowest part of the dictionary prep~~
- add headword browsing
- ~~remove /.dictionaries/, standardize on /dictionaries/, and~~ make the file browser opening a /dictionary/foo jump to the dict headword browser

---

### AI Usage


WarKittens:
> is it worth having the ability to override dictionary font size, given that we only ship the dictionary font at one point size?

Considering the size of the device and that you have people like me that read without glasses, even though we need them which leads down the path to people with vision disabilities, **_yes you should let this font size be overridden_**. If there's ever a question about something that could accommodate someone with a disability, and it could be included without much effort, you should include it.

WarKittens:
Question: I am guessing custom dictionaries will have to be under a certain size, e.g. my OED .mdd that is 135 MB will be too much of a burden for the device?

rmmh:
This should ideally work with dictionaries of any size, but it's strongly recommended that you decompress the .dict.gz files on a separate device-- SD card reads+writes to decompress the entire dictionary are *extremely* slow.

rmmh:
This branch ends up duplicating a huge amount of formatted text rendering logic, which makes it much harder to maintain parity (justification, focus words, etc). Once #2456 is merged (?), I think it's worth doing the minimal fixes at the beginning to extend the text renderer-- the only real change is being able to have parsed text with multiple fonts, which will be useful *anyways* for supporting more symbols in more books, combining fallback fonts between flash/SD, and potentially even more.

It just makes sense to have a single path that handles rendering styled text. Both the "use strip grayscale" and the "add focus reading" would have been free if this was already reusing the epub rendering.

rmmh:
Updates:
- replaced OFT/CPST index files with a much better index that requires around 1/4 as many block reads https://gist.github.com/rmmh/2c3d3d4382e9afadf3ef0c9ddcebf434
- dropped on-device dict decompression in favor of supporting lazy entry decompression

Uri-Tauber:
Wow. 1.2 million LoC! even github having troubles with viewing this PR.
You probably accidentally committed a lot of temp files... :wink:

rmmh:
Sorry about that mess. Was doing some rebase surgery to keep the commit history tidy and messed up.

rxmmah:
i have noticed the previews PR was faster in launching the word-select overlay.

itsthisjustin:
@Uri-Tauber should this be closed now?

Uri-Tauber:
@itsthisjustin It wont be merged, if that the question.
But there still number of dictionary improvements we can take from here.
I have this on my todo list, but lately I'm just reviewing PRs. I cant remember when was the last time I wrote a new feature myself...

Here some references Claude wrote for me, if someone interested. Its a bit outdated.  
[DICTIONARY_IMPORT_GUIDE.md](https://github.com/user-attachments/files/31547748/DICTIONARY_IMPORT_GUIDE.md)

PR #2481, opened 2026-06-29
feat: Add touch coordinate mapping and RTOS task yielding

itsthisjustin:
- Added touch navigation across main menus, settings, OPDS, file transfer, reader menus/subscreens, dialogs, keyboard, and passive/result screens.

 - Added touch-down highlighting, tap activation, swipe scrolling, and long-press handling

 - Added reader page turns as tap zones using FreeInk SDK tap-zones

 - Hid button hints and front-button-only settings on touch devices.

- Swaps keyboard out for freeinkui keyboard which has built in touch capability and is backwards compatible with button driven keyboard

 - Gated touch-only settings behind touch capability, and hid non-touch/device-incompatible settings like OTA on touch for now.

 - Added touch-compatible option popups and dismiss-on-outside- tap behavior.

 - Added direct touch controls for UTC offset with - / + buttons.

 - Added OPDS touch search with SDK-generated search icon.

 - Added RTOS task yielding around long-running work to avoid starving idle/watchdog tasks
 
 
 This new branch does NOT add any ui scaling or new icon system like the prior branch did. The ui scales just like it normally would without being aware of if the touch device is larger or taller


lib/hal/HalClock.cpp
```diff
@@ -18,33 +18,10 @@ static uint8_t bcdToDec(uint8_t bcd) { return ((bcd >> 4) * 10) + (bcd & 0x0F);
...
-  _available = true;
-  LOG_INF("CLK", "DS3231 RTC found");
-
-  // Prime the cache with an initial read
-  uint8_t h, m;
-  getTime(h, m);
+  _usesSdkRtc = false;
+  _available = _sdkRtc.begin();
+  _usesSdkRtc = _available;
+  LOG_INF("CLK", _available ? "SDK RTC found" : "RTC not found");
```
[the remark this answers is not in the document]
itsthisjustin: This is fixed now

itsthisjustin:
@Uri-Tauber tried to remove as much manual hit touch code in all the activities by building a re-usable touch mapper. I also swapped the keyboard out for a freeinkui one which yes, I know you'd want in a separate PR. However I'll argue for this branch having all the touch stuff in it's own PR makes it testable while yes maybe harder to review. Functionality is super important here and we need to make sure everything runs on the default env for x4/x3 without any regressions and everything plays nice on the new touch devices as well. The upgrade for the keyboard also gets us a few nicer things such as real UTF-8 keys so say a french user can type é in a search or password. We also get more symbol coverage and built in touch instead of manually rolling touch for each key.

lib/hal/HalTiltSensor.cpp
```diff
@@ -52,43 +61,31 @@ bool HalTiltSensor::readGyro(float& gx, float& gy, float& gz) const {
...
+    _lastPollMs = millis();
+    LOG_INF("GYR", "SDK IMU initialized");
     return;
   }
-
-  _available = true;
-  _initMs = millis();
-  _lastPollMs = millis();
-  LOG_INF("GYR", "QMI8658 gyro initialized and put to sleep");
+  LOG_ERR("GYR", "SDK IMU not found");
```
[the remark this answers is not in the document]
itsthisjustin: - HalTiltSensor has no _backend member and no QMI8658 code at all (lib/hal/HalTiltSensor.h:18-42). The class holds a single Imu _sdkImu and begin() sets _available from its one probe (HalTiltSensor.cpp:16-25).
- That's deliberate: commit d625b4dd ("Use SDK RTC and IMU backends") deleted the firmware-side QMI8658 backend (~267 lines removed across HalClock/HalTiltSensor) because the support moved into the SDK.
- X3 tilt still works through that single path: the SDK Imu driver natively handles the QMI8658 — the X3 board profile declares ImuType::Qmi8658 at address 0x6B (BoardConfig.h:627), and Imu::begin()/Imu::read() switch on that type with full QMI8658 init and register reads (freeink-sdk .../Imu.cpp:111-117, 135-137).

So there is no unreachable fallback to rewire; adding a second QMI8658 backend inside HalTiltSensor would duplicate the SDK driver and reintroduce exactly the code that commit removed. The one legitimate failure mode left (_sdkImu.begin() returning false on an X3 with a healthy IMU) would indicate a board-profile or I2C problem, which is already surfaced by the LOG_ERR("GYR", "SDK IMU not found") at HalTiltSensor.cpp:24.

lib/I18n/translations/hebrew.yaml
```diff
@@ -73,6 +73,7 @@ STR_IMAGES_SUPPRESS: "הסתר תמונות"
 STR_SHORT_PWR_BTN: "לחיצה קצרה על כפתור ההפעלה"
 STR_ORIENTATION: "כיוון קריאה (מסך)"
 STR_SIDE_BTN_LAYOUT: "פריסת כפתורי צד (בקריאה)"
+STR_TOUCH_READER_CONTROLS: "פקדי מגע לקורא"
```
Uri-Tauber:
```suggestion
STR_TOUCH_READER_CONTROLS: "שליטה במגע (קורא)"
```

src/activities/util/KeyboardEntryActivity.h
```diff
@@ -62,165 +73,48 @@ class KeyboardEntryActivity : public Activity {
...
-  static constexpr const char* const urlSnippets[URL_SNIPPET_COUNT] = {
-      "https://", "www.", ".com", "http://", "192.168.", ".org", "/opds", ":8080", ".net"};
+  // Tap/hold routing (threshold long-press, release swallow, slide re-arm)
+  // lives in the SDK; loop() feeds it the level-triggered touch state.
+  freeink::ui::TouchHoldRouter touchRouter;
+
+  // loop() runs on the main task while render() rebuilds the interaction
+  // table on the render task; routing against a half-built table would read
+  // torn entries, so taps are dropped during the rebuild window.
+  volatile bool interactionsReady = false;
```
Uri-Tauber: `volatile bool` doesn't provide inter-core memory ordering.
Fine for the single-core c3, could be a problem with the dial-core s3.
Maybe use `std::atomic<bool>`?

lib/hal/HalTiltSensor.cpp
```diff
@@ -4,127 +4,47 @@
...
-    LOG_INF("GYR", "QMI8658 entered sleep mode");
-    return true;
-  } else {
-    LOG_ERR("GYR", "Failed to put QMI8658 to sleep");
-    return false;
-  }
+  clearPendingEvents();
+  _inTilt = false;
+  _isAwake = false;
+  return true;
```
Uri-Tauber, line 62: The old implementation wrote `CTRL7_DISABLE_ALL` to put the QMI8658 into hardware standby. The new `deepSleep()`/`wake()` are software-only — they just flip `_isAwake`.
If I'm correct (I hope I didnt missed anything here), IMU samples continuously from boot, even with tilt page-turn disabled. Massive power drain.

src/activities/reader/EpubReaderActivity.cpp
```diff
@@ -1396,50 +1406,66 @@ void EpubReaderActivity::renderContents(std::unique_ptr<Page> page, const int or
...
+
+    // Tiered on heap pressure: two plane buffers hide both plane renders
+    // inside the refresh wait; one hides the LSB render (its buffer is reused
+    // for MSB after streaming); none falls back to the strip-scratch flow with
+    // no overlap. The MSB buffer is only attempted when it leaves ~60 KB free
+    // so the pass never starves concurrent allocations. Blocking panels skip
+    // the buffers entirely (nothing to overlap).
+    auto lsbPlaneBuf = overlapRefresh ? makeUniqueNoThrow<uint8_t[]>(planeBytes) : nullptr;
+    auto msbPlaneBuf =
+        (lsbPlaneBuf && ESP.getFreeHeap() >= planeBytes + 60000) ? makeUniqueNoThrow<uint8_t[]>(planeBytes) : nullptr;
```
Uri-Tauber, line 1491: This allocates (and frees) a 48 KB whole-plane buffer — up to 96 KB with the MSB tier — on **every page turn**.
Maybe allocate the plane buffer(s) once in `onEnter()` and reuse them across page turns?

Also, `ESP.getFreeHeap() checks *total* free heap, not the largest contiguous block.

src/components/OptionPopup.h
```diff
@@ -86,7 +121,90 @@ class OptionPopup {
 
   bool isActive() const { return active; }
 
+  std::vector<Rect> getOptionRects(const GfxRenderer& renderer) const {
```
Uri-Tauber: `getOptionRects` and `getDialogRect` are almost byte-to-byte copies. The first add ~10 lines the second dont have. And those calculations are very similar to what themes' virtual `drawOptionPopup()` computes independently. Should be refactored to one `computeLayout()` returning dialog rect + option rects.

src/components/OptionPopup.h
```diff
@@ -48,6 +50,39 @@ class OptionPopup {
     if (!active) return false;
 
     const int count = static_cast<int>(ownedStrings.size());
+    int tx = 0;
+    int ty = 0;
+    if (input.wasScreenTouchDown(tx, ty)) {
```
Uri-Tauber, line 58: `wasScreenTouchDown()` is level-triggered, so `handleInput()` heap-allocates a `std::vector<Rect>` and re-measures the title + every option string on every single iteration of the activity's `loop()`, not once per touch. 
Compute the rects once on first use and cache them.

src/components/OptionPopup.h
```diff
@@ -48,6 +50,39 @@ class OptionPopup {
...
+        if (contains(rects[i], tx, ty)) {
+          if (selectedIndex != i) {
+            selectedIndex = i;
+            requestUpdate();
+          }
+          return true;
+        }
+      }
+      if (contains(getDialogRect(input.getRenderer()), tx, ty)) return true;
+      return true;
```
Uri-Tauber: Dead condition.

src/components/UITheme.cpp
```diff
@@ -50,9 +51,17 @@ void UITheme::setTheme(CrossPointSettings::UI_THEME type) {
   }
 }
 
+ThemeMetrics UITheme::getMetrics() const {
+  ThemeMetrics metrics = *currentMetrics;
+  if (gpio.hasTouch()) {
+    metrics.buttonHintsHeight = 0;
+  }
+  return metrics;
```
Uri-Tauber: Any reason why return by value and not by reference?

src/activities/util/KeyboardEntryActivity.cpp
```diff
@@ -4,192 +4,512 @@
...
+const fui::KeyboardKey URL_SNIP_ROW1[] = {UK("https://", "https://", 2001), UK("www.", "www.", 2002),
+                                          UK(".com", ".com", 2003)};
+const fui::KeyboardKey URL_SNIP_ROW2[] = {UK("http://", "http://", 2004), UK("192.168.", "192.168.", 2005),
+                                          UK(".org", ".org", 2006)};
+const fui::KeyboardKey URL_SNIP_ROW3[] = {UK("/opds", "/opds", 2007), UK(":8080", ":8080", 2008),
+                                          UK(".net", ".net", 2009)};
+const fui::KeyboardKey URL_SNIP_BOTTOM[] = {UKS("abc", fui::KeyKind::Mode, fui::QWERTY_KEY_MODE, 2),
+                                            UKW("URL", nullptr, URL_PANEL_VALUE, 2),
+                                            UKS("Del", fui::KeyKind::Delete, fui::QWERTY_KEY_BACKSPACE, 2),
+                                            UKS("OK", fui::KeyKind::Ok, fui::QWERTY_KEY_ENTER, 2)};
```
Uri-Tauber, line 94: the legacy keyboard rendered OK via `tr(STR_OK_BUTTON)`.

src/activities/util/KeyboardEntryActivity.cpp
```diff
@@ -4,192 +4,512 @@
...
+  int lineStartIdx = 0;
+  int lineEndIdx = static_cast<int>(displayText.length());
+  int lineY = inputStartY;
+  int lastLineStartIdx = 0;
+  int lastLineEndIdx = lineEndIdx;
+  int lastLineStartX = effectiveMargin;
+  int lastLineWidth = 0;
+
+  while (true) {
+    std::string lineText = displayText.substr(lineStartIdx, lineEndIdx - lineStartIdx);
```
Uri-Tauber: If I'm not mistaken, This runs on **every screen tap**. Typing `maxLength` on a URL costs hundreds of short-lived allocations + redundant full-prefix measurements per keystroke.

Two cheap fixes: bail out early when the tap isn't inside the text-field rect (the common case — key taps), and measure incrementally with per-glyph advances over a `string_view`/fixed buffer instead of substr temporaries

Uri-Tauber:
in src/components/themes/BaseTheme.h, you left remains of the previous keyboard:  `drawKeyboardKey`, `KeyboardKeyType` enum, or the keyboard metrics fields (`keyboardKeyWidth`, `keyboardBottomKeyHeight`, `keyboardBottomKeySpacing`, `keyboardBottomAligned`, `keyboardFillUnselected`, `keyboardOutlineAllUnselected`, `keyboardKeyCornerRadius`). ~150 lines of per-theme rendering that no longer affect the screen.

Also, the keyboard now bypasses UITheme entirely, so RoundedRaff's key styling is silently ignored. I'm not sure if it's what we want...

src/MappedInputManager.cpp
```diff
@@ -74,17 +78,238 @@ bool MappedInputManager::mapButton(const Button button, bool (HalGPIO::*fn)(uint
...
+  float nxe = 0.0f;
+  float nye = 0.0f;
+  if (!gpio.wasSwipe(nxs, nys, nxe, nye)) return SwipeDir::None;
+
+  int sx = 0;
+  int sy = 0;
+  int ex = 0;
+  int ey = 0;
+  renderer.tapToLogical(nxs, nys, sx, sy);
+  renderer.tapToLogical(nxe, nye, ex, ey);
```
Uri-Tauber: Tha same ~10 lines of swipe decode is copy-pasted four times in this file. Extract to shared helper.

src/activities/settings/LanguageSelectActivity.cpp
```diff
@@ -27,17 +27,48 @@ void LanguageSelectActivity::onEnter() {
...
   }
 
   if (mappedInput.wasPressed(MappedInputManager::Button::Confirm)) {
-    handleSelection();
+    activateSelected();
+    return;
+  }
+
+  const auto& metrics = UITheme::getInstance().getMetrics();
+  const int contentTop = metrics.topPadding + metrics.headerHeight + metrics.verticalSpacing;
```
Uri-Tauber, line 43: This contentTop/contentHeight derivation + touchdown-select / tap-activate / swipe page-turn block is copy-pasted with slight variations into ~10 list activities. Maybe more.

Uri-Tauber:
One last thing: How much the new keyboard is scalable? 
If I will want tomorrow to add Hebrew layout, do I need to add it in freeink-sdk, or in the keyboardactivity?

itsthisjustin:
> Also, the keyboard now bypasses UITheme entirely, so RoundedRaff's key styling is silently ignored. I'm not sure if it's what we want...

I'm definitely less concerned about theming a keyboard. hurts usability imo and makes it heavy for no reason

itsthisjustin:
> One last thing: How much the new keyboard is scalable? If I will want tomorrow to add Hebrew layout, do I need to add it in freeink-sdk, or in the keyboardactivity?

for a language layout the right place is the SDK. add Hebrew to KeyboardLayoutId plus its tables in builtinKeyboardLayout() (where AzertyFr/QwertzDe live), then it's a one-line mapping in layoutForLanguage() here. Hebrew has no concept of shift layout as far as I can tell and RTL cursor editing would need work.

src/activities/reader/EpubReaderActivity.cpp
itsthisjustin: Solving in the other open branch for performance

src/activities/settings/LanguageSelectActivity.cpp
itsthisjustin: Should be consolidated now

src/activities/ActivityManager.cpp
```diff
@@ -61,6 +66,14 @@ void ActivityManager::renderTaskLoop() {
 
 void ActivityManager::loop() {
   if (currentActivity) {
+    if (currentActivity->name != "Home" && mappedInput.wasHomeGesture()) {
```
Uri-Tauber: Too fragile. 
Maybe add `virtual bool isHomeActivity() const { return false; }` on `Activity` and an override in `HomeActivity`. (like isReaderActivity we already have.
Then this line become `if (!currentActivity->isHomeActivity()...`

lib/hal/HalPowerManager.cpp
```diff
@@ -75,53 +72,46 @@ void HalPowerManager::startDeepSleep(HalGPIO& gpio) const {
...
+  // hold the enables off through deep sleep — otherwise the GT911 and SD card
+  // stay powered all through "off" and drain the battery. No-op on boards with
+  // no switched rails (X4/X3). Trade-off: no touch-to-wake; wake is the power
+  // button. Must run after display.deepSleep() so the panel controller gets its
+  // deep-sleep command while its rail is still up (enterDeepSleep() in main.cpp
+  // guarantees that ordering).
+  freeink::PowerManager::powerDownRailsForSleep();
+
+  freeink::PowerManager::armPowerButtonWakeup();
+  freeink::PowerManager::deepSleep();
```
Uri-Tauber: Maybe replace those two with `PowerManager::deepSleepUntilPowerButton()` from freeink?

Uri-Tauber (approved):
Working well, and the codebase is clean and functional (as much as I can say). 

@uxjulia do you want to try it before merging? Second pair of eyes would be welcome.

Uri-Tauber:
One small thing worth fixing before merging: when long-pressing item in file manager and the delete popup come up, pressing back (on the x4) is going two levels up - instead of one. On Sticky it behave fine.

PR #2490, opened 2026-06-30
feat: Add Both Next side button layout

jack-manuel:
## Summary

Implements Discussion #958

Adds "Both Next" option to Side Button Layout reader settings list. Either button advances. Front button behaviour is unchanged so can still page back.

## Additional Context

Tested on Xteink X3 only.

- Prev/Next, Next/Prev, Disabled are all unchanged
- Added a description of the new setting to the user guide

### AI Usage

Did you use AI tools to help write this code?  PARTIALLY

I used ChatGPT to help with dev setup and testing. Code was written, flashed and tested manually.


jack-manuel:
Added a release type to the title, sorry. Do I need to resubmit the pull request?

PR #875, opened 2026-02-13
feat: Initial support for the x3

itsthisjustin:
## Summary

Adds Xteink X3 hardware support to CrossPoint Reader. The X3 uses the same SSD1677 e-ink controller as the X4 but with a different panel (792x528 vs 800x480), different button layout, and an I2C fuel gauge (BQ27220) instead of ADC-based battery reading.

All X3-specific behavior is gated by runtime device detection — X4 behavior is unchanged.

Depends on community-sdk X3 support: open-x4-epaper/community-sdk#19 (merged).

### HAL Layer

**HalGPIO** (`lib/hal/HalGPIO.cpp/.h`)
- I2C-based device fingerprinting at boot: probes for BQ27220 fuel gauge, DS3231 RTC, and QMI8658 IMU to distinguish X3 from X4
- Detection result cached in NVS for fast subsequent boots
- Exposes `deviceIsX3()` / `deviceIsX4()` helpers used throughout the codebase
- X3 button mapping (7 GPIOs vs X4's layout)
- USB connection detection and wake classification for X3

**HalDisplay** (`lib/hal/HalDisplay.cpp/.h`)
- Calls `einkDisplay.setDisplayX3()` before init when X3 is detected
- Requests display resync after power button / flash wake events
- Runtime display dimension accessors (`getDisplayWidth()`, `getDisplayHeight()`, `getBufferSize()`)
- Exposed as global `display` instance for use by image converters

**HalPowerManager** (`lib/hal/HalPowerManager.cpp/.h`)
- X3 battery reading via I2C fuel gauge (BQ27220 at 0x55, SOC register)
- X3 power button uses GPIO hold for deep sleep

### Display & Rendering

**GfxRenderer** (`lib/GfxRenderer/GfxRenderer.cpp/.h`)
- Buffer size and display dimensions are now runtime values (not compile-time constants) to support both panel sizes
- X3 anti-aliasing tuning: only the darker grayscale level is applied to avoid washed-out text on the X3 panel. X4 retains both levels via `deviceIsX4()` gate

**Image Converters** (`lib/JpegToBmpConverter`, `lib/PngToBmpConverter`)
- Cover image prescale target uses runtime display dimensions from HAL instead of hardcoded 800x480

### UI Themes

**BaseTheme / LyraTheme** (`src/components/themes/`)
- X3 button position mapping for the different physical layout
- Adjusted UI element positioning for 792x528 viewport

### Boot & Init

**main.cpp**
- X3 hardware detection logging
[...]

PR #2492, opened 2026-06-30
fix: remove duplicate sleep logic

### LINKED ISSUE #2459: XTEINK-4 crashes during standby, drains battery
### Affected Version

v1.4.0

### Describe the Bug

When I go to standby, the screen switches to normal standby mode. Then after letting it sit for a couple of hours, when I try to start it, nothing happens. I have to soft-reset (reset button + 3 seconds power) it every time. It seems to draw power frequently. So after 1 day it is down from 100% (green led) to 10% (red led). 

### Steps to Reproduce

Read a book
Go standby
Wait
Turn on again

### Expected Behavior

I would expect the app to recover from standby using a simple power press. 

```shell

```

Uri-Tauber:
## Summary

**What is the goal of this PR?** fix #2459 

The bug was caused by two symptoms:
  - Device won't wake after ~2 hours in standby - GPIO13 (battery latch MOSFET) was left floating instead of being driven LOW, causing the MCU to lose power completely
  - 100% to 10% battery drain in one day - Without GPIO13 LOW during sleep, the battery latch remained partially engaged, drawing continuous power

There are **two different `startDeepSleep()` implementations** in the codebase, and they are **critically different**:
  - `HalPowerManager::startDeepSleep()` (correct — used by `enterDeepSleep()`)
  - `HalGPIO::startDeepSleep()` (broken — used by `verifyPowerButtonWakeup()`)

  1. Removed `HalGPIO::startDeepSleep()` method entirely from `lib/hal/HalGPIO.cpp/.h`
  2. Changed `verifyPowerButtonWakeup()` return type from void to bool - returns true for successful verification, false when device should re-enter sleep
  3. Updated `main.cpp` to call `powerManager.startDeepSleep(gpio)` when `verifyPowerButtonWakeup()` returns false

---

### AI Usage

### DIFF
--- a/lib/hal/HalGPIO.cpp
+++ b/lib/hal/HalGPIO.cpp
@@ -225,22 +225,10 @@ unsigned long HalGPIO::getHeldTime() const { return inputMgr.getHeldTime(); }
 
 unsigned long HalGPIO::getPowerButtonHeldTime() const { return inputMgr.getPowerButtonHeldTime(); }
 
-void HalGPIO::startDeepSleep() {
-  // Ensure that the power button has been released to avoid immediately turning back on if you're holding it
-  while (inputMgr.isPressed(BTN_POWER)) {
-    delay(50);
-    inputMgr.update();
-  }
-  // Arm the wakeup trigger *after* the button is released
-  esp_deep_sleep_enable_gpio_wakeup(1ULL << InputManager::POWER_BUTTON_PIN, ESP_GPIO_WAKEUP_GPIO_LOW);
-  // Enter Deep Sleep
-  esp_deep_sleep_start();
-}
-
-void HalGPIO::verifyPowerButtonWakeup(uint16_t requiredDurationMs, bool shortPressAllowed) {
+bool HalGPIO::verifyPowerButtonWakeup(uint16_t requiredDurationMs, bool shortPressAllowed) {
   if (shortPressAllowed) {
     // Fast path - no duration check needed
-    return;
+    return true;
   }
   // TODO: Intermittent edge case remains: a single tap followed by another single tap
   // can still power on the device. Tighten wake debounce/state handling here.
@@ -262,11 +250,12 @@ void HalGPIO::verifyPowerButtonWakeup(uint16_t requiredDurationMs, bool shortPre
       inputMgr.update();
     } while (inputMgr.isPressed(BTN_POWER) && inputMgr.getPowerButtonHeldTime() < calibratedDuration);
     if (inputMgr.getPowerButtonHeldTime() < calibratedDuration) {
-      startDeepSleep();
+      return false;
     }
   } else {
-    startDeepSleep();
+    return false;
   }
+  return true;
 }
 
 bool HalGPIO::isUsbConnected() const {
--- a/lib/hal/HalGPIO.cpp
+++ b/lib/hal/HalGPIO.cpp
@@ -233,9 +233,9 @@ bool HalGPIO::verifyPowerButtonWakeup(uint16_t requiredDurationMs, bool shortPre
   // TODO: Intermittent edge case remains: a single tap followed by another single tap
   // can still power on the device. Tighten wake debounce/state handling here.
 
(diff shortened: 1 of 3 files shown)


lib/hal/HalGPIO.cpp
```diff
@@ -225,22 +225,10 @@ unsigned long HalGPIO::getHeldTime() const { return inputMgr.getHeldTime(); }
 
 unsigned long HalGPIO::getPowerButtonHeldTime() const { return inputMgr.getPowerButtonHeldTime(); }
 
-void HalGPIO::startDeepSleep() {
```
Uri-Tauber, line 228: @itsthisjustin I love you to review this change, since this code was yours (from #875).
Need to make sure no regression was introduced here, and to wait for verification from the reporters in #2459.

itsthisjustin: I'm confused. This change has been in for months with zero reports of this?

Uri-Tauber: Me too. I'm not sure what exactly changed between 1.3.0 and 1.4.0 which trigger the bug. However, this rerouting should be the proper fix to whatever happen to accidently cause the bug.
I still prefer to trace the exact cause, but I haven't been able to reproduce the bug on my device...

itsthisjustin: Check this out: https://www.reddit.com/r/xteinkereader/comments/1uejxw1/comment/ouraisj/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button

Uri-Tauber: What am I looking at?
Doesn't seems to be connected. 
If I have to guess, I think this user's device having some hardware power-related problem.

itsthisjustin: So i could be very wrong, but i suspect they are related. It sounds like at idle cpu something is dropping power so low that it can't power back on or crashes. This happens in the reader when you don't turn pages and it's been reported in a number of threads where users are letting it sit on a screen for a long time

Uri-Tauber: Your theory explains the bug from the Reddit post, but it doesn't explain why, in #2459, they reported that the battery drains within hours. On the contrary, if the device is powered-down more than normal, I would expect it to consume less power, not more.

Uri-Tauber: My suspect is the bug from #2459 happen when the power button is accidently pressed shortly, the device wake-up, see the press was shorter than threshold, and go to sleep again, but this time via the wrong route.

itsthisjustin: Would an epaper device with no screen transitions and no wifi and nothing else happening really drain within hours?

Uri-Tauber: Thats a good question. From the [documentation](https://cdn.sparkfun.com/assets/1/a/5/a/6/esp32-c3_datasheet_en.pdf) page 30, we can learn that when the modem is off, but the cpu is active at 160Mhz, it draw 20mA. So it would drain the battery in ~30 hours. 
If the battery is already a bit old, the numbers fit.

itsthisjustin: Well we can merge this I suppose. It's been so long since it was added I can only say it may have been left over from where I tried to make x3 match x4 button hold stuff for sleep and it should have been removed when I later backed it out, but it wasn't. It's been in the code since early march though

Uri-Tauber: Lets wait to hear from the guys in #2459 first.

Uri-Tauber: Well, seems like no response there. I'm less convinced now it is the real cause of the bug they report, but it's an actual bug, nonetheless.

Uri-Tauber: And it clean the codebase - which is always nice.

itsthisjustin (approved):

PR #2500, opened 2026-07-01
feat: add server shortcut to OPDS picker

### LINKED ISSUE #2493: FR: Add shortcut to add new OPDS server
I think it would be great if at the end of the list of OPDS servers that you get to when you select OPDS Browser from the main menu, there was an "Add new server" option, that sent you directly to the Add server setting.

Thank you for your work! :)

ishaanlabs-gg:
## Summary
- include the existing Add Server row in the OPDS browser picker
- route that picker row to the existing new-server settings flow
- keep existing server selection routing to the OPDS browser unchanged

Fixes #2493.

## Testing
- ./bin/clang-format-fix -g
- git diff --check
- cppcheck on src/activities/settings/OpdsServerListActivity.cpp passed before cleaning local PlatformIO artifacts; it only reported generated I18nKeys.h was absent outside the build-generation step

I also attempted the recommended PlatformIO checks with PlatformIO and clang-format installed in an isolated SSD dependency environment. The full pio run/check path could not complete locally because the external volume generated thousands of AppleDouble ._* files while unpacking the ESP32 Arduino toolchain archives, causing PlatformIO package extraction/install noise before compilation.

### DIFF
--- a/src/activities/settings/OpdsServerListActivity.cpp
+++ b/src/activities/settings/OpdsServerListActivity.cpp
@@ -59,6 +59,13 @@ void OpdsServerListActivity::loop() {
 
 void OpdsServerListActivity::handleSelection() {
   const auto serverCount = static_cast<int>(OPDS_STORE.getCount());
+  auto resultHandler = [this](const ActivityResult&) {
+    OPDS_STORE.loadFromFile();
+    selectedIndex = 0;
+  };
+  auto openServerEditor = [this, resultHandler](int serverIndex) {
+    startActivityForResult(std::make_unique<OpdsSettingsActivity>(renderer, mappedInput, serverIndex), resultHandler);
+  };
 
   if (pickerMode) {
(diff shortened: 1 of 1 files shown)


Uri-Tauber:
Have you tested on real hardware?

ishaanlabs-gg:
I have not tested this on real hardware. I validated the code path with the project checks that are available on the PR: build, unit-tests, cppcheck, clang-format, and the title check are all green.\n\nThe change is limited to exposing the existing add-server flow from picker mode and keeping the same reload-on-return behavior used by the settings flow. If there is a hardware-specific navigation/input behavior you want covered, I can adjust the patch around that.

forteller:
Hi! May I ask if there's anything specific holding this back? Thank you very much!

Uri-Tauber:
> Hi! May I ask if there's anything specific holding this back? Thank you very much!

A. someone human to test this on real hardware - the PR author is obviously a bot. 
B. Resolve conflicts - this PR is 138 commit behind `develop`.

PR #2508, opened 2026-07-01
fix: footnote navigation on page border and footnote anchor drift when hyphenation enabled

lxyd:
## Summary

Fix two footnote bugs:
1. with hyphenation enabled footnotes on long paragraphs can break because anchors offset (which is measured in words) drift as it is measured in hyphenated, split words, but target anchor's offset is mesaured in original, not hyphenated words
2. when note text starts from a fresh page anchor still targets previous page

### Demo for bug 1: footnote 3 works fine while footnote 4 is not available in context menu

https://github.com/user-attachments/assets/aee1a765-df48-4afe-8eac-f61b34d4422b

### Demo for bug 2: footnote 1 tooks one full page, so footnote 2 target is exactly on page border, footnote 2 target misses

https://github.com/user-attachments/assets/b99985fc-7709-40c7-99ae-86ee4eecead3

## Additional Context

I was considering adding originalWordCount to TextBlock class, but TextBlock is serializing to cache and originalWordCount is not needed after initial line processing. So there would be 2 bytes extra cache size for each line. Not much, but unnecessary. So there was added a transient info class for it.

---

### AI Usage

### DIFF
--- a/lib/Epub/Epub/ParsedText.cpp
+++ b/lib/Epub/Epub/ParsedText.cpp
@@ -277,6 +277,7 @@ void ParsedText::addWord(std::string word, const EpdFontFamily::Style fontStyle,
     wordContinues.push_back(continues);
     wordNoSpaceBefore.push_back(noSpaceBefore);
     wordIsFocusSuffix.push_back(isFocusSuffix);
+    wordIsHyphenRemainder.push_back(false);
   };
 
   bool effectiveAttachToPrevious = attachToPrevious;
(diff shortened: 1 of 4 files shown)


itsthisjustin:
@lxyd can you solve conflicts and make sure this doesn't conflict with #2457 #2453

lxyd:
Sorry, it seems like I messed up things a bit.

First: this MR does NOT fix a #2336 bug as I mistakingly thought (the affected book has some anchor ids longer than FOOTNOTE_HREF_LEN. So there is nothing to do with anchor layouting).

Second: my first-try fix for one-page-early anchors (guessing anchor's page by summing up padding, margin and line height) turned out to be unstable and I had to re-work the approach: instead of pre-layout heuristics, an anchor is recorded only when its content is actually placed on a page.

I updated MR

unnamedd:
Hello @lxyd, do you have plans to fix the problems in your PR?

PR #2516, opened 2026-07-02
fix: nothrow allocs, RISC-V alignment, OTA wifi-ps, settings persist

sypianski:
## Summary

* **What is the goal of this PR?** Fix a set of correctness and stability bugs found via static analysis (cppcheck) and manual review. No new features, no behavior changes for correctly-functioning code paths.

### Crash fixes — bare `new` without nothrow guard

With `-fno-exceptions`, bare `new` on OOM calls `abort()` rather than returning `nullptr` (CLAUDE.md Rule 9). The following sites were converted to `new (std::nothrow)` / `makeUniqueNoThrow` with null checks:

- `lib/Epub/Epub/Page.cpp` — `PageLine`, `PageImage`, `Page` deserializers. Made consistent with the existing `PageHorizontalRule::deserialize` pattern in the same file. Null-check propagation added in the `Page::deserialize` loop for `TAG_PageLine` and `TAG_PageImage`.
- `lib/Epub/Epub/converters/PngToFramebufferConverter.cpp` — `HalFile` alloc in PNG open callback. Ownership passes to PNGdec C API so `new (std::nothrow)` used directly.
- `lib/PngToBmpConverter/PngToBmpConverter.cpp` — ditherer objects + `uint32_t[]`/`uint16_t[]` scaling accumulators.
- `src/activities/reader/EpubReaderActivity.cpp:848` — `Section` alloc; converted to `makeUniqueNoThrow<Section>`.

### RISC-V alignment crash

- `lib/ZipFile/ZipFile.cpp` — `loadZipDetails` used three `*reinterpret_cast<uintN_t*>(&buf[i])` reads on a `malloc`'d `uint8_t*` buffer. `buf` has no alignment guarantee; unaligned multi-byte loads fault on ESP32-C3 RISC-V (CLAUDE.md §ESP32-C3 Platform Pitfalls). Replaced with `memcpy` into typed locals. Triggered on every EPUB open.

### OTA updater

- `isUpdateNewer`: `sscanf` return value unchecked, all six `int` targets uninitialized. A GitHub tag with a `v` prefix (`v1.4.1`) causes `sscanf` to return 0 and leave stack garbage — could suppress a real update or trigger a spurious install. Guard: check `== 3`, initialize to 0.
- `installUpdate`: `esp_wifi_set_ps(WIFI_PS_NONE)` called before `esp_https_ota_begin`, but not restored on the early-return failure path. Radio stayed in max-power mode for the rest of the session, draining the 650 mAh battery. Fix: restore `WIFI_PS_MIN_MODEM` before the early return.

### Settings persistence

- `src/SdCardFontSystem.cpp` — four sites clear `SETTINGS.sdFontFamilyName` on missing/failed SD font load without calling `SETTINGS.saveToFile()`. The clear was lost on reboot, so the device retried the (still-failing) load on every boot. Fix: add `saveToFile()` after each clear. Pattern taken from `FontInstaller::deleteFamily` which already handles this correctly.

### Duplicate settings key

- `src/JsonSettingsIO.cpp` (`saveSettings`) — `doc["language"]` written twice in succession (copy-paste). ArduinoJson overwrites duplicates so output was correct, but future divergence between the two copies would be silent. Removed the second copy.

### Missing `reserve()` before `push_back` loops

Per CLAUDE.md Rule 7 (vector pre-allocation):
- `src/JsonSettingsIO.cpp` (`loadRecentBooks`) — `store.recentBooks.reserve(min(arr.size(), 10))` added.
- `src/WifiCredentialStore.cpp` (`loadFromBinaryFile`) — `credentials.reserve(min(count, MAX_NETWORKS))` added.

### OpdsParser handler re-registration

- `lib/OpdsParser/OpdsParser.cpp` — `XML_SetUserData`, `XML_SetElementHandler`, `XML_SetCharacterDataHandler` were called on every `write()` invocation (once per 1 KB chunk). Moved to constructor. Added `errorOccured || !parser` guard in `flush()` to avoid calling into a freed parser if `write()` already destroyed it on error.

### Minor

- `src/activities/home/FileBrowserActivity.cpp` — `getFileExtension(std::string)` → `const std::string&`.
- `lib/Txt/Txt.cpp` — `substr(0, len-4)` on a local string → `resize(len-4)` (avoids allocation).

## Additional Context

All changes pass `pio run -e default` with 0 errors, 0 warnings. RAM: 31% (101 KB / 328 KB). Formatted with clang-format-21.

No hardware testing was performed — the changes are all in code paths that are exercised on every book open (ZIP/PNG/EPUB deserialization) or on OTA update, so hardware verification would be straightforward.

---

### AI Usage

Used Claude (claude.ai/code) for static analysis fan-out across the codebase and to draft the fixes. All changes were reviewed and verified manually before committing. The project's CLAUDE.md was used as the primary correctness reference throughout.


sypianski:
Closing — submitted prematurely, not ready for review.

PR #2517, opened 2026-07-02
fix: use nothrow allocs and memcpy for RISC-V alignment

sypianski:
## Summary

* **What is the goal of this PR?** Fix five crash-level bugs where bare `new` or unaligned pointer casts would abort or fault the device on the ESP32-C3.
* **What changes are included?**

### Unaligned reads in ZipFile (RISC-V fault on every EPUB open)

`lib/ZipFile/ZipFile.cpp` — `loadZipDetails` used three `*reinterpret_cast<uintN_t*>(&buf[i])` reads on a `malloc`'d `uint8_t*` buffer. The buffer has no alignment guarantee; unaligned multi-byte loads fault on ESP32-C3 RISC-V (documented in your CLAUDE.md §RISC-V Alignment). This code runs on every EPUB open. Replace with `memcpy` into typed locals.

### Bare `new` without nothrow guard

With `-fno-exceptions`, bare `new` on OOM calls `abort()` rather than returning `nullptr` (CLAUDE.md Rule 9). Four sites fixed:

- `lib/Epub/Epub/Page.cpp` — `PageLine`, `PageImage`, `Page` deserializers used bare `new`. Convert to `new (std::nothrow)` + null-check, consistent with the existing `PageHorizontalRule` pattern in the same file. Null-check propagation added in `Page::deserialize` loop.
- `lib/Epub/Epub/converters/PngToFramebufferConverter.cpp` — `HalFile` alloc in the PNG open callback. Ownership passes to PNGdec C API, so `new (std::nothrow)` used directly.
- `lib/PngToBmpConverter/PngToBmpConverter.cpp` — ditherer objects and scaling accumulators (`uint32_t[]`, `uint16_t[]`). Convert to `new (std::nothrow)` with OOM error paths.
- `src/activities/reader/EpubReaderActivity.cpp:848` — `Section` alloc; convert to `makeUniqueNoThrow<Section>` with null-check before `loadSectionFile`.

## Additional Context

All changes follow the constraints in your CLAUDE.md (Rule 9, §RISC-V Alignment). Verified with `pio run` (0 errors, 0 warnings) and `pio check --fail-on-defect high` (no defects). Formatted with clang-format-21.

No hardware testing was performed — human verification on device is recommended before merging.

This is part of a series of 3 focused fix PRs. See also:
- #2518 — OTA version parse and WiFi power save
- #2519 — Settings persistence, reserve(), minor

---

### AI Usage

Did you use AI tools to help write this code? **YES**

[...]

PR #2518, opened 2026-07-02
fix: guard sscanf return and restore wifi-ps in OtaUpdater

sypianski:
## Summary

* **What is the goal of this PR?** Fix two bugs in OTA update handling: an sscanf-based version comparison that reads uninitialised stack values, and a WiFi power-save mode that is left at WIFI_PS_NONE on an early-exit error path.

### OTA version comparison reads uninitialised stack (OtaUpdater.cpp)

`isUpdateNewer()` declared six version-component `int` locals without initialiser values. The code then called `sscanf` on both version strings and compared the locals without checking whether `sscanf` succeeded. A malformed manifest URL or a dev build without a semver tag (e.g. `"dev"`) would leave all six values undefined; on RISC-V the comparison would use whatever bytes happened to be on the stack, potentially reporting a phantom "newer" version or suppressing a real update.

Fix: zero-initialise all six locals; guard both `sscanf` calls with a `!= 3` check that logs and returns `false` immediately on parse failure.

### WiFi power-save not restored on OTA begin failure (OtaUpdater.cpp)

`installUpdate()` disables modem sleep (`WIFI_PS_NONE`) for the duration of the OTA HTTP download, then re-enables it (`WIFI_PS_MIN_MODEM`) at the end. The function had a single early-exit path — `esp_https_ota_begin` failure — that returned `INTERNAL_UPDATE_ERROR` without restoring modem sleep, leaving the radio in high-power mode for the rest of the session.

Fix: add `esp_wifi_set_ps(WIFI_PS_MIN_MODEM)` before the early return.

## Additional Context

Changes verified with `pio run` (0 errors, 0 warnings) and `pio check --fail-on-defect high` (no defects). Formatted with clang-format-21.

No hardware testing was performed — human verification on device is recommended before merging.

This is part of a series of 3 focused fix PRs. See also:
- #2517 — nothrow allocations + RISC-V unaligned reads
- #2519 — Settings persistence, reserve(), minor

---

### AI Usage

These fixes were identified by Claude Fable 5 (Anthropic) via static analysis and manual review. All changes were verified against the project's own CLAUDE.md Resource Protocol. We flag this transparently in case AI-originated contributions require additional scrutiny.

### DIFF
--- a/src/network/OtaUpdater.cpp
+++ b/src/network/OtaUpdater.cpp
@@ -71,14 +71,19 @@ bool OtaUpdater::isUpdateNewer() const {
     return false;
   }
 
-  int currentMajor, currentMinor, currentPatch;
-  int latestMajor, latestMinor, latestPatch;
+  int currentMajor = 0, currentMinor = 0, currentPatch = 0;
+  int latestMajor = 0, latestMinor = 0, latestPatch = 0;
 
   const auto currentVersion = CROSSPOINT_VERSION;
 
   // semantic version check (only match on 3 segments)
-  sscanf(latestVersion.c_str(), "%d.%d.%d", &latestMajor, &latestMinor, &latestPatch);
-  sscanf(currentVersion, "%d.%d.%d", &currentMajor, &currentMinor, &currentPatch);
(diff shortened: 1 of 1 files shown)


itsthisjustin:
@sypianski can you solve conflcits and see what in this is still needed after the recent wolfssl changes that were just brought in?

PR #2519, opened 2026-07-02
fix: settings persist on font clear, reserve() before push_back, minor

sypianski:
## Summary

* **What is the goal of this PR?** Four small, independent fixes: persist the SD-card font family name across reboots, guard two reserve() calls against oversized inputs, fix a lost XML handler registration in OpdsParser, and a minor pass-by-value elimination.

### SD font selection not saved (SdCardFontSystem.cpp)

After a user picks a font family stored on the SD card, the system stores the family name in `sdFontFamilyName`. Four code sites (two in `begin()`, two in `ensureLoaded()`) clear this member on error or when the SD card is absent, but none called `SETTINGS.saveToFile()` afterward. On next boot the settings file still held the stale SD-font name, so the system attempted to load a font from a missing card and fell back to a built-in font silently.

Fix: add `SETTINGS.saveToFile()` after each of the four clear sites.

### reserve() without upper-bound guard (src/JsonSettingsIO.cpp, src/WifiCredentialStore.cpp)

Both files called `reserve(arr.size())` before a push_back loop. `arr.size()` comes from deserialised JSON; a corrupt settings file could provide an arbitrarily large value, triggering a large allocation before a single element is validated (Rule 7 in CLAUDE.md: always cap reserve()).

Fix: cap with `std::min(arr.size(), (size_t)10)` / `std::min<size_t>(count, MAX_NETWORKS)` respectively. Added `#include <algorithm>` where missing.

Also removed a duplicate `doc["language"]` assignment line in `JsonSettingsIO.cpp` (copy-paste leftover).

### OpdsParser XML handlers lost on re-use (OpdsParser.cpp)

`OpdsParser::write()` registered XML element and character-data handlers on every call. If `write()` was never called (parse error on the first chunk) or if a caller passed the parser through a code path that skipped `write()`, handlers were never registered and the parser would silently produce empty results.

Fix: move handler registration (`XML_SetElementHandler`, `XML_SetCharacterDataHandler`) to the constructor, which runs exactly once. Guard `flush()` with an early return if `errorOccured || !parser`.

### Minor: pass-by-value string (src/activities/home/FileBrowserActivity.cpp)

`getFileExtension` accepted `std::string filename` by value, copying on every directory-listing call. Changed to `const std::string&`.

Also: `lib/Txt/Txt.cpp` — replaced `filename.substr(0, n-4)` with `filename.resize(n-4)` (no intermediate allocation).

## Additional Context

Changes verified with `pio run` (0 errors, 0 warnings) and `pio check --fail-on-defect high` (no defects). Formatted with clang-format-21.

No hardware testing was performed — human verification on device is recommended before merging.

This is part of a series of 3 focused fix PRs. See also:
- #2517 — nothrow allocations + RISC-V unaligned reads
- #2518 — OTA version parse and WiFi power save

---

### AI Usage

These fixes were identified by Claude Fable 5 (Anthropic) via static analysis and manual review. All changes were verified against the project's own CLAUDE.md Resource Protocol. We flag this transparently in case AI-originated contributions require additional scrutiny.

### DIFF
--- a/lib/OpdsParser/OpdsParser.cpp
+++ b/lib/OpdsParser/OpdsParser.cpp
@@ -10,7 +10,11 @@ OpdsParser::OpdsParser() {
   if (!parser) {
     errorOccured = true;
     LOG_DBG("OPDS", "Couldn't allocate memory for parser");
+    return;
   }
+  XML_SetUserData(parser, this);
+  XML_SetElementHandler(parser, startElement, endElement);
+  XML_SetCharacterDataHandler(parser, characterData);
 }
 
 OpdsParser::~OpdsParser() { destroyXmlParser(parser); }
@@ -19,10 +23,6 @@ size_t OpdsParser::write(uint8_t c) { return write(&c, 1); }
 
 size_t OpdsParser::write(const uint8_t* xmlData, const size_t length) {
   if (errorOccured) return length;
-
-  XML_SetUserData(parser, this);
-  XML_SetElementHandler(parser, startElement, endElement);
-  XML_SetCharacterDataHandler(parser, characterData);
 
   const char* currentPos = reinterpret_cast<const char*>(xmlData);
   size_t remaining = length;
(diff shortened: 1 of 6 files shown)


Uri-Tauber (approved):
Nice Cleanups.

PR #2521, opened 2026-07-02
feat: cjk UI font fallback

szetszho:
# CJK fallback for UI text (size-matched, string-level)

## Summary

Loading a CJK-capable SD-card font already lets CrossPoint show CJK **book
content** correctly — but it did nothing for CJK in the **UI**: book titles in
the library, file names in the browser, list rows, and headers still showed
replacement boxes, because those are always drawn with the built-in Latin-only
flash fonts, independent of whatever SD font is loaded for the book text. This
PR closes that gap: any UI string containing a CJK codepoint is now
automatically routed to the user's already-selected SD-card font, at the same
point size as the surrounding Latin UI text. It also fixes a reader-size
regression that a naive implementation would otherwise introduce for CJK font
packs that ship small UI-only sizes.

## Background

CrossPoint already supports SD-card fonts with extended Unicode coverage, and
loading one is sufficient on its own to render CJK **book content** — the
reader's text-layout path already uses the loaded SD font family directly. The
renderer picks one physical point size per family (closest to the user's
reader-size setting: 12/14/16/18pt) and loads only that file to keep resident
memory down.

**UI chrome is a separate code path** and was unaffected by any of this: it
always draws with the built-in flash fonts (`SMALL_FONT_ID` @ 8pt,
`UI_10_FONT_ID` @ 10pt, `UI_12_FONT_ID` @ 12pt), which have no CJK glyphs at
all — regardless of which SD font, if any, is loaded for the book. So before
this PR, a user could select a CJK SD font, read CJK book content fine, and
still see boxes for a CJK book *title* on the library screen.

### 1. String-level fallback routing (`GfxRenderer`)

- Added `fallbackFontMap_` (`std::map<int,int>`): primary UI font id ->
  size-matched fallback font id.
- Added `resolveTextFontId(fontId, text, style)`: walks the codepoints of
  `text`; if any is CJK (`utf8IsCjkBreakable`) and the primary font's active
  style can't render it (`EpdFontFamily::hasCodepoint`), returns the
  registered fallback id, else returns `fontId` unchanged.
- Wired into every text entry point that needs consistent metrics:
  `getTextWidth`, `drawText`, `getTextAdvanceX`, `drawTextRotated90CW`. Each
  resolves once up front and uses the resolved id for lookup, ascender calc,
  and advance-table fast path — so a string is never split across two fonts
  mid-render.
- Public API: `setFallbackFont(primaryId, fallbackId)` /
  `clearFallbackFonts()`.

Routing is **per string, not per glyph** — a mixed title like `三体 Vol.1`
renders entirely in the fallback font (including the Latin portion). This is
called out in the docs since a `Mono` fallback family will render the Latin
part at half/full width.

### 2. Interval-only glyph check (`EpdFont` / `EpdFontFamily`)

- Added `EpdFont::hasCodepoint(cp)`: binary search over the font's interval
  table. Unlike `getGlyph()`, it does **not** consult the on-demand SD miss
  handler or fall back to the replacement glyph — it answers "can this font's
  own table draw this codepoint" so the fallback decision doesn't trigger a
  disk load just to check.
- Added `EpdFontFamily::hasCodepoint(cp, style)` forwarding to the
  style-resolved `EpdFont`.

### 3. Size-matched fallback loading (`SdCardFontManager`, `SdCardFontSystem`)

- Refactored `SdCardFontManager::loadFamily` to share a new private
  `loadFile(file, familyName, renderer)` helper (alloc, disk load, id
  compute/collision check, register, insert) instead of inlining it.
- Added `loadFamilyExtraSize(family, renderer, pointSize)`: additively loads
  the family's `.cpfont` at an exact physical size (used for UI sizes 8/10/12)
  without touching the already-loaded reader-size font. Reuses an existing
  loaded size instead of double-loading if one already matches. Returns 0 if
  the family has no file at that size.
- `unloadAll` now calls `renderer.clearFallbackFonts()` before freeing the SD
  fonts the map points to, so no dangling fallback ids survive a family
  switch.
- `SdCardFontSystem::setupUiFallbacks(renderer)`: for each built-in UI size
  (8/10/12pt), calls `loadFamilyExtraSize` on the active family and, if a file
  exists, registers it via `renderer.setFallbackFont(uiFontId, sdFontId)`.
  Called after `loadFamily` succeeds in both `begin()` and `ensureLoaded()`.
  No-op (and logged at DBG) for any size the family doesn't ship.

### 4. Reader-size selection fix (`SdCardFontRegistry::findClosestReaderSize`)

Families that add 8/10pt UI-fallback files alongside the four reader sizes
(12/14/16/18) would otherwise shift the existing ordinal (index-based)
mapping used for custom font packs — e.g. a family with files at
`8,10,12,14,16,18` has 6 sizes, so `MEDIUM` (enum 1) would index to the
*second* size (10pt) instead of 14pt, making body text too small.

Fix: when a family provides all four standard reader sizes (12/14/16/18) for
the requested style, map the enum straight to `{12,14,16,18}` regardless of
how many extra sizes exist. Ordinal selection is now only used as a fallback
for families that *don't* have the full standard set (genuinely custom-built
packs like 10/12/14/16), and closest-match-by-delta remains the last resort
for packs with fewer than 4 sizes.

## Files touched

| File | Change |
|---|---|
| `lib/EpdFont/EpdFont.{h,cpp}` | `hasCodepoint()` interval-only lookup |
| `lib/EpdFont/EpdFontFamily.{h,cpp}` | `hasCodepoint()` forwarding |
| `lib/EpdFont/SdCardFontManager.{h,cpp}` | `loadFile()` extraction, `loadFamilyExtraSize()`, fallback-map clear on unload |
| `lib/EpdFont/SdCardFontRegistry.cpp` | `findClosestReaderSize` standard-4-sizes fast path |
| `lib/GfxRenderer/GfxRenderer.{h,cpp}` | `fallbackFontMap_`, `resolveTextFontId()`, wiring into draw/measure paths |
| `src/SdCardFontSystem.{h,cpp}` | `setupUiFallbacks()`, called from `begin()`/`ensureLoaded()` |
| `docs/sd-card-fonts.md` | New "CJK in the User Interface" section |

## Memory impact

Additive only when a CJK SD family is active and ships UI sizes: up to 3
extra small `.cpfont` files resident (8/10/12pt), on top of the one
reader-size file already loaded. No change for users without an SD font
selected, or whose family lacks 8/10/12pt files (those sizes silently keep
showing boxes, as before this change).

## Test plan

- [x] Install a CJK family with reader sizes only (no 8/10/12) — verify UI
      still shows boxes for CJK (no crash, no regression) and body text sizes
      correctly.
- [x] Install a CJK family with 8/10/12/12/14/16/18 (per `docs/sd-card-fonts.md`
      conversion example) — verify library titles, file browser rows, and
      headers render CJK at the right size, and reader body text at
      SMALL/MEDIUM/LARGE/EXTRA_LARGE still maps to 12/14/16/18pt.
  Note: the WIP `NotoSansMonoCJKhk-Regular.otf` in `lib/EpdFont/scripts/` is a
  local conversion source, not part of this PR's shipped changes.
- [x] Switch SD font family while a CJK title is on screen — confirm no stale
      fallback font id (freed SD font) is referenced (`clearFallbackFonts()`
      ordering in `unloadAll`).
- [x] Mixed Latin+CJK title (e.g. `三体 Vol.1`) — confirm whole string renders
      in the fallback font, not split per-glyph.
- [x] No SD font selected — confirm UI CJK still shows boxes (fallback map
      empty, no behavior change).

## Out of scope

- CJK glyph coverage in the built-in flash fonts themselves (this stays a
  pure SD-fallback feature to avoid flash-size growth).
- Per-glyph (as opposed to per-string) fallback routing.
- Automatic UI-size font conversion for existing pre-built font packs in the
  `crosspoint-fonts` repo (tracked separately; packs need re-conversion with
  `--sizes 8,10,12,14,16,18` to benefit).

### DIFF
--- a/lib/GfxRenderer/GfxRenderer.cpp
+++ b/lib/GfxRenderer/GfxRenderer.cpp
@@ -179,6 +179,36 @@ void GfxRenderer::insertFont(const int fontId, EpdFontFamily font) {
   }
 }
 
+int GfxRenderer::resolveTextFontId(const int fontId, const char* text, const EpdFontFamily::Style style) const {
+  if (fallbackFontMap_.empty() || text == nullptr || *text == '\0') {
+    return fontId;
+  }
+  const auto fbIt = fallbackFontMap_.find(fontId);
+  if (fbIt == fallbackFontMap_.end()) {
+    return fontId;  // no fallback registered for this font
+  }
+  const int fallbackFontId = fbIt->second;
+  const auto fontIt = fontMap.find(fontId);
+  const auto fallbackIt = fontMap.find(fallbackFontId);
+  if (fontIt == fontMap.end() || fallbackIt == fontMap.end()) {
+    return fontId;  // unknown primary or fallback not loaded — let the caller handle it
+  }
+  const EpdFontFamily& primary = fontIt->second;
+  const EpdFontFamily& fallback = fallbackIt->second;
+  const char* cursor = text;
+  uint32_t cp;
+  while ((cp = utf8NextCodepoint(reinterpret_cast<const uint8_t**>(&cursor)))) {
+    // Only redirect for CJK the primary font cannot draw but the fallback can.
+    // Latin/symbol strings the built-in UI fonts already cover are left
+    // untouched, and a partial-coverage fallback (e.g. kana-only) is not worth
+    // dragging the whole string into for glyphs it would also miss.
+    if (utf8IsCjkCodepoint(cp) && !primary.hasCodepoint(cp, style) && fallback.hasCodepoint(cp, style)) {
+      return fallbackFontId;
+    }
+  }
+  return fontId;
+}
+
 // Translate logical (x,y) coordinates to physical panel coordinates based on current orientation
 // This should always be inlined for better performance
 static inline void rotateCoordinates(const GfxRenderer::Orientation orientation, const int x, const int y, int* phyX,
@@ -499,9 +529,12 @@ int GfxRenderer::getTextWidth(const int fontId, const char* text, const EpdFontF
     return 0;
   }
 
-  const auto fontIt = fontMap.find(fontId);
+  // Measure with the same font drawText would render with (see resolveTextFontId)
+  // so wrapping, truncation and centering of CJK strings stay consistent.
+  const int resolvedFontId = resolveTextFontId(fontId, text, style);
+  const auto fontIt = fontMap.find(resolvedFontId);
   if (fontIt == fontMap.end()) {
-    LOG_ERR("GFX", "Font %d not found", fontId);
+    LOG_ERR("GFX", "Font %d not found", resolvedFontId);
     return 0;
   }
 
@@ -526,10 +559,14 @@ void GfxRenderer::drawText(const int fontId, const int x, const int y, const cha
     return;
   }
 
+  // Route CJK-bearing strings to the fallback font when the requested font
+  // lacks the glyphs (e.g. Chinese book titles drawn with a Latin UI font).
+  const int resolvedFontId = resolveTextFontId(fontId, text, style);
+
   std::string visual;
   const char* renderedText = resolveVisualText(text, visual, baseDir);
 
-  const int yPos = y + getFontAscenderSize(fontId);
+  const int yPos = y + getFontAscenderSize(resolvedFontId);
   int lastBaseX = x;
   int lastBaseLeft = 0;
   int lastBaseWidth = 0;
@@ -537,13 +574,13 @@ void GfxRenderer::drawText(const int fontId, const int x, const int y, const cha
   int32_t prevAdvanceFP = 0;  // 12.4 fixed-point: prev glyph's advance + next kern for snap
 
   if (fontCacheManager_ && fontCacheManager_->isScanning()) {
-    fontCacheManager_->recordText(renderedText, fontId, style);
+    fontCacheManager_->recordText(renderedText, resolvedFontId, style);
     return;
   }
 
-  const auto fontIt = fontMap.find(fontId);
+  const auto fontIt = fontMap.find(resolvedFontId);
   if (fontIt == fontMap.end()) {
-    LOG_ERR("GFX", "Font %d not found", fontId);
+    LOG_ERR("GFX", "Font %d not found", resolvedFontId);
     return;
   }
   const auto& font = fontIt->second;
@@ -1828,6 +1865,8 @@ int GfxRenderer::getKerning(const int fontId, const uint32_t leftCp, const uint3
 }
 
 int GfxRenderer::getTextAdvanceX(const int fontId, const char* text, EpdFontFamily::Style style) const {
+  // Match the font drawText would use for CJK-bearing strings (see resolveTextFontId).
+  const int resolvedFontId = resolveTextFontId(fontId, text, style);
   // Measure the exact codepoint stream drawText renders: bidi-reordered and
   // Arabic-shaped (contextual presentation forms, Lam-Alef collapse).
   // Measuring the raw logical text counts the Alef a ligature absorbs and
@@ -1840,14 +1879,14 @@ int GfxRenderer::getTextAdvanceX(const int fontId, const char* text, EpdFontFami
   // Advance table fast-path for SD card fonts during layout.
   // No kerning/ligature lookup — consistent with previous metadataOnly behavior
   // where kern/lig data was not loaded.
-  auto sdIt = sdCardFonts_.find(fontId);
+  auto sdIt = sdCardFonts_.find(resolvedFontId);
   if (sdIt != sdCardFonts_.end() && sdIt->second->hasAdvanceTable()) {
     int32_t widthFP = 0;
     const bool isSupSub = (style & (EpdFontFamily::SUP | EpdFontFamily::SUB)) != 0;
     const uint8_t styleIdx = resolveSdCardStyle(*sdIt->second, style);
-    const auto fontIt = fontMap.find(fontId);
+    const auto fontIt = fontMap.find(resolvedFontId);
     if (fontIt == fontMap.end()) {
-      LOG_ERR("GFX", "Font %d not found", fontId);
+      LOG_ERR("GFX", "Font %d not found", resolvedFontId);
       return 0;
     }
     const auto& font = fontIt->second;
@@ -1866,9 +1905,9 @@ int GfxRenderer::getTextAdvanceX(const int fontId, const char* text, EpdFontFami
     return fp4::toPixel(widthFP);
   }
 
-  const auto fontIt = fontMap.find(fontId);
+  const auto fontIt = fontMap.find(resolvedFontId);
   if (fontIt == fontMap.end()) {
-    LOG_ERR("GFX", "Font %d not found", fontId);
+    LOG_ERR("GFX", "Font %d not found", resolvedFontId);
     return 0;
   }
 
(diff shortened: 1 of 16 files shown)


itsthisjustin:
I've been wanting to get a fix for this in for a bit. i planned to add font streaming into themes though which would allow you to have a theme with a UI font set which should solve the same kind of issue. Any thoughts @szetszho ?

szetszho:
> I've been wanting to get a fix for this in for a bit. i planned to add font streaming into themes though which would allow you to have a theme with a UI font set which should solve the same kind of issue. Any thoughts @szetszho ?

I think the main concern is the extra resources and speed impact if we were to load the different font for ui and content.

I have heard from user mentioning how this might take a bit longer to load each page of content.

eommisae01:
Thanks for working on this. My use case matches the problem this PR is solving: I know Korean-specific firmware builds exist, but I would prefer to keep using the original upstream CrossPoint firmware and only have user-content titles/file names render correctly when the selected SD-card font contains the needed glyphs.

I opened #2576 before noticing this PR, and it looks like this is the broader upstream direction. The main difference in #2576 was a narrower user-content/per-glyph fallback approach intended to avoid changing UI labels/settings/menu text and to keep fallback limited to book titles, recent-book entries, and file names. I'm closing #2576 as a duplicate and will follow/test this PR instead.

szetszho:
> Thanks for working on this. My use case matches the problem this PR is solving: I know Korean-specific firmware builds exist, but I would prefer to keep using the original upstream CrossPoint firmware and only have user-content titles/file names render correctly when the selected SD-card font contains the needed glyphs.
> 
> I opened #2576 before noticing this PR, and it looks like this is the broader upstream direction. The main difference in #2576 was a narrower user-content/per-glyph fallback approach intended to avoid changing UI labels/settings/menu text and to keep fallback limited to book titles, recent-book entries, and file names. I'm closing #2576 as a duplicate and will follow/test this PR instead.

Hey!, Thx for the interest! Would be nice if u can test it out in your usecase!

Uri-Tauber:
@szetszho Can you solve conflicts so we can get this reviewed soon?

szetszho:
@Uri-Tauber Conflict resolved

src/SdCardFontSystem.cpp
```diff
@@ -108,6 +124,23 @@ void SdCardFontSystem::ensureLoaded(GfxRenderer& renderer) {
   }
 }
 
+void SdCardFontSystem::setupUiFallbacks(GfxRenderer& renderer) {
+  const std::string& familyName = manager_.currentFamilyName();
+  if (familyName.empty()) return;  // no SD family loaded — nothing to fall back to
+
+  const auto* family = registry_.findFamily(familyName);
```
Uri-Tauber, line 131: This loads 1–3 extra SD font sizes and registers fallbacks unconditionally, even when the selected SD family is plain Latin.
On the 380 KB budget that's pure dead weight — resolveTextFontId only ever redirects on CJK the primary lacks, which a Latin font can't satisfy anyway. 

Probe the already-loaded reader font (no extra load) for Han/Hangul/Kana coverage and bail before loading any UI sizes.

lib/GfxRenderer/GfxRenderer.cpp
```diff
@@ -179,6 +179,32 @@ void GfxRenderer::insertFont(const int fontId, EpdFontFamily font) {
...
+  if (fontIt == fontMap.end() || fontMap.find(fallbackFontId) == fontMap.end()) {
+    return fontId;  // unknown primary or fallback not loaded — let the caller handle it
+  }
+  const EpdFontFamily& primary = fontIt->second;
+  const char* cursor = text;
+  uint32_t cp;
+  while ((cp = utf8NextCodepoint(reinterpret_cast<const uint8_t**>(&cursor)))) {
+    // Only redirect for CJK the primary font cannot draw. Latin/symbol strings
+    // the built-in UI fonts already cover are left untouched.
+    if (utf8IsCjkBreakable(cp) && !primary.hasCodepoint(cp, style)) {
```
Uri-Tauber: The whole string is routed to the fallback as soon as one CJK codepoint is missing from the primary, without checking the fallback can actually draw it. 
A partial-coverage SD font (e.g. kana-only) selected against a Hangul title trades a built-in replacement box for a fallback replacement box, and drags the Latin part of a mixed title into the SD font too. 

Require the fallback to cover the codepoint before redirecting.

Uri-Tauber: Using `utf8IsCjkBreakable(cp) could be problematic. This function skip some CJK ranges. Check it in `lib/Utf8/Utf8.h`.

Uri-Tauber:
`lib/EpdFont/scripts/NotoSansMonoCJKhk-Regular.otf`

This 16 MB .otf looks like an accidental commit — it's the source font for your local fontconvert_sdcard.py run, nothing in the tree references it.

README.md
```diff
@@ -1,5 +1,7 @@
 # CrossPoint Reader
 
+> **🌏 CJK Support Fork** — This fork adds size-matched Chinese, Japanese, and Korean (CJK) UI text support alongside book content. See [Reading CJK content](#reading-cjk-content) below for setup. For the official CrossPoint project, visit [crosspoint-reader/crosspoint-reader](https://github.com/crosspoint-reader/crosspoint-reader).
```
Uri-Tauber: edits in README.md should be minimal - one line top to reference this new feat.

szetszho: remove and add one liner

lib/GfxRenderer/GfxRenderer.cpp
szetszho: The redirect now requires the fallback to actually cover the codepoint: utf8IsCjkCodepoint(cp) && !primary.hasCodepoint(cp, style) && fallback.hasCodepoint(cp, style). So a kana-only SD font no longer captures a Hangul title — if neither font can draw the glyph, the string stays in the primary font instead of dragging its Latin part into the fallback for nothing.

szetszho: I added a broader utf8IsCjkCodepoint() for fallback font selection and left utf8IsCjkBreakable() alone, since its ranges are tuned for line-breaking and widening it could change layout.

src/SdCardFontSystem.cpp
szetszho: setupUiFallbacks() now probes the already-loaded reader-size font for one representative codepoint per script (Han U+4E00, Hiragana U+3042, Katakana U+30A2, Hangul U+AC00) before loading anything. If none are covered, it logs and returns without loading any UI sizes, so a plain Latin SD family costs nothing extra.

Uri-Tauber (approved):

brianhuster:
@szetszho Excuse me, does this PR support file picker and table of contents? I just updated to 1.5.0, it now shows CJK file name in the panel for most recently read book, but it shows nothing in Table of contents and file picker. FYI, my current reader font does have 12px variant

szetszho:
With my previous test on the PR build, I have CJK on all UI/File
explorer/Table of Contents. And that’s is the expected behaviour. I have
personally tested it and a fellow developer also tested it.

One thing to consider, do you have all the needed font size generated and
loaded to the SD card?

Iet me know how it goes and maybe upload a sample epud?

Uri-Tauber:
@brianhuster How did you generated your font?

```
    python3 lib/EpdFont/scripts/fontconvert_sdcard.py \
      MyCJKFont-Regular.otf \
      --intervals cjk \
      --sizes 8,10,12,14,16,18 \
      --style regular \
      --name MyCJKFont \
      --output-dir ./MyCJKFont/
```

szetszho:
However I did found a bug yesterday when testing the merged branch related  to “br” handling  affecting CJK epud, where most of CJK book use “br”differently, i will write up some thing tomorrow, i have a fix for that already.

brianhuster:
I generated my font from crosspoint
https://crosspointreader.com/fonts, with 4 font sizes 12px, 14px, 16px, 18px. It is the same font I used for other CJK screenshots in my PR. From the description, I thought it would have picked the 12px one for using in Table of Contents and File pickers, no?

Uri-Tauber:
Aha, I see the problem.

@szetszho, would you like to open a PR to https://github.com/crosspoint-reader/crosspoint-tools?

When CJK codepoints are selected under **Additional Unicode Coverage**, two additional **Font Size** options will appear, along with a note explaining why they are needed.

szetszho:
https://github.com/crosspoint-reader/crosspoint-tools/pull/26

itsthisjustin:
> https://github.com/crosspoint-reader/crosspoint-tools/pull/26
> 
> 

Merged

PR #2523, opened 2026-07-02
feat(opds): add optional per-server download folder

ArthurJordao:
## Summary

* **What is the goal of this PR?** Let each OPDS server optionally specify a folder on the SD card where its downloaded books are saved. When unset (the default), books download to the SD root exactly as they do today.

  * `OpdsServer` gains a `downloadFolder` field, persisted in `opds.json` under a new additive `folder` key. Legacy configs without the key load as root. Values are normalized to an absolute path with a single leading `/` and no trailing `/` (empty means root).
  * The on-device OPDS server editor gains a **Download Folder** field that opens a folder picker — a new `PickFolder` mode of `FileBrowserActivity` that lists directories only and returns the chosen folder (or root). It reuses the existing browser (as `PickFirmware` already does) rather than adding a new activity. No folder creation on device — you pick an existing folder or root.
  * The web UI OPDS form gains a **Download Folder** text field, backed by the same `folder` field, so both editors stay in sync.
  * OPDS downloads save into the configured folder (created with parents if missing), falling back to the SD root if the folder can't be created, so a bad config never blocks a download.

## Additional Context

* **Backward compatible.** Existing `opds.json` files and current download behavior are unchanged when no folder is set. No cache/file-format version bump — the download destination doesn't affect the render cache, and the `folder` JSON key is additive.
* **Memory.** No new persistent heap growth or per-render allocations. The picker adds one flash-resident `constexpr` sentinel and a single `vector::insert` at directory-load time; `std::string` use matches the surrounding code.
* **Suggested review focus:** `normalizeFolderPath` and its invariant (`src/util/StringUtils.cpp`), and the `PickFolder` handling in `FileBrowserActivity` (list index 0 is a synthetic "‹ Select this folder ›" entry that returns the current directory; it's special-cased in the loop and the render lambdas).
* **Verification:** `./bin/clang-format-fix`, `pio check --fail-on-defect low --fail-on-defect medium --fail-on-defect high`, and `pio run` all pass locally. Not yet flashed to hardware — on-device testing of the picker UX and across the four orientations would be appreciated.

---

### AI Usage


itsthisjustin:
Closing this in favor of #2571

PR #2524, opened 2026-07-02
perf: light-sleep idle + refresh downclock — ~3.2x active reading time (X3-measured)

BrianPugh:
## Summary

Reduces active-reading power draw ~3.2x, measured end-to-end on an X3 with a Nordic PPK2 at the battery terminals (slim build, cable out, fixed book/settings). Three independent changes:

1. **Light-sleep between input polls when idle** (the big one): after 1 s of inactivity the main loop replaces its busy `delay(50)` with 50 ms timer-paced `esp_light_sleep_start()` slices. Poll cadence and input latency are unchanged — the CPU sleeps between polls instead of spinning at 10 MHz. Guards: never sleeps while a `HalPowerManager::Lock` is held (render in flight), WiFi is up, or USB is connected (light sleep kills the CDC link).
2. **Two-stage idle backoff**: downclock at 500 ms, light sleep at 1 s, replacing the flat 3 s full-speed window. Renders re-raise the clock via `Lock` the moment they start, so the full-speed tail only ever served loop bookkeeping. (The 3 s value in #852 was a placeholder; responsiveness verified on device against stock.)
3. **Refresh BUSY-wait downclock**: drops the CPU to 10 MHz while `pollBusy()` polls the BUSY pin (0.3–2 s per refresh), via new optional hooks in the SDK — **depends on crosspoint-reader/community-sdk#21**. Plus a 1 Hz throttle on the X3's I2C-based USB detection (was polled every loop).

## Measured results (PPK2, X3)

| Metric | Before | After |
|---|---|---|
| Idle on a static page | 9.68 mA | **2.78 mA** |
| Fast page turn (incl. post-turn behavior) | ~100 mC | **28.98 mC** |
| Post-turn 160 MHz tail | 21.2 mA x 3 s | eliminated |
| Session average @ 30 s/page | ~11.6 mA | **~3.6 mA (~3.2x reading time)** |
| Deep sleep | 12.8 µA | unchanged |

Full measurement conditions, scenario suite, and battery-life model: `docs/power-benchmark.md` (added in this PR).

## Hardware hazard worth knowing about

First light-sleep attempt hard-powered-off the device: **GPIO13 is the flash SPIWP pad (unused in DIO mode), rewired to the battery-latch MOSFET gate** — and the prebuilt IDF libs ship `CONFIG_ESP_SLEEP_FLASH_LEAKAGE_WORKAROUND=y`, which pulls unused flash pads low on every light-sleep entry, releasing the latch. The fix drives GPIO13 high with a persistent pad hold, released only in `startDeepSleep()`. Any future light-sleep code must preserve this (documented in `HalPowerManager::lightSleep`).

## Notes for review

- Buttons are ADC resistor-ladder inputs (mid-rail when pressed), invisible to digital GPIO wake — hence timer-paced polling rather than GPIO wake. The 50 ms cadence matches the previous idle `delay(50)` exactly.
- #801's original light-sleep rejection (wake overhead, no ADC wake source) is addressed by measurement: 2.78 mA *includes* all wake overhead.
- Draft until community-sdk#21 merges: the submodule pointer currently references the fork commit, so CI cannot fetch it. Once merged I'll re-bump to the upstream SHA and mark ready.
- Verified on device: stability soaks, all refresh modes + grayscale, power-off/wake, auto-sleep, responsiveness A/B against master.

### DIFF
--- a/docs/power-benchmark.md
+++ b/docs/power-benchmark.md
@@ -74,6 +74,7 @@ off, no RAM/RTC retention), so this breakdown is where the responsiveness wins h
 |----|----------|-------------|------------------|--------------------|----------------------|
 | S1 | Idle static (slim) | | 9.68 | | slim, cable out, 2026-07-02 — awake on a reading page, NOT the sleep screen |
 | S1b | Idle static + Opp 1 light-sleep | | 3.45 | | slim, cable out, 2026-07-02 — 50 ms timer-paced light sleep; ~2.8× vs S1; stable after persistent latch-hold fix (runs: 3.38/3.45/3.47) |
+| S1c | Idle static + USB-poll throttle | | 2.78 | | slim, cable out, 2026-07-02 — BQ27220 USB-detect poll 50 ms → 1 s; tilt off; 3.5× vs S1 |
 | S2 | Deep sleep | | 0.0128 | | 12.77 µA — idling at the sleep-screen (latch off, image is bistable); slim, cable out, 2026-07-02 |
 | S3 | Turn, fast | | | 100.75 | slim, cable out, 2026-07-02 — window included the (since-removed) 160 MHz tail |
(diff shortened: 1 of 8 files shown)


BrianPugh:
sorry, not ready yet

PR #2525, opened 2026-07-02
perf: light-sleep idle + refresh downclock — ~3.2x active reading time (X3-measured)

BrianPugh:
## Summary

Reduces active-reading power draw ~3.2x, measured end-to-end on an X3 with a [Nordic PPK2](https://www.nordicsemi.com/Products/Development-hardware/Power-Profiler-Kit-2) at the battery terminals (slim build, USB cable removed, 3.8v). Three independent changes:

1. **Light-sleep between input polls when idle** (the big one): after 1 s of inactivity the main loop replaces its busy `delay(50)` with 50 ms timer-paced `esp_light_sleep_start()` slices. Poll cadence and input latency are unchanged — the CPU sleeps between polls instead of spinning at 10 MHz. Guards: never sleeps while a `HalPowerManager::Lock` is held (render in flight), WiFi is up, or USB is connected (light sleep kills the CDC link).
2. **Two-stage idle backoff**: downclock at 500 ms, light sleep at 1 s, replacing the flat 3 s full-speed window. Renders re-raise the clock via `Lock` the moment they start, so the full-speed tail only ever served loop bookkeeping. (The 3 s value in #852 was a placeholder; responsiveness verified on device against stock.)
3. **Refresh BUSY-wait downclock**: drops the CPU to 10 MHz while `pollBusy()` polls the BUSY pin (0.3–2 s per refresh), via new optional hooks in the SDK — **depends on Free-Ink/freeink-sdk/pull/8**. Plus a 1 Hz throttle on the X3's I2C-based USB detection (was polled every loop).

(1+2) reduced average static-page-current from 9.68mA to 3.45mA. Reducing the battery-monitor polling frequency further reduced it to 2.78mA.

Before merging, we'll need to update the submodule reference.

## Measured results (PPK2, X3)

| Metric | Before | After |
|---|---|---|
| Idle on a static page | 9.68 mA | **2.78 mA** |
| Fast page turn (incl. post-turn behavior) | ~100 mC | **28.98 mC** |
| Post-turn 160 MHz tail | 21.2 mA x 3 s | eliminated |
| Session average @ 30 s/page | ~11.6 mA | **~3.6 mA (~3.2x reading time)** |
| Deep sleep | 12.8 µA | unchanged |

## Typical-day profiles

Assuming someone spends 30s per page.

- **Light**: 1 h reading, 10 lock/wake cycles, remaining ~23 h deep sleep
- **Heavy**: 2 h reading, 20 lock/wake cycles, remaining ~22 h deep sleep
- Per-day cost components: reading 11.6 → 3.6 mA (session avg), lock+wake cycle
  0.113 → 0.099 mAh each (sleep-screen mode), deep sleep 12.8 µA (unchanged,
  ~0.29 mAh/day)
- Excluded: WiFi sessions, cold chapter indexing, self-discharge

### Days per charge (650mAh)

| Profile | Before | After | Ratio |
|---|---|---|---|
| Light (1 h/day, 10 locks) | 13.0 mAh/day → ~50 days | 4.9 mAh/day → ~133 days | 2.7× |
| Heavy (2 h/day, 20 locks) | 25.7 mAh/day → ~25 days | 9.5 mAh/day → ~68 days | 2.7× |
| Pure active reading | 11.6 mA → ~56 h | 3.6 mA → ~180 h | 3.2× |

### Caveats

- "After" saturates in reality: Li-ion self-discharge (~2–3%/month) is not modeled
  and becomes the limiter once firmware draw is this low.
- Faster page-turners trend toward ~2.8×, slower readers slightly above 3.2×.
- Frequent locking dilutes the ratio (wake = full cold boot, barely touched by the
  power branch): at 10–20 locks/day, cycles are ~20% of the "after" budget.
---

### AI Usage


BrianPugh:
I'll be running on this on my daily driver to see how it performs "out of the lab"

BrianPugh:
we can remove the changes to the debugging-monitor script if we want to make this PR smaller. The changes were just helpful when analyzing waking/boot behavior.

Uri-Tauber:
@BrianPugh Thanks for this PR.
Note we moved from open-x4-sdk to [freeink-sdk](https://github.com/Free-Ink/freeink-sdk).

BrianPugh:
I'll see what/if the equivalent change is over there! I might be able to mess with it tomorrow.

BrianPugh:
Also, if you have anything you want power profiled, I'm open to additional power optimization ideas!

BrianPugh:
@Uri-Tauber updated to freeink-sdk and rebased appropriately! All energy measurements are the same; i improved the polling logic around the debounce logic to make button presses less flakey. Basically poll 10ms after a state-change is detected (instead of waiting a full 50ms more). In an absolute worse case scenario, a user-button-press of 99ms could have been missed with the old logic. Now the button press has to be a minimum of 59ms in absolute worse-cast-scenario.

Uri-Tauber:
@BrianPugh Sound very nice!
I dont think I will have the time to properly review this in the next couple of days.
@itsthisjustin Want to check this?

BrianPugh:
additional note: i haven't tested any of this on an X4 (I don't have one), but i don't foresee any issues. I also haven't tested more complex interactions (e.g. wifi or bluetooth), but i believe they are accounted for.

itsthisjustin:
> additional note: i haven't tested any of this on an X4 (I don't have one), but i don't foresee any issues. I also haven't tested more complex interactions (e.g. wifi or bluetooth), but i believe they are accounted for.

I'll test on all of these

BrianPugh:
I just pushed an additional change (5fa9585) that saves around 3mC of energy during page turns. This is an additional 10% improvement over our previous best.

This would push the 1-hour-a-day reader from 133 days of battery life to 135 days. A minor savings. I think most future power-savings improvements just comes from algorithmic efficiency (e.g. faster anti-aliasing computations so that the device can get back to sleep quicker. Or faster boots, etc).

BrianPugh:
addressed coderabbitai feedback in the form of documentation in [be0409d](https://github.com/crosspoint-reader/crosspoint-reader/pull/2525/commits/be0409d421e8a3ebb07314671ec35047c2fcd8fe)

BrianPugh:
running the ADC polling at 160mhz instead of 10mhz dropped idle page average current 2.78mA -> 2.57mA

BrianPugh:
I just updated the free-ink submodule to official now that https://github.com/Free-Ink/freeink-sdk/pull/8 has been merged in.

BrianPugh:
Just an update, I've been using this on my daily driver for a week, and am still at 100% battery on my X3. Not sure if this speaks about how much I read, the efficiency of this PR, or the accurate of the battery percentage 😆 .

KnicKnic:
@BrianPugh
I was leveraging crosspoint for a bluetooth display project and had some similar observations as you. First I am curious around a few points.

I guessed idle perf simliliar to what you measured, great to see someone measure it. https://github.com/crosspoint-reader/crosspoint-reader/discussions/2468#discussioncomment-17600939

### GPIO wakeup
Have you considered sleeping longer and using the gpio pins to wake up? 

* On my X3, all buttons on right half of device to wake up gpio reliably. I then go into a polling mode for 10 seconds to allow reading buttons on left half of device. 
    * You have to be careful to not do an ADC read on them when in this mode though. Easy for manual light sleep, more annoying for auto light sleep
    * I know this may not be great for all users, but an option may be there

###  Auto Power Management
Have you considered **auto light sleep or auto power management** - https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-reference/system/power_management.html?
* Necessary for sleeping if bluetooth is there.
* You will need to protect all hardware calls.
    * I feel that this is a much easier approach of wrapping no light sleep, APB and MAX cpu in the appropriate places with power management locks. 
    * before freeink sdk some completion to hardware calls were missing. (no light sleep when writing to eink)
* I think this will allow every component to get what it needs and provide less coupling
* You have to be careful with UART I turn off serial & USB with auto sleep. 
    * I have a system option, and it can control sleep as well. Something like this may be useful in builds that allow logs.

### Bluetooth Power optimization
Now I found an annoying issue with ble & auto sleep that the default clock for this project causes it to add a 7% buffer to wakeup times meaning you can't achieve more than 93% idle if you have auto light sleep & bluetooth. 

* standard clock is untrustworthy https://github.com/espressif/esp-idf/blob/v5.5.2/components/bt/controller/esp32c3/bt.c#L1134-L1160
* To get around this I used the `CONFIG_RTC_CLK_SRC_INT_8MD256=y` https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-reference/system/system_time.html#rtc-clock-source-choice
* Did you consider using this clock source for more accurate wakeups?

This doesn't have any measured battery changes, due to the chip just having its power cut in deep sleep right?


### Deep sleep
You mentioned power draw during deep sleep. You are measuring isn't capable of being woken up from via timer is it, its more of just an off state right? 
* I assume above option doesnt hurt the deep sleep due to this, **but I would be curious if you have the capability of measuring if you could let me know**.

BrianPugh:
Hey @KnicKnic ! Let me see if I can answer your questions one-by-one.

---

> Have you considered sleeping longer and using the gpio pins to wake up?

So there's really two things here: the polling time, and (what i'm assuming you're referencing) using some of the buttons as an interrupt.

For reference, during idle (when we enter light-sleep), the power-floor is around 2.09mA. The current 20hz (50ms period) ADC polling interval results in an average current draw of 2.57mA. We could probably go as low as 10hz (100ms period) for an assumed average current draw of 2.33mA (~10% saving).

I didn't test it, but in theory (as you mentioned), the right/down (and power button) has enough voltage to serve as a digital interrupt. I didn't pursue it because it would offer an inconsistent experience or a complicated-to-a-user configuration-option (even if "next page" is probably the most common action). Presumably, this could reduce the average idle current from 2.59mA to basically 2.09mA (~20% saving). In this PR, I didn't really want to compromise the user-experience in anyway for the sake of even more ridiculous battery life.

---

> Have you considered auto light sleep or auto power management

Caveat: I'm actually not too familiar with esp32 programming with arduino/pio, all my experience is with directly using esp-idf. I briefly considered it, but it's my understanding that the precompiled esp32-arduino-libs don't have `CONFIG_PM_ENABLE=y`, so we cannot use it in this project. I figured that's probably a deep rabbit hole to go down, and that we can probably get 90% of the savings with a few manual power-management actions (this PR). If arduino can be successfully compiled with power-management and all the related items (and they work in the framework without issue), then I think that would definitely be the better route. Please let me know if this is a viable option!

 ---

> Bluetooth Power optimization

Sorry, I don't really understand the question in this section 😅 . I also didn't test bluetooth/wifi in this PR, what situation should i run? Are you mentioning timing accuracy for energy-saving reasons, or something else?

---

> You mentioned power draw during deep sleep. You are measuring isn't capable of being woken up from via timer is it, its more of just an off state right?

Correct, I think/assume how this device works is that there's a MOSFET that feeds power for all the major components (including the esp32-c3 main processor). The power button bypasses the mosfet and supplies power to the esp32-c3, which then itself keeps the mosfet on. When going into "deep sleep", the esp32 just stops keeping that mosfet on and cuts power to itself. So, in "deepsleep" the esp32-c3 literally doesn't have power going to it (it's not in a microcontroller deepsleep). It's physically powered off. So naturally no timer is powered/running and we cannot wake from this state from software.

I believe the esp32-c3 only consumes 5-40uA in "real" deepsleep, but I think the reason the board designers went with the mosfet route is to also reduce the parasitic power consumption of all the other chips on the board.

KnicKnic:
> Hey @KnicKnic ! Let me see if I can answer your questions one-by-one.
> 
> > Have you considered sleeping longer and using the gpio pins to wake up?
> 
> So there's really two things here: the polling time, and (what i'm assuming you're referencing) using some of the buttons as an interrupt.
> 
> For reference, during idle (when we enter light-sleep), the power-floor is around 2.09mA. The current 20hz (50ms period) ADC polling interval results in an average current draw of 2.57mA. We could probably go as low as 10hz (100ms period) for an assumed average current draw of 2.33mA (~10% saving).
> 
> I didn't test it, but in theory (as you mentioned), the right/down (and power button) has enough voltage to serve as a digital interrupt. I didn't pursue it because it would offer an inconsistent experience or a complicated-to-a-user configuration-option (even if "next page" is probably the most common action). Presumably, this could reduce the average idle current from 2.59mA to basically 2.09mA (~20% saving). In this PR, I didn't really want to compromise the user-experience in anyway for the sake of even more ridiculous battery life.

Fair enough! The reason I am doing it is when mixed with auto sleep for bluetooth I want a way to still have a responsive device and wake it up out of the auto sleep(although now that I think about it longer I guess I could just use delay with auto light sleep).


> 
> > Have you considered auto light sleep or auto power management
> 
> Caveat: I'm actually not too familiar with esp32 programming with arduino/pio, all my experience is with directly using esp-idf. I briefly considered it, but it's my understanding that the precompiled esp32-arduino-libs don't have `CONFIG_PM_ENABLE=y`, so we cannot use it in this project. I figured that's probably a deep rabbit hole to go down, and that we can probably get 90% of the savings with a few manual power-management actions (this PR). If arduino can be successfully compiled with power-management and all the related items (and they work in the framework without issue), then I think that would definitely be the better route. Please let me know if this is a viable option!

### Maintainer ask on above
@itsthisjustin you or another maintainer need to make the judgement call on this one. (is it acceptable to do `CONFIG_PM_ENABLE=y`)

Personally I think it makes the freeink sdk more valuable for people trying to do bluetooth & wifi projects and eek out more powersavings if the support is put in. As then they could enable auto light sleep. I started an issue about it here to find out if the project wants to consider it - https://github.com/Free-Ink/freeink-sdk/issues/11

Otherwise they are stuck not doing any light sleep, but light sleep is really useful for bluetooth. For example I left my project on today, and my X3 says I now have 98% battery in 12 hours of maintaining bluetooth connectivity because it is in light sleep 99% of the time. (even if the 98% battery is over optimistic by a few percent, I think my goal of lasting a work week has been achieved!)

_As for how hard is this... well I got it working! & I havent done c++/c coding in 5+ years, and not really ever worked on devices like this till playing around with crosspoint. Well AI knew to do it & did it for me, so I have no clue how hard it is 😊.  AI also figured out how to add sleep profiling and lock timing and so many cool features._

> 
> > Bluetooth Power optimization
> 
> Sorry, I don't really understand the question in this section 😅 . I also didn't test bluetooth/wifi in this PR, what situation should i run? Are you mentioning timing accuracy for energy-saving reasons, or something else?

No need for measurement

Mainly I am curious if `CONFIG_RTC_CLK_SRC_INT_8MD256=y` changes your measurements. If you eventually start to play with the flags I would be curious if it hurts any of them. The docs say it costs (5 μA) in deep sleep. My guess based on your answer with the mosfet is that it doesn't change anything. I was also curious how much -   `CONFIG_BT_CTRL_MAIN_XTAL_PU_DURING_LIGHT_SLEEP=y` costs in light sleep.

However I realized that unless you are actually doing a bluetooth project measuring usage wont work. And since I left my project on today I got a good enough measurement. So ignore any ask.

> 
> > You mentioned power draw during deep sleep. You are measuring isn't capable of being woken up from via timer is it, its more of just an off state right?
> 
> Correct, I think/assume how this device works is that there's a MOSFET that feeds power for all the major components (including the esp32-c3 main processor). The power button bypasses the mosfet and supplies power to the esp32-c3, which then itself keeps the mosfet on. When going into "deep sleep", the esp32 just stops keeping that mosfet on and cuts power to itself. So, in "deepsleep" the esp32-c3 literally doesn't have power going to it (it's not in a microcontroller deepsleep). It's physically powered off. So naturally no timer is powered/running and we cannot wake from this state from software.
> 
> I believe the esp32-c3 only consumes 5-40uA in "real" deepsleep, but I think the reason the board designers went with the mosfet route is to also reduce the parasitic power consumption of all the other chips on the board.

The only thing I think that can change is if you enable rtc or not right?

Thanks for taking the time to have responded to me!

itsthisjustin:
@BrianPugh i'm curious if you can test this on latest freeink sdk which bumps our SPI speed up and how this effects stuff like that

BrianPugh:
With latest freeink-sdk (9d4efea2d033) the key power measurements are:

* page-turn 26.00mC
* idle average power: 2.79mA
* idle power floor: 2.20mA

So about all as expected. I had to (unrelated) revert some of the changes that dropped the idle average power down to 2.38mA because it caused some lockups with some unrelated optimizations (i'll open up those PRs once this is merged).

I haven't done a soak test, but i imagine everything should still work just as well. Flashing to my daily driver now. All looks good to me!

diogo7dias:
Tested this on an **X4** and hit a reproducible input regression: short, fast repeated button presses get dropped.

I run a fork, so treat the surrounding code as mine, not yours. But the slice hook itself is unmodified from this PR — the only edit inside `onEinkBusyWaitSlice()` is that my battery-latch lines are folded into a helper, which on an Xteink board makes exactly the same three calls. Slice length, wake sources and the yield are yours.

### What I measured

Three builds, same X4, on battery (not USB, so light sleep is actually active):

| Build | Fast repeated taps |
|---|---|
| fork without this PR | 8 fast taps, all 8 registered — in a book and in menus |
| fork **with** this PR | roughly 2 of every 3 registered — in a book and in menus |
| fork with this PR, **`setBusyWaitSliceHook` not installed** | indistinguishable from the build without the PR |

The third build differs from the second by one line: I do not install the slice hook. `setBusyWaitHooks` (the downclock) and the idle `lightSleep()` path are both left fully active. So the idle work looks fine; it is specifically the in-refresh slice.

### Why I think it happens

`esp_light_sleep_start()` inside `onEinkBusyWaitSlice()` halts the whole chip, not just the render task, for up to `BUSY_SLEEP_SLICE_MS` (20 ms). After it returns, `vTaskDelay(1)` yields one tick — 1 ms at `CONFIG_FREERTOS_HZ=1000` — before the render task takes the next slice. So during a refresh the main loop samples buttons roughly once every 21 ms.

The slice arms a GPIO wake for the power button only. As your own comment in `lightSleep()` notes, the front and side buttons are ADC resistor-ladder inputs whose pressed level sits mid-rail and is invisible to a digital GPIO wake, so they can only be found by polling. A tap that begins and ends inside a single 20 ms slice is therefore never sampled at all.

That also explains why menus are affected as much as page turns: every screen update is a refresh.

I suspect this did not show up on X3 testing simply because it needs taps fast enough to fall inside a slice, and a page-turn button rather than the power button.

### Possible directions

I did not want to guess at a fix, since you have the power meter and I do not:

- Sample the ADC ladder across the slice — e.g. let the slice hook run `inputMgr.update()` itself before sleeping again, so presses are seen at the render task's cadence rather than the main loop's.
- Shorten `BUSY_SLEEP_SLICE_MS` enough that no realistic tap fits inside one, and take the smaller saving.
- Yield more than one tick per slice, so the main loop reliably gets a poll in between.

Happy to build and test any of these on the X4 and report back — including a power comparison if you tell me what you want measured. And thanks for the PR; the idle saving alone is a big improvement and it is the main reason I picked it up.

BrianPugh:
Hmmm interesting; thanks for reporting! I'm certainly able to reproduce; if I do 8 rapid clicks, typically only 3 get registered. You also provide good evidence/direction as to the cause. Most button debounce algorithms use a 20-50ms delay, so I don't think a button press falling fully within a 20ms window is the cause. I bet there's some adjacent button-sampling bug that this introduced. Since I'm able to reproduce, I think I should be able to get a fix in sometime within the next 2 days.

BrianPugh:
@diogo7dias I think this should be fixed in [05129e2](https://github.com/crosspoint-reader/crosspoint-reader/pull/2525/commits/05129e2fc04264660a2befb5b214fe7b683d672a) + [148ac36](https://github.com/crosspoint-reader/crosspoint-reader/pull/2525/commits/148ac36a92fde86ccc5145cd5f15678c3797a665); the fix has no measurable negative impact on average energy consumption during a page turn.

BrianPugh:
I also just brought this PR up to date with `develop`; power readings are mostly the same.

- 2.93mA idle-on-page.
- 30.24mC per page turn.

diogo7dias:
Confirmed fixed on hardware. Thank you for the quick turnaround.

I built `05129e2` + `148ac36` into my fork and flashed an Xteink X4. Repeating the same test as before:

- Eight rapid taps in a book: eight page turns.
- Eight rapid taps in a menu: eight moves.

Both were dropping roughly one press in three before these two commits, and both now feel identical to the stable build without light sleep. Power-off still work correctly on battery.

The only thing I have not measured yet is the battery win itself over normal reading, which needs a few days of ordinary use rather than a bench test. I will follow up if anything looks off.

BrianPugh:
Hey @itsthisjustin! Whats the priority-level/timeline for getting this merged in? Thanks!

itsthisjustin:
> Hey @itsthisjustin! Whats the priority-level/timeline for getting this merged in? Thanks!

I'm thinking we do this as a fast follow up to 1.5 as 1.5.1 along with my touch ui and x4pro support. Ideally we also fix the x3 display variants

BrianPugh:
sounds good! just let me know if you need anything from me!

itsthisjustin:
@BrianPugh i think we are gonna test this and get it merged soon if all is well. can you fix the conflict with the sdk? Also confirm we haven't regressed in the sdk in this area (don't think we have)

BrianPugh:
This branch is now up to date, however I think there's an odd interaction between the background section-indexing job and the power management; it seems like our CPU is pegged at full power until the device goes to sleep. After awaking from sleep, power-management is working as intended. Investigating now.

itsthisjustin:
> This branch is now up to date, however I think there's an odd interaction between the background section-indexing job and the power management; it seems like our CPU is pegged at full power until the device goes to sleep. After awaking from sleep, power-management is working as intended. Investigating now.

Could explain the complaints users had around worse battery life on the last update! Please fix without hurting smart indexing if you can cause that would be a big win

BrianPugh:
I think b38e973 fixes that issue (I'm not seeing it anymore). I bet it could be cherry-picked onto `develop` if you want to be more surgical. When the bug was triggered, the device was consuming around 31.13mA (on a static idle page), which is more than 3x more power than on 1.4.1, so thats probably what users were reporting.

Update number reports (this branch):
* Page turns take around 27.00mC-33.00mC
* Page idle is at around 2.82mA-2.93mA

BrianPugh:
@itsthisjustin I made #2996 for convenience if you want the surgical solution to immediately get v1.5's power consumption in check.

itsthisjustin:
> @itsthisjustin I made #2996 for convenience if you want the surgical solution to immediately get v1.5's power consumption in check.

Amazing thanks. I'll check it and merge asap. Any idea why that specifically caused the issue? (nvm nice explanation)

itsthisjustin (approved):

itsthisjustin:
FYI this breaks x4pro builds as gpio13 is used for something else there. I'm going to do a quick follow up patch

itsthisjustin:
@BrianPugh i'm having issues with this PR (I think). If I plug in an x4 while it's on, it immediately freezes up. If i boot with usb in already, it's fine but as soon as I plug it back in, entire device is dead

BrianPugh:
I'll investigate tonight

itsthisjustin:
> I'll investigate tonight

Thanks dude! I've reverted for now but kept your battery drain fix. If you could also double check that's good I'd appreciate it

BrianPugh:
Opened #3082 and responded to #3077

PR #2418, opened 2026-06-24
feat: Adds Bluetooth page turner support (not to be merged)

itsthisjustin:
Integrate BLE keyboard host functionality for page-turner remotes. Adds device pairing, button mapping, and persistent storage of mappings. Migrates from open-x4-sdk to freeink-sdk submodule which includes the BleKeyboardHost library. Implements CPU frequency locking during BLE operations to prevent watchdog timeouts, tears down bluetooth on chapter index to save on heap, and turns off bluetooth in any wifi activity or any non-reader activity.

PR #2526, opened 2026-07-02
fix: FontDecompressor OOM aborts on the render path (-fno-exceptions makes vector resize fatal)

k5njm:
## Summary

* **What is the goal of this PR?** Fix a class of OOM `abort()` crashes on the render path. `FontDecompressor::hotGroup` / `hotGlyphBuf` were `std::vector<uint8_t>` resized inside `getBitmap()` — the render path. Under `-fno-exceptions`, a resize that hits OOM throws `std::bad_alloc` with no handler → `std::terminate()` → `abort()`. The existing `if (hotGroup.empty())` "graceful failure" checks were unreachable: the abort happens inside `resize()` before they run.

* **What changes are included?** Both buffers converted to grow-only nothrow `malloc` buffers behind a shared `ensureCapacity()` helper (free-then-malloc, since callers fully rewrite the buffer after a grow — no point holding both blocks on a tight heap). On OOM, `getBitmap()` returns `nullptr`; both `renderCharImpl` call sites in GfxRenderer already skip the glyph on a null bitmap, so the worst case is a page with a missing glyph instead of a reboot. This matches the heap-discipline skill's rule ("Bare `new`/`new[]` is never correct here") — the vector just hid the `new`. The `malloc`-based `tempBuf` in `prewarmCache()` twenty lines away already does this correctly; this brings the hot-group path in line with it. One detail: the cached-group identity (`hotGroupFont`/`hotGroupIndex`) is now invalidated *before* the grow, so a failed allocation can't leave a stale "group N loaded" marker pointing at a freed buffer.

### Field crash (X4, 2026-07-02)

Reading with the BLE page-turner branch (#2418) resident, heap ~11 KB free after prewarm degraded gracefully. Symbolized stack from the on-device crash report:

```
abort() ← __terminate ← operator new
       ← std::__new_allocator<uint8_t>::allocate
       ← renderCharImpl (GfxRenderer.cpp:249 → getGlyphBitmap → FontDecompressor::getBitmap)
       ← GfxRenderer::drawText ← TextBlock::render ← EpubReaderActivity::renderContents
```

Last logs before the abort:
```
[FDC] Failed to allocate temp buffer (8480 bytes) for group 0   ← prewarm failing cleanly
[FDC] Prewarm: 55 glyphs in 0 bytes from 3 groups (3 missed)    ← forces hot-group fallback
→ abort() in hotGroup.resize()
```

Reproducible with a glyph-heavy EPUB ("Fundamentals of Data Engineering", many font groups + multiple styles per page) on any build where something else (NimBLE, WiFi) holds ~60 KB. Happy to share the crash report + book.

### Verification

Tested on X4 hardware: verified the same book renders through low-heap episodes (glyphs degrade, no abort). No behavior change when memory is fine.

### Related

- #2527 is the broader heap-resilience PR for BLE + reader coexistence; this PR fixes the same `-fno-exceptions` abort class inside the render path and is independent of it, but both are needed for the full story.

---

### AI Usage

Did you use AI tools to help write this code? **< YES >**

The code changes in this PR were written with AI assistance. Hardware testing and crash reproduction on the X4 were done manually by the author.

### DIFF
--- a/lib/EpdFont/FontDecompressor.cpp
+++ b/lib/EpdFont/FontDecompressor.cpp
@@ -33,12 +33,24 @@ void FontDecompressor::freePageBuffer() {
 }
 
 void FontDecompressor::freeHotGroup() {
-  hotGroup.clear();
-  hotGroup.shrink_to_fit();
+  free(hotGroup);
+  hotGroup = nullptr;
+  hotGroupCapacity = 0;
   hotGroupFont = nullptr;
   hotGroupIndex = UINT16_MAX;
-  hotGlyphBuf.clear();
-  hotGlyphBuf.shrink_to_fit();
+  free(hotGlyphBuf);
+  hotGlyphBuf = nullptr;
+  hotGlyphBufCapacity = 0;
+}
+
+bool FontDecompressor::ensureCapacity(uint8_t*& buf, uint32_t& capacity, uint32_t needed) {
+  if (capacity >= needed) return true;
+  // Grow-only, free-then-malloc: every caller fully rewrites the buffer after a grow, so the
+  // old contents are dead -- freeing first gives the allocator its best shot on a tight heap.
+  free(buf);
+  buf = static_cast<uint8_t*>(malloc(needed));  // owned by FontDecompressor, freed in freeHotGroup()
+  capacity = buf ? needed : 0;
+  return buf != nullptr;
 }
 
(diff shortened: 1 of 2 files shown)


itsthisjustin (approved):

itsthisjustin:
Found this same issue working the BLE branch. This was a pre-existing bug and this does in fact fix it

Uri-Tauber (approved):

Issue #2499, opened 2026-07-01
Feature Request: Automatically Open the Next Book in the Same Folder


chohoyeon-ops:
I have a small UX suggestion regarding reading multi-volume books.

At first I considered suggesting automatic series detection based on filenames, but I think that approach would be difficult to implement reliably.

People organize their books in many different ways, for example:

* 1, 2, 3
* 01, 02, 03
* Book 1, Book 2
* Vol.1, Vol.2
* Episode 01
* 上・中・下
* I, II, III
* Prologue, Main, Epilogue

Supporting every naming convention would require a lot of parsing logic and still wouldn't cover every case.

Instead, I think a much simpler solution would greatly improve the reading experience.

### Suggested behavior

Assume that books belonging to the same series are stored in the same folder.

When the reader reaches the **End of Book** screen, pressing the **Next Page** button could automatically open the next supported file (EPUB/TXT) in that folder, using the current file sorting order (alphabetical, which CrossPoint already uses in the file browser).

For example:

```
Series/
    01.epub
    02.epub
    03.epub
```

After finishing `01.epub`, pressing **Next Page** on the End of Book screen would immediately open `02.epub`.

If there is no next supported file, the reader would simply behave as it does now.

This avoids complicated filename parsing while making reading long series much more convenient.

Reading physical books naturally continues from one volume to the next. I think a file-based reader can offer a similar experience without implementing complex series detection.

brianhuster:
Epub has "belongs-to-collection", "group-position" for indexing books in the same series, franchise, etc

Uri-Tauber:
I think a good UX would be to have a list of three alphabetically-nearest books and 'return home' option.

@brianhuster's idea seems to me like over-engineering for our hardware.

tomlarse:
I'd like to take a stab at implementing this, along the lines @Uri-Tauber suggested. Here's what I have in mind:

No series detection at all. Candidates are just the supported files in the same folder, in the file browser's existing sort order, taking up to 3 entries after the current file. That sort is already a natural sort (numeric-aware, case-insensitive), so `1/2/10`, `01/02`, `Book 1/Book 2` and `Vol.1/Vol.2` all order correctly without any filename parsing.

Add an "End of book" setting with three values: `Home` / `Next book` / `Ask`.

`Home` (default): exactly what happens today, Next Page goes home. Updating changes nothing unless you opt in.

`Next book`: the end screen shows a hint, and Next Page opens the file directly.

[...]

PR #2532, opened 2026-07-03
feat(reader): End of Book next-book suggestions (#2499)

tomlarse:
## Summary

* **What is the goal of this PR?** Implements #2499: after finishing a book, the reader can continue straight into the next book in the same folder from the End of Book screen.
  * When the folder contains books that follow the current one, the End of Book screen shows a small menu with up to 3 of them plus a Home entry. When it doesn't, the screen looks and behaves exactly as it does today.
  * There is no series detection or filename parsing, as discussed in the issue. Candidates are the supported files (epub/xtc/txt/md) that follow the current file in the folder, using the file browser's natural sort. `1/2/10`, `01/02` and `Vol.1/Vol.2` all order correctly.
  * New `NextBookFinder` helper: a single directory pass that keeps only the 3 best candidates, so memory use stays flat no matter how many files are in the folder. A shared `EndOfBookOptions` drives the end screen for both the EPUB and XTC readers. The TXT reader has no end screen and is unchanged.
  * `FsHelpers::naturalLess` is extracted from `sortFileList` for reuse. While in there I fixed the ctype calls to cast to unsigned char first; passing raw chars was undefined behavior for non-ASCII filenames.

Buttons on the menu: side Up/Down or front Left/Right move the selection and wrap around, Confirm opens it, « Back returns to the last page of the book, long-press Back goes to the file browser. Home is an entry in the list.

## Additional Context

* The design was discussed in [this comment](https://github.com/crosspoint-reader/crosspoint-reader/issues/2499#issuecomment-4867269198) on the issue. Per review feedback there is no setting: the menu is the behavior, and folders without a next book keep the current end screen.
* Suggestions load once, on the render task, when the end screen first renders. The main task reads the list only after an atomic flag is published, and the final page turn doesn't get any slower.
* Memory: at most 4 short filename strings live at once, plus the same 500-byte name buffer pattern the file browser uses during the scan. I verified on device over serial that free heap stays around 130–149 KB while hopping through books via the menu, with no downward trend in min-free or largest free block.
* Tested on an X4: menu navigation and wrap-around, natural sort ordering (Book 2 → Book 3 → Book 10), UTF-8 filenames, hidden files, books in the SD root, epub→xtc→txt chains, all four orientations, the "Move Finished Books to Read Folder" and "Clear Read Books from Recent List" interactions, auto-sleep and auto page turn. Retested after the review changes.
* Skipping already-finished books is out of scope for now. A progress file existing only proves a book was opened, not finished, so doing this properly needs a real finished check. Noted in the issue as a possible follow-up.

---

### AI Usage


src/activities/reader/EndOfBookOptions.cpp
```diff
@@ -0,0 +1,168 @@
...
+  selector = 0;
+  // Release-publish so the main task, which gates all access on isLoaded, never
+  // observes a partially built list
+  isLoaded.store(true, std::memory_order_release);
+}
+
+bool EndOfBookOptions::askMenuActive() const {
+  return SETTINGS.endOfBookBehavior == CrossPointSettings::EOB_ASK && isLoaded.load(std::memory_order_acquire) &&
+         !names.empty();
+}
```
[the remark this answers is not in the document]
tomlarse: This is intentional. When there are no next-book candidates, the End of Book screen keeps today's behavior: next page goes home. That's how it was proposed in the design discussion on #2499. A menu whose only entry is Home would add a button press to an action the next-page button already performs.

src/activities/reader/EpubReaderActivity.cpp
```diff
@@ -298,33 +321,43 @@ void EpubReaderActivity::loop() {
...
+        return;
+      case EndOfBookOptions::Action::GoHome:
+        onGoHome();
+        return;
+      case EndOfBookOptions::Action::LastPage:
+        currentSpineIndex = epub->getSpineItemsCount() - 1;
+        nextPageNumber = 0;
+        pendingPageJump = std::numeric_limits<uint16_t>::max();
+        requestUpdate();
+        return;
```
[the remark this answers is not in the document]
tomlarse: Added the guard in bbd9a9cd, matching the one the XTC reader already had. Worth noting the case couldn't actually fire: this branch is gated on `atEndOfBook`, which requires `currentSpineIndex > 0`, so a zero-spine book never reaches it. The guard is there for consistency and defense in depth.

src/util/NextBookFinder.cpp
```diff
@@ -0,0 +1,83 @@
+#include "NextBookFinder.h"
+
+#include <FsHelpers.h>
+#include <HalStorage.h>
+#include <Logging.h>
+#include <Memory.h>
+
+#include <algorithm>
+
+#include "CrossPointSettings.h"
```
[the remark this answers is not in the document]
tomlarse: Fixed in bbd9a9cd, `<string_view>` is now included directly.

src/util/NextBookFinder.cpp
```diff
@@ -0,0 +1,84 @@
...
+  if (!dir || !dir.isDirectory()) {
+    LOG_ERR("NBF", "Cannot open folder: %s", folder.c_str());
+    return result;
+  }
+  dir.rewindDirectory();
+
+  const auto nameBuffer = makeUniqueNoThrow<char[]>(NAME_BUFFER_SIZE);
+  if (!nameBuffer) {
+    LOG_ERR("NBF", "OOM: %d bytes", static_cast<int>(NAME_BUFFER_SIZE));
+    return result;
```
Uri-Tauber, line 45: If this allocation fails, the function return without closing `dir`.
```suggestion
  if (!nameBuffer) {
    LOG_ERR("NBF", "OOM: %d bytes", static_cast<int>(NAME_BUFFER_SIZE));
    dir.close();
    return result;
```

src/activities/reader/EndOfBookOptions.cpp
```diff
@@ -0,0 +1,168 @@
...
+#include "util/NextBookFinder.h"
+
+namespace {
+// Vertical layout of the end screen. The plain title position matches the historical
+// end-of-book screen; the Ask layout starts higher so the list fits in landscape too.
+constexpr int TITLE_Y = 300;
+constexpr int ASK_TITLE_Y = 120;
+constexpr int ASK_SUBTITLE_Y = 170;
+constexpr int ASK_LIST_TOP = 200;
+constexpr int NEXT_HINT_Y = 345;
```
Uri-Tauber: I would prefer those numbers to be derived from the screen size, rather than hard-coded. 
I can look bad if we will support devices with different resolution (e.g. 960*540).

Uri-Tauber:
Small bug: on `endOfBookOptions` the two side buttons act as Select/Back instead of Up/Down.

Also, from UX perspective, I don't think we need to have special setting for this. Having `Ask` as the only and default option seems simpler and cleaner.

Apart from those, great job! Really liked this QoL addition.

tomlarse:
Thanks for the review! All addressed in 9fea14b3:

- Side buttons now move the selection like every other list (front Left/Right still work too). The Select/Back roles were a deliberate choice on my part, but I agree the list convention is the right call here.
- Setting removed. The menu shows whenever there are next books in the folder, and when there aren't, the end screen looks and behaves exactly as before.
- `dir.close()` added on the OOM path, and the end screen layout now derives from the safe-area height and font line heights instead of hardcoded positions.

I retested on device after the changes and ran cppcheck locally, which came back clean.

src/util/NextBookFinder.cpp
tomlarse: Fixed in 9fea14b3.

src/activities/reader/EndOfBookOptions.cpp
tomlarse: Fixed in 9fea14b3.

Uri-Tauber:
@tomlarse Sounds good.
Can you solve conflicts?

tomlarse:
@Uri-Tauber ok, thanks! Done!

Uri-Tauber (approved):
LGTM

itsthisjustin:
So is this skipped if I just use books in the root of the sd?

PR #2534, opened 2026-07-04
feat: add options to remember web upload settings & rename ebooks to `{title} - {author}`

victor141516:
## Summary

- Adds a `Remember Settings` toggle for the web EPUB upload options and persists the selected optimization settings in browser `localStorage` only when enabled.
- Adds `Rename from Book Metadata`, which reads EPUB OPF metadata before upload and renames the browser-side `File` to `Title - Author.epub` when possible, or `Title.epub` when no author is present.

## Additional Context

Nope

---

### AI Usage

Yes, but I reviewed and tested the changes it before sending the PR

### DIFF
--- a/src/network/CrossPointWebServer.cpp
+++ b/src/network/CrossPointWebServer.cpp
@@ -662,17 +662,15 @@ void CrossPointWebServer::handleUpload(UploadState& state) const {
     LOG_DBG("WEB", "[UPLOAD] START: %s to path: %s", state.fileName.c_str(), state.path.c_str());
     LOG_DBG("WEB", "[UPLOAD] Free heap: %d bytes", ESP.getFreeHeap());
 
-    // Create file path
     String filePath = state.path;
     if (!filePath.endsWith("/")) filePath += "/";
     filePath += state.fileName;
 
-    // Check if file already exists - SD operations can be slow
     esp_task_wdt_reset();
     if (Storage.exists(filePath.c_str())) {
-      LOG_DBG("WEB", "[UPLOAD] Overwriting existing file: %s", filePath.c_str());
-      esp_task_wdt_reset();
-      Storage.remove(filePath.c_str());
+      state.error = "File already exists: " + state.fileName;
+      LOG_DBG("WEB", "[UPLOAD] Collision: %s", filePath.c_str());
+      return;
     }
 
     // Open file for writing - this can be slow due to FAT cluster allocation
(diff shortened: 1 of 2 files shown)


itsthisjustin:
@victor141516 can you take a look at the coderabbit suggestions?

victor141516:
> @victor141516 can you take a look at the coderabbit suggestions?

@itsthisjustin I reviewed both CodeRabbit suggestions and pushed fixes for them:

1. Creator selection now ignores explicitly non-author roles and only falls back to a creator without role metadata.

2. For duplicate uploads, I added a suffixing for colliding files. It works also when this new "Rename from Book Metadata" feature is disabled. The backend also rejects any remaining collision rather than silently overwriting the existing file. Would you be okay with applying this behavior to all uploads? If you would prefer, I can limit it to renamed files.

itsthisjustin:
> > @victor141516 can you take a look at the coderabbit suggestions?
> 
> @itsthisjustin I reviewed both CodeRabbit suggestions and pushed fixes for them:
> 
> 1. Creator selection now ignores explicitly non-author roles and only falls back to a creator without role metadata.
> 2. For duplicate uploads, I added a suffixing for colliding files. It works also when this new "Rename from Book Metadata" feature is disabled. The backend also rejects any remaining collision rather than silently overwriting the existing file. Would you be okay with applying this behavior to all uploads? If you would prefer, I can limit it to renamed files.

Any scenario where parsing would fully fail? Does it just fallback to file name?

victor141516:
> > > @victor141516 can you take a look at the coderabbit suggestions?
> > 
> > 
> > @itsthisjustin I reviewed both CodeRabbit suggestions and pushed fixes for them:
> > 
> > 1. Creator selection now ignores explicitly non-author roles and only falls back to a creator without role metadata.
> > 2. For duplicate uploads, I added a suffixing for colliding files. It works also when this new "Rename from Book Metadata" feature is disabled. The backend also rejects any remaining collision rather than silently overwriting the existing file. Would you be okay with applying this behavior to all uploads? If you would prefer, I can limit it to renamed files.
> 
> Any scenario where parsing would fully fail? Does it just fallback to file name?

Yes. Parsing can fully fail if the EPUB cannot be opened as a ZIP, its OPF file cannot be found or read, the OPF XML is invalid, or no usable title is present. In those cases, it falls back to the original file and filename. 

The collision handling runs afterward regardless, so it will add a suffix to either the metadata generated filename, or the original filename (if necessary)

itsthisjustin (approved):

PR #2398, opened 2026-06-22
feat: add Arabic text shaping (contextual joining)

badr42:
Implements lightweight Arabic character shaping so letters render with correct connected forms (initial/medial/final/isolated) instead of as disconnected isolated glyphs.

- lib/ArabicShaper/: new library that applies UAX#9 joining rules and replaces base Arabic codepoints (U+0600-U+06FF) with Unicode Presentation Forms (U+FE70-U+FEFF for standard Arabic, U+FB50-U+FBFF for Persian/Urdu/Pashto). Handles the mandatory lam-alef ligature and treats harakat as transparent.

- lib/MiniBidi/bidiclasses.t: add Arabic block (AL/NSM/AN) and Arabic Presentation Forms (AL) so do_bidi() reorders shaped glyphs correctly.

- lib/GfxRenderer/GfxRenderer.cpp: call ArabicShaper::shape() inside resolveVisualText() before BiDi reordering.

- lib/EpdFont/scripts/fontconvert_sdcard.py: add "arabic" interval preset covering U+0600-U+06FF + Presentation Forms. Required when converting any Arabic font.

- lib/EpdFont/scripts/sd-fonts.yaml: add Amiri font entry using the arabic preset.

Tested on device with Amiri font and Arabic EPUB.
## Summary

* ** What is the goal of this PR? Implements Arabic text shaping so that Arabic characters render with correct contextual connected forms (initial/medial/final/isolated) instead of as disconnected isolated glyphs. Without this, Arabic fonts like Amiri render every letter in isolation regardless of its position in a word, making Arabic text unreadable.

* ** What changes are included?
lib/ArabicShaper/: new library that applies UAX#9 joining rules and replaces base Arabic codepoints (U+0600–U+06FF) with Unicode Presentation Forms (U+FE70–U+FEFF for standard Arabic, U+FB50–U+FBFF for Persian/Urdu/Pashto). Handles the mandatory lam-alef ligature and treats harakat as transparent.

lib/MiniBidi/bidiclasses.t: adds Arabic block (AL/NSM/AN) and Arabic Presentation Forms (AL) so do_bidi() reorders shaped glyphs correctly as RTL script.

lib/GfxRenderer/GfxRenderer.cpp: calls ArabicShaper::shape() inside resolveVisualText() before BiDi reordering, with a fast-path that skips non-Arabic text entirely.

[...]

PR #2222, opened 2026-05-31
feat: Add arabic support

YouHusam:
## Summary

* **What is the goal of this PR?**  
Add Arabic support across the firmware, including correct RTL rendering in the reader and UI, minimal built-in font coverage for Arabic UI text, and an Arabic UI translation so Arabic can be selected as a device language.  
Based on the work of @Uri-Tauber.

* **What changes are included?**  
  - Added Arabic bidi classification and shaping support in the MiniBidi pipeline.
  - Fixed RTL ordering issues that affected Arabic and Hebrew in UI text.
  - Routed Arabic shaping through the shared visual-text rendering path.
  - Added Arabic source font assets and regenerated the same built-in font header surface used by the Hebrew PR:
    - `notosans_8_regular.h`
    - `ubuntu_10_bold.h`
    - `ubuntu_10_regular.h`
    - `ubuntu_12_bold.h`
    - `ubuntu_12_regular.h`
  - Added a full Arabic UI translation YAML under translations.
  - Hardened the JPEGDEC patch helper so builds do not fail when the dependency is already partially patched.
  - Regenerated i18n outputs and the minimal built-in font assets needed for Arabic UI rendering.

## Additional Context

* Reader rendering is now correct for Arabic, including shaping and Lam-Alef handling.
* UI RTL behavior required a few follow-up fixes because the bidi class table is binary-searched and table ordering matters.
* Font changes were intentionally kept close to the Hebrew PR’s scope to avoid a large generated-font diff. The PR keeps only the minimal generated built-in font headers needed for Arabic UI coverage.
* Validation performed:
  - gen_i18n.py completed successfully with Arabic added and no fallbacks needed for the Arabic translation file.
  - `platformio run` completed successfully.
  - `platformio run --target upload` completed successfully on device.
* Reviewer focus areas:
  - RTL behavior in UI surfaces that use auto-direction.
  - Whether the minimal Arabic built-in font coverage is sufficient for all intended UI text.
  - Arabic translation wording and terminology consistency.

---

### AI Usage

Did you use AI tools to help write this code? _**YES**_

AI was used to help inspect the existing RTL implementation, prepare the Arabic translation file, draft code changes, and iterate on debugging/build fixes. Final changes were validated in-repo with the i18n generator and firmware build/upload.

I am a web developer, and the code was reviewed with my limited knowledge of C++.

Issue #1719, opened 2026-04-20
Support for Arabic, Urdu, and Farsi (Perso-Arabic script)


salad-dressing:
## Tracking issue: Arabic / Farsi / Urdu support

Splitting from #604 so that thread stays focused on the broader language picture.

### Why together

All three use the Perso-Arabic script, so a single font + shaper + RTL path covers them:

- Arabic — Naskh baseline
- Farsi — Arabic script + پ چ ژ گ
- Urdu — Farsi script + ٹ ڈ ڑ ں ہ ے, rendered in Naskh (Nastaliq is out of scope, it's a research-grade problem on this hardware)

Other Perso-Arabic-script languages (Pashto, Sindhi, Kashmiri, Uyghur, Kurdish variants, Punjabi in Shahmukhi, Balochi, etc.) would be largely covered by the same work, modulo a few additional codepoints per language.

### Prerequisites

- **#1327** — back-end SD card font infrastructure (.cpfont, prewarm glyph loading)
- **#1392** — font distribution; ships 22 fonts but none cover Arabic script

### Work after those land

1. Add Noto Naskh Arabic (or similar) to `sd-fonts.yaml` and regenerate .cpfont files
2. Arabic shaping (positional forms, Lam-Alef ligatures) — Papyrix has this working (bigbag confirmed in #604), worth studying before rolling our own
3. RTL layout in line-breaker, paragraph aligner, margins
4. Simplified paragraph-level RTL for v1; full UAX #9 BiDi as follow-up

### Out of scope

- Nastaliq for Urdu (separate hard problem)
- Hebrew (different shaping; could share RTL layout work)
- UI localization (book text only)

### Status

Picking this up after #1327 and #1392 merge. Papyrix remains the current option for Arabic readers — the Papyrix implementation (@bigbag) is the obvious reference point for the port.

Uri-Tauber:
My work in #1700 could be a good starting point for Arabic support. While I’m focusing on Hebrew, the same layout should work pretty well for Arabic too. Someone else can open a separate PR to handle Arabic shaping.

mcrosson:
Could anyone provide the following information. Even partial info would be helpful for me.

- A list of Google Noto fonts that would allow Arabic/Urdu/Farsi, in full as well as links to the font pages
- Any *open license* fonts readers may want to use beyond Noto fonts
[...]

PR #2541, opened 2026-07-05
feat: Arabic/Farsi/Urdu bidi reordering and contextual shaping — PR 1/3

YouHusam:
Arabic/Farsi/Urdu Bidi Reordering + Contextual Shaping — PR 1/3 (Foundation)

## Summary

This PR ports mintty's minibidi.c shaper (MIT licensed, 10+ years of production use) into CrossPoint's bidi pipeline to enable correct rendering of Arabic, Farsi, and Urdu text. Responsive to maintainer feedback on #2222 and #2398: we're splitting into three focused PRs, each standing alone; using proven upstream code rather than custom implementations; and keeping scope laser-focused.

What lands: UAX#9 bidi reordering + Arabic contextual shaping (joining types, presentation forms, Lam-Alef ligatures, diacritic transparency). Book-text rendering is now correct; UI glyphs follow in PR 2.

What's deferred: UI font glyphs (PR 2), Arabic translations (PR 3), Nastaliq (out of scope per #1719), built-in Harakat support (built-in fonts lack the glyphs; SD-card fonts can render them now via the combining-mark path).

Why 800+ Changes?

The diff is dominated by three load-bearing pieces:

1. lib/MiniBidi/minibidi.c (327 lines added) — the full do_shape() function from mintty, with CrossPoint deviations documented inline:
  - Joining-type lookups skip NSM marks (harakat) since they sit in our bidi_char array, unlike mintty's terminal cells.
  - Lam-Alef absorbed Alef is overwritten with LIGATURE_PLACEHOLDER (renderers drop it; mintty kept it as space).
  - Perso-Arabic table for Farsi/Urdu letters (ے ہ ں ڑ ڈ ٹ گ ژ چ پ…) outside mintty's native 0x0621–0x064A range, sourced from Unicode ArabicShaping.txt.
2. lib/MiniBidi/bidiclasses.t (54 lines added) — Unicode bidi-class tables for the entire Arabic block (0x0600–0x06FF, 0xFB50–FDFF, 0xFE70–FEFC), inserted after existing Hebrew ranges. Enables transparent harakat handling via isNSM classification.
3. test/minibidi_arabic/ (254 lines + 19 CMake) — host-side GTest suite exercising the full bidi+shape pipeline with 30 test cases covering:
  - Core Arabic contextual forms (isolated/initial/medial/final)
  - All four Lam-Alef ligature variants
  - Lam-diacritic-Alef (the specific bug from #2398 review)
  - Fully vocalized text (harakat contributes zero width)
  - Farsi & Urdu extra letters
  - ZWJ/ZWNJ forced forms
  - Mixed Arabic/Latin/numeral text

These three are essential; the rest is integration:

4. lib/MiniBidi/BidiUtils.cpp/h (39 + 7 lines changed) — wire shaping into the existing measurement+drawing pipeline; generalize niqqud range checks into isTransparentMark(cp).
5. lib/GfxRenderer/GfxRenderer.cpp (70 lines changed) — render harakat + niqqud as combining-mark overlays on the base glyph when the active font has the glyphs (SD fonts); skip silently for built-in fonts lacking them. Keeps measurement and drawing in sync (both see zero-width marks).
6. docs/file-formats.md + lib/Epub/Epub/Section.cpp — bump cache version to 29 (measurement semantics changed; v28 word positions no longer match shaped output).

Justification Against #2222/#2398 Feedback

- ✅ Split into independent PRs: This is PR 1/3; UI glyphs in PR 2 are optional and don't block book rendering.
- ✅ Human-verified upstream code: mintty's do_shape() has 10+ years of terminal use; we ported it (not invented a new shaper).
- ✅ Scope discipline: Only bidi+shaping+classification; no UI glyph changes, no translations, no infrastructure rewrites.
- ✅ Diacritic handling matches Hebrew: Harakat and niqqud now use the same transparent-mark path (#2398 rule 7).
- ✅ Covers Farsi/Urdu: Perso-Arabic table extends beyond mintty's core range (#2398 feedback).
- ✅ Measurement = drawing: Both paths see the same shaped+reordered codepoints (#2398 rule 6).
- ✅ Regression test suite: 30 test cases validate shaping correctness at the codepoint level; watch-list flags the two known "random space" bugs from #2398 testing.

Testing

- Host suite: test/minibidi_arabic/MiniBidiArabicTest.exe (all 30 tests pass)
- Firmware build: pio run -e default succeeds (no new binary bloat; tables in SRAM/Flash as before)
- Cache invalidation: Section.bin v28→v29 auto-invalidates old layouts (no manual cache clear needed)

What's NOT in this PR

- UI font glyphs (PR 2)
- Arabic UI strings (PR 3)
- Nastaliq rendering (out of scope per #1719)
- Built-in harakat glyphs (SD-card fonts render them; built-ins skip)

Follow-up (PR 2)

Once this merges, PR 2 will add a curated set of Arabic presentation-form glyphs to the built-in UI font (~28–30KB target, measured). Book text rendering is correct now; UI polish in PR 2.
---

### AI Usage

Did you use AI tools to help write this code? _**Yes with Fable**_

---

### Screenshots

I couldn't take normal screenshots due to unsupported characters, so I just photographed my device.

Arabic:
PXL_20260705_185848686

Bidi:
PXL_20260705_185951040


YouHusam:
@Uri-Tauber sorry for yet another Arabic PR, I used Fable for this to be written. I gave it instructions from previous MRs and discussions.

If you don't want to deal with it and just close it I understand

Uri-Tauber:
@YouHusam On the contrary! that's exactly what I had in mind. Thank you for putting this together.

I will need some time to review this, but it looks very promising. 
Could you do some benchmarking with English and Hebrew books, to check if the parsing time is affected by this?

Also, I would appreciate if other Arabic/Farsi speakers could test your work.

Uri-Tauber:
Two small fixes:

  1. Mutex for multi-core safety (not an issue right now, but could be in the future) — added  `std::mutex bidiMutex`  in  `BidiUtils.cpp`, locked in `applyBidiVisual()` and  `computeVisualWordOrder()` . The `static bidi_char[]`  arrays are shared mutable state that would corrupt under concurrent access. Also demoted the  `static std::string visual`  in `getTextAdvanceX()` to a local — it sat outside any lock and would still race.
  2.  joiners  init in  `computeVisualWordOrder`  — the new  `bidi_char.joiners`  field wasn't being initialised there. Harmless today (that path doesn't call `do_shape` ), but stale values from a prior `applyBidiVisual` call would bite if it ever does. Zeroed it out.

[0001-fix-potential-data-race.patch](https://github.com/user-attachments/files/29708558/0001-fix-potential-data-race.patch)

(apply with `git am 0001-fix-potential-data-race.patch`)

YouHusam:
> Two small fixes:
> 
>     1. Mutex for multi-core safety (not an issue right now, but could be in the future) — added  `std::mutex bidiMutex`  in  `BidiUtils.cpp`, locked in `applyBidiVisual()` and  `computeVisualWordOrder()` . The `static bidi_char[]`  arrays are shared mutable state that would corrupt under concurrent access. Also demoted the  `static std::string visual`  in `getTextAdvanceX()` to a local — it sat outside any lock and would still race.
> 
>     2. joiners  init in  `computeVisualWordOrder`  — the new  `bidi_char.joiners`  field wasn't being initialised there. Harmless today (that path doesn't call `do_shape` ), but stale values from a prior `applyBidiVisual` call would bite if it ever does. Zeroed it out.
> 
> 
> [0001-fix-potential-data-race.patch](https://github.com/user-attachments/files/29708558/0001-fix-potential-data-race.patch)
> 
> (apply with `git am 0001-fix-potential-data-race.patch`)

does your patch fix the 2 issues? if so I applied it and rebased with the latest develop, everything still works as expected

Uri-Tauber:
> does your patch fix the 2 issues? if so I applied it and rebased with the latest develop, everything still works as expected

Yes, it does.
Thanks. I will test on my device later for regressions with Hebrew, but I dont think there will be any problem.
Could other Arabic/Farsi speakers test this before we merge?

YouHusam:
@Uri-Tauber sorry for the wall of text, while trying to build an Arabic font I ran into this issue, we can ignore it for now, it talked about it in #1719 
It's an uncommon character used multiple times in the Quran, but can be found in other books on rare occasions. Outside the context of Quran, it's used as an opening for books or chapters most likely will exist in religious books. They use this character and not the actual phrase for decoration purposes in my opinion and that's why I think it's very minor. I don't know why anyone would want to read the Quran as an epub since it will not rendered as nicely as pdf or xtc or any better format suited for it.

I don't know if it's worth to change the character width to unit_16 since it might break old fonts, and it's such an edge case. However, I do think that we should take care of it in some way since it might break on other characters when using other fonts.

-----

here's a more detailed explanation from claude:

U+FDFD is a Unicode codepoint whose official name is "ARABIC LIGATURE BISMILLAH AR-RAHMAN AR-RAHEEM." It doesn't represent a letter or even a normal ligature — it's a single glyph that renders the entire opening phrase of the Quran, "بسم الله الرحمن الرحيم" ("In the name of God, the Most Gracious, the Most Merciful"), as one decorative unit. Font designers who want to support that traditional calligraphic rendering bake a whole-phrase glyph into the font at that codepoint. It's part of the "Arabic Presentation Forms-A" block (U+FB50–FDFF), tucked in among normal per-letter contextual forms.

Why it broke the converter

The .cpfont binary format (EpdGlyph struct in fontconvert_sdcard.py) stores each glyph's pixel width and height as a single unsigned byte (uint8_t, max value 255) — fine for ordinary letters, even wide ligatures like "lam-alef." But because U+FDFD encodes several words' worth of calligraphy in one glyph, its rasterized bitmap is enormous relative to a single character:


| Font size | Glyph width (px) | Glyph height (px) |
|-----------|------------------|-------------------|
| 12pt      | 304              | 43                |
| 14pt      | 355              | 50                |
| 16pt      | 404              | 58                |
| 18pt      | 455              | 64                |

Width alone exceeds 255 at every size you're generating, so struct.pack throws 'B' format requires 0 <= number <= 255 and the whole conversion aborts — not just for that glyph, but for the entire file, because packing happens after all glyphs are rasterized.

Why the fix is "just exclude it," not "widen the field"

I checked — it's genuinely the only offending codepoint in Noto Naskh Arabic across all four interval ranges and all four sizes you're building. Widening EpdGlyph.width/height to uint16_t would fix it structurally, but that changes the on-disk .cpfont format (currently a fixed 16-byte struct), which means every existing font file and the firmware's binary reader would need updating — a real format-version bump for one decorative glyph that isn't part of normal text shaping (readers don't type or encounter U+FDFD in running prose; it only shows up if someone deliberately embeds that specific ligature character).

- The bug: any font containing a glyph wider or taller than 255px at the target rasterization size crashes fontconvert_sdcard.py with an opaque struct.error, no indication of which glyph or codepoint caused it.
- The immediate workaround (what we did here): carve the offending cod argument by hand, e.g. splitting (0xFB50-0xFDFF) into(0xFB50-0xFDFC),(0xFDFE-0xFDFF).
- The maintainer's real decision is whether to make the script fail gracefully instead of hard-crashing — e.g., detect oversized glyphs during rasterization, print a clear warning naming the codepoint, and skip just that glyph (emit an empty placeholder) rather than aborting the whole build. That was the "Recommended" option I offered earlier; you chose the manual interval-exclusion instead, which works but means anyone converting an Arabic font hits the same opaque crash until they know to exclude U+FDFD.
- Worth noting for their triage: this isn't Arabic-specific in principle — any font/script with an oversized "whole word/phrase as one glyph" convention (some decorative Latin ligature sets, certain Indic conjuncts) could trip the same bug.

-----

I asked if there are other cases here's the answer

In this specific font: only FDFD

I checked every codepoint in your requested ranges (0600-06FF, 0750-077F, 08A0-08FF, FB00-FB06, FB50-FDFF, FE70-FEFF) across both regular and bold, all four sizes. U+FDFD is the only actual overflow, and nothing else is even close (I flagged anything over 200px as a near-miss — there were none).

Why it's specifically this one, and not its siblings

Unicode reserves U+FDF0–FDFF as a dedicated block of "religious phrase" ligatures for Arabic — Bismillah, Salamu Alaykum, Allahu Akbar, various Salawat (blessings on the Prophet), etc. It's the one place in Unicode where a font is explicitly expected to embed a multi-word phrase as a single glyph. All 14 of those codepoints exist in Noto Naskh Arabic — I confirmed that too — but only FDFD is drawn as the phrase spelled out inline at full width. The other 13 (including the closely-related FDFA, "peace be upon him," and FDF0, "peace be upon you") are drawn as compact stacked/monogram-style calligraphy that stays well under 255px. That's a font-design choice, not a Unicode requirement — some Arabic fonts draw all of these compactly, some draw several of them wide.

The broader answer for other fonts/languages

There's no fixed list I can hand the maintainer of "characters to always exclude" — it's font-dependent. The two things worth telling them:

1. Any Arabic font could trip this again, specifically via the FDF0–FDFF block, because that's the one Unicode-sanctioned spot for whole-phrase glyphs. Amiri, Scheherazade, and other calligraphic Arabic fonts commonly render more of these 14 slots as wide inline phrases than Noto does — so switching fonts later could resurface the same crash on a different codepoint in that block, not necessarily FDFD.
2. No other Unicode block has this same convention. Indic conjuncts, CJK, emoji ZWJ sequences, etc. don't have a single-codepoint "whole phrase baked into one glyph" pattern the way Arabic's religious ligatures do — those are built from combining sequences composed at shaping time, not oversized precomposed glyphs sitting in the font's cmap. Private-Use-Area codepoints (e.g. font vendor logos) are the other place this pattern shows up in the wild, but PUA isn't in any of the presets/ranges you're using, so it's not a live risk here.

Practical implication for the maintainer: manually excluding U+FDFD (what we did) is a one-off patch for this one font. If they plan to convert other Arabic fonts, they'll likely hit this again on a different codepoint each time, with the same opaque struct.error and no indication of which glyph caused it. That's the case for making the script fail gracefully — detect any glyph wider/taller than 255px during rasterization, print which codepoint and font caused it, skip just that glyph, and keep building — instead of hard-crashing the whole conversion.

Uri-Tauber:
The first time I see such a thing.

I suggest to address this in two directions: A: something similar to #2277, but in the other direction: Convert this glyph to individual characters on-the-fly.
B. Make the font generation script auto-skip code points which exceed 255px in the converted font. It will be cleaner solution than to hardcode specific codepoints.

Uri-Tauber:
@YouHusam I tested with Hebrew texts, and I noticed a bug with Hebrew Niqqud:
Currently every Niqqud is rendered in the middle above/below the letter, which works fine for some marks, like פתח or סגול, but break with דגש, שין ימנית/שמאלית or חולם. 
For example, lets take the following verse:

# כְּתֹנֶת־בַּד קֹדֶשׁ יִלְבָּשׁ וּמִכְנְסֵי־בַד יִהְיוּ עַל־בְּשָׂרוֹ וּבְאַבְנֵט בַּד יַחְגֹּר וּבְמִצְנֶפֶת בַּד יִצְנֹף בִּגְדֵי־קֹדֶשׁ הֵם וְרָחַץ בַּמַּיִם אֶת־בְּשָׂרוֹ וּלְבֵשָׁם׃

which rendered like this:
אחרי-מות_ch1_p4_6pct_172200.bmp


I'm assuming that you understand Hebrew. If I'm wrong, let me know and I'll try to explain better.

Here the epub for testing:
[אחרי מות.zip](https://github.com/user-attachments/files/29817738/default.zip)

YouHusam:
> The first time I see such a thing.
> 
> I suggest to address this in two directions: A: something similar to #2277, but in the other direction: Convert this glyph to individual characters on-the-fly. B. Make the font generation script auto-skip code points which exceed 255px in the converted font. It will be cleaner solution than to hardcode specific codepoints.

Yes, Arabic is a funny language and has a lot of "shortcut" characters in religious contexts




> @YouHusam I tested with Hebrew texts, and I noticed a bug with Hebrew Niqqud: Currently every Niqqud is rendered in the middle above/below the letter, which works fine for some marks, like פתח or סגול, but break with דגש, שין ימנית/שמאלית or חולם. For example, lets take the following verse:
> # כְּתֹנֶת־בַּד קֹדֶשׁ יִלְבָּשׁ וּמִכְנְסֵי־בַד יִהְיוּ עַל־בְּשָׂרוֹ וּבְאַבְנֵט בַּד יַחְגֹּר וּבְמִצְנֶפֶת בַּד יִצְנֹף בִּגְדֵי־קֹדֶשׁ הֵם וְרָחַץ בַּמַּיִם אֶת־בְּשָׂרוֹ וּלְבֵשָׁם׃
> 
> which rendered like this: אחרי-מות_ch1_p4_6pct_172200.bmp
> 
> I'm assuming that you understand Hebrew. If I'm wrong, let me know and I'll try to explain better.
> 
> Here the epub for testing: [אחרי מות.zip](https://github.com/user-attachments/files/29817738/default.zip)

I understand Hebrew and I see the issue, I'll try to take care of it tomorrow

YouHusam:
@Uri-Tauber I just did the nikkud fix, it's not perfect and I think there are also minor issues in Arabic Harakat too. In my opinion they are out of the scope of this PR and should be addressed on their own.

The issue I noticed in Hebrew is the positioning is inconsistent, if you look at יִהְיוּ  and בַּמַּיִם you can see in the screenshot that the Shuruk is too close to the letter and Dagesh inside the Mem is a little to the left. I'm not sure if it's a problem with the font I used (Noto Serif Hebrew) or something else.

Uri-Tauber:
@YouHusam Thats look very good for now! 
I don't fancy so much the manual hardcoded codepoints inside `anchorFor()`, It will be hard to scale, but for now it good enough.

I want to wait to @salad-dressing to test Urdu, and then we can merge this PR.

Uri-Tauber (approved):

Uri-Tauber:
Seems like we not going to get more testers soon, so I'm going to merge now.
Thanks @YouHusam for your work here!

Uri-Tauber:
@YouHusam Tag me when you create the two next PRs.
30kb for built-in Arabic glyphs is fine by me, so we can get it merged even today.

YouHusam:
> @YouHusam Tag me when you create the two next PRs. 30kb for built-in Arabic glyphs is fine by me, so we can get it merged even today.

Awesome thank you, I don't know if I'll have time today but I'll try

PR #2542, opened 2026-07-05
feat: Add anti-aliased font rendering with dithering

itsthisjustin:
Replace solid black rendering of anti-aliased font edges with period-2 dithering patterns. Full coverage pixels render as solid ink, while partial coverage pixels use checkerboard patterns (50% for dark grey, 25% for light grey) matching fillRectDither behavior. This prevents edge pixels from bolding to black on BW displays and produces proper anti-aliasing for 2-bit fonts. UI fonts are unchanged. SD fonts already were generated as 2bit but we werent using it.

Changes our Anti-Aliasing setting from On/Off to Off/Full/Fast

49 new lines of code before language files


Uri-Tauber:
Could you share some photos of before/after?

Also, what is the flash footprint of this?

itsthisjustin:
> Could you share some photos of before/after?
> 
> Also, what is the flash footprint of this?

Screenshots added. Doubles the UI font size on flash but still pretty small numbers

Uri-Tauber:
> Doubles the UI font size on flash but still pretty small numbers

Well, it add 106.6kb on flash. Thats a big number in my eyes.

itsthisjustin:
@Uri-Tauber added the compress flag and now they are actually SMALLER than they were prior to this change and still 2bit

uxjulia:
This looks terrible on device. The font looks washed out and too thin, especially the button labels, menu items, and status bars. The photos don't really convey how much worse it looks in person. Photo on my X4:

Before:

image

After:
image
image

Uri-Tauber:
> This looks terrible on device. The font looks washed out and too thin, especially the button labels, menu items, and status bars. The photos don't really convey how much worse it looks in person. Photo on my X4:

Agree. Just tested on my device. I don't like this. The zig-zag pattern on characters is quite ugly.

itsthisjustin:
Interesting! I found the thinner font to look better but can agree with preferences here. What if we only did it for reading font then?

Uri-Tauber:
> Interesting! I found the thinner font to look better but can agree with preferences here. What if we only did it for reading font then?

In my eyes, Noto Serif look awful with the zigzag edges. Noto Sans look a bit better, but still not something I will want to use for my day-to-day reading. 

How about keeping the built-in fonts as they are, but adding option to generate SD fonts with your AA?

itsthisjustin:
> > Interesting! I found the thinner font to look better but can agree with preferences here. What if we only did it for reading font then?
> 
> In my eyes, Noto Serif look awful with the zigzag edges. Noto Sans look a bit better, but still not something I will want to use for my day-to-day reading.
> 
> How about keeping the built-in fonts as they are, but adding option to generate SD fonts with your AA?

All the fonts already are generated with the AA but gfxrenderer ignores that and displays the full black anyway

Uri-Tauber:
> All the fonts already are generated with the AA but gfxrenderer ignores that and displays the full black anyway

So maybe to add setting for this? 

And to the other direction: maybe we should generate 1bit SD fonts? it probably will save some RAM...

lib/GfxRenderer/GfxRenderer.cpp
```diff
@@ -282,8 +282,15 @@ static void renderCharImpl(const GfxRenderer& renderer, GfxRenderer::RenderMode
...
-            renderer.drawPixel(screenX, screenY, pixelState);
+            // Full coverage is solid ink; partial (anti-aliased edge) coverage
+            // dithers with the same period-2 patterns as fillRectDither
+            // (dark grey = 50% checkerboard, light grey = 25%), so 2-bit fonts
+            // render anti-aliased on BW passes instead of bolding every edge
+            // pixel to black. Skipped pixels leave the background untouched.
+            if (bmpVal == 0 || (bmpVal == 1 && ((screenX + screenY) & 1) == 0) ||
+                (bmpVal == 2 && (screenX & 1) == 0 && (screenY & 1) == 0)) {
+              renderer.drawPixel(screenX, screenY, pixelState);
+            }
```
moben, line 295: I think with AA enabled this would cause the grayscale LUT to drive the white pixels (i.e. the ones not painted here) further towards white, not sure if that causes problems over time.

At least on X4 which I'm looking at currently.

There is the "01" LUT entry labeled "light gray", which drives white towards black and could be used here. It is currently unused in crosspoint, but it doesn't revert properly and might not be balanced and leave residual charge (not sure yet on the last point, need to check my math).

itsthisjustin: I'm now wondering if this looks bad for the other two folks here from testing on an x4 when I'm testing on an x3? @uxjulia what did you test on?

itsthisjustin: Confirmed issue with x4. Going to fix

itsthisjustin: Going to just revert the ui font changes and make this only for SD reading fonts

Uri-Tauber:
Something weird with `lib/I18n/translations/slovak.yaml`

itsthisjustin:
> Something weird with `lib/I18n/translations/slovak.yaml`

Weird. Are you able to fix? This was obviously just a mass Claude effort to translate strings

uxjulia:
I just tested this again and I'm not sure it brings any added value. Fonts look better with AA off than with this mode on. With this on, it reduces the weight of the font and then lightens it, so fonts look washed out and thinner... is it possible to maybe dial back how much it's dithering?

itsthisjustin:
> I just tested this again and I'm not sure it brings any added value. Fonts look better with AA off than with this mode on. With this on, it reduces the weight of the font and then lightens it, so fonts look washed out and thinner... is it possible to maybe dial back how much it's dithering?

It shouldn't do any lightening in fast mode as it's predithered. And this only effects reading fonts just to make sure we are testing the right thing?

itsthisjustin:
Closing this out as maintainers are not a fan of the look this provides

PR #2546, opened 2026-07-06
feat: Added OPDS Author folders

jb567:
## Summary

* Creates settings to allow automatically saving OPDS books under folders with an author name

## Additional Context

* I previously worked on this under #1214, which was closed because #1562 existed, which is now stale

---

### AI Usage


itsthisjustin:
@jb567 can you solve conflicts so we can get this in?

itsthisjustin:
I believe we can accomplish this with #2571 but let me know otherwise.

jb567:
@itsthisjustin

Unfortunately I don't think #2571 would achieve this, but I definitely agree that these changes are mutually exclusive.

#2571 does not allow for the folder to be dynamic, although minor changes allowing some string substitution in the config could help marry the two up, but it would be kinda annoying to type in on device. 

I do think it's technically a bit nicer having one set of config for downloads rather than it being per-server, so I'm not opposed to that being the PR that goes in, and I could work on follow up for changes. 

My branch isn't ready for a merge right now because I think I screwed up my merge conflicts

PR #2550, opened 2026-07-06
feat: implement captive portal redirects for auto-loading the management page on hotspot mode

latonis:
## Summary

* **What is the goal of this PR?** I had a bit of trouble of the network working on my android device, so I implemented a redirect for non API routes to allow for devices to automatically connect to the homepage once on the network

Without it, connecting to `CrossPoint-Reader` gives you a WiFi connection but nothing happens, you'd have to open a browser and manually type `192.168.4.1` or `crosspoint.local` or scan another qr code.

With the redir changes, the phone detects an unexpected response on its captive portal probe, recognizes it as a sign-in network, and auto-pops the browser straight to the CrossPoint home page without an additional step.

The main beneficiary is a non-technical user who just wants to drop an EPUB on the device. they scan the QR code on the e-ink screen to connect, their phone opens the page automatically, they upload the file, done.

* **What changes are included?** Handle captive endpoints for major device categories so that the management webpage automatically loads via instead of issuing a 404

image

## Additional Context

Docs for captive portal detection methods: https://en.wikipedia.org/wiki/Captive_portal#Detection

---

### AI Usage

Did you use AI tools to help write this code? _**no**_

### DIFF
--- a/src/network/CrossPointWebServer.cpp
+++ b/src/network/CrossPointWebServer.cpp
@@ -353,6 +353,15 @@ void CrossPointWebServer::handleJszip() const {
 }
 
 void CrossPointWebServer::handleNotFound() const {
+  // in AP mode, redirect all unmatched URLs to "/" — this handles captive portal detection probes
+  // (e.g. /generate_204, /hotspot-detect.html, /connecttest.txt) so the OS auto-opens the browser
+  // see https://en.wikipedia.org/wiki/Captive_portal#Detection
+  if (apMode) {
+    server->sendHeader("Location", "/", true);
+    server->send(302, "text/plain", "");
+    return;
+  }
+
   String message = "404 Not Found\n\n";
   message += "URI: " + server->uri() + "\n";
   server->send(404, "text/plain", message);
--- a/src/network/CrossPointWebServer.cpp
+++ b/src/network/CrossPointWebServer.cpp
@@ -353,10 +353,10 @@ void CrossPointWebServer::handleJszip() const {
 }
 
 void CrossPointWebServer::handleNotFound() const {
-  // in AP mode, redirect all unmatched URLs to "/" — this handles captive portal detection probes
-  // (e.g. /generate_204, /hotspot-detect.html, /connecttest.txt) so the OS auto-opens the browser
+  // in AP mode, redirect unmatched browser/captive-portal requests to "/" so the OS auto-opens the browser
+  // API requests (/api/*) still return 404 so XHR errors surface correctly
   // see https://en.wikipedia.org/wiki/Captive_portal#Detection
-  if (apMode) {
+  if (apMode && !server->uri().startsWith("/api/")) {
     server->sendHeader("Location", "/", true);
     server->send(302, "text/plain", "");
     return;


latonis:
> Caution
> 
> Some comments are outside the diff and can’t be posted inline due to platform limitations.
> 
> ⚠️ Outside diff range comments (1)
> > src/network/CrossPointWebServer.cpp (1)> `365-376`: _🎯 Functional Correctness_ | _🟡 Minor_ | _⚡ Quick win_
> > > **Blanket 404→redirect in AP mode will mask real API errors as 200 responses.**
> > > `handleNotFound()` now redirects _every_ unmatched path to `/` when `apMode` is true — not just captive-portal probe URLs. Since fetch/XHR follow same-origin 302 redirects transparently by default, any frontend call to a removed/mistyped `/api/...` endpoint while running as an AP will silently resolve to a 200 response with the home page body instead of a 404, breaking client-side error handling that branches on response status.
> > > Consider scoping the AP-mode redirect to likely browser-navigation requests (e.g., only when the request has no JSON/XHR-style `Accept` header, or restrict it to a known captive-portal probe path list) rather than every unmatched route.
> > > 
> > > 💡 Example scoping
> > > ```diff
> > >  void CrossPointWebServer::handleNotFound() const {
> > > -  // if the device is in AP mode, redirect to the root page so any unreachable URLs are handled correctly for the client
> > > -  if (apMode) {
> > > +  // if the device is in AP mode and this looks like a browser navigation (not an XHR/API call),
> > > +  // redirect to the root page so captive-portal-style probes are handled correctly for the client
> > > +  const bool looksLikeApiCall = server->uri().startsWith("/api/");
> > > +  if (apMode && !looksLikeApiCall) {
> > >      server->sendHeader("Location", "/", true);
> > >      server->send(302, "text/plain", "");
> > >      return;
> > >    }
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > 
> > > 
> > > 🤖 Prompt for AI Agents
> > > ```
> > > Verify each finding against current code. Fix only still-valid issues, skip the
> > > rest with a brief reason, keep changes minimal, and validate.
> > > 
> > > In `@src/network/CrossPointWebServer.cpp` around lines 365 - 376, The blanket
> > > AP-mode redirect in handleNotFound() is masking real 404s for API/XHR requests.
> > > Update CrossPointWebServer::handleNotFound() so the redirect to “/” only applies
> > > to browser/captive-portal navigation cases, and let unmatched API-style requests
> > > still return a 404. Use request context from the existing server object (for
> > > example headers such as Accept, or a known probe-path allowlist) to decide when
> > > to send the 302 versus the current 404 response.
> > > ```
> 
> 🧹 Nitpick comments (1)
> > src/network/CrossPointWebServer.cpp (1)> `181-189`: _📐 Maintainability & Code Quality_ | _🔵 Trivial_ | _💤 Low value_
> > > **Captive-probe routes registered unconditionally, not gated by `apMode`.**
> > > These routes are registered regardless of network mode. In practice they're only reachable in AP mode since the wildcard DNS redirect only runs there, so this is low risk, but consider gating registration (or the handler body) on `apMode` for clarity and to avoid unintended redirects if a client ever hits these exact paths in STA mode.
> > > 
> > > 🤖 Prompt for AI Agents
> > > ```
> > > Verify each finding against current code. Fix only still-valid issues, skip the
> > > rest with a brief reason, keep changes minimal, and validate.
> > > 
> > > In `@src/network/CrossPointWebServer.cpp` around lines 181 - 189, The
> > > captive-probe routes are currently registered unconditionally in
> > > CrossPointWebServer setup, even though they only make sense in AP mode. Update
> > > the route registration around the captiveRedirect lambda and the server->on
> > > handlers so they are gated by apMode, or add an explicit apMode check inside the
> > > handler body before sending the redirect. Keep the change localized to the
> > > captive portal setup in CrossPointWebServer to avoid unexpected redirects in STA
> > > mode.
> > > ```
> 
> 🤖 Prompt for all review comments with AI agents
> ```
> Verify each finding against current code. Fix only still-valid issues, skip the
> rest with a brief reason, keep changes minimal, and validate.
> 
> Outside diff comments:
> In `@src/network/CrossPointWebServer.cpp`:
> - Around line 365-376: The blanket AP-mode redirect in handleNotFound() is
> masking real 404s for API/XHR requests. Update
> CrossPointWebServer::handleNotFound() so the redirect to “/” only applies to
> browser/captive-portal navigation cases, and let unmatched API-style requests
> still return a 404. Use request context from the existing server object (for
> example headers such as Accept, or a known probe-path allowlist) to decide when
> to send the 302 versus the current 404 response.
> 
> ---
> 
> Nitpick comments:
> In `@src/network/CrossPointWebServer.cpp`:
> - Around line 181-189: The captive-probe routes are currently registered
> unconditionally in CrossPointWebServer setup, even though they only make sense
> in AP mode. Update the route registration around the captiveRedirect lambda and
> the server->on handlers so they are gated by apMode, or add an explicit apMode
> check inside the handler body before sending the redirect. Keep the change
> localized to the captive portal setup in CrossPointWebServer to avoid unexpected
> redirects in STA mode.
> ```
> 
> ℹ️ Review info
> ⚙️ Run configuration
> **Configuration used**: Organization UI
> 
> **Review profile**: CHILL
> 
> **Plan**: Pro
> 
> **Run ID**: `1de1c29b-e6cf-4d5d-99f1-a1d47a7e4446`
> 
> 📥 Commits
> Reviewing files that changed from the base of the PR and between [f2b00f6](https://github.com/crosspoint-reader/crosspoint-reader/commit/f2b00f6b930032e2b2f69b13e2f816a81143be12) and [415460a](https://github.com/crosspoint-reader/crosspoint-reader/commit/415460a0e0549626ddbd8cc061779c972bb385eb).
> 
> 📒 Files selected for processing (1)
> * `src/network/CrossPointWebServer.cpp`
> 
> 📜 Review details

addressed in 5929efedd016c924d992c2fb25307e9da7a103fa

itsthisjustin:
Super sorry if this is a dumb question, but how does your android hotspot have a captive portal? In principal I agree with this PR but I'm confused by the use case that causes it

latonis:
> Super sorry if this is a dumb question, but how does your android hotspot have a captive portal? In principal I agree with this PR but I'm confused by the use case that causes it

My android hotspot doesn't have a captive portal, I connected to the hotspot created via the X4 and it took a few retries to get it to connect reliably. I am proposing using this method to get the upload page to load immediately via captive portal detection on the device

itsthisjustin:
> > Super sorry if this is a dumb question, but how does your android hotspot have a captive portal? In principal I agree with this PR but I'm confused by the use case that causes it
> 
> My android hotspot doesn't have a captive portal, I connected to the hotspot created via the X4 and it took a few retries to get it to connect reliably. I am proposing using this method to get the upload page to load immediately via captive portal detection on the device

Ah! You're using the captive portal detection to bypass the issue of not being able to bring up .local pages. Neat. Got it

itsthisjustin (approved):

PR #2552, opened 2026-07-07
feat: add Switch Firmware option to System settings

iammoen:
The following MR was created with Claude Sonnet. It has been built and tested.

Claude summary follows:
> ## What this does
> 
> Adds a **Switch firmware** action to Settings → System that lets users switch between firmware images stored in the two OTA partitions — without needing a USB cable or the web debug tools.
> 
> The option only appears when the inactive OTA partition holds a valid app image (checked via `esp_ota_get_partition_description`), so it is completely invisible to users who only ever run CrossPoint.
> 
> ## Why this fits
> 
> CrossPoint already has this exact capability — `ota_boot::switchTo()` exists, and the web debug page exposes a working \"Swap boot partitions\" button. This PR surfaces that same mechanism in the on-device UI.
> 
> The use case it enables: a user installs CrossPoint alongside another firmware (e.g. a custom display app) and wants to switch between them without hunting for a cable. The feature does not add connectivity, background tasks, or any persistent state — it is a one-shot action that calls `ota_boot::switchTo()` and reboots.
> 
> ## Behaviour
> 
> - **Detected as CrossPoint**: shows the firmware name + version, confirms the same Settings menu can be used to return.
> - **Unknown firmware**: shows the name + version and a warning that returning *may* require web debug tools or USB if the alternate firmware has no built-in switch-back option.
> - All text uses `renderer.wrappedText()` so arbitrary firmware names and long strings render cleanly at any screen width.
> - Works symmetrically — CrossPoint in app0 targets app1, CrossPoint in app1 targets app0.
> 
> ## Changes
> 
> - `src/activities/settings/BootAlternateFirmwareActivity.{h,cpp}` — new activity (confirm → switch → reboot, or fail gracefully)
> - `src/activities/settings/SettingsActivity.{h,cpp}` — adds `BootAlternateFirmware` action; conditionally appends menu item
> - `lib/I18n/translations/english.yaml` — 9 new strings (English only; other languages fall back per existing i18n policy)
> 
> ## Testing
> 
> Built and tested on hardware (Xteink X3):
> - Option hidden when app1 has no valid firmware ✓
> - Option visible when app1 has valid firmware ✓
> - Unknown firmware warning path renders correctly with wrapped text ✓
> - CrossPoint → CrossPoint path shows return hint ✓
> - Switch completes and device reboots into alternate firmware ✓
> 
> I'm happy to update strings, adjust wording, or scope this down if the maintainers feel it belongs behind a developer/advanced section rather than the main System tab.


itsthisjustin:
My only fear with this, is someone booting into a broken OTA partition or a locked user doing this and then getting locked out of a firmware update path

iammoen:
> My only fear with this, is someone booting into a broken OTA partition or a locked user doing this and then getting locked out of a firmware update path

First off, I totally agree. That is a concern of mine as well. I am not quite sure what a "locked" user would be, but with the broken ota partition or whatever I do understand. I was thinking that gating the visibility of this option with there being something else on the other partition would significantly lessen this chance. If someone had put something in app1 instead of just app0 partition, i had been thinking that would show that they are doing things beyond what the average user does. I was using that as a gauge of intention.

I am certainly open to more dialog about this in general and anything else we could do (other than the visibility thing above and perhaps a more well thought out warning message than the one I built) to make this give everyone the warm and fuzzies. And at the end of the day, I am certainly ok with saying "this just isn't something we want a part of crosspoint". I made this for myself and figured other people might benefit from it.

A quick thank you also for this software you all maintain. I have been digging it tremendously.

itsthisjustin:
> > My only fear with this, is someone booting into a broken OTA partition or a locked user doing this and then getting locked out of a firmware update path
> 
> First off, I totally agree. That is a concern of mine as well. I am not quite sure what a "locked" user would be, but with the broken ota partition or whatever I do understand. I was thinking that gating the visibility of this option with there being something else on the other partition would significantly lessen this chance. If someone had put something in app1 instead of just app0 partition, i had been thinking that would show that they are doing things beyond what the average user does. I was using that as a gauge of intention.
> 
> I am certainly open to more dialog about this in general and anything else we could do (other than the visibility thing above and perhaps a more well thought out warning message than the one I built) to make this give everyone the warm and fuzzies. And at the end of the day, I am certainly ok with saying "this just isn't something we want a part of crosspoint". I made this for myself and figured other people might benefit from it.
> 
> A quick thank you also for this software you all maintain. I have been digging it tremendously.

Technically anytime they OTA update stock/crosspoint or do a flash from the website, the old version is resident on the other ota partition i believe. So they dont need to do anything overly technical to end up in that state

iammoen:
> > > My only fear with this, is someone booting into a broken OTA partition or a locked user doing this and then getting locked out of a firmware update path
> > 
> > 
> > First off, I totally agree. That is a concern of mine as well. I am not quite sure what a "locked" user would be, but with the broken ota partition or whatever I do understand. I was thinking that gating the visibility of this option with there being something else on the other partition would significantly lessen this chance. If someone had put something in app1 instead of just app0 partition, i had been thinking that would show that they are doing things beyond what the average user does. I was using that as a gauge of intention.
> > I am certainly open to more dialog about this in general and anything else we could do (other than the visibility thing above and perhaps a more well thought out warning message than the one I built) to make this give everyone the warm and fuzzies. And at the end of the day, I am certainly ok with saying "this just isn't something we want a part of crosspoint". I made this for myself and figured other people might benefit from it.
> > A quick thank you also for this software you all maintain. I have been digging it tremendously.
> 
> Technically anytime they OTA update stock/crosspoint or do a flash from the website, the old version is resident on the other ota partition i believe. So they dont need to do anything overly technical to end up in that state

Ah ok. That's a pretty low bar then. 

There is another option here I think if we wanted to pursue it. If the bootloader is compiled with CONFIG_BOOTLOADER_APP_ROLLBACK_ENABLE=y (which i think is the default), we could use the `esp_ota_mark_app_valid_cancel_rollback()` function at a point of "success" in crosspoint booting, and if the function isn't called the bootloader would go back to the last used partition?

But again maybe this isn't something crosspoint maintainers feel inclined to want to pursue. I am certainly willing to work on that if people feel like it would be a benefit. Happy to do whatever you feel is best (including starting a conversation somewhere more appropriate for a change like this)

itsthisjustin:
I think for now I'm going to close this but it's something we should open up for discussion. I'm just very wary of adding something that could brick a locked device with a single accidental button press.

PR #2553, opened 2026-07-07
fix: prevent font manager OOM abort with many SD fonts installed

pillarsdotnet:
## Summary

Opening **Settings > Fonts > Manage Fonts** can `abort()` during font-manifest parsing on low-heap devices that have an **SD reader font active** and **many font families installed**. This fixes it by freeing the active font and the font registry before the network/parse work, and by not holding the parsed manifest and the registry in memory at the same time.

**Verified on hardware (X3)** — see the A/B below.

## Root cause

At the parse peak, four large things were resident in RAM simultaneously:

1. the discovered SD-font **registry** (populated in `SdCardFontSystem::begin()` at boot and never released),
2. the **active reader font** (`onEnter()` freed nothing),
3. the parsed manifest **`JsonDocument`** (~40 KB for the ~29 KB `fonts.json`), and
4. the growing **`families_`** list.

With `-fno-exceptions`, the failing allocation doesn't return `nullptr` — it throws `bad_alloc`, which goes to `__terminate` → `abort()` (the hazard in `CLAUDE.md` rule #9).

Symbolicated crash (unfixed, `riscv32-esp-elf-addr2line`):

```
__terminate ← std::bad_alloc ← __cxa_allocate_exception
  ← std::vector<ManifestFile>::_M_realloc_append(ManifestFile&&)
  ← FontDownloadActivity::fetchAndParseManifest  (onWifiSelectionComplete)
```

The reproduction needs the heap already squeezed low (~48 KB free), which happens in normal use after reading with an SD font selected — not on a fresh boot.

## Changes

- **`SdCardFontRegistry::clear()`** — release discovered family metadata (repopulated by `discover()`).
- **`SdCardFontSystem::releaseForNetwork()`** — unload the active font (`manager_.unloadAll`) and clear the registry before heavy network work; both restore on demand.
- **`FontDownloadActivity::onEnter()`** calls `releaseForNetwork()`; **`onExit()`** restores via `ensureLoaded()` when WiFi selection is cancelled (the WiFi-connected path already reboots via `silentRestart()`).
- **`fetchAndParseManifest()`** now parses in **two passes**: build `families_` from the JSON, free the `JsonDocument`, then re-discover the registry for installed/update detection — so the JSON and the registry are never resident together.

No behavior change to the resulting font list; this only lowers peak heap during load.

## Testing

Built `env:default`, flashed to an X3 with an SD reader font active (NotoSansExtended, XLarge) and 24 SD font families (192 `.cpfont` files) installed, heap squeezed by reading a book first. Same steps each time → Settings > Fonts > Manage Fonts:

| Build | Result |
|---|---|
| `develop` (unfixed) | ❌ `abort()` — `bad_alloc` in `vector<ManifestFile>` append |
| two-pass parse only | ❌ `abort()` — `bad_alloc` in a `std::string` (registry is resident from boot, so reordering alone didn't help) |
| **this PR** (`releaseForNetwork` + two-pass) | ✅ Font Browser loads, no abort |

(For on-device reproduction the manifest was served from an HTTP mirror, since the ESP32‑C3 stalls on the GitHub‑HTTPS manifest redirect before reaching the parse — a separate transport issue, not part of this change.)

## Related

This is the upstream counterpart of the same crash reported on the CrossInk fork: [uxjulia/CrossInk#344](https://github.com/uxjulia/CrossInk/issues/344) ("Crash abort 1.4.0 RC in manage fonts"). The mechanism is identical (`bad_alloc` → `__terminate` during font-manifest parse under low heap). CrossInk already had a `releaseForNetwork()` helper that frees the registry, so its fix was smaller (call it in `onEnter()` + the two-pass parse); this repository had no font/registry release path, so this PR adds `SdCardFontRegistry::clear()` and `SdCardFontSystem::releaseForNetwork()` before applying the same approach. The CrossInk fix was verified on an X3; this upstream fix was verified on the same device against `develop`.

### DIFF
--- a/lib/EpdFont/SdCardFontRegistry.h
+++ b/lib/EpdFont/SdCardFontRegistry.h
@@ -49,6 +49,10 @@ class SdCardFontRegistry {
   int getFamilyIndex(const std::string& name) const;
   int getFamilyCount() const { return static_cast<int>(families_.size()); }
 
+  // Release all discovered family metadata, freeing its heap. Call discover()
+  // again to repopulate. Used to reclaim memory before heavy network work.
+  void clear() { std::vector<SdCardFontFamilyInfo>().swap(families_); }
+
  private:
   std::vector<SdCardFontFamilyInfo> families_;  // sorted alphabetically
 
--- a/src/SdCardFontSystem.h
+++ b/src/SdCardFontSystem.h
@@ -45,6 +45,16 @@ class SdCardFontSystem {
     }
   }
 
+  /// Free the active SD font and the discovered registry before a memory-heavy
+  /// network operation (e.g. the font manager). Both are restored on demand
+  /// afterwards: the registry via discover()/refreshIfDirty(), the active font
+  /// via ensureLoaded().
+  void releaseForNetwork(GfxRenderer& renderer) {
+    manager_.unloadAll(renderer);
+    registry_.clear();
+    registryDirty_.store(true, std::memory_order_release);
+  }
+
  private:
   SdCardFontRegistry registry_;
   SdCardFontManager manager_;
--- a/src/activities/settings/FontDownloadActivity.cpp
+++ b/src/activities/settings/FontDownloadActivity.cpp
@@ -24,6 +24,11 @@ FontDownloadActivity::FontDownloadActivity(GfxRenderer& renderer, MappedInputMan
 
 void FontDownloadActivity::onEnter() {
   Activity::onEnter();
+  // Free the active reader font and the SD font registry before the network +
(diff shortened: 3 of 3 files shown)


Uri-Tauber:
@uxjulia Have you tested this fix?

uxjulia:
> @uxjulia Have you tested this fix?

Not with the excessive fonts situation, but it's a similar fix to another issue I had where KOReader sync would fail with too many fonts. The user in that situation tested it and said it worked.

itsthisjustin:
This may already be fixed by #2475

PR #2554, opened 2026-07-08
perf: share one static bidi scratch buffer to save ~1.5 KB RAM

pillarsdotnet:
## Summary

`applyBidiVisual()` and `computeVisualWordOrder()` in `lib/MiniBidi/BidiUtils.cpp` each held their own `static bidi_char line[BIDI_MAX_LINE]` reorder buffer (1.5 KB each), both permanently resident in DRAM — even for LTR-only reading.

They run on the single render task and never call each other, so they can share **one** file-scope scratch buffer instead of two, freeing **1,536 bytes** of always-resident RAM with no behavior change (the buffers were already non-reentrant statics; this preserves that same assumption).

## Verification

Built `env:default`: `.bss` now has a single `sharedBidiLine` symbol where there were two per-function statics; measured RAM dropped by exactly 1,536 bytes (e.g. 100,060 vs the prior baseline). No functional change to bidi reordering.

## Mechanism

The two functions populate an identically-typed `bidi_char[BIDI_MAX_LINE]` line buffer and are only ever invoked serially from the render task, so a shared file-static is safe and removes one full copy from DRAM.

### DIFF
--- a/lib/MiniBidi/BidiUtils.cpp
+++ b/lib/MiniBidi/BidiUtils.cpp
@@ -34,6 +34,12 @@ bool isNaturalDirectionClass(const uchar cls) {
   }
 }
 
+// Visual-reorder scratch shared by applyBidiVisual() and
+// computeVisualWordOrder(). Neither function calls the other, and bidiMutex
+// already serialises both, so a single buffer serves both instead of a
+// per-function static — saving ~1.5 KB of always-resident RAM.
+bidi_char sharedBidiLine[BIDI_MAX_LINE];
+
 }  // namespace
 
 namespace BidiUtils {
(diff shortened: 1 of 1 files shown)


Uri-Tauber:
Nice finding. 
Will look into this after #2541 is merged.

Uri-Tauber:
@pillarsdotnet Could you fix conflicts?

pillarsdotnet:
> @pillarsdotnet Could you fix conflicts?

@Uri-Tauber Absolutely. Rebased and re-pushed. Verified that a build with this change still saves 1536 bytes relative to baseline.

Uri-Tauber (approved):

PR #2560, opened 2026-07-09
feat: HTTP serve static with Cache-Control and ETag headers

shirok1:
## Summary

The WebUI is currently a static web page, relying on JS and refreshing to display new data. But since these static resources are not served with Cache-Control header, the browser is not willing to cache them but requesting them each time, making performance impact.

This PR set `Cache-Control` and `ETag` headers, asking browser to request with the `ETag` later so server side can return `304 Not Modified` instead of full file.

Apart from server, `scripts/build_html.py` is also changed to compute sha256 of file content and generate consts like `HomePageHtmlETag`.

## Additional Context

`if (server->header("If-None-Match") == etag)` might add a string compare cost, but since only [:16] of sha256 is taken as `ETag` that's probably fine.

---

### AI Usage

Claude Code + Claude Sonnet 5

### DIFF
--- a/scripts/build_html.py
+++ b/scripts/build_html.py
@@ -61,8 +61,9 @@ for root, _, files in os.walk(SRC_DIR):
                 processed = content
 
             # Compress with gzip (compresslevel 9 is maximum compression)
+            # mtime=0 keeps the output reproducible across builds
             # IMPORTANT: we don't use brotli because Firefox doesn't support brotli with insecured context (only supported on HTTPS)
-            compressed = gzip.compress(processed.encode('utf-8'), compresslevel=9)
+            compressed = gzip.compress(processed.encode('utf-8'), compresslevel=9, mtime=0)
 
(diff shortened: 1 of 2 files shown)


itsthisjustin:
@shirok1 any regressions this could cause? any known issues that might arise?

shirok1:
> @shirok1 any regressions this could cause? any known issues that might arise? 

the webui static serving be changed so worth checking out the whole webui. also webdav cloud be affected for extra headers added.

shirok1:
@itsthisjustin got any time to test this around?

itsthisjustin:
> @itsthisjustin got any time to test this around?

Will take a look later tonight! Working on x4pro port

itsthisjustin:
I'm going to try to test this against my new sd plugins system for webui and see if there's any impact.

itsthisjustin (approved):

PR #2576, opened 2026-07-11
Use selected SD font as glyph fallback for user-content titles

eommisae01:
## Summary

Fixes missing CJK/Korean glyph rendering for book titles and file names by using the selected SD-card reader font as a glyph fallback only for user-content text.

I'm aware that Korean-specific firmware builds exist, but I'd like to keep using the original upstream CrossPoint firmware and only have user-content titles and file names render correctly when an SD-card font contains the needed glyphs. This keeps the UI and firmware behavior upstream-compatible while fixing missing title glyphs for Korean/CJK users.

## Problem

Reader content can render Korean correctly when an SD `.cpfont` is selected, but Home, Recent Books, and File Browser titles are rendered with built-in UI fonts. Those fonts may not contain Hangul/CJK glyphs, so titles can appear blank or as replacement glyphs.

## Approach

- Keep built-in UI fonts as the primary UI font.
- Add exact glyph coverage checks that do not treat U+FFFD replacement fallback as a real glyph.
- Use the selected SD-card font only for missing glyphs in user-content strings.
- Apply fallback to Home recent titles, Recent Books list title/author, and File Browser names.
- Keep UI labels, settings, menus, and other application text unchanged.
- Use the same fallback rule for measurement, truncation/wrapping, and drawing.

## Validation

- `git diff --check`
- `pio check -e default`
- `pio run -t unit-tests` - 86/86 tests passed
- `pio run -e default` - firmware build succeeded

A local firmware build was produced for validation only. No device flashing, unlock/re-lock, partition changes, or OTA install were performed.

### DIFF
--- a/docs/korean-title-fallback.md
+++ b/docs/korean-title-fallback.md
@@ -1,0 +1,46 @@
+# Korean title fallback
+
+This patch keeps the built-in UI fonts as the primary fonts, and uses the
+currently selected SD-card reader `.cpfont` only for user-content titles and
+file names that contain codepoints missing from the UI font.
+
+## Traced rendering paths
+
+- Home recent title: `HomeActivity::render()` passes the latest book title to
+  `BaseTheme::drawHeader()` when `homeContinueReadingInMenu` is active, and to
+  `drawRecentBookCover()` for cover tiles.
+- Recent books list: `RecentBooksActivity::render()` passes `RecentBook::title`
+  and `RecentBook::author` to `drawList()`.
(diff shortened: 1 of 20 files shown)


Uri-Tauber:
I believe this is a duplicate of #2521

eommisae01:
Thanks for pointing this out. I missed #2521 when opening this PR.

#2521 covers the same underlying issue in a broader upstream direction, so I'm closing this as a duplicate and will follow that PR instead.

Issue #2570, opened 2026-07-10
OTA Unlocker Tool Update Failing


Washbowl7980:
### Affected Version

1.4.1

### Describe the Bug

I'm immediately get "Update Failed" when using the unlocker to flash when 1.4.1 is installed. (I'm mostly sure it's not my set up due to it working with an older release.)

This isn't urgent due to SD flashing working fine. But i have seen the failure in forks.

### Steps to Reproduce

1. Set up unlocker tool.
2. Connect device to shared network.
3. Navigate to "Check for Updates"
4. Next screen is "Update failed"

### Expected Behavior

Pressing "Check for updates" provides an update available screen, listing 99.9 as the update. Confirming update leads to device updating to new firmware.

### Relevant Log Output/Screenshots

```shell

```

eszter007:
I have the same issue with v 1.4.1

ituri:
I have spent quite some time trying to find a solution for the failing OTA and I got it working again. 

While my PR is actually intended for Crossink, it should also work for CrossPoint itself: https://github.com/uxjulia/CrossInk/issues/312#issuecomment-4946141620

PR #2579, opened 2026-07-12
fix: run OTA check and install at boot to give TLS enough heap (#2570)

### LINKED ISSUE #2018: X3 OTA fails with `ESP_ERR_NO_MEM` at `esp_https_ota_begin`

### LINKED ISSUE #2419: OTA update not working on X4 1.3.0 unlocked device
### Affected Version

1.3.0

### Describe the Bug

I am trying to update X4 from 1.3.0 to 1.4.0 over wifi but it keeps saying update failed after 1 second.

I have a device bought from xteink directly unlocked. I can transfer files over wifi wifh no issues.

### Steps to Reproduce

1. Go to settings and then check for updates
2. Press check for updates and then update to 1.4.0 after connected to network
3. After 1 second updates fails.

### Expected Behavior

It should download and update the device over wifi to 1.4.0

```shell

```

ituri:
Addresses #2570 (and the broader "Update failed" class of OTA reports).

## TL;DR

OTA failures are a heap problem: a TLS handshake needs tens of KB of free heap at its peak, but inside the settings UI with WiFi up only ~40 KB remain. This PR moves the OTA check and install to a boot-time stage that runs right after display init with >100 KB free, and lends the framebuffer to the download while the e-ink panel retains the "Updating…" screen.

## Background

I debugged this extensively in the CrossInk fork (uxjulia/CrossInk#312) on an X4 that had never completed an OTA, with serial logs and packet captures. The measured mechanics, which apply to any TLS engine:

- During a handshake, free heap bottoms out near ~1 KB: record buffers plus certificate processing plus crypto scratch, while received TLS records pile up in RAM because the crypto runs for hundreds of ms in software. Once the heap floor is hit, lwIP can't allocate RX pbufs, the handshake stalls into a timeout — or an allocation fails outright. Which symptom you get depends on where exactly the memory runs out, which is why reports are so inconsistent across devices and networks.
- During the firmware download the same backpressure applies continuously: with ~8 KB free the transfer trickles at ~3 KB/s and eventually times out; with the framebuffer's KBs freed it runs at ~90 KB/s.

With the boot-time stage plus framebuffer lending, that device completed its first ever end-to-end OTA (check → confirm → 5.5 MB download → install → reboot). The fork's fix has additional mbedTLS-specific parts (certificate-chain memory, GitHub API asset endpoint); those don't apply here since this repo's HTTP path runs over wolfSSL with `setInsecure()`, so this PR ports only the engine-agnostic architecture.

## Changes

- **`src/network/OtaBootCheck`** (new): records the requested stage (check/install) in RTC memory, silent-restarts (existing mechanism), reconnects to the last saved WiFi network headlessly, and runs the updater with a minimal status UI. Read-and-clear flag semantics so a panic can't boot-loop.
- **`OtaUpdateActivity`**: no longer performs network work — consumes the boot result, collects confirmation, requests the install stage. The WiFi selection UI only runs when no saved credential exists yet.
- **`HalDisplay::releaseBuffers()/reallocBuffers()`**: passthroughs to the existing FreeInk SDK API; the install stage hands the framebuffer heap to the download and the panel retains the drawn frame.
- UX: one brief silent reboot before the check result appears, and the install shows a static "Updating…" screen instead of a live progress bar (rendering is offline while its memory is lent out).

## Testing

- `pio run -e default` builds clean.
- The architecture is verified end-to-end on X4 hardware in the CrossInk fork (same activity/boot structure); I don't have a stock-Crosspoint test setup, so testers with affected devices — especially the unlocker-tool flow from #2570 — would be very welcome.

### DIFF
--- a/lib/hal/HalDisplay.cpp
+++ b/lib/hal/HalDisplay.cpp
@@ -77,6 +77,10 @@ void HalDisplay::deepSleep() { einkDisplay.deepSleep(); }
 
 uint8_t* HalDisplay::getFrameBuffer() const { return einkDisplay.getFrameBuffer(); }
 
+void HalDisplay::releaseBuffers() { einkDisplay.releaseBuffers(); }
+
+bool HalDisplay::reallocBuffers() { return einkDisplay.reallocBuffers(); }
+
 void HalDisplay::copyGrayscaleBuffers(const uint8_t* lsbBuffer, const uint8_t* msbBuffer) {
   einkDisplay.copyGrayscaleBuffers(lsbBuffer, msbBuffer);
 }
--- a/lib/hal/HalDisplay.h
+++ b/lib/hal/HalDisplay.h
@@ -47,6 +47,13 @@ class HalDisplay {
   // Access to frame buffer
   uint8_t* getFrameBuffer() const;
 
+  // Hand the framebuffer heap back to the system while the panel keeps
+  // showing its last refresh (e-ink retains the image unpowered). Rendering
+  // is unavailable until reallocBuffers() succeeds again. Used by the
+  // boot-time OTA download, which needs every contiguous block it can get.
+  void releaseBuffers();
+  bool reallocBuffers();
+
   // X3 grayscale preconditioning (OEM "AA-pre-BW(mid)" settle pass), windowed
   // to the gray region in physical panel coordinates (no-arg = full frame).
   // Call after the BW base frame is displayed and before the grayscale planes
--- a/src/activities/settings/OtaUpdateActivity.cpp
+++ b/src/activities/settings/OtaUpdateActivity.cpp
@@ -9,7 +9,22 @@
 #include "activities/network/WifiSelectionActivity.h"
 #include "components/UITheme.h"
 #include "fontIds.h"
-#include "network/OtaUpdater.h"
+#include "network/OtaBootCheck.h"
+
+void OtaUpdateActivity::consumeBootResult(const OtaBootCheck::Result& result) {
+  {
+    RenderLock lock(*this);
+    if (result.error == OtaUpdater::OK) {
+      updater.adoptManifest(result.version, result.url, result.size);
+      state = updater.isUpdateNewer() ? WAITING_CONFIRMATION : NO_UPDATE;
+    } else if (result.error == OtaUpdater::NO_UPDATE) {
+      state = NO_UPDATE;
+    } else {
+      state = FAILED;
+    }
+  }
+  requestUpdate(true);
+}
 
 void OtaUpdateActivity::onWifiSelectionComplete(const bool success) {
   if (!success) {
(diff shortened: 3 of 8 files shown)


ituri:
Both findings addressed in 2ede7023: the renderer's cached framebuffer pointer is refreshed via `renderer.begin()` after `reallocBuffers()` (real issue — the old code only worked when the allocator happened to return the same block), and the boot stage now picks its status message by stage. Build verified.

ituri:
Retested as requested (cc @Diirge from Discord): flashed **stock develop @ d8139e96** (includes the wolfSSL OTA work from #2475) onto the X4 that previously failed every OTA check, and the check now **succeeds in the activity context** — "No update available", full 32 KB release JSON fetched and parsed, no crash.

The serial numbers add useful context though: during the check, minimum free heap dropped to **11.6 KB** (from ~73 KB free before the fetch). So the wolfSSL migration shrinks the TLS footprint enough to fix the common case — confirming your read — but the margin in the activity context is ~10 KB, and it shrinks further with fatter device states (more SD fonts/themes; for comparison, the CrossInk fork has ~30 KB *less* free at the same point and still fails there without this PR) and weak-WiFi retransmit pileups. The install path (5+ MB streaming download in the same context) wasn't exercised since no newer release was offered.

So I'd keep this PR as defense-in-depth: the boot-time stage makes OTA independent of whatever the UI heap looks like, and the framebuffer lending helps download throughput regardless of TLS engine. Happy to adjust scope if you'd rather take only parts of it.

PR #2588, opened 2026-07-13
fix: Use HALF_REFRESH for sleep and boot instead of FULL

itsthisjustin:
Changes all sleep screen display refreshes and boot screens from FULL_REFRESH to HALF_REFRESH to match OEM firmware behavior. The stock firmware uses a single-pass 0xD7 waveform for sleep screens, not the multi-flash GC waveform (0xF7) that FULL_REFRESH triggers. For grayscale mode, HALF_REFRESH is required because the gray nudge LUT is calibrated against the pixel charge state left by the single-pass waveform; using FULL_REFRESH causes blotchy noise in gray areas. Also adds X3 wb_gc passive byte (the one-byte 0x54 to 0x00, matching stock's passive WB table in the gc nudge bank): Every X3 grayscale render, reader AA, covers, and the sleep-cover nudge, gets it.

### DIFF
--- a/src/activities/boot_sleep/SleepActivity.cpp
+++ b/src/activities/boot_sleep/SleepActivity.cpp
@@ -149,6 +149,10 @@ void SleepActivity::renderCustomSleepScreen() const {
   renderDefaultSleepScreen();
 }
 
+// Sleep screens paint with a single HALF refresh (stock parity): the OEM X4
+// firmware's only clean refresh in normal operation is the single-pass 0xD7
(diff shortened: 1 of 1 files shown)


Uri-Tauber:
What are the changes in the SDK?

uxjulia (approved):

itsthisjustin:
> What are the changes in the SDK?

Listed above in description. There's a change in the byte we send for gray

PR #1041, opened 2026-02-20
feat: List improvements (WIP)

sam-lord:
## Summary

**What is the goal of this PR?** (e.g., Implements the new feature for file uploading.)
* Full compatibility with lists in epub view

**What changes are included?**
* Ordered lists support
* Nested list support
* Paragraphs within lists
* Pixel perfect alignment for blocks within lists

## Additional Context

* WIP PR right now, this was fully vibe coded, but well tested -- I'll do some cleanup for the commit history & code before requesting review
* The attached `list-test.zip` file is an epub (needs renaming, but you can't upload epubs to GH), containing a bunch of tests for lists
[list-test.zip](https://github.com/user-attachments/files/25447294/list-test.zip)

---

### AI Usage

While CrossPoint doesn't have restrictions on AI tools in contributing, please be transparent about their usage as it 
helps set the right context for reviewers.

Did you use AI tools to help write this code? _**< YES >**_

PR #965, opened 2026-02-18
fix: Fixed formatting of paragraph under a list item

sam-lord:
## Summary

**What is the goal of this PR?**
* Fixing the layout of paragraphs inside list items (Issue #956)

**What changes are included?**
* Using the existing `elemUntilDepth` pattern, adds a state for when a paragraph is is a list item
* When in a paragraph inside a list item, skip creating a new text block

---

### AI Usage

While CrossPoint doesn't have restrictions on AI tools in contributing, please be transparent about their usage as it 
helps set the right context for reviewers.

Did you use AI tools to help write this code? _**< YES >**_

PR #2589, opened 2026-07-13
fix: keep list item bullet inline with nested paragraph text

### LINKED ISSUE #956: Paragraphs in bullet points end up on a new line
### Affected Version

v1.0.0

### Describe the Bug

An epub with the following:

```html
<ul>
<li>
<p>The fundamental difference between changes to the behavior of a system and changes to its structure</p>
</li>
<li>
<p>The enabling magic of alternating investment in structure and investment in behavior, as a lone programmer changing code</p>
</li>
<li>
<p>The basics of the theory of how software design works and the forces that act 
<span class="keep-together">on it</span></p>
</li>
</ul>
```

Would be displayed as a series of bullets, each followed by a new line, and then the paragraph. 

### Steps to Reproduce

1. Create an epub containing the HTML I pasted in the issue
2. View it in CrossPoint
3. Note that each bullet is followed by a line break

### Expected Behavior

Each bullet should have the respective text in front of it, without the line break.

```shell

```

jan-xyz:
## Summary

  Fix issue #956: when `<li>` contains a block-level child like `<p>`, the bullet `•` was rendered on its own line above the paragraph text instead of inline with it.

  - Added a `listItemBulletOnly` flag to `ChapterHtmlSlimParser` that tracks when the current text block contains only the `<li>` bullet
  - In `startNewTextBlock()`: when a nested block tag (`<p>`, `<div>`, etc.) opens and the current block is bullet-only, the block is reused (styles merged) instead of being flushed via `makePages()` — same as the existing empty-block path
  - The flag is cleared when real text arrives (`flushPartWordBuffer`), when a new block is created, and when `</li>` closes (covering empty `<li></li>`)
  - Added 3 test chapters to the `test_br_section_break.epub` generator covering `<li><p>` with bold/italic, nested `<ul>`, and direct-text baseline

## Root cause

Both `<li>` and `<p>` are in `BLOCK_TAGS`. When `<li>` opens, it creates a new block and adds the bullet `•` (block becomes non-empty). When the nested `<p>` opens, `startNewTextBlock()` sees a non-empty block → calls `makePages()` → the bullet gets flushed to its own `PageLine` → a new block is created for the `<p>` text. Result: bullet and text on separate lines.

## How this fix handles the edge cases raised in prior PR reviews

| Edge case | Behavior |
|---|---|
| `<li><p>text</p></li>` | Bullet and text in same block (fixed) |
| Multiple `<p>` in one `<li>` | First `<p>` merges with bullet; second `<p>` sees `listItemBulletOnly=false` → `makePages()` flushes → new block (correct paragraph separation) |
| Nested `<ul>` inside `<li>` | Inner `<li>` triggers `startNewTextBlock`; outer block is non-empty, `listItemBulletOnly=false` → outer content flushed, inner `<li>` starts fresh |
| `<li><div>text</div></li>` | Any block tag reuses the bullet-only block, not just `<p>` |
| Empty `<li></li>` | Flag cleared at `</li>`, next `<li>` flushes normally |

## Comparison to prior PRs

This is a minimal fix (21 lines) that resolves the specific bug in #956 without the scope of prior attempts:

- **#965** — depth-based suppression (`listItemUntilDepth`). Closed because it merged multiple `<p>` in one `<li>` and didn't handle nested blocks. Our flag-based approach clears after the first nested block, so subsequent blocks flush normally.
- **#1041** — full list rendering overhaul (ordered lists, hanging indents, `ListContext` stack, `TextBlock` serialization changes). Closed when author's device broke. Much higher risk (cache format change).
- **#2264** — built on #1041's ideas. Closed, no maintainer review. Same cache format concerns.

This PR intentionally does **not** include ordered list numbering, hanging indents, or per-depth indentation — those are features that could be follow-up PRs. The fix here has no cache format changes, no serialization impact, and minimal new state.

Fixes #956
Supersedes #965, #1041, #2264

---

### AI Usage


Uri-Tauber (approved):
Neat!

PR #2593, opened 2026-07-13
feat: add Turkish hyphenation support

metoli86:
Adds Turkish to the hyphenation language registry:

- `hyph-tr.trie.h` generated from typst/hypher via the existing `scripts/update_hyphenation.sh` flow (`process tr` added). The trie payload is only 526 B — Turkish hyphenation is highly regular.
- A Turkish-specific lowercasing function (`toLowerTurkish`) in HyphenationCommon: the generic Latin Extended-A pairing maps dotted/dotless I the wrong way around for Turkish (`I`→`i`, `İ`→`ı`); Turkish needs `I`→`ı` (U+0131) and `İ` (U+0130)→`i`. Other languages are unaffected — the generic `toLowerLatin` is unchanged.
- Evaluation data: 3,609 words extracted from public-domain Ömer Seyfettin short stories (Turkish Wikisource), with ground truth generated from the deterministic TDK syllabification rule (boundary before the last consonant of each intervocalic cluster; between adjacent vowels) — deliberately independent of the TeX patterns being tested. Measured F1 is **99.76%**; the test threshold is set ~1pp below (98.75) per the existing convention.

`ctest` passes (all hyphenation evals including the new Turkish one), full `pio run` build OK.

Note: not yet tested on hardware.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2595, opened 2026-07-13
feat: offline dictionary lookup in the reader

metoli86:
Implements the in-scope "Dictionary Lookup" roadmap item (README "coming soon" / SCOPE.md Reference Tools): Kindle-style inline word lookup without leaving the reader, fully offline.

**Note: includes the commits of #2593** (the lookup reuses its Turkish-aware lowercasing helper). Happy to rebase once that merges — the diff then shrinks to the dictionary feature itself.

## UX

Reader Menu → **Dictionary** enters word-selection mode: a frame is drawn around the first word of the page. Left/Right (or the side buttons) move by word, Up/Down by line (nearest word by horizontal center), Confirm looks the word up and opens a scrollable definition popup (Up/Down scroll, Back closes), Back leaves the mode. Punctuation around the selected word is ignored, matching is case-insensitive, and on an exact miss the lookup falls back to the longest indexed key that is a prefix of the word — inflected forms like *kitaplarımızdan* resolve to *kitap*, with the matched headword shown as the popup title. See the new `docs/dictionaries.md` and the USER_GUIDE section.

## Design notes (RAM budget)

- **`.cpd` format + on-SD binary search.** Dictionaries are converted on a PC with the new `scripts/make_dictionary.py` (StarDict and dictd input, stdlib only) into a sorted fixed-width index + plaintext blob. Lookups binary-search the index directly on the SD card — one 36-byte read per probe, ~16 probes for the 36k-entry FreeDict en→tr — so RAM usage is O(1) in dictionary size. Nothing is ever bulk-loaded.
- **Overlay, not re-render.** The selection highlight and popup snapshot the pixels underneath via `copyRegionToBuffer` and restore them on close (`FAST_REFRESH` updates), so moving the selection never re-renders the page. If a snapshot allocation fails, the code falls back to a full page re-render on exit.
- **No layout changes.** Word bounding boxes are derived from the existing `Page`/`TextBlock` data (`wordXpos` + `getTextWidth`); the page is re-loaded from the section cache and kept alive only while the mode is active.
- **Deferred entry.** Entering from the reader menu is deferred to the end of the next `render()`, since the menu close triggers a full re-render that would wipe an overlay drawn earlier.

## Testing

- 12 host unit tests (`test/dictionary/`, wired into CTest) against a checked-in fixture: header validation and corruption rejection, exact/case-insensitive/Turkish-normalized lookups, duplicate-headword merging, prefix fallback (longest-stem selection), first/last-entry boundaries, miss behavior.
- **Tested on X4 hardware** with the FreeDict English–Turkish dictionary (34,405 entries, 4.7 MB): selection navigation, lookups, popup scrolling, and the no-dictionary message all work; RAM usage in the build is unchanged (15.2%).
- `pio run` clean, clang-format applied.


Uri-Tauber:
Duplicate of #2583.

metoli86:
**Hardware verification round complete** — exercised the full feature end-to-end on an X4 (v1.4.1 base, FreeDict en→tr, 34k entries). Real-device testing surfaced three issues that host tests could not catch, all fixed in the commits above:

1. `fix: don't close a never-opened dictionary file` — `HalFile::close()` asserts on a never-opened file; `reset()` runs on every reader render, so opening **any** book panicked the device. Reproduced via crash_report.txt, fixed with an `isOpen()` guard.
2. `fix: word-box and popup coordinates` — `drawText`'s `y` is the **top** of the text line, not the baseline (TextBlock draws underlines at `y + ascender + 2`), so the highlight frames sat one ascender above the words and popup text overflowed its frame. Also added periodic HALF refreshes (every 8 selection moves, on popup close and on mode exit) to clear e-ink ghosting from the fast outline updates.
3. `fix: skip hidden files and keep scanning` — macOS Finder copies leave AppleDouble `._*.cpd` junk next to the real dictionary; the scan now skips dotfiles and tries every candidate instead of giving up on the first unparseable file.

Current state on device: selection navigation, exact/prefix lookups, popup scrolling, no-dictionary message, mode exit and ghost cleanup all verified working. CI is green.

metoli86:
Fair call — #2583 predates this PR, consolidates months of community work, and reading it through, using StarDict directly (no PC-side conversion step) is the better UX for users. Closing in its favor.

Three pieces of this PR may still be worth carrying over into #2583's architecture as follow-ups, since they target languages the mini-stemmer can't reach:

1. **Longest-prefix stem fallback for agglutinative languages** — on an exact miss, walk back from the bisect insertion point to the longest indexed key that is a prefix of the word (min 3 bytes, bounded scan). Turkish *kitaplarımızdan* → *kitap*, Finnish/Hungarian get the same benefit; suffix-stripping rules can't cover these. It's ~30 lines against a sorted index and composes with the `.qidx` sampling.
2. **Turkish dotted/dotless-I normalization** — `I`→`ı`, `İ`→`i` during headword matching (the generic Latin mapping folds them the wrong way for Turkish headwords/selections).
3. **X4 device-test findings** that may save others a round-trip: `HalFile::close()` asserts on a never-opened file (bit me in a cleanup path), `drawText`'s `y` is the top of the line rather than the baseline (bit my word-highlight boxes), and dictionary folder scans need to skip macOS `._*` AppleDouble junk or Finder-copied dictionaries fail to load.

Happy to send 1+2 as a small PR on top of #2583 once it lands, and to test #2583 on X4 hardware in the meantime. Thanks for the consolidation work — the slim-slice approach makes this actually reviewable.

PR #2603, opened 2026-07-14
fix: fall back to NCX when EPUB 3 nav TOC is sparse

Cabalist:
## Summary

* **What is the goal of this PR?** Show the full chapter list for books whose EPUB 3 nav TOC is coarse while their NCX carries the real chapters. Converted books (e.g. Amazon exports like *A Parade of Horribles*) ship a `nav.xhtml` with only part/front-matter entries (15) alongside an NCX with every chapter (125). Since the firmware prefers nav and only falls back to NCX when nav *parsing fails*, these books get a chapterless TOC and the reader header shows the part title for most of the book.

  - `Epub.cpp` (TOC pass): richness-based source selection. After the nav parses, if its entry count covers less than half the spine and an NCX exists, the TOC pass restarts and parses the NCX; the NCX is adopted only if it yields *more* entries, otherwise the nav is re-parsed and kept. The decision is logged (`Nav TOC sparse (15 entries for 126 spine items), trying NCX` / `NCX TOC adopted (125 entries)`).
  - `BookMetadataCache.cpp`: `beginTocPass()` now resets the TOC entry counter, making the pass restartable. This also fixes a latent bug: previously, a nav that errored mid-parse *after* emitting entries left them in the temp file, and the NCX fallback **appended** to them — producing a duplicated TOC with an inflated count.

## Additional Context

* **Not a regression** — investigated as one and empirically disproved: a debug build of the 1.4.1 tag fresh-indexes the same book to the same 15 entries on-device (tested with both plain and "Optimize" web uploads; the optimizer doesn't touch nav/NCX/OPF navigation data). This is longstanding selection behavior.
* **Scope of the heuristic:** fires only when nav coverage is below 50% of the spine count. Books without an NCX, or with adequate navs, are completely untouched. If the NCX turns out sparser than the nav (or fails), the nav is restored. Nested NCX entries need no UI work — entries already carry a `level` and the chapter-selection screen indents by it.
* **Cost:** one or two extra streamed TOC parses (~200 ms each), once, at first index of an affected book. No RAM impact, no cache-format change, no version bump — books already indexed keep their current TOC unless their cache is rebuilt.
* **Verification (X4, this book):** serial shows `Nav TOC sparse (15 entries for 126 spine items), trying NCX` → `NCX TOC adopted (125 entries)` → `Wrote 126 spine, 125 TOC entries`; the TOC screen lists all chapters ("Heat 1 of 7", "Chapter 1", …) nested under the part entries, and jump-to-chapter works. `pio run` zero warnings; clang-format applied.
* **Review focus:** the pass-restart flow (`endTocPass()`/`beginTocPass()` error paths) and the adoption condition (`ncxParsed && (!navParsed || ncxCount > navCount)`).

---

### AI Usage

Diagnosed and implemented with Claude Code (Claude Fable 5); on-device flashing and visual verification done by hand.


barbarhan (changes requested):
The NCX adoption decision compares raw TOC entry counts, but `createTocEntry()` increments `tocCount` even when its href cannot be matched to the spine and `spineIndex` remains `-1`.

That means a valid sparse nav can be replaced by a longer stale/unmapped NCX: for example, five working nav entries versus eight NCX entries whose old URIs no longer match the spine. The condition here sees `8 > 5` and adopts the NCX, but selecting any of those entries in `EpubReaderChapterSelectionActivity` immediately cancels because `spineIndex == -1`.

Please compare mapped/usable entry counts (or otherwise require the candidate NCX entries to resolve to the spine) before allowing NCX to supersede a successfully parsed nav TOC. A regression fixture with a valid nav plus a longer stale NCX would cover this deterministic case.

Cabalist:
Updated!  Thanks for the review.  :) 

  The nav-vs-NCX TOC decision now compares spine-mapped entry counts instead of raw entry counts.                                                                                                                    
                                                                                                                                                                                                                     
  - BookMetadataCache tracks a second counter, tocMappedCount, incremented in createTocEntry() only when the entry's href actually resolved to a spine item (spineIndex >= 0). It resets with tocCount on each pass  
  restart and is never written to book.bin.                                                                                                                                                                          
  - The two decisions in Epub::load() — "is the nav sparse?" and "should the NCX replace it?" — were extracted into pure constexpr predicates in a new header, TocFallbackPolicy.h, and both now take mapped counts. 
  - A new host gtest suite (test/toc_fallback/, 14 cases, runs in CI) pins the decision table, headlined by the reported scenario: nav with 5 mapped entries vs NCX with 0 mapped → keep the nav.

barbarhan (approved):
Re-reviewed the latest PR head (`d49f9bd`). The mapped-entry accounting addresses my earlier blocker: a longer stale NCX with unresolved spine links can no longer replace a smaller usable nav TOC, while a genuinely richer NCX still wins. The pass counters are reset consistently, and the decision predicates cover the relevant failure/tie cases.

Verification:

- PR head: all 135 host tests pass, including the 14 TOC fallback tests.
- Local integration with current `develop`: the only conflict is the append-only list in `test/CMakeLists.txt`. Keeping `add_subdirectory(toc_fallback)` alongside the newer test directories resolves it cleanly; all 152 host tests then pass.

No remaining issue with the requested change. Approving. GitHub still marks the branch as conflicting, so it needs a refresh against `develop` before merge.

PR #2605, opened 2026-07-14
feat: unified Text Settings screen with live preview

PaulDelestrac:
## Summary
* Consolidates the reader's text-appearance settings into a single **Text Settings** screen with a shared live preview pane with 4 tabs: Font, Size, Layout and Style.
  * **Font** and **Size**: pick-one lists, applied on Confirm, current value tagged.
  * **Layout** (Line Spacing, Extra Paragraph Spacing, Alignment, Screen Margin): multi-value rows open the standard value picker, the toggle flips in place.
  * **Style** (Focus Reading, Hyphenation, Embedded Style, Anti-Aliasing): moved in from the flat Reader list so all text-appearance settings live together.
  * Merged settings are hidden from **Settings › Reader** but kept in the shared list, so the web/JSON settings API is unchanged.
* The preview mirrors the reader's actual font family, font size, layout and focus reading setting.
  * A **"Not shown in preview"** caption appears for settings the preview can't reflect (Hyphenation, Embedded Style, Anti-Aliasing).

### Known limitations
* Anti-Aliasing and Embedded Style aren't reflected in the preview (hence the caption).
* Hyphenation is applied through the reader engine, but only visibly changes the sample when a word actually breaks at the preview width, so it stays flagged as not-shown.

### Update (addressing review)
* **Preview now routed through the real reader engine** (`ParsedText` + `TextBlock`) instead of the hand-rolled layout — justification, alignment, kerning, **RTL**, CJK, hyphenation and focus reading are now pixel-exact with the book. Cached and glyph-prewarmed so navigation stays responsive.
* **Touch support** added (tab tap, list tap/drag, swipe paging); inert on non-touch boards.
* Folded settings now use a self-describing `inTextSettings` flag instead of a hardcoded skip-list; minor fixes: SD-font unload on family switch, preview label spacing derived from theme metrics.
* Merged latest `develop`.

## Screenshots
Screenshot 2026-07-18 at 01-28-48
Screenshot 2026-07-18 at 01-29-18
Screenshot 2026-07-18 at 01-29-52
Screenshot 2026-07-18 at 01-31-26

---
### AI Usage
AI use for assisting in coding and in writing the PR description.


Uri-Tauber:
Hey @PaulDelestrac, thanks for the contribution—this is a really nice addition!

One UX concern I have is that this new pane feels like a duplicate of the existing font family preview pane. Instead of having two separate panes, I think it would be make sense to consolidate them into a single **"Reader Font"** submenu, with two tabs beneath the preview pane for the different settings.

What do you think?

PaulDelestrac:
> I think it would be make sense to consolidate them into a single **"Reader Font"** submenu, with two tabs beneath the preview pane for the different settings.

@Uri-Tauber I agree conceptually, let me see if I can come up with some nice UI/UX to merge both previews!

PaulDelestrac:
# Updated implementation

## Summary
- Created a unified ReaderFontActivity which combines font family and size pickers with a shared preview panel (as suggested by @Uri-Tauber https://github.com/crosspoint-reader/crosspoint-reader/pull/2605#issuecomment-4972068114)
  - Merged the two setting rows in one "Reader Font" row, showing also the currently set values on the right (with S, M, L and XL abbreviations for size)
  - In the ReaderFontActivity, made changes to the UX
    - Clicking 'Confirm' will apply the selection instantly instead of just previewing the font; it will not automatically close the activity so that the user can see the font preview and maybe select a size (or vice versa)
    - Clicking 'Back' closes the activity (the last selected font stays applied)
    - Long pressing 'Up' or 'Down' allows for switching tabs between 'Family' and 'Size'

Screenshot 2026-07-17 at 11-24-18
Screenshot 2026-07-17 at 11-24-36
Screenshot 2026-07-17 at 11-24-45

## Potential Issues
- Size abbreviations are only implemented in English for now
- Very long font family names can cut into the setting text with certain themes (e.g., see below with RoundedRaff)
  - This could be fixed by changing the truncation policy to prioritize the settings name (@Uri-Tauber)?
Screenshot 2026-07-17 at 11-25-26

## Follow-up
- More settings could be merged with this new 'Reader Font' activity like Line spacing or Focused Reading or Anti Aliasing etc. (see below some prototypes)
Screenshot 2026-07-17 at 11-42-25
Screenshot 2026-07-17 at 11-43-02
Screenshot 2026-07-17 at 11-43-24

Uri-Tauber:
@PaulDelestrac It looks fantastic on my device!
Regarding the truncation, maybe we should change the entry from "reader font" to "font settings" and remove the _font,size_ info from there? Similar to "Customize status bar" entry.
Moving spacing and style settings seems like a very good idea. Your call if you want to do it now, or do a seperate PR.

I haven't reviewed the code yet, I will probably I will have time for this tomorrow evening.

PaulDelestrac:
> maybe we should change the entry from "reader font" to "font settings" and remove the _font,size_ info from there? 

I think it makes sense, especially if we move spacing and style settings in this common view.

> Your call if you want to do it now, or do a seperate PR.
> I haven't reviewed the code yet, I will probably I will have time for this tomorrow evening.

I'll try to include the changes in this PR before then so you can review the whole thing at once. If I don't find the time, then I'll go for a second PR :)

PaulDelestrac:
pushed green implementation + updated PR description

Uri-Tauber:
@PaulDelestrac Ready for review?

Uri-Tauber:
Something is weird with the PR diff of `lib/I18n/translations/slovak.yaml`

PaulDelestrac:
> Something is weird with the PR diff of `lib/I18n/translations/slovak.yaml`

@Uri-Tauber yes, weirdly `slovak.yaml` was the only file with CRLF line endings before this PR, all other were just LF. So I made it the same as the rest of the I18n files.
If you want to see the diff without the whitespaces, there is an option at the top of the GH page

Screenshot 2026-07-20 at 11-35-21

src/activities/settings/TextSettingsActivity.cpp
```diff
@@ -0,0 +1,386 @@
...
+  } else if (registry_) {
+    const int sdIdx = font.settingIndex - CrossPointSettings::BUILTIN_FONT_COUNT;
+    const auto& families = registry_->getFamilies();
+    if (sdIdx < static_cast<int>(families.size())) {
+      strncpy(SETTINGS.sdFontFamilyName, families[sdIdx].name.c_str(), sizeof(SETTINGS.sdFontFamilyName) - 1);
+      SETTINGS.sdFontFamilyName[sizeof(SETTINGS.sdFontFamilyName) - 1] = '\0';
+      sdFontSystem.ensureLoaded(renderer);
+    }
+  }
+}
```
Uri-Tauber, line 309: Two issues here:

1. **Builtin branch never unloads the SD font.** `sdFontSystem.ensureLoaded(renderer)` is the only path that unloads a resident SD font once the wanted family is empty. Switching **SD font → Noto** leaves the SD font's glyph data resident while the user keeps previewing in this activity.
2. **`currentFamilyIndex_` is updated before validation.** If the SD branch guard fails, the list shows the new font as *Selected* and the preview label names it while `SETTINGS` still holds the old family.

```suggestion
void TextSettingsActivity::applyFamily(int listIndex) {
  const auto& font = fonts_[listIndex];
  if (font.isBuiltin) {
    SETTINGS.fontFamily = font.settingIndex;
    SETTINGS.sdFontFamilyName[0] = '\0';
    sdFontSystem.ensureLoaded(renderer);  // unloads the previously resident SD font
    currentFamilyIndex_ = listIndex;
  } else if (registry_) {
    const int sdIdx = font.settingIndex - CrossPointSettings::BUILTIN_FONT_COUNT;
    const auto& families = registry_->getFamilies();
    if (sdIdx < static_cast<int>(families.size())) {
      strncpy(SETTINGS.sdFontFamilyName, families[sdIdx].name.c_str(), sizeof(SETTINGS.sdFontFamilyName) - 1);
      SETTINGS.sdFontFamilyName[sizeof(SETTINGS.sdFontFamilyName) - 1] = '\0';
      sdFontSystem.ensureLoaded(renderer);
      currentFamilyIndex_ = listIndex;
    }
  }
}
```

src/activities/settings/TextSettingsPreview.cpp
```diff
@@ -0,0 +1,179 @@
...
+  if (boldChars >= charCount) return wordBytes;
+  size_t chars = 0;
+  for (size_t i = 0; i < wordBytes; i++) {
+    if ((wordStart[i] & 0xC0) != 0x80) {
+      if (chars == boldChars) return i;
+      chars++;
+    }
+  }
+  return wordBytes;
+}
```
Uri-Tauber: Could we route this via the existing reader's layout engine. I don't like the duplication here.

src/activities/settings/SettingsActivity.cpp
```diff
@@ -44,6 +44,15 @@ void SettingsActivity::rebuildSettingsLists() {
...
     } else if (setting.category == StrId::STR_CAT_READER) {
+      // Settings merged into "Text Settings"
+      // (they stay in the shared list for the web settings API)
+      if (setting.nameId == StrId::STR_FONT_FAMILY || setting.nameId == StrId::STR_FONT_SIZE ||
+          setting.nameId == StrId::STR_LINE_SPACING || setting.nameId == StrId::STR_EXTRA_SPACING ||
+          setting.nameId == StrId::STR_PARA_ALIGNMENT || setting.nameId == StrId::STR_SCREEN_MARGIN ||
+          setting.nameId == StrId::STR_FOCUS_READING || setting.nameId == StrId::STR_HYPHENATION ||
+          setting.nameId == StrId::STR_EMBEDDED_STYLE || setting.nameId == StrId::STR_TEXT_AA) {
+        continue;
+      }
```
Uri-Tauber: I dont like the duplication between here and `settingslist`. can you find more elegant solution?

src/activities/settings/TextSettingsPreview.cpp
```diff
@@ -0,0 +1,179 @@
...
+namespace textsettings {
+
+void renderPreview(GfxRenderer& renderer, int previewPadding, int top, int height, const char* familyName,
+                   const char* sizeName) {
+  const int left = previewPadding;
+  const int width = renderer.getScreenWidth() - (previewPadding * 2);
+  if (width <= 0 || height <= 0) return;
+
+  const int labelH = renderer.getTextHeight(UI_10_FONT_ID);
+  const int labelGap = 4;
```
Uri-Tauber: Maybe derive from `themeMetrics`, rather than hardcode specific value?

lib/I18n/translations/english.yaml
```diff
@@ -88,12 +88,20 @@ STR_LONG_PRESS_BEHAVIOR_OFF: "OFF"
 STR_LONG_PRESS_BEHAVIOR_SKIP: "Chapter skip"
 STR_LONG_PRESS_BEHAVIOR_ORIENTATION: "Orientation change"
 STR_LONG_PRESS_MENU: "Long-press Menu"
-STR_FONT_PREVIEW_TEXT: "The quick brown fox jumps over the lazy dog"
+STR_FONT_PREVIEW_TEXT: "The quick brown fox jumps over the lazy dog."
+STR_FONT_PREVIEW_TEXT_EXT: "Pack my box with five dozen liquor jugs."
```
Uri-Tauber: How about displaying `STR_FONT_PREVIEW_TEXT` twice instead of introducing a new string across ~30 languages? It would avoid adding ~30 new localized strings, saving several KB.

src/activities/settings/TextSettingsPreview.cpp
```diff
@@ -0,0 +1,179 @@
...
+      snprintf(buf, sizeof(buf), "%.*s", static_cast<int>(wordBytes - boldBytes), wordStart + boldBytes);
+      w += renderer.getTextAdvanceX(fontId, buf, EpdFontFamily::REGULAR);
+    }
+  }
+  return w;
+}
+
+// Draws a line word by word: adds extraSpace to each gap (justification) and
+// applies the focus-reading bold split per word.
+void drawLine(const GfxRenderer& renderer, int fontId, int x, int y, const char* line, bool focusReading,
```
Uri-Tauber: drawLine() and lineStartX() always lay text out left-to-right, while the real reader auto-detects RTL paragraphs and lays them out right-to-left.
It was also the case in the previous design - not something you introduced, but it supports my other comment about routing via the reader layout parser.

Uri-Tauber:
@PaulDelestrac I hope my comments aren't too much!
I really want to get this feature in, but fixing those now can save a lot of debugging headache later.

PaulDelestrac:
> @PaulDelestrac I hope my comments aren't too much! I really want to get this feature in, but fixing those now can save a lot of debugging headache later.

@Uri-Tauber Thanks for taking the time to review! I totally get it, let's take the time to make this robust :)
I'll go through your comments tomorrow.

Uri-Tauber:
I'm afraid merging the touch support branch introduce quite a few conflicts here. Sorry for the mess.

src/activities/settings/SettingsActivity.cpp
PaulDelestrac: > I dont like the duplication between here and `settingslist`. can you find more elegant solution?

I addressed in commit https://github.com/crosspoint-reader/crosspoint-reader/pull/2605/commits/45d4063df3a38edca46c8d0cfdf5a4698da95550 by using a flag, I wanted to keep those settings into the READER category instead of adding a new one so that they appear in the current version of the WebUI without any change needed.
I hope the flag approach is a good-enough middle ground.

PaulDelestrac:
Updated the implementation. Some points that may be discussed:
- the deduplication of the settings (see this comment: https://github.com/crosspoint-reader/crosspoint-reader/pull/2605#discussion_r3631615751) -> left the comment unresolved for now
- for the use of the rendering engine, I used ParsedText and TextBlock. It seems to work well on my X4, you may want to test it on an X3 if you have one around. I tried in the simulator, but we never know.
- I merged the develop branch and added a handleTouch function that I call in the loop to try and support touch on this new screen
  - @Uri-Tauber regarding the support of touch in tab bars, I followed the implementation in SettingsActivity, but this might deserve a separated touch helper (similarly to handleListTouch)

+ updated PR description

Uri-Tauber:
@PaulDelestrac Awesome. Thanks.
I will do a final review later today.

Uri-Tauber:
Tested on my device. If I choose several SD fonts one after the other, the device crash on the 9th one. Its reproducible - I get the same crash for the same actions.
I will try to dig into this now. 

Here's the crash log:

```
[22:22:08] [DBG] [GFX] Time = 90 ms from clearScreen to displayBuffer
[22:22:10] [DBG] [SDFS] Reloading Alef: size 12 -> 14 (enum 1)
[22:22:10] [DBG] [SDCF] Loaded: /fonts/Alef/Alef_14.cpfont (v4, 2 styles)
[22:22:10] [DBG] [SDCF]   style[0]: 49 intervals, 411 glyphs, advY=40, asc=30, desc=-11, kernL=0, kernR=0, ligs=2
[22:22:10] [DBG] [SDCF]   style[1]: 49 intervals, 411 glyphs, advY=40, asc=30, desc=-11, kernL=0, kernR=0, ligs=2
[22:22:10] [DBG] [SDMGR] Loaded /fonts/Alef/Alef_14.cpfont size=14 id=-471696821 styles=2 (sizeEnum=1)
[22:22:10] [DBG] [SDFS] Loaded SD font family: Alef
[22:22:10] [DBG] [SDCF] Kern classes + lig loaded: kernL=0, kernR=0, ligs=2
[22:22:10] [DBG] [FCM] prewarmCache(SD): 1 glyph(s) not found (styleMask=0x01)
[22:22:10] [DBG] [SDCF] Advance table style 0: +30 from SD, total=30/768
[22:22:10] [DBG] [GFX] Time = 91 ms from clearScreen to displayBuffer
[22:22:11] [DBG] [FCM] prewarmCache(SD): 1 glyph(s) not found (styleMask=0x01)
[22:22:11] [DBG] [GFX] Time = 54 ms from clearScreen to displayBuffer
[22:22:12] [DBG] [SDCF] Loaded: /fonts/FrankRuehlCLM/FrankRuehlCLM_14.cpfont (v4, 4 styles)
[22:22:12] [DBG] [SDCF]   style[0]: 53 intervals, 335 glyphs, advY=31, asc=24, desc=-8, kernL=87, kernR=86, ligs=0
[22:22:12] [DBG] [SDCF]   style[1]: 53 intervals, 335 glyphs, advY=31, asc=24, desc=-8, kernL=81, kernR=79, ligs=0
[22:22:12] [DBG] [SDCF]   style[2]: 53 intervals, 335 glyphs, advY=31, asc=24, desc=-8, kernL=87, kernR=86, ligs=0
[22:22:12] [DBG] [SDCF]   style[3]: 53 intervals, 335 glyphs, advY=31, asc=24, desc=-8, kernL=82, kernR=79, ligs=0
[22:22:12] [DBG] [SDMGR] Loaded /fonts/FrankRuehlCLM/FrankRuehlCLM_14.cpfont size=14 id=-2079643216 styles=4 (sizeEnum=1)
[22:22:12] [DBG] [SDFS] Loaded SD font family: FrankRuehlCLM
[22:22:12] [DBG] [LOOP] New max loop duration: 93 ms (activity: 92 ms)
[22:22:12] [DBG] [SDCF] Kern classes + lig loaded: kernL=87, kernR=86, ligs=0
[22:22:12] [DBG] [SDCF] Built mini kern: 25×28 matrix (700 bytes, full was 87×82 = 7134 bytes)
[22:22:12] [DBG] [FCM] prewarmCache(SD): 1 glyph(s) not found (styleMask=0x01)
[22:22:12] [DBG] [SDCF] Advance table style 0: +30 from SD, total=30/768
[22:22:12] [DBG] [GFX] Time = 122 ms from clearScreen to displayBuffer
[22:22:15] [DBG] [SDCF] Loaded: /fonts/Alef/Alef_14.cpfont (v4, 2 styles)
[22:22:15] [DBG] [SDCF]   style[0]: 49 intervals, 411 glyphs, advY=40, asc=30, desc=-11, kernL=0, kernR=0, ligs=2
[22:22:15] [DBG] [SDCF]   style[1]: 49 intervals, 411 glyphs, advY=40, asc=30, desc=-11, kernL=0, kernR=0, ligs=2
[22:22:15] [DBG] [SDMGR] Loaded /fonts/Alef/Alef_14.cpfont size=14 id=-471696821 styles=2 (sizeEnum=1)
[22:22:15] [DBG] [SDFS] Loaded SD font family: Alef
[22:22:15] [DBG] [SDCF] Kern classes + lig loaded: kernL=0, kernR=0, ligs=2
[22:22:15] [DBG] [FCM] prewarmCache(SD): 1 glyph(s) not found (styleMask=0x01)
[22:22:15] [DBG] [SDCF] Advance table style 0: +30 from SD, total=30/768
[22:22:15] [DBG] [GFX] Time = 95 ms from clearScreen to displayBuffer
Guru Meditation Error: Core  1 panic'ed (StoreProhibited). Exception was unhandled.
Core  1 register dump:
PC      : 0x40056f60  PS      : 0x00060630  A0      : 0x820076d4  A1      : 0x3fcb0fa0
A2      : 0x00000000  A3      : 0x3fc9e654  A4      : 0x00000010  A5      : 0x00000000
A6      : 0x4f465043  A7      : 0x0000544e  A8      : 0x00000000  A9      : 0x3fcb0f70
A10     : 0x00000001  A11     : 0x00059200  A12     : 0x3fc9e654  A13     : 0x00060023
A14     : 0x00000001  A15     : 0x0000abab  SAR     : 0x00000018  EXCCAUSE: 0x0000001d
EXCVADDR: 0x00000000  LBEG    : 0x40056f5c  LEND    : 0x40056f72  LCOUNT  : 0x00000000
Backtrace: 0x40056f5d:0x3fcb0fa0 0x420076d1:0x3fcb0fb0 0x42009a55:0x3fcb0ff0 0x42037fb1:0x3fcb1010 0x4203b176:0x3fcb1030 0x4203b5cf:0x3fcb10a0 0x4204177a:0x3fcb10f0 0x420b680f:0x3fcb1110 0x420b51b2:0x3fcb1350 0x420b5767:0x3fcb1450 0x4208e5da:0x3fcb1470 0x42182471:0x3fcb14a0 0x4037f5dd:0x3fcb14c0
ELF file SHA256: 0ae1451e6
Rebooting...
���ESP-ROM:esp32s3-20210327
Build:Mar 27 2021
rst:0xc (RTC_SW_CPU_RST),boot:0xa (SPI_FAST_FLASH_BOOT)
Saved PC:0x4037b3b6
SPIWP:0xee
mode:DIO, clock div:1
load:0x3fce2820,len:0xfb0
load:0x403c8700,len:0xaf4
load:0x403cb700,len:0x2e34
entry 0x403c8898
```

Uri-Tauber:
Found it!
Data race between the two cores of the ESP32-S3. Easy fix. 
I'm going to apply it myself so we can merge this now.

Uri-Tauber (approved):

PaulDelestrac:
> Found it!
> 
> Data race between the two cores of the ESP32-S3. Easy fix. 
> 
> I'm going to apply it myself so we can merge this now.

Great @Uri-Tauber !
Thanks for your reviews and for your time!

PR #2606, opened 2026-07-14
fix: reduce CSS parse-time OOM risk in chapter layout

brianhuster:
Continues from https://github.com/crosspoint-reader/crosspoint-reader/pull/2604#issuecomment-4970178049

## Summary

**What is the goal of this PR?** Reduce heap pressure during chapter parsing when embedded CSS is enabled, to  lower vector reallocation spikes in CSS-heavy content and helps prevent operator new OOM aborts during section build.

- Add shared token-capacity reservation in ParsedText::addWord() and apply it to both CJK token splitting and focus-reading tokenization paths.
- Replace ad-hoc per-branch reserve logic with ensureTokenCapacity().
- Add CSS-aware text block flush thresholds in ChapterHtmlSlimParser: soft flush at 320 words with CSS, 750 without CSS

## Additional Context

This PR was created because of this crash 
[crash_report.txt](https://github.com/user-attachments/files/30017903/crash_report.txt)

@Uri-Tauber with this PR, I can now open any chapter in that osanaikoro.epub file

---

### AI Usage

Did you use AI tools to help write this code? _**YES**_, however it was cleaned up by me

### DIFF
--- a/lib/Epub/Epub/parsers/ChapterHtmlSlimParser.cpp
+++ b/lib/Epub/Epub/parsers/ChapterHtmlSlimParser.cpp
@@ -22,6 +22,17 @@
 constexpr size_t MIN_SIZE_FOR_POPUP = 10 * 1024;  // 10KB
 constexpr size_t PARSE_BUFFER_SIZE = 1024;
 
+// This number comes from PR #73
+// If we have > 750 words buffered up, perform the layout and consume out all but the last line
+// There should be enough here to build out 1-2 full pages and doing this will free up a lot of
+// memory.
+// Spotted when reading Intermezzo, there are some really long text blocks in there.
+constexpr size_t TEXT_BLOCK_SOFT_FLUSH_WORDS = 750;
+
+// When CSS is enabled, flush earlier to save RAM. 320 is still more than enough to build a CJK
+// page at font size 14
+constexpr size_t TEXT_BLOCK_SOFT_FLUSH_WORDS_WITH_CSS = 320;
+
 // Hard cap on the number of anchor IDs recorded per chapter. Legitimate navigation
 // anchors (TOC entries, footnotes, cross-references) rarely exceed a few hundred per
 // chapter. A runaway count usually means a converter injected machine-generated IDs on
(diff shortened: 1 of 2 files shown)


lib/EpdFont/SdCardFont.cpp
```diff
@@ -1204,7 +1204,7 @@ int SdCardFont::buildAdvanceTableRange(Iter begin, Iter end, bool includeSpace,
   unsigned long startMs = millis();
 
   // +2 reserved slots for space and hyphen injected after the main scan.
-  static constexpr uint32_t MAX_UNIQUE_CODEPOINTS = 4096;
+  static constexpr uint32_t MAX_UNIQUE_CODEPOINTS = 1024;
```
Uri-Tauber: I imagine this would be quite the regression for CJK users.

brianhuster: I could be wrong since this is my very first time digging into the code, but the way I understand is that Unique codepoints list is refreshed after each page. And I don't think a page in this very small e-ink screen can have more than 1024 unique characters, at font size 14 it only holds fewer than 300 characters. I think even at font size 8 it won't have more than 600. Korean characters are a bit smaller by width than Chinese or Japanese, but I don't think the screen can display more than 900.

brianhuster: @Uri-Tauber I reverted this very change for now, because it doesn't seem to affect the device's ability to render osanaikoro.epub

lib/Epub/Epub/parsers/ChapterHtmlSlimParser.cpp
```diff
@@ -21,6 +21,8 @@
 // Minimum file size (in bytes) to show indexing popup - smaller chapters don't benefit from it
 constexpr size_t MIN_SIZE_FOR_POPUP = 10 * 1024;  // 10KB
 constexpr size_t PARSE_BUFFER_SIZE = 1024;
+constexpr size_t TEXT_BLOCK_SOFT_FLUSH_WORDS = 750;
+constexpr size_t TEXT_BLOCK_SOFT_FLUSH_WORDS_WITH_CSS = 320;
```
Uri-Tauber, line 34: Overall nice work. 
I would like to know where those two magic numbers come from?
Small comment will be a good idea.

brianhuster: 750 was not chosen by me, but it comes from #73

320 is a random number chosen between 750/2 and 16*19 (the number of characters in a page of [osanaikoro.epub](https://www.daido-life-fd.or.jp/assets/files/ebooks/osanaikoro.epub) when displayed at font size 14)

Uri-Tauber (approved):

PR #2611, opened 2026-07-14
feat: deferred refresh, memory work port, and first-open speedups

### LINKED ISSUE #2535: Repeated crashes on a chapter with heavy footnoting.
### Affected Version

1.4.0-dev-feat-smart-indexing-23f3ff77

### Describe the Bug

Repeated crashes on a chapter with heavy footnoting when I go back and forth between the footnotes and the main body text a few times. Happens on a book which has been stable on 1.4.0 and 1.4.1 but which is heavily footnoted enough to cause problems in the latest CrossInk release (s/o @uxjulia for responding to my bug reports).

### Steps to Reproduce

1. Open the attached epub
2. Navigate to chapter "Scylla and Charybdis"
3. Advance until you see the word "jennyass" and open that footnote.
4. Return to the main text and back to the same footnote page a few times.

[Ulysses_ Annotated Edition (Alma Classics - James Joyce.epub.zip](https://github.com/user-attachments/files/29653880/Ulysses_.Annotated.Edition.Alma.Classics.-.James.Joyce.epub.zip)

### Expected Behavior

Navigating between footnotes and main text should not crash.

```shell
CrossPoint version: 1.4.0-dev-feat-smart-indexing-23f3ff77

Panic reason: abort() was called at PC 0x4217cc53 on core 0

Last logs:
[191985] [DBG] [SDCF] Kern classes + lig loaded: kernL=828, kernR=844, ligs=2
[192007] [DBG] [SDCF] Kern classes + lig loaded: kernL=828, kernR=844, ligs=2
[192028] [DBG] [SDCF] Kern classes + lig loaded: kernL=819, kernR=822, ligs=2
[192049] [DBG] [SDCF] Kern classes + lig loaded: kernL=820, kernR=822, ligs=2
[192109] [DBG] [SDCF] Built mini kern: 57×50 matrix (2850 bytes, full was 162×160 = 25920 bytes)
[192171] [DBG] [SDCF] Built mini kern: 57×51 matrix (2907 bytes, full was 161×160 = 25760 bytes)
[192229] [DBG] [SDCF] Built mini kern: 58×47 matrix (2726 bytes, full was 139×118 = 16402 bytes)
[192286] [DBG] [SDCF] Built mini kern: 58×47 matrix (2726 bytes, full was 139×119 = 16541 bytes)
[192345] [DBG] [GFX] Time = 411 ms from clearScreen to displayBuffer
[194398] [ERR] [ERS] OOM: grayscale strip scratch (8000 bytes); skipping AA this page
[194401] [DBG] [ERS] Rendered page in 2441ms
[194473] [DBG] [ERS] Progress saved: spine=36 page=78
[194574] [DBG] [SCT] Page 80 processed
[1] [DBG] [UI] Using Lyra theme
[276] [INF] [HW] Using cached device type: X4
[276] [INF] [MAIN] Hardware detect: X4


Stack memory:
0x3FCB2970: 0x00000000 0x00000000 0x3FCB2988 0x4038DEF2 0x00000000 0x00000000 0x3FCA0030 0x3FC94A40 
0x3FCB2990: 0x3FCB299C 0x3FC94A5C 0x3FCB2988 0x37313234 0x33356363 0x00000000 0x726F6261 0x20292874 
0x3FCB29B0: 0x20736177 0x6C6C6163 0x61206465 0x43502074 0x34783020 0x63373132 0x20333563 0x63206E6F 
0x3FCB29D0: 0x2065726F 0x00000030 0x3FCA0000 0xCDC6567D 0x00007000 0x00000051 0x3FCCDA74 0x4217CC56 
0x3FCB29F0: 0x00003800 0x00000051 0x3FCCDA74 0x4217CCA2 0x3FCCDA74 0x00000051 0x3FCCDA74 0x4217CDC4 
0x3FCB2A10: 0x00003800 0x00000051 0x4217FA26 0x3C4F1070 0x00003800 0x00000051 0x00007000 0x4217CF0E 
0x3FCB2A30: 0x00000004 0x3FCBBFB8 0x3FCBDC70 0x420596DA 0x00000000 0x3FCAA000 0x3FCBDC7C 0x00000000 
0x3FCB2A50: 0x00000000 0x00000051 0x3FCBDA68 0x42059CC6 0x3FCB2CC4 0x3FCC7550 0x3FCBDA68 0x420595A0 
0x3FCB2A70: 0x3FCB2A78 0x00000009 0x35316770 0x336E6631 0x00000036 0x420580F2 0x42050051 0xCDC6567D 
0x3FCB2A90: 0x00000000 0x00000000 0x41880000 0x00000000 0x3FCB2CC4 0x3FCBDA68 0x3FCC7550 0x42059ED6 
0x3FCB2AB0: 0x41880000 0x3FCB2BF4 0x3FCB2BC4 0x4205823C 0x000000E0 0x3FCB2ACC 0x00000000 0x42058246 
0x3FCB2AD0: 0x41880000 0x3FCB2BF4 0x3FCB2BC4 0x4205836C 0x00000000 0x00000000 0x41880000 0x00000000 
0x3FCB2AF0: 0x00000000 0x3FCBDD84 0x3FCBDA68 0x4205C004 0x3FCDE194 0x000001B2 0x00000000 0xBFC00000 
0x3FCB2B10: 0x00000000 0x00000000 0x00000000 0x00000000 0x3FC00000 0x00000001 0x00000000 0x00000000 
0x3FCB2B30: 0x00000000 0x00000000 0x00000000 0x00000000 0x00000000 0x00000000 0x00000000 0x00000000 
0x3FCB2B50: 0x00000000 0x00000000 0x00000001 0x00000000 0x00000004 0x00060000 0xC0380000 0x3FCAA000 
0x3FCB2B70: 0x3FCDC710 0x00000004 0x3FCB2D34 0x4038C562 0x3FCBA280 0x00000008 0x3FCC7590 0xBAAD5678 
0x3FCB2B90: 0x00000000 0x00000038 0x3FCDCE18 0x4038C72C 0x00000004 0x00000001 0x3FCAA07C 0x000000BA 
0x3FCB2BB0: 0x00000000 0x00000000 0x00000000 0x00000000 0x00000000 0x00000001 0x00190000 0x00000000 
0x3FCB2BD0: 0x00000000 0xFFE70000 0x00000101 0x00000000 0x3FCBBFB8 0x00000008 0x3FCBBFB8 0x00000100 
0x3FCB2BF0: 0x3FCDCE20 0x00000000 0x3FC90000 0xBFC00000 0xFF7F7F01 0x00000000 0xCDC65600 0x00000000 
0x3FCB2C10: 0x00000000 0x3FC00000 0x00000001 0x00000000 0x00000000 0x00000000 0x00000000 0x00000000 
0x3FCB2C30: 0x00000000 0x00000000 0x3FCC7500 0x00000000 0x00000000 0x00000000 0x42058B00 0x00000000 
0x3FCB2C50: 0x00000000 0x81F50000 0x3FCBDAAC 0x00000001 0x00190000 0x00000000 0x00000000 0xFFE70000 
0x3FCB2C70: 0x00000101 0x3FCBE050 0x3FCC638C 0x00000000 0x00000002 0x3FCAA000 0x00000001 0x3FCC7550 
0x3FCB2C90: 0x3FCAA000 0x00000000 0x3FCBDA68 0x42058B32 0x3FCB3130 0x02000000 0x4038DA80 0x3FCBDAAC 
0x3FCB2CB0: 0x00000000 0x3FCDD6D3 0x3FCBE2BC 0x3FCB2CE0 0x3FCDD6D3 0x00000001 0x00190000 0x00000000 
0x3FCB2CD0: 0x00000000 0xFFE70000 0x00000101 0x00000000 0x00000000 0x3FC00000 0x00000001 0x00000000 
0x3FCB2CF0: 0x00000000 0x00000000 0x00000000 0x00000000 0x00000000 0x00000000 0x3FCC7500 0x00000000 
0x3FCB2D10: 0x00000000 0x00000000 0x42058B00 0x00000000 0x00000000 0x81F50000 0x3FCBDAAC 0x3FCB2D34 
0x3FCB2D30: 0x00000004 0x65746F6E 0x3FCBDF00 0x42033CEA 0x00000000 0x3FCB2D4C 0x00000000 0x3FCC6300 
0x3FCB2D50: 0x00000000 0x00000001 0x00000001 0x3FCB2D64 0x00000000 0x3C217900 0x3FCBE0DC 0x42033CEA
```

itsthisjustin:
This PR ports the memory work measured and field-tested on the BLE branch (all fixes, no BLE code and no diagnostic instrumentation), adds async display refreshing, bumps SPI speed to 20 MHz, adds idle-time glyph prewarming, and cuts first-open time for chapter-per-stylesheet books.

## Steady-state heap: ~10 KB more free at rest (device-measured: 107K to 117K)

- **Framebuffer sized to the runtime panel** (SDK a01bb76): the buffer was allocated at the X3 maximum (52,272 B) even on X4 (48,000 B). Allocation drops 53,248 to 49,152, saving 4.1 KB.

- **Timer task stacks right-sized** via custom_sdkconfig, from measured high-water marks: esp_timer used ~0.8 KB of 8 KB, the FreeRTOS timer service ~0.5 KB of 4 KB. Trimmed to 4096/2560, returning ~7 KB.

- **WiFi hot paths moved out of IRAM** (`CONFIG_ESP_WIFI_IRAM_OPT=n`, `CONFIG_ESP_WIFI_RX_IRAM_OPT=n`): on the C3, IRAM and DRAM share one SRAM pool, **so this returns ~25-30 KB to the heap permanently** (the cost is static, so the gain applies even with WiFi off).

  _Trade-off: lower WiFi throughput during transfers, acceptable for occasional sync/OTA use._

Largest allocatable block at rest improved from ~82-86 KB to 94-98 KB with the same trims.

> Note: custom_sdkconfig makes the first build (locally and once per CI cache) rebuild the Arduino core libs. The interrupted-rebuild failure mode and its cleanup are documented in platformio.ini.

## Fragmentation: keep-if-fits font buffers

The per-page mini font arenas (intervals, glyphs, bitmap, kern matrix) were freed and reallocated at slightly different sizes on every page. Each freed hole rarely fit the next page's need, so maxAlloc eroded across a session. Buffers now track capacity and only reallocate when a page needs more; after a few pages the capacities converge on the book's maximum and page turns stop touching the allocator entirely.

## TLS: KOSync sync now works at reading-session heap levels

Root cause of the KOSync failures: the sync server is TLS 1.3 only and requests a P-256 key share via HelloRetryRequest. Key generation ran on fast-math bignums, which WOLFSSL_SMALL_STACK heap-allocates at FP_MAX_BITS size, dozens of multi-KB temporaries at once. At the ~50 KB free a reading session leaves, keygen failed with MP_MEM and the handshake died before the hello was ever sent.

- **Single-precision ECC enabled** (`WOLFSSL_HAVE_SP_ECC` + `WOLFSSL_SP_SMALL`): P-256 operations (keygen, ECDHE, ECDSA cert verify) use fixed 256-bit arrays. A full TLS 1.3 handshake was measured succeeding with 56 KB free and a 43 KB largest block, where it previously failed outright.

- **FP_MAX_BITS 16384 to 8192**: halves every remaining fast-math temporary. 8192 still verifies RSA-4096 roots.

- **Removed inert WOLFSSL_SP_RISCV32**: Arduino-wolfSSL 5.7.2 ships no RISC-V SP file, so the flag did nothing.

- **KOSync heap gate split**: free-heap floor 50 KB, largest-block floor 20 KB (was 50 KB contiguous, which refused syncs that would have succeeded; the largest single TLS allocation is now the ~17 KB record buffer).

- **SDK side (ac5db6f)**: TLS 1.3 key share pinned to X25519 (fixed 32-byte field math, no bignum temporaries), and isWantIo no longer matches the WOLFSSL_CBIO_ERR_* callback codes, which collide with wolfSSL's own error codes (MP_MEM is -2) and made a failed handshake spin for the full 15 s deadline instead of failing fast.

## Crash fixes

- **Background build heap gate**: the incremental section build's parse path grows word vectors of heap strings through throwing operators under -fno-exceptions (field crash: bad_alloc in ParsedText::addWord during a background tick). Background ticks now require 32 KB free and a 16 KB largest block. This is deferrable work, so it simply resumes when page-turn transients free up. The render path still builds the page it actually needs regardless.

- **Web server watchdog registration**: request handlers call esp_task_wdt_reset() but the handler task was never added to the watchdog, so the resets were no-ops. The task is now registered in begin() and removed in stop().

## First-open time

Two fixes target the cold first open of a book (no `.crosspoint` cache yet). Both device-measured on the X4 with the same test book, The Divine Comedy (Gutenberg wrap: 242 spine items, 140 CSS files, illustration JPEGs throughout), from "Loading ePub" to the first page on glass:

| Phase | Before | After |
|---|---|---|
| Metadata indexing (spine/TOC) | 2.8 s | 2.8 s (unchanged) |
| CSS parsing (140 files) | 24.4 s | 0.6 s (3 parsed, 137 skipped) |
| JPEG dimension probe, per image | 3.8 to 4.5 s | ~25 ms |
| First section build (35 pages, 5 images) | 18.8 s | 1.6 s |
| **Open to first rendered page** | **~41 s** | **~8.7 s** |

### Duplicate stylesheet skip

Some converters emit one byte-identical stylesheet per chapter. The test book (The Divine Comedy, Gutenberg wrap) carries 140 CSS files of which 137 are identical; each one cost a zip central-directory locate plus an extract-to-SD round trip, ~155 ms apiece, 24.4 s of the ~41 s first open on device.

CSS parsing now makes a single central-directory pass that maps every stylesheet to its (CRC32, compressed size) already stored in the zip, and parses only the first of each identical pair. Rules merge into one global set, so dropping exact duplicates cannot lose styles; a stylesheet that never matches a directory entry always parses, as before. For the test book this collapses 140 parses to 3, dropping us from 24.4 s to 0.6 s on device.

### Image dimension probe without full extraction

Layout only needs an image's width and height, but the probe previously inflated the entire image out of the EPUB to SD before the header could be read: 3 to 4.5 s per image on device, and paid again when a partial section build resumed. Dimension probes now feed the zip entry's inflate stream through a small header parser and abort the read as soon as the dimensions are known. If no frame header is found in the stream (rare), the old path runs unchanged as a fallback: full extraction, file-based probe, and the indexing popup shown up front since that can take seconds.

`ImageBlock` now serializes the book-internal source href alongside its SD cache path (`section.bin` v31 to v32, so existing section caches regenerate once on update), and the reader registers an extractor hook so the image file is pulled out of the EPUB the first time a page that shows it is actually rendered.

### Image page render: pixel cache held in RAM across passes

The buffer is chunked (16 KB pieces, heap-gated per chunk) because a single 60-96 KB block rarely fits the fragmented mid-render heap; if it doesn't fit, the old streaming path runs unchanged. The reader releases the slot when the page render completes, so nothing stays resident across page turns. Device-measured on a full-page illustration: Gray plane renders dropped from 695/693 ms to 145/165 ms. 

## Page-turn latency: idle next-page glyph prewarm

After a page settles (400 ms debounce), the likely next page is scanned in font-cache scan mode (no pixels drawn) and its glyphs are loaded from SD during idle time. The next turn's in-render prewarm then hits cache instead of paying ~100 ms of SD reads on the page-turn critical path. Prewarm is skipped while a render or build owns the CPU or free heap is below 24 KB.

## Async display refresh: overlap CPU work with the panel's refresh

An e-ink refresh is two phases: the MCU streams the framebuffer to the controller over SPI, then the panel physically drives its pixels through the waveform, which takes hundreds of milliseconds during which the controller needs nothing from the CPU. The blocking displayBuffer() sits in a busy-wait for that whole window, so on every page turn the CPU idles exactly when it has its next chunk of work queued up.

This PR adds a non-blocking path that splits the two phases:

- **displayBufferAsync(mode)** streams the framebuffer and starts the waveform, then returns immediately instead of polling BUSY.
- **waitRefreshComplete()** blocks until the panel finishes (no-op if nothing is pending). Any later blocking display call also drains the pending refresh first, so ordering is preserved.
- **supportsAsyncRefresh()** reports whether the panel driver genuinely defers; callers can skip overlap scaffolding when it falls back to blocking (fadingFix forces the blocking path, and panels without deferral support fall back transparently).

The EPUB reader uses it on the page-turn path: the black-and-white page starts its refresh, and while the panel is busy the CPU renders the grayscale anti-aliasing planes for the same page, work that previously ran only after the refresh wait returned. The panel's ~500 ms refresh window absorbs that render time instead of adding to it.

The contract is documented on the API: the framebuffer must stay untouched until waitRefreshComplete() (single-buffer mode has no second buffer to hide behind), and the caller must rebuild the differential baseline before the next differential update; the tiled grayscale cleanup pass does this. displayWithRefreshCycle() takes an async flag so the reader's refresh-cycle bookkeeping (fast vs. half refresh cadence) is shared between both paths, and the blocking readers (Txt/Xtc) are unchanged.


src/activities/reader/EpubReaderActivity.h
```diff
@@ -86,6 +93,27 @@ class EpubReaderActivity final : public Activity {
...
+  // tick aborted anyway. Free heap says how much memory exists; maxAlloc says whether
+  // any single allocation can actually have it. 16 KB also keeps the advance-table
+  // batch path (16 KB scratch) viable during builds.
+  static constexpr size_t BACKGROUND_BUILD_MIN_MAX_ALLOC = 16 * 1024;
+  // Gate for a background build tick: true when the heap can take parse allocations.
+  bool buildTickHeapGate();
+  // Heap floor for optional render-adjacent work (idle prewarm). Page
+  // deserialization (TextBlock word vectors/strings) and glyph caching allocate
+  // through throwing paths that abort() on OOM; skip deferrable work below it.
+  static constexpr size_t RENDER_MIN_FREE_HEAP = 24 * 1024;
```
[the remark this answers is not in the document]
itsthisjustin: The guard now reads:

ESP.getFreeHeap() > RENDER_MIN_FREE_HEAP && ESP.getMaxAllocHeap() > BACKGROUND_BUILD_MIN_MAX_ALLOC &&

So this is already done

Uri-Tauber:
> The per-page mini font arenas (intervals, glyphs, bitmap, kern matrix) were freed and reallocated at slightly different sizes on every page. Each freed hole rarely fit the next page's need, so maxAlloc eroded across a session. Buffers now track capacity and only reallocate when a page needs more; after a few pages the capacities converge on the book's maximum and page turns stop touching the allocator entirely.

@itsthisjustin I played with this idea before. And it had one big drawback. If you have a book in which one page need **much** more RAM space than the others (for example, a page with normal, bold, and italic text cramped together.) with `develop`, this page take higher allocation, but it's freed next page, allowing other things to use this memory. 
With this "optimization" this space is going to stay half-empty for the rest of the book, but no-one can use it... 

I had thoughts to track the avg of page allocations so far, and use it as factor when deciding if to free the arena or not. The arena can grow only to some percentage above the avg, and get freed above it.
Unfortunately, I didnt implemented this yet - you are welcome to play with this idea, but I would prefer it to be a seperate PR, so measurements can be accurate.

lib/Epub/Epub/parsers/ChapterHtmlSlimParser.cpp
```diff
@@ -722,7 +742,8 @@ void XMLCALL ChapterHtmlSlimParser::startElement(void* userData, const XML_Char*
                 self->currentPageNextY += imageMarginTop;
 
                 // Create ImageBlock and add to page
-                auto imageBlock = std::make_shared<ImageBlock>(cachedImagePath, displayWidth, displayHeight);
+                auto imageBlock =
+                    std::make_shared<ImageBlock>(cachedImagePath, resolvedPath, displayWidth, displayHeight);
```
Uri-Tauber: Using nothrow make sense here.
```suggestion
                    auto imageBlock = std::shared_ptr<ImageBlock>(
                    new (std::nothrow) ImageBlock(cachedImagePath, resolvedPath, displayWidth, displayHeight));

File: lib/Epub/Epub/blocks/ImageBlock.cpp [hunk 3]
```diff
@@ -79,8 +90,113 @@ void rememberImageFailure(const std::string& path) {
...
+  }
+  const size_t firstPart = PXC_CHUNK_SIZE - offset;
+  memcpy(tempRow, pxcChunks[chunk].get() + offset, firstPart);
+  memcpy(tempRow + firstPart, pxcChunks[chunk + 1].get(), bytesPerRow - firstPart);
+  return tempRow;
+}
+
+// cacheFile is positioned just past the header. True when the slot holds the
+// full pixel payload for this cache path afterward.
+bool loadPxcSlot(uint64_t cacheHash, HalFile& cacheFile, uint16_t cachedWidth, uint16_t cachedHeight, int bytesPerRow) {
```
Uri-Tauber, line 140: `pxcChunks`/`pxcSlotHash` is a **single** global slot; `loadPxcSlot()` unconditionally evicts whatever's cached before loading the new image. On a page with 2+ images, each render pass evicts the other image's slot. My thinking is either implement multi-slot cache, or falling back to streaming when a page has >1 image.

lib/Epub/Epub/converters/ImageDimsProbe.h
```diff
@@ -0,0 +1,49 @@
...
+  };
+  State state = State::Sniff;
+  uint32_t pos = 0;       // absolute stream offset (PNG fixed-offset parsing)
+  uint32_t skipLeft = 0;  // remaining segment bytes to skip
+  uint16_t segLen = 0;
+  bool sofPending = false;  // current segment is a SOF frame header
+  uint8_t sofBuf[5] = {0};
+  uint8_t sofFill = 0;
+  uint16_t width = 0;
+  uint16_t height = 0;
```
Uri-Tauber: Those should be uint32_t - PNG width/height are 4-bytes fields. Also fix the relevant `static_cast<uint16_t>` in the .cpp.

platformio.ini
```diff
@@ -40,7 +43,14 @@ build_flags =
...
   -DWOLFSSL_TLS13
-  -DWOLFSSL_SP_RISCV32
+# MEMFIX-PORT: single-precision ECC (sp_c32.c). Without it every P-256 operation
+# (TLS 1.3 key_share keygen, ECDHE, ECDSA cert verify) runs on fast-math bignums
+# that WOLFSSL_SMALL_STACK heap-allocates at FP_MAX_BITS size -- tens of KB of
+# temporaries, which OOMs (MP_MEM) at the ~50KB free heap a reading session
+# leaves. SP uses fixed 256-bit arrays: a few KB, and several times faster.
+# SP_SMALL trades the large precomputed point tables for smaller flash.
+  -DWOLFSSL_HAVE_SP_ECC
+  -DWOLFSSL_SP_SMALL
```
Uri-Tauber, line 53: Claude:
This change trades away a RISC-V asm speedup with no replacement

`sp_int.c:4374` guards this block with `#if defined(WOLFSSL_SP_RISCV32) && SP_WORD_SIZE == 32`, and inside it's raw inline GCC assembly using real RISC-V instructions:

```c
#define SP_ASM_MUL(vl, vh, va, vb)
    __asm__ __volatile__ (
        "mul    %[l], %[a], %[b]   \n\t"
        "mulhu  %[h], %[a], %[b]   \n\t"
        ...
```

`mul`/`mulhu` are RV32M-extension instructions (multiply, multiply-high-unsigned) for 32×32→64-bit widening multiplication — the core primitive bignum math is built on. This code is compiled by the RISC-V cross-toolchain PlatformIO invokes for this target (`riscv32-esp-elf-gcc`) and runs on the ESP32-C3's own RISC-V core at runtime.

So concretely: with `WOLFSSL_SP_RISCV32` defined, wolfSSL's generic bignum path (`sp_int.c`) uses these hand-written `mul`/`mulhu` sequences for the ECC/RSA math. With it removed and `WOLFSSL_HAVE_SP_ECC`/`WOLFSSL_SP_SMALL` taking over (routing to `sp_c32.c`, confirmed zero RISCV32 references), that ECC math instead goes through sp_c32.c's portable C implementation, which the compiler auto-vectorizes/optimizes generically rather than using these explicit RV32 opcodes. The chip is still RISC-V either way — the difference is hand-tuned asm vs. compiler-generated C for the same P-256 operations.

lib/Epub/Epub.cpp
```diff
@@ -256,8 +256,44 @@ void Epub::parseCssFiles() const {
     return;
   }
 
+  // Some converters emit one byte-identical stylesheet per chapter (100+ .css
+  // entries), and each parse costs a zip locate plus an SD extract round-trip.
+  // Map every CSS path to its central-directory (CRC32, compressed size) in a
+  // single scan and parse only the first of each identical pair. Rules merge
+  // into one global set, so dropping exact duplicates cannot lose styles. A
+  // path that never matches a directory entry keeps key 0 and always parses.
+  std::vector<uint64_t> dedupKeys(cssFiles.size(), 0);
```
Uri-Tauber, line 265: A nice optimization here: compare CSS file sizes from the already-available `fileStatSlimCache` before paying for the full CRC scan.

itsthisjustin:
> > The per-page mini font arenas (intervals, glyphs, bitmap, kern matrix) were freed and reallocated at slightly different sizes on every page. Each freed hole rarely fit the next page's need, so maxAlloc eroded across a session. Buffers now track capacity and only reallocate when a page needs more; after a few pages the capacities converge on the book's maximum and page turns stop touching the allocator entirely.
> 
> @itsthisjustin I played with this idea before. And it had one big drawback. If you have a book in which one page need **much** more RAM space than the others (for example, a page with normal, bold, and italic text cramped together.) with `develop`, this page take higher allocation, but it's freed next page, allowing other things to use this memory. With this "optimization" this space is going to stay half-empty for the rest of the book, but no-one can use it...
> 
> I had thoughts to track the avg of page allocations so far, and use it as factor when deciding if to free the arena or not. The arena can grow only to some percentage above the avg, and get freed above it. Unfortunately, I didnt implemented this yet - you are welcome to play with this idea, but I would prefer it to be a seperate PR, so measurements can be accurate.

Yeah that CAN happen but: the waste here is bounded (outlier page minus typical, per style, in one stable block, and it still frees on style eviction / font unload), while the free+realloc pattern it replaces eroded maxAlloc without bound over a session because each freed hole rarely fit the next page's need.

I like the avg idea. One caveat: a book alternating styled and plain pages would thrash a naive average, which is exactly the churn this change killed.

platformio.ini
itsthisjustin: with the flag defined, P-256 keygen ran ~570ms on fast math and OOM'd at reading-session heap levels; with SP ECC (sp_c32) it completes and the handshake succeeds.

itsthisjustin: That block never compiled here: sp_int.c is gated top-to-bottom on WOLFSSL_SP_MATH / WOLFSSL_SP_MATH_ALL (line 35), and we define neither

lib/Epub/Epub.cpp
itsthisjustin: There's no populated fileStatSlimCache at this point (all the Epub ZipFile uses are tmp files), and creating that cache takes the full scan. The CRC here isn't computed, it's read in the same single pass that would give us sizes, and that pass only runs once per book before the CSS cache exists. So a size pre-filter would cost the same timing I am bypassing here

lib/Epub/Epub/converters/ImageDimsProbe.h
itsthisjustin: Fixed

lib/Epub/Epub/blocks/ImageBlock.cpp
itsthisjustin: Fixed

lib/Epub/Epub/parsers/ChapterHtmlSlimParser.cpp
itsthisjustin: Fixed

lib/Epub/Epub.cpp
Uri-Tauber: Hmm. Okay. I stand corrected.

platformio.ini
Uri-Tauber: ok

Uri-Tauber: > That block never compiled here: sp_int.c is gated top-to-bottom on WOLFSSL_SP_MATH / WOLFSSL_SP_MATH_ALL (line 35), and we define neither 

Are you sure of that?
sp_int.c **does** get compiled. it can be found in `/.pio/build/default/libdea/Arduino-wolfSSL/wolfcrypt/src/sp_int.c.o`.

src/activities/reader/EpubReaderActivity.cpp
```diff
@@ -157,6 +157,11 @@ void EpubReaderActivity::onEnter() {
   }
 
   ImageBlock::clearSessionRenderFailures();
+  // Lazy image extraction: section builds only header-probe images, so the first
+  // render of an image page pulls the file out of the EPUB through this hook.
+  ImageBlock::setExtractor(epub.get(), [](void* ctx, const char* src, const char* dest) {
```
Uri-Tauber, line 162: This is a new pointer into epub that outlives a single `onEnter()`/`onExit()` cycle in more places than those two.
onExit() (line 219) correctly clears it before the activity's epub goes away — but `launchKOReaderSync()` (around line 925-926) also frees this same epub early, before `onExit()` runs. 

Worth adding ImageBlock::setExtractor(nullptr, nullptr); there too, right before its `epub.reset()`, mirroring the pattern at line 219.

lib/EpdFont/SdCardFont.cpp
```diff
@@ -311,13 +376,13 @@ bool SdCardFont::buildMiniKernMatrix(PerStyle& s, const uint32_t* codepoints, ui
...
+  // per-page sizes vary by a few entries, which as free+realloc churn was punching
+  // non-coalescing holes in the heap every page turn).
   const uint32_t matrixBytes = static_cast<uint32_t>(numLeft) * numRight;
-  s.miniKernLeftClasses = new (std::nothrow) EpdKernClassEntry[miniLeftCount];
-  s.miniKernRightClasses = new (std::nothrow) EpdKernClassEntry[miniRightCount];
-  s.miniKernMatrix = new (std::nothrow) int8_t[matrixBytes];
-  if (!s.miniKernLeftClasses || !s.miniKernRightClasses || !s.miniKernMatrix) {
+  if (!ensureArrayCapacity(s.miniKernLeftClasses, s.miniKernLeftCapacity, miniLeftCount) ||
+      !ensureArrayCapacity(s.miniKernRightClasses, s.miniKernRightCapacity, miniRightCount) ||
+      !ensureArrayCapacity(s.miniKernMatrix, s.miniKernMatrixCapacity, matrixBytes)) {
```
Uri-Tauber, line 395: This "reused across pages when they fit" comment doesn't currently hold: `buildMiniKernMatrix()` unconditionally calls `freeStyleMiniKern(s)` at its own top (line 331), which resets all three kern buffer capacities to 0 before these `ensureArrayCapacity()` calls run — so they always reallocate.

This `freeStyleMiniKern()` call may now be redundant: `resetStyleMiniData()` (added in 2ec11eb29) already clears the entry/class counts between pages without freeing the buffers, and `buildMiniKernMatrix()` appears to be called only once per style per page, from `prewarmStyle()`, after that reset has already run. If that ordering is guaranteed on every call path, the `freeStyleMiniKern()` call at line 331 could likely be dropped, letting these buffers actually reuse capacity as intended. I haven't verified every path into `buildMiniKernMatrix()`, so worth checking there's no route that reaches it without a prior `resetStyleMiniData()` — if there is, dropping that call could leave stale mini kern data from a previous page's build.

ceryen:
I am new to this code, so apologies if this is a dumb question =P

So when we are idle on a page, we try to prewarm the font cache with glyphs for the next page.  When we then call EpubReaderActivity::renderContents() for the next page, I believe that fcm->createPrewarmScope() creates a PrewarmScope.  The constructor for PrewarmScope seems to call manager->clearCache().  Would that throw out the prewarming done by the prior page?

EDIT: Hm, I'm thinking that "the next turn's in-render prewarm then hits cache instead of paying ~100 ms of SD reads on the page-turn critical path" meant hitting the SD card cache rather than hitting the font cache.  If so, that ought to work, please ignore my question =P

itsthisjustin:
> I am new to this code, so apologies if this is a dumb question =P
> 
> So when we are idle on a page, we try to prewarm the font cache with glyphs for the next page. When we then call EpubReaderActivity::renderContents() for the next page, I believe that fcm->createPrewarmScope() creates a PrewarmScope. The constructor for PrewarmScope seems to call manager->clearCache(). Would that throw out the prewarming done by the prior page?
> 
> EDIT: Hm, I'm thinking that "the next turn's in-render prewarm then hits cache instead of paying ~100 ms of SD reads on the page-turn critical path" meant hitting the SD card cache rather than hitting the font cache. If so, that ought to work, please ignore my question =P

Thats correct but this is still a valid call out. I'm looking at improving it now

Uri-Tauber (approved):
LGTM

PR #2614, opened 2026-07-15
fix: improve ZIP EOCD scanning for large/fragmented EPUBs

yemM:
## Summary

* **What is the goal of this PR?** Make the ZIP End-Of-Central-Directory (EOCD) lookup robust for large/fragmented EPUBs. The previous implementation only scanned the last 1KB of the file, so EPUBs with a larger archive comment (or fragmented trailing data) failed to open.
* **What changes are included?** `ZipFile::loadZipDetails()` now scans the full spec-allowed EOCD region (up to 65557 bytes from EOF) using a fixed 4KB window that starts at end-of-file and steps backwards only when no signature is found:
  * The window nearest EOF is checked first and searched high-to-low, so the EOCD nearest end-of-file always wins — a false `PK\x05\x06` byte sequence earlier in the file (compressed data, nested ZIP) can never be selected over the real record.
  * Common case (no archive comment, EOCD in the last 4KB) costs exactly one 4KB read.
  * Windows overlap by 21 bytes so a 22-byte EOCD record spanning a window boundary is never missed.
  * Each window is a fresh seek+read with a fill loop; a short read fails loudly (`LOG_ERR` + `return false`) instead of scanning stale buffer bytes.
  * Multi-byte fields are extracted with `memcpy` (RISC-V unaligned-load safety).
  * Scan buffer is allocated with `makeUniqueNoThrow` (no manual `free` on the five return paths).

## Additional Context

* **Memory**: one transient 4KB heap allocation during `loadZipDetails()`, freed on return; no persistent RAM cost.
* **I/O**: worst case (no EOCD found / max comment) is ~17 windows of 4KB; typical case is a single 4KB read. Note the pre-change code read 1KB in the typical case, so this is a small increase there, traded for actually finding the record when a comment exists.
* **Static checks**: `pio run` clean, `pio check --fail-on-defect low|medium|high` passes, clang-format applied.

### Verification (on device, Xteink X4)

Tested with purpose-built EPUBs that isolate each behaviour. Serial log evidence:

| Test EPUB | Property | Result |
|---|---|---|
| 60KB archive comment | real EOCD sits **60,049 bytes from EOF** | `[ZIP] EOCD found at offset 4219` — correct. The pre-change 1KB scan could not reach this record at all, so this book previously failed to open. Exercises the multi-window backwards walk. |
| Decoy signature | stored entry containing `PK\x05\x06` at offset **3567**, real EOCD at **4459** | `[ZIP] EOCD found at offset 4459` — correct. The decoy (lower offset, same 4KB window) is passed over in favour of the record nearest EOF. |
| Normal EPUBs (1.2MB, 645KB) | EOCD in final 22 bytes | Found and opened normally; chapters build and render. No regression. |

Both books indexed, built `book.bin`, and rendered. No `ZIP` errors, no watchdog, heap stable.

**Not covered by this testing**, flagged for reviewers: the short-read / genuinely fragmented-SD path is hard to force deliberately, so the fill-loop error branch is reasoned about rather than observed. Happy to add a test if someone has a reliable way to reproduce it.

The generator for these test EPUBs is a short Python script; I can attach it or open it as a follow-up if that's useful to the project.

---

### AI Usage

(Implemented and adversarially reviewed with Claude; findings from the review pass were fixed before opening this PR. Hardware testing and the test-EPUB design were done to validate the result rather than trusting the review.)

### DIFF
--- a/lib/ZipFile/ZipFile.cpp
+++ b/lib/ZipFile/ZipFile.cpp
@@ -3,8 +3,10 @@
 #include <HalStorage.h>
 #include <InflateReader.h>
 #include <Logging.h>
+#include <Memory.h>
 
 #include <algorithm>
+#include <cstring>
 
 struct ZipInflateCtx {
   InflateReader reader;  // Must be first — callback casts uzlib_uncomp* to ZipInflateCtx*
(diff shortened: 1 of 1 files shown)


barbarhan (changes requested):
The signature alone isn't enough to identify the EOCD here. A valid ZIP comment may contain arbitrary bytes, including `PK\x05\x06`; because this loop scans from EOF backward, such a sequence in the comment is selected before the real EOCD and the bytes at `i + 10` / `i + 16` become garbage metadata.

I reproduced this with a structurally valid ZIP: its EOCD comment length and central-directory boundaries were consistent, but this algorithm selected an embedded `PK\x05\x06` marker in the comment instead of the real EOCD.

Please validate each candidate before returning—at minimum, read the 16-bit comment length at `i + 20` and require `candidateOffset + 22 + commentLength == fileSize`, plus a sensible central-directory bound such as `centralDirOffset <= candidateOffset`. If validation fails, continue scanning backward. A regression test with the signature embedded in the ZIP comment would lock this down.

Uri-Tauber:
@yemM Can you share the test epubs?

hooligan333:
Hi @yemM — following the maintainers' call for help with the PR backlog, I picked this one up: it fixes a real failure mode (books whose EOCD sits beyond the last 1KB never open), and I'd like to help it land. Everything below is offered for you to take into this PR as you see fit.

## Continuation addressing the review

Branch: [`hooligan333:fix/zip-eocd-scan-2614`](https://github.com/hooligan333/crosspoint-reader/tree/fix/zip-eocd-scan-2614) — your commit cherry-picked unchanged (authorship preserved), plus [one commit](https://github.com/hooligan333/crosspoint-reader/commit/0bc8dc24) implementing @barbarhan's requested validation:

- Every `PK\x05\x06` match is validated before acceptance:
  - **Self-consistency** (new host-testable `lib/ZipFile/ZipEocd.h`): `recordOffset + 22 + commentLength == fileSize` and `centralDirOffset <= recordOffset` — the checks requested in review.
  - **Central-directory probe**: the 4 bytes at `centralDirOffset` must be `PK\x01\x02` (or the EOCD's own signature for an empty archive). One extra 4-byte read, and only for candidates that match the signature.
- A record passing both wins immediately. If the scan exhausts without one, the nearest-EOF candidate that passed the probe alone is used as a fallback — so archives with trailing data after the EOCD (which the old fixed-1KB scan happened to tolerate) keep opening.
- `test/zip_eocd/`: 7 host gtests, including the decoy-in-the-comment regression case requested in review, plus a nastier variant (a byte-exact copy of the real EOCD embedded in the comment).

## Verification

@Uri-Tauber asked for test epubs — I generated six synthetic but fully valid, readable EPUBs whose ZIP tails each exercise one scanner path (book titles on-device name the case). Happy to attach them plus the generator script here if useful. Verified behavior:

| book | develop (1KB scan) | this PR as submitted | PR + validation |
|---|---|---|---|
| 01 control | opens | opens | opens |
| 02 decoy signature in comment | **garbage EOCD** | **garbage EOCD** | opens |
| 03 60KB comment | fails to open | opens | opens |
| 04 EOCD copy in comment | **garbage EOCD** | **garbage EOCD** | opens |
| 05 trailing data after archive | opens | opens | opens (fallback tier) |
| 06 EOCD straddles the 4KB window edge | fails to open | opens | opens |

Two findings worth noting:

- The decoy failure mode **predates this PR**: develop's fixed-1KB scan also accepts the first signature from EOF without validation, so a decoy within the last 1KB breaks current releases the same way. This PR + validation fixes an existing bug, not just a new-code risk.
- Python's own `zipfile` mis-parses books 02 and 04 (same naive backward scan), so this bug class has respectable company.

On-device A/B on an Xteink X4 Pro: with develop head (`eef20504`), books 01/05 open and 02/03/04/06 fail; with the branch, **all six open**, and existing library books regression-check fine. `pio run` for `x4pro` and `default` (C3) both build; `pio check --fail-on-defect low` reports no defects; clang-format clean.

## How to proceed

@yemM if you're around: feel free to cherry-pick the commit from the branch above into this PR — it's yours to take. If you're busy, I'm equally happy to open a successor PR that keeps your commit and authorship intact and links back here. Whichever you and the maintainers prefer.

*Transparency: the continuation was developed with AI assistance (Claude Code); I reviewed the changes and verified them on real hardware.*

PR #2615, opened 2026-07-15
fix: self-close void HTML elements before expat parsing

yemM:
## Summary

* **What is the goal of this PR?** Fix chapter parse failures on EPUBs whose generators emit HTML5 void elements without self-closing slashes (`<br>`, ``, `<hr>` …). Expat is a strict XML parser and rejects these, making affected chapters unreadable.
  * New `ChapterHtmlSlimParser::selfCloseVoidElements(src, dst, bool& modified)`: a single-pass streaming byte state machine (512-byte chunks, two `makeUniqueNoThrow` buffers, no whole-file buffering) that:
    * inserts the missing `/` before `>` on void start tags (`<br>` → `<br/>`), with quote-state tracking so a literal `>` or `/` inside an attribute value is never misinterpreted (`<img alt="a>b">` handled correctly);
    * drops void end tags (`<br></br>` → `<br/>`), failing open (verbatim passthrough) on anything unexpected;
    * passes comments (`<!-- -->`), CDATA (`<![CDATA[ ]]>`), DOCTYPE and processing instructions through byte-identically via dedicated states, with **no** quote tracking inside them — so an apostrophe in a comment (`<!-- don't -->`) cannot derail the machine and silently suppress fixups for the rest of the file;
    * propagates I/O errors (write counts checked, short read treated as failure) so the caller falls back to the original file instead of parsing a truncated copy;
    * reports whether anything was actually modified.
  * `Section::createSectionFile()` runs the sanitizer on the temp chapter HTML before expat: on modified → parse the sanitized copy; on unmodified or failure → parse the original (sidecar removed in all non-use paths, no orphan files left on SD).

## Additional Context

* **Memory**: two transient 512-byte heap buffers during section build, ~50 bytes of state-machine locals; no persistent RAM cost.
* **I/O**: one extra read+write pass of the temp chapter HTML per section build (section builds are already I/O-heavy and results are cached under `.crosspoint/`). Chapters needing no changes still parse the original file, so the sanitized sidecar is never kept for the common case.
* **Static checks**: `pio run` clean, `pio check --fail-on-defect low|medium|high` passes, clang-format applied.
* No cache-format change: chapters that previously failed were never cached, so no version bump is needed. Happy to bump if maintainers prefer.

### Verification (on device, Xteink X4)

Built a test EPUB with one chapter per behaviour. Four of the six chapters are **rejected by expat before this change** (verified independently against expat: `mismatched tag`), and two are controls that already parse and must keep working. All six render correctly on device:

| Chapter | Content | Pre-change | Result |
|---|---|---|---|
| C1 | bare `<br>`, `<hr>`, `` | expat rejects | renders; image decodes and draws |
| C2 | `<br></br>` and `<BR></BR>` | **parses OK** (valid XML — regression control) | renders; end tags dropped, no mismatched-tag error |
| C3 | `<!-- don't edit -->` followed by `<br>` | expat rejects | renders; the apostrophe does not suppress the following fixup |
| C4 | `>` and `/` inside double- and single-quoted attributes | expat rejects | renders; no leaked `b">` / `d'>` / `x/"` text |
| C5 | already well-formed (`<br/>`, `<hr/>`) | parses OK (control) | renders; takes the `modified=false` path, original parsed |
| C6 | `<brand>`, `<b>`, `base`-prefixed non-void names | expat rejects | renders; lookalikes not self-closed |

Every chapter carries an XHTML DOCTYPE with quoted public/system identifiers and an `<?xml … ?>` declaration, so the declaration and PI passthrough paths are exercised on every single chapter build.

Serial log for C1: `[SCT] Streamed temp HTML to .tmp_0.html (481 bytes)` → `[EHP] Time to parse and build pages: 142 ms` → `[SCT] Page 0 processed`. No `Void-element sanitization failed` and no `Failed to parse XML` anywhere across the run. Existing real-world EPUBs continue to open and render normally (regression check); heap stable, no watchdog.

**Not covered by this testing**, flagged for reviewers:
* CDATA passthrough is implemented and hand-traced (`]]>` matcher) but no test chapter exercised it on device.
* C4 confirms quote handling by the *absence* of leaked attribute text rather than by reading the attribute value back — with default image rendering the firmware draws the image instead of the `alt` text, so the value isn't user-visible.

The generator for the test EPUB is a short Python script; I can attach it or contribute it as a follow-up if that's useful.

---

### AI Usage

(Implemented and adversarially reviewed with Claude. The review caught a real defect in an earlier iteration of this state machine — quote tracking applied inside comments, which silently suppressed fixups after any apostrophe in a comment — which is why C3 exists as a dedicated test case. Hardware testing was done to validate the result rather than trusting the review.)


Uri-Tauber:
I'm not convinced this belongs in Crosspoint. The parser should be able to assume it's receiving valid XHTML. Adding HTML5 compatibility workarounds to `ChapterHtmlSlimParser` increases the complexity of an already difficult-to-maintain component.

If we want to support malformed EPUBs (which I don't think we do), I'd rather see them normalized by the Calibre plugin or the web server's EPUB optimizer.

Uri-Tauber:
Closing as not planned. 
Thanks for your effort @yemM.

PR #2619, opened 2026-07-15
feat: Back on home menu opens the most recent book

rxmmah:
## Summary

The Back button had no function or hint label on the home menu. It now opens the most recently read book (first entry of the pruned recents list) and the button hint shows "Read". Hidden when there are no recent books.

only changes to 'src/activities/home/HomeActivity.cpp' and 'lib/I18n/translations/english.yaml'
---

### AI Usage


Uri-Tauber:
@rxmmah Nice UX improvement.
One thing: In my opinion, "resume" sounds better than "read". What do you think?

rxmmah:
> @rxmmah Nice UX improvement. One thing: In my opinion, "resume" sounds better than "read". What do you think?

i agree with you. it's more natural.

Uri-Tauber (approved):

PR #2634, opened 2026-07-17
fix: prevent SD font advance cache overlap

### LINKED ISSUE #2002: Large SD card CJK fonts can render overlapping text when advance metrics are missing
### Affected Version

v1.3.0

## Large SD card CJK fonts can render overlapping text when advance metrics are missing

When using a large SD card font, such as a broad Unicode/CJK NotoSansCJK `.cpfont`, some EPUB text can render with overlapping characters.

This seems to happen when the SD card font advance cache does not contain metrics for some glyphs. If a missing advance effectively falls back to `0`, the measured text width becomes too small and later characters can be positioned on top of earlier ones.

In local testing, using the largest cached `advanceX` value as the fallback for missing advance-cache entries avoided the overlap. This is probably only a defensive workaround, not necessarily the best final solution.

Possible fixes:

- ensure SD font advance-cache misses never return zero width during text measurement
- use a conservative fallback advance, such as max cached advance, replacement glyph advance, average advance, or a font-level default
- consider storing a fallback advance in the `.cpfont` metadata
- consider improving the advance cache strategy for very large CJK fonts

The max-cached-advance fallback fixes the overlap in my test case, but it may overestimate width and cause earlier line breaks. A better long-term solution may be a font-provided default or CJK-specific average advance.

### Steps to Reproduce

1. install attached font
2. change reader font family to the font
3. open large CJK combined text

[noto_sans_cjk_kr_14_2bit.epdfont.zip](https://github.com/user-attachments/files/27855809/noto_sans_cjk_kr_14_2bit.epdfont.zip)

### Expected Behavior

render text with proper width

### v1.3.0
Image

### local tested version - using the largest cached `advanceX`
Image

uxjulia:
## Summary

  * Prevent overlapping EPUB text when large SD-card fonts exhaust the glyph-advance cache. Closes #2002.
  * Refresh a full advance cache when it does not cover the active text.
  * Read missing glyph advances directly from `.cpfont` metadata instead of treating them as zero-width.
  * Apply the same fallback to space measurement.
  * Bump `section.bin` from version 30 to 31 so affected EPUB layouts rebuild automatically.
  * Document the cache-version change.

## Additional Context

* Ported from uxjulia/CrossInk@78a94233dbd9e82369f796a475a9989d58a27e0f.
* The persistent cache remains bounded at 768 entries per style and no long-lived allocation was added.
* Cache misses may require an additional SD metadata read during layout. Each file is closed immediately; this favors correct spacing over silently returning a zero width.
* Hardware reproduction with the issue’s CJK font and EPUB is still recommended.

---

### AI Usage

### DIFF
--- a/docs/file-formats.md
+++ b/docs/file-formats.md
@@ -90,16 +90,19 @@ if (parsedSize != fileSize) {
 
 ## `section.bin`
 
-### Version 30
+### Version 31
 
 Each file in `sections/*.bin` stores one laid-out spine section. The header is
 also the cache-busting key: if any layout-affecting setting differs from the
 current reader settings, the section is discarded and rebuilt.
 
-Version 30 is binary-identical to version 29. The version was bumped because
-Arabic contextual shaping changed text measurement (`getTextAdvanceX` now
-measures the shaped visual text), so word positions cached by v29 no longer
-match what `drawText` renders.
+Version 31 is binary-identical to version 30. The version was bumped so
+SD-card font layouts are rebuilt with corrected glyph-advance measurement for
+CJK-heavy text.
+
+Version 30 was bumped because Arabic contextual shaping changed text
+measurement (`getTextAdvanceX` measures the shaped visual text), so word
+positions cached by v29 no longer match what `drawText` renders.
 
 Version 28 introduced serialized word style bits for underline, strikethrough,
 superscript, and subscript. The format also includes:
--- a/lib/EpdFont/SdCardFont.cpp
+++ b/lib/EpdFont/SdCardFont.cpp
@@ -655,6 +655,41 @@ int32_t SdCardFont::findGlobalGlyphIndex(const PerStyle& s, uint32_t codepoint) 
   return -1;
 }
 
+bool SdCardFont::readAdvance(uint32_t codepoint, uint8_t style, uint16_t* outAdvance) const {
+  if (!outAdvance || !loaded_) return false;
+
+  const uint8_t styleIdx = resolveStyle(style);
+  if (styleIdx >= MAX_STYLES || !styles_[styleIdx].present) return false;
+  if (advanceTableLookup(styleIdx, codepoint, outAdvance)) return true;
+
+  const auto& s = styles_[styleIdx];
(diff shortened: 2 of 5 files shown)


Uri-Tauber:
@uxjulia Have you tested with the files attached in #2002?

uxjulia:
> @uxjulia Have you tested with the files attached in #2002?

There is only an `.epdfont` attached to that issue. I tried installing it as-is and renamed as `.cpfont` and it doesn't work. I need a korean epub to test with and I don't know of any.

brianhuster:
>  I need a korean epub to test with and I don't know of any

I've found a txt file for the Korean translation of "The Little Prince". It has 906 unique characters.

[어린 왕자 (텍스트txt 최종)(한글 최종) 171220.txt](https://github.com/user-attachments/files/30160582/txt.171220.txt)

uxjulia:
> > I need a korean epub to test with and I don't know of any
> 
> I've found a txt file for the Korean translation of "The Little Prince". It has 906 unique characters.
> 
> [어린 왕자 (텍스트txt 최종)(한글 최종) 171220.txt](https://github.com/user-attachments/files/30160582/txt.171220.txt)

@brianhuster - I had to convert the `.txt` file to an EPUB to even get it to render. I used a generated Merriweather + NotoSans KR + NotoSans JP, and the text renders fine without any overlap on the current `develop` branch before and after line height/margin changes. I can't reproduce the overlapping. Can you verify that this is still an issue in the nightly Crosspoint builds?

brianhuster:
Hm, I cannot reproduce the bug on nightly anymore. Antigravity found why, because the function `GfxRenderer::getTextAdvanceX` already has fallback (to read SDfont) in case `sdIt->second->getAdvance` returns 0

https://github.com/crosspoint-reader/crosspoint-reader/blob/b1d1d757ba7b393f88e5ffca02b1013e64e6393b/lib/GfxRenderer/GfxRenderer.cpp#L1817

uxjulia:
Closing since the issue this was meant to fix seems to have already been mitigated from updates in v1.4.0

PR #2640, opened 2026-07-17
feat: folder move/rename and bulk move in web file manager

rxmmah:
## Summary

The web file manager could move and rename individual files (#630) but rejected folders ("Only files can be renamed/moved"), and the only bulk action was Delete Selected. This PR closes both gaps:

- **Folder move and rename**: `/rename` and `/move` now accept directories. SdFat's `FatFile::rename` supports subdirectories natively (including the dot-dot entry fix-up when reparenting), so this is a single FAT metadata operation — no copying. Folder rows in the file table get the same 📂 move / ✏️ rename buttons files already have.
- **Self-move guard**: `/move` refuses to move a folder into itself or its own subtree (which would orphan the subtree on the FAT chain); the same check runs client-side for a friendlier message.
- **Move Selected**: new toolbar action next to Delete Selected, reusing the existing checkbox selection. The move modal now handles multiple items and posts `/move` sequentially per item (SD access is serialized by the storage mutex, so parallel requests would only add contention); failures are collected and reported together.
- `clearBookCache` is skipped for directories — it only understands book file paths.

### DIFF
--- a/src/network/CrossPointWebServer.cpp
+++ b/src/network/CrossPointWebServer.cpp
@@ -884,14 +884,10 @@ void CrossPointWebServer::handleRename() const {
 
   HalFile file = Storage.open(itemPath.c_str());
   if (!file) {
-    server->send(500, "text/plain", "Failed to open file");
+    server->send(500, "text/plain", "Failed to open item");
     return;
   }
-  if (file.isDirectory()) {
-    file.close();
-    server->send(400, "text/plain", "Only files can be renamed");
-    return;
-  }
+  const bool isDir = file.isDirectory();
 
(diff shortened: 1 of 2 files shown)


uxjulia:
In the current implementation, any renamed or moved folder would cause bookmarks, recent books, and progress to be lost, resulting in a poor user experience. Users will wonder where their bookmarks went and why opening a book puts them back to the start. There should either be a warning to let the user know what will happen from this action or implement shared book-relocation helper that proactively remaps those files to the new location. I'd prefer the remapping to make this more seamless.

fain182:
#3166 is related

PR #2464, opened 2026-06-28
chore: Refactor stores to use PersistableStore CRTP template

Uri-Tauber:
### Summary

`KOReaderCredentialStore`, `WifiCredentialStore`, `RecentBooksStore`, and `OpdsServerStore` all shared the same boilerplate: a static singleton instance, `saveToFile`/`loadFromFile`, and a password-deobfuscation fallback. This PR eliminates that duplication.

**What changed:**

Introduces `PersistableStore<T>` (CRTP base class, lives in `lib/Serialization/`) that provides `getInstance()`, `saveToFile()`, `loadFromFile()`, and `extractPassword()` once for all stores. Each store now only implements three things: `getFilePath()`, `toJson()`, and `fromJson()`.

As a consequence, `KOReaderJsonIO` and the WiFi/RecentBooks/OPDS functions in `JsonSettingsIO` are deleted — their logic moved directly into each store's `toJson`/`fromJson`. **The old binary-migration paths (`.bin → .json`) are also removed, as we can safely asume no devices remain on binary-only storage.**

The legacy OPDS fields (`opdsServerUrl`, `opdsUsername`, `opdsPassword`) are removed from `CrossPointSettings` along with their binary deserialization, and `SETTINGS_FILE_VERSION` is bumped to 2 accordingly.

Flash footprint increased in ~1.5kb, expected when using templates. 

`develop`:
RAM:   [===       ]  31.0% (used 101564 bytes from 327680 bytes)
Flash: [========  ]  79.7% (used 5225945 bytes from 6553600 bytes)

PR: 
RAM:   [===       ]  30.9% (used 101340 bytes from 327680 bytes)
Flash: [========  ]  79.8% (used 5227537 bytes from 6553600 bytes)

---

### AI Usage

While CrossPoint doesn't have restrictions on AI tools in contributing, please be transparent about their usage as it 
helps set the right context for reviewers.

Did you use AI tools to help write this code? _**< PARTIALLY >**_

PR #2647, opened 2026-07-19
chore: Migrate Settings/State onto PersistableStore;

Uri-Tauber:
# Summary

A follow up of #2464.

`CrossPointSettings` and `CrossPointState` were the last two persisted objects still carrying bespoke serialization. Each had three code paths for the same data: a hand-rolled binary reader with positional field ordering, a JSON reader, and a JSON writer — the latter two living in a separate `JsonSettingsIO` module. Every new setting meant touching all three, and the binary reader's `if (++settingsRead >= fileSettingsCount) break;` ladder made field order load-bearing in a way nothing enforced.

## What changed

1. Settings/State onto PersistableStore
Both now implement toJson/fromJson/getFilePath and inherit file I/O, locking, and the singleton. JsonSettingsIO is deleted entirely (−340 lines). Binary migration is dropped rather than ported.

2. `StatusBarSpec` / `ReaderRenderSpec` — resolved value objects
Consumers now ask for a resolved snapshot instead of reading ~10 raw fields and re-deriving. `ReaderRenderSpec` replaces the 10-parameter signatures throughout Section; 

3. BookmarkFile
Bookmark load/save was duplicated between `EpubReaderActivity` and `EpubReaderBookmarksActivity`, each open-coding path derivation, mkdir, and the exists/empty/parse ladder. Now one namespace taking a book

## Flash impact

~1.5kb win.

---

### AI Usage


Uri-Tauber:
This PR also removes the remaining code related to the binary-to-JSON migration.

As a result, users upgrading directly from v1.0.0 will lose all of their settings. We should add a warning to the release notes advising users to first upgrade to v1.4.1, and only then update to the latest version, so their settings are migrated correctly.

Uri-Tauber:

itsthisjustin (approved):

PR #2651, opened 2026-07-19
chore: update Spanish, Catalan, and Valencian translations

lpla:
## Summary

  Add the Spanish, Catalan, and Valencian translations missing from current `develop`, and keep longer localized Wi-Fi connection statuses readable in portrait orientation.

  - Translate the new dictionary labels and status messages.
  - Translate the KOReader sync behavior and account-registration messages.
  - Translate the recent-book resume hint, short-Back navigation setting, already-synced status, and OPDS SD-root label.
  - Translate the four touch-control strings added by [#2481](https://github.com/crosspoint-reader/crosspoint-reader/pull/2481), including the current [English source keys](https://github.com/lpla/crosspoint-reader/blob/048d6d65d7ac47ab516685c955b2b8febf78bbdc/lib/I18n/translations/english.yaml#L234-L249).
  - Bring all three locale files to 426 translated strings with zero English fallbacks.
  - Add a reusable, bounds-aware [`UITheme::drawCenteredWrappedText`](https://github.com/lpla/crosspoint-reader/blob/048d6d65d7ac47ab516685c955b2b8febf78bbdc/src/components/UITheme.cpp#L176-L209) helper for centered text that may need multiple lines.
  - Use the helper for saved-network scanning and connection statuses.
  - Keep the existing single-line `drawCenteredText` API unchanged for current callers.

## Additional Context

* This does not add a locale or a new localization subsystem; it keeps three existing shipped locales aligned with the current English UI.
* The patch applies cleanly to current upstream `develop`.
* The helper checks text width first and only invokes the existing UTF-8-safe wrapper when text overflows.
* The long Catalan saved-network connection status was verified on an Xteink X4 and renders correctly across two centered lines.
* Verified with a parsed YAML key and C-format-placeholder audit: 426/426 own strings and zero fallbacks in ES, CA, and CAV.
* Verified with `python3 scripts/gen_i18n.py lib/I18n/translations lib/I18n/`, `pio run`, `pio check`, and a fresh host build running all 129 tests.
* The measured build kept static RAM unchanged and added 1,882 bytes of flash.

---

### AI Usage


Uri-Tauber:
What are the changes in `src/activities/network/WifiSelectionActivity.cpp`?

lpla:
Those are for wrapping the "connecting to the saved Wi-Fi" line in File Transfer into two lines, because on these languages particularly they are long translations that don't fit the screen (in portrait at least). I tested it and it wraps the line fine, without regressions in English.

Uri-Tauber:
@lpla This looks to me like code duplication - the two new blocks are almost identical.
Also, if you will find tomorrow somewhere else wrapping is needed - would you add the exact same lines there?
The clean solution would be to change `drawCenteredText` so it will warp lines automatically, so individual activities wont have to do so manually.

lpla:
@Uri-Tauber Good point. I moved the wrapping and per-line drawing out of `WifiSelectionActivity` into the reusable `UITheme::drawCenteredWrappedText` helper in 05ef367e.

The helper takes bounded layout space, a line limit, and vertical alignment, so the activity now only defines where the scanning and connecting status blocks belong. I kept the existing single-line `drawCenteredText` contract unchanged because its current callers use fixed `y` positions; the new helper also retains the width-check fast path so short strings do not incur wrapping allocations.

Validated with `pio run`, `pio check`, all 129 host tests, and a successful flash to the X4.

Uri-Tauber (approved):
Thanks @lpla

PR #2652, opened 2026-07-20
fix(epub): preserve word continuation when splitting CJK text on MAX_WORD_SIZE

brianhuster:
…WORD_SIZE

## Summary

* **What is the goal of this PR?** fix random spaces appearing in Chinese and Japanese epub
* **What changes are included?** Set nextWordContinues to true when flushing partWordBuffer due to buffer overflow, preventing random spaces in CJK text; Bump SECTION_FILE_VERSION to 31 to invalidate older cached section files.

## Additional Context

There should be no performance regression

---

### AI Usage

### DIFF
--- a/lib/Epub/Epub/Section.cpp
+++ b/lib/Epub/Epub/Section.cpp
@@ -17,7 +17,7 @@ namespace {
 // v30: Arabic shaping changed both drawing and measurement (getTextAdvanceX now
 //      measures the shaped visual text); cached word positions from v29 no longer
 //      match what drawText renders.
-constexpr uint8_t SECTION_FILE_VERSION = 30;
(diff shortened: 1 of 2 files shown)


Uri-Tauber:
Could you share a screenshot of book with this bug before/after, so I can understand better the fix?

brianhuster:
| Before | After |
| :---: | :---: |
| kanji_ch1_p1_0pct_144899 | kanji_ch1_p1_0pct_87623 |



@Uri-Tauber In the first image (before the fix), the book is not well justified, because some unexpected spaces are inserted in 5th, 9th, 13rd and 17th lines

In the second image (after the fix), the book is perfectly justified

Uri-Tauber (approved):

PR #2278, opened 2026-06-06
feat(epub): basic table support for row/colspan and for solid/none border style

WuTofu:
## Summary

* **What is the goal of this PR?**  
  Implements basic EPUB table rendering support in the reader, including row/colspan handling and border rendering for `solid`/`none` border styles.

* **What changes are included?**  
  * Added `TableBlock` and `PageTable` support for table rendering and page serialization.
  * Extended `ChapterHtmlSlimParser` to parse `<table>`, `<tr>`, `<td>`, and `<th>` with cell padding, width, colspan, rowspan, and border style accumulation.
  * Added CSS parsing for `border-style`, `border-width`, and shorthand `border`/`border-top`/`border-bottom`/`border-left`/`border-right`.
  * Added `CssBorderStyle` support and visibility checks in `CssStyle`.
  * Added table serialization/deserialization and page element tagging in Page.cpp / `Page.h`.

## Additional Context

1. Basic two-column table

| Table  | Screenshot  | 
|--------|--------|
| image | image |

2. Basic three-column table

| Table | Screenshot | 
|--------|--------|
| image | image|

3. Center-aligned table with no border

|  Table  | Screenshot | 
|--------|--------|
| image | image |

4. Table with no border

| Table | Screenshot | 
|--------|--------|
|image | image|

5. Table with border-{top,bottom} style

| Table | Screenshot | 
|--------|--------|
| image | image |

6. Table with border-{top,bottom} style and center-aligned text

| Table | Screenshot | 
|--------|--------|
| image  | image|

7. Table with row/colspan

| Table | Screenshot | 
|--------|--------|
| image | image |

8. Table with border shorthands style
image

| Screenshot 1 | Screenshot 2 | Screenshot 3 |
|--------|--------|--------|
| image | image |  image | 

9. Table with border shorthands style (postive pixels + none)

image

| Screenshot 1 | Screenshot 2 | Screenshot 3 |
|--------|--------|--------|
| image |image | image |

10. Table with border shorthands style (none + postive pixels)

image

| Screenshot 1 | Screenshot 2 | Screenshot 3 |
|--------|--------|--------|
| image |  image | image |

11. Table with special case of rowspan="0"

Note: colspan must be a positive number, only rowspan can be zero or a positive number.

image

| Screenshot 1 | Screenshot 2 | Screenshot 3 |
|--------|--------|--------|
| image | image | image |

---

### AI Usage

While CrossPoint doesn't have restrictions on AI tools in contributing, please be transparent about their usage as it 
[...]

PR #2654, opened 2026-07-20
feat: render EPUB tables as columns

### LINKED ISSUE #876: [Feature Request] table support
I see that tables are intentionally omitted right now. [A Study In Scarlet](https://standardebooks.org/ebooks/arthur-conan-doyle/a-study-in-scarlet/downloads/arthur-conan-doyle_a-study-in-scarlet.epub) is an example of a novel that contains a table (in chapter 2) that is part of the story, as I noticed when I was reading it and came across `[Table omitted]`.

I can see how these would be challenging to format on a small device but it would be nice to find a way to include them.

lpla:
Fixes #876

## Summary

* **What is the goal of this PR?** Replace the current synthetic `Tab Row N, Cell N:` paragraphs with a lightweight column layout for the simple tables commonly found in EPUBs.
  - Renders rows with 2-4 simple cells as equal-width positioned columns using the existing `PageLine` and `PageHorizontalRule` primitives.
  - Keeps `<th>` text bold, respects cell text alignment/direction, mirrors column order for RTL rows, preserves image alt text, and accounts for native ruby height when advancing table rows.
  - Falls back to ordinary full-width flow when a row is too narrow or complex: more than four cells, more than 32 words or 512 text bytes in a cell, an internal link, or other unsupported structure. Content is preserved rather than truncated.
  - Flattens nested-table text into the enclosing cell without creating recursive table state.
  - Bumps `section.bin` to version 38 so existing cached word positions are rebuilt once; version 37 is used by current `develop`.

### Intentionally limited scope

This is a deliberately smaller alternative to #2533 and follows the maintainer direction in #2278 to start with simple table cases without border/CSS machinery. Those are also the limited cases CrossInk supports, as noted in [the same discussion](https://github.com/crosspoint-reader/crosspoint-reader/pull/2278#issuecomment-4640473055), and match the [maintainer request](https://github.com/crosspoint-reader/crosspoint-reader/pull/2278#issuecomment-4640297495) for a reviewable core implementation.

The implementation does **not** add a table block type, new page serialization tags, CSS border parsing, intrinsic column sizing, `colspan`, or `rowspan`. Unsupported rows keep the existing readable stacked behavior, so those can be added incrementally if there is demand.

Grid borders are intentionally kept in the separate follow-up #2954.

### Memory and firmware impact

The parser buffers only the active row. The row vector is pre-reserved for four cells; each grid cell is bounded to 32 words and 512 text bytes before the row switches to streaming full-width flow. `ParsedText` cells are allocated with `makeUniqueNoThrow` because they must survive across Expat callbacks; stack lifetime is insufficient and static storage would make parser instances share mutable state.

Compared with `origin/develop` at `d0359edf` using clean `default` builds and the same submodule revisions:

| Metric | `develop` | This PR | Delta |
|---|---:|---:|---:|
| Static DRAM | 119,021 B | 119,021 B | 0 B |
| PlatformIO RAM | 52,036 B | 52,036 B | 0 B |
| Firmware image | 5,503,841 B | 5,507,667 B | +3,826 B |

### Validation

- Current GitHub CI: firmware build, unit tests, cppcheck, clang-format, title check, and CodeRabbit all pass
- `pio run -e default`: passed
- `pio check -e default`: passed, no defects
- Fresh host CMake/CTest build: 129/129 tests passed
- `clang-format` and `git diff --check`: passed
- Flashed to an Xteink X4 and tested with real three-column tables in *Momo* after a fresh section-cache rebuild:
  - portrait and landscape
  - Small and default Medium reader fonts
  - default Medium portrait keeps the three cells on the same row
  - no reset, OOM, or parser error observed while navigating the table pages

### Known limitations

- Equal-width columns only; no content-proportional sizing.
- Grid mode supports at most four cells per row.
- Narrow or complex rows intentionally render stacked.
- No borders, `colspan`, `rowspan`, or table-specific CSS sizing.

### Screenshots

Before/after screenshots from the Xteink X4, including portrait and landscape results, are in [this PR comment](https://github.com/crosspoint-reader/crosspoint-reader/pull/2654#issuecomment-5021045573).

---

### AI Usage

Implemented and reviewed collaboratively with OpenAI Codex; the contributor performed the manual device and orientation/font validation.


lpla:
Some screenshots.

Before PR (develop branch and v1.4.1, same behaviour):
Momo-(ed.-original)_ch15_p23_22pct_12621.bmp
Momo-(ed.-original)_ch15_p24_22pct_19621.bmp


After (tested both portrait and landscape modes):
Momo-(ed.-original)_ch15_p24_23pct_122809.bmp
Momo-(ed.-original)_ch15_p26_23pct_137822.bmp

thiagokokada:
CC @Uri-Tauber, could you have a look at this?

Uri-Tauber:
@thiagokokada Tables is definitely high on my priorities.
There are three competing PRs for that. this one, #2533 and #2278. 
This one seems the lightest of them all, which I like. The question is how well it handle real world cases. 
Another thing, I think normal grid is better than rows only. So, if the only reason is aesthetics, I would prefer to have normal grids.

If you have the time to test this, and the other two, it would be a great help. I'm busy above my head.

lpla:
Thanks for the feedback, @Uri-Tauber. The horizontal-rule version wasn’t only an aesthetic choice. It was the smallest approach that could reuse the current serialized page elements and follow the narrow scope you suggested on #2278.
I recently tested it against a varied set of public-domain EPUBs, including _A Study in Scarlet_. Simple 2–4-cell rows remain tabular, while wide, spanning, or unusually long rows preserve all text through the stacked fallback. But I understand that vertical boundaries would make long rows clearer.
I think the best next step is a small equal-width grid primitive for supported rows, keeping the same fallback and still excluding CSS borders, intrinsic sizing, and span layout. I can update the PR in that direction if you really feel like it is worth that extra code.

Uri-Tauber:
@lpla Thanks for the clarification. 
I think the grid should be a followup PR.

Uri-Tauber (approved):
LGTM.

Uri-Tauber:
Tested on my device. Work fine for english and hebrew (RTL) tables with 3 columns max.
four columns fall back to list, so the PR description is wrong here.
Other than that, looks awesome!

Sorry it took me so much time to review. I have tons of other stuff on my head right now.

PR #2663, opened 2026-07-21
feat: ability to get definitions from the dictionary screen recursively

ryanbarouki:
## Summary

  - To give the reader the ability to search for another word in the dictionary that appears on the dictionary definition screen. 
  - Classic use case: Look-up: **stupefying**: _Adjective: That causes stupefaction._ - Not useful!
  - Would be great if we could look up **stupefaction**. This PR implements that.

## Scope Check
- In scope with the dictionary feature.

---

### AI Usage

Did you use AI tools to help write this code? _**NO**_ (pinkie promise - I do enough of that at work so hobby projects are sacred!)


ryanbarouki:
@coderabbitai review

Uri-Tauber:
Hey @ryanbarouki, thanks for the PR!

This feature was actually part of #1402, which the final slim dictionary implementation was based on.

The main reason I left it out was memory safety. Following definitions recursively can lead to heap fragmentation after a few levels, depending on the configuration (Custom fonts are big factor too).

I definitely understand the use case you're describing. I can think of a couple of possible approaches:

- Simply use a dictionary with simpler definitions - avoiding the need for long chains of lookups.
- Limit recursive lookups to a small depth (e.g. 2–3 levels). That would likely cover most real-world cases while reducing the risk of excessive memory fragmentation.

If we go with the second approach, I'd want to see thorough testing across different configurations to make sure it doesn't introduce instability.

ryanbarouki:
> Hey @ryanbarouki, thanks for the PR!
> 
> This feature was actually part of #1402, which the final slim dictionary implementation was based on.
> 
> The main reason I left it out was memory safety. Following definitions recursively can lead to heap fragmentation after a few levels, depending on the configuration (Custom fonts are big factor too).
> 
> I definitely understand the use case you're describing. I can think of a couple of possible approaches:
> 
> * Simply use a dictionary with simpler definitions - avoiding the need for long chains of lookups.
> * Limit recursive lookups to a small depth (e.g. 2–3 levels). That would likely cover most real-world cases while reducing the risk of excessive memory fragmentation.
> 
> If we go with the second approach, I'd want to see thorough testing across different configurations to make sure it doesn't introduce instability.

Thanks for the comments @Uri-Tauber! 
Ahh I see -- that makes sense!
I'm very happy to implement the second suggestion and restrict to 2-3 jumps. I agree this would cover the vast majority of use-cases.

ryanbarouki:
@coderabbitai review

Uri-Tauber:
OK, so I looked from above on how you implemented this, and the big problem here is pushing activities on `stackActivities` one on top of the other. each retaining its full definition string, lines vector, and words vector.

On top of that, `WordBox.text` is now an owned `std::string` per word (it used to be a borrowed `const char*` into the `Page` arena), and words gets cleared and rebuilt on every page turn — so even a single definition churns ~100 tiny allocations per turn, and stacked hops multiply it.

#1402 took a different approach worth exploring: commit ba75de118 introduced `replaceActivityKeepingStack()` / `PendingAction::ReplaceCurrent`, so a lookup replaces the current activity in place instead of stacking. That's the cheap, high-value part (~36 lines, no feature logic) and it kills the stack growth on its own. It also pools word text as {offset, len} into a single string instead of one `std::string` per word.

The bigger `LookupChain` mechanism there is more aspirational — it re-resolves words from the persisted lookup-history log on Back, so importing it fully drags in the history feature too. So I'd steal the pattern (replace-in-place + a compact back-stack), not copy `LookupChain.h` wholesale. If you have the time, check commits ba75de118 + f489197b6, `LookupChain.h`.

A word of caution: I didn't actually tested the chained lookup in #1402. It is possible it have other problems - so copy-pasting isn't recommended.

ryanbarouki:
Thanks @Uri-Tauber I'll have a look at these and hopefully get some time to do this soon!

ryanbarouki:
@coderabbit review

Uri-Tauber:
@ryanbarouki Sorry about the delay. We currently merging only bug fixes, so we can get release 1.5 out soon. 
I will review this after the release.

ryanbarouki:
No worries @Uri-Tauber! Thank you!

ryanbarouki:
@coderabbit.ai review

Uri-Tauber:
With release 1.5.0 shipped, I could review this properly in the next couple of days. The core problem as I see it is that the PR uses the screen-navigation history as the dictionary’s lookup history.
Normally, when you open a child screen, the previous screen is kept underneath so you can return to it. Here, you choose to chains dictionary lookups by repeatedly placing complete definition screens on that stack. But to support Back, it only needs a small breadcrumb such as:

> previous word: “stupefying”, page 2

Instead, it keeps the entire previous definition screen: its full definition text, wrapped lines, selectable words, and other display state. Every additional lookup carries another packed suitcase when a small tag would suffice.

This also causes one related problem: The definition page is described twice: once for normal display and again for word selection. A future change to fonts, spacing, headers, or styled definitions must update both descriptions. If someone updates only one, highlighting can become misaligned without any compiler warning.

The Cleaner design is to keep one dictionary screen alive, replace its displayed definition when another word is looked up, and maintain a tiny dictionary-specific Back history.

Then navigation owns navigation, the dictionary owns its lookup chain, and the definition screen remains the single authority for how its contents are laid out.

PR #2673, opened 2026-07-21
fix: preserve custom font ligatures

### LINKED ISSUE #2672: Issues with custom fonts and ligatures (e.g., ff)
### Affected Version

v1.4.1

### Describe the Bug

Using a custom serif font like Bitter or Literata (both downloaded via the "Manage fonts" inside "Settings"), I can see words like "off", where "ff" being rendered with a sans serif font. Probably an image is better than words here (the issue highlighted in yellow, what was expected in red):

Image

The issue doesn't happen with the internal Noto Serif font.

### Steps to Reproduce

1. Select a custom serif font like Bitter or Literata from "Manage fonts" in Settings (this issue may also happen with a sans serif font, but it is easier to detect with a serif font)
2. I can reproduce this with "WorldEnd v02", around 53% of the book, look at the "Is it because you wanted to show o**ff** or some other stupid reason like that"?

### Expected Behavior

"ff" should render with the proper font.

```shell

```

thiagokokada:
## Summary

* **What is the goal of this PR?** Fix custom-font ligatures rendering in a fallback font instead of the selected font. Fixes #2672.
* **What changes are included?** Update the SD-card font converter to keep supported ligatures in the custom font when the font stores them internally rather than as separately encoded characters.

## Scope Check

Please confirm:

## Additional Context

Bitter and Literata represent `ff` with an internal `f_f` glyph. The converter previously substituted the `ff` ligature codepoint, then drew it using the configured fallback font because the selected font did not expose that codepoint directly. This produced the mixed-font ligature shown in #2672.

This covers every standard ligature supported by the converter: `ff`, `fi`, `fl`, `ffi`, `ffl`, long-s + `t`, and `st`, not only `ff`.

- The converter must be re-run for the fonts distributed through **Settings > System > Manage Fonts**.
- Memory / flash impact: none at firmware runtime; this changes the host-side font conversion script only.

Validation:

- `python3 -m py_compile lib/EpdFont/scripts/fontconvert_sdcard.py`
- Converted Bitter and Literata with their real font sources and verified their internal `f_f` glyphs are kept in the output.
- Tested on device and confirmed the issue is fixed (see how it was before in https://github.com/crosspoint-reader/crosspoint-reader/issues/2672):

PXL_20260721_224350432

---

### AI Usage

### DIFF
--- a/lib/EpdFont/scripts/fontconvert_sdcard.py
+++ b/lib/EpdFont/scripts/fontconvert_sdcard.py
@@ -193,6 +193,51 @@ STANDARD_LIGATURE_MAP = {
 }
 
 
+def extract_ligature_glyph_indices_fonttools(font_path):
+    """Map standard ligature codepoints without a cmap entry to their glyph IDs.
+
+    Most fonts use unencoded glyphs such as ``f_f`` for standard ligatures.
+    The reader represents those substitutions with the Unicode compatibility
+    codepoints, so rasterize the GSUB target directly instead of falling back
+    to another font for a codepoint that the primary cmap does not contain.
+    """
+    from fontTools.ttLib import TTFont
+
+    font = TTFont(font_path)
+    cmap = font.getBestCmap() or {}
+    glyph_to_cp = {gname: cp for cp, gname in cmap.items()}
+    glyph_indices = {gname: index for index, gname in enumerate(font.getGlyphOrder())}
+    overrides = {}
+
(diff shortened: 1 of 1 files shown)


thiagokokada:
CC @itsthisjustin, can you take a look at this PR?

uxjulia (approved):

Uri-Tauber (approved):

PR #2683, opened 2026-07-22
fix: restore antialiasing after manual refresh

thiagokokada:
## Summary

* **What is the goal of this PR?** Restore text anti-aliasing after a short power-button press configured to refresh the screen.
* **What changes are included?** The forced refresh retains its clean 1-bit pass, then schedules the active EPUB, TXT, or XTC reader page's normal render pipeline to restore its grayscale overlay. Menus and non-reader screens keep the existing one-pass behavior.

## Scope Check

Please confirm:

This is a focused fix for an existing CrossPoint reader refresh behavior; it does not add a new feature category or touch the SDK/HAL, bootloader, OTA, or recovery code.

## Additional Context

The root cause is that the manual half refresh displays only the retained 1-bit framebuffer. The reader activities generate anti-aliasing/grayscale through additional render passes, which were not scheduled after the forced refresh.

Validation completed:

- `pio check`
- `pio run --target upload`
- On-device verification of the configured short power-button refresh

Memory/flash impact is limited to one activity capability and the redraw scheduling path; no persistent image buffers or settings are added.

---

### AI Usage

Did you use AI tools to help write this code? **PARTIALLY** — AI assisted implementation and review; the change was validated on device by the contributor.

### DIFF
--- a/nix/flake.nix
+++ b/nix/flake.nix
@@ -69,6 +69,7 @@
             packages = [
               pio
               fhsEnv
+              pkgs.clang-tools # for clang-format
             ];
 
             shellHook = setEnvs;
--- a/src/activities/reader/EpubReaderActivity.h
+++ b/src/activities/reader/EpubReaderActivity.h
@@ -21,6 +21,9 @@ class EpubReaderActivity final : public Activity {
   // Cleared on the next render after the new section loads and resolves it to a page.
   std::string pendingAnchor;
   int pagesUntilFullRefresh = 0;
+  // Image pages use a dedicated double-FAST refresh path, so retain a manual
+  // refresh request until renderContents can issue its clean base pass.
+  bool forcedRefreshPending = false;
   int cachedSpineIndex = 0;
   int cachedChapterTotalPageCount = 0;
   unsigned long lastPageTurnTime = 0UL;
@@ -198,6 +201,7 @@ class EpubReaderActivity final : public Activity {
   bool isReaderActivity() const override { return true; }
   bool handleForcedRefresh() override {
     pagesUntilFullRefresh = 1;
+    forcedRefreshPending = true;
     requestUpdate();
     return true;
   }
--- a/src/activities/reader/EpubReaderActivity.h
+++ b/src/activities/reader/EpubReaderActivity.h
@@ -196,6 +196,7 @@ class EpubReaderActivity final : public Activity {
   // speed would only burn battery; the paused gate still retries every loop pass).
   bool skipLoopDelay() override { return section && section->isBuilding() && !buildHeapPaused; }
   bool isReaderActivity() const override { return true; }
+  bool restoresGrayscaleAfterForcedRefresh() const override { return true; }
   ScreenshotInfo getScreenshotInfo() const override;
   CrossPointPosition getCurrentPosition() const;
 };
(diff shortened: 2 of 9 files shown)


Uri-Tauber:
Thx for the contribution.

One suggestion on the mechanism: Right now we drive the panel BW at `main.cpp:539` and then queue a render that drives it again, so a single press costs two-plus panel cycles.
Would it be worth letting the activity handle the forced refresh rather than following it? The readers already own the right hook:
```
// Activity.h — default keeps today's behavior for non-readers
virtual bool handleForcedRefresh() { return false; }

// readers:
bool handleForcedRefresh() override {
  pagesUntilFullRefresh = 1;
  requestUpdate();
  return true;
}

// main.cpp
if (!activityManager.handleForcedRefresh()) {
  RenderLock lock;
  renderer.displayBuffer(HalDisplay::HALF_REFRESH);
}
```

That gives one HALF_REFRESH with the grayscale pass on top instead of two-to-four cycles.

thiagokokada:
> Thx for the contribution.
> 
> One suggestion on the mechanism: Right now we drive the panel BW at `main.cpp:539` and then queue a render that drives it again, so a single press costs two-plus panel cycles. Would it be worth letting the activity handle the forced refresh rather than following it? The readers already own the right hook:
> 
> ```
> // Activity.h — default keeps today's behavior for non-readers
> virtual bool handleForcedRefresh() { return false; }
> 
> // readers:
> bool handleForcedRefresh() override {
>   pagesUntilFullRefresh = 1;
>   requestUpdate();
>   return true;
> }
> 
> // main.cpp
> if (!activityManager.handleForcedRefresh()) {
>   RenderLock lock;
>   renderer.displayBuffer(HalDisplay::HALF_REFRESH);
> }
> ```
> 
> That gives one HALF_REFRESH with the grayscale pass on top instead of two-to-four cycles.

Done, needed one additional change for pages with images. When rendering an EPUB page with an image, it performs the clean half-refresh immediately before the existing image pipeline, then applies the usual grayscale passes. Text-only pages continue to use the normal single cleanup refresh path.

nix/flake.nix
```diff
@@ -69,6 +69,7 @@
             packages = [
               pio
               fhsEnv
+              pkgs.clang-tools # for clang-format
```
thiagokokada, line 72: Adding this so I can run `./bin/clang-format-fix` successfully and looks line a minimal change, but if you want I can make this a separate PR.

src/activities/reader/EpubReaderActivity.h
```diff
@@ -196,6 +199,12 @@ class EpubReaderActivity final : public Activity {
   // speed would only burn battery; the paused gate still retries every loop pass).
   bool skipLoopDelay() override { return section && section->isBuilding() && !buildHeapPaused; }
   bool isReaderActivity() const override { return true; }
+  bool handleForcedRefresh() override {
+    pagesUntilFullRefresh = 1;
+    forcedRefreshPending = true;
+    requestUpdate();
```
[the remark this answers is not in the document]
thiagokokada: @coderabbitai Can you recheck the code?

thiagokokada:
CC @Uri-Tauber, I think this is ready to review.

nix/flake.nix
Uri-Tauber: Nah, Thats fine.

Uri-Tauber (approved):
LGTM

thiagokokada:
> LGTM

Hey thanks for merging that fast.

@Uri-Tauber Just a question, is this only going to be part to `release/1.5.0` branch or also you will also merge in `develop` branch?

Uri-Tauber:
> @Uri-Tauber Just a question, is this only going to be part to release/1.5.0 branch or also you will also merge in develop branch?

We're adopting [GitFlow](https://nvie.com/posts/a-successful-git-branching-model/) now. I'm still getting familiar with the workflow myself, but as I understand it, for the next week or so, bug fixes will go into `release/1.5.0`, while new features will be merged into `develop`.

Once `release/1.5.0` is considered stable enough for release, it will be merged back into `develop`.

Issue #2395, opened 2026-06-22
Specific page of ebook reliably crashes Crosspoint when generating QR code


QuestionableTangerine:
### Affected Version

1.3.0, XTEINK 4

### Describe the Bug

<img width="480" height="800" alt="Image" src="https://github.com/user-attachments/assets/b628ef28-14c8-490e-b1a7-3c0b997f24c2" />
When attempting to generate a QR code of the above page, Crosspoint has a 100% crash rate (across 5+ tests after various methods of attempting a workaround). Crash persists after chapter file is reindexed but some pages displaying subsets of the text (eg. changing orientation to landscape) do not crash. Other pages in the same ebook (including the pages directly before and after this one) do not crash the device.

### Steps to Reproduce

1. [Load Kings Avatar epub](https://filebin.net/h6yq7g0a67frutzw)
2. Navigate to Chapter 408, page 18 (using stock font Noto Sans, Font Size Medium, Screen Margin 15) which starts "Sun Xiang is a piece of trash!"
3. Attempt to take QR of page
4. Crosspoint crashes and reboots

### Expected Behavior

QR code will load so I can transfer text quote to my phone, Crosspoint does not crash.

### Relevant Log Output/Screenshots

```shell
CrossPoint version: 1.3.0

Panic reason: 

Last logs:
[22] [INF] [HW] Using cached device type: X4
[23] [INF] [MAIN] Hardware detect: X4
[2665] [ERR] [GFX] Failed to read row 0 from bitmap
[36475] [ERR] [SCT] Deserialization failed: Parameters do not match
[22] [INF] [HW] Using cached device type: X4
[23] [INF] [MAIN] Hardware detect: X4


Stack memory:
0x3FCB6020: 0x288E1840 0x5637C509 0x968605E2 0x0672E616 0x12063297 0x56960607 0xF6065236 0x26470762 
0x3FCB6040: 0x1E823617 0x04D20928 0xE2568636 0xF2567704 0xD6D69606 0x17964656 0x92C75646 0x32974606 
0x3FCB6060: 0x164606D2 0x4256E696 0x09280EE2 0x066296C4 0x26075286 0xC7C61656 0x17760792 0x26470732 
0x3FCB6080: 0xC2823617 0x56864607 0x864607E2 0x285E2657 0x06329709 0xF606F2E6 0x960652E6 0x864607E2 
[...]

PR #2686, opened 2026-07-23
fix: empty panic reason on raw faults + garbled Last logs: from unsynchronized ring buffer

DarkStarDS9:
## Summary

Two bugs in the panic/crash-log infrastructure (`lib/hal/HalSystem.cpp`, `lib/Logging/Logging.cpp`), found while pulling crash reports off a device via a serial debug command.

### 1. `Panic reason:` empty for raw architectural faults

`__wrap_panic_abort` (the existing hook that populates `panicMessage`) only runs for explicit `abort()`/assert paths. Raw faults (LoadProhibited, StoreProhibited, illegal instruction, ...) never call `panic_abort()` — they go straight through esp-idf's `panic_arch_fill_info()`, which derives the reason purely from the RISC-V `mcause` CSR. Reproduced live in #2395 (`Panic reason: ` is blank there too, unrelated crash).

Fix: `__wrap_panic_print_backtrace` (the other existing hook, which runs unconditionally for every panic type via esp-idf's `print_state()`) now decodes `mcause` from the frame it already receives and backfills `panicMessage`, but only when it's still empty — so a real `abort()`/assert message is never overwritten. The decode table mirrors esp-idf's own `panic_arch_fill_info()` (`components/esp_system/port/arch/riscv/panic_arch.c`) exactly; only standard architectural exceptions are covered, SoC-level pseudo-causes (watchdog, cache error) are intentionally left undecoded rather than guessed at.

### 2. `Last logs:` garbled/spliced entries

Traced a real corrupted report back to two separate causes in `Logging.cpp`:

- **Unsynchronized ring buffer writes.** `LOG_*` is called from multiple FreeRTOS tasks (e.g. main loop + a render task) with no locking around `logHead`/`logMessages`. Two tasks racing on the same slot — one preempted mid-`strncpy` by the other — can leave a shorter new message's write interrupted before its null-padding finishes clearing a longer previous occupant's tail, producing exactly this in a live report:
  `"Time = 141 ms from clearScreen to displayBuffers from clearScreen to displayBuffer"`
  (the trailing `s from clearScreen to displayBuffer` is leftover from an earlier, longer message in the same slot — not truncation, since that message is far under the 256-byte entry limit). Fixed with a mutex around all ring-buffer access, mirroring the existing `HalStorage::storageMutex` convention. Confirmed `LOG_*` is never called from ISR context, so a plain (non-critical-section) mutex is sufficient.
- **Truncation splice.** Several call sites interpolate unbounded strings (XPath/doc-hash strings, file paths) into the 256-byte entry buffer. If the formatted line exceeds that, `vsnprintf` truncates and the trailing `\n` (last char of the format string) is exactly what gets cut, so `getLastLogs()` would concatenate that entry straight into the next one with no separator. Fixed by having `getLastLogs()` always emit exactly one `\n` per stored entry regardless of what was actually stored.

Both fixes are needed — they cover disjoint failure modes, not overlapping ones.

## Scope Check

- [x] The stock firmware does not already handle this well — these are genuine bugs (confirmed live against #2395 and a real garbled crash report), not a new feature.
- [ ] If this PR touches freeink-sdk/, lib/hal/, the bootloader, OTA, or recovery code, I have coordinated with the relevant maintainer.

This PR touches `lib/hal/` (`HalSystem.cpp`). **I have not coordinated with a maintainer beforehand** — this is a first-time external contribution. The change is narrowly scoped to the existing panic/log-capture hooks (decoding an already-available `mcause` value, adding a mutex around an existing ring buffer) and does not touch peripheral drivers, the bootloader, OTA, or recovery code, but flagging this honestly per the checklist rather than checking a box I can't back up. Happy to loop in whoever owns that area if that's the expected process.

## Testing performed

No automated test harness for this firmware, per project convention — verified against real hardware (an X3) via temporary serial debug commands (not included in this diff), then reverted before opening this PR:

- **Panic reason fix:** intentionally triggered a raw fault (`volatile uint32_t* p = (uint32_t*)0x1; *p = 0xDEADBEEF;`). Device panicked with `MCAUSE=0x00000005` (Load access fault, confirmed by esp-idf's own onboard "Guru Meditation Error" text). After reboot, the crash report's `Panic reason:` field correctly showed `Load access fault` instead of being blank.
- **Ring buffer race fix:** ran 16 stress rounds, each spawning two FreeRTOS tasks hammering `LOG_DBG` concurrently with zero throttling (~10,000-20,000 log calls per task in under 2 seconds each — a much heavier load than whatever intermittent real-world race originally produced the garbled report). Dumped `getLastLogs()` after each round and validated every entry against the exact expected format (correct prefix, correct payload, correct size). 16/16 runs clean — no splicing, no interleaving, no truncated entries.
- Confirmed a normal `pio run` build succeeds with no new warnings, and `bin/clang-format-fix` produces a clean diff.

Given the ring-buffer bug is timing-dependent, a clean report is strong evidence but not absolute proof the race can never occur — flagging that honestly rather than overclaiming. One residual, accepted edge case: a crash occurring exactly while a writer holds the mutex mid-`strncpy` would still leave that one slot half-written on the next boot.

## Additional Context

- **Memory/flash impact:** negligible. The mcause decode table adds a handful of small `DRAM_ATTR` string constants (well under 100 bytes); the logging mutex adds one `SemaphoreHandle_t` plus its FreeRTOS control block. No measurable change to flash usage beyond the new code itself.
- **Risk:** both changes are additive/defensive — they only run on already-broken paths (empty panic reason, garbled logs) and don't alter normal control flow. The mutex does add a small amount of serialization to every `LOG_*` call, which was already effectively serialized by the single physical core; no deadlock path exists since none of the guarded functions call each other reentrantly.
- Related: #2395 (empty `Panic reason:`, same symptom on unrelated code).

## AI Usage

YES

This was diagnosed, fixed and tested by Claude (Sonnet 5). Apart from connecting the hardware and turning it on, no human was involved. Feel free to refuse / ignore this PR on that grounds.

### DIFF
--- a/lib/Logging/Logging.cpp
+++ b/lib/Logging/Logging.cpp
@@ -2,6 +2,8 @@
 
 #include <BoardConfig.h>
 #include <esp_rom_sys.h>
+#include <freertos/FreeRTOS.h>
+#include <freertos/semphr.h>
 
 #include <string>
 
@@ -19,7 +21,40 @@ RTC_NOINIT_ATTR size_t logHead = 0;
 RTC_NOINIT_ATTR uint32_t rtcLogMagic;
 static constexpr uint32_t LOG_RTC_MAGIC = 0xDEADBEEF;
 
+// LOG_* is called from multiple FreeRTOS tasks (main loop task + ActivityManager's render task, see
+// ActivityManager::renderTaskLoop()) without any prior synchronization. Two calls racing on the same
+// logHead slot -- one preempted mid-strncpy by the other -- can leave a shorter new message's write
+// interrupted before its null-padding finishes clearing the slot, so a stale tail from the previous
+// occupant survives right after it. That produced exactly this in a real crash report:
+//   "Time = 141 ms from clearScreen to displayBuffers from clearScreen to displayBuffer"
+// (the "s from clearScreen to displayBuffer" tail is leftover from a longer earlier message in the same
+// slot, not truncation -- the message itself is far under MAX_ENTRY_LEN). This mutex serializes all
+// ring-buffer access. Never called from ISR context (grep for LOG_* inside IRAM_ATTR functions turns up
+// only HalSystem's non-IRAM checkPanic()), so a plain (non-ISR) mutex is sufficient.
+static SemaphoreHandle_t logMutex = xSemaphoreCreateMutex();
+
+namespace {
+// RAII helper; falls back to no locking if logMutex hasn't been constructed yet (only possible if some
+// other translation unit's global constructor calls a LOG_* macro before this file's own global
+// initializer has run -- C++ does not guarantee cross-TU init order).
+struct LogLock {
+  bool locked = false;
+  LogLock() {
+    if (logMutex) {
+      xSemaphoreTake(logMutex, portMAX_DELAY);
+      locked = true;
+    }
+  }
+  ~LogLock() {
+    if (locked) {
+      xSemaphoreGive(logMutex);
+    }
+  }
+};
+}  // namespace
+
 void addToLogRingBuffer(const char* message) {
+  LogLock lock;
   // Add the message to the ring buffer, overwriting old messages if necessary.
   // If the magic is wrong or logHead is out of range (RTC_NOINIT_ATTR garbage
   // on cold boot), clear the entire buffer so subsequent reads are safe.
@@ -76,6 +111,7 @@ void logPrintf(const char* level, const char* origin, const char* format, ...) {
 }
 
 std::string getLastLogs() {
+  LogLock lock;
   if (rtcLogMagic != LOG_RTC_MAGIC) {
     return {};
   }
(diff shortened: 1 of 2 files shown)


DarkStarDS9:
Per the PR template's `lib/hal/` coordination checkbox: this PR touches `lib/hal/HalSystem.cpp` (panic-reason decoding) and `lib/Logging/Logging.cpp` (log ring buffer). Tagging @itsthisjustin as the most active recent contributor to `lib/hal/`, and @daveallie per GOVERNANCE.md, since I couldn't find a published owner for this area and wanted to coordinate before this leaves draft rather than assume it's fine.

Happy to answer questions or make changes based on feedback — no rush, just flagging per the checklist.

itsthisjustin:
@Uri-Tauber you are probably better fit than I am to look at this one. Also looks like we need to update the governance file haha

PR #2696, opened 2026-07-24
feat: StarDict .syn synonym lookup support

W-Floyd:
## Summary

* **Goal:** Add StarDict `.syn` synonym-file support to the reader's dictionary lookup — the follow-up explicitly listed as "deliberately cut" in the Slim Dictionary PR (#2583).
* **Changes:** Resolve alternate spellings and irregular forms (e.g. `theatre`→`theater`, `oxen`→`ox`) via a dictionary's optional `.syn` file, slotting in between the direct index match and the English stemmer in `Dictionary::lookup()`.

**How it works.** A `.syn` entry carries an *ordinal* (the N-th `.idx` entry), not a byte offset. The existing `.qidx` sidecar already samples byte offsets at fixed entry-count boundaries, so it doubles as an ordinal→offset map: `locateByOrdinal()` reads sample `N / SAMPLE_INTERVAL` then scans the remainder. A parallel `.sidx` sidecar makes the `.syn` itself binary-searchable, mirroring `.qidx` over `.idx`. The sidecar header is generalized (`QidxHeader`→`SidecarHeader`) and gains an `entryCount` field to bounds-check ordinals; version bumped 1→2 so stale `.qidx` files auto-rebuild. **Zero new persistent RAM** — everything streams from SD reusing the existing scan buffer. All BE32 reads go through the existing byte-assembly helper (RISC-V alignment-safe).

**Composing with #2733.** Rebased onto current `develop`, so this now builds on the `LookupSession` from *"perf: open dictionary index files once per lookup"* rather than reopening files per probe. All four search entry points (`locate`, `locateByOrdinal`, `locateSynonym`) take the session, so a synonym lookup adds **no** SD opens or `std::string` path temporaries to the hot path:

* `.syn` / `.sidx` are opened **lazily**, by `openSynonyms()` on first use. Only an exact-match miss consults them, so a direct hit never pays for the two extra opens.
* `buildSidecar()` propagates #2706's `IndexResult`, so an OOM during indexing still reports "Not enough memory" rather than a generic failure. The `.syn` pass is best-effort and passes `nullptr` — a failed synonym index disables synonyms for the session but cannot downgrade the build result or break the dictionary.
* `sidecarIsStale()` checks each sidecar independently, so editing a `.syn` doesn't force a needless rescan of the (much larger) `.idx`, and vice versa.
* `locateSynonym()` / `locateByOrdinal()` set `readError` on seek failure, preserving #2706's "a search that never reached a verdict is not a miss" contract.

## Scope Check

CrossPoint is intentionally narrow. See [SCOPE.md](../blob/develop/SCOPE.md) and [ROADMAP.md](../blob/develop/ROADMAP.md).

- [x] The stock firmware does not already handle this — it improves an already-in-scope feature (dictionary lookup, #917) by raising hit-rate on inflected/variant forms at ~zero RAM cost.
- [x] Does **not** touch `freeink-sdk/`, `lib/hal/`, the bootloader, OTA, or recovery code — only `src/util/Dictionary.*`, a docs update, and a test fixture. No maintainer coordination required on that basis (happy to coordinate with the dictionary-area maintainers regardless — see below).

This is the `.syn` follow-up seam from #2583. Not currently on a ROADMAP milestone or the #1599 to-do list, so flagging for a maintainer nod. Rebased onto `develop` — #2733, #2706 and #2877 have landed since this was opened and are all preserved (see "Composing with #2733" above). Still aware of dictionary rework in flight (#2663, #2690, mcrosson's orchestration branch) — glad to rebase/coordinate again.

## Additional Context

* **Memory/flash impact:** No new persistent RAM. The per-lookup `LookupSession` is a stack local and grows by two `HalFile` handles plus three `uint32_t` — `HalFile` is pimpl with a defaulted constructor, so that is 16 bytes of stack and **zero heap** unless a miss actually opens the synonym files. Two additional small SD sidecars (`.sidx`, and the existing `.qidx` gains one `uint32`). Streaming-only; reuses the 256-byte scan buffer.
* **Cache format:** `.qidx` version 1→2, so existing dictionaries re-index once on first open (existing "Indexing dictionary…" flow). Sidecars are safe to delete; they rebuild.
* Docs updated in `docs/dictionary.md` (`.syn` moved from "Not supported" to supported, lookup order and per-sidecar rebuild documented).

### Testing

Tested on device (X3). Using the **GNU Collaborative International Dictionary of English (GCIDE)** StarDict build — which ships an 11,466-entry `.syn` — from:
https://tuxor1337.frama.io/firedict/dictionaries.html

A test EPUB is included at `test/epubs/test_dictionary_synonyms.epub` (generator: `scripts/generate_dictionary_synonyms_test_epub.py`). Every word in it is a direct-index **miss** resolvable only through `.syn`, and also unreachable by the stemmer, so a hit exercises the synonym path specifically:

| Look up | Resolves to |
|---|---|
| theatre | theater |
| sceptre | scepter |
| absinthe | absinth |
| accoutre | accouter |
| accurst | accursed |
| abettor | abetter |
| aardwolf | aard-wolf |

Control words (`dog`, `cat`) are direct hits, confirming the reader isn't just failing everything. Verified each test word resolves to its headword on device; control words behave as before.

**Post-rebase status:** builds clean (`pio run -e default`, 0 warnings) and clang-format clean. The on-device pass above was run before the rebase onto #2733's session refactor; re-verifying the synonym path and the one-time `.qidx` v1→v2 rebuild on hardware before merge.

---

### AI Usage

Did you use AI tools to help write this code? _**YES**_ — implemented with AI assistance (design, code, and this test fixture); reviewed and tested on hardware by me.


W-Floyd:
Thanks @coderabbitai — addressed in 18950f42.

**`buildIndex()` (Dictionary.cpp) — fixed.** Both points were valid:
- On a `.sidx` build failure, `hasSyn` now flips to `false`. Previously the log claimed "synonyms disabled" but the flag stayed `true`, so — with the failed `.sidx` already removed — every subsequent miss fell into an unbisected from-byte-0 linear scan of `.syn` for the rest of the session. It now skips `locateSynonym()` until the next `open()` retries the build.
- Factored the per-sidecar freshness check into a shared `sidecarIsStale()` helper used by both `needsIndex()` and `buildIndex()`, so each sidecar rebuilds only when actually stale. A `.syn`-only change no longer forces a full rescan of the larger `.idx`, and vice versa. Original edge cases preserved (missing `.idx` / vanished `.syn` → not stale → graceful degradation).

**`handleForcedRefresh()` (EpubReaderActivity.h) — not changed (non-issue).** Clearing `forcedRefreshPending` without using it on non-image render paths is intentional: `handleForcedRefresh()` also sets `pagesUntilFullRefresh = 1`, so the clean refresh still happens through the normal cadence; the flag only matters for the image double-FAST path that bypasses it. (This is also #2683's plumbing, which will drop out of this diff once the branch rebases onto `develop`.)

Uri-Tauber:
Hey @W-Floyd. Thanks for doing this. Was on my TODO list.
I noticed you branched off `release/1.5.0`, which add two commits unrelated to this PR. can you rebase on `develop`?

W-Floyd:
@Uri-Tauber done

Uri-Tauber:
@W-Floyd Thanks. I will review this after release 1.5.0. Probably a week or so. 
Ping me if you think I forgot.

W-Floyd:
Will do, and no problem, there's no rush

Uri-Tauber:
@W-Floyd Could you rebase on `develop` and solve conflicts?

W-Floyd:
Sure, I'll take a look in the next few days at least.

src/util/Dictionary.cpp
```diff
@@ -308,6 +380,94 @@ DictLocation Dictionary::locate(LookupSession& session, const char* target, std:
...
+        break;
+      }
+      if (StringUtils::asciiCaseCmp(wordBuf, target) <= 0) {
+        lo = mid;
+      } else {
+        hi = mid - 1;
+      }
+    }
+    readSampleOffset(session.sidx, lo, &startByte);
+  }
```
Uri-Tauber: This section looks identical to the sidecar bisection loop in `locate()` (L334–L352).
Can you extract shared helper to clean the code?

Uri-Tauber (approved):
I haven't tested yet. but the code LGTM.

PR #2705, opened 2026-07-24
feat: Adds dark mode with image polarity preservation

itsthisjustin:
Implements dark mode functionality by adding display inversion at the driver level while preserving original image polarity. Images are counter-inverted in the framebuffer so they maintain their original appearance when output inversion is applied. Includes HAL methods to toggle and query inversion state.


Uri-Tauber:
@itsthisjustin Can you rebase on latest develop and fix conflicts?

Also, I think it will be clearer if the setting will be called "Night Mode"-> off/on. (like koreader)

Uri-Tauber:
Great!

Uri-Tauber:
Running on my device and working well. 
Just one thing to verify before merging: when setting Night Mode ON and entering the reader, it take half a second before the night mode start - you see the book in normal mode, and after a second the night mode start. when entering the reader menu or going back to home, the effect reversed.

I find it nice-looking, but I want to make sure it's intentional.

itsthisjustin:
> Running on my device and working well. 
> Just one thing to verify before merging: when setting Night Mode ON and entering the reader, it take half a second before the night mode start - you see the book in normal mode, and after a second the night mode start. when entering the reader menu or going back to home, the effect reversed.
> 
> I find it nice-looking, but I want to make sure it's intentional. 

If you notice, light mode also does this but obviously the opposite of that. It's how it's always worked prior to this change

Uri-Tauber (approved):

PR #2706, opened 2026-07-24
fix: name dictionary lookup failures (low-memory vs decompress vs read) instead of 'Not found'

W-Floyd:
## Summary

* **Goal:** Stop a real, transient dictionary failure from masquerading as a missing word. A word that **is** in the `.idx` but whose definition can't be read used to return the same `false` as a genuine miss, so the reader showed **"Not found"**. This is the cause of the reported *"dictionary just stopped working… simply stopped finding words… after a reset it's working again"* — for a `.dict.dz` dictionary, each lookup inflates the definition on the fly, which needs a ~32 KB **contiguous** buffer (the zlib window); once a long reading session fragments the heap, that allocation fails, decompression fails, and the lookup silently reports "not found". A reset restores an unfragmented heap.
* **Changes:** `Dictionary::lookup()` now reports a precise `LookupResult` and names the failure — and is **certain when the cause is low memory**:

  | Result | Cause | Message |
  |---|---|---|
  | `LowMemory` | an allocation failed (the ~32 KB inflate window / chunk buffer, or the definition text buffer) — heap fragmentation, the "stops finding words until restart" case | **Not enough memory** |
  | `Decompress` | decompression genuinely failed: corrupt/truncated `.dict.dz` (read/inflate error, *not* memory) | **Couldn't decompress definition** |
  | `ReadError` | file open / bounds / IO error | **Couldn't read definition** |
  | `NotFound` | word genuinely absent | **Not found** (unchanged) |

  To tell low-memory apart from a genuinely corrupt `.dz`, `DictZip::extractEntry()` gained an `outLowMemory` flag set only at its allocation points (`InflateReader::init()` returns false solely when the 32 KB window `malloc` fails; the `makeUniqueNoThrow` chunk buffer). Serial logging is disambiguated too: `dictzip extraction failed … (low memory)` vs `(read/inflate error)`. A normal hit or genuine miss is unchanged — this only splits the previously-overloaded `false`.

## Scope Check

Please confirm:

- [x] The stock firmware does not already handle this well, **and** no other popular CrossPoint fork already does. This is a small correctness/diagnostics fix to an already-in-scope feature (dictionary lookup) — stock currently reports a memory failure as "Not found".

## Additional Context

* **Behaviour-only on the failure path.** A hit and a genuine miss are byte-for-byte the same as before; only the *reason* reported on a read failure changed.
* **Memory / flash impact:** negligible — no new persistent RAM; one `bool` out-param threaded through the lookup path; three new English UI strings (`STR_DICT_LOW_MEMORY`, `STR_DICT_DECOMPRESS_ERROR`, `STR_DICT_READ_FAILED`; other languages fall back).
* **Not a fix for the fragmentation itself** — that's a follow-up (an opt-in Settings action that decompresses `.dict.dz` to a plain sidecar so lookups never inflate on the fly), kept out of this minimal, low-risk change.
* Files: `src/util/Dictionary.{h,cpp}`, `src/util/DictZip.{h,cpp}`, `src/activities/reader/DictionaryWordSelectActivity.cpp`, `lib/I18n/translations/english.yaml`.

### Testing

* Builds clean for the device `default` env (ESP32-C3).
* **Tested and verified on hardware (Xteink X3):** a `.dict.dz` lookup that previously reported "Not found" under heap pressure now reports the accurate memory error.

---

### AI Usage

Did you use AI tools to help write this code? _**YES**_ — implemented with AI assistance (design + code + this description); reviewed and tested on hardware (X3) by me.


W-Floyd:
PXL_20260724_210138615

W-Floyd:
Thanks @coderabbitai — went through all three:

**✅ Typed extraction result (DictZip.h / .cpp / Dictionary.cpp) — done in 9d2c6b5.** Replaced the `outLowMemory` bool with a typed `DictZip::ExtractError { LowMemory, ReadError, Decompress }`. `extractEntry`/`extractChunkSlice` now return `ReadError` for file open/read/write/bad-offset, `Decompress` only for inflate/parse failures, and `LowMemory` for the allocation failures; `readDefinition` maps each to the matching `LookupResult`. This is squarely on-theme (the PR is about accurate failure naming), so an IO glitch mid-extraction is no longer mislabeled "decompress."

**⏭️ Failed seek / short read (Dictionary.cpp) — skipped.** The `seekSet(offset)` is guarded two lines above by the bounds check (`offset > dictSize || size > dictSize - offset` → `ReadError`), so on a valid file it can't fail here — adding a check would be an unreachable guard, which this repo's conventions discourage. The short-read `resize` is intentional pre-existing tolerance (returns the bytes actually present); turning it into a hard failure is a behavior change beyond this reporting-only fix and risks regressing dictionaries whose final entry size overshoots EOF.

**⏭️ `parse()` `chunkOffsets.reserve` abort under `-fno-exceptions` — skipped here.** Real but **pre-existing** (predates this PR) and low-probability: `chunkCount` is ~900 for real dictionaries (GCIDE/reader.dict ≈ a few KB reserve), not the 8192 max. Hardening `parse()` against OOM (a heap pre-check before building the chunk table) is worth doing but is a separate concern from naming lookup failures — happy to do it as a follow-up rather than widen this minimal, hardware-tested change.

Verified: device `default` env builds clean; clang-format-21 clean. Tested on X3 (the LowMemory path is unchanged from the verified build).

W-Floyd:
Good catch @coderabbitai — done in d1cc410. Added the guard in `extractChunkSlice`:

```cpp
if (!file.seekSet(compressedOffset)) return fail(ExtractError::ReadError);
```

This one is worth it (unlike the `readDefinition` seek, which is already covered by the preceding bounds check): `compressedOffset` is derived from the untrusted `.dz` chunk table, so a malformed table could seek out of range and the subsequent read would otherwise come from the wrong position. Now a failed seek reports `ReadError` deterministically.

Device `default` builds clean; clang-format-21 clean.

W-Floyd:
@coderabbitai Done in 4e98153 — went the full route rather than just documenting, and audited **every** `return false` in the extraction path.

**`parse()` now reports a typed cause** (`ExtractError* outError`), so `Dictionary::readDefinition` gets the right classification:
- header/table/trailer **read** shortfalls (the initial 10-byte read, `readLe16` for xlen/version/chunkLen/chunkCount/compLen, FNAME/FCOMMENT/FHCRC reads, the ISIZE read) → `ReadError`.
- malformed/unsupported values (bad magic, missing FEXTRA, `subLen`/`xlen` mismatches, bad RA version/chunkLen/chunkCount, `!foundRa`, `fileSize < 4`, `totalSize == 0`) → `Decompress`.

**The audit surfaced two more issues I fixed:**
1. `extractEntry` ended with `return remaining == 0;` — a `false` return that left `outError` unset. Now returns `Decompress` (chunk table inconsistent with the requested range).
2. Two unchecked `seekSet` calls in `parse()` (skipping a non-RA subfield; seeking to the ISIZE trailer) now return `ReadError` on failure, matching the `extractChunkSlice` guard.

Every remaining bare `return false` is either the `readLe16` helper (its call sites map to `ReadError`), a `fail(...)` lambda body, or a direct propagation of `parse`/`extractChunkSlice` (which already set `outError`).

Device `default` builds clean; clang-format-21 clean.

W-Floyd:
@coderabbitai closed the last open item too — the `chunkOffsets.reserve()` OOM (the Major) is now fixed in c8d7e54 rather than deferred, since it's a potential `abort()` in the exact low-memory path this PR is about.

`parse()` now guards the chunk-table allocation before reserving, mirroring `Dictionary::readDefinition`'s existing heap check:

```cpp
const size_t chunkTableBytes = (static_cast<size_t>(chunkCount) + 1) * sizeof(uint32_t);
if (ESP.getMaxAllocHeap() < chunkTableBytes + 1024) return fail(ExtractError::LowMemory);
info->chunkOffsets.reserve(static_cast<size_t>(chunkCount) + 1);
```

So a fragmented heap surfaces `LowMemory` → "Not enough memory" instead of crashing. (`chunkCount <= MAX_CHUNK_COUNT` caps the table at ~32 KB; in practice it fails only at a far more degraded heap state than the ~30 KB-largest-block bug this PR was opened for, where the inflate window fails first.)

That resolves all four review comments:
- ✅ typed `ExtractError` (DictZip.h)
- ✅ parse-time IO vs decompress classification + full return-false audit (DictZip.cpp)
- ✅ `reserve()` OOM guard (DictZip.cpp) — this one
- ⏭️ `readDefinition` seek / short-read (Dictionary.cpp) — skipped with rationale you accepted (unreachable seek; intentional short-final-entry tolerance)

Device `default` builds clean; clang-format-21 clean.

W-Floyd:
@coderabbitai done in a2ab3b4 — `parse()` now sets the out-param on the success path too:

```cpp
if (outError) *outError = ExtractError::None;  // establish the out-param on every path
```

Good contract-hygiene catch: the sole current caller (`extractEntry`) pre-initializes it, but `parse()` is public, so it shouldn't leave a reused `outError` stale after a successful parse.

Device `default` builds clean; clang-format-21 clean.

src/util/Dictionary.cpp
```diff
@@ -279,38 +283,49 @@ bool Dictionary::readDefinition(const DictLocation& location, std::string& out)
...
 
   // std::string growth aborts on OOM (-fno-exceptions); refuse up front unless
   // the allocation fits comfortably in the largest free block.
   if (ESP.getMaxAllocHeap() < size + 8 * 1024) {
     LOG_ERR("DICT", "Low heap for %lu byte definition", static_cast<unsigned long>(size));
-    return false;
+    return fail(LookupResult::LowMemory);
   }
 
   dict.seekSet(offset);
```
Uri-Tauber: This PR hardens exactly this pattern three times in `DictZip.cpp` — `if (!file.seekSet(...)) return fail(ReadError);`

```suggestion
  if (!dict.seekSet(offset)) {
    LOG_ERR("DICT", "Seek to %lu failed", static_cast<unsigned long>(offset));
    return fail(LookupResult::ReadError);
  }
```

Same in line 250, outside the diff.

src/util/Dictionary.cpp
```diff
@@ -373,6 +393,13 @@ bool Dictionary::lookup(const char* word, std::string& definitionOut, std::strin
       if (location.found) break;
     }
   }
-  if (!location.found) return false;
-  return readDefinition(location, definitionOut);
+  if (!location.found) return false;  // genuinely not in the dictionary
```
Uri-Tauber: Hmm. without exploring too much, I can see a case where the location isn't found, but because of failure to open file (line 218).

src/util/DictZip.cpp
```diff
@@ -52,108 +60,132 @@ bool extractChunkSlice(HalFile& file, uint32_t compressedOffset, uint32_t compre
...
+    if (file.read(subHeader, sizeof(subHeader)) != static_cast<int>(sizeof(subHeader)))
+      return fail(ExtractError::ReadError);
     extraRead += 4;
     const uint16_t subLen = static_cast<uint16_t>(subHeader[2] | (static_cast<uint16_t>(subHeader[3]) << 8));
-    if (extraRead + subLen > xlen) return false;
+    if (extraRead + subLen > xlen) return fail(ExtractError::Decompress);
 
     if (subHeader[0] == 'R' && subHeader[1] == 'A') {
-      if (subLen < 6) return false;
+      if (subLen < 6) return fail(ExtractError::Decompress);
```
Uri-Tauber, line 105: `foundRa` is only consulted after the loop (line 131), never as a re-entry guard inside it — and the branch doesn't break — so a .dz whose FEXTRA carries two 'RA' subfields runs this whole block twice. The line-114 guard passes on both passes, but `reserve()` is a no-op the second time (capacity is already `chunkCount + 1`), so the `push_back`s grow through `std::vector`'s bare operator new, which aborts under -fno-exceptions — the exact crash line 114 was added to prevent.

```suggestion
      if (foundRa) return fail(ExtractError::Decompress);  // a second RA table would append past reserve()
      if (subLen < 6) return fail(ExtractError::Decompress);
```

Guarding at the top rather than `break`ing after line 125 is deliberate: breaking out of the loop would leave `extraRead < xlen` and trip the `extraRead != xlen` check at line 131, rejecting valid files that carry other FEXTRA subfields after the RA one.

src/activities/reader/DictionaryWordSelectActivity.cpp
```diff
@@ -172,8 +173,33 @@ void DictionaryWordSelectActivity::performLookup() {
                            [this](const ActivityResult&) { requestUpdate(); });
     return;
   }
-  popup = ok ? Popup::NotFound : Popup::Error;
-  popupMsg = ok ? StrId::STR_DICT_NOT_FOUND : StrId::STR_DICT_ERROR;
+  // Name the failure: a genuine miss is "Not found"; a word that WAS found but
+  // couldn't be read is a real error — and we distinguish decompression from a
+  // low-memory allocation from a generic read error.
+  if (!ok) {
```
Uri-Tauber, line 180: in line 162, `ok` is used to store the result of `dict.buildIndex()`. 
So a low-memory failure during index building shows "Dictionary error", not "Not enough memory".

src/util/Dictionary.cpp
```diff
@@ -266,8 +266,12 @@ DictLocation Dictionary::locate(const char* target, std::string* matchedHeadword
...
 
-bool Dictionary::readDefinition(const DictLocation& location, std::string& out) {
-  if (!location.found) return false;
+bool Dictionary::readDefinition(const DictLocation& location, std::string& out, LookupResult* outResult) {
+  const auto fail = [outResult](LookupResult r) {
+    if (outResult) *outResult = r;
+    return false;
+  };
+  if (!location.found) return fail(LookupResult::ReadError);
   const uint32_t size = std::min(location.size, MAX_DEFINITION_BYTES);
```
Uri-Tauber, line 295: Since we know the size here, two cheap improvments:

```suggestion
  const uint32_t size = std::min(location.size, MAX_DEFINITION_BYTES);
  if (size == 0) {
    LOG_ERR("DICT", "Zero-length definition entry");
    return fail(LookupResult::ReadError);
  }

  // Refuse before touching the heap or the SD: std::string growth aborts on OOM
  // (-fno-exceptions), and the extraction below transiently holds a chunk buffer
  // plus a 32KB inflate window we'd rather not commit to a doomed lookup.
  if (ESP.getMaxAllocHeap() < size + 8 * 1024) {
    LOG_ERR("DICT", "Low heap for %lu byte definition", static_cast<unsigned long>(size));
    return fail(LookupResult::LowMemory);
  }
```

With this, the block at 316-321 become redundant.

src/util/Dictionary.cpp
```diff
@@ -266,8 +266,12 @@ DictLocation Dictionary::locate(const char* target, std::string* matchedHeadword
...
 }
 
-bool Dictionary::readDefinition(const DictLocation& location, std::string& out) {
-  if (!location.found) return false;
+bool Dictionary::readDefinition(const DictLocation& location, std::string& out, LookupResult* outResult) {
+  const auto fail = [outResult](LookupResult r) {
+    if (outResult) *outResult = r;
+    return false;
+  };
+  if (!location.found) return fail(LookupResult::ReadError);
```
Uri-Tauber:
```suggestion
  if (!location.found) return fail(LookupResult::NotFound);
```

W-Floyd:
@Uri-Tauber thanks! Pushed changes. See below for something the LLMs flagged, maybe worth [a separate PR](https://github.com/crosspoint-reader/crosspoint-reader/pull/2733)? Pure Claude below the fold...

***

`locate()` builds `basePath + ".idx"` and `basePath + ".qidx"` as `std::string` temporaries and opens **both** files on every call — and `lookup()` calls it once per stem variant, up to ~5 times for a word like "stories". That's roughly 10 SD opens and 10 string allocations per single lookup.

That feels relevant rather than incidental: this PR's whole premise is that fragmentation is what makes lookups start failing mid-session, and the lookup path is itself a fragmentation source. It also cuts against CLAUDE.md's "avoid repeated `new`/`delete` in loops" and the `std::string` hot-path rule. Hoisting the two handles and two path strings into a per-lookup scope would take it to two opens and two allocations regardless of variant count.

My instinct is that it doesn't belong in this PR — different concern, and this one is scoped to naming causes.

Uri-Tauber (approved):

PR #2710, opened 2026-07-25
fix: treat <br> in CJK flowing text as a paragraph separator

szetszho:
## Summary

Many CJK EPUBs (especially web-novel conversions) use `<br>` as the paragraph separator inside one flowing block, and `<br><br>` as a scene break. The parser previously treated every `<br>` as a scene-break separator that could inject a full line-height gap, which inflated vertical spacing between every paragraph and collapsed page capacity in these books.

This PR classifies each `<br>` by its text-flow context:

- `ChapterHtmlSlimParser` now tracks a text-flow context: `flowHasText` (words flushed since the current flow started) and `flowHasCjk` (any flushed word contained a CJK codepoint, detected via `utf8IsCjkCodepoint` and latched so the per-word scan short-circuits once found). Flow state resets when a block element opens or closes and is preserved across `<br>`s.
- The first `<br>` after flowed CJK text becomes a plain block boundary with top and bottom margins/padding stripped (new `BlockStyle::withoutTop()` helper alongside `withoutBottom()`), so the container's spacing isn't re-applied at every `<br>` paragraph.
- A consecutive `<br>` (the current block is still empty) takes the legacy scene-break path, so `<br><br>` blank-line scene breaks are preserved.
- Non-CJK text and standalone `<br>`s between blocks go through code identical to the legacy path.

## Behavior

| Scenario | Before | After |
| --- | --- | --- |
| `中文A<br/>中文B` in one block | Scene-break gap per `<br>`, container margins re-applied | Plain paragraph boundary, no gap, margins stripped |
| `中文A<br/><br/>中文B` | Gap between every line | One blank-line scene break (single/double `<br>` distinction kept) |
| `<p>English<br/>English</p>` | Legacy behavior | Unchanged |
| Standalone `<br/>` between `<p>` blocks | Scene-break gap | Unchanged |
| `<br/>` inside a table cell with CJK text | — | Uses the cell block's style (td/th blocks don't push the style stack) |

## Cache invalidation

`SECTION_FILE_VERSION` bumped 32 → 33 in `Section.cpp` so cached page layouts built with the old `<br>` handling rebuild on next open. The partial-build sentinel (`SECTION_FILE_PARTIAL_VERSION`) is derived from the main version, so full and partial section caches invalidate in lockstep.

## Testing / Review

The change was code-reviewed and the five scenarios above were traced through the final code, including the `startNewTextBlock` empty-block gap-injection path (`fromBrElement` is consumed only there, so tagging a block that receives text is inert). No host/native test environment exists in this repo and no embedded build was run as part of this change.


szetszho:
@Uri-Tauber fix for the issues found on 1.5.0 CJK testing,  #2548  broke the support for CJK content

Uri-Tauber:
@szetszho Can you rebase on current develop and fix conflicts?

lib/Epub/Epub/parsers/ChapterHtmlSlimParser.cpp
```diff
@@ -957,14 +926,35 @@ void XMLCALL ChapterHtmlSlimParser::startElement(void* userData, const XML_Char*
...
-      BlockStyle brStyle =
-          self->currentTextBlock ? self->currentTextBlock->getBlockStyle() : self->blockStyleStack.back();
-      brStyle.fromBrElement = true;
-      self->startNewTextBlock(brStyle);
+      // A <br> separates lines within a text flow rather than ending it: snapshot the
+      // flow context and restore it after the block switch (startNewTextBlock resets
+      // it), so consecutive <br>s classify the same way.
+      const bool savedFlowHasText = self->flowHasText;
+      const bool savedFlowHasCjk = self->flowHasCjk;
+      if (savedFlowHasText && savedFlowHasCjk && self->currentTextBlock && !self->currentTextBlock->isEmpty()) {
```
brianhuster: I don't understand why this `if` block need `savedFlowHasCjk` condition. Does `<br/>` as in `text1<br/>text2` really work differently in CJK languages than in English? I think no CJk conditions are needed in this PR

Uri-Tauber (approved):
LGTM

PR #2716, opened 2026-07-25
fix: ignore ambiguous EPUB guide text so books open at the right location

### LINKED ISSUE #2674: Some epubs opens directly to the last chapter
### Affected Version

1.4.1

### Describe the Bug

A few of my epubs just opens directly to the last chapter when I open them for the first time. I've tested this on both CrossPoint and CrossInk, and it happens on both.

### Steps to Reproduce

1. Open for example this
[epub.zip](https://github.com/user-attachments/files/30246937/epub.zip)
2. Notice how it opens on "About the author", the last chapter.

### Expected Behavior

Should open at the beginning, of course ;)

When this happened at first I didn't understand that it was at the end of the book, so I thought the whole file was almost empty, as I went just a few pages forward and then it said End of book. So this was very confusing to me, and I thought CrossPoint just couldn't read almost any of the book at all. :)

Thank you for your work!

```shell

```

uxjulia:
## Summary

  * Fix first-open EPUB positioning for malformed EPUB 2 guides that label multiple unrelated files as `type="text"`.
  * Fixes #2674 

  * Treat only an explicit `type="start"` guide reference as a custom reading start; otherwise open from the first spine item.
  * Invalidate existing `book.bin` metadata caches so affected books rebuild with the corrected behavior.
  * Update the documented cache-format version.

## Additional Context

* Some EPUBs use `type="text"` broadly, including for end matter such as “About the Author.” Using those entries as a reading-start signal could open the book at the wrong location.
* An explicit EPUB `start` reference still takes precedence when present.
* Cache invalidation happens automatically after updating; users do not need to delete files manually.
* This build was validated by the user who reported the original issue.

---

### AI Usage

### DIFF
--- a/docs/file-formats.md
+++ b/docs/file-formats.md
@@ -6,7 +6,7 @@ All POD fields are written in the ESP32 little-endian representation used by
 
 ## `book.bin`
 
-### Version 7
+### Version 10
 
 `book.bin` stores EPUB metadata plus lookup tables for spine and TOC entries.
 The current firmware writes this version from `BookMetadataCache`.
@@ -18,7 +18,7 @@ import std.mem;
 import std.string;
 import std.core;
 
-#define EXPECTED_VERSION 7
+#define EXPECTED_VERSION 10
 #define MAX_STRING_LENGTH 65535
 
 struct String {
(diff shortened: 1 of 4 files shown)


Uri-Tauber (approved):
Awesome. thanks for taking care of this.

PR #1723, opened 2026-04-20
fix: Erroneous navigation with long filenames in footnote links

CSCMe:
## Summary

* **What is the goal of this PR?** 
* Increase the allowed link length for footnotes
* **What changes are included?** 
* Un-magic-numbers parts of the footnote parser
* Increases max href length

## Additional Context

* Additional RAM usage was not noticable to me, worst case 16 * 128 * 2 = 4KB instead of previously 16 * 88 * 2 ~ 2.8KB, see #1031 
* Motivation: My book navigated me to the start of the footnote chapter, which required manual navigating for every footnote due to ungodly file name choices which pushed the required href length to 72, which caused truncating, which caused Crosspoint to not find the anchor (Sacred and Terrible Air, available via reddit)

---

### AI Usage

While CrossPoint doesn't have restrictions on AI tools in contributing, please be transparent about their usage as it 
helps set the right context for reviewers.

Did you use AI tools to help write this code? _**NO**_

PR #2722, opened 2026-07-25
fix(epub): bump FOOTNOTE_HREF_LEN 96→256 for long calibre paths

### LINKED ISSUE #2555: Footnote link jumps to the beginning of the notes chapter instead of the specific footnote (still on 1.4.1)
**Affected Version**
1.4.1 (Xteink X4)

**Describe the Bug**
When following a footnote reference in an EPUB, the reader opens the
beginning of the notes chapter (page 0) instead of jumping to the
specific footnote anchor.

This looks like the same symptom as #2336, which was fixed by #2382
and shipped in 1.4.0, but it still reproduces for me on 1.4.1.

What I already tried:
- Updated to 1.4.1
- Cleared the reading cache (Settings → Clear Reading Cache) and also
  deleted the `.crosspoint/` folder from the SD card so the anchor map
  was rebuilt from scratch
- Ran the book through the built-in EPUB Optimizer

None of these helped — the footnote link still lands on the first page
of the notes chapter.

**Steps to Reproduce**
1. Open the attached EPUB
2. Navigate to a page with a footnote reference
3. Select the footnote reference and confirm
4. The reader opens the start of the notes chapter, not the footnote

**Expected Behavior**
The reader should jump to the specific footnote anchor, as described
in the #2382 fix.

**Additional Context**
The book is a Russian-language EPUB (likely converted from FB2), where
all footnotes live in a single separate notes chapter. Sample EPUB
attached. Possibly a different anchor-loss scenario than the one
covered by #2382 / #2271.

[Sapolski_Biologiya-dobra-i-zla-Kak-nauka-obyasnyaet-nashi-postupki.534294.fb2.epub.zip](https://github.com/user-attachments/files/29790164/Sapolski_Biologiya-dobra-i-zla-Kak-nauka-obyasnyaet-nashi-postupki.534294.fb2.epub.zip)

fchaix:
Calibre-generated EPUBs with long filenames and URL-encoded characters can produce footnote hrefs exceeding 96 chars (observed: 150 chars on a Terry Pratchett EPUB). The truncated href loses the `#anchor` fragment and the tail of the filename, so `resolveHrefToSpineIndex` fails and the reader stays on the current page instead of jumping to the footnote.

This is the second time this ceiling is hit — the previous fix (#1723, 8154f88) bumped from 64 to 96. 256 provides enough headroom for the longest reasonable epub paths while remaining a fixed-size buffer (preferable to `std::string` on ESP32's constrained heap).

Fixes #2555 — same symptom (footnote link lands on page 0 / current page instead of the specific anchor), different root cause than #2382.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,3 +32,4 @@ lib/EpdFont/scripts/output/
 /sdkconfig.default
 /sdkconfig.defaults
 sdkconfig.sticky
+flake.*
--- a/lib/Epub/Epub/FootnoteEntry.h
+++ b/lib/Epub/Epub/FootnoteEntry.h
@@ -3,7 +3,10 @@
 #include <cstring>
 
 #define FOOTNOTE_NUMBER_LEN 32
-#define FOOTNOTE_HREF_LEN 96
+#define FOOTNOTE_HREF_LEN 256
+// ponytail: bumped from 96 to 256; calibre-generated EPUBs with long filenames
+// and URL-encoded characters routinely exceed 96 chars (e.g.
+// "Author-Title_split_NNN.html#_ftnN" encoded is ~150 chars).
 
 struct FootnoteEntry {
   char number[FOOTNOTE_NUMBER_LEN];


Uri-Tauber (approved):

lpla:
Follow-up: changing `FOOTNOTE_HREF_LEN` changed the serialized footnote record from 128 to 288 bytes, while the section-cache version remained unchanged. Existing caches could therefore be read using the new layout instead of being rebuilt. I opened #2935 to bump the format version and update the documentation; the existing version-mismatch path handles rebuilding automatically. This addresses the unresolved CodeRabbit finding: https://github.com/crosspoint-reader/crosspoint-reader/pull/2722#discussion_r3651173998

PR #2733, opened 2026-07-26
perf: open dictionary index files once per lookup

W-Floyd:
## Summary

* **Goal:** Stop the dictionary lookup path from churning the heap it depends on. `Dictionary::locate()` opened **both** `.idx` and `.qidx` and built both path strings as `std::string` temporaries on *every call* — and `lookup()` calls it once for the cleaned word and then once per stem variant, up to ~5 probes for a word like "stories". That's roughly **10 SD opens and 10 string allocations for a single lookup**.

  This is the supply side of the bug named in #2706: fragmentation is what makes dictionary lookups start failing mid-session (the ~32 KB contiguous inflate window for `.dict.dz` can no longer be allocated, and a reset "fixes" it). The lookup path was itself a fragmentation source. It also cuts against CLAUDE.md's *"avoid repeated `new`/`delete` in loops"* and the `std::string` hot-path rule.

* **Changes:** a private `LookupSession` holds the two `HalFile` handles plus the validated sidecar state, opened once by `openSession()` and passed into `locate()`.

  | | Before | After |
  |---|---|---|
  | SD opens per lookup | ~10 | **2** |
  | `std::string` path temporaries | ~10 | **2** |
  | `.qidx` header validations | once per probe | **once per lookup** |
  | Peak open file handles | 2 | 2 (unchanged) |

  Counts are per lookup regardless of how many stem variants get probed.

## Scope Check

Please confirm:

- [x] No new feature surface at all — this is a pure internal refactor of an existing in-scope feature (dictionary lookup). Behaviour is unchanged.

## Additional Context

* **Why sharing the handles is behaviour-preserving:** every read in `locate()` seeks absolutely before reading — `readSampleOffset()` seeks to its computed offset, the bisect seeks per probe, and the linear scan seeks to `startByte` before the loop. No call inherits file position from a previous one, so reuse changes nothing observable.
* **Peak handles deliberately unchanged:** the session is scoped in a block so `.idx`/`.qidx` close before `readDefinition()` opens the data file. Without that scoping the peak would have gone 2 → 4.
* **Memory / flash impact:** no new persistent RAM. `LookupSession` is a stack struct in `lookup()` holding two `HalFile` handles that previously lived on `locate()`'s stack anyway. The win is fewer transient heap allocations and fewer SD transactions per lookup, not a static saving.
* **Relationship to #2706:** that PR touches the same `locate()` — it adds failure-cause plumbing (`readError`) so an `.idx` failure isn't reported as "Not found". This PR changes where the handles come from. **They will conflict.** Deliberately opened as a draft: I'd suggest landing #2706 first, then I'll rebase this on top. Raising it now so the approach can be sanity-checked before I do that work.

### Testing

* Builds clean for the device `default` env (ESP32-C3).
* `pio check` clean, clang-format clean, 129/129 unit tests pass.
* **Not yet tested on hardware.** No dictionary lookup has been exercised on a device with this change. That verification is still outstanding and I'll do it before marking ready for review.

---

### AI Usage

Did you use AI tools to help write this code? _**YES**_ — the churn was identified and the refactor implemented with Claude Code. Not yet hardware-verified by me (see Testing).

### DIFF
--- a/src/util/Dictionary.cpp
+++ b/src/util/Dictionary.cpp
@@ -6,6 +6,7 @@
 
 #include <algorithm>
 #include <cctype>
+#include <cstdio>
 #include <cstring>
 
 #include "DictZip.h"
@@ -96,8 +97,23 @@ bool Dictionary::open(const char* folderName) {
     LOG_ERR("DICT", "%s uses 64-bit index offsets (unsupported)", resolved.c_str());
     return false;
   }
+  // Checked once here so buildPath() can never fail on the lookup path.
+  if (resolved.size() + strlen(LONGEST_SUFFIX) + 1 > PATH_BUF_BYTES) {
+    LOG_ERR("DICT", "Dictionary path too long (%u chars, max %u)", static_cast<unsigned>(resolved.size()),
+            static_cast<unsigned>(PATH_BUF_BYTES - strlen(LONGEST_SUFFIX) - 1));
+    return false;
+  }
 
   basePath = std::move(resolved);
+  return true;
+}
+
+bool Dictionary::buildPath(char* buf, size_t bufSize, const char* suffix) const {
+  const int n = snprintf(buf, bufSize, "%s%s", basePath.c_str(), suffix);
+  if (n < 0 || static_cast<size_t>(n) >= bufSize) {
+    LOG_ERR("DICT", "Path too long: %s%s", basePath.c_str(), suffix);
+    return false;
+  }
   return true;
 }
 
@@ -221,12 +237,17 @@ int Dictionary::readWordInto(HalFile& file, char* buf, size_t bufSize) {
 
 bool Dictionary::openSession(LookupSession& session) {
   if (!isOpen()) return false;
-  if (!Storage.openFileForRead("DICT", basePath + ".idx", session.idx)) return false;
+
+  // One buffer reused for both paths — no transient heap on the lookup path.
+  char path[PATH_BUF_BYTES];
+  if (!buildPath(path, sizeof(path), ".idx")) return false;
+  if (!Storage.openFileForRead("DICT", path, session.idx)) return false;
   session.idxSize = static_cast<uint32_t>(session.idx.fileSize());
 
   // The sidecar is optional and its header only has to be validated once per
   // lookup; without a usable one locate() scans from byte 0.
-  if (Storage.openFileForRead("DICT", basePath + ".qidx", session.qidx)) {
+  if (!buildPath(path, sizeof(path), ".qidx")) return true;
+  if (Storage.openFileForRead("DICT", path, session.qidx)) {
     const QidxHeader header = readQidxHeader(session.qidx, SAMPLE_INTERVAL);
     if (header.valid && header.idxFileSize == session.idxSize) session.sampleCount = header.sampleCount;
   }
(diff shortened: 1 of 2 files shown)


W-Floyd:
@coderabbitai ready for review

Uri-Tauber:
@W-Floyd Have you tested on your hardware?

A user reported a dictionary which cause a crash was successfully opened with this optimization.

W-Floyd:
I may have briefly tried it - getting a dictionary to crash reliably is hard 😅 I'm flashing now, I'll give it a whirl again.

Uri-Tauber:
Try with focus reading + book style + AA on.

W-Floyd:
Ah, cool, that ran out of memory first try once I changed those settings lol
"Not enough memory"

W-Floyd:
I suspect it's the decompression/inflate causing the failure - I'll test later with an uncompressed `.dict`. I'd tried an experiment around this doing on-device decompression ahead of first use, which worked but was really slow - better to just load the decompressed file from the start (maybe WebUI could offer to decompress?)

Uri-Tauber:
@W-Floyd Is this PR ready for review, or do you want to optimize more?

W-Floyd:
@Uri-Tauber probably ready

src/util/Dictionary.cpp
```diff
@@ -96,11 +97,26 @@ bool Dictionary::open(const char* folderName) {
...
+bool Dictionary::buildPath(char* buf, size_t bufSize, const char* suffix) const {
+  const int n = snprintf(buf, bufSize, "%s%s", basePath.c_str(), suffix);
+  if (n < 0 || static_cast<size_t>(n) >= bufSize) {
+    LOG_ERR("DICT", "Path too long: %s%s", basePath.c_str(), suffix);
+    return false;
+  }
+  return true;
+}
+
 bool Dictionary::needsIndex() {
```
Uri-Tauber, line 120: `needsIndex()`'s only caller is `DictionaryWordSelectActivity.cpp:159`, which sits outside the `dictOpenAttempted` guard and runs on every tapped word. It opens basePath + ".idx", sizes it, opens basePath + ".qidx" and re-reads the same QidxHeader — then `openSession()` immediately opens the same two files and re-reads the same header.

hoist it behind `dictOpenAttempted`, or fold the sidecar-freshness check into `openSession()`.

src/util/Dictionary.cpp
```diff
@@ -96,11 +97,26 @@ bool Dictionary::open(const char* folderName) {
     LOG_ERR("DICT", "%s uses 64-bit index offsets (unsupported)", resolved.c_str());
     return false;
   }
+  // Checked once here so buildPath() can never fail on the lookup path.
+  if (resolved.size() + strlen(LONGEST_SUFFIX) + 1 > PATH_BUF_BYTES) {
```
Uri-Tauber, line 101: This is the one new failure mode in the branch. 
`DictionaryRegistry::findStem()` reads both components through `char[128]` buffers (`DictionaryRegistry.cpp:26-27`), so a base path of ~266 chars is legal by the registry's own limits — roughly double the 151 allowed here. A dictionary that exceeds it worked before (`std::string` had no ceiling, and SdFat walks path components without a fixed limit); now `open()` returns false and every lookup shows the generic `STR_DICT_ERROR`.

Uri-Tauber:
Ok. I'm going to fix those myself in small followup commit so we can merge now.

Uri-Tauber (approved):

W-Floyd:
Thanks!

PR #2665, opened 2026-07-21
feat(epub): native <ruby> support for Chinese, Japanese

brianhuster:
## Summary

**What is the goal of this PR?**: Support `<ruby>` and `<rt>` tags, as *required* by EPUB 3 spec.
**What changes are included?**
- XHTML Parser: Added support for parsing `<ruby>` and `<rt>` tags, and skipping fallback `<rp>` tags (which are discouraged by EPUB 3 standard). Pre-reserves heap capacity upon entering ruby tags to minimize fragmentation. Trims and normalizes all annotation text by collapsing redundant whitespace/newlines.
- Text Layout & Rendering: Implemented robust Group Ruby, Mono-ruby, and complex multi-rt tabular layouts. Words in a Group Ruby block (e.g. `<ruby>明日<rt>あした</rt></ruby>`) are kept together to prevent incorrect line-breaking.
- Dictionary: shift the y coordinate of the word selection popup down an amount of `rubyShift` if it is on a line that has ruby characters

## Scope Check

CrossPoint is intentionally narrow. See [SCOPE.md](../blob/master/SCOPE.md) and [ROADMAP.md](../blob/master/ROADMAP.md).
Please confirm:

- [x] I have read SCOPE.md and ROADMAP.md.
- [x] This PR is **not** a new built-in theme (themes are temporarily closed pending the move to SD-loaded themes).
- [x] This PR is **not** a new external network connector (sync engine, cloud storage, remote file access, etc.).
- [x] This PR is **not** an interactive app, writing tool, RSS/news/browser, media playback, or PDF feature.
- [] The stock firmware does not already handle this well, **and** no other popular CrossPoint fork already does
      (or, if one does, I explain why CrossPoint still needs it below).
- [x] If this PR touches `freeink-sdk/`, `lib/hal/`, the bootloader, OTA, or recovery code, I have coordinated with
      the relevant maintainer.

**Why Crosspoint still needs it**: 

This feature is a part of the [core reading experience](https://github.com/crosspoint-reader/crosspoint-reader/blob/973a90233232f03bf1a697f4375e931cf84f6dab/SCOPE.md?plain=1#L43-L45). [EPUB 3 spec](https://www.w3.org/TR/epub-33/#sec-xhtml-deviations-rp) also requires readers to be "ruby aware"

Without ruby support, the following code block
```
<ruby>
  漢<rt>かん</rt>
  字<rt>じ</rt>
</ruby>は
<ruby>
  難<rt>むずか</rt>
[...]

PR #2547, opened 2026-07-06
fix: Flatten TextBlock word storage into single allocation

itsthisjustin:
Replace per-word string vectors with a single NUL-terminated text blob and offset arrays. This allows firmware to read an entire TextBlock with one SD read and one allocation. Add validation checks for allocation failures during TextBlock construction and PageLine deserialization.

## Why

A rendered page is held in RAM as ~25-30 `TextBlock`s (one per line), and each block stored its words as `std::vector<std::string>` plus five parallel vectors. On a ~380 KB device with no heap compaction, this causes crashes and heap fragmentation as you turn pages.

## What changed

TextBlock now stores its entire payload in **one heap allocation**: a flat arena holding the per-word metadata arrays (text offsets, x positions, styles, and the focus-reading annotations when present) followed by all word text, each word NUL-terminated in place.

## Measured results (X3, Bitter 14 SD font, 66-page chapter)

| Metric | Before | After |
|---|---|---|
| Allocations per page load (word storage) | ~250 | ~30 |
| maxAlloc / free ratio during reading | 35-91%, eroding over the session | 86-94%, sustained |
| Worst mid-session minimum free heap | under 1 KB | 55.9 KB |
| Resident word-storage overhead per page | ~10-12 KB | ~4-5 KB |

Fragmentation no longer accumulates from page turns, so long sessions keep the same large-allocation headroom as fresh boots

PR #2735, opened 2026-07-27
perf: skip ruby scratch allocations on ruby-less lines

W-Floyd:
## Summary

* **What is the goal of this PR?** Remove the per-line heap churn that `<ruby>` support reintroduced into the render hot path, and give `TextBlock` its first host test coverage.
  - `TextBlock::render()` no longer allocates on ruby-less lines (**2 → 0 allocations per line, 56 → 0 per page render**).
  - `Page::deserialize()` reserves the element vector before its `push_back` loop.
  - New host tests: 12 behavioural + 3 allocation-accounting.

## Background

[#2547](https://github.com/crosspoint-reader/crosspoint-reader/pull/2547) flattened `TextBlock` word storage into a single arena, taking word-storage allocations per page load from ~250 to ~30. [#2665](https://github.com/crosspoint-reader/crosspoint-reader/pull/2665) (merged 19 days later) added `<ruby>` support, which reintroduced per-word storage into the same path.

CodeRabbit flagged both instances in review on #2665 four days before merge, tagged *🔵 Trivial / Quick win*, and neither was addressed. This is that follow-up.

## What was happening

`TextBlock::render()` allocated two per-word vectors *before* the `hasRuby()` gate that decides whether either is ever read:

```cpp
std::vector<int> wordShiftArr(numWords, 0);
std::vector<RubyDrawInfo> rubies(numWords);   // RubyDrawInfo holds a std::string
if (hasRuby()) {
```

Every line of every non-CJK book paid for both, on **every render** — not just page load. These are transient: taken and freed per line, so on a device with no heap compaction they churn DRAM rather than occupying it.

`getRubyShift()` was also called once per word, and it rescans `rubyTexts` each time, making that part of the loop O(numWords²). The value is loop-invariant.

## Measured

Host-side allocation counting (`TextBlockAllocTest`), 12-word ruby-less line:

| | before | after |
|---|---|---|
| allocations per line render | 2 | **0** |
| allocations per 28-line page render | 56 | **0** |

Reproducible by anyone via `ctest` — no hardware needed. `RubyRenderStillAllocatesAndDraws` passes on both sides, pinning that the ruby path was made conditional rather than disabled.

## Behaviour is unchanged

- `wordShiftArr` was all-zero whenever `hasRuby()` was false, so `drawX` is identical.
- `rubyShift` is the same expression `getRubyShift()` would return, hoisted out of the loop.
- The ruby draw site was already unreachable without a non-empty `rubyTexts[i]`, so the added `blockHasRuby &&` is a short-circuit no-op that only makes the vector access safe.

## Tests

`TextBlock` had **no host coverage**, so ruby layout (group ruby, adjacent-annotation collision resolution, the baseline shift) was only checkable on device. Follows the `minibidi_arabic` pattern: a `stubs/` dir first on the include path shadows the four headers that can't build on the host (`Logging.h`, `HalStorage.h`, `EpdFontFamily.h`, `GfxRenderer.h`); the real `TextBlock.cpp` is compiled against them.

The 12 behavioural tests were **written and verified green on `develop` first**, then mutation-tested to confirm they can fail:

| Mutation on develop | Caught by |
|---|---|
| `getRubyShift()` → always `0` | `RubyLineShiftsBaselineAndDrawsAnnotation` |
| drop per-word `wordShiftArr[i]` offset | `WideRubyRecentresBaseWord` |

Those are exactly the two axes this change touches.

## Scope Check

- [x] The stock firmware does not already handle this well — this is a regression against #2547's measured baseline.

## Additional Context

**Memory / flash impact.** No change to static footprint — DRAM stays at 119,013 bytes (`.text` 66,988 / `.bss` 35,416 / `.data` 16,609). The win is heap churn, not residency.

**The `Page::deserialize` reserve is clamped**, deliberately. `count` is read straight off the SD cache and is `uint16_t`, so an unclamped `reserve(count)` would let a corrupt header request 65535 × `sizeof(shared_ptr)` ≈ 512 KB — and since `std::vector`'s `operator new` is throwing under `-fno-exceptions`, that is an `abort()`, converting a corrupt-cache miss into a hard crash. Capped at 256; under-reserving just falls back to normal growth. Same pattern as #2435 and #2519.

**Not hardware-verified.** The allocation counts are exact and host-measured, but reduced fragmentation is an inference from them — glibc's allocator behaves nothing like ESP-IDF's. Confirming the actual DRAM benefit wants a device run in #2547's terms (`maxAlloc / free` ratio sampled during steady-state reading, `ESP.getMinFreeHeap()` across ~100 page turns, same book, `.crosspoint/` cleared between runs). **Left as draft pending that.**

**One pre-existing warning surfaced.** The test target's `-Wall -Wextra -pedantic` exposes a sign-compare at `TextBlock.cpp:415` (`file.read(...)` returns `int`, compared against `size_t`). Benign — a `-1` error return converts to a huge unsigned and still takes the error path — and pre-existing, so production code was left alone. Happy to silence it with a cast if reviewers prefer a quiet CI log.

**Deliberately not included:** the third finding from the same audit — `Page::elements` is `vector<shared_ptr<PageElement>>` and `PageLine` holds `shared_ptr<TextBlock>`, both constructed *from* `unique_ptr`, so each pays a separately allocated control block plus atomic refcounting on a single-core RISC-V. Ownership on the read path is unique. Left out because 10 open PRs touch `Page.cpp`/`Page.h`/`TextBlock.*` and the conflict surface isn't worth it right now.

**Related open work:** [#2554](https://github.com/crosspoint-reader/crosspoint-reader/pull/2554) (shared bidi scratch buffer, ~1.5 KB static) and [#2437](https://github.com/crosspoint-reader/crosspoint-reader/pull/2437) (cache section LUT offset) are both complementary and non-conflicting.

---

### AI Usage

Did you use AI tools to help write this code? _**YES**_ — written with Claude Code. Build (`pio run -e default`) and `pio check` clean, host tests green, mutation-tested for sensitivity. Not yet flash-tested; see the draft note above.

### DIFF
--- a/lib/Epub/Epub/Page.cpp
+++ b/lib/Epub/Epub/Page.cpp
@@ -174,6 +174,16 @@ std::unique_ptr<Page> Page::deserialize(HalFile& file) {
   uint16_t count;
   serialization::readPod(file, count);
 
+  // Reserve up front so a page load costs one allocation for the element vector
+  // instead of a grow-copy-free cycle every doubling. `count` is untrusted (it
+  // comes straight off the SD cache), so clamp it: a real page holds a few dozen
+  // elements, while a corrupt header could ask for 65535 * sizeof(shared_ptr) and
+  // abort() on the failed allocation (vector's operator new is throwing, and this
+  // firmware builds with -fno-exceptions). Under-reserving is harmless -- the
+  // push_back path below still grows normally.
+  static constexpr uint16_t RESERVE_CAP = 256;
+  page->elements.reserve(std::min(count, RESERVE_CAP));
+
   for (uint16_t i = 0; i < count; i++) {
     uint8_t tag;
     serialization::readPod(file, tag);
--- a/lib/Epub/Epub/blocks/TextBlock.cpp
+++ b/lib/Epub/Epub/blocks/TextBlock.cpp
@@ -142,9 +142,16 @@ void TextBlock::render(const GfxRenderer& renderer, const int fontId, const int 
     std::string text;
     BidiUtils::BidiBaseDir baseDir;
   };
-  std::vector<int> wordShiftArr(numWords, 0);
-  std::vector<RubyDrawInfo> rubies(numWords);
-  if (hasRuby()) {
+  // hasRuby() is an O(numWords) scan, so resolve it once here rather than per word.
+  // Both arrays below are only ever read when the line carries ruby, so they stay
+  // empty (zero allocations) for the ruby-less case, which is every line of a
+  // non-CJK book. Sized lazily inside the branch.
+  const bool blockHasRuby = hasRuby();
+  std::vector<int> wordShiftArr;
+  std::vector<RubyDrawInfo> rubies;
+  if (blockHasRuby) {
+    wordShiftArr.assign(numWords, 0);
+    rubies.resize(numWords);
     int accumulatedShift = 0;
     int lastEnd = -9999;
     for (uint16_t i = 0; i < numWords; i++) {
@@ -245,6 +252,10 @@ void TextBlock::render(const GfxRenderer& renderer, const int fontId, const int 
     }
   };
 
+  // Loop-invariant: hoisted out of the word loop so rubyTexts is scanned once,
+  // not once per word.
+  const int rubyShift = getRubyShift(ascender);
+
   for (uint16_t i = 0; i < numWords; i++) {
     const char* word = wordText(i);
     const int wordX = xposArr[i] + x;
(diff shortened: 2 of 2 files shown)


Uri-Tauber:
The fixes are sound, but do we really need all those tests?

W-Floyd:
@Uri-Tauber, fair, I've reshaped this branch to be just the fix

I'll move tests to a separate PR rather than being dropped. `TextBlock` has no host coverage at all today, and I'd rather that be judged on its own merits than ride along with a perf change. I'll open that shortly.

Two tests will fail until (if) this is merged: `RubylessRenderAllocatesNothing` and `RubylessPageRenderAllocatesNothing` assert *zero* allocations, so by design they fail on `develop` and only pass with this change.

The reduced-fragmentation claim is still an inference, but casual testing has been fine on ruby-less reading of my own.

Uri-Tauber (approved):
LGTM

PR #2747, opened 2026-07-27
fix: clean image page after sync return

thiagokokada:
## Summary

* **What is the goal of this PR?** Prevent stale Sync Progress UI content from remaining visible when returning to an EPUB page containing an image. The issue can be seen in the image below:

PXL_20260727_202809106 MP

* **What changes are included?** Image pages now perform the pending clean HALF refresh before their double-FAST image-rendering sequence.

## Scope Check

Please confirm:

- [x] The stock firmware does not already handle this issue well, **and** no other popular CrossPoint fork already does (or, if one does, I explain why CrossPoint still needs it below).

## Additional Context

The root cause is that image pages bypass the normal refresh cadence. After Sync Progress exits via a silent restart, the e-ink panel retains the Sync Progress frame; the image page's double-FAST pipeline started from that stale frame, which could leave the UI and book image visually mixed. The clean base refresh synchronizes the panel before the image pipeline runs.

Memory / flash impact: negligible; one existing refresh condition is extended.

Validation: tested on Xteink X4. This is a bug for the release 1.5.0, so it should probably be backported.

---

### AI Usage

### DIFF
--- a/src/activities/reader/EpubReaderActivity.cpp
+++ b/src/activities/reader/EpubReaderActivity.cpp
@@ -1480,6 +1480,12 @@ void EpubReaderActivity::renderContents(std::unique_ptr<Page> page, const int or
   const bool pageHasImagesNeedingDecode = pageHasImages && page->hasImagesNeedingDecode();
   const bool manualRefreshPending = forcedRefreshPending;
   forcedRefreshPending = false;
+  // The reader starts with zero here, which means the normal refresh cycle
+  // would use a HALF refresh for its first page. Keep that same clean base for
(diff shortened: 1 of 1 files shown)


thiagokokada:
CC @Uri-Tauber, this is a bug affecting the soon to be released 1.5.0 release, and the worst thing from my experience is that it is difficult to fix once triggered (unless you have full refresh mapped to the power button you either need to move enough pages until the refresh triggers manually, close and open a book or power cycle the device).

Uri-Tauber (approved):
Nice catch, and clean fix.

PR #2755, opened 2026-07-27
fix: batch-load SD fallback glyphs for UI text to fix Hangul slowdown

### LINKED ISSUE #2725: File Browser becomes very slow when rendering Hangul filenames
### Affected Version

v1.5.0 rc1

### Describe the Bug

The File Browser becomes significantly slower when rendering Hangul (Korean) filenames.

Initially I suspected the issue was caused by the size of my custom UI font, since it originally included the full CJK Unicode ranges. However, after rebuilding the font with only the following ranges, the performance remained exactly the same:

(0x0020-0x007E),(0x00A0-0x00FF),(0x2000-0x206F),(0x2460-0x24FF),(0x25A0-0x25FF),(0x3000-0x303F),(0x3130-0x318F),(0xAC00-0xD7A3),(0xFF00-0xFFEF)

The slowdown only occurs when Hangul characters are actually rendered. If the same files are renamed to ASCII-only filenames, the File Browser becomes responsive again.

This appears to be different from the recently reported CJK TXT reader performance issue, as it occurs only in the File Browser while rendering filenames.

### Steps to Reproduce

1. Install a custom UI font (.cpfont) with Hangul support.
2. Copy several files with Hangul filenames to the SD card.
3. Open the folder in the File Browser.
4. Navigate through the file list.

For comparison:

1. Rename the same files to ASCII-only filenames.
2. Open the same folder again.

The File Browser becomes noticeably faster.

### Expected Behavior

Rendering Hangul filenames should have performance comparable to rendering ASCII filenames.

Reducing the font to only Hangul (without CJK ideographs) does not improve the performance, suggesting the slowdown is related to the Hangul rendering path rather than the total number of glyphs in the font.

```shell
No crash or error messages.
```

szetszho:
Fixes #2725

## Root cause

When a UI string (filename, book title) contains CJK/Hangul, `resolveTextFontId()`
redirects the whole string to the size-matched SD-card fallback font. SD-card
fonts only keep glyphs resident after a prewarm — but prewarm only happens
inside the reader/dictionary activities' `PrewarmScope`. The File Browser, Home
screen, and Recent Books never prewarm, so every Hangul glyph faulted through
`SdCardFont::onGlyphMiss()`:

- a full `.cpfont` **file open + 2 seeks + 2 reads + close per glyph**,
- cached in an **8-slot overflow ring**, which any screenful of Korean text
  (far more than 8 unique syllables) thrashes completely,
- paid **per measure and per draw** — `getTextBounds()` walks `getGlyph()` per
  codepoint, and `truncatedText()`'s shrink loop re-measures repeatedly.

One File Browser redraw with Hangul filenames meant hundreds of SD file opens.
This matches every reported symptom: slowdown scales with the amount of Hangul
rendered on screen, ASCII rows are unaffected, and shrinking the font (fewer
glyphs) doesn't help.

## Fix

New private helper `GfxRenderer::ensureSdGlyphsResident()`: when (and only
when) `resolveTextFontId()` redirected a string to the SD fallback
(`resolvedFontId != fontId`), batch-load the string's glyphs into the SD
font's resident mini tables via the existing `SdCardFont::prewarm()` machinery
before the per-glyph loop runs. Called from `drawText()`, `getTextWidth()`,
and `drawTextRotated90CW()`.

- **One file open per new string** instead of one per glyph per redraw.
- Repeat draws / truncation re-measures hit `prewarmStyle()`'s RAM-only
  subset check — zero SD I/O.
- The redirect gate keeps the reader's hot path (primary SD reading font,
  already scope-prewarmed) completely untouched, and built-in fonts are a
  map-lookup no-op.
- Prewarm during a `FontCacheManager` scan pass is naturally skipped in
  `drawText()` (the scan early-returns before the new call).

## Why not the patches drafted in the issue

- *Hoisting `resolveTextFontId()` out of `truncatedText()`'s loop*: real but
  minor — the resolver is RAM-only (coverage index, no SD I/O) and was never
  the bottleneck; unnecessary once glyph loading is batched.
- *Replacing the `utf8IsCjkCodepoint()` block-list with `cp > 0x7F`*: doesn't
  address the bottleneck, and silently changes fallback routing for non-CJK
  scripts (e.g. Greek/Cyrillic strings dragged into the SD font). The
  block-list is a deliberate scope decision; left as is.

## Testing

- Full `pio run -e default` build passes (ESP32-C3 image).
- Not yet verified on hardware — the reporter's scenarios are the acceptance
  tests: a File Browser folder of Hangul filenames, and the Home screen with
  multiple Hangul Recent Books titles (Lyra Extended), both compared against
  ASCII responsiveness.

## Scope

2 files, +40 lines: `lib/GfxRenderer/GfxRenderer.h`, `lib/GfxRenderer/GfxRenderer.cpp`.
No serialized-format or section-cache changes (rendering only, no layout change),
so no `SECTION_FILE_VERSION` bump is needed.


lib/GfxRenderer/GfxRenderer.cpp
```diff
@@ -209,6 +209,17 @@ int GfxRenderer::resolveTextFontId(const int fontId, const char* text, const Epd
...
 
+void GfxRenderer::ensureSdGlyphsResident(const int fontId, const char* text, const EpdFontFamily::Style style) const {
+  const auto sdIt = sdCardFonts_.find(fontId);
+  if (sdIt == sdCardFonts_.end()) {
+    return;
+  }
+  // SUP/SUB bits don't select a distinct .cpfont style bitstream — mask to the
+  // base style. resolveStyleMask() inside prewarm folds absent styles.
+  const uint8_t styleMask = static_cast<uint8_t>(1u << (static_cast<uint8_t>(style) & 0x03));
+  sdIt->second->prewarm(text, styleMask);
```
Uri-Tauber, line 220: `ensureSdGlyphsResident` runs on every redirected `drawText`/`getTextWidth`, but `SdCardFont::prewarm` unconditionally does new `uint32_t[512]` (2 KB) + an O(n²) dedup before its resident subset-check can short-circuit. For the UI screens this PR targets (a few title strings per redraw) that's fine. But the same call is reachable per-word from the reader render loop when a built-in-font book has CJK redirected to the SD fallback — hundreds of 2 KB `malloc`/`free` per page.
Could prewarm early-out on a cheap residency check before allocating (or reuse a member scratch buffer), so the resident-page case is truly RAM-only?

lib/GfxRenderer/GfxRenderer.cpp
```diff
@@ -578,6 +596,12 @@ void GfxRenderer::drawText(const int fontId, const int x, const int y, const cha
     return;
   }
 
+  // Redirected to the SD fallback: batch-load the string's glyphs so the draw
+  // loop below doesn't fault them in one SD read at a time (#2725).
+  if (resolvedFontId != fontId) {
+    ensureSdGlyphsResident(resolvedFontId, renderedText, style);
+  }
```
Uri-Tauber, line 603: The batch-load only fires when `resolveTextFontId` redirected to a fallback. A directly-selected SD-card font has `resolvedFontId == fontId`, so it skips this and still faults one glyph at a time on any screen without a PrewarmScope — the exact #2725 symptom. `ensureSdGlyphsResident` already no-ops for built-in fonts (`sdCardFonts_.find miss`), so the guard can just be dropped to cover both cases.

```suggestion
  ensureSdGlyphsResident(resolvedFontId, renderedText, style);
```

Same guard removal applies at the other two sites: getTextWidth (lines 555–560, uses renderedText) and drawTextRotated90CW (lines 2013–2017, uses text).

Uri-Tauber (changes requested):

Uri-Tauber:
@szetszho Are you planning to come back to this?

RayWarthog:
I've tested a version of the firmware which is the latest develop branch + the changes from this PR manually applied on both my Xteink X3 and X4, with multiple files/books that has Traditional Chinese as filenames, in ToC and in metadata, and it solved the slowness in the Home screen, the Menu screen and the file browser, that was basically unusably slow in 1.5.0. I've faced no issues so far.

Sharing it here because I'm hoping this PR wouldn't just close and die out cause the fixes did solve my issue.

szetszho:
Sorry, been busy at life, will be back on this tomorrow
Best Regards,
Max Sze

szetszho:
All good 
@Uri-Tauber

itsthisjustin:
> All good @Uri-Tauber

So I don't know how but you somehow made this pr show 12,000 lines of code changes haha. I will probably cherry pick your fix into a clean PR and credit you

itsthisjustin:
Closing in favor of #3026

PR #2756, opened 2026-07-27
feat: add opt-in automatic KOReader progress checks

rigogsilva:
## Summary

  Add an opt-in, pull-only automatic progress check to the existing KOReader Sync integration so readers can be offered newer remote progress when they open an EPUB or wake with one already open.

  - Add **Automatic progress check** to the device and web KOReader Sync settings. It defaults to **Off**, including on upgrades.
  - Trigger automatic checks on explicit EPUB open and normal reader wake without retriggering during the internal return-to-reader path.
  - Reuse the existing KOReader Sync client, credentials, matching modes, and progress mapper.
  - Try only the last connected saved Wi-Fi network, without scanning or showing the network picker, using a hardware-tested six-second association timeout.
  - Keep connection failures, server failures, missing progress, and no-op results silent.
  - Show **Resume** / **No** only when the mapped remote spine/page is ahead of a stable, render-locked local snapshot.
  - Keep automatic mode pull-only. It never uploads and does not replace manual **Sync Progress**.
  - Disconnect and stop the Wi-Fi radio when the check finishes or while its confirmation prompt is shown.
  - Add automatic-decision and mapped-position policy tests and document the behavior in the existing KOReader Sync user-guide section.

## Scope Check

Please confirm:

      the relevant maintainer.

This is an in-reader progress-tracking UX improvement and does not add a connector or server protocol. CrossPoint already ships KOReader Sync; this PR only adds an optional trigger and pull-only decision policy around that existing integration. The Xteink stock firmware does not provide KOReader-compatible cross-device progress syncing. Kobo Nickel has comparable automatic-resume UX, but it is neither Xteink stock firmware nor a CrossPoint fork.

I also checked the ten most-starred direct CrossPoint forks on 2026-07-27. `franssjz/cpr-vcodex` does have automatic KOReader pull-on-open (and push-on-close), so it is the relevant existing fork implementation. CrossPoint still benefits from this PR's narrower behavior: it never uploads automatically, compares remote progress with the current local position, prompts only when remote is meaningfully ahead, and supports waking with an EPUB already open. The other inspected popular forks did not expose an equivalent automatic KOReader progress workflow.

The PR does not touch `freeink-sdk/`, `lib/hal/`, the bootloader, OTA, or recovery code.

### Behavior and risk controls

- EPUB only.
- Disabled by default and persisted in the existing KOReader credential store.
- Manual **Ask every time** and **Smart sync** behavior remains unchanged.
- No upload path is reachable from automatic mode.
- Automatic network attempts do not scan, show the Wi-Fi picker, or run NTP.
- The existing e-ink page remains displayed during network work; returning to the reader can re-render that page.
- Invalid remote percentages are ignored only in automatic mode.
- The reader remains the sleep context while the confirmation prompt is visible.

### Validation

- `ctest --test-dir build/test --output-on-failure` — 135/135 passed.
- `pio run -e default` — successful X3/X4 firmware build.
- Apple clang-format 21 and `git diff --check`.
- Final build totals: 52,020 bytes static RAM (15.9%) and 5,511,229 bytes application flash (84.1%).
- Hardware-tested on an Xteink X4 upgraded from CrossPoint 1.4.1:
  - automatic check triggered on EPUB open;
  - saved 2.4 GHz Wi-Fi connected within the six-second window;
  - TLS 1.3 handshake and KOReader Sync progress request succeeded with HTTP 200;
  - remote-ahead progress produced the **Resume** / **No** confirmation prompt;
  - declining or completing the check returned to the same book.

The timeout was initially three seconds. Hardware logs showed the ESP32-C3 consistently reaching the cutoff just before associating, so it was raised to six seconds and successfully retested on the same X4.

### Review focus

- Reader lifecycle behavior across explicit open, normal wake, and the internal return-to-reader path.
- Wi-Fi ownership and shutdown on success, failure, and while the prompt is shown.
- Guarantee that automatic mode cannot upload.
- Stable local-position capture, mapped spine/page comparison, and invalid-remote handling.

---

### AI Usage

Codex assisted with repository investigation, implementation, tests, documentation, code review, build/flash verification, and preparation of this PR. The behavior was validated interactively on the contributor's physical Xteink X4, including review of serial logs from the real Wi-Fi and KOReader Sync flow.


rigogsilva:
@coderabbitai review

PR #2783, opened 2026-07-29
feat: adding check_translations.py script to help discover untranslated i18n strings

dasrecht:
## Summary
Adding a script that makes it easier to discover untranslated i18n strings

It checks against the English translation strings by default and lists out missing, untranslated and extra strings.

## Scope Check

Please confirm:

## Additional Context

  specific areas to focus on).
* Memory / flash impact, if known.

---

### AI Usage

Did you use AI tools to help write this code? _**< YES  >**_

### DIFF
--- a/scripts/check_translation.py
+++ b/scripts/check_translation.py
@@ -1,0 +1,102 @@
+#!/usr/bin/env python3
+"""
+Compare a language's translation YAML against english.yaml and report
+missing / extra / untranslated (same as English) keys.
+
+Usage:
+    python scripts/check_translation.py [language] [translations_dir]
+
+Examples:
+    python scripts/check_translation.py german
+    python scripts/check_translation.py lib/I18n/translations/german.yaml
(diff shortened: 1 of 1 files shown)


wlhlm:
Maybe I misunderstand what your new script is doing, but looks to me like all of this is already implemented in `gen_i18n.py` if you run it with `--verbose`. It shows both when a translation falls back to `english.yaml` and when it has a string not present in `english.yaml`.

Here are some output samples:
```
  INFO: 'STR_RECOVERY_MODE_HINT' missing in SI, using English fallback
  WARNING: SV has keys not in English: STR_SEAMLESS_SLEEP
```

dasrecht:
@wlhlm hah thanks ... i completely missed that existing script :( the only benefit my script has is that it's a tad easier to take over the missing strings and translate them as it only targets one language. but happy to close and abandon this.

Issue #2744, opened 2026-07-27
X3 dict lookup runs out of memory


cartomichelines:
### Affected Version

1.5.0 RC02

### Describe the Bug

X3 runs out of memory when text anti-aliasing is enabled.

NOTE: used a spanish-spanish dictionary from reader.dict. Book used: Lord of the Rings Trilogy in spanish (don’t if the book’s size may be a controburing factor)

### Steps to Reproduce

1. Enable text anti-aliasing
2. Go into a book, look up a word and it fails with message that it ran out of memory.

NOTE: used a spanish-spanish dictionary from reader.dict. Book used: Lord of the Rings Trilogy in spanish (don’t if the book’s size may be a controburing factor)

### Expected Behavior

Dictionary lookup should say that it didn’t find the word or show the definition if successful.

### Relevant Log Output/Screenshots

```shell

```

Uri-Tauber:
@itsthisjustin Does anti-aliasing consume more RAM than BW rendering?

itsthisjustin:
> @itsthisjustin Does anti-aliasing consume more RAM than BW rendering?

Yes and on x3 even more so I believe. I'd need to trace it but yes I think so

kevinkaw:
I've encountered a similar issue with X4, the dictionary runs out of memory when looking-up a word on a newly opened book. 

To fix it:
(1) I had to return to the homescreen; (2) then wait for the book thumbnail to loaded; (3) return to the book and dictionary works without issue.

PR #2791, opened 2026-07-29
fix: dictionary lookup OOM on .dict.dz definitions

W-Floyd:
## Summary

* **Goal:** Fix `.dict.dz` dictionary lookups failing with **"Not enough memory"** (#2744). Two independent causes on the definition-extraction path, one commit each.

* **Changes:**

  **1. Stop buffering the whole compressed chunk.** `extractChunkSlice()` read the entire compressed chunk into one heap buffer before inflating it. A dictzip chunk is ~58 KB uncompressed — **18,428 bytes compressed** for a real dictionary — and that block was held live while the 32 KB inflate window was allocated on top of it. Two large contiguous requests at once, on a heap sitting under 50 KB free mid-book. uzlib supports a pull callback, so the input is now fed from the file through a 2 KB refill buffer.

  **2. Allocate largest-first.** Even after (1), the lookup still failed — and the instrumented margin was *twelve bytes*:

  ```
  chunkSlice:enter          largest=36852   <- would have fit the ring
  after chunkSource (3348)  largest=32756
  inflateWindow need=32768  WILL FAIL
  ```

  Every buffer the lookup needs is carved from the same big free run, so **allocation order decides the outcome**. The ~3 KB `ChunkSource` was taken first, out of the only block large enough for the 32 KB ring. Reversing the order gives the ring the big block while it is still whole.

  | | Before | After |
  |---|---|---|
  | Largest block alongside the 32 KB ring | ~18 KB (whole chunk) | **~2.2 KB** |
  | Ring allocated | after the small buffers | **first** |

  Note the heap here is **not** badly fragmented — `frag=15%`, 14 free blocks. This was a shortfall-and-ordering bug, not a fragmentation one.

## Scope Check

Please confirm:

- [x] No new feature surface at all — this is a bugfix to an already-in-scope feature (dictionary lookup). Stock reports this case as an error the user can do nothing about.

## Additional Context

* **Behaviour is unchanged on success.** A hit returns the same bytes; only the peak heap needed to produce them changed.
* **The chunk boundary is still enforced.** `ChunkSource::remaining` stops the callback at `compressedSize`, so it never reads into the next chunk — exactly the hard limit the whole-chunk buffer imposed.
* **IO errors stay distinguishable from corruption.** uzlib collapses a callback `-1` into a generic decode failure, which would have turned every SD hiccup into `Decompress` and partly undone #2706. `readFailed` records the cause and maps it back to `ReadError`.
* **`InflateReader::initWithRing()`** is new: streaming mode over a caller-owned ring, needed because the reader is embedded in the struct whose allocation must come second. `ownsRing` keeps `deinit()` from freeing a buffer it doesn't own, so the existing `init(true)` callers in `ZipFile` and `PngToBmpConverter` are unaffected.
* **Memory / flash impact:** no new persistent RAM; peak transient demand per lookup drops by ~16 KB, and the 32 KB ring now comes out of the largest free block rather than whatever is left after it. `ChunkSource` (~2.2 KB, mostly the 2 KB refill buffer) is heap-allocated rather than stacked, to stay clear of the task stack budget.
* **Reviewer focus:** the uzlib callback contract in `chunkReadCb` (refill, set `source`/`source_limit` to the tail, return the first byte) and the `window`/`src` declaration order in `extractChunkSlice` — the ring must outlive the reader.
* **Possible latency shift:** SD access changes from one bulk chunk read to interleaved 2 KB reads. The discard phase stays pure sequential reads; only the extract phase (definition-sized, typically a few KB) alternates reads with writes to the temp file. No slowdown noticed in use, but it isn't benchmarked.
* **Relationship to #2733:** disjoint files (that PR is `Dictionary.{h,cpp}`, the index-search path; this is `DictZip.cpp` + `InflateReader`). Neither depends on the other and they should not conflict. #2733 does **not** fix this OOM — the failure is entirely downstream of what it changes.

### Testing

* Builds clean for the device `default` env (ESP32-C3); clang-format clean.
* **Host-verified against `lib/uzlib`** with a real 58,315-byte chunk: full-chunk decode, mid-chunk slice (the case every lookup takes), and a slice ending exactly at the chunk boundary all match the source bytes; a truncated `compressedSize` is refused rather than over-read.
* **Tested on hardware (Xteink X3):** the lookup that reliably reproduced "Not enough memory" (focus reading + book style + anti-aliasing on) now succeeds, and no OOM has recurred in use since. This is hands-on use rather than a systematic soak, so I can't claim the failure mode is fully eliminated — but the specific reproducer is fixed.

---

### AI Usage

Did you use AI tools to help write this code? _**YES**_ — diagnosed with Claude Code using temporary heap instrumentation (largest-free-block probes, not included in this PR), and the fix was implemented with it. Reviewed and tested on hardware (X3) by me.

### DIFF
--- a/lib/InflateReader/InflateReader.cpp
+++ b/lib/InflateReader/InflateReader.cpp
@@ -27,12 +27,14 @@ bool InflateReader::init(const bool streaming) {
   return true;
 }
 
-void InflateReader::initWithRing(uint8_t* ring) {
+bool InflateReader::initWithRing(uint8_t* ring) {
   deinit();  // free any owned ring buffer and reset state
+  if (!ring) return false;
   ringBuffer = ring;
   ownsRing = false;  // caller's buffer: never freed here
   memset(ringBuffer, 0, INFLATE_DICT_SIZE);
   uzlib_uncompress_init(&decomp, ringBuffer, INFLATE_DICT_SIZE);
+  return true;
 }
 
 void InflateReader::deinit() {
--- a/lib/InflateReader/InflateReader.cpp
+++ b/lib/InflateReader/InflateReader.cpp
@@ -4,7 +4,7 @@
 #include <type_traits>
 
 namespace {
-constexpr size_t INFLATE_DICT_SIZE = 32768;
+constexpr size_t INFLATE_DICT_SIZE = InflateReader::RING_BYTES;
 }
 
 // Guarantee the cast pattern in the header comment is valid.
@@ -19,6 +19,7 @@ bool InflateReader::init(const bool streaming) {
   if (streaming) {
     ringBuffer = static_cast<uint8_t*>(malloc(INFLATE_DICT_SIZE));
     if (!ringBuffer) return false;
+    ownsRing = true;
     memset(ringBuffer, 0, INFLATE_DICT_SIZE);
   }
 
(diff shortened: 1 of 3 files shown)


W-Floyd:
Still doesn't (completely) fix the dictionary loading failures when under pressure/fragmentation.

barbarhan (approved):
I reviewed the new caller-owned ring-buffer path, `ChunkSource` callback streaming, allocation/destruction order, short-read handling, compressed-chunk boundary enforcement, and error classification.

The externally owned ring outlives `InflateReader`, `ownsRing` prevents both leaks and double-free, and every new DictZip `read()` call is made with a positive batch size. Short positive file reads are handled by refilling, while exhausted chunk input and actual I/O failures remain distinguishable.

CI is green, and I found no actionable issue introduced by this change.

Uri-Tauber (approved):
LGTM. Thanks @barbarhan

PR #2793, opened 2026-07-29
refactor: resume by paragraph anchor after a re-pagination

### LINKED ISSUE #2778: 1.5.0 RC2: Book doesn't resume at correct page after changing text settings
### Affected Version

1.5.0 RC2

### Describe the Bug

Changing text settings & resuming a long chapter (mid-incremental-build) doesn't jump to the correct page; the reader resumes at the same absolute page number.

I suspect the underlying issue is that `applyDeferredReposition` always short-circuits during an incremental section build (`section->isBuilding()`): https://github.com/crosspoint-reader/crosspoint-reader/blob/fe37bf80ab0ebe07ecaf87747969eab2f6c3a1a9/src/activities/reader/EpubReaderActivity.cpp#L1433

..and the wrong `cachedChapterTotalPageCount` persists until the incremental build completes.

---

Another fun corollary of this bug: if you make the text smaller & read toward the end of the chapter, the reader will jump to the last page of the chapter as soon as the incremental build completes :^)

---

### Steps to Reproduce

1. Go to the beginning of a long chapter
2. Clear book cache
3. Resume book; verify incremental build (look for `~` before page number)
4. Move forward a handful of pages (note page number)
5. Go to settings; change font size
6. Resume book; notice that the same absolute page number was used as a resume point, rather than seeking to the correct page.

### Expected Behavior

The reader should resume approximately at the same point in the chapter after changing text settings.

```shell

```

Uri-Tauber:
# Summary

## The problem

When you're reading and you change something that reflows the text — font, font size, line spacing, screen orientation, auto page-turn — the book has to be re-laid-out from scratch. All the page breaks land in different places than before.

The reader used to remember your position as a page number ("you were on page 142"). But after a reflow, page 142 is no longer the same content — smaller text means more pages, so the old page 142 might now be a completely different passage. The result: you'd change a setting and get dumped somewhere other than where you were reading, and have to hunt to find your spot again.

## The fix

This branch changes what "your place in the book" means. Instead of remembering a page number, the reader now remembers which paragraph you were reading. A paragraph is the same paragraph no matter how the text is re-laid-out, so after any setting change the reader re-finds that paragraph under the new layout and puts you right back on it.

To keep things fast, it still uses the exact page number as a shortcut whenever nothing changed — so a normal "close and reopen" is instant, and the paragraph re-matching only kicks in when a setting actually forced a reflow.

fix #2778 

---

### AI Usage

### DIFF
--- a/docs/file-formats.md
+++ b/docs/file-formats.md
@@ -330,3 +330,50 @@ if (parsedSize != fileSize) {
     std::warning(std::format("Unparsed data detected: {} bytes remaining at offset 0x{:X}", fileSize - parsedSize, parsedSize));
 }
 ```
+
+## `progress.bin`
+
+Stores the reader's resume position for a book (one file per book, at the cache
+root). Unlike `book.bin` / `section.bin`, `progress.bin` is **irreplaceable user
+state**, so a format change is *migrated, never discarded*: older records are still
+read and upgraded to the current version on the next save.
+
+Records are discriminated by **length**. Legacy records have no version field:
+
+- **4 bytes** — `{ u16 spineIndex; u16 page; }`
+- **6 bytes** — `{ u16 spineIndex; u16 page; u16 totalPages; }`
+
+Versioned records lead with a `version` byte (currently `1`). A legacy record is
+never 27 bytes, so length alone separates the two; the version byte then carries
+format evolution past v1.
+
+### Version 1 (27 bytes)
+
+Adds the paragraph anchor and the render spec so a resume can tell *same
+pagination* (use the exact page) from *re-paginated* (remap to the saved
+paragraph). `totalPages` is display-only (the "page X of Y" total shown before the
+section finishes loading). `paragraphIndex == 0xFFFF` means "no anchor recorded".
+See `src/activities/reader/EpubReaderUtils.h`.
+
+```c++
+struct ProgressBinV1 {
+    u8  version;          // == 1
+    u16 spineIndex;
+    u16 page;
+    u16 paragraphIndex;   // 0xFFFF = unknown
+    u16 totalPages;       // display-only
+    // ReaderRenderSpec, in the same field order Section serializes its header:
+    s32 fontId;
+    float lineCompression;
+    u8  extraParagraphSpacing;
+    u8  paragraphAlignment;
+    u16 viewportWidth;
+    u16 viewportHeight;
+    u8  hyphenationEnabled;
+    u8  embeddedStyle;
+    u8  imageRendering;
+    u8  focusReadingEnabled;
+};
+
+ProgressBinV1 progress @ 0x00;
+```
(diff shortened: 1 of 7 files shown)


brianhuster:
There are Chinese and Japanese epubs that use a `<br/>` instead of `<p>` to seperate paragraphs btw

Uri-Tauber:
> There are Chinese and Japanese epubs that use a <br/> instead of <p> to seperate paragraphs btw

I also saw English epubs with such structure. Its definitely count as illformed epub. The question how common it is?

brianhuster:
It is very common in Chinese, that's why we had this PR https://github.com/crosspoint-reader/crosspoint-reader/pull/2710

For Japanese, it is also common, which is why even as recent as 2023, they had to write an article to encourage people _not_ to use `<br />` as paragraph seperator https://www.japandaisy.org/139. Even though it is discouraged, it proves how common it is prior to 2023. In fact, most books from Aozora Bunko (the Japanese equivalent of Project Gutenberg) use `<br />` as paragraph seperator. Given that most Japanese e-books users can read on XTEINK are converted from Aozora Bunko (I think book piracy is almost non-existent in Japan), this is definitely worth considering.

I think it is probably worth explaining why China and Japan used `<br/>` as paragraph separator instead of standard `<p>`. In both countries, space between paragraphs are the same as space between consecutive lines inside a paragraph, they also use text indentation as visual separator between paragraphs. Combining `<br />` and "full-width space" (U+3000, HTML standard requires it not to be trimmed) allow them to achieve such visual result without any CSS.

Also #2805 seems like a good alternative on positioning in epub IMO

Uri-Tauber:
@brianhuster Thank you for your detailed explanation.

I believe #2805 only touch Kosync paths.

thiagokokada:
> I believe #2805 only touch Kosync paths.

True, but maybe the same idea could be applied here. If it is a idea good enough to sync between cross devices accurately, it should be a good idea to also sync in device after re-pagination (that IMO it is a easier problem to solve).

Uri-Tauber:
@thiagokokada your implementation is more precise than mine.
I plan to release 1.5.0_RC03 tomorrow night (about 35 hours from now, in my timezone).
Do you think it's enough time for you to extend your PR to support re-pagination, and test it thoroughly?
Thx.

thiagokokada:
> @thiagokokada your implementation is more precise than mine. I plan to release 1.5.0_RC03 tomorrow night (about 35 hours from now, in my timezone). Do you think it's enough time for you to extend your PR to support re-pagination, and test it thoroughly? Thx.

Let's try.

lukestein:
Because this issue has been bugging me, I'm testing on device and so far it seems to work well and as expected. Thanks!

(I realize there may be an attempt to use something based on #2805 instead though.)

itsthisjustin:
@Uri-Tauber you good to get this merged?

Uri-Tauber:
> @Uri-Tauber you good to get this merged?

#2805 was better than mine.

PR #2797, opened 2026-07-30
fix: don't rewrite settings and re-layout when text settings close unchanged

W-Floyd:
## Summary

* **Goal:** opening the reader's text settings and backing out without changing anything should cost nothing.
* **Changes:** snapshot the settings `TextSettingsActivity` can edit, compare on return, and skip the save and the re-layout when nothing moved.

The `TEXT_SETTINGS` callback added in #2788 runs unconditionally:

```cpp
[this](const ActivityResult&) {
  SETTINGS.saveToFile();
  ...
  section.reset();
}
```

So merely opening the menu to look at it and pressing back costs a `SETTINGS.saveToFile()` and a `section.reset()`. Two separate problems:

* **The write burns a SPIFFS erase cycle for no change.** CLAUDE.md's write-throttling rule asks for a `if (newVal == _current) return;` guard on exactly this shape.
* **The reset discards the laid-out chapter.** On a long chapter that means a full re-layout and a fresh incremental build, for a menu the user only looked at.

`applyOrientation()` already guards this way — it early-returns when the selected orientation matches settings. This does the same.

## Additional Context

The snapshot covers all eleven settings `TextSettingsActivity` can write (`fontFamily`, `fontPointSize`, `sdFontFamilyName`, `lineSpacing`, `paragraphAlignment`, `extraParagraphSpacing`, `screenMargin`, `hyphenationEnabled`, `embeddedStyle`, `focusReadingEnabled`, `textAntiAliasing`).

`textAntiAliasing` is included even though it affects rendering rather than geometry, so a change to it still triggers the re-layout path. Over-invalidating on one flag is cheaper than letting a real change slip through; happy to special-case it to a repaint-only path if a reviewer prefers.

Memory: ~42 bytes of stack, captured by value into the existing callback. No heap.

**Related but explicitly not fixed:** #2778 (wrong resume page after a text settings change). This removes one *trigger* — closing unchanged no longer resets the section and runs the reposition path — but the root cause that issue identifies, `applyDeferredReposition` short-circuiting while `section->isBuilding()`, is untouched.

**Not tested on hardware.** Needs: open text settings and back out unchanged (page should not re-layout, no SPIFFS write in the log), then change font size and confirm the re-layout still happens and position is preserved.

## Scope Check

- [x] Not a new built-in theme.
- [x] Not a new external network connector.
- [x] Bug fix to an existing menu path; no new surface.
- [x] Does not touch `freeink-sdk/`, `lib/hal/`, bootloader, OTA, or recovery.

---

### AI Usage

### DIFF
--- a/src/activities/reader/EpubReaderActivity.cpp
+++ b/src/activities/reader/EpubReaderActivity.cpp
@@ -12,6 +12,8 @@
 #include <esp_system.h>
 
 #include <algorithm>
+#include <cstdio>
+#include <cstring>
 #include <functional>
 #include <iterator>
 #include <limits>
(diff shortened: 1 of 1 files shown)


W-Floyd:
@coderabbitai review

Uri-Tauber:
> The write burns a SPIFFS erase cycle for no change. CLAUDE.md's write-throttling rule asks for a if (newVal == _current) return; guard on exactly this shape.

Nothing in this firmware writes SPIFFS.

Uri-Tauber:
Also, I think #2793 does a similar thing along the way.

PR #2803, opened 2026-07-30
fix: order languages by native name instead of language code

### LINKED ISSUE #2799: Wrong order languages
### Affected Version

1.5.0

### Describe the Bug

Wrong order based on native name of each language

Correct reordering:
- Polish, then Portuguese (Portugal)
- Slovenčina (Slovak), then Slovenščina (Slovenian)

### Steps to Reproduce

Select Language in Settings

### Expected Behavior

Incorrect order

```shell

```

dasrecht:
## Summary
order languages by native name instead of language code for better sorting. e.g. the Português (Portugal) has language code `P2` which throws it out of alphabetical order. This way the sorting works and we don't need to touch the `_order` of the I18n files.

Closes #2799

## Scope Check

Please confirm:

## Additional Context

* Ordering the languages based on the ISO639 Language code works mostly except when we deviate from it (e.g. with Português (Portugal) that has the code `P2` assigned and gets thrown out of order) or with Slovenščina (SI) and Slovenčina (SK)
* A knock on is that ISO639 does not differentiate between different regions of a language, so Português in theory would be `pt` no matter the region - this has been noted noted in #1506
* The language sorting was mostly refined a few weeks ago in #1071 
---

### AI Usage

Did you use AI tools to help write this code? _**< PARTIALLY  >**_


Uri-Tauber (approved):
LGTM

osteotek:
And now Cyrillic languages are shown after Latin languages. Sorting by language code was introduced for a reason. There is no "native order" for languages to be sorted by, since they don't have the same alphabet. This was discussed extensively when translations mechanism was developed. Sorting by language code is the least bad option out of other more bad options.
@dasrecht Let's revert this

dasrecht:
I agree it's not perfect. This is why I would also suggest that we add the BCP47 language tags (mentioned here https://github.com/crosspoint-reader/crosspoint-reader/pull/1506#issuecomment-4130245356). Because at some point, we will need to solve this issue for good and while we are working on language improvements, we could also do that improvement in the process.

Uri-Tauber:
@osteotek I don't really mind either way.

I suggest to revert this now, and migrate to BCP-47 after release 1.5.0.

EDIT: @osteotek do you still have write permissions here?

dasrecht:
Also fine with that. Shall we discuss the way forward in the initial issue? I've a rough POC that i'm testing but want to make sure we don't break anything especially regarding the `_order` which seems to be used in places

PR #2805, opened 2026-07-30
fix: make EPUB sync positions content-based

thiagokokada:
## Summary

  Make EPUB reading positions independent of local pagination by treating the spine index plus visible-text offset as
  authoritative. CrossPoint continues to speak the existing KOReader sync protocol; the `progress` field remains a
  KOReader-compatible XPath/XPointer and `percentage` remains metadata and fallback.

  * Record the visible Unicode-codepoint offset at the start of every rendered page in the section cache.
  * Upload progress as `page -> exact page-start offset -> KOReader XPath`, without a page-percentage approximation.
  * Resolve downloaded XPaths to exact content offsets and derive the page from the current local layout.
  * Persist content offsets locally so font, orientation, and other pagination changes reopen at the same text.
  * Keep percentage, paragraph/list-item LUTs, and the existing CrossPoint rich position fields as compatibility
    fallbacks only.
  * Preserve lightweight support for nested `text()[N]`, character offsets, image anchors, and omitted unindexed
    wrappers. This covers the interoperability cases from #2777 without adding its full image-XPath scan.
  * Document the new section-cache LUT and content-based sync behavior.

## Scope Check

Please confirm:

      the relevant maintainer.

Not applicable. This is an in-scope EPUB rendering/progress-tracking correction and does not add a connector or touch
`freeink-sdk/`, `lib/hal/`, the bootloader, OTA, or recovery code.

## Additional Context

The root cause is that CrossPoint previously converted a rendered page number to a chapter fraction, approximated a
text location from that fraction, and generated an XPath. On download it reversed the same approximation. Since page
counts depend on font and layout, the position could move even when a valid content anchor was available.

The section cache format is bumped to v35, so existing rendered section caches rebuild automatically. Existing 4-byte
and 6-byte local progress files remain readable; newly written EPUB progress files add a 32-bit content offset.

Resource impact:

* Section cache: 4 additional bytes per rendered page.
* Pagination working memory: 4 additional bytes per logical text token while laying out a chapter.
* Measured static RAM usage is unchanged; the default firmware build grows by roughly 1.4 KB of flash.

Validation:

* `clang-format --dry-run --Werror` on all changed C/C++ files.
* `git diff --check` with the repository-compatible whitespace configuration.
* `pio run -e default` — successful ESP32-C3 firmware build and link.
* Tested on Xteink X4 and a Kindle PW3 running KOReader

---

### AI Usage

### DIFF
--- a/lib/Epub/Epub/ParsedText.h
+++ b/lib/Epub/Epub/ParsedText.h
@@ -28,10 +28,17 @@ class ParsedText {
   std::vector<bool> wordContinues;      // true = word attaches to previous with no break
   std::vector<bool> wordNoSpaceBefore;  // true = may break before token, but no synthetic space when joined
   std::vector<bool> wordIsFocusSuffix;  // true = token is the regular tail of a focus bold-prefix split
-  // Zero-based visible Unicode-codepoint offset in the spine body for each logical
-  // token. This is layout-only metadata: page-start offsets are persisted in the
-  // section LUT, so rendered TextBlocks do not need to carry it.
-  std::vector<uint32_t> wordVisibleOffsets;
+  // Zero-based visible Unicode-codepoint offsets in the spine body, stored as
+  // uint16_t deltas from a shared base to keep this layout-only metadata small.
+  // Pathological spans wider than uint16_t use sparse rebases; rendered
+  // TextBlocks do not carry any of this metadata.
+  struct VisibleOffsetRebase {
+    size_t wordIndex;
+    uint32_t base;
+  };
+  std::vector<uint16_t> wordVisibleOffsetDeltas;
+  uint32_t visibleOffsetBase = 0;
+  std::vector<VisibleOffsetRebase> visibleOffsetRebases;
   std::deque<std::string> rubyTexts;
   BlockStyle blockStyle;
   bool extraParagraphSpacing;
@@ -47,6 +54,11 @@ class ParsedText {
   std::vector<bool> reorderedFocusSuffixScratch;
   std::vector<uint16_t> visualOrderScratch;
 
+  uint32_t visibleOffsetBaseAt(size_t wordIndex) const;
+  uint32_t visibleOffsetAt(size_t wordIndex) const;
+  void pushVisibleOffset(uint32_t offset);
+  void insertVisibleOffset(size_t wordIndex, uint32_t offset);
+  void eraseVisibleOffsetPrefix(size_t count);
   int resolveFirstLineIndent(bool isFirstLine, const GfxRenderer& renderer, int fontId) const;
   std::vector<size_t> computeLineBreaks(const GfxRenderer& renderer, int fontId, int pageWidth,
                                         std::vector<uint16_t>& wordWidths, std::vector<bool>& continuesVec,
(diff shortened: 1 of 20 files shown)


thiagokokada:
For reviewers: I am using this as a daily drive between my Kindle PW3 (KOreader) and Xteink X4 (CrossPoint), and I am getting much more accurate sync position between KOreader -> CrossPoint (the other way was already accurate).

Before pushing the sync in KOReader and pulling in CrossPoint in a book I am reading (WorldEnd Vol. 3) would result in the sync being off by ~10+ more pages. Now it is pretty much on point or very near (maybe off by ~1 page).

I understand though that this PR is big enough that this is probably something that will be post 1.5.0 release.

lib/Epub/Epub/ParsedText.h
```diff
@@ -28,6 +28,10 @@ class ParsedText {
   std::vector<bool> wordContinues;      // true = word attaches to previous with no break
   std::vector<bool> wordNoSpaceBefore;  // true = may break before token, but no synthetic space when joined
   std::vector<bool> wordIsFocusSuffix;  // true = token is the regular tail of a focus bold-prefix split
+  // Zero-based visible Unicode-codepoint offset in the spine body for each logical
+  // token. This is layout-only metadata: page-start offsets are persisted in the
+  // section LUT, so rendered TextBlocks do not need to carry it.
+  std::vector<uint32_t> wordVisibleOffsets;
```
Uri-Tauber: `wordVisibleOffsets` (`lib/Epub/Epub/ParsedText.h:34`)
stores a full `uint32_t` per token and stays alive through layout alongside the other per-word arrays — roughly 3KB at peak on a 750-token flush. 
Since it has a single consumer (`ParsedText.cpp:1097`, `lineVisibleOffset`), the offsets could be `uint16_t` deltas from a `uint32_t visibleOffsetBase` member, halving it to ~1.5KB. 
`addWord`'s parameter and the `processLine` callback would keep taking absolute `uint32_t`, so the change stays entirely inside `ParsedText`.

thiagokokada: Done.

thiagokokada:
Tested the problematic cases that I knew:
- Sub-chapter start by title
- Chapter division using images
- Long jumps between sync

Everything looks good for me.

Edit: wait, this PR broke the CrossPoint -> KOReader part instead, investigating...

thiagokokada:
Ok, now I tested both the CrossPoint -> KOReader, CrossPoint -> CrossPoint and KOReader -> CrossPoint. All looks good.

Just let me check if `@coderabbitai` will have any other concerns about the code.

thiagokokada:
CC @Uri-Tauber for review.

Uri-Tauber:
@thiagokokada Thats was quick. 
Have you tested with changing font and size?

thiagokokada:
> @thiagokokada Thats was quick. Have you tested with changing font and size?

Tested and confirmed working.

I found an issue where where if the chapter is not indexed (for example, right I changed the fonts) we will fallback to the old behavior (based by percentage), but I don't think there is much we can do in this case.

Otherwise I tested different font and font size from both CrossPoint -> CrossPoint and KOReader -> CrossPoint, both working fine too.

Uri-Tauber:
@thiagokokada Awesome. 
I will have time to properly review tomorrow night.

Uri-Tauber (approved):
LGTM. some minor optimizations, but I will just open new PR for those.

PR #2806, opened 2026-07-30
fix: Save text settings immediately on each change

itsthisjustin:
Previously, text settings were only saved when the activity finished normally, causing changes to be lost when exiting via home gesture or sleep. Now each setting change (font family, line spacing, paragraph alignment, screen margin) immediately persists to disk, matching the behavior of other settings screens.

### DIFF
--- a/src/activities/reader/EpubReaderActivity.cpp
+++ b/src/activities/reader/EpubReaderActivity.cpp
@@ -811,7 +811,7 @@ void EpubReaderActivity::onReaderMenuConfirm(EpubReaderMenuActivity::MenuAction 
       startActivityForResult(std::make_unique<TextSettingsActivity>(renderer, mappedInput, &sdFontSystem.registry(),
                                                                     TextSettingsActivity::Tab::Family),
                              [this](const ActivityResult&) {
-                               SETTINGS.saveToFile();
+                               // TextSettingsActivity saves on each change; no save needed here.
                                // Font/size/spacing/margin changes invalidate the current
                                // layout: preserve position and force a re-layout, mirroring
                                // applyOrientation()'s reflow.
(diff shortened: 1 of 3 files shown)


Uri-Tauber:
The problem is users will want to change things to see how they look, and each change will cost a disk-write. 
Could you do instead, saving when exiting the text settings?
If the user press back, and the activity stack pop the text settings away - store to disk.

itsthisjustin:
> The problem is users will want to change things to see how they look, and each change will cost a disk-write. Could you do instead, saving when exiting the text settings? If the user press back, and the activity stack pop the text settings away - store to disk.

Why is that a big deal though? That's literally the bug though. Doing it only on back means it doesn't save if you sleep or exit to home directly

Uri-Tauber:
> Why is that a big deal though?

Slower navigation mainly.
If you could find cleaner way to solve this, it will be nice.
If not, its good enough as it is.

itsthisjustin:
> > Why is that a big deal though?
> 
> Slower navigation mainly. If you could find cleaner way to solve this, it will be nice. If not, its good enough as it is.

I had another way of doing it which was to detect if settings is dirty or not and save regardless of exit method. It just seemed way more messy than just writing it when i change it.

itsthisjustin:
> > > Why is that a big deal though?
> > 
> > 
> > Slower navigation mainly. If you could find cleaner way to solve this, it will be nice. If not, its good enough as it is.
> 
> I had another way of doing it which was to detect if settings is dirty or not and save regardless of exit method. It just seemed way more messy than just writing it when i change it.

Plus if we ever have settings that cascade depending on which setting you toggle another setting might toggle differently, this prevents collissions

Uri-Tauber (approved):

PR #2814, opened 2026-07-31
fix: Replace vector with deque for text token storage

itsthisjustin:
Prevents crashes when parsing CJK text with thousands of tokens. Vector's contiguous reallocation fails under fragmented heap, while deque's chunked growth (~512B nodes) avoids large contiguous allocations. 

ParsedText stored a paragraph's tokens in a std::vector<std::string>, which reallocates its whole array into one contiguous block as it grows. This is technically not specific to CJK but will trigger more often there. CJK makes one token per character, but any long Latin paragraph can reach the same count. Switching to std::deque fixes both, since it grows in fixed ~512 B nodes and never needs that one giant contiguous allocation.

This builds upon #1038 

Moving to std::deque keeps the indexing, near-vector locality wins from Zach's original PR but grows in fixed ~512 B nodes, so it never needs one big contiguous block.  So no crashing vector and faster than the slow pre-#1038 list.


Uri-Tauber (approved):
LGTM

PR #2828, opened 2026-08-01
feat: add Cyrillic (ЙЦУКЕН) keyboard layout for Russian UI

winst0niuss:
## Problem

The on-screen keyboard ships Latin layouts only. With the UI set to Russian, there is no way to type Cyrillic at all — book search, KOSync usernames, OPDS/file names and any other free-text field are Latin-only. The SDK's implicit case flip does not help either: it handles ASCII `a-z`/`A-Z` and returns null for anything else.

## What this does

Adds ЙЦУКЕН letter layers (normal + shift) plus a **РУС/ABC** key that toggles between the Cyrillic and Latin letter layers in a single press.

- **Layer tables are app-defined**, built over the same public `fui::KeyboardKey` / `KeyboardLayout` structs the existing URL layers already use — no SDK change needed.
- **Key ids are the letters' Unicode code points** (U+0410..U+044F). They fit `int16_t` and cannot collide with the ASCII ids (8/13/32, `'a'`-`'z'`), the negative control ids (-1..-4) or the URL snippet ids (2001+).
- **Long-press mirrors the Latin hold-for-opposite-case behaviour**, supplied explicitly via `alt` since the SDK cannot derive it for Cyrillic. One deliberate exception: `е`/`Е` long-presses to `ё`/`Ё` — that letter has no key of its own in the ЙЦУКЕН arrangement, and reaching it matters more than a redundant case flip.
- **Latin side gets its own bottom row**, because a Cyrillic UI still has to type Wi-Fi passwords and server URLs and needs a way back. Its letter rows are the existing URL tables — identical QWERTY, already defined.
- **Scoped narrowly**: the layer appears only when the UI language is Russian (`languageUsesCyrillic()`), and never for URL fields, which keep the EN-arranged tables with their `:` / `/` / snippet keys. Text fields start on Cyrillic; credential fields start on Latin, which is what Wi-Fi passwords and usernames overwhelmingly are.
- UK/BE/KK are Cyrillic too but each swaps in letters this table lacks (і/ї/є/ґ, ў, ә/ң/ө/ұ/…), so they are deliberately not wired up — they need their own rows first.

## Memory cost

- Layout tables are `const` → flash, no DRAM.
- `InteractionBuffer<48>` → `<56>`: ЙЦУКЕН registers 48 key rects vs 41 for the EN layout, so 56 preserves the same headroom for both. That is ~160 bytes, one-off per `KeyboardEntryActivity` instance, on the activity object itself — no new heap allocation, nothing held across activities.

## Notes for review

- The mode key's label follows the layer it leads *back* to, so it reads `АБВ` while the Cyrillic layer is active — `abc` would point at the Latin layer the key does not lead to. That literal, like the static tables' `ABC`/`РУС`, is not run through `tr()`: it names an alphabet rather than translating a word, the same reasoning as the SDK's `?123`.
- No new i18n keys. The existing `STR_KEY_*` values are reused as-is.

## Testing

- `clang-format` 21 clean.
- Not yet verified on hardware — device testing (all orientations, touch hit-testing on the wider 12-key rows) still pending.

### DIFF
--- a/src/activities/util/KeyboardEntryActivity.cpp
+++ b/src/activities/util/KeyboardEntryActivity.cpp
@@ -93,6 +93,74 @@ const fui::KeyboardKey URL_SNIP_BOTTOM[] = {UKS("abc", fui::KeyKind::Mode, fui::
                                             UKS("Del", fui::KeyKind::Delete, fui::QWERTY_KEY_BACKSPACE, 2),
                                             UKS("OK", fui::KeyKind::Ok, fui::QWERTY_KEY_ENTER, 2)};
 
+// ---------------------------------------------------------------------------
+// Cyrillic (ЙЦУКЕН) layers. The SDK ships Latin layouts only, so — like the URL
+// layers above — these are app-defined tables over the same public structs.
+//
+// Key ids are the Unicode code points of the letters (U+0410..U+044F). They fit
+// int16_t and cannot collide with the ASCII ids (8/13/32, 'a'-'z'), the
+// negative control ids (-1..-4) or the URL snippet ids (2001+).
+//
+// `alt` (long-press) mirrors the Latin behaviour of hold-for-opposite-case,
+// which the SDK's implicit case-flip cannot provide here: it only handles ASCII
+// a-z/A-Z and returns null for anything else. The one deliberate exception is
+// е/Е, whose long-press yields ё/Ё — the letter has no key of its own in the
+// ЙЦУКЕН arrangement, and reaching it matters more than a redundant case flip.
+// ---------------------------------------------------------------------------
+
+constexpr int16_t LANG_TOGGLE_VALUE = -4;  // mirrors KeyboardEntryActivity::LANG_KEY
+
+const fui::KeyboardKey RU_ROW1[] = {UKA("й", "й", 0x439, "Й"), UKA("ц", "ц", 0x446, "Ц"), UKA("у", "у", 0x443, "У"),
+                                    UKA("к", "к", 0x43A, "К"), UKA("е", "е", 0x435, "ё"), UKA("н", "н", 0x43D, "Н"),
+                                    UKA("г", "г", 0x433, "Г"), UKA("ш", "ш", 0x448, "Ш"), UKA("щ", "щ", 0x449, "Щ"),
+                                    UKA("з", "з", 0x437, "З"), UKA("х", "х", 0x445, "Х"), UKA("ъ", "ъ", 0x44A, "Ъ")};
+const fui::KeyboardKey RU_ROW2[] = {UKA("ф", "ф", 0x444, "Ф"), UKA("ы", "ы", 0x44B, "Ы"), UKA("в", "в", 0x432, "В"),
+                                    UKA("а", "а", 0x430, "А"), UKA("п", "п", 0x43F, "П"), UKA("р", "р", 0x440, "Р"),
+                                    UKA("о", "о", 0x43E, "О"), UKA("л", "л", 0x43B, "Л"), UKA("д", "д", 0x434, "Д"),
+                                    UKA("ж", "ж", 0x436, "Ж"), UKA("э", "э", 0x44D, "Э")};
+const fui::KeyboardKey RU_ROW3[] = {UKS("Shift", fui::KeyKind::Shift, fui::QWERTY_KEY_SHIFT, 2),
+                                    UKA("я", "я", 0x44F, "Я"),
+                                    UKA("ч", "ч", 0x447, "Ч"),
+                                    UKA("с", "с", 0x441, "С"),
+                                    UKA("м", "м", 0x43C, "М"),
+                                    UKA("и", "и", 0x438, "И"),
+                                    UKA("т", "т", 0x442, "Т"),
+                                    UKA("ь", "ь", 0x44C, "Ь"),
+                                    UKA("б", "б", 0x431, "Б"),
+                                    UKA("ю", "ю", 0x44E, "Ю"),
+                                    UKS("Del", fui::KeyKind::Delete, fui::QWERTY_KEY_BACKSPACE, 2)};
+
+const fui::KeyboardKey RU_SHIFT_ROW1[] = {
+    UKA("Й", "Й", 0x419, "й"), UKA("Ц", "Ц", 0x426, "ц"), UKA("У", "У", 0x423, "у"), UKA("К", "К", 0x41A, "к"),
+    UKA("Е", "Е", 0x415, "Ё"), UKA("Н", "Н", 0x41D, "н"), UKA("Г", "Г", 0x413, "г"), UKA("Ш", "Ш", 0x428, "ш"),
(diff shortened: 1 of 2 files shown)


Uri-Tauber:
Additional layouts should live in the SDK.
See comment https://github.com/crosspoint-reader/crosspoint-reader/pull/2481#issuecomment-5005425662

winst0niuss:
Good catch on the cursor-mode byte slicing — confirmed and fixed in 4289b44d.

`cursorPos` is kept on a code point boundary, but the block cursor sliced exactly one byte at that offset in four places (block width measurement, kerning offset, the two-part advance calculation, and drawing the character inside the block). For a two-byte Cyrillic letter that is a lone lead byte, so the renderer received an incomplete sequence and the block was sized to the wrong advance. Harmless while only ASCII was reachable — which is exactly what this branch changes.

Now uses the existing `utf8Next()` helper to take the whole code point. The masked-password path is unchanged: `displayText` is all `*` there, so the span stays one byte.

I did not extend this to the wider `cursorPos` handling (movement, insertion, deletion, line breaking) — those already go through `utf8Prev`/`utf8Next` or operate on spans whose boundaries are code point boundaries, so they are byte-offset-based by design rather than byte-oriented bugs.

winst0niuss:
Understood, and no argument — the ЙЦУКЕН tables belong next to AzertyFr/QwertzDe in `builtinKeyboardLayout()`, not in this activity. I'll move them into the SDK and cut this PR down to the one-line `layoutForLanguage()` mapping.

Before I open the SDK PR, one design question, because Cyrillic differs from the layouts already there in a way the Hebrew answer doesn't cover.

FR/DE/ES are Latin, so a user on those layouts can always type a Wi-Fi password, a server URL or an OPDS username. Cyrillic can't — those fields are Latin in practice, so a ЙЦУКЕН layout is only usable if there's a way back to QWERTY. That's what the РУС/ABC key in this PR does. The SDK currently has no way to express it: `KeyKind` has no language/alphabet member, and nothing marks a layout as "non-Latin, has a Latin companion".

Two ways I can see:

1. **`KeyKind::Lang` + a slot in the bottom row.** The non-Latin layout and the QWERTY layout it pairs with each carry the key; `builtinKeyboardLayout()` returns whichever the app asks for, and the app supplies the label per frame through `KeyboardProps`, exactly like `okLabel`/`shiftLabel`/`modeLabel` do today. Keeps the toggle a property of the layout, so every future non-Latin layout (Hebrew, Greek, Ukrainian) gets it for free.

2. **A `latin` flag on `builtinKeyboardLayout()`.** The SDK stays as-is apart from the tables, and the activity owns the toggle state. Smaller SDK change, but each app re-implements the same key.

I lean towards 1 — the pairing is a property of the layout, and it keeps `KeyboardEntryActivity` down to a bool. But it does add a `KeyKind` member and a label field, so it's your call. Happy to go with 2 if you'd rather keep the SDK surface unchanged.

Which would you prefer? I'll hold off on the SDK PR until then.

(Repo note: the two commits here also fix a pre-existing cursor-mode bug — the block cursor sliced one byte at `cursorPos` for measuring and drawing, which is a lone lead byte for any multi-byte character. That part is independent of where the tables live, so I can split it into its own PR if that's easier to land.)

Uri-Tauber:
I'm not sure I understood the bug you mentioned. But whatever it is, open a seperate PR for it.

I do like the idea of switch-lang key. Although it should be more general than russian-english only.
My thinking was to allow switching between English layout and current language layout.
So, for example, if your UI lang is Hebrew, you could switch between English and Hebrew. Same for Arabic, Russian, and any other languages. 
It will be a bigger SDK change than what you planned here - essentially adding a dozen of new layouts.
The firmware side will be to show the correct layouts alone (Arabic user don't have any use for Russian layout)

I hope it clear things. I planned to do it myself in the near future, but feel free to take it on yourself.

winst0niuss:
Split the cursor fix out into #2840 as you asked, and dropped it from this branch — this PR is now just the keyboard layout commit.

The bug description there should be clearer: it has before/after screenshots and a reproduction that needs no Cyrillic at all. It turns out the layouts already in the SDK are enough to trigger it — `é` on AzertyFr, `ü`/`ß` on QwertzDe, `ñ` on SpanishEs are all multi-byte, and the block cursor sliced a single byte off them.

On the switch-lang design: understood, and your framing is better than mine — pairing English with whatever the UI language is, rather than a Russian-specific toggle. I'll wait for your call on whether you're taking it on yourself before I start anything on the SDK side.

winst0niuss:
One idea on top of that, from how I actually use the reader.

English + the UI language is the right default, but for me it wouldn't be enough. My UI is Russian, and I read books in Spanish — so I want Spanish available too, for OPDS searches and filenames. Someone with a German UI who reads in French is in the same position. The language you read in is not necessarily the language your UI is in, and neither is necessarily English.

So I'd suggest:

- **Default: two layouts** — the UI language's layout plus English. That is your proposal and it covers most users with zero configuration.
- **Optional: let the user enable additional layouts** in settings, from the same set the SDK ships. In my case that would be RU + EN + ES.
- The lang key then **cycles through the enabled layouts** rather than toggling between two. With the default of two it behaves exactly like a toggle, so nothing changes for users who never open that setting.

The cost looks small on the firmware side: the layout tables are `const` and live in flash whether or not they are enabled, so enabling one costs no RAM. The enabled set is a bit mask in settings — one `uint16_t` covers 16 layouts. The key's label is already supplied by the app per frame through `KeyboardProps`, so it can show the code of the next layout in the cycle (`EN` → `RU` → `ES`).

It also keeps your rule that an Arabic user shouldn't see a Russian layout: nothing is shown unless the user asks for it, and the default stays two.

Not arguing for scope creep — if you'd rather ship the default-two version first and treat the "add more layouts" setting as a follow-up, that ordering makes sense to me. Wanted to raise it now in case it changes how you'd store or model the layout selection.

winst0niuss:
Closing in favour of #2858, which does this the way you asked, @Uri-Tauber.

The layouts moved into the SDK — [Free-Ink/freeink-sdk#29](https://github.com/Free-Ink/freeink-sdk/pull/29), merged — and the switch key became general rather than Russian-specific. `KeyKind::Lang` is a new key kind there, distinct from `Mode`, with its label supplied by the app since which layout is active is app state.

What grew beyond the original plan:

- **Nine layouts, not one.** Russian, Ukrainian, Belarusian, Kazakh and Hebrew are new; the existing French, German and Spanish gained an uppercase layer, which they had been drawing a Shift key for without one.
- **A user-selectable set** rather than a fixed UI-language + English pair. The reasoning is in #2858: the language someone reads in is not necessarily the language of their interface, and pairing cannot express that. The default is still your pairing, and it needs no setup.
- **Cyrillic case-flip in `keyboardAltOutputFor`**, so those layouts do not each need an explicit `alt` per key.

The cursor fix that was originally bundled here is #2840, still open separately.

Thanks for the review that redirected this — the SDK version is a great deal better than what was here.

PR #2831, opened 2026-08-01
fix: load Settings page's Wi-Fi/OPDS/settings data sequentially

obra:
## Summary

* **What is the goal of this PR?** Fix a reproducible crash: opening the web UI's Settings page reliably crashes the device (silent reboot to Home) under real-world conditions.
* **What changes are included?** `SettingsPage.html` fired three unawaited `fetch()` calls on page load (`/api/settings`, `/api/wifi`, `/api/opds`), opening simultaneous connections to the device's web server. That server (Arduino's `WebServer`) handles one client at a time; a genuinely concurrent second/third connection can stall a response mid-stream for multiple seconds — long enough to trip the firmware's watchdogs (300ms interrupt WDT / 5s task WDT, both panic-on-timeout in this project's config). This PR makes the three initial loads run sequentially via a small async IIFE instead of firing them concurrently. Same end state for the page, no visible behavior change.

## Scope Check

- [x] This is a bug fix for a reproducible crash, not new functionality.
- [x] Does not touch `freeink-sdk/`, `lib/hal/`, the bootloader, OTA, or recovery code.

## Testing performed

Verified live on an X4 (driven via the `CMD:WIFI` serial debug command from the companion PR, since I didn't have physical access to the device for part of this investigation):

- Root-caused via a temporary per-entry heap/timing log in `handleGetSettings()` (not included in this diff): under concurrent load, individual `sendContent()` calls stalled 1-3+ seconds between simple, static settings entries that should serialize in well under a millisecond each.
- Reproduced the crash directly: firing `/api/settings`, `/api/wifi`, and `/api/opds` concurrently (mirroring the page's old JS behavior) crashed the device in 2/2 attempts. The companion PR (watchdog reset detection) let the second crash actually get logged: reset reason `ESP_RST_TASK_WDT`, panic confirmed.
- After the fix: same three endpoints, hit sequentially (matching the new JS), succeeded cleanly across dozens of repeats (single-shot and a 60-request hammer loop) with no crashes and stable free-heap.
- Confirmed the live device served the updated, gzip-compressed generated header after a normal `pio run` (source HTML → `SettingsPageHtml.generated.h` via the existing `scripts/build_html.py` build step).

## Additional Context

Memory/flash impact: none — same JS logic, just sequenced instead of concurrent. No new allocations, no change to the generated header's size beyond the few added bytes of source text.

Risk: low. This only changes the *order* three existing fetches happen in on page load; no new endpoints, no changed request/response shapes.

---

### AI Usage

YES — diagnosed and fixed by Claude (Sonnet 5), including live reproduction and verification against real hardware.

### DIFF
--- a/src/network/CrossPointWebServer.cpp
+++ b/src/network/CrossPointWebServer.cpp
@@ -1248,6 +1248,8 @@ void CrossPointWebServer::handleGetSettings() const {
       seenFirst = true;
     }
     server->sendContent(output);
+    yield();                          // Yield to allow WiFi and other tasks to process during a slow send
+    resetTaskWatchdogIfSubscribed();  // Reset watchdog: each sendContent() is a blocking network write
   }
 
   server->sendContent("]");
@@ -1360,6 +1362,8 @@ void CrossPointWebServer::handleGetOpdsServers() const {
 
     if (i > 0) server->sendContent(",");
     server->sendContent(output);
+    yield();                          // Yield to allow WiFi and other tasks to process during a slow send
+    resetTaskWatchdogIfSubscribed();  // Reset watchdog: each sendContent() is a blocking network write
   }
 
   server->sendContent("]");
@@ -1476,6 +1480,8 @@ void CrossPointWebServer::handleGetWifiNetworks() const {
 
     if (i > 0) server->sendContent(",");
     server->sendContent(output);
+    yield();                          // Yield to allow WiFi and other tasks to process during a slow send
+    resetTaskWatchdogIfSubscribed();  // Reset watchdog: each sendContent() is a blocking network write
   }
 
   server->sendContent("]");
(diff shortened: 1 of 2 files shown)


obra:
Update: the JS-sequencing fix alone wasn't sufficient. Confirmed live — after that fix, loading the Settings page still crashed the device via watchdog panic under a slightly different concurrency trigger (a browser's automatic `favicon.ico` request racing the page's own sequential fetches, most likely — this project has no `<link rel="icon">` on any page).

Pushed a second commit that addresses the actual gap: `handleGetSettings()`, `handleGetOpdsServers()`, and `handleGetWifiNetworks()` were the only per-entry streaming-response loops in `CrossPointWebServer.cpp` that never called `resetTaskWatchdogIfSubscribed()` — every other similar loop in this file (file listing, uploads, font handling) already does. This is the more durable fix: it protects these three handlers against *any* slow-response scenario, not just the one the JS change avoids.

Re-verified on the same X4 hardware: Settings page now loads reliably.

itsthisjustin:
This matches a similar fix I had on another branch so makes sense to me

itsthisjustin (approved):

PR #2834, opened 2026-08-02
fix: thread-safety issues in credential store

itsthisjustin:
Replace hand-rolled lazy initialization with C++11 function-local static to ensure thread-safe key initialization on dual-core targets. Add mutex protection to credential store operations to prevent race conditions. Include password length validation to detect corrupted credentials during deserialization.

Wifi passwords could become corrupted when recent wifi network changed, a new wifi network was added, or if you edited your wifi info in the web server


src/WifiCredentialStore.h
```diff
@@ -36,17 +41,17 @@ class WifiCredentialStore : public PersistableStore<WifiCredentialStore> {
   // Credential management
   bool addCredential(const std::string& ssid, const std::string& password);
   bool removeCredential(const std::string& ssid);
-  const WifiCredential* findCredential(const std::string& ssid) const;
+  std::optional<WifiCredential> findCredential(const std::string& ssid) const;
 
-  // Get all stored credentials (for UI display)
-  const std::vector<WifiCredential>& getCredentials() const { return credentials; }
+  // Return snapshots so callers cannot retain references across mutations.
+  std::vector<WifiCredential> getCredentials() const;
```
Uri-Tauber: `getCredentials()` returns by value, this have some unwanted effects.
`.size()` WifiSelectionActivity.cpp line 43, deep-copies the entire credentials vector (up to 8 structs, each with an SSID and a plaintext-password `std::string`) just to read a count.
`!getCredentials().empty()` in `WifiSelectionActivity.cpp` line 59.
`handleGetWifiNetworks` in `CrossPointWebServer.cpp` lune 1454. also copies all credentials into the web-server task's RAM on every GET.

Uri-Tauber (approved):
Approving, but fix the deep-copy regression before merging.

Uri-Tauber:
The problem you trying to solve here is very similar to the one #2305 solved. maybe you could reuse that pattern?

itsthisjustin:
> The problem you trying to solve here is very similar to the one #2305 solved. maybe you could reuse that pattern?

This is slightly different. The way our current system worked is it rewrote the whole file on every network connect. So if you changed your most recent network or failed to connect to the last one it rejiggers the whole file. This avoids writing the whole file every time.

PR #2837, opened 2026-08-02
fix: serialize display and SD on the shared SPI bus

Uri-Tauber:
## Summary

`HalStorage` serialized SD access with `storageMutex`, but `HalDisplay` took no lock. The e-ink panel and the SD card share one physical SPI bus, so a display refresh interleaving with a background SD transfer (indexing, thumbnail generation, OTA) could confuse `SdSpiCard`'s unsynchronised `m_spiActive` state machine and corrupt reads or trigger a FreeRTOS panic.

Add `HalSpiBus`: a recursive-mutex singleton with an RAII Lock. `HalDisplay` acquires it around every panel operation (begin, `displayBuffer`, `refreshDisplay`, `deepSleep`, `displayGrayBuffer`, `writeGrayscalePlaneStrip`). `HalStorage::StorageLock` acquires it as its outermost member so the order is always SPI-outer, storage-inner, giving a single consistent lock order across display and storage code. The mutex is recursive so a storage path that re-enters `StorageLock` (e.g. `HalFile::Impl close`) does not self-deadlock.

---

### AI Usage

### DIFF
--- a/lib/hal/HalDisplay.cpp
+++ b/lib/hal/HalDisplay.cpp
@@ -1,6 +1,8 @@
 #include <HalDisplay.h>
 #include <HalGPIO.h>
 
+#include "HalSpiBus.h"
+
 // Global HalDisplay instance
 HalDisplay display;
 
@@ -11,6 +13,8 @@ HalDisplay::HalDisplay() : einkDisplay(EPD_SCLK, EPD_MOSI, EPD_CS, EPD_DC, EPD_R
 HalDisplay::~HalDisplay() {}
 
 void HalDisplay::begin(bool seamless) {
+  HalSpiBus::Lock spiLock;
+
   // Set X3-specific panel mode before initializing.
   if (gpio.deviceIsX3()) {
     einkDisplay.setDisplayX3();
@@ -58,6 +62,8 @@ EInkDisplay::RefreshMode convertRefreshMode(HalDisplay::RefreshMode mode) {
 }
 
 void HalDisplay::displayBuffer(HalDisplay::RefreshMode mode, bool turnOffScreen) {
+  HalSpiBus::Lock spiLock;
+
   if (gpio.deviceIsX3() && mode == RefreshMode::HALF_REFRESH) {
     einkDisplay.requestResync(1);
   }
@@ -78,6 +84,8 @@ void HalDisplay::waitRefreshComplete() { einkDisplay.waitRefreshComplete(); }
 bool HalDisplay::supportsAsyncRefresh() const { return einkDisplay.supportsAsyncRefresh(); }
 
 void HalDisplay::refreshDisplay(HalDisplay::RefreshMode mode, bool turnOffScreen) {
+  HalSpiBus::Lock spiLock;
+
   if (gpio.deviceIsX3() && mode == RefreshMode::HALF_REFRESH) {
     einkDisplay.requestResync(1);
   }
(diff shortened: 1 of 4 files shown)


jpirnay:
Adopted this downstream in witchhunt-reader and it works well — thanks. One note from doing so, about the PR description rather than the code.

The description says the recursive bus mutex is what makes storage reentrance safe:

> The mutex is recursive so a storage path that re-enters `StorageLock` (e.g. `HalFile::Impl close`) does not self-deadlock.

The reentrant path is real — `openFileForRead` holds `StorageLock`, then assigns to the `HalFile&` out-param, and the prior `Impl`'s destructor takes the lock again — but what makes it safe is that `storageMutex` is *already* recursive, as documented in the `HalStorage()` constructor. `HalSpiBus`'s recursiveness doesn't contribute: they're two independent mutexes, and a nested `StorageLock` re-takes `storageMutex`, not the bus lock. If `storageMutex` were plain, a recursive bus mutex wouldn't prevent that deadlock.

No bug on your tree — the code is correct, the attribution in the description just points at the wrong mechanism. Worth tightening because it's load-bearing for anyone adopting this: it reads as "the bus lock handles storage reentrance," which is a property the bus lock doesn't have.

Concretely, it cost us a review cycle. Our `storageMutex` is plain (`xSemaphoreCreateMutex`), so we started from your sentence and had to work out whether adding the recursive bus lock had introduced a latent deadlock. It hadn't, but again for an unrelated reason: our `HalFile::Impl` is a bare struct with a defaulted destructor, so it destroys `FsFile` directly and calls SdFat's `close()` rather than the locking `HalFile::close()` — no reentry. Yours has an explicit `~Impl` that takes the lock, which is exactly why you need the recursive `storageMutex` and we don't.

Suggested rewording:

> `storageMutex` is already recursive (see the `HalStorage()` constructor), so the existing `StorageLock` reentrancy is unaffected. `HalSpiBus`'s mutex is recursive for the same reason — so any path that takes the bus lock twice cannot self-deadlock — not because it participates in storage reentrancy.

Two other things we changed downstream, both because our tree differs rather than anything wrong here, in case they're useful:

- **`HalStorage::begin()` takes the bus lock.** Your comment notes `begin()`/`ready()` are setup-only. In our tree the render task is already running by the time `begin()` is called, so SD init genuinely races the panel. Might be worth confirming that ordering holds on yours.
- **We time out instead of `portMAX_DELAY`.** 5s, then log and proceed unlocked. A single transaction is short, so anything past that is a lock-order mistake, and `portMAX_DELAY` turns it into a silent hang. This is a deliberate tradeoff, not a strict improvement — the degraded path does the exact unsynchronized access this PR prevents, so it's only defensible for something that should never fire. Reasonable to prefer the hang plus a watchdog reset.

lpla:
Hey @Uri-Tauber, had a quick look at this. The SPI-outer/storage-inner direction makes sense 👍

One thing to revisit while rebasing: `lib/hal/HalDisplay.cpp:74-146` still has a few `_bus` paths outside the shared lock, mainly async finish plus the grayscale base/copy/cleanup paths. Also, `lib/hal/HalSpiBus.cpp:5-28` currently continues unlocked if mutex setup/acquisition fails; I think this one should fail closed.

It’s conflicting with current `develop` rn. Want a hand rebasing and finishing the wrapper audit? Happy to help.

PR #2841, opened 2026-08-02
fix(i18n): fix Russian and Ukrainian strings that don't fit their widgets

winst0niuss:
Completes the Russian and Ukrainian translations, and fixes strings in both that did not fit their widgets. Everything here was checked on screen in the simulator, not just in the YAML.

## What's added

| Language | Keys filled in |
|---|---|
| Russian | 31 |
| Ukrainian | 33 |

These were falling back to English: the dictionary lookup flow and its error states, the KOSync sign-up flow and sync behaviour, end-of-book actions, saved-Wi-Fi progress messages, `SD root`, `Resume`, `Bookmark`, `Look Up` and assorted status strings.

Both files now cover every key in `english.yaml` except `STR_KEY_SHIFT`, `STR_KEY_MODE_SYMBOLS`, `STR_KEY_MODE_ABC` (and `STR_KOSYNC` for Ukrainian) — their values are identical to the English ones, so translating them would add noise without changing a pixel.

## What's fixed

Three of these were pre-existing strings that render broken today.

**Russian settings list** — three labels were truncated with an ellipsis:

ru settings

`STR_QUICK_RESUME_TIMEOUT`, `STR_REMOVE_READ_FROM_RECENTS` and `STR_MOVE_FINISHED_TO_READ` are now short enough to fit. The Ukrainian file already used shortened forms for the same three settings, which is what made the difference visible.

**Russian keyboard hint** — at 69 characters the line ran past both screen edges, and `очитисть` is a typo for `очистить`:

ru keyboard hint

**Ukrainian sync mode** — the value column truncated `Розумна синхронізація`; shortened to `Розумна`, which reads correctly under the `Режим синхронізації` label:

uk kosync

**Ukrainian network count** — `STR_NETWORKS_FOUND` was `мереж %zu`, rendering as a bare noun and a number with no verb:

uk wifi

Also fixes a typo in `STR_LONG_PRESS_BEHAVIOR`: `довгому настику` → `довгому натиску` (`настик` is not a word; neighbouring strings already use `натиск`).

## How it was checked

Built with the simulator and walked through both languages on 11 screens each: home, all four settings tabs, text settings, language picker, OPDS server list and server card, the keyboard, Wi-Fi, and the KOSync screen — the tightest case, where a setting name and its value share one row. No truncation left in either language.

Not verified on screen: the KOSync sign-up popups, the saved-Wi-Fi progress messages and the end-of-book screen, which need a live network session and an open book. Their strings have margin, but I did not see them rendered.

New translations were kept at or below the English length wherever the string is a menu label or a setting value, per `docs/translators.md`.


winst0niuss:
I'm a native speaker of both Russian and Ukrainian, and I don't agree with the CodeRabbit suggestions here — the current wording is what I'd keep.

On `STR_QUICK_RESUME_TIMEOUT`: it labels an ON/OFF toggle, not the mode. The mode itself keeps its full name ("Быстрое возобновление"), shown as a Sleep Screen value two rows above in the same tab, so "Возобновление по таймауту" is unambiguous in context. The literal form is what upstream has today and it renders truncated ("Быстрое возобновление по тайма…"); keeping the adjective would force "возобн.", which reads worse in Russian than the clean phrase.

On the `←` suggestion: no built-in font in `lib/EpdFont/builtinFonts/` contains U+2190 — there are no U+21xx glyphs at all — so it would render as a replacement glyph. That is why the translations use surrogates (`<—` in Russian, `<--` in Ukrainian, spelled-out key names in Spanish and Polish). Worth noting the German file already contains a real `←`, which is presumably broken on screen today.

muhas:
Nice optimize russian translate

merge to  [chore/update-translations](https://github.com/crosspoint-reader/crosspoint-reader/tree/chore/update-translations)

I'm too translated resume as "Читать" 😂


---
Вас не бесит слово "Управление" не влазящее в блок кнопки?  Каждый раз хочется его заменить на "Кнопки", "Клавиши" или что-то подобное что-бы в теме лира влезло, но вроде давно уже так переведено, никто не трогает и самому боязно, так сказать, нарушать традиции. Переводил из-за лёгкого приступа перфекционизма, сам английский плохо знаю и рад что есть кто-то ещё взявшийся за это дело да ещё и за клавиатуру 👍

winst0niuss:
@muhas, thanks for the suggestion. 
Renamed "Управление" to "Кнопки" — it reads better and looks better on screen.

winst0niuss:
Closing in favour of two smaller PRs, one per language, both targeting `chore/update-translations` as @muhas suggested:

- Russian: #2851
- Ukrainian: #2852

Rebasing onto `chore/update-translations` showed that Russian had been completed there independently — both files carry the same 424 keys, so the "fills gaps" framing of this PR no longer holds. What is left is strings that do not fit their widgets, plus a couple of typos, and that reads better split per language.

@muhas thanks for pointing at the right base branch. Where our wordings differed I kept yours for `STR_LOOKUP`, `STR_EOB_HOME`, `STR_INVALID_BMP_FILE` and `STR_FINDING_SAVED_WIFI`; the rest are changed because the current string overflows its row, with measured widths in #2851.

PR #2852, opened 2026-08-03
chore: update Ukrainian translation

winst0niuss:
Split out of #2841 so Russian and Ukrainian can be reviewed separately. Russian is in its companion PR.

The key set is unchanged — 427 keys on both this branch and `chore/update-translations`, and `STR_KOSYNC` is preserved. This fixes strings that render broken on screen.

## What is fixed

**Sync mode** — the value column truncated `Розумна синхронізація`; shortened to `Розумна`, which reads correctly under the `Режим синхронізації` label:

uk kosync

**Network count** — `STR_NETWORKS_FOUND` was `мереж %zu`, rendering as a bare noun and a number with no verb:

uk wifi

**Typo** — `STR_LONG_PRESS_BEHAVIOR` had `довгому настику`; `настик` is not a word, and neighbouring strings already use `натиск`.

**Dictionary popups** — reworded to match the Russian PR's reasoning: `drawPopup` neither wraps nor truncates, so an over-long string leaves the screen.

| Key | Was | Now |
|---|---|---|
| `STR_DICT_READ_FAILED` | Не вдалося прочитати визначення | Визначення не прочитано |
| `STR_DICT_DECOMPRESS_ERROR` | Не вдалося розпакувати визначення | Визначення не розпаковано |

**Rows that overflow** — `STR_BACK_SHORT_TO_FILE_BROWSER`, `STR_SYNC_BEHAVIOR`, `STR_ASK_EVERY_TIME` and a few others shortened so the label and its value fit one row.

`STR_EOB_HOME` follows the Russian PR and keeps the base branch's `Додому`.

## Checked

Walked Ukrainian through the simulator on 11 screens: home, all four settings tabs, text settings, language picker, OPDS server list and server card, the keyboard, Wi-Fi, and the KOSync screen. `python scripts/gen_i18n.py` regenerates cleanly: 31 languages, 427 keys.

Not verified on screen: the KOSync sign-up popups, saved-Wi-Fi progress messages and the end-of-book screen, which need a live network session and an open book.


mirus-ua:
Except for a typo with `настик,` I don't think that this AI-driven translation is correct, especially since it shouldn't be aligned with Russian. translation 

@KymAndriy, you as a corrector of the OG translation, WDYT?

KymAndriy:
I need to check this all myself. Basically translation looks just ok.
Some cases use passive language that makes filling like people don't speak like that.

Alignment with russian language isn't correct at all because semantic and the others rules dramatically different between those languages.
I still wait for my device to make a real tests.

@Uri-Tauber please wait untill end of the wee with 1.5.0 release (in case you have a release plan), i have to made changes to ukrainian translation

Uri-Tauber:
@KymAndriy Thx for letting me know. I planned to release 1.5.0 tomorrow, but another day is fine. Do you think you could finished your changes until friday?

KymAndriy:
I will try
Thanks @Uri-Tauber <3

PR #752, opened 2026-02-07
feat: Connect to last wifi by default

el:
## Summary

* **What is the goal of this PR?** 

Use last connected network as default

* **What changes are included?**

- Refactor how an action type of Settings are handled
- Add a new System Settings option → Network
- Add the ability to forget a network in the Network Selection Screen
- Add the ability to Refresh network list
- Save the last connected network SSID
- Use the last connection whenever network is needed (OPDS, Koreader sync, update etc)

## Additional Context

* Add any other information that might be helpful for the reviewer (e.g., performance implications, potential risks, 
  specific areas to focus on).

IMG_6504
IMG_6503

https://github.com/user-attachments/assets/95bf34a8-44ce-4279-8cd8-f78524ce745b

---

### AI Usage

Did you use AI tools to help write this code? _** PARTIALLY: I wrote most of it but I also used Gemini as assist.

PR #2858, opened 2026-08-03
feat(keyboard): user-selectable layout set with a language key

winst0niuss:
> Ready: [Free-Ink/freeink-sdk#69](https://github.com/Free-Ink/freeink-sdk/pull/69) has landed and the submodule pointer tracks upstream `main` again.

Firmware side of multilingual keyboard support, following the direction @Uri-Tauber set in #2828: layouts belong in the SDK, and switching should be general rather than one language paired with English.

Since review, freeink-sdk#69 also gives ף its own key in the Hebrew top row instead of a long-press off פ, and turns the language key into a globe glyph. The firmware half of that is small: it no longer supplies a label for the key.

## What a user sees

With a Russian interface and no configuration, the keyboard offers ЙЦУКЕН and QWERTY, and a key in the bottom row switches between them. With an English interface and no configuration, nothing changes at all — the set collapses to one layout and the key is not drawn.

**Settings → System → Keyboard Layouts** lists every layout the SDK ships and toggles each on or off. The language key then cycles through whatever is enabled; with the default two it behaves as a plain toggle.

That second part goes beyond the "UI language + English" pairing discussed in #2828, and deliberately. The language someone reads in is not necessarily the language of their interface: a Russian UI with Spanish books wants a Spanish layout for OPDS searches and filenames. Pairing cannot express that, so the set is the user's to pick. The rule that an Arabic user has no use for a Russian layout still holds — nothing extra is shown unless asked for.

## How it is stored

`CrossPointSettings::keyboardLayouts`, a `uint16_t` bit mask, one bit per `KeyboardLayoutId`. Zero means unconfigured and is resolved to the derived default at read time rather than being written out, so a later UI language change still brings its layout along.

Layout tables are `const` and live in flash whether or not their bit is set, so enabling a layout costs no RAM — only the bit changes.

The mask is filtered against the layouts the build actually has. It comes from a user-editable file and survives firmware downgrades, so it can carry bits for layouts that are not there; unfiltered, such a mask would read as "configured" while enabling nothing, leaving the keyboard with no letters.

## Notes for review

`KeyboardLayoutSet` (new, `src/activities/util/`) owns the default and the cycle order. The key itself needs nothing from it: the SDK draws a globe there, so no label crosses the boundary.

The settings screen writes once on exit, and only if something actually changed. Writing per keypress would burn SPIFFS erase cycles; writing unconditionally would persist the derived default on a mere visit and freeze the set.

`InteractionBuffer` goes 48 → 56 because Cyrillic rows run 12/11/11 against Latin's 10/9/9, landing on exactly 48 hit rects with no headroom. `Frame<>` is typed by the same parameter and moves with it.

`STR_KEYBOARD_LAYOUTS` is the only new string. English, Russian and Ukrainian are filled in; the rest fall back to English.

## Testing

Exercised in the desktop simulator across all nine layouts: switching in and out of each, shift on each, the settings screen, and the default set with a Russian and an English UI. Host test suite passes (129/129). `pio check` with the CI flags reports no defects.

Verified by script over the SDK tables: every layout reachable with the language key carries it in every variant (an early version stranded the user on French, German and Spanish), and upper/lower layers correspond one-to-one.

Right-to-left behaviour of the Hebrew layout in the text field — cursor placement, deletion — has only been seen in the simulator. @Uri-Tauber has since exercised the Hebrew keyboard on a Sticky.


winst0niuss:
Good catch, fixed in 76fcc3d2 — the layout screen now gets a no-op handler and owns its own write.

Worth noting for whoever picks this up later: the same double-save applies to the language, status bar and button remap screens. They all save themselves and also get the shared handler, so each of them writes twice. That predates this branch — the handler comes from #752 back in February — so I have left them alone rather than widen this PR.

winst0niuss:
Fixed in 7bba1096 — took the second option and gave each layout its own persistent bit in the table rather than deriving it from the enum.

The mask lives in `settings.json` and outlives firmware updates, while the ordering of `KeyboardLayoutId` belongs to the SDK. Deriving one from the other meant a layout inserted mid-enum would silently reinterpret every saved mask: enable Cyrillic, come back to Hebrew after an update. The bit indices are now documented as append-only, and `static_assert`s reject a duplicate bit, a bit that does not fit the `uint16_t`, and a table longer than 16 entries.

Uri-Tauber:
@winst0niuss Thanks for putting this together. I hope I will have next week the time to review.

winst0niuss:
**Why one Latin layout always stays enabled**

Not covered in the description above, and the one thing on the screen that can surprise: a row whose removal would leave the set without a Latin layout reads "Default" and does not toggle.

The symbol layers are shared across layouts and hold only digits and punctuation, so a Cyrillic-only set cannot type a Wi-Fi passphrase or an OPDS URL. `isLocked()` refuses the last Latin layout; `enabled()` puts English back if a hand-edited settings file arrives without one. AZERTY, QWERTZ and Spanish count as Latin and can stand alone; Cyrillic and Hebrew cannot.

This does not make English the default keyboard — the starting layout still follows the UI language, so a Russian UI opens on ЙЦУКЕН. Lifting the restriction belongs in the SDK (a Latin row in the symbol layer), not here.

winst0niuss:
@Uri-Tauber, pls check it when you can.
I tested it and used it with both rus and eng keyboard layouts, and didn't notice any problems

Uri-Tauber:
Ohh, sorry. I completely forgot about this!
I'm juggling between so many things!

Uri-Tauber:
I tested on my Sticky, and its working very well for English+Hebrew. Awesome work!
Two small things: 
a. I think its very confusing to have ף as long press on פ. Would make more sense to remove ' from the left and move the entire line one step left, leaving space for ף.
b. I would prefer if the **change lang** key would have [globe](https://lucide.dev/icons/globe) icon, instead of **ENG**/**HEB**. FreeinkUI let you add icons.

winst0niuss:
@Uri-Tauber, add actual keyboard screens on [freeink-sdk/pull/69](https://github.com/Free-Ink/freeink-sdk/pull/69)

winst0niuss:
@Uri-Tauber, check please.
 [freeink-sdk/pull/69](https://github.com/Free-Ink/freeink-sdk/pull/69) merged and its ready to review

Uri-Tauber:
I cleaned the code a bit. There is a RTL small bug I want to solve before merging, I hope I could do it tomorrow.

Can you test it again with my cleanup? It working well as far as I can tell, but another pair of eyes would be better.

winst0niuss:
I tested it. The keyboard, the layouts screen and the globe key all behave as before. 
Looks good to me.

Uri-Tauber (approved):
LGTM.

PR #2860, opened 2026-08-03
fix(reader): let the end-of-book menu be tapped

winst0niuss:
The end-of-book menu lists the next books in the folder plus a Home entry, and on a touch device those rows look tappable. They are not: the component has no touch handling at all — no hit test, no reference to `mappedInput`. Only the buttons move the selection.

## What changed

`handleMenuInput` now checks touch before buttons, following the same idiom as every other list in the firmware: a tap acts on the row, a touch-down only moves the highlight.

The list geometry moves out of `render()` into a shared `listGeometry()`. Computing hit rects separately from what is drawn is exactly how the two drift apart, and this menu's layout derives from the safe area, so it is not a constant that can be duplicated safely.

Both callers pass the renderer through — the EPUB reader and the XTC reader.

## Scope

This is the same gap the small-panel work in #2802 describes for lists generally. Of the firmware's list screens, four had no shared hit test: this one, `EpubReaderBookmarksActivity`, `SettingsActivity` (which has its own equivalent) and `ButtonRemapActivity` (which is about physical buttons and correctly has none). Bookmarks can get the same treatment; kept out of this PR to keep it to one screen.

## Testing

Exercised in the desktop simulator on an X4 Pro profile: tapping a suggested book opens it, tapping Home leaves the reader, touch-and-drag moves the highlight without acting, and the buttons behave exactly as before. Host test suite passes (129/129), `pio check` with the CI flags reports no defects.

Not tested on hardware.

### DIFF
--- a/src/activities/reader/EndOfBookOptions.cpp
+++ b/src/activities/reader/EndOfBookOptions.cpp
@@ -44,7 +44,42 @@ std::string EndOfBookOptions::fullPath(const size_t index) const {
   return folder == "/" ? "/" + names[index] : folder + "/" + names[index];
 }
 
-EndOfBookOptions::Action EndOfBookOptions::handleMenuInput(const MappedInputManager& input, std::string* openPath) {
+void EndOfBookOptions::listGeometry(const GfxRenderer& renderer, int& listTop, int& listHeight) const {
+  const auto& metrics = UITheme::getInstance().getMetrics();
+  const Rect safe = UITheme::getInstance().getScreenSafeArea(renderer, true, false);
+  const int titleY = safe.y + safe.height / 8;
+  const int subtitleY = titleY + renderer.getLineHeight(UI_12_FONT_ID) + metrics.verticalSpacing;
+  listTop = subtitleY + renderer.getLineHeight(UI_10_FONT_ID) + metrics.verticalSpacing * 2;
+  listHeight = safe.y + safe.height - listTop - metrics.verticalSpacing;
+}
+
+EndOfBookOptions::Action EndOfBookOptions::handleMenuInput(const MappedInputManager& input, const GfxRenderer& renderer,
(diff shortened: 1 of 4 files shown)


Uri-Tauber:
@itsthisjustin What do you think?

itsthisjustin:
> @itsthisjustin What do you think?

I missed this in the existing release but I'll fix this in the upcoming touch ui branch

Uri-Tauber:
> I missed this in the existing release but I'll fix this in the upcoming touch ui branch

What bothering me is the touch handling should live in some central place (probably in the SDK), and be automatically useable from every screen. Adding touch handling to every screen will be a maintenance burdon.

itsthisjustin:
> > I missed this in the existing release but I'll fix this in the upcoming touch ui branch
> 
> What bothering me is the touch handling should live in some central place (probably in the SDK), and be automatically useable from every screen. Adding touch handling to every screen will be a maintenance burdon.

That IS true when using freeink components. However all the handrolled stuff needs to have stuff manually added.

winst0niuss:
Superseded by #2957.

That PR converted the lists to FreeInkUI, and `EndOfBookOptions` came along with it: it now derives from `UiAppHost`, its list is declared with `props.action = ACTION_ROW` and `props.inputMask = fui::InputTouch` (so FreeInkUI registers a hit rect per row at render time), and `handleMenuInput()` starts with `routeTouch()`, dispatching through `onRowEvent`. Tapping a suggestion opens it and tapping "Home" leaves the reader, so the gap this PR was about is closed on current `develop` (618608e8).

One detail from this PR isn't carried over: touch-down moving the highlight without acting. FreeInkUI shows a tap flash instead, which is consistent across every list now, so it doesn't seem worth a follow-up.

Closing.

Issue #2656, opened 2026-07-20
Allow users to select where a word starts and ends for dictionary feature.


brianhuster:
### Problem

In English there are a lot of idioms that are not easy to understand at first sight by those unfamiliar with them, even if you know the meaning of its individual words. So there will be a demand for looking up these phrases in dictionaries. 

There are a lot of languages that don't use space to separate words, like CJK languages, Vietnamese, Thai, Lao, Khmer, Burmese, Tibetan, etc. As a result, Dictionary feature cannot detect a word correctly in these languages, e.g for Chinese and Japanese it highlights a character instead of a word, and for Vietnamese it highlights a syllable instead of a word. This makes the feature not very useful for these languages. 

### Suggested behaviors

My suggestion is making dictionary feature has 2 options: "lookup word" and "lookup phrase". "Lookup words" will work the same as current Dictionary feature. "Lookup phrase" will prompt users to select the "word" that starts the phrase, and then from the position that start the phrase it will prompt users to select the "word" that ends the phrase. In this paragraph, "word" means an English word, a Chinese character or a Vietnamese syllable, etc. Finally it shows the result from StarDict dictionary for the chosen phrase.

badbob001:
How about instead of pressing 'select' to instantly lookup the highlighted word, have the first 'select' mark the first word for lookup. Then the cursor keys allows you to highlight the last word. Now pressing select a second time will lookup the phrase between the first and last words.

Pressing select twice without moving the cursor in between will just lookup a single word.

For languages without spaces, have a lookup option that lets you select by character instead of word. Does the 'select' button support long-presses? A long-press can toggle the cursor between word and character mode.

This start/end word selection interface could be used for a feature to select words for export to a quotes file.

brianhuster:
> How about instead of pressing 'select' to instantly lookup the highlighted word, have the first 'select' mark the first word for lookup. Then the cursor keys allows you to highlight the last word. Now pressing select a second time will lookup the phrase between the first and last words.

> Pressing select twice without moving the cursor in between will just lookup a single word.

Sounds like good idea to me

badbob001:
I just saw a video of variant CrossInk, which has a highlight/save clipping feature where the selection process works exactly as I described it.
https://www.youtube.com/watch?v=NAaglgBNOlA

brianhuster:
@Uri-Tauber @rmmh what do you think of badbob001's idea?

Uri-Tauber:
[...]

PR #2877, opened 2026-08-05
fix: trim curly quotes and dashes in dictionary lookup

Uri-Tauber:
# Summary

`cleanWord` treated any byte >= 0x80 as a word byte to preserve accented letters, but General Punctuation (U+2000-U+206F: curly quotes, dashes) is UTF-8 E2 80/81 xx, all >= 0x80. Trailing/leading quotes were kept, so EPUB text like `garage."` or `possible,"` failed to match a headword.

Extend the edge-trim loops to step over leading/trailing 3-byte General Punctuation codepoints, matching the range `isSelectableToken` already special-cases in `DictionaryWordSelectActivity`.

---

### AI Usage

### DIFF
--- a/src/util/Dictionary.cpp
+++ b/src/util/Dictionary.cpp
@@ -392,14 +392,20 @@ std::string Dictionary::cleanWord(const char* word) {
   // are all >= 0x80, so isWordByte keeps them; strip those 3-byte codepoints
   // from the edges too, or EPUB text like garage.” never matches a headword.
   while (start < end) {
-    if (!isWordByte(b[start])) start++;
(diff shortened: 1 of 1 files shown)


Uri-Tauber:
should fix https://github.com/crosspoint-reader/crosspoint-reader/issues/2656#issuecomment-5194377704 and https://github.com/crosspoint-reader/crosspoint-reader/discussions/2693#discussioncomment-17908152

itsthisjustin:
Does this fix hyphens or no?

Uri-Tauber:
Unrelated. Other bug two user complained about in dictionary selection.

itsthisjustin (approved):

PR #2884, opened 2026-08-06
feat: Chunked allocation for glyph bitmaps

itsthisjustin:
# Chunked mini glyph-bitmap cache

Splits the per-page SD-font glyph-bitmap cache (the "mini bitmap" arena built during prewarm) from a single contiguous allocation into fixed 4 KB chunks.

## Why

A whole page's glyph bitmaps can total tens of KB. The old code allocated that as **one contiguous block** sized to the page.

Chunking assembles the arena from 4 KB blocks the allocator can actually hand out. Each glyph's bitmap is placed wholly within one chunk (never straddling a boundary); a glyph is addressed by a virtual offset that maps to `(chunk index, offset-in-chunk)`.

## Mini glyph-bitmap cache: contiguous (before) vs chunked (after)

Latin figures are grounded in device logs + font geometry.

| Metric | Latin (measured/derived) | CJK (estimated) |
|---|---|---|
| Unique glyphs per page | ~60–100 | ~250–450 (cap 512) |
| Avg glyph bitmap size | ~140 B | ~50–200 B |
| **Per-page arena (per style)** | **~12–17 KB** | **~30–60 KB** |
| Peak single allocation — **before** | up to ~17 KB (one block) | up to ~60 KB (one block) |
| Peak single allocation — **after** | **4 KB** (fixed) | **4 KB** (fixed) |
| Reduction in peak allocation | ~4× | **~8–15×** |
| Largest free block for fast path — **before** | ≥ ~17 KB | ≥ ~30–60 KB |
| Largest free block for fast path — **after** | ≥ 4 KB | ≥ 4 KB |
| Chunks used per page — **after** | 1–5 | ~8–15 |
| On allocation failure (fragmented heap) | falls back to per-glyph SD reads (~60–100 glyphs) | falls back to per-glyph SD reads (**300+ glyphs**) |

## Render-cost impact (same font/page, fast path vs fragmentation fallback)

| | Fast path (RAM cache) | Fallback (per-glyph SD) |
|---|---|---|
| `bw_render`, per page | ~30 ms *(measured)* | ~4,200 ms *(measured, pressured build)* |
| Grayscale plane render | ~1 s | ~13 s *(measured, pressured build)* |
| Relative cost | 1× | **~100×** |

Fallback figures were measured on a memory-pressured build (BLE resident), not
reproduced standalone. 

## Cost of chunking when memory is plentiful (no regression)

| Metric | Value |
|---|---|
| Extra render time at full heap | none — `bw_render` 23–39 ms, identical to contiguous *(measured)* |
| Memory overhead | ≤ one extra 4 KB block per ~4 KB of bitmap |
| Boundary padding | ≤ ~43 B per chunk *(measured: `style3 4648 need → 4691 used`)* |

### DIFF
--- a/lib/EpdFont/SdCardFont.cpp
+++ b/lib/EpdFont/SdCardFont.cpp
@@ -99,8 +99,11 @@ void SdCardFont::freeStyleMiniData(PerStyle& s) {
   s.miniIntervals = nullptr;
   delete[] s.miniGlyphs;
   s.miniGlyphs = nullptr;
-  delete[] s.miniBitmap;
-  s.miniBitmap = nullptr;
+  for (auto& chunk : s.miniBitmapChunks) {
+    delete[] chunk;
+    chunk = nullptr;
+  }
+  s.miniBitmapChunkCount = 0;
   s.miniIntervalCount = 0;
   s.miniGlyphCount = 0;
   s.miniIntervalCapacity = 0;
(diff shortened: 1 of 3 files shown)


lib/EpdFont/SdCardFont.cpp
```diff
@@ -1010,30 +1013,66 @@ int SdCardFont::prewarmStyle(uint8_t styleIdx, const uint32_t* codepoints, uint3
...
-        glyph.dataOffset = miniBitmapOffset;
+      if (len == 0) {
+        glyph.dataOffset = span;
         continue;
       }
+      if (len > MINI_BM_CHUNK_SIZE) {
+        // A single glyph larger than one chunk cannot be placed. Real fonts never
+        // hit this (a 4 KB 2bpp glyph would be enormous); bail rather than corrupt.
+        LOG_ERR("SDCF", "Prewarm: glyph %u B exceeds chunk %u B (style %u)", len, MINI_BM_CHUNK_SIZE, styleIdx);
+        file.close();
```
Uri-Tauber, line 1041: I believe we don't need to manually close, since we using `DESTRUCTOR_CLOSES_FILE` flag.
Also applied to two other places here.

Uri-Tauber:
Do you have some reproducible test case in which `develop` failed to allocate, and here its working?

itsthisjustin:
> Do you have some reproducible test case in which `develop` failed to allocate, and here its working?

CJK can and this happens when you use BLE due to heap pressure

Uri-Tauber:
Do you mind waiting with this after the release?
I want more testing.

itsthisjustin:
> Do you mind waiting with this after the release?
> I want more testing.

Yep totally not priority

PR #2886, opened 2026-08-06
feat: add the CLX1 on-disk index format and its reader

oreglio:
**Part 2/5 of the Library view** — design and measurements in #2878. Stacked on part 1 (the diff shows both until it merges; this part's own changes are the last commit).

The on-disk index (`CLX1`) and its read side:

- **Fixed 128-byte records** in folded-title order: record *k* is always at `recordStart + 128k`, so every sort order is O(1) seeks per row and browsing RAM stays ~440 bytes regardless of library size. Author/date orders are u16 permutation arrays.
- **Strict validation**: the index is treated as untrusted input — magic/version checks, self-size vs real file size (a power-cut build cannot pass), section-layout recomputation, length caps so forged headers cannot wrap the arithmetic. Per-record lengths are clamped at the single point records enter the program.
- **Bounded reads** for name / author / title / path, with `docs/file-formats.md` documenting the full layout as the contract.

**How to review:** `docs/file-formats.md` first (the contract), then `LibraryFormat.h` (layout + validation), then `LibraryIndexFile` (reads). Host tests: `test/library_format` (17 tests) — layout arithmetic, truncation both directions, version rejection, flag round-trips.

**Series:** #2878 (tracking) → text core → **this** → index builder → shelf activity → wiring.

### AI Usage
YES — see #2878.


Uri-Tauber:
@oreglio I need you to rebase on top of `feat/library-view`, so I could see what was added here. I think the things you removed/fixed in #2885 are still here.
Apart from that, see my response in #2885.

Thanks for your work.

Uri-Tauber:
@oreglio Is this still a draft on purpose?

oreglio:
@Uri-Tauber Not on purpose, thanks for the nudge. #2886 and #2887 are ready for review now: both rebased on top of feat/library-view as you asked, so the parts #2885 already covered are gone from their diffs.

I'm deliberately keeping #2888 and #2889 in draft a little longer: following your freeInk-components direction, I've already converted the shelf to fui::list on a shared tab-list base on my fork, and I'm validating it on hardware right now. I'd rather update #2888 with the converted shelf than have you review the pre-conversion UI twice.

PR #2887, opened 2026-08-06
feat: add the library index builder and EPUB metadata reader

oreglio:
## Summary

This is the index-builder part of the Library stack. It adds the persistent `CLIX` index, bounded SD-card walking, stable ordering, optional EPUB metadata, and recovery from interrupted rebuilds.

The round-two review fixes are included here rather than deferred to the shelf PR:

- the builder services the watchdog in every long phase; the callback is now only for progress observation and cancellation;
- optional metadata reuses `Epub`/`ContentOpfParser` in a metadata-only mode, with one `ZipFile` open per book and an early stop at `</metadata>` before `<manifest>` (no spine, TOC, CSS, section cache, or `.items.bin` work);
- author text is accumulated across Expat callbacks before separators are inserted, and OPF text is normalized consistently;
- every staging write and reread is exact-checked, including the folder stage, so a truncated stage cannot be padded into an apparently valid index;
- per-directory duplicate tracking is a fixed, fallible 8 KiB allocation instead of a growing vector; an allocation failure or a directory beyond 1,024 remembered identities is persisted as `DEDUP_DEGRADED` instead of aborting;
- sort scratch is bounded to 512 books (7 KiB for keys and ranks); larger libraries remain usable in explicit degraded walk order;
- root-level EPUB paths are reconstructed correctly;
- installation is transactional: an interrupted replacement can recover the previous valid index from `.bak`.

The on-card format and all degraded-state flags are documented in `docs/file-formats.md`.

## Verification completed

- Host suite: 191/191 tests passed, including metadata-only early stop, split Expat callbacks, root-level paths, staging truncation, and install recovery.
- `pio run -e default` (ESP32-C3): passed.
- `pio run -e sticky` (ESP32-S3): passed.
- `pio check -e default --fail-on-defect low --fail-on-defect medium --fail-on-defect high`: no defects found.

`x4-pro` is not a configured PlatformIO environment in this checkout, so it is not claimed as tested.

## Hardware follow-up

Before requesting another review pass I will post X3 measurements for rebuild duration, `freeHeap`, and `maxAlloc` on a large card, and verify that metadata rebuilds create no per-book `.items.bin` files. The earlier timing/heap numbers have deliberately been removed because the parser and scratch-memory paths changed during review.


Uri-Tauber:
@oreglio Can you rebase?
Thx.

oreglio:
Done, rebased on the merged #2886, single commit, ready for review.

hamaluik (commented):
Ran into a few things here while building the series branch on top of this code — sending them while the PR is still open in case they're useful. Push back on any of them.

lib/LibraryIndex/LibraryBuilder.cpp
```diff
@@ -0,0 +1,1005 @@
...
+      // book get an id and the ids stay dense.
+      myFolderId = st.folderId++;
+      const uint8_t pathLen = static_cast<uint8_t>(std::min<size_t>(path.size(), 255));
+      st.folders.write(&pathLen, 1);
+      st.folders.write(reinterpret_cast<const uint8_t*>(path.data()), pathLen);
+      st.folderBytes += 1u + pathLen;
+      folderEmitted = true;
+      if (depth == 0) st.booksAtRoot = true;
+    }
+    stageRecord(st, name, size, myFolderId, joinPath(path, name));
```
hamaluik: I think this can starve the watchdog on flat folders.

`onProgress()` only runs once per directory, but with metadata enabled each book can spend a while in `stageRecord()`. A large flat folder can then go past the TWDT timeout with nothing feeding the watchdog and since the missing index retriggers the rebuild on the next entry, it can reboot-loop.

Would it make sense to call `onProgress()` per staged book, or at least yield inside the entry loop? Something like:

```suggestion
    stageRecord(st, name, size, myFolderId, joinPath(path, name));
    if (st.onProgress != nullptr && !st.onProgress(st.books, path.c_str(), st.progressCtx)) {
      st.aborted = true;
      return;
    }
```

(The per-directory call after the loop can stay for directories that hold no books.)

lib/LibraryIndex/LibraryBuilder.cpp
```diff
@@ -0,0 +1,1005 @@
...
+    HalFile folders;
+    if (Storage.openFileForRead("LIBIDX", folderStagePath, folders)) {
+      uint8_t buf[256];
+      // read() returns int: a -1 error must fail the emit, not wrap into a
+      // huge unsigned length.
+      int got = 0;
+      while ((got = folders.read(buf, sizeof(buf))) > 0) put(buf, static_cast<size_t>(got));
+      if (got < 0) writeFailed = true;
+      folders.close();
+    }
```
hamaluik, line 676: It looks like a failed open of `folderStagePath` is silently ignored here.

We still `padTo(header.recordStart)` and publish an index whose folder section is all zeroes. `readPath()` then fails for every book, but `selfSize` still matches, so validation passes and the index never self-repairs.

```suggestion
    } else {
      writeFailed = true;
    }
```

lib/LibraryIndex/LibraryBuilder.cpp
```diff
@@ -0,0 +1,1005 @@
...
+    Storage.remove(NEW_PATH);
+    return false;
+  }
+  StagedEntry& entry = staged[0];
+  StagedEntry& canonical = staged[1];
+
+  uint32_t nameCursor = 0;
+  for (uint16_t i = 0; i < n; i++) {
+    stage.seekSet(static_cast<uint64_t>(order[i]) * STAGE_STRIDE);
+    stage.read(reinterpret_cast<uint8_t*>(&entry), STAGE_STRIDE);
```
hamaluik: These loops don't check `stage.read()`.

On a read failure, `entry` keeps the previous book's bytes, so we emit a duplicate row and `written == selfSize` can still be true so the corrupt index would pass. The spelling-vote reads above already check the same kind of read; extending that here and setting `writeFailed` on a short read would let the guard handle it. For the record loop:

```suggestion
    if (stage.read(reinterpret_cast<uint8_t*>(&entry), STAGE_STRIDE) != static_cast<int>(STAGE_STRIDE)) {
      writeFailed = true;
      break;
    }
```

Same treatment for the `canonical` read below it and the two reads in the name-blob loop.

lib/LibraryIndex/LibraryBuilder.cpp
```diff
@@ -0,0 +1,1005 @@
...
+  auto resolvedFirstSeen = makeUniqueNoThrow<uint16_t[]>(st.books == 0 ? 1 : st.books);
+  if (!resolvedFirstSeen) {
+    LOG_ERR("LIBIDX", "firstSeen array alloc failed");
+    Storage.remove(STAGE_PATH);
+    Storage.remove(folderStagePath.c_str());
+    return false;
+  }
+  if (!st.aborted && st.books > 0) {
+    HalFile read;
+    if (Storage.openFileForRead("LIBIDX", STAGE_PATH, read)) {
```
hamaluik: A failed `firstSeen` reconciliation silently falls back to zero.

If opening `STAGE_PATH` fails or a later read is short, affected books will have `resolvedFirstSeen == 0`. That breaks "Recently added", and the next rebuild would read those zeroes back as prior truth corrupting things.

Failing the build but keeping the previous index seems like the safer answer, unless I'm missing a reason zero is preferable?

lib/LibraryIndex/LibraryBuilder.cpp
```diff
@@ -0,0 +1,1005 @@
...
+    st.prior[priorIndex].matched = true;
+    entry.record.firstSeen = st.prior[priorIndex].firstSeen;
+    st.reused++;
+  } else {
+    // Might be a rename rather than a new book; resolved after the walk, when
+    // the set of genuinely unmatched previous entries is known.
+    entry.record.firstSeen = FIRST_SEEN_UNRESOLVED;
+  }
+  entry.record.folderId = folderId;
+  entry.record.nameLen = static_cast<uint8_t>(std::min<size_t>(name.size(), STAGE_NAME_BYTES));
```
hamaluik: These length clamps can stage books that can never be opened.

`nameLen` and `pathLen` both truncate at 255 bytes with no skip and the name cut is not UTF-8-safe. A long Unicode filename or deep path would render on the shelf but reconstruct to an incorrect path.

Skipping these entries and counting them in `unreadableSkipped` might fit better than staging rows that cannot open?

lib/LibraryIndex/LibraryMeta.cpp
```diff
@@ -0,0 +1,242 @@
...
+  return !out.empty();
+}
+
+}  // namespace
+
+std::string opfPathFromContainer(const std::string& containerXml) {
+  // <rootfile full-path="OEBPS/content.opf" .../>. Scanned for rather than
+  // parsed: the attribute is unambiguous, and a container.xml with anything else
+  // wrong should still yield the path.
+  const size_t attr = containerXml.find("full-path=");
```
hamaluik: This takes the first `full-path=` anywhere in `container.xml`.

That can be a commented-out `rootfile` or a non-OPF alternative listed first. The reader resolves the OPF through `ContainerParser` so the shelf and the reader can end up reading different package documents for the same book.

Skipping comments and anchoring on a real OPF `rootfile` here too would keep the two sides agreeing.

lib/LibraryIndex/LibraryMeta.cpp
```diff
@@ -0,0 +1,242 @@
...
+      out.push_back('<');
+    } else if (name == "gt") {
+      out.push_back('>');
+    } else if (name == "quot") {
+      out.push_back('"');
+    } else if (name == "apos" || name == "#39" || name == "#x27" || name == "#8217" || name == "#x2019") {
+      out.push_back('\'');
+    } else if (name == "#160" || name == "#xa0" || name == "nbsp") {
+      out.push_back(' ');
+    } else {
```
hamaluik: This doesn't decode general numeric character references.

A perfectly valid title like `&#201;mile Zola` renders raw on the shelf and the raw entity text also pollutes things so searching `emile` would miss it. The reader-side XML path decodes these fine so the two screens disagree about the same book.

I think we could add a general `&#NNN;` / `&#xNN;` branch via `utf8AppendCodepoint` (the earlier explicit `#39`/`#160` cases still win first, which keeps their ASCII mappings):

```suggestion
    } else if (name.size() > 1 && name[0] == '#') {
      // numeric character reference: &#201; or &#xC9;
      const bool hex = name[1] == 'x' || name[1] == 'X';
      uint32_t cp = 0;
      bool valid = name.size() > (hex ? 2u : 1u);
      for (size_t d = hex ? 2 : 1; valid && d < name.size(); d++) {
        const char c = name[d];
        uint32_t v = 0;
        if (c >= '0' && c <= '9') {
          v = static_cast<uint32_t>(c - '0');
        } else if (hex && c >= 'a' && c <= 'f') {
          v = static_cast<uint32_t>(c - 'a' + 10);
        } else if (hex && c >= 'A' && c <= 'F') {
          v = static_cast<uint32_t>(c - 'A' + 10);
        } else {
          valid = false;
        }
        cp = cp * (hex ? 16u : 10u) + v;
      }
      // reject NUL, surrogates and out-of-range rather than emit bad UTF-8
      if (valid && cp > 0 && cp <= 0x10FFFF && !(cp >= 0xD800 && cp <= 0xDFFF)) {
        utf8AppendCodepoint(cp, out);
      } else {
        out.append(in, i, semi - i + 1);
      }
    } else {
```

(Needs `#include <Utf8.h>` added to `LibraryMeta.cpp` as well)

Uri-Tauber: Manual parsing code should live in UTF library somewhere, not in the library indexer...

lib/LibraryIndex/LibraryBuilder.cpp
oreglio: Confirmed — with metadata enabled the per-directory call is the only feed point, and the caller's watchdog feed never runs inside a large flat folder, so the TWDT fires and the missing index retriggers the rebuild on reboot. Fixed in ebf0d3fe: the callback now runs after every staged book, and the per-directory call is kept (guarded by `!st.aborted`) so trees of book-less folders still feed it. One deviation from the suggestion: the per-book abort `break`s instead of returning, so the directory handle still reaches `dir.close()`.

lib/LibraryIndex/LibraryBuilder.cpp
oreglio: Fixed in 87e729bd — a failed open now sets `writeFailed` (with a LOG_ERR naming the stage file), so the emit is discarded and the previous index stays live instead of an all-zero folder section that validates.

lib/LibraryIndex/LibraryBuilder.cpp
oreglio: Fixed in 87e729bd, same commit as the folder-stage guard — it is the same failure surface. The four seek+read pairs in the record and blob loops now go through a `fetch()` helper that mirrors `put()`: a short read sets `writeFailed` and breaks, and the final guard discards the emit. The rank-key reads further up are left unchecked deliberately: a short read there yields a zeroed key, which can only degrade ordering for that book, never corrupt a record — and the record loop's `fetch()` still fails the emit if the card is actually unreadable.

lib/LibraryIndex/LibraryBuilder.cpp
oreglio: No reason zero is preferable — it is wrong twice, as you said: "Recently added" breaks now, and the next rebuild reads the zeros back as prior truth. Fixed in 081dac3d: an unopenable stage or a short read logs, removes the staging files and returns false, keeping the previous index — the same policy the aborted walk already follows.

lib/LibraryIndex/LibraryBuilder.cpp
oreglio: Fixed in 7d10b931 — walk() now skips (and counts in `unreadableSkipped`) any entry whose basename exceeds 255 bytes or whose folder path exceeds the folder section's one length byte, before dedup and staging. The former clamps became plain casts with the invariant stated at the skip site. Title and author spellings are still truncated, but those are display-only — they never reconstruct a path.

lib/LibraryIndex/LibraryMeta.cpp
oreglio: Fixed in c67a09c1. Still a bounded scan — pulling expat in here felt disproportionate for two attributes — but it now anchors on what ContainerParser anchors on: XML comments are skipped, only `<rootfile>` elements count, and the media-type must be `application/oebps-package+xml`. Quoted attribute values may contain `>`, and `=` tolerates surrounding whitespace. One residual divergence, called out in a comment: this takes the first matching rootfile (the spec's default rendition) while ContainerParser keeps the last match it sees, so the two can still disagree on a multi-rendition EPUB. If that case matters, the cleaner fix is on the reader's side.

lib/LibraryIndex/LibraryMeta.cpp
oreglio: Implemented in 9876ced1, with the parsing placed per @Uri-Tauber's note: the decoder lives in lib/Utf8 as `utf8AppendNumericCharRef()` beside `utf8AppendCodepoint()`, and `decodeEntities()` just calls it. Two changes from the inline version suggested here: the range check runs per digit, so an arbitrarily long digit run rejects instead of wrapping the 32-bit accumulator back into the valid range, and the helper has host-side gtest coverage (`test/utf8_compose/Utf8NumericCharRefTest.cpp`) for the decimal/hex forms, NUL/surrogate/out-of-range rejection, and that overflow case. The named XML entities stay local to LibraryMeta — they are XML, not UTF-8, so lib/Utf8 did not seem like their home either. @Uri-Tauber if you meant the whole entity decoder should move somewhere shared, say the word and I will lift it out.

Uri-Tauber:
Sorry for the wait. I'm trying to juggle between dozen different PRs.
Here some comments on the current stage:

1. rebuild invariants are delegated to callers
The builder already opens and reads the previous index at `lib/LibraryIndex/LibraryBuilder.cpp:879`, but initializes its counter from a caller-supplied value at `lib/LibraryIndex/LibraryBuilder.cpp:907`. It also makes watchdog servicing conditional on the optional callback at `lib/LibraryIndex/LibraryBuilder.cpp:389`. The two callers already stacked in #2889 duplicate both the prior-index read and the same `delay(1)` lambda.

  - Remove `previousNextFirstSeen` from `lib/LibraryIndex/LibraryBuilder.h:65`.
  - Capture `previous.header().nextFirstSeen` inside the existing successful `previous.open()` block.
  - Yield internally at the established interval.
  - Retain the callback only for observation/cancellation.

2. `LibraryMeta` reimplements container resolution, OPF element parsing, XML entities, and metadata extraction; `stageRecord()` chooses between Epub and that parallel parser.

Epub already owns container and OPF parsing at `lib/Epub/Epub.cpp:16` and `lib/Epub/Epub.cpp:48`. The PR duplicates those rules in `lib/LibraryIndex/LibraryMeta.cpp:68` and `lib/LibraryIndex/LibraryMeta.cpp:233`, then explicitly documents a remaining disagreement over multi-rendition EPUBs at line 271.

Suggestion: Expose `Epub::loadForMetadata(MetadataSource*)` beside `lib/Epub/Epub.h:47`. It owns cache-first fallback, package resolution, XML decoding, and provenance. Then `lib/LibraryIndex/LibraryBuilder.cpp:191` becomes one EPUB call and `LibraryMeta.*` disappears.

This should reuse the existing streaming parser and its chunk buffers rather than introduce another heap allocation; it removes the current 8 KB whole-metadata buffer. **Heap must still be measured on-device because Expat’s internal allocation varies by document.**

3. Nitpick: The builder first writes folder paths to a temporary file. It does not check whether those writes actually succeeded.
check both folder-stage `write()` calls and mark the build failed if either writes fewer bytes than requested. Do the same for the record-stage write, then discard the temporary files and retain the previous index.

4. Large directory listings can crash the firmware instead of degrading gracefully. 
While scanning one directory, the code accumulates every book already seen in that directory, to filter duplicates. 
To keep the PR small. either accept duplications or add a hard cap of the number of files allowed in one folder. This could be done better, but should deserve a follow-up.

5. Metadata reading keeps decompressing after it has found the title and author

For EPUBs, the code reads the package metadata—the part that contains title and author. It detects the closing `</metadata>` tag, but does not actually tell the ZIP reader to stop at that point. Instead, it continues decompressing until it reaches the 8 KiB safety cap. That is still memory-safe, but it performs much more SD and CPU work than needed for every EPUB with a large manifest.

This is a small local fix: once `BoundedSink` finds the metadata closing tag, make its `write()` return a short result. `ZipFile::readFileToStream(..., allowEarlyStop=true)` already interprets that as “the caller has enough data; stop successfully.”

oreglio:
Thanks @Uri-Tauber, I checked all five points against the exact PR head and pushed the fixes as separate commits:

1. `buildLibraryIndex()` now feeds the watchdog internally throughout the rebuild; caller callbacks only handle progress/cancellation.
2. `LibraryMeta` is removed. Metadata extraction now reuses `Epub`/`ContentOpfParser`, with consistent NFC normalization and correct handling of Expat-split creator text.
3. Record and folder staging use exact I/O. Any short write/read aborts the build instead of producing a padded index.
4. Dedup storage is now one fixed, fallible 8 KiB allocation. Failure/overflow degrades explicitly with a persisted flag; sort scratch is bounded too.
5. Metadata-only parsing keeps one ZIP open and stops at `</metadata>`, before the manifest, so it creates no `.items.bin` or section caches.

I also fixed root-level `//book.epub` paths and made index replacement recoverable through a `.bak` file.

Verification on the final stacked head: 191/191 host tests, C3 and Sticky builds, static analysis, and all CI jobs pass. I separately tested a local X3/X4 beta containing these changes; OTA installation and the device smoke test passed.

I’ll still collect fresh large-card timing, `freeHeap`/`maxAlloc`, and the on-card no-`.items.bin` check before requesting another review.

Uri-Tauber (approved):
LGTM.

PR #2888, opened 2026-08-06
feat: add the Library shelf activity

oreglio:
## Summary

This is the shelf UI on top of the index from #2887.

- Adds a paged Library activity with four strip entries: Recently added, Titles, Author, and Search.
- Shows a one-line title with the author below it and opens the original book path from the index.
- Supports ascending/descending title order, author order, recent order, and a keypad search over folded title/author text.
- Keeps navigation valid when a search returns zero rows: the Search tab remains selected, Left/Right still move between tabs, and the empty message is query-specific.
- Uses a fixed 256-byte page-history buffer. Search matches use a fallible allocation capped at 4,096 `uint16_t` entries (8 KiB); allocation failure displays an explicit translated error instead of risking an abort.
- Logs rank/dedup degraded states reported by the index instead of presenting them as fully sorted data.

This branch is restacked directly on the reviewed/fixed head of #2887.

## Verification completed

- Host suite: 191/191 tests passed.
- `pio run -e default` (ESP32-C3): passed.
- `pio run -e sticky` (ESP32-S3): passed.
- `pio check -e default --fail-on-defect low --fail-on-defect medium --fail-on-defect high`: no defects found.

The zero-result keyboard-navigation case raised in review is covered by the revised state handling and remains part of the X3 hardware checklist.


src/activities/library/LibraryListActivity.cpp
```diff
@@ -0,0 +1,1114 @@
...
+                                      static_cast<int16_t>(metrics.buttonHintsHeight + readoutReserved), 0});
+
+  if (letterGrid) {
+    screen.spacer(static_cast<int16_t>(sortStripHeight(screen) + metrics.verticalSpacing));
+    buildLetterGrid(screen);
+    return;
+  }
+
+  if (rowCount() == 0) {
+    screen.centeredText(tr(STR_LIBRARY_EMPTY));
```
hamaluik: The empty-search state seems to strand the query.

With an active query and zero matches this shows `STR_LIBRARY_EMPTY`, which reads as "no books on the card" even when the card is full. It also returns before `buildSortTabs()` so the Search tab becomes unreachable and Back is the only way out which just throws the query away rather than letting a typo be fixed.

Maybe a query-aware empty message, and keeping the strip reachable while a query is active?

hamaluik (commented):
Ran into a few things here while building the series branch on top of this code — sending them while the PR is still open in case they're useful. Push back on any of them.

Uri-Tauber:
@oreglio Before you rebase this stage, I want to rebase `feat/library-view` on develop, so we will be in sync with the additions from the last three weeks.

Uri-Tauber:
@oreglio Ready to rebase.

Also, take a look at the following suggestions. They all related to code already in `feat/library-view` which can be cleaned without any change in functionality. Treat them with suspicion, its possible I missed something from 2888 and 2889 which consume those.

`lib/Utf8/Utf8.cpp:L135-161`: delete: utf8AppendNumericCharRef has no production caller; Expat already decodes character references. Remove helper, declaration, dedicated test, and CMake entry.

`lib/LibraryIndex/LibraryFormat.h:L58-77`: delete: ClixRecord format/provenance/title-origin flags are never read. Reserve the byte; remove flag packing, MetadataSource plumbing, and tests.

`lib/LibraryIndex/LibraryFormat.h:L120-121`: delete: authorRank/dateRank duplicate the permutation arrays and no reader consumes them. Reserve those bytes; remove both inverse-rank arrays and passes.

`lib/LibraryIndex/LibraryIndexFile.h:L69`: shrink: heap-allocated HalFile exists only to permit a forward declaration. Include HalStorage.h and store HalFile directly; delete new/delete/null plumbing.

`lib/LibraryIndex/LibraryText.cpp:L243-260`: shrink: authorKey repeats cleanPersonName’s bracket/secondary-author cleanup. fold(cleanPersonName(author)) replaces the duplicate pass.

`lib/LibraryIndex/LibraryText.cpp:L329-357`: stdlib: manual reverse scan finds the last word. string::rfind plus two substr calls.

src/activities/library/LibraryListActivity.cpp
oreglio: Fixed in `d44f19c5`. A zero-result search now remains in the Search tab, renders the query-specific empty state, and keeps Left/Right tab navigation active; row navigation is simply a no-op while there are no matches. I also made the search-result buffer fallible (maximum 8 KiB at the 4,096-book format cap), with an explicit translated low-memory error instead of a possible allocation abort. The final stacked branch passes all GitHub checks.

oreglio:
Restacked on the rebased base. At that point #2888 was back to the 5 files it should contain (+1,364); the earlier 32-file / +3,276 diff was the old merge base replaying #2887.

**Update after #3256 merged:** restacked again on `f9c3257d` (new head `5decc101`). GitHub now shows the intended 5 files / +1,364 and reports the branch mergeable. All GitHub checks are green.

I checked the `ContentOpfParser` resolution because namespace handling is easy to get plausibly but silently wrong. The final use of `xmlLocalNameEquals` for the package/metadata boundaries and for the metadata-only stop conditions is correct.

Correction to my earlier explanation: failure to recognize only `</metadata>` would leave the parser in `IN_METADATA`; it does not by itself prove that the parser would enter the manifest or write a per-book `.items.bin`. The confirmed issue is narrower: namespace-sensitive comparisons could prevent the intended boundary recognition and early stop. The existing `manifest`/`spine`/`guide` fallback is now namespace-aware as well, so an OPF with a missing metadata close still stops, and that behavior has a test.

One conflict in the earlier restack was in `SettingsActivity.cpp`: #2983 removed the `hasTouch()` guard around `CheckForUpdates` because OTA now fetches each board's own release asset. I preserved that behavior and placed the library rebuild line above it.

### The six suggestions — merged as #3256

Four changes are straightforward and remain sound: removing the unused numeric-character-reference helper, removing the inverse-rank scratch arrays and passes, reserving the retired record fields without changing offsets or `CLIX_FORMAT_VERSION`, and holding `HalFile` by value.

Two points need a narrow follow-up on `feat/library-view`:

- `authorKey(cleanPersonName(author))` is not proven equivalent for every input. Token sorting makes ordinary comma inversion inert, but `authorKey` retains at most 12 tokens before sorting; on a long comma-form name, inversion can change which 12 tokens survive. It also repeats the allocation-heavy cleanup because `stageRecord` cleans the display author separately. The underscore test covers one case, not the full equivalence claim.
- Replacing the manual surname scan with `rfind` is semantically sound because `fold()` normalizes spaces. However, the two `folded.substr(...)` calls create temporary strings once per book. They should use positional `append` (or a view) on the C3 path.

#3256 contains 67 additions and 292 deletions, i.e. **225 net lines removed**, not 222. The record layout is still 128 bytes and the removed members stay reserved at the same offsets. The reserved region discussed here is two `uint16_t` fields plus one flag byte (**5 bytes**, not 12).

Removing the ranks does save two `uint16_t[N]` scratch arrays (up to 8 KB each at the 4,096-record cap) and three full passes over the record set per rebuild.

### Verification

The restacked commits are patch-equivalent to the previous three commits (`git range-diff` marks all three `=`). clang-format 21 is clean; `pio run -e default` and `pio run -e sticky` both pass; all 35 `LibraryText` tests and the 20 targeted `LibraryFormat` / `ContentOpfParser` tests pass. The complete macOS host build still stops in the unrelated upstream `CssParser.cpp` because this libc++ has no `std::from_chars(float)`; Linux CI covers that suite. All GitHub checks on `5decc101` are green, including all four firmware builds, clang-format, cppcheck, unit tests and the aggregate status check.

X3 measurements are still outstanding: rebuild duration, `freeHeap`, `maxAlloc`, and confirmation that a metadata rebuild creates no per-book `.items.bin`.

Uri-Tauber:
Ok. I merged #3256 and fixed clang-format plus rebased on develop.
Should be ready for you to rebase again.

oreglio:
Rebased again onto `f9c3257d`; the new head is `5decc101`.

`git range-diff` marks all three PR commits patch-equivalent, GitHub is back to the intended 5 files / +1,364, and all checks are green.

Uri-Tauber:
`LibraryListActivity` currently builds a second navigation and layout system alongside `FreeInkUI`. Please simplify these boundaries before merging:

1. `src/activities/library/LibraryListActivity.cpp:224-271,598-615,697-781,939-984`

Remove the second tab state machine. Make A–Z and Z–A ordinary explicit sort tabs, move Search to a separate FUI action, and use `UiTabListActivity::buildTabBar()`. Delete `sTitleDescending`, `flipTitleDirection()`, the title-mode controls, long-press handling, triangle decoration, and the separate active-tab/cursor logic. This keeps tab styling and future navigation fixes owned by the shared base.

2. `src/activities/library/LibraryListActivity.cpp:783-925`

Stop reproducing `FreeInkUI`’s list-layout algorithm inside the activity. The code currently duplicates header height, section spacing, row fitting, selection clamping, and viewport feedback. Extend the list component so a selectable book row can carry an optional section heading, then provide one `ListItem` per book and use the existing virtual-window and `ListNav` APIs. This preserves bounded RAM without coupling the activity to the component’s internal geometry.

3. `src/activities/library/LibraryListActivity.h:145-151`, `src/activities/library/LibraryListActivity.cpp:1101-1144`

Delete the 260-byte page-history state and its maintenance code. Exact restoration of previously visited page boundaries was not requested. Use the existing `visibleRows` fallback for backward paging; it is smaller and avoids a second pagination mechanism alongside `ListNav`.

4. `src/activities/library/LibraryListActivity.cpp:119-132`, `src/activities/library/LibraryListActivity.h:58,115`

Let `LibraryIndexFile`’s RAII state be authoritative. `open()` already closes any previous handle, `isOpen()` already exposes readiness, and member destruction closes the file. Remove `openIndex()`, `indexReady`, and the teardown-only resets in `onExit()`. Keep the explicit close immediately before opening a book, where the handle must be released early.

5. `src/activities/library/LibraryListActivity.h:13-28`, `src/activities/library/LibraryListActivity.cpp:24-1144`

Reduce ~300 lines of comments. Much of this is PR rationale, measurements, rejected alternatives, or narration of the following code. Retain only comments that a future developer looking at the code after it was merged will need to see.

Uri-Tauber:
Thanks for addressing my comment, @oreglio.

another thing: I think it would be better to simplify the Library shelf’s tab handling before merging.

Use four ordinary sort tabs: Recent, Title A–Z, Title Z–A, and Author. Let `UiTabListActivity::buildTabBar()` own their rendering and navigation.

Move Search to a separate FreeInkUI action, following the existing header-action pattern in `OpdsBookBrowserActivity`. Remove the Library-specific tab cursor, title-direction state, long-press direction toggle, triangle decoration, custom tab drawing, and duplicate tab navigation.

This makes the shared UI layer authoritative, so future changes — such as adding the Series tab — only need to add a tab instead of updating several parallel state machines.

The tradeoff is that title direction is no longer changed by holding Confirm or from the A–Z grid. Users select A–Z or Z–A directly, which keeps the behavior visible and removes the duplicated controls.

Uri-Tauber (approved):

Uri-Tauber:
Thanks @oreglio.

oreglio:
Thanks for the thorough review and for merging!

PR #2892, opened 2026-08-06
fix: hyphenate focus-split words and break after visible hyphens

### LINKED ISSUE #2115: Hyphenation works poorly in Focus Reading mode
### Affected Version

main (post 1.3.0)

### Describe the Bug

Hyphenation (English-language) does not work well with Focus Reading mode.

In particular, the hyphenation engine doesn't seem willing to break at many acceptable points (including both within word and after some punctuation, including explicit "-" characters) when Focus Reading is turned on.

In particular, **in FR mode, hyphenation seems not allowed at points adjacent to the bolded text.** This seems to include no breaking allowed when a word immediately follows a hyphen or em dash, and no hyphenation immediately after the bolded bit.

### Steps to Reproduce

1. Turn on Focus Reading
2. Turn on Hyphenation
3. Read

### Expected Behavior

Hyphenate more aggressively

**Several examples:**

## Breaking after explicit hyphens

Image

See the phrase "Lord-Who-Is-Seen-Within"

Image

Image

## Breaking within words

Image

Image

## General line wrapping

Image

```shell

```

Uri-Tauber:
# Summary

Fixes #2115 — hyphenation barely works when Focus Reading is on.

This is an alternative to #2165, which diagnoses the same bug but took a different route.

Credit where it's due: @lukestein found this bug, characterised it precisely, built the stress-test EPUB, and kept a fix alive on the tree for two months. The diagnosis below is his; only the repair strategy differs. ThePR is co-authored to him.

## What a reader sees today

Turn Focus Reading on and hyphenation quietly stops doing its job. Long words refuse to break, so lines end early and the right margin goes ragged — worst at large font sizes, where each line holds fewer words and a single unbreakable word wastes a lot of space. Compound words joined by dashes ("Lord-Who-Is-Seen-Within") never wrap at all; the whole thing jumps to the next line.

Turn Focus Reading off and the same book hyphenates fine.

## Why it happens

Focus Reading bolds the first few letters of every word to give the eye an anchor:

> **cer**tainly

The way it did that was to cut the word into two pieces — a bold piece _cer_ and a normal piece _tainly_ — and glue them back together at drawing time.

Everything that happens between those two moments sees two short fragments instead of one real word. The hyphenation dictionary is one of those things. Asked where to break _tainly_, it has no answer, because _tainly_ is not a word. Asked about _certainly_, it would have said cer-tain-ly immediately.

So hyphenation didn't break. It was simply never shown anything it could work with.

## What this PR changes

Stop cutting the word up. Keep it whole and attach a note: "render the first 4 letters of this word in bold." The reader sees exactly the same thing — the drawing layer already worked this way and had to reassemble the two pieces anyway, so this actually removes a step.

With the word intact, the hyphenator sees *certainly*, and every legal break point is available again, including the one that falls exactly where the bold ends. That last case matters more than it sounds: for a large share of English words it is the only place the word can break.

Let dashed compounds wrap. *Lord-Who-Is-Seen-Within* may now break after any of its dashes, which is what the same text already does with Focus Reading off. The mechanism for "you may break here, but don't insert a space" already existed in the codebase for Korean text; this reuses it rather than inventing a second one.

## Why not #2165's approach

#2165 keeps the word cut in two and teaches each later stage how to cope: reassemble the halves before consulting the dictionary, translate the answer back, add a special case for breaking exactly at the bold edge, another for compound words, another for punctuation that trails a word. It works — it reaches comparable quality, and @lukestein has been reading on it daily for months, which is real evidence.

The concern is what stays behind. Those special cases are permanent, they live in two separate line-breaking algorithms that must agree, and each new one exists only because words arrive pre-broken. The next piece of typography work in this area pays the same tax again, and a future contributor has no way to tell that any of it is optional. Removing the cut removes the reason all of them exist.

To be fair to the other approach: it is lower-risk in the narrow sense. It only adds code, so anything it doesn't touch cannot regress. This PR changes how every word is represented during layout, which is a wider blast radius — it touches word measurement, line breaking, right-to-left reordering and the drawing hand-off. That is the real trade, and it is why this needs proper device testing rather than a quick look.

---

### AI Usage


lukestein:
Thanks very much @Uri-Tauber for taking this on and for the thoughtful write-up above!

I installed the draft PR and I'm doing some testing on device. It looks great so far, with the only immediate problem I notice is it introduces an extra horizontal space after hyphens and em dashes:

IMG_5461

IMG_5462

lukestein:
Hi @Uri-Tauber .

(With agent help) I traced the extra spacing to the new dash-boundary representation:

> The token following the dash gets `noSpaceBefore=true`. Although this removes the normal space, `extractLine()` counts every `noSpaceBefore` boundary as a stretchable justification gap and adds `justifyExtra` there on justified non-final lines. That appears to explain the device screenshots.
>
> To clarify the intended behavior: all characters covered by `isExplicitHyphen()`—including underscores, ellipses, minus signs, and the various dashes—should create line-break opportunities. When no break is taken, however, the following text should remain directly adjacent, with neither a normal space nor justification expansion. Soft hyphens and U+2011 non-breaking hyphens should presumably retain their special semantics.
>
> One compact way to represent this without another per-token allocation may be to use the currently unused `continues=true, noSpaceBefore=true` combination for a “breakable, zero-width, non-stretchable boundary.” The line breakers could recognize that combination as breakable, while positioning could treat it like an attached/kerning boundary when the break is not taken. This would leave the existing `continues=false, noSpaceBefore=true` behavior available for stretchable CJK/Korean boundaries.


Happy to prepare a small PR into your branch that rebases onto current `develop` and tries to fix. Would that be helpful?

Uri-Tauber:
@lukestein Go ahead.
I will have time tomorrow night to work on this.

lukestein:
> @lukestein Go ahead. I will have time tomorrow night to work on this.

@Uri-Tauber thanks. Added as https://github.com/Uri-Tauber/crosspoint-reader-fork/pull/22 which I'm also testing on device

itsthisjustin (approved):

PR #2897, opened 2026-08-06
fix: read platformio.ini as UTF-8 in git_branch.py

lranslot:
configparser.read() without an explicit encoding decodes using
locale.getpreferredencoding(), which is the ANSI code page on Windows
(cp932 on Japanese systems). Reading platformio.ini then raises
UnicodeDecodeError on any non-ASCII byte and the pre-build script aborts.
platformio.ini is UTF-8, so decode it as UTF-8 regardless of locale.

## Summary

* **What is the goal of this PR?** Make `pio run` work on Windows systems
  whose default locale is not UTF-8 (Japanese, Chinese, Korean).
* **What changes are included?** One line in `scripts/git_branch.py`:
  pass `encoding='utf-8'` to `config.read()`.

### Error before the fix
```
UnicodeDecodeError: 'cp932' codec can't decode byte 0x94 in position 3011:
illegal multibyte sequence
File "scripts/git_branch.py", line 72: config.read(ini_path)
```

The only workaround is setting `PYTHONUTF8=1` in every terminal session,
and the error gives no hint that locale is the cause.

### Environment

* Windows 11, Japanese locale (cp932)
* Python 3.12.10, PlatformIO Core via the VS Code extension
* Reproduced on tags 1.4.1 and 1.5.0

macOS, Linux and English Windows are unaffected because they default to
UTF-8, which is likely why this has gone unnoticed.

## Scope Check

Please confirm:

## Additional Context

Build-tooling only; no runtime code is touched. No memory or flash impact.

Verified after the change:
```
CrossPoint build version: 1.5.0-dev-fix/git-branch-encoding-fecb9eb3
```

---

### AI Usage

I used Claude to investigate the traceback and to help draft this
description. The change itself is a one-line encoding argument that I
verified by reproducing the failure and confirming the fix on my own
hardware.

### DIFF
--- a/scripts/git_branch.py
+++ b/scripts/git_branch.py
@@ -69,7 +69,7 @@ def get_base_version(project_dir):
         warn(f'platformio.ini not found at {ini_path}; base version will be "0.0.0"')
         return '0.0.0'
     config = configparser.ConfigParser()
-    config.read(ini_path)
+    config.read(ini_path, encoding='utf-8')
     if not config.has_option('crosspoint', 'version'):
         warn('No [crosspoint] version in platformio.ini; base version will be "0.0.0"')
         return '0.0.0'


Uri-Tauber (approved):
LGTM

PR #2928, opened 2026-08-08
fix(ui): wrap button hint labels that overflow their button

vishae:
## Summary

* **What is the goal of this PR?** Button hint labels wider than their button no longer overflow the border and run into the neighbouring hint.
* **What changes are included?** The fit-or-wrap logic moved into a shared `BaseTheme::drawHintLabel()`, called by both `BaseTheme` (106px buttons) and `LyraTheme` (80px). Labels that already fit render identically to before. Wider ones wrap to at most two lines, centred within the button, with `wrappedText()` ellipsising anything that still doesn't fit. Lines are spaced by the glyph height rather than `getLineHeight()`, whose full `advanceY` stacks two lines taller than the button. Lyra's side (rotated) hints are untouched.

## Scope Check

Please confirm:

      the relevant maintainer. (Not applicable — this touches neither.)

## Additional Context

Found with a two-word label ("Update now") on an 80px Lyra button: the text was drawn centred and unclipped, so it ran past the button border into its neighbour.

* No new theme — this is a rendering fix inside the existing theming surface, and it applies equally to both themes shipped today.
* Flash impact negligible (a few hundred bytes); no measurable RAM change.
* Verified on an X4 with both the base theme and Lyra, across single-word and two-word labels.
* Reviewers may want to look at the vertical placement in particular: on an 80px/40px Lyra button the second line sits slightly tight against the bottom edge.

---

### AI Usage

### DIFF
--- a/src/components/themes/BaseTheme.cpp
+++ b/src/components/themes/BaseTheme.cpp
@@ -149,6 +149,38 @@ void BaseTheme::drawProgressBar(const GfxRenderer& renderer, Rect rect, const si
   renderer.drawCenteredText(UI_10_FONT_ID, rect.y + rect.height + 15, percentText.c_str());
 }
 
+// Centre a button-hint label inside its box. A label that fits is drawn on the
+// single baseline it always was; one too wide used to overflow the button border
+// and run into the neighbouring hint, and now wraps to at most two centred lines
+// (wrappedText() ellipsises anything that still doesn't fit). Shared so every
+// theme's drawButtonHints() gets the same behaviour.
+void BaseTheme::drawHintLabel(GfxRenderer& renderer, const int fontId, const char* label, const int x,
+                              const int boxWidth, const int boxTop, const int boxHeight,
+                              const int singleLineYOffset) {
+  constexpr int textPadding = 4;  // keeps a wrapped label off the button's border
+  const int maxTextWidth = boxWidth - (textPadding * 2);
+
+  const int textWidth = renderer.getTextWidth(fontId, label);
+  if (textWidth <= maxTextWidth) {
+    renderer.drawText(fontId, x + (boxWidth - 1 - textWidth) / 2, boxTop + singleLineYOffset, label);
+    return;
+  }
+
+  // Spaced by the glyph height, not getLineHeight() — that returns the font's
+  // full advanceY (leading included), which stacks two lines taller than the
+  // button and clips the second one.
+  constexpr int lineGap = 2;
+  const int step = renderer.getTextHeight(fontId) + lineGap;
+  const auto lines = renderer.wrappedText(fontId, label, maxTextWidth, 2);
+  const int block = static_cast<int>(lines.size()) * step - lineGap;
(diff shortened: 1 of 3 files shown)


Uri-Tauber:
Please share before/after screenshots.

vishae:
> Please share before/after screenshots.

Not sure if I'm supposed to crop the screenshots or do the full screen. It's the "Update now" button at the bottom.
screenshot-91682
screenshot-176998

The device screen crop the "now" a little bit at the bottom, but it's still legible. I tried making the line spacing even smaller, but it didn't make any noticeable difference.

Uri-Tauber:
@vishae This screen is not a part of crosspoint. Where is it from?
Anyway, I think it would be simpler to shorten to "Update".

vishae:
> @vishae This screen is not a part of crosspoint. Where is it from? Anyway, I think it would be simpler to shorten to "Update".

No, it's not part of crosspoint. I'm trying to create a fork, and noticed that the labels don't wrap. Although it would be easier to change the label to say "Update", I thought it would be better for the code to wrap text in that area in general.

Uri-Tauber:
@vishae First of all, thanks for porting fixes upstream. 
Specifically for this fix, I'm hesitant to merge, because its a hot path, and the extra competition could cost several ms.
How about to manually insert `\n`? I think the renderer already handle those, but I'm not sure.

vishae:
> @vishae First of all, thanks for porting fixes upstream. Specifically for this fix, I'm hesitant to merge, because its a hot path, and the extra competition could cost several ms. How about to manually insert `\n`? I think the renderer already handle those, but I'm not sure.

Thanks — I tested the `\n` suggestion on device, and the renderer doesn't handle it.

`GfxRenderer::drawText()` walks the string codepoint by codepoint and draws a glyph for each, with no case for `\n` and no baseline advance. U+000A isn't in any font's interval table, so `EpdFont::getGlyph()` falls through to the
replacement glyph. `getTextWidth()` goes through the same walk, so the newline even counts toward the centring maths, and `wrappedText()` splits on spaces only. The result is one line reading `Update□now`, still overflowing the button:

screenshot-55012

On the hot-path concern: for a label that fits, the helper does exactly what the code did before — one `getTextWidth()`, one `drawText()`. The only additions are a function call and two integer comparisons; the wrapping branch is never
entered unless the label already overflows, which no label upstream does today.
I measured the build cost as well: +274 bytes of flash, RAM unchanged.

If the call itself is the worry, I'm happy to move `drawHintLabel()` into the header as an inline function so the fitting path compiles to what it was.

Which path were you thinking of specifically? `drawButtonHints()` runs once per screen render, immediately before a display refresh, so I'd like to make sure I'm measuring the case you have in mind.

I'm not actually that married to this PR; I'm just addressing your concerns in case there's anything I can do before I discard the idea.

Uri-Tauber (approved):

Uri-Tauber:
> Thanks — I tested the \n suggestion on device, and the renderer doesn't handle it.

Hmm. That's weird. Anyway, if so, your code is fine.

PR #2938, opened 2026-08-09
fix(ui): wrap the confirmation body instead of truncating it

vishae:
## Summary

* **What is the goal of this PR?** A confirmation prompt longer than one screen width no longer loses its tail.
* **What changes are included?** `ConfirmationActivity` ran its body through `truncatedText()`, so anything past one line was cut — including the part explaining what is about to happen. The body now wraps to at most six lines via `wrappedText()`, and the vertical centring accounts for their real height. The heading is deliberately left single-line: it's a title, not prose. Existing single-line prompts render exactly as before.

## Scope Check

Please confirm:

      the relevant maintainer. (Not applicable — it touches neither.)

## Additional Context

I tried to create a prompt in my fork that reads "Delete every book's cached covers and metadata? Books and reading progress are untouched..." — it was cut mid-sentence, hiding the reassurance that made it safe to answer.

* Affects every caller that passes a body longer than one line; short prompts are unchanged.
* Negligible flash impact; no measurable RAM change.
* Same class of defect as #2928 (text drawn without regard for the space available), but a separate concern in a separate file, so it's a separate PR.
* Verified on an X4 with both a long and a short prompt.

---

### AI Usage

### DIFF
--- a/src/activities/util/ConfirmationActivity.cpp
+++ b/src/activities/util/ConfirmationActivity.cpp
@@ -18,8 +18,11 @@ void ConfirmationActivity::onEnter() {
   if (!heading.empty()) {
     safeHeading = renderer.truncatedText(fontId, heading.c_str(), maxWidth, EpdFontFamily::BOLD);
   }
+  // The body wraps rather than truncating: a prompt explaining what is about to
+  // happen is routinely longer than one line, and truncating it cut off the part
+  // that mattered. The heading stays single-line — it's a title, not prose.
   if (!body.empty()) {
-    safeBody = renderer.truncatedText(fontId, body.c_str(), maxWidth, EpdFontFamily::REGULAR);
+    bodyLines = renderer.wrappedText(fontId, body.c_str(), maxWidth, maxBodyLines, EpdFontFamily::REGULAR);
   }
 
   // Text sits in the upper part of the screen so the confirmation popup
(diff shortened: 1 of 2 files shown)


vishae:
Screenshots (from an XTEINK X4):
screenshot-35115 2
screenshot-71756 2

Update: Corrected the screenshots to show the default delete confirmation screen.

vishae:
On the `startY` comment: the text isn't meant to be vertically centred here. `startY = screenHeight / 6` is deliberate — the confirmation popup is centred, so centring the text block would put it behind the popup. Recomputing `startY` from the wrapped height would reintroduce exactly what that line avoids.

There is a real issue underneath it though, so thanks for the prod. The block grows downward from a fixed start, and I hadn't checked what the tallest case does. Measuring on an X4 (800px), base theme metrics:

- The popup's dialog is 141px tall, so its top edge sits at `(800 - 141) / 2 = 329`.
- The text starts at `800 / 6 = 133`; heading 24 + spacing 30 + six body lines at 24 each = 198, ending at 331.

So a heading plus the full six lines just reaches the popup — about 2px, within descender space rather than a visible overlap, but with no margin at all. I've capped the body at five lines, which ends at 307 and clears by roughly 20px. Pushed as a separate commit with the numbers in the message.

Five is enough for the cases this actually fixes upstream: the confirmation screens here pass a book title, a font family name, or a filename as the body, and those were being truncated mid-word. The screenshots are being retaken against the five-line build.

Issue #2916, opened 2026-08-07
Dictionary look up not returning base word forms in German


luki531:
### Affected Version

1.5.0

### Describe the Bug

In German language words can take up a multitude of forms due to declension, conjugation, adding zu, etc. As is the lookup feature doesn't display the base form of such words in German, unlike in English where for example when searching for plural form it returns the singular. This fact renders the feature borderline useless for German and possibly other non English languages.

### Steps to Reproduce

Dictionary from reader.dict de-de
1. Look up words Verhandlungen, fortzusetzen, hätten
2. Look up returns not found 

### Expected Behavior

Look up returning base forms Verhandlung, fortsetzen, haben as it does while using the same dictionary in ColorDict.

<img width="1080" height="1503" alt="Image" src="https://github.com/user-attachments/assets/9d2ba0c2-0cf6-4472-9b42-9d264119c061" />

### Relevant Log Output/Screenshots

```shell

```

metoli86:
#2942 improves this: on a miss the reader now falls back to the longest indexed headword that the selected word begins with, so declension/plural forms land on their base form — `Verhandlungen` → `Verhandlung` works with the reader.dict de-de dictionary.

Being transparent about scope: forms that change *inside or in front of* the word rather than at the end — `fortzusetzen` → `fortsetzen` (infix `zu`), `hätten` → `haben` (umlaut), `gesagt` → `sagen` (ge- prefix) — are not covered; those need real German morphology rather than index-order tricks. But the large class of suffix-built forms (genitives, plurals, -ung nominalizations and the like) should now resolve instead of returning "Not found".

sorig:
Same experience here. The dictionary feature almost always gives me "not found" because German words rarely appear in their base forms. I.e. the dictionary feature is not really usable for German in its current state.

I know other platforms like iOS have APIs that go beyond simple heuristics to identify the base form of words for lookups. See [NLTagger](https://developer.apple.com/documentation/naturallanguage/nltagscheme). Not sure if there are any open source equivalents to this.

sorig:
Had the same issue. Just tested 1.6.0rc which has greatly improved the usability of the dictionary in German via #2696 which now ensures that the lemmatization table (`.syn` file), which already ships with the reader.dict dictionaries, is being used.

sorig:
I found a bug that causes some words with capitalised umlauts to return "Not found". Fixed in #3123

brianhuster:
[...]

PR #2942, opened 2026-08-09
feat: longest-prefix dictionary fallback and Turkish dotted/dotless-I lookup

metoli86:
Follow-up promised when closing #2595 in favor of Slim dictionary (#2583): the two pieces targeting languages the mini-stemmer can't reach. Improves #2916.

## What changes

Lookup now tries, in order: exact match → **Turkish I-fold reading** → English stem variants → **longest-prefix stem fallback**.

**1. Longest-prefix stem fallback.** During the existing exact-match scan toward the target's insertion point, `locate()` remembers the longest headword that is a proper prefix (≥ 3 letters, counted in codepoints) of the target. If the exact probe and stem variants all miss, that prefix match is returned — the definition screen already shows the matched headword, so the reader sees what was actually found.

- Turkish `kitaplarımızdan` → `kitap`; German declensions like `Verhandlungen` → `Verhandlung` (#2916); Finnish/Hungarian get the same benefit. Suffix-stripping rule lists can't cover these.
- **Zero extra SD reads**: tracking rides the scan window the exact probe already walks (≤ SAMPLE_INTERVAL entries after the qidx bisect). A stem sitting further than that behind the insertion point is not seen — in practice a stem sorts directly among its own derived forms, so the window covers it.
- Prefixes of a word are nested, so the last one seen in the sorted scan is the longest — a plain overwrite, no extra state.

**2. Turkish dotted/dotless-I second reading.** ASCII case folding maps `I`→`i`, but Turkish pairs `I` with `ı` (U+0131) and `İ` (U+0130) with `i`. Words containing either now get a second probe with the Turkish fold applied — `İstanbul` finds `istanbul`, `ILIK` finds `ılık` instead of the unrelated `ilik`. The fold runs on the raw word (after `cleanWord`'s tolower the original `I` is gone) and only costs a probe when the word actually contains `I`/`İ` and the exact match missed.

## What this does *not* cover

- German infix/umlaut forms (`fortzusetzen` → `fortsetzen`, `hätten` → `haben`) — those need language-aware morphology, out of scope for a slim reader.
- Full non-ASCII case folding (all-caps `IŞIK` still has Ş left unfolded) — pre-existing limitation, unchanged here.

## Testing

- Pure matching helpers extracted to `src/util/DictionaryMatch.h` (HAL-free, same pattern as `OpdsFilename`) with a new host suite `test/dictionary_match/` — I-fold mappings (including the `ğ` = 0xC4 0x9F lead-byte near-miss and a truncated trailing lead byte) and prefix semantics (proper-prefix, ASCII-case-fold, multibyte byte-exactness).
- `locate()` wiring compiles clean under `-Wall -Wextra`; firmware build via CI.


metoli86:
Good catch @coderabbitai — fixed in ee313e58 by making the implementation match the docs rather than the other way around: the minimum is now counted in codepoints (`utf8Length`, non-continuation bytes), so "at least 3 letters" is accurate for non-ASCII scripts too. A byte threshold was also inconsistent on its own terms: two-letter `aş` (3 bytes) passed while two-letter `at` (2 bytes) didn't. Test added.

PR #2943, opened 2026-08-09
feat: retain custom sleep screen during wake

BrigBryu:
### Summary

When `Sleep Screen` is set to `Custom`, retain the existing e-ink sleep image while waking instead of drawing the normal boot splash before restoring the reader/Home UI.

This extends CrossPoint's existing splash-suppressed/seamless display initialization behavior to custom sleep screens.

### Why

On e-ink, the sleep image remains physically visible without power. Redrawing a boot splash during every normal wake creates an unnecessary intermediate refresh. Keeping the retained frame visible provides a more seamless sleep/wake experience.

### Behavior

Before:

`custom sleep image → CrossPoint boot splash → book/Home`

After:

`custom sleep image → book/Home`

Normal cold boots and ordinary restarts continue to use the normal CrossPoint splash.

### Implementation

The change intentionally reuses the existing splashless/seamless display initialization path rather than introducing new panel-driver behavior.

A stale transient Quick Resume framebuffer is removed when entering Custom sleep so it cannot replace the selected custom wallpaper during wake.

### Testing

* clang-format 21 passes
* cppcheck passes with no defects
* all 130 host unit tests pass
* the regular X4/default firmware environment builds successfully
* the equivalent v1.5.0 implementation has been physically tested successfully on a regular Xteink X4

The `src/main.cpp` change on this branch is identical to the hardware-tested v1.5.0 patch because that upstream file has not changed since v1.5.0. This PR branch is based on current `develop` and has been built and host-tested, but the complete development-branch firmware has not itself been flashed to hardware.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -123,9 +123,9 @@ constexpr uint32_t SILENT_REBOOT_TARGET_READER = 1;
 // flows suppress the splash and leave the panel holding its pre-boot frame; a
 // plain boot shows the splash. See setup() for the resolution.
 enum class BootResume : uint8_t {
-  Splash,       // cold boot, flash, panic, or plain reboot
-  Silent,       // heap-defrag ESP.restart() (RTC flag; lost on power loss)
-  QuickResume,  // wake from a quick-resume deep sleep (SD flag; survives power loss)
+  Splash,          // cold boot, flash, panic, or plain reboot
+  Silent,          // heap-defrag ESP.restart() (RTC flag; lost on power loss)
+  SplashlessWake,  // wake from deep sleep with the splash suppressed by the SD flag
 };
 
 // Latched true once enterDeepSleep() commits to sleeping, before it tears down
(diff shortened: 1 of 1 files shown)


Uri-Tauber:
I think UX-wise the user will want to see some feedback.
@uxjulia What do you think?

uxjulia:
> I think UX-wise the user will want to see some feedback. @uxjulia What do you think?

I just tested this on device and made one change, which will eliminate the need for separate visual feedback. The previous implementation required the button to be released before waking. For short-press, this is fine. For long-press, users could press and hold forever waiting for visual feedback. 

With this change, I tested it side by side with the old splash screen enabled and both devices react in roughly the same amount of time (~2 seconds for the screen to change), but this branch just removes the splash screen. Looks good to me with the change from a UX perspective.

Uri-Tauber (approved):

Uri-Tauber:
Thanks @uxjulia.

RouquinBlanc:
Note that I'm quite annoyed by this change because I never know if I press long enough for the device to turn on. Half of the time I release it too quickly and nothing happens... So I've taken the habit to hold the power button until seeing my book, which on current book + X4 + 1.6.0rc gives me ~6 seconds holding the button.

If not redrawing everything, _some_ kind of visual feedback would be appreciated (along with a faster start, it seems to have degraded between 1.4.0 and now (and I switched to get the dictionary)

Uri-Tauber:
@RouquinBlanc Please mention it in #2964. I'm afraid its going to get lost here.

PR #2946, opened 2026-08-10
chore: update translations

Uri-Tauber:
I'm going to direct here all translations-related PRs, and merge before the next release. 
The commits history become very hard to navigate when every one-line addition have its own commit...


lpla:
This PR branch needs to be updated with latest develop changes, as there are new strings in there that didn't reach this PR and users cannot propose translations for them. See this one as an example: https://github.com/crosspoint-reader/crosspoint-reader/pull/3165

PR #2960, opened 2026-08-10
fix: MAC address reading when WiFi is off

itsthisjustin:
Replace WiFi.macAddress() with esp_read_mac() to read the hardware MAC address directly. WiFi.macAddress() requires the STA network interface to exist, which fails when WiFi is off. A race condition between ui updates and wifi radio on caused the mac address to display as all zeros. Add error handling for MAC read failures.

### DIFF
--- a/src/activities/network/WifiSelectionActivity.cpp
+++ b/src/activities/network/WifiSelectionActivity.cpp
@@ -5,6 +5,7 @@
 #include <I18n.h>
 #include <Logging.h>
 #include <WiFi.h>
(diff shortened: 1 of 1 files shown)


Uri-Tauber:
Can you link the issue this close?

Uri-Tauber (approved):
The code LGTM, just link the relevant issues before merging

itsthisjustin:
> The code LGTM, just link the relevant issues before merging

I don't think there is one. It was reported on discord.

PR #2973, opened 2026-08-11
chore: remove unnesessary swallowing code

Uri-Tauber:
## Summary

AI coding agents love piling bloat onto downstream code instead of just fixing the upstream bug.

This PR cleans up one class of that.
The bug is dumb simple. One check does `isPressed`, the one next activity does `wasReleased` — so a single press fires as two events, press then release.

The real fix is one word: `isPressed` → `wasReleased`. Instead, agents (Claude Code, Codex, etc.) bolt on ~20 lines of swallowing logic to eat the stray release. Fragile, and it bloats the firmware for nothing. #2969 is a good example — it patches the symptom with such swallowing logic, but the bug only exists because someone swallowed the release in the reader menu instead of making `popupOption` react to it directly.

---

### AI Usage

### DIFF
--- a/src/activities/reader/EpubReaderActivity.cpp
+++ b/src/activities/reader/EpubReaderActivity.cpp
@@ -500,13 +500,38 @@ void EpubReaderActivity::loop() {
     requestUpdate();
   }
 
+  const bool confirmReleased = mappedInput.wasReleased(MappedInputManager::Button::Confirm);
+  if (confirmReleased) {
+    switch (SETTINGS.longPressMenuFunction) {
+      case CrossPointSettings::LP_MENU_BOOKMARK:
+        if (mappedInput.getHeldTime() >= ReaderUtils::BOOKMARK_HOLD_MS) {
+          addBookmark();
+          showBookmarkMessage = true;
+          bookmarkMessageTime = millis();
+          requestUpdate();
+          return;
+        }
+        break;
+      case CrossPointSettings::LP_MENU_KOSYNC:
+        if (mappedInput.getHeldTime() >= ReaderUtils::GO_HOME_MS && launchKOReaderSync()) return;
+        break;
+      case CrossPointSettings::LP_MENU_DICTIONARY:
+        if (mappedInput.getHeldTime() >= ReaderUtils::BOOKMARK_HOLD_MS) {
+          openDictionaryWordSelect();
+          return;
+        }
+        break;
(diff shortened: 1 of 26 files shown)


Uri-Tauber:
@coderabbitai check if there are other similar swallowing code blocks in the project, that I might missed.

src/activities/reader/EpubReaderActivity.cpp
```diff
@@ -486,13 +486,40 @@ void EpubReaderActivity::loop() {
...
+        if (mappedInput.getHeldTime() >= ReaderUtils::BOOKMARK_HOLD_MS) {
+          openDictionaryWordSelect();
+          return;
+        }
+        break;
+      case CrossPointSettings::LP_MENU_DISABLED:
+      default:
+        break;
+    }
+  }
```
thiagokokada, line 528: Any reason why only the `LP_MENU_KOSYNC` case has a different held time than the other ones. Asking this because if they were all the same, we could simplify this code by checking `const bool confirmReleased = mappedInput.wasReleased(MappedInputManager::Button::Confirm) && mappedInput.getHeldTime() >= ReaderUtils::BOOKMARK_HOLD_MS` instead.

CC @coderabbitai.

thiagokokada: @coderabbit Yes, I assumed the thresholds were different. My question is more "why"? If there is no reason, is my analysis correct or not?

Uri-Tauber:

itsthisjustin:
Haha I noticed my own llm doing this same exact thing whenever I'd report a bug related to this.

itsthisjustin (approved):

src/activities/reader/EpubReaderActivity.cpp
itsthisjustin: @Uri-Tauber Thoughts?

Uri-Tauber: I believe kosync take more time for long press on purpose (introduced in d4069aeae).
Anyway, its not a part of this PR.

thiagokokada:
@Uri-Tauber A side-effect that I am seeing of this change is, e.g., before if I held the menu button it would eventually would go to my selected shortcut (KOReader sync), but now I need to release the menu button and there is no feedback (so I sometimes keep holding the menu button until realising I need to release to trigger my action).

PR #2983, opened 2026-08-12
feat: Add support for x4pro & papermono devices 

itsthisjustin:
Adds support for the x4pro and m5stack papermono devices including frontlight controls and swipe gestures in reading along with support for the Pro's homekey. Also fixes comments by CodeRabbit related to a missing board support for x3


src/components/themes/roundedraff/RoundedRaffTheme.cpp
```diff
@@ -204,13 +177,17 @@ void RoundedRaffTheme::drawRecentBookCover(GfxRenderer& renderer, Rect rect, con
...
+}
+
 void RoundedRaffTheme::drawButtonMenu(GfxRenderer& renderer, Rect rect, int buttonCount, int selectedIndex,
                                       const std::function<std::string(int index)>& buttonLabel,
                                       const std::function<UIIcon(int index)>& rowIcon) const {
   (void)rowIcon;
   const int sidePadding = RoundedRaffMetrics::values.contentSidePadding;
   const int rowX = rect.x + sidePadding;
-  const int rowHeight = renderer.getLineHeight(kTitleFontId) + 20;  // 10px top + 10px bottom
+  const int rowHeight = getMenuRowHeight(renderer);  // shared with HomeActivity's touch grid
```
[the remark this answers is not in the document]
itsthisjustin: This finding is incorrect

itsthisjustin:
@uxjulia tagging you as Uri can't test this

platformio.ini
```diff
@@ -224,3 +225,41 @@ build_flags =
   -DCROSSPOINT_VERSION=\"${crosspoint.version}-rc+${sysenv.CROSSPOINT_RC_HASH}\"
   -DENABLE_SERIAL_LOG
   -DLOG_LEVEL=1 ; Set log level to info for release candidate builds
+
+; --- Xteink X4 Pro — ESP32-S3, 16MB flash / 8MB PSRAM, SSD1677 + GT911 --------
+; Minimal X4 Pro target: the SDK profile supplies the board pinout, display,
+; touch controller, and capacitive Home key.
+;   pio run -e x4pro -t upload
+[env:x4pro]
```
Uri-Tauber, line 233: So you decided to go with seperate envs for each device?

I do think it will be cleaner to have one env for sticky+x4pro+papermono.

Uri-Tauber: For example, we added code which prevent users from installing c3 firmware on sticky and vice-versa, but we cant identify which S3 varial is the device, so users could install x4pro firmware on sticky, and get stuck...

itsthisjustin: I don't think that's actually true though? It just means they might get an extra lighting panel on a device without a frontlight

Uri-Tauber: Does sticky come with display drivers for the three x4pro variants?

itsthisjustin: so it does in fact cause an issue for something like SDmmc. We need the different profiles or we are adding a lot of device capabilities and boot bring up for stuff that doesn't exist on other devices.

itsthisjustin: I don't think we can combine the profiles. I've looked into that a bunch to reduce my rebuilding. it would cause SO MANY if else statements on boot and on devices like the papermono it uses a completely different power button system

Uri-Tauber: Hmmm.
Does the x4 work slower because it have x3's gyro and RTC drivers?

itsthisjustin: No but we are talking about way more destructive stuff like basic sd card bring up, the display itself, usb gpios, etc

itsthisjustin: Ok so I added a board tag in the ota image to prevent someone from flashing the wrong board. Checks for this are added to ota and sd flash and it's set to accept firmware builds without the tag so alternative firmware isn't forced to do this, plus older versions of the software can still update to the traditional tag if missing. So fails on missmatch, passes on missing entirely, passes on matching.

unnamedd:
@itsthisjustin tomorrow I will test on my X4 Pro and let you know the outcome.

itsthisjustin:
> @itsthisjustin tomorrow I will test on my X4 Pro and let you know the outcome.

Thank you!

platformio.ini
Uri-Tauber: Sounds good.

itsthisjustin: If you can check all the ci and yml stuff I added, the device support is done and tested on the hardware and already locally merged with develop to validate that. I just need an extra set of eyes on build pipeline for the additions and ota file names

Uri-Tauber (approved):
Approving based on users reporting its working. Haven't looked at the code yet.

PR #2988, opened 2026-08-12
fix(ui): keep the option popup's closing press from reaching its host

jetaudio:
## Summary

  Stop the button press that closes an `OptionPopup` from being handled a second
  time by the activity behind it.

  `OptionPopup` closes on the button **press** edge, while the activities hosting
  it act on the **release** edge. The release that dismissed the popup therefore
  falls through to the host and gets handled again:

  - **Settings → Text**: closing a picker with Back also leaves the Text settings
    screen (`TextSettingsActivity::loop()` exits on `wasReleased(Back)`).
  - **Reader → Bookmarks**: cancelling the delete confirmation with Back also
    closes the bookmarks list (`EpubReaderBookmarksActivity::loop()` finishes on
    `wasReleased(Back)`).

  `EpubReaderMenuActivity` already worked around exactly this with a local
  `popupClosing` latch. This PR moves that behaviour into `OptionPopup` itself, so
  every host gets it: the popup keeps owning the input until the button that
  closed it comes back up, so one press is handled by one layer. The local latch
  in `EpubReaderMenuActivity` is then redundant and is removed.

  Closing by tap arms nothing — a tap has no split press/release across the two
  layers.

## Scope Check

- [x] The stock firmware does not already handle this well — it handles it in one host and not the others.

### Behaviour by host

Hosts that already act on the press edge (`SettingsActivity`,
`StatusBarSettingsActivity`, `ClearCacheActivity`, `ConfirmationActivity`) never
saw the trailing release, so nothing changes for them — the guard just makes them
immune if they ever move to release edges.

### Cost

Two extra members on `OptionPopup` (a `bool` and a `Button`), no allocation, no
new dependency. Net −24/+30 lines with the local latch removed.

### Testing

- [x] `clang-format`
- [x] Every `OptionPopup` host compiles clean: `TextSettingsActivity`,
      `StatusBarSettingsActivity`, `SettingsActivity`, `ClearCacheActivity`,
      `ConfirmationActivity`, `EpubReaderMenuActivity`, `EpubReaderBookmarksActivity`
- [ ] Tested on Xteink X3 hardware
- [x] Tested on Xteink X4 hardware

### Reviewer focus

1. `swallowClosingRelease()` returning `true` while the closing button is still
   held — a long hold on Back after closing a popup is swallowed until release,
   which is the intent (it must not become a long-press-to-home in the host).
2. Reopening the popup before the closing button is released: all three `show()`
   overloads clear the latch, so the new popup owns the input from that point.

---

### AI Usage

AI tools were used for the investigation, the patch, and this PR description.
The result was reviewed, formatted, and built locally by the contributor.

### DIFF
--- a/src/activities/reader/EpubReaderMenuActivity.cpp
+++ b/src/activities/reader/EpubReaderMenuActivity.cpp
@@ -65,24 +65,10 @@ bool EpubReaderMenuActivity::handleHomeGesture() {
 }
 
 void EpubReaderMenuActivity::loop() {
-  if (optionPopup.handleInput(mappedInput, [this] { requestUpdate(); })) {
-    // The popup acts on button press; if that input closed it, the trailing
-    // release must be swallowed below (Back would close the menu, Confirm
-    // would re-activate the selected item).
-    popupClosing = !optionPopup.isActive();
-    return;
-  }
-  if (popupClosing) {
-    if (mappedInput.isPressed(MappedInputManager::Button::Back) ||
-        mappedInput.isPressed(MappedInputManager::Button::Confirm)) {
-      return;  // closing press still held
-    }
-    popupClosing = false;
-    if (mappedInput.wasReleased(MappedInputManager::Button::Back) ||
-        mappedInput.wasReleased(MappedInputManager::Button::Confirm)) {
-      return;  // swallow the release that closed the popup
-    }
-  }
+  // OptionPopup keeps owning the input until the button that closed it comes back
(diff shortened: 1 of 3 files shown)


jetaudio:
Superseded: develop's OptionPopup now closes on the release edge (`wasReleased(Confirm/Back)`) and the `popupClosing` latch in EpubReaderMenuActivity is gone, so the double-handling this fixed no longer occurs. Closing.

PR #2996, opened 2026-08-13
fix: avoid full-clock spin while a section build is parked at its watermark

BrianPugh:
## Summary
Cherry-pick of [the fix mentioned here](https://github.com/crosspoint-reader/crosspoint-reader/pull/2525#issuecomment-5274356229) for high power usage in v1.5.0. Just making this PR incase you want a smaller change for an immediate fix.

skipLoopDelay() keyed only on section->isBuilding(), but loop()'s background-build pump also requires the reader to be within BUILD_WINDOW_AHEAD pages of the build watermark (or the section to be a partial). Once a first-open build lays out its look-ahead window it parks: isBuilding() stays true while loop() stops pumping it, so skipLoopDelay() forced full clock + yield() with no delay for the whole time the reader sits on a page mid-chapter -- defeating the loop's idle downclock + delay(50) path and pinning the core at full speed doing no build work.

Gate skipLoopDelay() on the same watermark window as the build pump so it returns full-speed only when a build tick will actually run this pass.

## Scope Check

Please confirm:

      the relevant maintainer.

## Additional Context

https://github.com/crosspoint-reader/crosspoint-reader/pull/2525 is still the complete power solution for ~3x more battery life, this fix should just bring v1.5.0 power consumption more inline with what it was on v1.4.0 (9.72mA on an idle page). Without this we were seeing 31.13mA on an idle page.

If we merge this in, we'll have to resolve the trivial merge-conflict in #2525

---

### AI Usage

### DIFF
--- a/src/activities/reader/EpubReaderActivity.h
+++ b/src/activities/reader/EpubReaderActivity.h
@@ -224,7 +224,17 @@ class EpubReaderActivity final : public Activity {
   // it from page 0. Reverts to normal power behavior the moment the build finishes,
   // and while the build is heap-paused (no work is happening, so spinning at full
   // speed would only burn battery; the paused gate still retries every loop pass).
-  bool skipLoopDelay() override { return section && section->isBuilding() && !buildHeapPaused; }
+  // The watermark window below MUST mirror the background-build gate in loop() (the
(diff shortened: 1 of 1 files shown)


BrianPugh:
closing since #2525 was merged!

PR #2990, opened 2026-08-12
fix: backport File Transfer guard from CrossInk

lpla:
Fixes #2825

## Summary

* **What is the goal of this PR?** Backport a stability fix for existing File Transfer from the CrossInk fork.
* **What changes are included?**
  * Do not register the File Transfer serving task with the task watchdog. Arduino WebServer client and ACK waits can consume the full watchdog timeout on a weak connection; the interrupt watchdog remains enabled. This adapts Julia's [CrossInk watchdog fix](https://github.com/uxjulia/CrossInk/commit/119f0f306ae2fc1d304cc0651e9933718c16ce90), reported fixed in [CrossInk #432](https://github.com/uxjulia/CrossInk/issues/432).

## Scope Check

CrossPoint is intentionally narrow. See [SCOPE.md](../blob/master/SCOPE.md) and [ROADMAP.md](../blob/master/ROADMAP.md).
Please confirm:

- [x] I have read SCOPE.md and ROADMAP.md.
- [x] This PR is **not** a new built-in theme (themes are temporarily closed pending the move to SD-loaded themes).
- [x] This PR is **not** a new external network connector (sync engine, cloud storage, remote file access, etc.).
- [x] This PR is **not** an interactive app, writing tool, RSS/news/browser, media playback, or PDF feature.
- [x] The stock firmware does not already handle this well, **and** no other popular CrossPoint fork already does (or, if one does, I explain why CrossPoint still needs it below).
- [x] This PR does not touch `freeink-sdk/`, `lib/hal/`, the bootloader, OTA, or recovery code.

**Fork context:** CrossInk contains the source fixes, but CrossPoint has the same existing affected code paths without these safeguards. This is a narrow backport to improve stability in the core File Transfer, not a new feature.

## Additional Context

Validation completed:

* X4 simulator build.
* File Transfer simulator flow and a live `/api/status` request.
* Embedded `default` build.
* X4 hardware: the previously observed File Transfer failure no longer reproduces.

No new dependencies or persistent heap allocations are introduced. A binary-size delta was not measured.

---

### AI Usage

While CrossPoint doesn't have restrictions on AI tools in contributing, please be transparent about their usage as it helps set the right context for reviewers.

Did you use AI tools to help write this code? **PARTIALLY** - AI assisted with review and preparing this PR; implementation and test execution were performed by the author.

PR #3005, opened 2026-08-13
fix: retry low-memory CSS cache safely

lpla:
## Summary

* **What is the goal of this PR?** Prevent CSS parsing and cache loading from aborting or leaving an unusable cache when heap is constrained or input is malformed.
  * Replace the allocation-prone rules map with bounded flat storage for rule entries, selector text, and deduplicated styles. Every growth allocation is fallible.
  * Replace CSS deduplication's throwing STL scratch containers with one optional nothrow table; allocation failure falls back to parsing every stylesheet.
  * Return explicit complete, partial, and error outcomes. Oversized stylesheets and truncated selectors or declarations cannot be cached as complete.
  * Validate cache headers, lengths, and serialized enum values before hydration.
  * Write caches through a temporary file, preserve complete caches when a retry remains partial, and restore the previous backup if promotion fails.

## Scope Check

Please confirm:

- [x] The stock firmware does not already handle this well, **and** no other popular CrossPoint fork already does without this backport.

Not applicable; this is reliability work in the existing EPUB CSS path.

## Additional Context

This is the CSS follow-up separated from #2990 after its File Transfer fix merged. It began as an adaptation of Julia's [CrossInk fix](https://github.com/uxjulia/CrossInk/commit/9fb247545f1f9917a4246794f43b29ad9ff78238), then incorporated the explicit partial-cache and retry behavior discussed in #2990.

The bounded flat rule store follows the direction of #2602 while retaining this PR's cache-status, retry, validation, and rollback work. #3125 addresses only the old map-based cache loader and is therefore narrower than the current implementation here.

Regression coverage includes:

* continuing cascade updates for existing selectors after new-rule growth stops;
* marking oversized selectors, declarations, and stylesheets partial;
* complete and partial cache round trips;
* rejection of corrupted serialized style values;
* restoration of the only usable backup after a failed cache promotion.

Validation passes 157/157 host tests, cppcheck, the repository formatting wrapper, `git diff --check`, and the ESP32-C3 default build. The final default image uses 54,540 bytes static RAM and 5,709,391 bytes flash. GitHub CI also passes all four firmware targets and CodeRabbit.

No allocation in the rule-store growth path can abort on OOM: failures return a partial result and keep already accepted rules usable for the current session.

---

### AI Usage

AI tools were used to trace the earlier review findings, implement the backport, review failure paths, and run the test matrix.


lpla:
@Uri-Tauber this is the continuation of the work I was seeing that was growing too much in #2990. It ended up being a quite big patch to backport the whole hardening from CrossInk, so let's take a thought on the safety of CSS.

Uri-Tauber:
I passed your work here via agent with some special reviewing skills I'm working on. Here some suggestion it come with:

### 1 
The rule table cannot report allocation failure safely, so the PR predicts whether each future allocation will succeed. That prediction currently understands `unordered-map` buckets, nodes, strings, allocator overhead, and `libstdc++`’s rehash policy. The current patch may work, but every storage change must keep that model accurate.

WHAT’S WRONG
`lib/Epub/Epub/css/CssParser.cpp:495` hand-calculate bucket and rule allocation requirements, after which parsing and hydration still cross the throwing `unordered_map::emplace()` boundary at `lib/Epub/Epub/css/CssParser.cpp:491` and `lib/Epub/Epub/css/CssParser.cpp:1148`. The map representation is private at `lib/Epub/Epub/css/CssParser.h:166`; callers do not require it.

  BREAKS WHEN
  The next selector expansion—descendant, ID, or richer combined selectors already anticipated at `lib/Epub/Epub/css/CssParser.cpp:458` — adds another container or allocation pattern. It then needs another synchronized heap estimate with no compile-time enforcement. Repository history already contains successive changes widening CSS cache heap estimates, demonstrating this maintenance loop.
  
  
  SUGGESTION:
  Give `CssParser` a bounded, fallible rule store whose `insertOrMerge()` returns Inserted, Merged, Limit, or `OutOfMemory`. Use `makeUniqueNoThrow`-owned flat `selector`/`index`/`style` pools, adapting the suggested direction in PR #2602, while retaining this PR’s complete/partial outcomes and crash-safe replacement.

  The dynamic, book-dependent data is too large and variable for the task stack or static DRAM; use explicit caps such as a 16KB entry index plus 32KB selector pool and a bounded style pool, with every bulk allocation fallible. This removes the allocator-model formulas and thousands of small map allocations.
  
  
### 2 The cache inspector, writer, and loader each independently know the byte layout of one CSS rule. Adding a persisted CSS property means editing all three in lockstep.

  WHAT’S WRONG
  The layout is summarized by constants at `lib/Epub/Epub/css/CssParser.cpp:758`, reconstructed by `inspectCache()` at `lib/Epub/Epub/css/CssParser.cpp:798`, written field-by-field at `lib/Epub/Epub/css/CssParser.cpp:835`, and reconstructed again by `loadFromCache()` at `lib/Epub/Epub/css/CssParser.cpp:960`. A prior CSS branch already omitted a newly added `cssFloat` byte from one writer, requiring another version bump.
  
  Create one canonical fixed-size style codec: `encodeStyleWire()` and `decodeStyleWire()`, backed by a roughly 66-byte stack buffer. The writer uses it, the loader decodes it, and inspection skips exactly `STYLE_WIRE_BYTES`. Keep the separate inspect/load workflows but give the binary representation one owner.
  
### 3 Avoid querying heap statistics for every cached selector —`lib/Epub/Epub/css/CssParser.cpp:1033`

  After reserving the complete bucket array, every one of up to 1,500 rules still calls `prepareRuleInsertion()`, issuing both free-heap and largest-block queries. This repeats whenever an uncached section hydrates CSS. Amortize the node-allocation checks using a conservative batch budget while retaining the 64 KB safety floor.

### 4 Avoid scanning complete caches twice during EPUB load — `lib/Epub/Epub.cpp:401`

  `inspectCache()` walks the entire cache payload, after which `loadFromCache()` reads it again. Partial caches are similarly inspected here and again inside `parseCssFiles()`. Reuse the first result or limit complete-cache inspection to the header and let hydration perform full validation.
  
### 5 Do not commit section caches after CSS hydration fails — `lib/Epub/Epub/Section.cpp:367`

  When `loadFromCache()` returns `LowMemory`, the build merely logs and continues with an empty rule map. `finalizeBuild()` then commits pagination without external CSS. Because the complete CSS cache remains valid, future loads never invalidate this degraded section cache. Abort/suspend the section build on `LowMemory`, or mark the section as CSS-degraded so it is retried.

lpla:
Thanks for the detailed review. I agreed with the direction and reworked the implementation around a bounded rule store inspired by #2602, while retaining this PR’s partial-cache recovery and crash-safe replacement behavior.

The five points from the agent are now addressed:

1. The `unordered_map` and allocator prediction are gone. Rules use a sorted, fallible flat store with a maximum 1,500-entry index, a 32 KB selector pool, and a 256-entry deduplicated style pool. Growth uses `makeUniqueNoThrow`, and insertion reports `Inserted`, `Merged`, `Limit`, or `OutOfMemory`.

2. Cache style serialization now has one owner through the fixed-size `encodeStyleWire()` / `decodeStyleWire()` codec. Inspection skips exactly `STYLE_WIRE_BYTES`, while writing and hydration use the same representation.

3. The per-selector heap-stat queries are gone.

4. Complete caches receive a header-only inspection and are fully validated during hydration. Partial caches still receive a full inspection because we need to determine whether they can safely be preserved for another low-memory retry.

5. If CSS hydration returns `LowMemory`, the section build now aborts and removes its temporary files instead of promoting an unstyled section cache.

Two additional bounded-store edge cases were covered during the follow-up:

- Repeated overrides of an unshared selector reuse its style slot instead of consuming the bounded pool with obsolete intermediate styles. Shared styles remain copy-on-write.
- After new-rule growth stops, declarations for selectors already in the store continue to participate in the cascade. A binary exact-match probe allows those merges while unseen selectors are skipped without retrying failed allocations.

The branch is updated with current `develop` and is conflict-free.

lpla:
@coderabbitai full review

Uri-Tauber (approved):
LGTM

Issue #2879, opened 2026-08-05
Custom sleep screen renders, then refreshes to solid white when the image contains greyscale (X3, v1.4.1)


busseo:
### Affected Version

v1.4.1 (Community), Xteink X3

Describe the Bug

Any custom sleep image containing greyscale pixels (e.g. antialiased text or a greyscale palette) briefly renders and then the panel refreshes to solid white. The same happens in the BMP viewer. The default CrossPoint sleep logo and pure black-and-white images render fine.

Tracing the render path, the difference is in [`SleepActivity.cpp` L217–L251](https://github.com/crosspoint-reader/crosspoint-reader/blob/fecb9eb3a468/src/activities/boot_sleep/SleepActivity.cpp#L217-L251): when `bitmap.hasGreyscale()` is true and Sleep Screen Cover Filter is "None", rendering goes through the multi-pass grayscale pipeline (`displayGrayscaleBase` → LSB/MSB passes → `displayGrayBuffer`). On my unit that sequence ends with a white panel — the visible "flash" is the base
HALF-refresh pass, and the grey passes then wipe it. Images with only pure black/white pixels take the single-pass `displayBuffer(HALF_REFRESH)` branch (same as the stock logo) and work every time.

Consistent with this, either of these workarounds fixes it on my unit:
- converting the image to strict 1-bit black/white (no grey pixels or palette entries), or
- setting Sleep Screen Cover Filter to "Contrast"/"Inverted" (both bypass the grayscale branch).

Possibly related reports: #307 (custom wallpapers not showing, closed cant-replicate — the reporter's covers worked, matching the hasGreyscale split), #418 (double flash on sleep image), #1110 (grey conversion delay).

Steps to Reproduce

1. Create a 528×792 24-bit BMP that includes greyscale pixels (e.g. white background + black antialiased text) and copy it to the SD root as `sleep.bmp`
2. Settings → Sleep Screen → Custom; Sleep Screen Cover Filter → None
3. Sleep the device: the image appears for a moment, then the screen refreshes to solid white
4. Threshold the same image to pure black/white (no greys), replace `sleep.bmp`, sleep again: renders correctly and persists

Expected Behavior:
Greyscale sleep images render and persist, as the 4-level pipeline intends.

dmellok:
Second hardware datapoint: **Xteink X4, same pipeline, opposite symptom — the greys render as solid black instead of the panel wiping to white.**

[...]

PR #3007, opened 2026-08-13
fix: request a clean grayscale base on X4 sleep screens

dmellok:
Fixes the X4 half of #2879: a sleep screen whose image contains grey renders every grey level as solid black.

## Cause

The sleep grayscale pipeline paints a B/W base, writes the LSB/MSB planes, then fires the AA nudge refresh (`SleepActivity::renderBitmapSleepScreen`). The planes are built correctly — decoding the repro BMP through `Bitmap::parseHeaders`/`readNextRow` yields all four levels (`nativePalette == true`, ~108k level-1 and ~108k level-2 pixels on a 480x800 test card), so `drawBitmap()` in `GRAYSCALE_LSB`/`GRAYSCALE_MSB` is not at fault.

The base is. `HalDisplay::displayGrayscaleBase()` gated its "a HALF fallback means the caller wants a clean base" resync on `gpio.deviceIsX3()`:

```cpp
if (gpio.deviceIsX3() && fallback == RefreshMode::HALF_REFRESH) {
  einkDisplay.requestResync(1);
}
```

On X4 the base therefore ran as a differential DU partial against whatever the panel last showed (home or reader). The UC8279 AA LUTs are calibrated against a GC-conditioned pixel state, so the nudge refresh lands invisibly and every grey stays at the black the base pass painted it.

This matches the history: the grayscale branch here used `FULL_REFRESH` only between #2471 and #2588, and that window is the last one in which X4 greys rendered.

## Fix

Drop the X3 gate so the clean-base request is panel-agnostic. On X4 this produces the same waveform `FULL_REFRESH` did — `Uc8279X4Driver::displayStart` takes the identical `!fast` branch (white-seeded DTM1 + GC absolute waveform) for `mode == Full` and for `_needFullClear`, which is what `requestResync()` sets:

```cpp
const bool fast = (mode != RefreshMode::Full) && !_needFullClear && _oldPlaneValid;
```

X3 behaviour is unchanged (it already made this call), and drivers with no differential baseline no-op `requestResync()`.

Scope is limited to the grayscale base. The single-pass B/W sleep and boot paints keep the HALF stock-parity waveform from #2588, and the reader's `FAST` grayscale base stays on the cheap differential path, so per-page AA is untouched.

## Testing

- `pio run` (default env) — SUCCESS. RAM 16.5%, Flash 86.6%. No RAM cost: the change is one condition.
- Xteink X4, flashed from this branch. A 4-level test card (palette `#000000/#555555/#AAAAAA/#FFFFFF`) as `/sleep.bmp` with Sleep Screen Cover Filter = None now renders black / dark grey / light grey / white, and persists through deep sleep. Before this change the same card rendered black / black / black / white.
- Non-grey sleep images take the `hasGreyscale == false` branch and are unaffected.

Tradeoff worth calling out: a sleep image containing grey now costs one GC base flash where it previously ran a silent DU partial. That is the #2471/#2588 tension, but it only applies to images that actually enter the grayscale branch, and without it those images do not render at all.

## Not covered

The X3 symptom in #2879 (grey sleep images wipe to solid white) is a separate fault. X3 already requested the resync, so its base is already clean; the failure is downstream in the gc nudge bank, in the SDK submodule. Unchanged here.

### DIFF
--- a/lib/hal/HalDisplay.cpp
+++ b/lib/hal/HalDisplay.cpp
@@ -108,14 +108,18 @@ void HalDisplay::copyGrayscaleBuffers(const uint8_t* lsbBuffer, const uint8_t* m
 }
 
 void HalDisplay::displayGrayscaleBase(RefreshMode fallback, bool turnOffScreen) {
-  // X3: a HALF fallback means the caller wants a clean base (e.g. the sleep
-  // cover, a full-screen swap from arbitrary prior content). Without this, the
-  // X3 grayscale base takes its gentle differential happy path and the prior
-  // home/reader frame ghosts through the soft aa_pre_bw_mid waveform. Forcing a
-  // resync makes displayGrayscaleBase clear first, matching displayBuffer(HALF).
+  // A HALF fallback means the caller wants a clean base (e.g. the sleep cover,
+  // a full-screen swap from arbitrary prior content), so request a resync on
+  // every panel; drivers with no differential baseline no-op the request.
+  // Without it the base takes the differential happy path: on X3 the prior
+  // home/reader frame ghosts through the soft aa_pre_bw_mid waveform, and on X4
+  // the base runs as a DU partial diff, leaving pixels in a charge state the
+  // AA nudge LUTs are not calibrated for — the LSB/MSB planes then land
+  // invisibly and greys stay at their black base. The resync makes
+  // displayGrayscaleBase clear first, matching displayBuffer(HALF).
   // The reader's FAST path is deliberately left on the differential path so
   // per-page grayscale stays cheap.
-  if (gpio.deviceIsX3() && fallback == RefreshMode::HALF_REFRESH) {
+  if (fallback == RefreshMode::HALF_REFRESH) {
     einkDisplay.requestResync(1);
   }
 


theZiz:
I tested your change on my X4 and nothing changed -- neither better nor worse.

However, I am confused anyway, but maybe I am overlooking something.

The original X4 is using the Ssd1677Driver, where `requestResync` is a noop anyway. But I read that some newer or repaired X4 are using a different controller, the UC8179, which driver indeed enforces an FULL REFRESH. Are refering to it?

dmellok:
You were right to push back, and I owe you an apology — you spent time flashing and testing a patch that could never have done anything, because I never verified the premise. Sorry for that.

Closing this. There was no bug. Here's the full post-mortem.

**Your question answers itself.** I probed my own X4 and it's SSD1677, same as yours:

```
[XTDET] no NVS screenType; bus probe VER=FF FF FF FF FF FLG=FF -> default controller
[MAIN] Hardware detect: X4
```

`PrimaryAssumed`, not `Uc81xxConfirmed`, so `applyXteinkDisplayController()` returns early and `selectDriver()` lands on `ssd1677Driver()` — where `requestResync()` is the inherited `PanelDriver` no-op. The patch is inert on my hardware exactly as it is on yours. So my claimed before/after could not have come from this change, which I should have caught before opening the PR rather than after you tested it.

**What actually produced the black bands.** I had **Sleep Screen Cover Filter** set to *Black and White*. That flag doesn't just post-process — it disables the greyscale pipeline outright, in `SleepActivity::renderBitmapSleepScreen`:

```cpp
const bool hasGreyscale =
    bitmap.hasGreyscale() && (preserveBackground || SETTINGS.sleepScreenCoverFilter ==
                                  CrossPointSettings::SLEEP_SCREEN_COVER_FILTER::NO_FILTER);
```

With `hasGreyscale` false there's no `displayGrayscaleBase()`, no LSB/MSB planes, no AA nudge — it falls through to a plain 1-bit `displayBuffer(HALF_REFRESH)`. And the BW path in `GfxRenderer` is `if (renderMode == BW && val < 3)`, i.e. every level that isn't pure white is drawn black. My test card's levels 0/1/2/3 therefore render black / black / black / white. That is the entire symptom, reproduced exactly, with no driver involvement at all.

I then reformatted my SD card, which wiped `/.crosspoint/` and regenerated settings at defaults — `sleepScreenCoverFilter = NO_FILTER`. The greyscale path switched back on, the test card rendered correctly, and I attributed that to my patch. It was a settings reset. The code I was "fixing" never executed in either run.

This also explains why the card rendered fine for you on stock nightly: you were on defaults the whole time.

**The one thing worth keeping.** A non-default cover filter silently disabling greyscale rendering entirely is genuinely confusing — the setting reads like a post-process, not a pipeline switch, and there's no indication in the UI that picking it turns anti-aliased sleep covers into hard 1-bit. I'll open a separate issue for documenting or renaming it. That interaction is the only real finding here, and it came out of your question, not my patch.

Thanks for testing it and for asking the obvious question I should have asked myself.

theZiz:
No problem, I am learning about the whole grayscale rendering process myself right now and tbh your pull request helped a lot in understanding more of it. :)

PR #3009, opened 2026-08-13
fix: clear sleep image after wake

### LINKED ISSUE #3002: Sleep screen image overlays menu after wake-up
### Affected Version

1.5.0-dev+9b1fb71, 13/08/2026

### Describe the Bug

When waking the X4 up from sleep, the "booting" logo doesn't appear anymore (which is a new, welcome feature #2989) and the device switches from the sleep image directly to the main menu. 

However in the current nightly build, the sleep image is not cleared, so the main menu is drawn on top of it and stays like this until you open a book, which fixes the issue until the next wake-up.

Image

### Steps to Reproduce

1. Hold the X4 side button to put the device to sleep
2. Hold the button again to wake the device
3. The main menu is now drawn on top of the sleep screen image

### Expected Behavior

The sleep screen image should be replaced by the usual white menu background.

```shell

```

thiagokokada:
## Summary

* **What is the goal of this PR?** Fix the sleep-screen image remaining visible underneath the Home screen after a splashless wake.
  * Pass a one-shot clean-initial-refresh option to HomeActivity for splashless wakes without a retained Quick Resume frame.
  * Use `HALF_REFRESH` for that first Home render instead of issuing a second post-render refresh.
  * Preserve Quick Resume and normal Home navigation behavior.

Fix https://github.com/crosspoint-reader/crosspoint-reader/issues/3002.

## Scope Check

- [x] The stock firmware does not already handle this well, and no other popular CrossPoint fork already does.

## Additional Context

Tested in Xteink X4:
- Custom sleep image + waking up from main menu
  + Double checked that we are only refreshing once (the first fix I did the screen flashed twice)
- Custom sleep image + waking up from book (already working, just checking for regressions)
- Quick sleep + waking up from book (not really related to the changes here but just to make sure there were no regressions)

---

### AI Usage

### DIFF
--- a/src/activities/ActivityManager.cpp
+++ b/src/activities/ActivityManager.cpp
@@ -256,7 +256,7 @@ void ActivityManager::goToFullScreenMessage(std::string message, EpdFontFamily::
   replaceActivity(std::make_unique<FullScreenMessageActivity>(renderer, mappedInput, std::move(message), style));
 }
 
-void ActivityManager::goHome(HomeMenuItem initialMenuItem) {
+void ActivityManager::goHome(HomeMenuItem initialMenuItem, bool cleanInitialRefresh) {
   if (initialMenuItem == HomeMenuItem::NONE && currentActivity) {
     const auto& activityName = currentActivity->name;
     if (activityName == "FileBrowser") {
@@ -271,7 +271,7 @@ void ActivityManager::goHome(HomeMenuItem initialMenuItem) {
       initialMenuItem = HomeMenuItem::SETTINGS_MENU;
     }
   }
-  replaceActivity(std::make_unique<HomeActivity>(renderer, mappedInput, initialMenuItem));
+  replaceActivity(std::make_unique<HomeActivity>(renderer, mappedInput, initialMenuItem, cleanInitialRefresh));
 }
 void ActivityManager::goToCrashReport() { replaceActivity(std::make_unique<CrashActivity>(renderer, mappedInput)); }
 
--- a/src/activities/ActivityManager.h
+++ b/src/activities/ActivityManager.h
@@ -91,7 +91,7 @@ class ActivityManager {
   void goToBoot();
   void goToFullScreenMessage(std::string message, EpdFontFamily::Style style = EpdFontFamily::REGULAR);
   void goToCrashReport();
-  void goHome(HomeMenuItem initialMenuItem = HomeMenuItem::NONE);
+  void goHome(HomeMenuItem initialMenuItem = HomeMenuItem::NONE, bool cleanInitialRefresh = false);
 
   // This will move current activity to stack instead of deleting it
   void pushActivity(std::unique_ptr<Activity>&& activity);
--- a/src/activities/home/HomeActivity.cpp
+++ b/src/activities/home/HomeActivity.cpp
@@ -4,6 +4,7 @@
 #include <Epub.h>
 #include <FsHelpers.h>
 #include <GfxRenderer.h>
+#include <HalDisplay.h>
 #include <HalStorage.h>
 #include <I18n.h>
 #include <Utf8.h>
@@ -335,7 +336,7 @@ void HomeActivity::render(RenderLock&&) {
                                             tr(STR_DIR_DOWN));
   GUI.drawButtonHints(renderer, labels.btn1, labels.btn2, labels.btn3, labels.btn4);
 
-  renderer.displayBuffer();
+  renderer.displayBuffer(cleanInitialRefresh && !firstRenderDone ? HalDisplay::HALF_REFRESH : HalDisplay::FAST_REFRESH);
 
   if (!firstRenderDone) {
     firstRenderDone = true;
--- a/src/activities/home/HomeActivity.h
+++ b/src/activities/home/HomeActivity.h
@@ -29,6 +29,7 @@ class HomeActivity final : public Activity {
   int coverRectH = 0;
   std::vector<RecentBook> recentBooks;
   const HomeMenuItem initialMenuItem;
+  const bool cleanInitialRefresh;
 
   // Convert HomeMenuItem to menu index (used in onEnter)
   static int menuItemToIndex(HomeMenuItem item, bool hasOpdsUrl) {
@@ -71,8 +72,10 @@ class HomeActivity final : public Activity {
 
  public:
   explicit HomeActivity(GfxRenderer& renderer, MappedInputManager& mappedInput,
-                        HomeMenuItem initialMenuItemValue = HomeMenuItem::NONE)
-      : Activity("Home", renderer, mappedInput), initialMenuItem(initialMenuItemValue) {}
+                        HomeMenuItem initialMenuItemValue = HomeMenuItem::NONE, bool cleanInitialRefresh = false)
+      : Activity("Home", renderer, mappedInput),
+        initialMenuItem(initialMenuItemValue),
+        cleanInitialRefresh(cleanInitialRefresh) {}
   void onEnter() override;
   void onExit() override;
   void loop() override;
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -464,6 +464,7 @@ void setup() {
                             : isSleepWake && !APP_STATE.showBootScreen ? BootResume::SplashlessWake
                                                                        : BootResume::Splash;
   bool allowFastInitialReaderRefresh = false;
+  bool needsWakeRefresh = false;
 
   setupDisplayAndFonts(resume != BootResume::Splash);
 
@@ -494,6 +495,10 @@ void setup() {
         } else {
           renderer.displayBuffer(HalDisplay::HALF_REFRESH);
         }
+      } else {
+        // The first Home/Reader paint is followed by an explicit clean refresh
+        // because the panel still physically shows the sleep image.
+        needsWakeRefresh = true;
       }
       break;
     case BootResume::Splash:
@@ -523,7 +528,7 @@ void setup() {
              mappedInputManager.isPressed(MappedInputManager::Button::Back) || APP_STATE.readerActivityLoadCount > 0) {
     // Boot to home screen if no book is open, last sleep was not from reader, back button is held, or reader activity
     // crashed (indicated by readerActivityLoadCount > 0)
-    activityManager.goHome();
+    activityManager.goHome(HomeMenuItem::NONE, needsWakeRefresh);
   } else {
     // Clear app state to avoid getting into a boot loop if the epub doesn't load
     const auto path = APP_STATE.openEpubPath;


theZiz:
I wonder why I don't have the same problem on my X4. Do you know what kind of display controller your X4 has?

thiagokokada:
> I wonder why I don't have the same problem on my X4. Do you know what kind of display controller your X4 has?

To be honest I have no idea. Are you using custom sleep screens? This is how I got this bug in the first place.

theZiz:
> > I wonder why I don't have the same problem on my X4. Do you know what kind of display controller your X4 has?
> 
> To be honest I have no idea. Are you using custom sleep screens? This is how I got this bug in the first place.

I played around with them the last couple of days and didn't run in such a problem. 

But why should it be different between custom sleep screens and covers? Are you using a sleep screen filter?

Regarding the display controller version: You could check the monitoring while booting, mine e.g. states:

```
[XTDET] bus probe VER=FF FF FF FF FF FLG=FF -> default controller
```

If yours is not SSD1677, is will say something about UltraChip and give details about the specific version of the UltraChip afterwards.

itsthisjustin:
You see ghosting on the pro or regular x4? Custom sleep screens DO seem to do this more than regular covers and I think it's due to the way the file is setup and white pixels vs gray

itsthisjustin:
I'm in favor of this, but pretty much any time we flash the screen users complain...

thiagokokada:
> You see ghosting on the pro or regular x4? Custom sleep screens DO seem to do this more than regular covers and I think it's due to the way the file is setup and white pixels vs gray

In my case regular X4. To be clear this is less like ghosting and more like the sleep screen completely overlapping with the menu (making it unusable), see the original issue.

theZiz:
> I'm in favor of this, but pretty much any time we flash the screen users complain...

I checked both versions: His and the same version but without his commit (to make the difference as small as possible).

And I didn't see any difference, my X4 is always (!) doing a half refresh after boot. I think the reason is this code:
https://github.com/Free-Ink/freeink-sdk/blob/5dd02bcfa7d34c01190d334849878cc0b4cba760/libs/display/FreeInkDisplay/src/driver/Ssd1677Driver.cpp#L429
```
void Ssd1677Driver::displayImpl(EpdBus& bus, const uint8_t* fb, const uint8_t* prev, RefreshMode mode, bool turnOff,
                                bool async) {
  // The first paint after boot/wake must be an absolute clean, not a partial/DU
  // refresh: a partial only drives pixels that differ from the RED "old" plane, so
  // it can't clear what is physically on the panel at boot (e.g. the sleep screen).
  // Stock parity: the OEM firmware's only clean primitive in normal operation is
  // the single-pass HALF (0xD7 "_updateFull"); it never runs the multi-flash OTP
  // full waveform (0xF7 is a dead fallback branch there). Promote to HALF where
  // the board has one; boards without (Sticky) promote to their vendor FULL.
  if (!turnOff) {
    if (_needsInitialFull) {
      // First paint after boot/wake must not be a differential FAST: it only drives
      // pixels that differ from the RED baseline, so it can't clear whatever is
      // physically on the panel (the boot screen) and that ghosts through. But a HALF
      // or FULL the caller already asked for is non-differential (BYPASS_RED) and
      // clears the panel + seeds the baseline on its own — honor it and just consume
      // the one-shot. Only upgrade a FAST request. This keeps the boot logo (a HALF)
      // from paying an extra multi-inversion FULL-waveform flash on top of its own.
      if (mode == RefreshMode::Fast) mode = RefreshMode::Full;
      _needsInitialFull = false;
      mode = (_cfg.halfSeqOverride != 0) ? RefreshMode::Half : RefreshMode::Full;
    } else if (!_isScreenOn && _cfg.fullSeqOverride == 0) {
      // X4-class cold start: panel asleep -> a (warmed) HALF full-clear. Override
      // boards skip this — their fast sequence self-powers, so _isScreenOn is false
      // every page and forcing HALF would make every page a slow full-waveform flash.
      mode = RefreshMode::Half;
    }
  }
  // ...
```

The the driver basically already does, what this fix here for every driver. In the SSD1677 case it doesn't change anything.

I don't see such a handling in the UC8179 driver, so I still assume, that @thiagokokada got an X4 without SSD1677 display controller. :sweat_smile: 

Anyway, tldr: I think the fix is valid and needed. It doesn't change behavior on SSD1677 devices, but fixes it for other drivers.

thiagokokada:
> Regarding the display controller version: You could check the monitoring while booting, mine e.g. states:
> 
> ```
> [XTDET] bus probe VER=FF FF FF FF FF FLG=FF -> default controller
> ```
> 
> If yours is not SSD1677, is will say something about UltraChip and give details about the specific version of the UltraChip afterwards.

```
[14:04:19] [INF] [HW] Using cached device type: X4
[14:04:19] [XTDET] NVS hw_calib/screenType: not set [info only]
[14:04:19] [XTDET] bus probe VER=FF FF FF FF FF FLG=FF -> default controller
[14:04:19] [ERR] [GYR] SDK IMU not found
[14:04:19] [INF] [CLK] RTC not found
[14:04:19] [INF] [MAIN] Hardware detect: X4
```

Strange, mine seems to be the default controller but I definitely have this issue running `develop`.

thiagokokada:
Just flashed the latest nightly directly from Crosspoint website (version `1.5.0-dev+618608e`) and can confirm I can still reproduce this issue with my X4.



Edit: also confirmed that rebasing this branch with current `develop` and flashing the result in my device fixes the issue.

theZiz:
> Strange, mine seems to be the default controller but I definitely have this issue running `develop`.

This is strange, I tested the vanilla develop again with custom cover and were not able to reproduce your issue. Could you give me your custom sleep screen, maybe it has something to do with the image format.

However, you are not using any kind of sleep screen filter, do you?

thiagokokada:
> > Strange, mine seems to be the default controller but I definitely have this issue running `develop`.
> 
> This is strange, I tested the vanilla develop again with custom cover and were not able to reproduce your issue. Could you give me your custom sleep screen, maybe it has something to do with the image format.
> 
> However, you are not using any kind of sleep screen filter, do you?

Yes, no filter. I am using those: https://github.com/dmellok/eink-screens/tree/main/Pok%C3%A9dex/x4.

theZiz:
Nice ones! Took the Raichu. Favorite Pokemon. Anyway. I think I got the problem. It was not the image, but could it be, that you have the anti sunlight fading fix activated?
imageimage

thiagokokada:
> that you have the anti sunlight fading fix activated? 

Yes, I do.

theZiz:
> > that you have the anti sunlight fading fix activated?
> 
> Yes, I do.

For some reason the first half refresh in the SSD1677 after boot is not happening if the fading fix is on. It's called `turnOff` in the driver.

```
void Ssd1677Driver::displayImpl(EpdBus& bus, const uint8_t* fb, const uint8_t* prev, RefreshMode mode, bool turnOff,
                                bool async) {
  // The first paint after boot/wake must be an absolute clean, not a partial/DU
  // refresh: a partial only drives pixels that differ from the RED "old" plane, so
  // it can't clear what is physically on the panel at boot (e.g. the sleep screen).
  // Stock parity: the OEM firmware's only clean primitive in normal operation is
  // the single-pass HALF (0xD7 "_updateFull"); it never runs the multi-flash OTP
  // full waveform (0xF7 is a dead fallback branch there). Promote to HALF where
  // the board has one; boards without (Sticky) promote to their vendor FULL.
  if (!turnOff) {
    if (_needsInitialFull) {
      // First paint after boot/wake must not be a differential FAST: it only drives
      // pixels that differ from the RED baseline, so it can't clear whatever is
      // physically on the panel (the boot screen) and that ghosts through. But a HALF
      // or FULL the caller already asked for is non-differential (BYPASS_RED) and
      // clears the panel + seeds the baseline on its own — honor it and just consume
      // the one-shot. Only upgrade a FAST request. This keeps the boot logo (a HALF)
      // from paying an extra multi-inversion FULL-waveform flash on top of its own.
      if (mode == RefreshMode::Fast) mode = RefreshMode::Full;
      _needsInitialFull = false;
      mode = (_cfg.halfSeqOverride != 0) ? RefreshMode::Half : RefreshMode::Full;
    } else if (!_isScreenOn && _cfg.fullSeqOverride == 0) {
      // X4-class cold start: panel asleep -> a (warmed) HALF full-clear. Override
      // boards skip this — their fast sequence self-powers, so _isScreenOn is false
      // every page and forcing HALF would make every page a slow full-waveform flash.
      mode = RefreshMode::Half;
    }
  }
  // ...
```


@itsthisjustin Does this even make sense? I think the `if (!turnOff) {` should be in the else-path after the "refresh on first draw"-part, which should happen even with `turnOff` activated.

So imho the fix would be better in the FreeInk SDK. I don't know whether force refreshing in the Crosspoint Reader Itself does an unnecessary refresh on other display controller types.

On the other hand, I see less risk in your change.

itsthisjustin:
> > > that you have the anti sunlight fading fix activated?
> > 
> > 
> > Yes, I do.
> 
> For some reason the first half refresh in the SSD1677 after boot is not happening if the fading fix is on. It's called `turnOff` in the driver.
> 
> ```
> void Ssd1677Driver::displayImpl(EpdBus& bus, const uint8_t* fb, const uint8_t* prev, RefreshMode mode, bool turnOff,
>                                 bool async) {
>   // The first paint after boot/wake must be an absolute clean, not a partial/DU
>   // refresh: a partial only drives pixels that differ from the RED "old" plane, so
>   // it can't clear what is physically on the panel at boot (e.g. the sleep screen).
>   // Stock parity: the OEM firmware's only clean primitive in normal operation is
>   // the single-pass HALF (0xD7 "_updateFull"); it never runs the multi-flash OTP
>   // full waveform (0xF7 is a dead fallback branch there). Promote to HALF where
>   // the board has one; boards without (Sticky) promote to their vendor FULL.
>   if (!turnOff) {
>     if (_needsInitialFull) {
>       // First paint after boot/wake must not be a differential FAST: it only drives
>       // pixels that differ from the RED baseline, so it can't clear whatever is
>       // physically on the panel (the boot screen) and that ghosts through. But a HALF
>       // or FULL the caller already asked for is non-differential (BYPASS_RED) and
>       // clears the panel + seeds the baseline on its own — honor it and just consume
>       // the one-shot. Only upgrade a FAST request. This keeps the boot logo (a HALF)
>       // from paying an extra multi-inversion FULL-waveform flash on top of its own.
>       if (mode == RefreshMode::Fast) mode = RefreshMode::Full;
>       _needsInitialFull = false;
>       mode = (_cfg.halfSeqOverride != 0) ? RefreshMode::Half : RefreshMode::Full;
>     } else if (!_isScreenOn && _cfg.fullSeqOverride == 0) {
>       // X4-class cold start: panel asleep -> a (warmed) HALF full-clear. Override
>       // boards skip this — their fast sequence self-powers, so _isScreenOn is false
>       // every page and forcing HALF would make every page a slow full-waveform flash.
>       mode = RefreshMode::Half;
>     }
>   }
>   // ...
> ```
> 
> @itsthisjustin Does this even make sense? I think the `if (!turnOff) {` should be in the else-path after the "refresh on first draw"-part, which should happen even with `turnOff` activated.
> 
> So imho the fix would be better in the FreeInk SDK. I don't know whether force refreshing in the Crosspoint Reader Itself does an unnecessary refresh on other display controller types.
> 
> On the other hand, I see less risk in your change.

I think this is because now due to the boot screen being skipped, this first refresh counts as a normal operation refresh and is where the sunlight fix changes the power state of the display. I'd consider this a bug though.

I'm not sure if this belongs in the SDK though as the sunlight fix is a specific xteink c3 board issue. I think i'd probably just add a refresh at the device level in the firmware over a driver config in the SDK as the sunlight fading fix doesn't even exist on the x4pro which uses the same display on some variants

theZiz:
If I remember correctly, also the Sticky and Murphy M4 might use the driver, so they might have a similar problem? I mean, the fading fix does not make much sense on the Sticky, but is it still available as option?

itsthisjustin:
> If I remember correctly, also the Sticky and Murphy M4 might use the driver, so they might have a similar problem? I mean, the fading fix does not make much sense on the Sticky, but is it still available as option?

No it's removed on sticky as well

theZiz:
> I'm not sure if this belongs in the SDK though as the sunlight fix is a specific xteink c3 board issue.

Well, good point, but it is already in there, isn't it?

thiagokokada:
> On the other hand, I see less risk in your change.

Can confirm that disabling the "Sunlight Fading Fix" fix the issue on `develop` branch, but I still think that maybe we should merge this since this is the more general fix.

thiagokokada:
Considering that https://github.com/Free-Ink/freeink-sdk/pull/46 fix has some issues (I reported it in the PR) and we are near 1.6.0 release, I think this is probably the safest fix. Also it is not like this PR introduces any additional flashing:

- Before this PR, with "Sunlight Fade Fix" off we already flashed (once) with a custom sleep image
- Before this PR, with "Sunlight Fade Fix" on we don't flash for some reason
- With this PR, with "Sunlight Fade Fix" on/off we flash (once)

@itsthisjustin @Uri-Tauber

Uri-Tauber (approved):
LGTM. 
@itsthisjustin merge when you're ready.

PR #3018, opened 2026-08-13
feat: Using factory lut for less fizzy cover image

theZiz:
### Making use of the factory lut grayscale renderings for covers

This PR is based on the work of zgredex, who unfortunately didn't continue his efforts: https://github.com/crosspoint-reader/crosspoint-reader/pull/1614

He added way more stuff, optimized, played around and also added a lot of refactorings.

I tried to boil it down to an as small as possible change request for the most important feature (for me): Nice looking covers.

For me a cover is an essential part of the reading experience, especially as eBook Reader can keep them when turned of. But those changes could of course also be extended to images in the book as well as the image viewer of Crosspoint Reader as zgredex did it.

#### develop
strength_before

#### my change
strength_after

#### develop
jade_before

#### my change
jade_after

### Open discussion

I had to touch lib/hal, although I just extended the forwarding interface. However, I also had to add driver specific compile time  branching in the Sleep activity. I could move some of the stuff to the freeink repo, however I would still need some knowledge about the used driver as not every display supports the factory lut render mode and the framebuffer handling is inverted in that case.

## Scope Check

Please confirm:

- [ ] If this PR touches `freeink-sdk/`, `lib/hal/`, the bootloader, OTA, or recovery code, I have coordinated with
      the relevant maintainer. -> not yet, see discussion above.

## Additional Context

  specific areas to focus on).
* Memory / flash impact, if known.

---

### AI Usage

### DIFF
--- a/lib/GfxRenderer/GfxRenderer.cpp
+++ b/lib/GfxRenderer/GfxRenderer.cpp
@@ -1352,6 +1352,10 @@ void GfxRenderer::drawBitmap(const Bitmap& bitmap, const int x, const int y, con
         drawPixel(screenX, screenY, false);
       } else if (renderMode == GRAYSCALE_LSB && val == 1) {
         drawPixel(screenX, screenY, false);
+      } else if (renderMode == FACTORY_GRAY_LSB && !(val & 1)) {
+        drawPixel(screenX, screenY, false);
+      } else if (renderMode == FACTORY_GRAY_MSB && val < 2) {
+        drawPixel(screenX, screenY, false);
       }
     }
   }
@@ -1391,6 +1395,9 @@ void GfxRenderer::drawBitmap1Bit(const Bitmap& bitmap, const int x, const int y,
     return;
   }
 
+  // Factory grayscale inverts the pixel values
+  const bool full = !(renderMode == FACTORY_GRAY_LSB || renderMode == FACTORY_GRAY_MSB);
+
   for (int bmpY = 0; bmpY < bitmap.getHeight(); bmpY++) {
     // Read rows sequentially using readNextRow
     if (bitmap.readNextRow(outputRow, rowBytes) != BmpReaderError::Ok) {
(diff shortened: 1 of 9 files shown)


jb567:
I've given this a go on my x4. I found it definitely helped with some weirdness around some overly bright blotches on some covers I saw, but none looked quite as dramatically better as your SOTF example.

theZiz:
I think, especially very bright covers look better. For "Jade Legacy" the difference is way smaller, but e.g. the "stemp" "Spiegel Beststeller author" is sharper.

itsthisjustin:
Oh hell yeah this looks great

lib/GfxRenderer/GfxRenderer.cpp
```diff
@@ -1589,7 +1596,14 @@ void GfxRenderer::invertScreen() const {
 void GfxRenderer::displayBuffer(const HalDisplay::RefreshMode refreshMode) const {
   auto elapsed = millis() - start_ms;
   LOG_DBG("GFX", "Time = %lu ms from clearScreen to displayBuffer", elapsed);
-  display.displayBuffer(refreshMode, fadingFix);
+  // After a factory LUT render, RED RAM still contains the grayscale MSB plane.
```
theZiz, line 1599: Regarding the comment of the AI:

> The change improves cover rendering, but the asynchronous display path can retain the factory-LUT state and use the wrong refresh mode, potentially affecting subsequent screen updates. This is a bounded issue that should receive explicit owner follow-up.

This should be handled here.

However, I didn't add such a handling for the `displayBufferAsync` - mostly because zgredex didn't do it back then as the method was not existing then.

So, maybe this should also handled there?

For the use case right now (cover image) it doesn't matter as the device goes into sleep state anyway, but if the method wants to be used for the image viewer or images in books, this might be more important.

Uri-Tauber:
@theZiz Just add zgredex as co-author, since this is based on their work.

theZiz:
> @theZiz Just add zgredex as co-author, since this is based on their work.

Didn't know, this was possible at all! Very good point, I added them. :)

itsthisjustin (approved):

PR #3025, opened 2026-08-14
refactor(reader): consolidate reader activities

Uri-Tauber:
# Summary

The reader flow is now structured around one shared abstract reader base plus three concrete readers, instead of using `ReaderActivity` as a trampoline into separate format activities.

Before, opening a book passed through `ReaderActivity`, which routed into standalone EPUB, TXT, or XTC activities. Shared behavior had to be coordinated across those separate activity implementations.

Now, `ReaderActivity` is the shared reading shell, and `EpubReaderActivity`, `TxtReaderActivity`, and `XtcReaderActivity` inherit from it.

```
  ActivityManager::goToReader(path)
    ├─ empty path      → FileBrowserActivity
    ├─ BMP / PNG       → BmpViewerActivity
    └─ ebook path      → concrete ReaderActivity
                         ├─ EpubReaderActivity
                         ├─ TxtReaderActivity
                         └─ XtcReaderActivity
```

The shared base now owns the behavior that should be identical for every ebook format:

  - reader lifecycle, orientation reset, current-book and recents updates;
  - generic Back/page navigation;
  - end-of-book screen and next-book suggestions;
  - refresh scheduling and forced refresh handling.

  Each format activity owns the behavior that actually varies:

  - EPUB keeps its ordered input flow, including footnote Back, auto-page-turn cancellation, manual-turn debounce, orientation long-press, and its image/grayscale rendering.

  - TXT keeps text indexing, pagination, status-bar/anti-alias rendering, and can enter the shared end-of-book state after its last page.
  - XTC keeps fixed-page bitmap rendering, chapter selection, and portrait-only display policy.

A concrete reader now owns its complete format render, while the base owns only shared reader behavior.

`ActivityManager` also owns non-reader routing. `ReaderActivity::create` returns only reader subclasses, so the reader module no longer depends on the file browser or image viewer.

---

### AI Usage


itsthisjustin:
Ran on Sticky and seems to be working. We can RC it and users can report anything

itsthisjustin (approved):

itsthisjustin (approved):

PR #3027, opened 2026-08-14
fix: stop long chapters aborting on the build tables' growth

s4y:
> [!IMPORTANT]
> This entire PR was written by Claude Opus 5

## The bug

A section build keeps two tables in RAM for the whole parse: the per-page LUT in `Section::BuildContext` and the anchor map in `ChapterHtmlSlimParser`. Both were `std::vector`s, so both grew geometrically into a single contiguous block and copied themselves at every doubling — while the live parse was already holding most of the heap.

In a book whose spine is one giant XHTML file, that is fatal. A crash report from a 1300-page section decodes to:

```
operator new(unsigned int)                        <- new_op.cc:55, allocation failed
__cxxabiv1::__terminate                           <- abort()
vector<pair<string,uint16_t>>::_M_realloc_append
ChapterHtmlSlimParser::flushPendingAnchor()       ChapterHtmlSlimParser.cpp:241
```

The allocator arguments are on the stack next to those frames: `0x3800` (14,336) and `0x7000` (28,672) — the anchor map doubling from 896 to 1792 entries of 16 bytes, both blocks live during the copy.

`std::vector` allocates through the *throwing* `operator new`, which under `-fno-exceptions` is `abort()`. So an allocation failure reboots the device rather than returning null. On resume the chapter re-parses from page 0 and dies in the same place, giving a reboot every few pages.

`EpubReaderActivity::buildTickHeapGate()` cannot catch this: it admits a build tick when the largest free block is ≥16KB, and one page inside that tick then demanded 28KB.

## The fix

New `lib/Memory/ChunkedVector.h`: append-only, fixed-size chunks allocated with `new (std::nothrow)`, never reallocated or copied, O(1) random access, `push_back()` returns `false` instead of aborting. Both tables use it — 1.5KB chunks for the LUT, 1KB for the anchor map.

At the crash point this turns a 28,672-byte contiguous demand into 1,024 bytes, and drops steady-state build RAM by ~10KB (no vector slack). On failure it degrades instead of dying: a dropped anchor costs link precision, an exhausted LUT suspends the build and keeps the pages already laid out as a partial.

## Verification

Built clean (no new warnings), `pio check` passes, and the crashing `_M_realloc_append<pair<string,unsigned short>>` instantiation is gone from the ELF.

On device (X3), the chapter that previously aborted at page 1308 now builds to completion:

```
[392786] [DBG] [EHP] Time to parse and build pages: 152771 ms
[392792] [DBG] [SCT] Page 1987 processed
```

1988 pages in one 153-second parse, no `abort`, no panic, no reset. Heap through the deep part of the build:

| uptime | free | max alloc |
|---|---|---|
| 301s (page ~830) | 73,656 | 59,380 |
| 391s (page ~1900) | 42,088 | 32,756 |
| 401s (build freed) | 72,904 | 61,428 |

At the deepest point the largest contiguous block available was 32,756 bytes — under what the old code was asking for.

Not exercised by this run: the OOM fallback paths (the anchor map stayed under its 1024 cap and the LUT under 16384 pages), so those are reasoned-through but untested.

### DIFF
--- a/lib/Epub/Epub/Section.cpp
+++ b/lib/Epub/Epub/Section.cpp
@@ -70,18 +70,36 @@ Section::Section(const std::shared_ptr<Epub>& epub, const int spineIndex, GfxRen
 // (no-op once a build has completed or never started).
 Section::~Section() { suspendBuild(); }
 
-uint32_t Section::onPageComplete(std::unique_ptr<Page> page) {
+void Section::onPageComplete(std::unique_ptr<Page> page, const uint16_t paragraphIndex, const uint16_t listItemIndex,
+                             const uint32_t visibleTextOffset) {
+  // Already stopping (buildSomeMore suspends as soon as it regains control): don't
+  // write pages that can no longer be indexed.
+  if (build_->lutExhausted) return;
+
+  // A zero file offset marks a page that failed to write; commitBuildFile rejects the
+  // build when it sees one, so the entry is still recorded to keep the LUT aligned
+  // with the pages emitted so far.
+  uint32_t position = 0;
   if (!file) {
     LOG_ERR("SCT", "File not open for writing page %d", builtPageCount_);
-    return 0;
+  } else {
+    position = file.position();
+    if (!page->serialize(file)) {
+      LOG_ERR("SCT", "Failed to serialize page %d", builtPageCount_);
+      position = 0;
+    } else {
+      LOG_DBG("SCT", "Page %d processed", builtPageCount_);
+    }
   }
 
-  const uint32_t position = file.position();
-  if (!page->serialize(file)) {
-    LOG_ERR("SCT", "Failed to serialize page %d", builtPageCount_);
-    return 0;
+  if (!build_->lut.push_back({position, paragraphIndex, listItemIndex, visibleTextOffset})) {
+    // Out of memory, or past the LUT's ceiling. builtPageCount_ is deliberately not
+    // advanced: this page is unreachable without an index entry, so the build stops
+    // here and keeps the pages it can still address.
+    LOG_ERR("SCT", "OOM: page LUT at %u pages", build_->lut.size());
+    build_->lutExhausted = true;
+    return;
   }
-  LOG_DBG("SCT", "Page %d processed", builtPageCount_);
 
   builtPageCount_++;
   // pageCount is the pages available to read: a rebuild over a partial only raises it
@@ -89,7 +107,6 @@ uint32_t Section::onPageComplete(std::unique_ptr<Page> page) {
   if (builtPageCount_ > pageCount) {
     pageCount = builtPageCount_;
   }
-  return position;
 }
 
 void Section::writeSectionFileHeader(const ReaderRenderSpec& spec) {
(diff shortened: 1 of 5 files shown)


Uri-Tauber:
Can you share the epub which crashed before, and working with this fix?

s4y:
@Uri-Tauber In this case, no problem — it’s a long export from Google Docs.

[Surviving%20Burning%20Man-%20What%20to%20Bring%20...%20and%20Why,%20by%20Jessie%.zip](https://github.com/user-attachments/files/31111186/Surviving.20Burning.20Man-.20What.20to.20Bring.20.20and.20Why.20by.20Jessie.zip)

serialx:
> @Uri-Tauber In this case, no problem — it’s a long export from Google Docs.
> 
> [Surviving%20Burning%20Man-%20What%20to%20Bring%20...%20and%20Why,%20by%20Jessie%.zip](https://github.com/user-attachments/files/31111186/Surviving.20Burning.20Man-.20What.20to.20Bring.20.20and.20Why.20by.20Jessie.zip)

You attached a PDF file. Can you upload the epub file you've tested?

s4y:
> > @Uri-Tauber In this case, no problem — it’s a long export from Google Docs.
> > [Surviving%20Burning%20Man-%20What%20to%20Bring%20...%20and%20Why,%20by%20Jessie%.zip](https://github.com/user-attachments/files/31111186/Surviving.20Burning.20Man-.20What.20to.20Bring.20.20and.20Why.20by.20Jessie.zip)
> 
> You attached a PDF file. Can you upload the epub file you've tested?

@serialx Whoops, sorry. This should be it: [Surviving Burning Man.epub.zip](https://github.com/user-attachments/files/32388101/Surviving.Burning.Man.epub.zip)

serialx:
PR #3027 — X3 device QA

**The reported crash is fixed in this test, but memory usage and fragmentation should be optimized before merging.** The candidate leaves less free heap and a smaller largest free block, with new Home cover-buffer allocation failures. Re-run the paired tests after optimization to confirm improved memory headroom and no new cover-buffer failures while preserving the long-chapter fix.

Using the author's [attached EPUB](https://github.com/crosspoint-reader/crosspoint-reader/pull/3027#issuecomment-5732440418), a cold open followed by Go to 100% produces:

| Original EPUB functional check | Baseline | Candidate |
|---|---|---|
| Cold first-page open | Pass | Pass |
| Go to 100%, complete the single chapter | Abort after zero-based page 1632 | Pass: page 2482/2482 |
| Completed cache reopened after reset | Unavailable: build aborted | Pass: 2482 pages deserialized, build skipped |
| Reading-content image, first page | Reference | Byte-identical content; only battery-icon pixels differ |

The candidate also passed twelve backward page turns and early/late TOC navigation. Baseline's decoded crash is an anchor-vector reallocation: 28,672 B requested while the old 14,336 B block remains live. The candidate's fixed-size chunks avoid that allocation peak.

**Regression:** returning Home logs `OOM: cover buffer (16368 bytes)` in all three candidate runs (three occurrences per run), versus none on baseline. This is separate from thumbnail-decoding failures already present on both builds. The candidate's measured largest free block is below the cover-buffer request.

Demian comparison: three alternating cold boots per build, six warmup and six measured turns per boot (18 measured renders per build), identical page sequences, and a 13-second heap-sampling wait. Settings: Pretendard 16 pt, Extra Wide, justified, margin 5, portrait; embedded styles, extra paragraph spacing and AA enabled; full refresh every 15 pages.

| Measurement | Baseline | Candidate | Candidate − baseline |
|---|---:|---:|---:|
| Median page drawing, bw_render | 100.5 ms | 99.0 ms | −1.5 ms |
| Median total rendering, including refresh | 1,622 ms | 1,634 ms | +12 ms |
| Median cold layout preparation | 4,951 ms | 4,955 ms | +4 ms |
| Median free heap after measured pages | 51,356 B | 48,032 B | −3,324 B |
| Median largest free block after measured pages | 21,492 B | 15,348 B | −6,144 B |
| Lowest boot Min Free watermark sampled in measurement window | 18,724 B | 17,452 B | −1,272 B |
| Static RAM | 56,696 B | 56,696 B | 0 B |
| Flash | 5,568,435 B | 5,568,733 B | +298 B |

Reading content was byte-identical across all three screenshot pairs; differences were confined to footer estimates. Timing differences were small. Chunk directories and minimum chunk allocations increase short-section storage: target measurements show +584 B across the live build/parser objects. Allocation order also affects fragmentation, so that object-size increase alone does not explain the full heap delta.

Baseline: develop `4b17a7bb`; candidate: PR temporarily rebased as `13072887`. Both include serial-control commit `f039422c` (measured heads `4c3cf885` / `d8ac3df6`). Both `default` builds passed. No commits were pushed to the PR.

Coverage: X3 with the settings above. A screenshot-transfer timeout caused a monitor-driven USB reset after chapter completion; cache and navigation checks passed after reconnecting. Host container OOM/capacity tests passed, but firmware allocation-failure and capacity-limit paths were not forced on the device.

Issue #3021, opened 2026-08-14
CJK: "Wide" line spacing is inconsistent within word-wrapped paragraphs


chohoyeon-ops:
### Affected Version

master branch, v1.5.0 (reference comparison: KO fork)

### Describe the Bug

Line breaking (word wrap) itself works correctly — no complaints there. The problem is that the "Wide" line-spacing setting does not apply uniformly to wrapped lines *within a single paragraph* on master 1.5.0.

Test conditions: same EPUB (Korean web-novel, Chapter 6, ~5%), same font, same font size, Line spacing = Wide, on the same device (Xteink X4).

As a reference point, the KO fork keeps a consistent gap between wrapped lines inside one paragraph, and a distinct, also-consistent larger gap between paragraphs. On master, the gap between wrapped lines varies line-to-line and doesn't cleanly separate into "within-paragraph" vs "between-paragraph" spacing at all.

Measured vertical gap (px) between consecutive text-line centers, same paragraph / same crop region:

| | Gaps (px) |
|---|---|
| KO fork (reference) | 43, 43, 43, 43, 43, 43 (within paragraph) / ~64 (between paragraphs) — consistent, two clean tiers |
| master 1.5.0 | 58, 58, 43.5, 43.5, 72, 57.5, 57.5, 58, 57.5, 29, 29, 29, 29.5 — irregular, no consistent tiering |

Hypothesis: line breaking is fine, but the line-spacing calculation may be applied inconsistently between logical/source lines and physically rendered (wrapped) lines. This could explain why spacing appears consistent for short, unwrapped text but becomes irregular once a long sentence wraps into multiple physical lines — which is common in CJK text.

Screenshots attached:
- line-spacing-ko-fork.png (reference build, consistent spacing)
- line-spacing-master-1.5.0.png (affected build, irregular spacing)

<img width="480" height="800" alt="Image" src="https://github.com/user-attachments/assets/2eeeaaf3-d68b-4df8-887c-a3db4d843f1d" />
<img width="480" height="800" alt="Image" src="https://github.com/user-attachments/assets/838d6389-9bfe-4eef-9817-7f422813e08d" />

### Steps to Reproduce

1. Load the attached EPUB with identical settings (same font/size, Line spacing = Wide) on master 1.5.0.
2. Go to Chapter 6, page 1 (~5%).
3. Find a paragraph long enough to wrap across 3+ lines and compare the spacing against the surrounding paragraph breaks.
[...]

PR #3037, opened 2026-08-15
feat: add Extra Wide reader line spacing option

serialx:

CJK text needs more leading than Latin text, and the current line-spacing scale tops out at values tuned for Latin. Latin typesetting conventionally defaults to ~120% line-height; CJK glyphs are dense squares with no ascender/descender rhythm to visually separate lines, so the same leading reads cramped. From Peter Biľak's [*Typesetting principles of Chinese, Japanese, and Korean (CJK) text*](https://www.typotheque.com/articles/typesetting-cjk-text) (Typotheque):

> For example, in a 10-point font, the default leading would typically be 12 points. This works well for most Latin fonts, with adjustments only needed for fonts with unusually large x-heights. For CJK (Chinese, Japanese and Korean) text, which tends to have a higher density of information per character, a leading value of around 1.7 improves both readability and aesthetic balance.

With SD-card CJK fonts now supported (v1.5.0), readers of Korean/Chinese/Japanese books hit this wall: even the Wide setting is visibly tight for Hangul and Han text. This PR adds one more step, **Extra Wide**, so CJK readers can move toward comfortable leading — while staying useful as a plain accessibility/preference option for Latin readers too.

Each font family's existing increment progression is continued rather than inventing a new scale (values are the internal line-compression factor applied to the font's built-in line height, not absolute leading):

* Noto Serif / SD-card fonts: 0.95 / 1.0 / 1.1 / **1.2** (existing 0.1 steps at the top of the scale)
* Noto Sans: 0.90 / 0.95 / 1.0 / **1.05** (uniform 0.05 steps, matching that family's existing increments)

Concretely:

* One new enum value `EXTRA_WIDE`, inserted before `LINE_COMPRESSION_COUNT` so stored-settings range validation keeps working; older firmware settings load unchanged.
* One entry in the declarative `SettingsList`, which drives both the on-device settings UI and the web settings page.
* `STR_EXTRA_WIDE` labels for **all 32 languages** (translations provided for every locale; native-speaker corrections welcome).
* No cache-format change: the compression factor is already part of the section layout cache header, so selecting the new step invalidates and re-lays-out automatically.

## Preview
hp_wide_vs_extra_wide

## Scope Check

- [x] This extends an existing reader typography setting (line spacing) by one step; it adds no new subsystem.
- [x] This PR does not touch `freeink-sdk/`, `lib/hal/`, bootloader, OTA, or recovery code.

## Additional Context

* **Memory / flash impact:** effectively none — one enum value, one settings-list entry, and one short string per language in the flash-resident string tables.
* **Settings compatibility:** values 0–2 keep their meaning; 3 is new. `LINE_COMPRESSION_COUNT` moves from 3 to 4, so validation accepts the new value and nothing previously stored becomes invalid.
* **Hardware testing:** X3, SD-card font (Pretendard), Korean EPUBs: Extra Wide lays out correctly, section cache regenerates on switch, and the setting round-trips through both the device UI and the web settings page.

---

### AI Usage

Did you use AI tools to help write this code? _**YES**_ — the patch and translations were done with Claude Code with Fable 5; hardware testing was manual. Non-English `STR_EXTRA_WIDE` labels are AI-provided translations phrased to match each locale's existing Tight/Normal/Wide labels.


chohoyeon-ops:
Thanks for the quick response. Extra Wide as a new step is a nice addition on its own, but I don't think it addresses what #3021 is reporting, so I wanted to clarify the reproduction focus.

The bug isn't that "Wide" is too tight overall (a magnitude issue). The issue is that the gap **varies inconsistently between lines within the same paragraph** (a variance issue). Applying a larger fixed compression factor (1.2 instead of 1.1) would likely just scale the same irregular pattern up, rather than making the spacing consistent.

Could you retest specifically by comparing these two cases?

1. A long sentence on X3/X4 that gets **forced word-wrapped** across 3+ physical lines. Please check the gap between each of those wrapped lines.
2. The line spacing immediately after that wrapped sentence, where the **next sentence/paragraph begins**.

On the reference KO fork build, both cases are internally consistent: all wrapped lines share the same line-gap value, while the paragraph-break gap is a different but also consistent value.

On master, neither is consistent. The per-line pixel measurements in the issue body show the gap changing from line to line, even within what should be the same type of spacing.

If Extra Wide reproduces that same line-to-line irregularity, just at a larger scale, that would confirm that this is a separate calculation issue from what #3037 addresses, and is probably worth tracking separately from the new setting.

serialx:
I wasn't implying that this PR fixes your issue. I was saying that I tried to reproduce your issue in the latest development tip and were unable to reproduce it. I even bought the same ebook you screenshot. Maybe it was a font issue? I don't know. We need to find what's different between your device and mine to find the issue.

chohoyeon-ops:
> I wasn't implying that this PR fixes your issue. I was saying that I tried to reproduce your issue in the latest development tip and were unable to reproduce it. I even bought the same ebook you screenshot. Maybe it was a font issue? I don't know. We need to find what's different between your device and mine to find the issue.

Thanks for checking, especially for buying the same ebook. Since you can’t reproduce it on the latest development tip, I’ll try the latest dev build on my X4 as well. That should help distinguish whether this is a device/environment difference or something that has already changed in the development branch.

Uri-Tauber (approved):

PR #3046, opened 2026-08-15
fix: Add runtime check for factory grayscale support

itsthisjustin:
Introduce supportsFactoryGrayscale() method to HalDisplay to determine if the device supports factory grayscale mode at runtime. Update SleepActivity to conditionally use factory grayscale settings only when supported, falling back to standard grayscale mode otherwise.

In #3018 we set the sleep covers to run based off a driver availablity but x4pro can have both UC and SSD based controllers so on UC based Pros all sleep covers were failing to generate

### DIFF
--- a/lib/hal/HalDisplay.cpp
+++ b/lib/hal/HalDisplay.cpp
@@ -145,6 +145,8 @@ void HalDisplay::writeGrayscalePlaneStrip(bool lsbPlane, const uint8_t* rows, ui
 
 bool HalDisplay::supportsStripGrayscale() const { return einkDisplay.supportsStripGrayscale(); }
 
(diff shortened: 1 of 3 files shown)


theZiz:
Also handled here: https://github.com/crosspoint-reader/crosspoint-reader/pull/3031 with some more corner case handling

theZiz:
I just saw you reverted it anyway. In that case I would propose to make first the needed changes in the freeink sdk and afterwards I do a fixed PR which handles the corner cases right from the beginning :)

itsthisjustin:
> I just saw you reverted it anyway. In that case I would propose to make first the needed changes in the freeink sdk and afterwards I do a fixed PR which handles the corner cases right from the beginning :)

So I added a commit to main on freeink that at least handles the driver detection properly

itsthisjustin:
Closing this for now

PR #3049, opened 2026-08-15
feat: PSRAM image cache with one-shot pixel-cache writes for in-book images

hooligan333:
## Summary

* **What is the goal of this PR?** Stop in-book images re-streaming their `.pxc` off SD on every render pass. An anti-aliased image page renders ~13 times (BW double-refresh plus every band of both gray planes), and today each pass re-reads the whole pixel cache from the card. Column clipping can't help — the row stride (~100 B) is smaller than an SD sector, so a strip pass touches every sector anyway. The current mitigation is a single per-page-render slot in internal RAM, chunked into 6×16 KB because a contiguous 96 KB rarely fits a fragmented mid-render heap, and dropped when the page render ends. It only ever holds **one** image, so the second image on a page re-streams itself ~13 times, and a back-turn re-streams everything.

  - New flag `CROSSPOINT_PSRAM_IMAGE_CACHE`, **default off**, enabled in `[env:x4pro]` only (PSRAM board — same precedent as #3032 enabling `FREEINK_FRONTLIGHT_LS` there). Every other env compiles and behaves exactly as today.
  - `lib/Epub/Epub/blocks/ImageBlock.{h,cpp}`: under the flag, the single internal-RAM slot becomes an **8-entry LRU of whole `.pxc` images** in PSRAM (one contiguous `heap_caps_malloc(MALLOC_CAP_SPIRAM)` block each, keyed by the hash of the cache path, storing width/height) that **outlives the page render**. Every pass of every image on a page hits RAM, and a back-turn or re-visit while an image is resident costs no SD access at all. Eviction is LRU (own entry → empty slot → least recently used); a slot reuses its block when the new image fits. `hasValidCache()` consults the LRU before the file, so the page-turn decode probe costs no SD access for resident images. Without the flag the chunked internal-RAM slot is compiled in unchanged.
  - `lib/Epub/Epub/converters/PixelCache.h`: **one-shot `.pxc` writes**. With the payload cap in reach, the band can simply *be* the whole image — one PSRAM block holds a complete `.pxc` (4-byte header + payload), `advanceTo()` has nothing to flush, and `finalize()` emits the file in a **single write** instead of the ~40 interleaved 3 KB band flushes a full-page image costs today. The block is pre-zeroed so rows a clipped image never covers stay black, matching the banded path's zero-fill exactly.
  - **Donation**: on a successful one-shot finalize the block is handed straight to the render cache instead of making the ~12 passes that follow read back the file the decode just wrote. Ownership transfers only on success; the callee re-derives width/height from the block's own header and rejects anything whose size is not exactly `4 + ((w+3)/4)*h`.
  - `lib/Epub/Epub/converters/PxcFormat.h` (new): the on-disk `.pxc` constants and the donation hook, so the converter no longer has to include the block layer for what is only a file format.
  - `src/activities/reader/ReaderActivity.cpp`: one line in `onExit()` — `ImageBlock::clearPxcLru()` gives the PSRAM back when leaving the reader, where the slots can no longer be hit. `onExit` runs with the RenderLock held (`ActivityManager::exitActivity`), which is the cache's access rule.

Memory budget: 8 slots × at most 96,004 B ≈ **750 KB of the 8 MB PSRAM**, held only while the reader is open. Internal RAM use goes *down* on the flagged build (the 96 KB chunked slot is gone). The 96,000 B payload cap is a full-screen 800×480 image at 2 bpp; anything larger is refused and takes the streaming path.

Staleness is handled the same way the on-disk header check does it: a hit is only used when the slot's width/height match the block's expected size within ±1, because re-layout (font size, orientation, margins) recomputes an image's box without changing its path. A mismatch drops the entry and falls through to the file, which is itself about to be rejected and re-decoded.

## Scope Check

- [x] The stock firmware does not already handle this well, and no other popular CrossPoint fork already does. This is a pure performance change to the existing image path — no new user-facing surface, no new setting.
- [x] This PR touches neither `freeink-sdk/`, `lib/hal/`, the bootloader, OTA, nor recovery code.

## Additional Context

* **Tested working on X4 Pro hardware by the submitter.** Multi-image pages, back-turns, AA rendering on image pages, and multi-day reading — including a combined build that also runs light sleep. Image pages no longer stall re-reading the card on every grayscale band, and the second image on a page is now as fast as the first.
* **Adversarially reviewed** before submission, with the findings folded in: LRU lock discipline (every access is on a task holding the RenderLock — verified for both the render path and the decode path; the cache has no lock of its own by design), donation validation (the donated block must be exactly one `.pxc` image or it is refused and the caller keeps it), and eviction safety (a reserve that reuses an oversized block zeroes the slot dimensions first, so a slot can never advertise a new hash with the previous occupant's size).
* **Memory impact**: ~750 KB of PSRAM at full occupancy on the 8 MB part, freed on reader exit. Internal RAM is *reduced* on the flagged build — the 6×16 KB chunked slot no longer exists. Flash delta is small (the two paths are mutually exclusive at compile time). Flag-off builds are byte-for-byte the same design as today.
* **Graceful fallback everywhere.** An image past the payload cap, a failed PSRAM allocation, a short read while filling a slot, or a refused donation all land silently on today's streaming behavior. Nothing here is load-bearing for correctness — worst case is the current performance.
* **Areas to focus review on**: the RenderLock invariant on the LRU (`ImageBlock.cpp`'s threading note), the size validation in `adoptPxcImage`, and the `#else` branch of `ImageBlock.cpp` — the only shared edit outside the flag is that the streaming render loop moved into a helper both branches call.
* **Builds**: `pio run -e x4pro` (flag on) and `pio run -e default` (ESP32-C3, flag off) — both SUCCESS.

---

### AI Usage

Did you use AI tools to help write this code? _**YES**_ — implementation and review with Claude Code (Opus subagents under orchestration); all hardware testing done by the submitter on a physical X4 Pro.

https://claude.ai/code/session_01UUgQffv7GGVCWQ9gijsVbp

### DIFF
--- a/lib/Epub/Epub/blocks/ImageBlock.cpp
+++ b/lib/Epub/Epub/blocks/ImageBlock.cpp
@@ -12,6 +12,11 @@
 
 #include "Epub/converters/DirectPixelWriter.h"
 #include "Epub/converters/ImageDecoderFactory.h"
+#include "Epub/converters/PxcFormat.h"
+
+#ifdef CROSSPOINT_PSRAM_IMAGE_CACHE
+#include <esp_heap_caps.h>
+#endif
 
 // Cache file format:
 // - uint16_t width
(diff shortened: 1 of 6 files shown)


itsthisjustin:
I think I'll merge my existing x4pro branch into develop before we further bloat it with all these PRs you opened and then you can rebase on develop instead? That way we get focused PRs into develop instead into one giant branch?

hooligan333:
> I think I'll merge my existing x4pro branch into develop before we further bloat it with all these PRs you opened and then you can rebase on develop instead? That way we get focused PRs into develop instead into one giant branch?

Yes, of course. My apologies!

PR #3053, opened 2026-08-15
feat: Toggle Light short-press power action

hooligan333:
## Summary

* **What is the goal of this PR?** Give the X4 Pro a one-press blackout for the reading lights. Reading in a dark
  room with the frontlight on and night mode on, the only way to go dark today is to walk into the reader menu or the
  frontlight panel and turn each one off, then walk back in and turn each one on again. This adds **"Toggle Light"** as
  an option on the existing **Short Power Button Click** setting: one quick press turns off whichever of
  {frontlight, night mode} are on and *remembers the set*; the next press brings back exactly that set.
* **What changes are included?** One flag-gated feature, `CROSSPOINT_PWR_TOGGLE_LIGHT`, enabled on `[env:x4pro]`.
  Every other environment (the C3 `default` included) is byte-identical to before.

  * **`TOGGLE_LIGHT = 6`** appended at the **end** of `SHORT_PWRBTN`, with the matching option appended at the end of
    the settings row — so every stored value keeps the index it has always had.
  * **`pwrToggleLightRemembered`** — a two-bit persisted field (bit0 frontlight, bit1 night mode) holding the
    remembered set, so a restore survives sleep and reboots. Cleared once a restore consumes it.
  * **Empty state** (first use, or the press right after a restore) switches the frontlight on at its saved level
    rather than doing nothing, so the action always behaves like a light button. A level parked at 0% is lifted to a
    visible default first, so the gesture is never inert.
  * **`handleX4ProFrontlightDoubleClick` is suppressed** while the option is selected — two quick presses would
    otherwise toggle the light twice and then have the double-click handler flip it a third time.

**Sleep is untouched.** Only `SLEEP` drops `getPowerButtonDuration()` to 10 ms; `TOGGLE_LIGHT` leaves it at the 400 ms
default, so a long hold still sleeps exactly as it does in a stock build. This binding adds no sleep dead-end.

**Levels survive the round trip.** `HalFrontlight::setOn(false)` drives the PWM to 0 without touching
`lastBrightness`, so `setOn(true)` re-applies the selected brightness untouched — the remembered set stores *which*
lights were on, never *how bright*. This is also what makes it compose with the Night Light dim steps in #3033: the
blackout and the restore both go through `setOn()`, which that PR's `applyLit` round-trips the dim step through, so an
active dim level comes back as it was. #3033 is not on this branch yet, so the dim-step-aware variant of the
empty-state 0% guard (don't bump a deliberate sub-1% dark-room level) is not in this diff; it is one `&&` when #3033
lands, and the hardware testing below was done on a tree carrying that code.

**One porting note worth a reviewer's eye.** The settings row is split by `#if FREEINK_CAP_TOUCH`, and the stored
value **is** the index into that array. The non-touch variant omits Confirm (`PWR_CONFIRM = 5`), so an appended entry
there would land on 5 and be read as `PWR_CONFIRM` by the dispatch sites. The option is therefore added to the
cap-touch variant only, with a comment on the other. The X4 Pro is a `FREEINK_CAP_TOUCH` board, so the flag's only
shipping target takes that branch; enabling the flag on a non-touch board leaves the action unreachable rather than
wrong.

## Scope Check

Please confirm:

      (or, if one does, I explain why CrossPoint still needs it below). — The two lights are individually reachable
      from the reader menu and the frontlight panel; nothing binds them to one gesture, which is the whole point in a
      dark room.
      the relevant maintainer. — **N/A: this PR touches none of them.** No submodule bump, no HAL change, no
      bootloader/OTA/recovery code. `lib/hal/HalFrontlight` is **read only** — the action calls its existing
      `present()` / `isOn()` / `brightness()` / `setBrightness()` / `setOn()` surface and adds nothing to it. The diff
      is `src/main.cpp`, two headers, `english.yaml`, and the `[env:x4pro]` flag list.

## Additional Context

**Tested working on X4 Pro hardware by the submitter** — the capture/restore cycle in both orders (frontlight only,
night mode only, both), the empty-state fallback from a 0% level, persistence of the remembered set across a deep
sleep and a reboot, the long hold still sleeping, and the double-click shortcut correctly going quiet while the option
is selected. Also exercised **with the Night Light dim steps of #3033** (an active dim level came back as it was after
a blackout) and on the **combined light-sleep build** with #3032's `FREEINK_FRONTLIGHT_LS`.

**Adversarially reviewed.** Three areas got specific attention:

* *Release-edge dispatch.* The dispatch block is deliberately the same shape as the `FORCE_REFRESH` short-press block
  directly above it: the raw release edge via `mappedInputManager.wasReleased(Button::Power)`, no held-time gate, and
  **no early return**, so the rest of the frame still runs. A hold long enough to sleep never reaches it — the
  held-time block above calls `enterDeepSleep()`, which does not return. A **wake hold** cannot fire it either: the
  existing `wakePowerReleasePending` guard returns earlier in `loop()` on the frame that ends the wake gesture, and
  `gpio.update()` recomputes edges every frame, so the release is gone by the next one. Every `wasReleased` overload is
  a `const` non-consuming read, so the suppression path in the double-click handler cannot starve a downstream
  consumer of an edge.
* *Settings clamp safety.* On a flag-off build the option is absent from `enumValues`, and `fromJson()`'s existing
  clamp (`value >= enumValues.size()` → the field's struct default) folds a hand-edited `settings.json` carrying 6 back
  to Ignore; the web settings POST handler rejects it on the same bound. `pwrToggleLightRemembered` is a `VALUE` entry
  ranged `{0, TOGGLE_LIGHT_MASK, 1}`, so a corrupt file clamps to the two valid bits. It is category-less so it
  persists and round-trips through the web API while staying out of the on-device Settings screen — and it is placed
  with the other category-less state next to `frontlightBrightness`/`Warmth`/`On`, not in Controls, because the web
  page groups in list order and a category-less entry inside Controls would open the catch-all card early and split
  the page.
* *Deferred repaint rationale.* The night-mode flip is queued to a pending flag and consumed on the first frame the
  **raw** power button is up, rather than repainted inline. `handleForcedRefresh()` takes a `RenderLock`, which blocks
  on the rendering mutex with `portMAX_DELAY`, so the loop task can sit there for as long as the render task holds it
  — and for that whole stretch `loop()` stops calling `gpio.update()`, leaving a still-down power button unread,
  including a hold meant to sleep the device. The release edge does not guarantee the button is up either: it is a
  logical `MappedInputManager` edge, which on shared confirm/power boards is not the same signal as the raw
  `BTN_POWER` level. Waiting costs nothing — e-ink cannot show the flip any sooner than the user can see it, and the
  frontlight half of the gesture already applied at the dispatch. Sleeping in between discards the pending flip
  harmlessly, since `screenInverted` is already persisted and the next boot renders in the new polarity.

**Memory / flash impact.** Measured on `[env:x4pro]` against the same tree with the flag off: **+732 B flash, 0 RAM**
(5,545,874 → 5,546,606 flash; 64,372 static RAM unchanged). The persisted field is one `uint8_t` in an existing
struct. Every other environment is unchanged — the whole feature is behind `#ifdef CROSSPOINT_PWR_TOGGLE_LIGHT`.

Builds: `x4pro`, `default` — both SUCCESS.

---

### AI Usage

Did you use AI tools to help write this code? _**YES**_ — implementation and adversarial review with Claude Code;
hardware testing on an X4 Pro by the submitter.

https://claude.ai/code/session_01UUgQffv7GGVCWQ9gijsVbp


itsthisjustin:
Why not just use the existing double click behavior?

PR #3059, opened 2026-08-15
feat: add series grouping and a Series order to the Library shelf

hamaluik:
## Summary

* **What is the goal of this PR?** Add series browsing to the Library view. Targets `feat/library-view` (#3366).

  * Read EPUB 3 and Calibre series metadata.
  * Add a Series tab, book positions, series counts, and series-name search.
  * Update CLX1 to version 3. Preserve series when reusing book records.
  * Add English and French strings, format documentation, and 38 host tests.

## Scope Check

Please confirm:

* [x] I have read SCOPE.md and ROADMAP.md.
* [x] This PR is **not** a new built-in theme (themes are temporarily closed pending the move to SD-loaded themes).
* [x] This PR is **not** a new external network connector (sync engine, cloud storage, remote file access, etc.).
* [x] This PR is **not** an interactive app, writing tool, RSS/news/browser, media playback, or PDF feature.
* [x] The stock firmware does not already handle this well, **and** no other popular CrossPoint fork already does
* [x] If this PR touches `freeink-sdk/`, `lib/hal/`, the bootloader, OTA, or recovery code, I have coordinated with
  the relevant maintainer.

This extends Library Management. Stock firmware has no Library view. This PR changes none of the components that require maintainer coordination.

## Additional Context

* All 295 host tests pass. Firmware builds successfully; static analysis reports no defects.
* Tested on an Xteink X4, including metadata reuse, metadata settings, and index migration in both directions.
* The current base branch cannot build the simulator: its shim lacks `HalFile::modificationTime()`. Screenshots came from an earlier simulator build.
* Index storage adds series order, 4 bytes per book for series references, and 64 bytes per series. No permanent memory allocation is added.
* Existing indexes require a rebuild. Book caches do not change. Performance overhead is unmeasured.

**Series order**

Books grouped by series and sorted by position

**Collapsed groups**

Collapsed series groups

**Standalone books**

Standalone books after the series groups

---

### AI Usage

I used Claude Opus 5 under my direction. I edited, reviewed, and tested the code myself.


oreglio:
@hamaluik Reviewed, really impressive work! You used every extension point exactly as intended, and the code reads like it grew out of the same design: the 64-byte series stride, the reserved-header-bytes reuse, positions in hundredths for Calibre's fractional numbering, the single-memcmp grouping key. Measurements and the honest limitations section are appreciated too.

Deep pass on your eight commits — nothing structural, but worth fixing before review:

Bugs
- onEnter(): the metadata-toggle path calls rebuildIndex() while index still holds library.idx open, so the emit ends with remove+rename under a live SdFat handle — corruption risk on device, masked by POSIX in the simulator. index.close() first.
- Same path, rebuild failure: indexReady stays false with the old valid index still open — the shelf draws but Confirm is a silent no-op. Fall back to the old index.
- Emit: a failed staged-series read writes a valid entry with nameLen=0, rendering that series as a phantom mid-shelf "Standalone" group forever. Better to fail the emit — the selfSize guard then keeps the old index.
- A series whose folded name is empty (Cyrillic/Greek/CJK — fold() is ASCII-only, a base-feature limitation) is silently demoted to standalone; at minimum a log + a known-limitations line.

Parser tolerance (all small, all in LibraryMetaParse)
- collection-type is compared case-sensitively: Series disqualifies the collection — strictly worse than declaring no type.
- A <meta> inside an XML comment is parsed as live metadata.
- A raw > in an attribute value (legal XML) desyncs MetaScanner mid-tag.
- An untyped collection whose name tidies to empty commits and then gives up instead of falling back to the next one — the Calibre branch handles that case, the EPUB3 branch doesn't.

Perf on the C3
- The cost you flagged is mostly avoidable: findPrior() already matches unchanged books and the previous index stays readable until the rename, so an unchanged book's series is one 4-byte ClixSeriesRef + one 64-byte entry read from the old index — series survive rebuilds with the pre-series cost profile and no names held in RAM.
- buildSeriesOrder runs (n staged reads + a sort) even when the walk read no metadata — gate it on st.readMetadata.
- In series order, buildRows/letterOf re-read the same 64-byte entry for every adjacent row of a group — the one-entry memo you used in applyFilter works there too (and in non-series orders that memo thrashes; a pre-pass over the ≤512-entry series table is cheaper).

Happy to detail any of these. On sequencing you called it right — draft until the base lands; it started moving today (#2886 merged).

-- 

I used CC to assist me in the review, don't hesitate to push back where you see fit

Uri-Tauber:
Hey @hamaluik,

You’ve clearly put a lot of effort into series grouping, so I’m a bit hesitant to say this, but I don’t think adding fixed series grouping is the right direction.

Instead, I think we should give users the option to choose how they want the library grouped/sorted. You want Series; I prefer Subjects; another user might want Languages, Genres, etc.

I have a half-working attempt at this here:
https://github.com/Uri-Tauber/crosspoint-reader-fork/tree/feat/library-view. Unfortunately, it’s currently pretty messy, bloated, and slooow.

If you’d be interested in taking this direction on, that would be great.

hamaluik:
@Uri-Tauber @itsthisjustin well that sucks. I'd have appreciated a heads up in the original feature PR (or anytime since) before spending weeks on it to just be closed without comment, but it's your projec; I get it and you don't owe me anything and still I appreciate this project.

Such a broad rework isn't something I'm going to take on, especially on the back of this, but I will maintain a patch for myself in the meantime if anyone wants it.

There's a couple things that I had to include in this PR that don't seem fixed in `develop` yet if you want them:

1. `test/library_builder/stubs/HalStorage.h ` uses `uint*_t` without including `<cstdint>` which doesn't compile on newer GCC (16)
2. The simulator isn't working with the library feature merged (`LibraryBuilder.cpp` calls `HalFile::modificationTime()` but the crosspointer-reader/crosspoint-simulator repo doesn't expose it).

itsthisjustin:
So I think maybe this got auto closed when I merged the other library PR? I absolutely did not mean to close this on purpose. Honestly hadn't been following the discussion too closely here

itsthisjustin:
I do know series grouping has been a requested feature. Any thoughts on an sd plugin that could automatically do it for you and place them into folders instead?

hamaluik:
I'm sorry, that comment was unduly salty. And yes that is what happened, I forgot github does that.

I still don't have the ability to work on that expanded feature anytime soon unfortunately as I'm having surgery and won't have use of my hands for a couple months.

A plugin would be interesting I guess, that's similar to how I managed my library before this so would be helpful, but the library feature is there to use book metadata - series (and other collection metadata) definitely feels the province of a library feature.

itsthisjustin:
> I'm sorry, that comment was unduly salty. And yes that is what happened, I forgot github does that.
> 
> I still don't have the ability to work on that expanded feature anytime soon unfortunately as I'm having surgery and won't have use of my hands for a couple months.
> 
> A plugin would be interesting I guess, that's similar to how I managed my library before this so would be helpful, but the library feature is there to use book metadata - series (and other collection metadata) definitely feels the province of a library feature.

If you give me your blessing I'll certainly see if there's something I can do with your branch here!

hamaluik:
Please do! My fork has rebased the changes and fixed conflicts. I can continue to try to help where I can, but like I said, getting surgery next week so all I can reasonably do for a few weeks at least is beg an LLM to do things for me.

PR #3060, opened 2026-08-15
fix: keep reader background build at full speed

loic-daigle:
## Summary

* **What is the goal of this PR?** Fix slow page turns into not-yet-built pages during normal one-page-at-a-time reading, caused by the background section-build tick being starved by the idle CPU downclock.
* **What changes are included?** `EpubReaderActivity::loop()`'s background `buildSomeMore()` tick — meant to keep a small page-count cushion built ahead of the reader (`BUILD_WINDOW_AHEAD`) — now wraps that call in `HalPowerManager::Lock`, the same mechanism the render task already uses to force full CPU speed while it works.

### Why

That tick runs on the main-loop task, which has no clock lock of its own, so it silently inherited the idle downclock (as low as 10 MHz on non-PSRAM boards, after 500ms of no button input — see `HalPowerManager::IDLE_DOWNCLOCK_MS` / `LOW_POWER_FREQ`). Since a reader typically dwells on a page for several seconds between turns, nearly every background-build tick ran at that throttled clock instead of full speed, so the lookahead cushion rarely finished building in time. The practical effect: normal forward reading kept falling back to the synchronous catch-up build in `renderCurrentPage()` (`startBuild` + `buildSomeMore` loop) instead of hitting an already-built page — a visible stall on pages that should have been pre-built silently in the background.

The lock is non-blocking (won't contend with the render task's own lock) and scoped tightly to just the one `buildSomeMore()` call, so the main loop's own idle-downclock check — which re-evaluates every idle iteration, right after `activityManager.loop()` runs — pulls the clock back down again within the same loop pass. No sustained full-speed hold.

## Scope Check

Please confirm:

Notes on the checks above:
- This is a regression/gap specific to CrossPoint's own lazy incremental section-indexing design (introduced in `685d4e88`), so it isn't something stock firmware or another fork's differently-architected pagination would even exhibit — I haven't audited every fork, but the bug is intrinsic to this codebase's own background-build loop, not something upstream elsewhere.
- The change is entirely within `src/activities/reader/EpubReaderActivity.cpp`. It *consumes* the existing `HalPowerManager::Lock` public API but does not modify `freeink-sdk/`, `lib/hal/`, the bootloader, OTA, or recovery code, so no maintainer coordination was needed.

## Additional Context

* **Performance/battery implications**: the fix trades a small amount of extra full-speed CPU time (bounded to the duration of building 1-2 pages of text layout per idle tick, only while the cushion isn't yet full) for eliminating the synchronous catch-up stalls. No sustained full-speed hold — see mechanism above.
* **Memory/flash impact**: none. No new allocations; the change is one `#include` and one stack-allocated RAII lock in an existing function.
* **Build verification**: I was not able to run `pio run` / `pio check` in this environment (PlatformIO CLI not installed here), so this has only been verified by reading/tracing the code, not compiled. Please run CI and flag if it doesn't build.
* **Hardware verification**: tested on a physical device — no problem found.

---

### AI Usage

Did you use AI tools to help write this code? **YES** — Claude Code was used to trace the root cause across `EpubReaderActivity.cpp`, `HalPowerManager.cpp`, and `main.cpp`, and to write and verify the fix.

### DIFF
--- a/src/activities/reader/EpubReaderActivity.cpp
+++ b/src/activities/reader/EpubReaderActivity.cpp
@@ -5,6 +5,7 @@
 #include <FontCacheManager.h>
 #include <FsHelpers.h>
 #include <GfxRenderer.h>
+#include <HalPowerManager.h>
 #include <HalStorage.h>
 #include <I18n.h>
 #include <Logging.h>
@@ -338,6 +339,10 @@ void EpubReaderActivity::loop() {
       buildTickHeapGate()) {
     RenderLock lock;
     if (section->isBuilding() && buildTickHeapGate()) {
+      // The main-loop task inherits the idle downclock (as low as 10 MHz)
+      // unless raised here; the build must run at full speed to stay ahead
+      // of normal reading pace.
+      HalPowerManager::Lock powerLock;
       if (!section->buildSomeMore(BACKGROUND_BUILD_PAGES_PER_TICK)) {
         LOG_ERR("ERS", "Background section build failed");
         section.reset();


itsthisjustin:
This was on my list of things to do but I was concerned it would impact battery life too much. Seems like a good way to do it (better than my plan)

loic-daigle:
> This was on my list of things to do but I was concerned it would impact battery life too much. Seems like a good way to do it (better than my plan)

Does this fix also worry you about battery life? Or does it seem like a resonable tradeoff?

itsthisjustin:
> > This was on my list of things to do but I was concerned it would impact battery life too much. Seems like a good way to do it (better than my plan)
> 
> Does this fix also worry you about battery life? Or does it seem like a resonable tradeoff?

Everything worries me about battery life cause I currently have a horrible way to test battery life haha. We'd need to write stuff to sd card because having it plugged in all the time during develop is not helping us catch regression in battery usage

loic-daigle:
@itsthisjustin Anything more needed before you can review/merge this PR?

itsthisjustin:
So I think we need to close this as the cpu being held high while idle did in fact cause drain...correct me if I'm misunderstanding the PR

loic-daigle:
Well, the cpu speed is supposed to go down after the bg build has ended, so theoretically, it shouldn't affect very much, to what I remember.

BrianPugh:
>So I think we need to close this as the cpu being held high while idle did in fact cause drain...correct me if I'm misunderstanding the PR

I didn't read this PR at all, but IIRC [the previous battery drain was the result of a bug](https://github.com/crosspoint-reader/crosspoint-reader/pull/2996) in the "are we doing background tasks" detection logic. Basically it thought we were still doing background tasks perpetually, so we never downclocked. Doing background builds at fullspeed and then downclocking is optimal from both a user perspective, as well as from an energy-consumption perspective.

PR #3068, opened 2026-08-16
perf: make CJK UI fast with a bounded compressed glyph pool

serialx:
This PR makes CJK UI finally **usable**. After #3026 we got CJK in the UI, but the latency was unbearable — every Korean list row rebuilt its glyphs from SD, so a chapter-list repaint took ~928ms and repeated on every keypress. This PR delivers native-feeling CJK UI performance while using **~3KB more heap at worst and ~16KB less heap at best** (measured on device, table below).

Since #3026, UI strings containing CJK are redirected to SD-card fallback fonts, with glyphs cached in a per-string "mini arena" that is **replaced, not merged**, whenever a string's codepoints aren't fully resident. List screens draw many small disjoint strings, so each row evicts its neighbors' glyphs and rebuilds from SD. Measured on an Xteink X3 with a Korean library:

- a Korean browsing session performed **376 arena rebuilds totalling 14.2s of SD I/O** (equivalent English session: 8 rebuilds, 0.12s);
- a single chapter-list repaint took **928ms** (34 rebuilds), repeated on every selection keypress;
- each row cost ~3 rebuilds (measure → ellipsis → draw), ~75ms.

The goal: glyphs loaded once should be reused across rows, screens, and repaints — within a strict, fixed memory budget. The reader path keeps the existing arena unchanged; its workload (one page-sized union, subset hits on page turns) is what the arena is good at.

### The glyph pool

`UiGlyphPool` is one bounded LRU-style cache shared by the 8/10/12pt UI fallback instances: a single ~14.3KB block allocated at font setup, holding a sorted 384-entry index (16B/entry: key, metrics, kern classes) and an 8KB arena of compressed bitmaps.

- **Per-glyph granularity, merge-only.** A miss loads one glyph; nothing else is disturbed. Row 2 reuses what row 1 loaded; a keypress repaint is ~100% pool hits — pure RAM blits, zero SD I/O.
- **Second-chance eviction over a compacted arena.** Each `find()` sets a ref bit; when space is needed, the oldest entry is dropped unless its ref bit earns it another lap. Bytes stay contiguous by construction — no fragmentation, no runtime growth. Entries are born referenced so a fill can never evict the string it is loading before the blit reads it.
- **Whole-block lifecycle.** For heap-critical transitions (web server + WiFi) the pool releases as a single `free()` and revives lazily behind a heap guard. One deliberate trade-off: while the heap stays below the revive floor (typically exactly the web-server screens the release exists for), CJK UI strings render blank instead of rebuilding caches into a heap that cannot hold them — the degraded-but-alive counterpart to the OOM aborts #3035 fixed. Recovery is automatic once the heap clears the floor.
- **Minimal integration surface.** `SdCardFont` gains a pool mode: `prewarm()` batch-fills the pool with chunked, offset-sorted SD reads, and per-glyph lookups are served through the existing glyph-miss path. `GfxRenderer` is untouched. No `.cpfont` format changes — existing font files work as-is.

Compared with develop's arena path:

| | Arena (develop) | Glyph pool (this PR) |
|---|---|---|
| Cache unit | whole string set, replace-on-rebuild | single glyph, merge-only |
| Resident glyph capacity | one string per instance (typically 10–30 glyphs) | 384 glyphs shared across instances (≈ a full screen plus session vocabulary; largest measured screen used ~170 unique) |
| Reuse across rows / screens | none (each string evicts the last) | full (Zipfian CJK vocabulary shares glyphs) |
| Warm repaint | ~928ms of SD rebuilds | RAM blits, ~0 SD I/O |
| Allocation behavior | per-string realloc churn | one block at font setup, never resized |
| Resident kern tables | ~6.7KB per UI instance, lazily accumulated | none (transient reads + 1KB pair cache) |
| Memory bound | emergent (capacities converge on usage) | hard (14.3KB, fixed) |

### Glyph compression: 2D context modeling + arithmetic coding

To fit useful capacity into 8KB, pooled bitmaps are compressed by `UiGlyphCodec`:

1. **1-bit threshold conversion** (2× smaller): BW-mode rendering draws any non-white 2-bit level as black, and UI text only ever renders in BW mode (grayscale passes draw book content and bitmaps only) — so dropping the gray levels at load time is pixel-exact for everything the pool serves.
2. **2D context model**: each pixel is predicted from its three causal neighbors (west, north, northwest). Glyph bitmaps are strongly 2D-correlated — a pixel continuing a stroke is ~90–98% predictable — which byte-oriented compressors can't see. The model is a static 8-entry probability table (16 bytes of flash), trained offline on 38k CJK glyphs; it captures stroke continuation, not font-specific shapes.
3. **Binary arithmetic coding**: the prediction error stream is entropy-coded with an LZMA-style range coder. The decoder reads its context from its own already-decoded output, so encode and decode need no per-font state and no heap.

Measured on a 501-glyph Korean UI corpus: **19.2 B/glyph stored vs 75.2 B/glyph for the 2-bit originals (3.9×)** — this is what lets 384 glyphs (a full screen plus session vocabulary) fit in 8KB. Glyphs that don't compress below raw 1-bit are stored raw, so the codec can never lose to plain conversion. Decode cost is a few tens of µs per glyph, invisible under the e-ink refresh.

The design space was measured before implementation. All candidates below were evaluated on the same 167-glyph Korean UI corpus (Pretendard 10pt — the list-row size), cross-validated wherever a dictionary or table is trained, after the 1-bit conversion where applicable:

| Approach | Stored B/glyph | Ratio | Why rejected |
|---|---|---|---|
| 2-bit raw (what the arena caches today) | 73.8 | 1.0× | baseline |
| 1-bit threshold only | 37.1 | 2.0× | kept — stage 1 of the codec |
| per-glyph deflate | 37.0 | 2.0× | no gain at ~37-byte payloads |
| zstd-19 + 2KB trained dictionary | 33.1 | 2.2× | +2KB dictionary RAM — net loser |
| RLE + Golomb-Rice coded runs (best scan order) | 23.4 | 3.2× | tableless runner-up, beaten by 2D context |
| per-glyph deflate + 4KB dictionary | 23.0 | 3.2× | +4KB dictionary RAM — net loser |
| grouped deflate (32 glyphs/group) | 22.2 | 3.3× | eviction granularity becomes the group |
| whole-set LZMA (theoretical bound) | 18.6 | 4.0× | no random access — unreachable in a cache |
| **ctx3 + arithmetic coding (this PR)** | **19.3** | **3.8×** | **chosen: per-glyph random access, zero RAM overhead, 16B flash table** |

The chosen codec lands within 4% of the whole-set LZMA bound while keeping per-glyph random access — the 2D context model sees the stroke-level correlation that byte-oriented compressors structurally cannot.

### Kerning

The vast majority of CJK glyphs have no kerning at all — Pretendard's kern tables contain zero Hangul syllable entries, 97% of kerned codepoints are Latin/punctuation, and 86% of class-pair matrix cells are zero. Kern data is that sparse, so this PR changes how it is kept in the heap: instead of the arena path's ~6.7KB of resident class tables per UI instance, kern class IDs ride in each pool entry (two bytes that fit the entry's existing layout), and pair values live in a 256-entry sparse cache — zeros included, so a "no kern here" answer never re-probes SD — prefetched during prewarm as single-byte reads of the on-file matrix. Total: ~1KB, ~4% of the arena path's kern memory, with kerning fully preserved.

The render path stays RAM-only via a new `EpdFontData::kernHandler`, consulted only when no kern matrix is resident (nullptr for every other font); the common CJK–CJK pair short-circuits on its zero class IDs without any lookup. Verified in hardware testing.

### Memory usage observations

Measured with an on-screen free-heap overlay (not part of this PR), same navigation route on both builds; the chapter-list pair used a controlled protocol (fresh boot → same book and position → 3 page turns → chapter list) because reader session state dominates that screen otherwise. FH = free heap, MA = largest contiguous free block.

| Screen | develop FH / MA | this PR FH / MA | Δ FH |
|---|---|---|---|
| Home (Korean book card) | 96,736 / 90,100 | 93,472 / 81,908 | −3,264 |
| File browser (Korean titles) | 103,408 / 90,100 | 101,076 / **98,292** | −2,332 |
| **Chapter list (book resident)** | 49,384 / 25,588 | **65,544 / 51,188** | **+16,160** |

The pattern: the pool pays its fixed ~14.3KB up front at font setup, while the arena path accumulates its costs lazily (kern tables load on first CJK draw per instance, arenas grow toward their high-water). On light screens with ~100KB free, the pool is ~2–3KB more expensive; on the worst-case screen — where memory actually matters — it frees **16KB** and **doubles the largest contiguous block** (the fragmentation difference between per-string realloc churn and a block that never moves).

Six commits, reviewable independently:

1. `feat: add ctx3 glyph codec for UI font caching` — pure, no dependencies
2. `feat: add bounded shared glyph pool for UI fonts` — pure, host-testable
3. `feat: serve UI fallback glyphs from shared pool` — the only commit touching existing behavior (`SdCardFont` pool mode + `SdCardFontSystem` wiring)
4. `feat: kern support for pool-served UI glyphs` — adds `EpdFontData::kernHandler`
5. `perf: right-size UI glyph pool arena to 8KB` — measured tuning
6. `test: host coverage for UI glyph codec and pool`

### Testing

- **Host**: new gtest suite (`test/ui_glyph_pool`) — codec roundtrip pixel-exactness across densities and packing edge cases, pool content integrity under eviction fuzz against a shadow model, same-string fill survival, kern pair cache mechanics, release/reinit. All existing suites unaffected.
- **On device** (Xteink X3, Korean library + mixed-script titles): list/browser/chapter-list rendering verified pixel-identical and dramatically faster; kerning verified on Latin pairs inside mixed strings; reader regression sweep clean (pool mode never touches the reader path); web-server transition releases and revives the pool; heap A/B measured as above. Primarily tested in portrait — the pool serves glyphs orientation-independently (no coordinate math), but a sweep of the other three orientations by testers would be welcome.

### AI usage disclosure

This PR was done with Claude Code with Fable 5: the compression design space (RLE, Huffman, dictionaries, zstd, context modeling) was explored and measured on real font data host-side before implementation, and the implementation was host-tested against a 501-glyph corpus before any firmware flash. All on-device measurements and visual verification were performed by a human on real hardware.


serialx:
@coderabbitai Review feedback addressed in the incoming commit:

- **Slot-buffer leak on decode failure**: fixed via your suggested robust option — `clearOverflow()` now sweeps all `OVERFLOW_CAPACITY` slots, and the rollback only decrements when the serve actually incremented the count.
- **Stale `enableUiGlyphPool` doc**: corrected — kerning is preserved, only ligatures are skipped.
- **Per-test LCG seeding**: added `seedLcg()` at the start of every test using the synthetic corpus or the op stream.
- **Arena size vs `uint16_t` offsets**: `reinit()` now rejects arenas over 0xFFFF bytes.
- **Shared heap floor**: both gates now reference `UiGlyphPool::MIN_FREE_HEAP_FOR_ALLOC`.

One item deliberately not changed: **zero-length entries holding index slots**. Zero-area glyphs are bounded to a handful of codepoints (space, per instance/style) — at most ~1–2% of the 384 slots for the pool lifetime — and they are also among the most frequently drawn glyphs, so evicting them preferentially would be LRU-backwards. The eviction loop is fuzz-verified as-is; special-casing it for that capacity margin isn't worth the risk. Happy to revisit if maintainers disagree.

serialx:
@coderabbitai Check [8694a47](https://github.com/crosspoint-reader/crosspoint-reader/pull/3068/commits/8694a47a368cdf081431048d87fb92e55ac2cf92)

itsthisjustin:
Ha! I think we were both working on the same thing at different times. Can you take a look at this and tell me what's overlap or if there's a preferred one to merge? https://github.com/crosspoint-reader/crosspoint-reader/pull/3071

itsthisjustin:
> Ha! I think we were both working on the same thing at different times. Can you take a look at this and tell me what's overlap or if there's a preferred one to merge? #3071

My PR packs two changes where I could redo a smaller PR to only fix the list overlap issues. I am not sure if you do this but lazy loading the TOC was a huge win on mine

serialx:
> > Ha! I think we were both working on the same thing at different times. Can you take a look at this and tell me what's overlap or if there's a preferred one to merge? #3071
> 
> My PR packs two changes where I could redo a smaller PR to only fix the list overlap issues. I am not sure if you do this but lazy loading the TOC was a huge win on mine

lol! Didn't think you were working on CJK optimizations. I was developing and testing this PR this whole day. :'D

I didn't do any of lazy loading the TOC. Rather optimization comes naturally from glyph cache. I thought UI refactoring PR be hard for maintainers to merge, as I am not familiar with the UI architecture and maintainer's tastes.

> Two-line wrapped labels meant the selection could sit on a never-drawn row, page jumps skipped rows entirely, and rows past count - visibleRows were unreachable.

I definitely experienced this bug, but didn't have time to find root cause!

Your PR will help with same screen latency. My PR, at the cost of 3KB heap usage, will help with multi-screen CJK font rendering. So I think my patch is ~kinda orthogonal to yours. Arguably, mine is better because cross-screen latency is better in some cases.

serialx:
@itsthisjustin One side note: as I was developing this feature, I thought of using this glyph compression in the system fonts would also save flash space. At little more cost to CPU cycle per glyph rendering, we can save some firmware size.

Currently 34% of the entire firmware is builtin font data (1,955 KB). Of that 1,336 KB is glyph bitmaps, 2-bit, DEFLATE group-compressed.

But I cannot directly apply this PR's compression algorithm, because it assumes 1-bit BW bitmaps. I have some ideas on compressing 2-bit grayscale AA glyphs. If you're interested, I can experiment it too.

itsthisjustin:
> @itsthisjustin One side note: as I was developing this feature, I thought of using this glyph compression in the system fonts would also save flash space. At little more cost to CPU cycle per glyph rendering, we can save some firmware size.
> 
> Currently 34% of the entire firmware is builtin font data (1,955 KB). Of that 1,336 KB is glyph bitmaps, 2-bit, DEFLATE group-compressed.
> 
> But I cannot directly apply this PR's compression algorithm, because it assumes 1-bit BW bitmaps. I have some ideas on compressing 2-bit grayscale AA glyphs. If you're interested, I can experiment it too.

I had done that compression in another pr actually, it does save space. I don't think we even need 2 bit fonts, we aren't rendering them as such. We could save a lot via compression and just using 1 bit

serialx:
> > @itsthisjustin One side note: as I was developing this feature, I thought of using this glyph compression in the system fonts would also save flash space. At little more cost to CPU cycle per glyph rendering, we can save some firmware size.
> > Currently 34% of the entire firmware is builtin font data (1,955 KB). Of that 1,336 KB is glyph bitmaps, 2-bit, DEFLATE group-compressed.
> > But I cannot directly apply this PR's compression algorithm, because it assumes 1-bit BW bitmaps. I have some ideas on compressing 2-bit grayscale AA glyphs. If you're interested, I can experiment it too.
> 
> I had done that compression in another pr actually, it does save space. I don't think we even need 2 bit fonts, we aren't rendering them as such. We could save a lot via compression and just using 1 bit

But if you burn system font as 1 bit, wouldn't users using system fonts to render ebooks experience worse font rendering?

itsthisjustin:
> > > @itsthisjustin One side note: as I was developing this feature, I thought of using this glyph compression in the system fonts would also save flash space. At little more cost to CPU cycle per glyph rendering, we can save some firmware size.
> > > Currently 34% of the entire firmware is builtin font data (1,955 KB). Of that 1,336 KB is glyph bitmaps, 2-bit, DEFLATE group-compressed.
> > > But I cannot directly apply this PR's compression algorithm, because it assumes 1-bit BW bitmaps. I have some ideas on compressing 2-bit grayscale AA glyphs. If you're interested, I can experiment it too.
> > 
> > I had done that compression in another pr actually, it does save space. I don't think we even need 2 bit fonts, we aren't rendering them as such. We could save a lot via compression and just using 1 bit
> 
> But if you burn system font as 1 bit, wouldn't users using system fonts to render ebooks experience worse font rendering?

I don't think we even render them at 2bit currently. I could be wrong but it's what I remember from working on this a month or so ago

serialx:
> > > > @itsthisjustin One side note: as I was developing this feature, I thought of using this glyph compression in the system fonts would also save flash space. At little more cost to CPU cycle per glyph rendering, we can save some firmware size.
> > > > Currently 34% of the entire firmware is builtin font data (1,955 KB). Of that 1,336 KB is glyph bitmaps, 2-bit, DEFLATE group-compressed.
> > > > But I cannot directly apply this PR's compression algorithm, because it assumes 1-bit BW bitmaps. I have some ideas on compressing 2-bit grayscale AA glyphs. If you're interested, I can experiment it too.
> > > 
> > > 
> > > I had done that compression in another pr actually, it does save space. I don't think we even need 2 bit fonts, we aren't rendering them as such. We could save a lot via compression and just using 1 bit
> > 
> > 
> > But if you burn system font as 1 bit, wouldn't users using system fonts to render ebooks experience worse font rendering?
> 
> I don't think we even render them at 2bit currently. I could be wrong but it's what I remember from working on this a month or so ago

I told codex to find it out. We definitely use 2 bit fonts for rendering books. That was my experience too.

---

# Embedded Firmware Fonts

CrossPoint embeds separate fonts for reader content and firmware UI text. The
reader fonts retain four coverage levels for anti-aliasing, while the UI fonts
use monochrome bitmaps.

The flash sizes below were measured from the linked `.rodata` sections in the
default build's `firmware.map` on 2026-08-16. **Bitmap payload** is the actual
bitmap array stored in the firmware, compressed where indicated. **Total linked
flash** also includes glyph metrics, Unicode intervals, kerning tables,
ligatures, and compressed-group indexes. Small alignment padding is excluded.

| Embedded font | Sizes and styles | Primary use | Embedded / rendered | Bitmap payload | Total linked flash |
|---|---|---|---|---:|---:|
| **Noto Serif** | 12, 14, 16, 18 pt; regular, bold, italic, bold-italic | Default EPUB and TXT body font | 2-bit compressed; rendered 2-bit with Text AA, otherwise 1-bit | **623.8 KiB compressed** | **1,054.6 KiB** |
| **Noto Sans Reader** | 12, 14, 16, 18 pt; regular, bold, italic, bold-italic | Alternative EPUB and TXT body font | 2-bit compressed; rendered 2-bit with Text AA, otherwise 1-bit | **577.2 KiB compressed** | **1,009.1 KiB** |
| **Noto Sans Small** | 8 pt; regular | Status bar, hints, paths, and secondary labels | 1-bit uncompressed; rendered 1-bit | **18.2 KiB** | **52.6 KiB** |
| **Ubuntu UI** | 10 and 12 pt; regular and bold | Menus, lists, titles, dialogs, home screen, keyboard, network/settings screens, and reader messages | 1-bit uncompressed; rendered 1-bit | **116.5 KiB** | **298.6 KiB** |
| **Total** | **37 font assets** | | | **1,335.7 KiB** | **2,414.9 KiB / 2.36 MiB** |

## Compressed reader bitmap payload by point size

Each value includes all four styles at that point size.

| Family | 12 pt | 14 pt | 16 pt | 18 pt |
|---|---:|---:|---:|---:|
| Noto Serif | 115.9 KiB | 143.3 KiB | 167.7 KiB | 196.9 KiB |
| Noto Sans | 107.2 KiB | 132.5 KiB | 155.3 KiB | 182.2 KiB |

## Rendering details

The firmware always starts with a 1-bit framebuffer. When Text AA is enabled,
the reader renders the 2-bit Noto glyph coverage through a black-and-white base
and two grayscale bit-plane passes, producing four display levels. Text AA is
enabled by default. If it is disabled or the grayscale scratch allocation
fails, every non-white glyph coverage level is flattened to black and the text
is displayed as 1-bit.

The Ubuntu UI and Noto Sans Small glyphs are embedded as 1-bit bitmaps and
remain 1-bit when rendered.

The font families and styles are instantiated in
[`src/main.cpp`](../src/main.cpp). Generated font headers under
[`lib/EpdFont/builtinFonts/`](../lib/EpdFont/builtinFonts/) record their source
bit depth and compression mode. The 2-bit glyph groups are decompressed on
demand by `GfxRenderer::getGlyphBitmap()`.

Builds defining `OMIT_FONTS` retain Noto Serif 14, Noto Sans Small, and Ubuntu
UI, but omit the other Noto Serif sizes and all Noto Sans reader sizes. Their
flash totals therefore differ from the standard build shown above.

serialx:
@itsthisjustin Did some experiments, I think I can improve current system font compression by using an improved compression algorithm like in this PR and save at least 264KB of flash back. Does this kind of improvement warrant a PR? I want to know what you think.

itsthisjustin:
> @itsthisjustin Did some experiments, I think I can improve current system font compression by using an improved compression algorithm like in this PR and save at least 264KB of flash back. Does this kind of improvement warrant a PR? I want to know what you think.

Works for me. I didn't see any performance loss when I did compression in the past

itsthisjustin:
> > @itsthisjustin Did some experiments, I think I can improve current system font compression by using an improved compression algorithm like in this PR and save at least 264KB of flash back. Does this kind of improvement warrant a PR? I want to know what you think.
> 
> Works for me. I didn't see any performance loss when I did compression in the past

See if you can find my closed pr for AA fast mode. I flipped compression on in it

serialx:
> > @itsthisjustin Did some experiments, I think I can improve current system font compression by using an improved compression algorithm like in this PR and save at least 264KB of flash back. Does this kind of improvement warrant a PR? I want to know what you think.
> 
> Works for me. I didn't see any performance loss when I did compression in the past

I've just submitted #3083 which saves 556 KB flash space and also helps with heap usage too.

If you choose to merge #3083 then please don't merge this PR. (#3068) I have ideas to improve this PR when #3083 is merged.

serialx:
I'll close this PR in favor of #3071. I'll come back with a better CJK optimizations soon!

PR #3071, opened 2026-08-16
fix: Slow CJK TOC & Lists

itsthisjustin:
With an SD fallback font installed for CJK glyphs any screen containing more than one CJK string was effectively unusable.

1. Every distinct CJK string evicted every other one, so full-screen immediate-mode repaints re-read the SD forever. The reader status bar (drawn after the page's PrewarmScope) wiped the page cache each turn, and AA strip rendering re-draws the page 20 times, multiplying the damage. 
2. The chapter list runs over an open book. Materializing all 547 TOC entries up front cost ~60KB, kern/ligature class tables (~3.3KB, useless for Hangul: the mini matrix is 1x1) loaded on every prewarm

### Font cache (lib/EpdFont, lib/GfxRenderer)

- `prewarmStyle()` unions resident + requested codepoints instead of replacing
- screens batch-prewarm all their static strings in one SD pass at data-load time with zero transient allocation. 
- UI-driven prewarms skip kern/ligature loading (`loadKernLig=false`) 
- `FontCacheManager` scans per font id (4 slots) instead of first-font-wins, so mixed-font pages batch-prewarm the SD fallback even when Latin text is recorded first.
- `GfxRenderer::prewarmFallbackText()` resolves the fallback and batch-loads a screenful of strings (including the truncation ellipsis automatically). Wired into the file browser (rows + path band), recent books (bold titles, regular authors), and the chapter list window. 
- SDK: `ListProps::itemsWindowFirst` lets `items` be a small window around the viewport. Entry cost no longer scales with TOC size; the 547-entry TOC dropped ~60KB of RAM. 

| Scenario | Before | After |
| --- | --- | --- |
| Chapter list row step | 3.2s + SD reads every press | ~800ms, zero SD |
| Chapter list first paint | 2.2s + OOM errors | 239ms |
| Entering chapter list | abort() (string OOM) | heap stable |
| Free heap in chapter list | 10.7K free / 2.5K max-alloc | 33K free |
| Missing/unreachable last file | not listed, highlight lost | reachable |

### DIFF
--- a/lib/EpdFont/SdCardFont.cpp
+++ b/lib/EpdFont/SdCardFont.cpp
@@ -767,6 +767,42 @@ int SdCardFont::prewarm(TextGetter getter, void* ctx, uint32_t textCount, uint8_
 
   unsigned long startMs = millis();
 
+  // Cap the unique-codepoint budget by what the heap can actually hold as a
+  // full mini arena (glyph structs + bitmaps, working headroom left over).
+  // Multi-string batches only: a several-hundred-chapter CJK table of
+  // contents would otherwise extract up to MAX_PAGE_GLYPHS and fail the whole
+  // arena allocation — better to load the first screens' worth and let
+  // scrolling union-in the rest page by page. Per-string requests are small
+  // and already bounded by the union gate in prewarmStyle (running the check
+  // there would also log per draw call); metadata-only prewarms load no
+  // bitmaps. Bytes/glyph prefers the measured average from the resident mini:
+  // Hangul ink boxes run well under the advanceY-squared em estimate, which
+  // otherwise roughly halves the usable budget.
+  uint32_t cpBudget = MAX_PAGE_GLYPHS;
+  if (!metadataOnly && textCount > 1) {
+    uint8_t refStyle = MAX_STYLES;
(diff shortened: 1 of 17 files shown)


Uri-Tauber:
How is it compared with #3068?

Which solve the problem better, and which take less flash (here I think you win).

itsthisjustin:
> How is it compared with #3068?
> 
> Which solve the problem better, and which take less flash (here I think you win).

In talking with contributor they seem to solve different edges of the same problem. I will say my fixes are technically not limited to just CJK here. The kerning stuff is but there's other general improvements especially around TOC that will help all languages. This book I was testing with had over 500 toc entries and were huge multi line strings too. Definitely a good pressure test epub

serialx:
I have not tested this yet, but I think we should take this over my #3068. My PR is strictly more large, needs more testing, etc. This PR is directly from the maintainer, minimal, better for many more cases.

Uri-Tauber (approved):
Approving base on code review.
I haven't tested it, since I'm not CJK speaker. @serialx Could you test this, and compare vs your implementation?
If they are both the same from usability perspective, I will go with this one.

PR #3077, opened 2026-08-16
chore: revert light sleep code but keep battery drain fix

itsthisjustin:
Simplifies USB connection detection by removing SOF-based sampling and I2C throttling logic. USB state is now determined by a single isUsbConnected() call instead of complex updateUsbState() polling. Also removes the display busy wait slice hook mechanism and associated debounce pending checks.

Reverts #2525

### DIFF
--- a/lib/hal/HalDisplay.cpp
+++ b/lib/hal/HalDisplay.cpp
@@ -93,10 +93,6 @@ bool HalDisplay::isInverted() const { return einkDisplay.isInverted(); }
 
 void HalDisplay::deepSleep() { einkDisplay.deepSleep(); }
 
-void HalDisplay::setBusyWaitSliceHook(bool (*sliceHook)(int8_t busyPin, uint8_t busyLevel)) {
-  einkDisplay.setBusyWaitSliceHook(sliceHook);
-}
-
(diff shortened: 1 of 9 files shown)


Uri-Tauber:
Anything from coderabbit comments is relevant?

itsthisjustin:
Confirmed fixes my issue of freezing when I plug in my usb and another locked user who can't use nightly without usb always plugged in

itsthisjustin:
> Anything from coderabbit comments is relevant?

yeah I find these odd. It's reverting code...so these issues would have pre-existed?

Uri-Tauber:
> yeah I find these odd. It's reverting code...so these issues would have pre-existed?

Could be.
It kill me [they raised 143M](https://www.coderabbit.ai/blog/introducing-agentic-change-management)...

Uri-Tauber (approved):

BrianPugh:
I can confirm (on an x3) that this restores the original v1.4.0 idle power consumption (9.79mA). I can't get the v1.5.0 max-power-consumption (~40mA) to trigger. So 👍 from me if we cannot get a solution like #3082 to work.

PR #2596, opened 2026-07-13
feat: Arabic/Farsi/Urdu glyphs in built-in UI fonts - PR 2/3

YouHusam:
Add Noto Sans Arabic as a fontstack layer to the five built-in UI fonts (ubuntu 10/12 regular+bold, notosans 8) so shaped Arabic UI text renders on menus, settings, and file-browser titles.

Coverage targets the *output* of MiniBidi do_shape() rather than base letters: Presentation Forms-B U+FE80-FEFC (all core Arabic contextual forms plus the eight Lam-Alef ligature forms), the Presentation Forms-A blocks of the Farsi/Urdu extra letters, the few codepoints shaping leaves unshaped (hamza, tatweel, noon ghunna base), Arabic punctuation, and both digit sets. No harakat and no Sindhi/Pashto/Kurdish forms - book text gets full coverage from SD-card fonts.

Measured firmware size delta (pio run -e default, real before/after builds): 49,824 bytes (+48.7 KB), of which core Arabic Forms-B is 29.2 KB, Farsi/Urdu Forms-A 9.9 KB, digits 3.6 KB, punctuation 0.9 KB, and kerning-class/interval table growth ~5 KB.

## Summary

* **What is the goal of this PR?** 
Add Arabic/Farsi/Urdu support to the UI components.

* **What changes are included?**
Changes to the UI fonts.

## Additional Context

* Add any other information that might be helpful for the reviewer (e.g., performance implications, potential risks, 
  specific areas to focus on).
No

---

### AI Usage

Did you use AI tools to help write this code? _**YES**_

Chapter title renders correctly in status bar

Random Urdu Book, I can see that it's rendering correctly but I don't understand so I can't verify correctness

PR #3079, opened 2026-08-16
feat(fonts): add --mono flag for crisper 1-bit UI fonts

fain182:
## Summary

* **What is the goal of this PR?** Make the built-in 1-bit UI fonts render crisply instead of over-inked.
  * `fontconvert.py`: new `--mono` flag that rasterises 1-bit fonts with FreeType's native monochrome renderer (`FT_LOAD_TARGET_MONO`) instead of rendering antialiased greyscale and thresholding it.
  * `convert-builtin-fonts.sh`: pass `--mono` for the Ubuntu UI fonts only.
  * Regenerated `ubuntu_{10,12}_{regular,bold}.h`.

Without the flag the 1-bit path behaves exactly as before, so `notosans_8_regular` and all 2-bit reading fonts are untouched.

### Background

The 1-bit UI fonts are rendered antialiased and then downsampled with "treat any 2+ as black" — roughly 12% coverage is enough to turn a pixel black. That over-inks stems: at UI sizes Ubuntu's 2px stems come out 3–4px wide with uneven edges.

`FT_LOAD_TARGET_MONO` is hinted and drop-out controlled, so stems land on whole pixels. Glyphs stay 1 bit/pixel, and advance widths come from `linearHoriAdvance`, which is independent of the render mode — text layout and line breaking are bit-for-bit identical.

| Before — antialiased + threshold | After — `--mono` (native monochrome) |
| --- | --- |
| IMG_9395 | IMG_9400 |
| IMG_9396 | IMG_9403 |
| IMG_9397 | IMG_9402 |
| IMG_9399 | IMG_9401 |

This supersedes #2466, which made monochrome the default for *every* 1-bit font and was declined because thinner faces then look too thin. That matches the analysis from @jpirnay and @rmmh on that PR: monochrome suits well-hinted faces such as Ubuntu, but not thin ones whose sub-pixel stems need to be biased back to a solid pixel. So here it is opt-in per font, and applied only to Ubuntu.

## Scope Check

- [x] If this PR touches `freeink-sdk/`, `lib/hal/`, the bootloader, OTA, or recovery code, I have coordinated with the relevant maintainer. *(N/A — none of these are touched.)*

## Additional Context

* **Flash:** −10,288 bytes (5,714,032 → 5,703,744, `default` env). The old thresholding inflated glyph bitmaps, so dropping it shrinks them by ~7%.
* **RAM:** unchanged — glyph bitmaps live in flash, and the free-heap reading on device is identical within noise.
* **Glyph coverage is unchanged.** The regenerated headers keep byte-identical interval tables, including the Arabic/Farsi/Urdu presentation forms added in #2596; only the bitmap data differs.
* Does not touch `freeink-sdk/`, `lib/hal/`, the bootloader, OTA, or recovery code.
* Built and verified on X4 hardware.

---

### AI Usage


itsthisjustin:
I'm in favor of this but I think the rest of the maintainers felt it was too thin. @uxjulia @Uri-Tauber ?

Uri-Tauber:
@itsthisjustin you know me well.

fain182:
How heavy the UI should look is your call — that's taste, and I won't argue it.

What isn't taste is that the current rendering is wrong. The 1-bit downsample marks a pixel black at ≥12.5% coverage, so every stem edge picks up an extra column: Regular renders 3px stems where the font designs 2px, Bold 5px where it designs 4px. Over ASCII 0x21–0x7E that's +17% to +34% more ink than the outline actually contains, depending on size and weight. `--mono` lands within ±8%. So it isn't thin, it's correct — what we're used to is Ubuntu Regular plus a pixel.

So if you want the UI heavier, take it from the weight, not from the threshold: Ubuntu Medium (same family, same UFL licence) with `--mono` gives 3px stems at both sizes — exactly today's width — and stays crisp. Happy to open that as a follow-up.

uxjulia:
I do think it looks better from a font integrity standpoint, but as a UI font, it does appear too thin and I think we'd be doing low-vision users a disservice. @fain182 can you generate this same PR with the Ubuntu weight bumped up to medium? I'd love to see how that looks.

fain182:
Opened #3096 with the Medium variant, which should answer the weight question: same stem width as the UI has today, screenshots in the PR.

It's built on top of this branch, so it already contains this commit: merging it alone covers both, and this PR can just be closed.

fain182:
Superseded by #3096, merged as 66a7ae1 — it was built on this branch, so `--mono` is already on `develop`.

PR #3089, opened 2026-08-17
fix: unblock X4 Pro frontlight double-click when short power press = Sleep, add Home key option

### LINKED ISSUE #3087: X4 Pro - Physical Button to toggle Front Light
### Scope Self-Check (required)

- [x] This is **not** a new theme or theming change (themes are temporarily closed pending the move to SD-loaded themes).
- [x] This is **not** a new external network connector (sync engine, cloud storage, remote file access, OPDS extensions beyond what exists, or any new "talk to a server" feature).
- [x] This is **not** an interactive app (game, calculator, notepad), writing/authoring tool, RSS/news/browser feature, media playback feature, or PDF rendering.
- [x] The stock firmware does **not** already handle this well.
- [x] No other popular CrossPoint fork already handles this well (or, if one does, I explain below why CrossPoint still needs it).

### Problem this solves

If I want to read something, but I'm already in a dark room, I can't easily navigate the swipe menus to turn on the front light. I'd love to have a "long press some button" to turn on front light.

### Why the stock firmware (and other forks) do not already solve this

I have not yet found this feature in the stock firmware, and I'm a crosspoint purist, so I'd prefer it to exist here :) 

### Proposed change

One of the "Long Press" options should be to "Toggle Frontlight"

### Memory / flash / complexity cost

_No response_

### Additional context

_No response_

naydichev:
## Summary

* **What is the goal of this PR?** Fix the X4 Pro's power-button double-click frontlight toggle so it actually works when `Short Power Button Click` is set to `Sleep`, and add a setting to move the gesture to the Home key instead.
  * `getPowerButtonDuration()` returns a 10ms hold threshold for the `Sleep` short-press action, so the existing level-triggered sleep check fired on button-down, before the double-click handler ever saw a release. The device slept on the very first click and the second click never landed.
  * The sleep (and existing `PWR_CONFIRM`) actions are now deferred until the double-click window passes, mirroring the deferral pattern already used for `PWR_CONFIRM`.
  * Added a `Frontlight Toggle` setting (Power / Home) so the double-click/double-tap gesture can live on the Home key instead of Power, sidestepping the conflict entirely. Home's own single-tap "go home" gesture is deferred the same way while its double-click window is pending.

## Scope Check

Please confirm:

Closes #3087

## Additional Context

Built and flashed to a physical X4 Pro; confirmed double-click-to-toggle-frontlight now works with `Short Power Button Click: Sleep`, and that switching `Frontlight Toggle` to Home works as a double-tap while still allowing a single Home tap (delayed up to the ~500ms double-click window) to exit as normal.

`pio run` (x4pro + default envs) and `pio check` (x4pro) pass. `./bin/clang-format-fix` was not run — `clang-format` isn't installed in this environment; happy to run it or take a follow-up push if CI's format check flags anything.

No flash/RAM impact beyond one new `uint8_t` setting field and a few extra i18n strings.

---

### AI Usage


canardleteer:
> **Human Note:**, just testing my updated skills from another PR, all the rest of this is :robot: generated, @naydichev .

`frontlightToggleButton` is added in `getSettingsList()` for every binary, then hidden only in `SettingsActivity::rebuildSettingsLists` when `!BoardConfig::isX4Pro()`. That on-device filter does not apply to `CrossPointWebServer::handleGetSettings()`, which serves the same list.

On `default` / `sticky` / `papermono` the web UI can therefore offer **Frontlight Toggle → Home**. Those boards have no Home key (`BoardConfig::hasHomeKey` is only true on X4 Pro in this compile set), and only X4 Pro runs `handleX4ProFrontlightDoubleClick()`. Paper Mono is the easy misread: it has a frontlight (`pmic-pwm`) so the label looks real, but the gesture never arms.

Please drop the row in `getSettingsList()` when `!BoardConfig::isX4Pro()`, same as the existing `hasHomeKey()` removal of `STR_TAP_FOR_READER_MENU`. The `uint8_t` on `CrossPointSettings` can stay so a copied `settings.json` still round-trips.

naydichev:
> > **Human Note:**, just testing my updated skills from another PR, all the rest of this is 🤖 generated, @naydichev .
> 
> `frontlightToggleButton` is added in `getSettingsList()` for every binary, then hidden only in `SettingsActivity::rebuildSettingsLists` when `!BoardConfig::isX4Pro()`. That on-device filter does not apply to `CrossPointWebServer::handleGetSettings()`, which serves the same list.
> 
> On `default` / `sticky` / `papermono` the web UI can therefore offer **Frontlight Toggle → Home**. Those boards have no Home key (`BoardConfig::hasHomeKey` is only true on X4 Pro in this compile set), and only X4 Pro runs `handleX4ProFrontlightDoubleClick()`. Paper Mono is the easy misread: it has a frontlight (`pmic-pwm`) so the label looks real, but the gesture never arms.
> 
> Please drop the row in `getSettingsList()` when `!BoardConfig::isX4Pro()`, same as the existing `hasHomeKey()` removal of `STR_TAP_FOR_READER_MENU`. The `uint8_t` on `CrossPointSettings` can stay so a copied `settings.json` still round-trips.

> thanks for the feedback, my 🤖 has read your comments and adapted the change. this message brought to you by human clickity-clacking. :)

itsthisjustin:
Can you pare this down to just the unblock thing as #3516 already adds the light to the home key?

naydichev:
> Can you pare this down to just the unblock thing as #3516 already adds the light to the home key?

done!

itsthisjustin (approved):

PR #3094, opened 2026-08-17
fix: block cursor mangles non-ASCII characters in cursor mode

winst0niuss:
## Symptom

In cursor mode, the block cursor over any non-ASCII character is drawn one glyph too narrow and the character underneath is not rendered inside the block. This affects every layout that has accented letters — `é` on AzertyFr, `ü`/`ß` on QwertzDe, `ñ` on SpanishEs — and any text loaded from outside (OPDS server names, download folder names).

**Before** — French UI, server name `Café`, cursor on `é`:

before

**After** — same screen, same keypresses:

after

The block should be as wide as the character and show it reversed out, which is exactly what it already does for ASCII.

## Steps to reproduce

No special hardware or data needed — this is reproducible on `develop` in the simulator:

1. Set the UI language to French (`Settings → System → Language`).
2. Open any text field, e.g. `Settings → System → OPDS Servers → <server> → Server Name`.
3. Type `é` (bottom letter row on the AZERTY layout).
4. Hold **Up** to enter cursor mode, then press **Left** until the cursor sits on the `é`.

## Cause

`cursorPos` is a byte offset that is always kept on a code point boundary, but four places in `render()` sliced exactly **one byte** at that offset:

| Code | Purpose |
|---|---|
| `text.substr(cursorPos, 1)` | block width |
| `displayText.substr(cursorPos, 1)` (×2) | kerning offset |
| `cursorPos + 1` | start of the trailing run |
| `char buf[2] = {text[cursorPos], '\0'}` | drawing the character in the block |

For a two-byte character that single byte is a lone lead byte. `utf8NextCodepoint()` (`lib/Utf8/Utf8.cpp`) validates continuation bytes, finds none, and returns `REPLACEMENT_GLYPH` (`0xFFFD`) — so the width is measured for the replacement glyph and the draw call paints that instead of the letter.

Byte-identical to the old behaviour for ASCII, where one byte *is* one code point.

## Fix

Uses the existing `utf8Next()` helper to take the whole code point in all four places. No new helper, no behaviour change for ASCII.

The masked-password path is unaffected: `displayText` is all `*` there, so the span is still one byte — that is why the fix computes the span separately for `text` and `displayText`.

## Testing

- Verified in the simulator on `develop` (bug) and on this branch (fixed) — the screenshots above are those two runs, same input script, only the branch differs.
- ASCII behaviour unchanged (checked with a Latin server name in the same screen).
- `clang-format` 21 clean.
- Not tested on hardware.

<details>
<summary>Full screenshots</summary>

Before:

before-full

After:

after-full

</details>

Split out of #2828 at @Uri-Tauber's request — this fix is independent of where the keyboard layout tables live.

Replaces #2840, which carried this fix and the line-wrapping one together; the wrapping half is now a separate PR.

### DIFF
--- a/src/activities/util/KeyboardEntryActivity.cpp
+++ b/src/activities/util/KeyboardEntryActivity.cpp
@@ -716,9 +716,20 @@ void KeyboardEntryActivity::render(RenderLock&&) {
   const int maxLineWidth = textAreaWidth;
   const bool centerText = metrics.keyboardCenteredText;
 
+  // The cursor spans a whole code point: a lone byte of it renders as a replacement glyph.
+  // Masking is per byte, so displayText keeps text's length and the same span applies to both.
+  const size_t cursorCharBytes = (cursorPos < text.length()) ? utf8Next(text, cursorPos) - cursorPos : 0;
+  char cursorChar[8] = {};         // the character under the cursor
+  char displayCursorChar[8] = {};  // same span of displayText, masked for passwords
+  if (cursorCharBytes > 0) {
+    const size_t n = std::min(cursorCharBytes, sizeof(cursorChar) - 1);
+    memcpy(cursorChar, text.data() + cursorPos, n);
+    memcpy(displayCursorChar, displayText.data() + cursorPos, n);
+  }
+
   int cursorCharWidth = 6;
-  if (cursorPos < text.length()) {
-    int w = renderer.getTextWidth(UI_12_FONT_ID, text.substr(cursorPos, 1).c_str());
+  if (cursorCharBytes > 0) {
+    int w = renderer.getTextWidth(UI_12_FONT_ID, cursorChar);
     if (w > cursorCharWidth) cursorCharWidth = w;
   }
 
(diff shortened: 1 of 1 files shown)


winst0niuss:
Both addressed.

**Masked passwords** — correct catch. Masking is per byte, so `displayText` always matches `text` byte-for-byte; `displayCursorCharBytes` was redundant as well as wrong, and is gone. This half was pre-existing, though — `develop` also advanced by exactly one byte there.

**Allocations** — the code point under the cursor is now a stack `char[8]`, so the diff adds no allocation over `develop`, and the kerning path drops one. `beforeCursor`, `part1` and `part3` are pre-existing and out of scope here; `string_view` could not replace them anyway, since `drawText`/`getTextWidth`/`getTextAdvanceX` all take `const char*`.

barbarhan (approved):
Reviewed the current head (`6d68207d`). The fix consistently derives the UTF-8 span once at `src/activities/util/KeyboardEntryActivity.cpp:719` and uses that same complete code point for block sizing, kerning, masked-password trailing-run placement, and the reversed glyph draw (`:730-733`, `:759-765`, `:785`, `:808-812`). ASCII remains a one-byte special case naturally, and the 8-byte stack buffers are comfortably bounded for valid UTF-8 without adding a render-loop heap allocation.

Verification:

- all 138 host tests pass on the PR head;
- clean merge rehearsal against current `develop`;
- all 138 host tests pass again after that merge rehearsal;
- CI, cppcheck, and all four hardware builds are green;
- structured and manual review found no remaining blocker.

The simulator before/after evidence in the PR covers the user-visible path. Approving.

lpla (approved):

PR #3095, opened 2026-08-17
fix: text-field line wrapping splits multi-byte characters

winst0niuss:
## Symptom

A text-field value long enough to wrap loses a character at every wrap point, if that point falls inside a multi-byte character. A Cyrillic OPDS server name shows it plainly — `литературы` renders as `литера` / `уры`, the `т` is simply gone:

line wrap

## Steps to reproduce

Reproducible on `develop` in the simulator, no special hardware or data:

1. Open a text field, e.g. `Settings → System → OPDS Servers → <server> → Server Name`.
2. Enter a non-ASCII value long enough to wrap onto a second line.
3. A character disappears wherever the wrap lands mid-sequence.

## Cause

`lineBreakEnd()` (`src/activities/util/KeyboardEntryActivity.cpp`) binary-searches over **byte** indices for the widest prefix that fits, so the returned wrap point can land inside a multi-byte character. `render()` then slices the line at that byte and hands both halves to the renderer as incomplete sequences. Each decodes to `U+FFFD`, and since no built-in font carries that glyph, the character is not drawn at all.

## Fix

Snap the wrap point back to a code point boundary. Because widths measured mid-sequence are unreliable, the snapped line is re-measured and stepped back whole characters until it fits, and a line always keeps at least one whole character so the wrap loop still advances.

ASCII is unaffected — the snap is a no-op there.

## Testing

- Verified in the simulator on `develop` (character lost) and on this branch (intact).
- Long Latin name still fills the line to the same edge as before.
- `clang-format` 21 clean.
- Not tested on hardware.

Found by @coderabbitai's review on #2840, which carried this fix together with the cursor-block one. That PR is now split: the cursor half is a separate PR, this is the wrapping half.

### DIFF
--- a/src/activities/util/KeyboardEntryActivity.cpp
+++ b/src/activities/util/KeyboardEntryActivity.cpp
@@ -386,7 +386,16 @@ int KeyboardEntryActivity::lineBreakEnd(std::string& s, const int start, const i
       hi = mid - 1;
     }
   }
-  return best;
+
+  // The byte-index search can stop inside a character; snap back to a boundary,
+  // keeping one whole character so the wrap loop always advances.
+  const int firstCharEnd = static_cast<int>(utf8Next(s, static_cast<size_t>(start)));
+  while (best > start && (static_cast<uint8_t>(s[best]) & 0xC0) == 0x80) best--;
+  // Widths measured mid-character are unreliable, so the search can overshoot.
+  while (best > firstCharEnd && measureRange(s, start, best) > maxWidth) {
+    best = static_cast<int>(utf8Prev(s, static_cast<size_t>(best)));
+  }
+  return best < firstCharEnd ? firstCharEnd : best;
 }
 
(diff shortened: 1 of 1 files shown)


barbarhan (approved):
Reviewed the current head (`b743d5ee`). `lineBreakEnd()` now normalizes the byte-index search result to a UTF-8 boundary at `src/activities/util/KeyboardEntryActivity.cpp:389-398`, re-measures after correcting an unreliable mid-sequence width, and backs up by whole code points when needed. The `firstCharEnd` floor preserves the existing guarantee that even an oversized first glyph advances the wrap loop. ASCII remains unchanged because every byte is already a boundary.

Verification:

- all 138 host tests pass on the PR head;
- clean merge rehearsal against current `develop`;
- all 138 host tests pass again after that merge rehearsal;
- CI, cppcheck, and all four hardware builds are green;
- structured and manual review found no remaining blocker.

The simulator evidence covers the reported Cyrillic failure and the Latin regression check. Approving.

lpla (approved):

PR #3096, opened 2026-08-17
feat(fonts): use Ubuntu Medium as the UI text weight

fain182:
## Summary

* **What is the goal of this PR?** Follow-up to #3079, addressing the "too thin" concern directly: give the UI the weight it has today, but as a designed weight rather than a rasterisation artefact.
  * `Ubuntu-Medium.ttf` (official 0.83 release) as the UI text source, stripped of `DSIG`, `LTSH` and the legacy `kern` table to match the Regular and Bold sources already in tree. Stripping `kern` is required: `fontconvert.py` sums the legacy and GPOS kern tables, which carry identical values here, so keeping both would double every pair.
  * `Ubuntu-Vietnamese-Medium.ttf`, rebuilt from the existing cut by the new `reweight-vietnamese-cut.py`, which swaps in the target weight's base letters and accents and re-centres the marks. The hook-above mark, which Ubuntu does not ship, is carried over unchanged. Regenerating cut and headers is byte-reproducible.
  * `ubuntu_1x_regular.h` → `ubuntu_1x_medium.h`, with `all.h`, `main.cpp` and `build-font-ids.sh` updated.

Measured at the sizes the UI uses: Medium with `--mono` gives 3px stems at both 10pt and 12pt — the same width the current threshold rendering produces — at +11.2% / +2.0% ink against the outline instead of +34.0% / +28.3%.

Noto Sans Hebrew and Arabic cover scripts Ubuntu does not and ship only Regular and Bold, so the Medium slot keeps their Regular.

| Before — Regular + threshold (current) | After — Medium + `--mono` |
| --- | --- |
| IMG_9395 | IMG_9405 |
| IMG_9396 | IMG_9406 |
| IMG_9397 | IMG_9407 |

## Scope Check

- [x] If this PR touches `freeink-sdk/`, `lib/hal/`, the bootloader, OTA, or recovery code, I have coordinated with the relevant maintainer. *(N/A — none are touched.)*

## Additional Context

* Stacked on #3079: this branch carries that commit too, so it should land after it.
* Flash: +3.5 KB image. RAM unchanged. Verified building and running on X4 hardware.
* `src/fontIds.h`: the two UI IDs are updated for the new header contents. Two things worth knowing — `build-font-ids.sh` does not emit the `static_assert` block the committed file carries, so regenerating it blindly drops those assertions; and `SMALL_FONT_ID` on `develop` already does not match the hash of `notosans_8_regular.h`, which predates this PR and is left alone.

---

### AI Usage


theZiz:
Hard to distinguish on your screenshots, it somehow feels better, but hard to point with the finger on it.

So I flashed your changes and made some more detailed photos:

Before / After

imageimage

I really like the difference! :smiley:

uxjulia:
Oh yeah, this is much better. Ubuntu's personality definitely shows through more now and the weight aligns more with what we're used to. I dig it.

Edit: just want to add that I tested this on device and it looks great.

uxjulia (approved):
Approving from a UI standpoint. I'll let @Uri-Tauber @itsthisjustin weigh in

lukestein (approved):
Testing on-device side by side with two x4's (one with this PR, one without), I definitely think UI rendering appearance is improved.

PR #3113, opened 2026-08-18
fix: re-check section under the render lock before starting the deferred partial-extension build

hooligan333:
## Summary

* **What is the goal of this PR?** Fix a latent crash in the reader's main loop: two sites in
  `EpubReaderActivity::loop()` check `section` *before* taking the `RenderLock`, then dereference it *after* acquiring
  the lock without re-checking. The peek and the acquire are not atomic — the render task (which runs on its own
  FreeRTOS task and takes the same lock) can win the race and finalize the partial extension, replace the section, or
  reset it to null on its page-load-failure paths (`renderBook()`'s build-start/persist failures). After that, the
  main loop dereferences a stale or null `section` pointer.
* **What changes are included?** One file, `src/activities/reader/EpubReaderActivity.cpp`:
  1. **Deferred partial-extension start**: the outer guard evaluates only lock-free state (the `!RenderLock::peek()`
     no-wait fast gate, `buildViewportWidth`, `partialRebuildStartFailed`); the `RenderLock` is then acquired and
     **every `section` state read — null check, `isBuilding()`, `isPartial()`, watermark — happens under the lock**
     before `section->startBuild()`.
  2. **The build pump below it**: same structure — heap fast gate outside, lock acquired, all `section` eligibility
     (null, `isBuilding()`, the partial/window condition) and the heap re-check evaluated under the lock.

  *(Revision history: the first version kept the unlocked eligibility peeks and added locked re-checks; review
  correctly pointed out the peeks themselves dereference `section` unsynchronised — a concurrent render-task reset
  frees the Section mid-expression. The second revision moved all `section` reads under the lock. The third makes
  both acquisitions non-blocking (`RenderLock::TryAcquire`): the restructured outer gates pass on nearly every
  loop pass, and a blocking acquire that lost the peek→acquire race would park the loop task behind an entire
  page render, stalling input polling — deferrable work now skips the contended pass and retries.)*

No behavior change on any healthy path: the re-checks pass whenever the unlocked peeks' state is still true under the
lock, and the build proceeds exactly as before. The only paths that change are the ones that previously crashed.

## Scope Check

Please confirm:

      (or, if one does, I explain why CrossPoint still needs it below). — It's a bug fix to CrossPoint's own lazy
      partial-extension mechanism.

## Additional Context

* **How the race is reached**: `loop()` runs on the main task; rendering runs on the ActivityManager render task
  (pinned to core 1 on multi-core boards). The unlocked peeks read `!RenderLock::peek()` at one instant; a render
  notified earlier can acquire the lock in the window between that peek and the main task's own blocking acquire.
  If that render hits a failure path (`renderBook()` resets `section` when `startBuild`/`createSectionFile` fail) or
  replaces/finalizes the section, the main task then operates on freed/null state.
* **Why the fix is shaped this way**: the build pump below the fixed site already re-checks under its lock — the fix
  extends the same discipline to the partial-extension start (which had no re-check at all) and completes the pump's
  re-check (which was missing only the null test).
* **cppcheck**: the re-checks are flagged `knownConditionTrueFalse` because cppcheck can't see the cross-task mutation;
  suppressed inline, following the existing convention.
* **Memory/flash impact**: none. No new allocations; the change adds two conditions and re-indents a block.
* **Build verification**: `pio run -e x4pro` and `pio run -e default` both pass. `pio check -e x4pro
  --fail-on-defect low` passes (no defects). `./bin/clang-format-fix -g` run, no reflow beyond the change.
* **Hardware verification**: tested on a physical X4 Pro (flashed to the OTA slot): a large partial-cached book read
  past the watermark — the deferred extension resumed normally (serial debug confirmed the fixed site fired); page
  turns during an in-flight background build and chapter jumps all behave as before. No crashes, no regressions.

Found by adversarial review during background-build work on an X4 Pro branch; the bug itself is unconditional and
affects all boards — no flags, no board-specific code.

---

### AI Usage

Did you use AI tools to help write this code? **YES** — the race was found by an AI adversarial review pass and the
fix was written with Claude Code; I reviewed and tested it on real hardware.

### DIFF
--- a/src/activities/ActivityManager.cpp
+++ b/src/activities/ActivityManager.cpp
@@ -361,6 +361,8 @@ RenderLock::RenderLock() {
   isLocked = true;
 }
 
+RenderLock::RenderLock(TryAcquire) { isLocked = xSemaphoreTake(activityManager.renderingMutex, 0) == pdTRUE; }
+
 RenderLock::RenderLock([[maybe_unused]] Activity&) {
   xSemaphoreTake(activityManager.renderingMutex, portMAX_DELAY);
   isLocked = true;
--- a/src/activities/RenderLock.h
+++ b/src/activities/RenderLock.h
@@ -9,6 +9,14 @@ class RenderLock {
  public:
   explicit RenderLock();
   explicit RenderLock(Activity&);  // unused for now, but keep for compatibility
+  // Non-blocking acquisition for deferrable work. The loop task's background
+  // chores (deferred build starts, the build pump) must not park on the
+  // rendering mutex: losing the peek->acquire race would block the loop behind
+  // an entire page render, stalling input polling for its duration. Check
+  // locked() before touching guarded state; on false, retry the next pass.
+  struct TryAcquire {};
+  explicit RenderLock(TryAcquire);
+  bool locked() const { return isLocked; }
   RenderLock(const RenderLock&) = delete;
   RenderLock& operator=(const RenderLock&) = delete;
   ~RenderLock();
--- a/src/activities/reader/EpubReaderActivity.cpp
+++ b/src/activities/reader/EpubReaderActivity.cpp
@@ -322,11 +322,13 @@ void EpubReaderActivity::loop() {
   // section is owned by the RenderLock: the render task resets or replaces it on its
   // failure/load paths while holding the lock, so even eligibility peeks must not
   // dereference it unlocked -- a concurrent reset would leave this task reading a freed
-  // Section mid-expression. Only lock-free state is checked outside; RenderLock::peek()
-  // is a no-wait fast gate that skips the acquire while a render is in flight.
+  // Section mid-expression. Only lock-free state is checked outside. The acquire is
+  // NON-BLOCKING: this outer gate passes on virtually every pass, and a blocking
+  // acquire that loses the peek->acquire race would park the loop task behind a whole
+  // page render, stalling input polling. Deferrable work retries next pass instead.
   if (!RenderLock::peek() && buildViewportWidth > 0 && !partialRebuildStartFailed) {
-    RenderLock lock;
-    if (section && !section->isBuilding() && section->isPartial() &&
+    RenderLock lock{RenderLock::TryAcquire{}};
+    if (lock.locked() && section && !section->isBuilding() && section->isPartial() &&
         section->currentPage + PARTIAL_REBUILD_START_MARGIN >= static_cast<int>(section->pageCount)) {
       const ReaderRenderSpec buildSpec = SETTINGS.readerRenderSpec(buildViewportWidth, buildViewportHeight);
       if (!section->startBuild(buildSpec)) {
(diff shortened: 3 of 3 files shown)


barbarhan (changes requested):
The under-lock rechecks are necessary, but they do not fully close the race described by this PR. Both outer guards still read and dereference `section` before taking `RenderLock`:

- the deferred-start guard reads `section`, `isBuilding()`, `isPartial()`, `currentPage`, and `pageCount` before `RenderLock::peek()`;
- the build-pump guard similarly reads `section` state before acquiring the lock.

`renderBook()` runs on the render task while holding the same lock and can reset or replace `section` on its failure/load paths. It can therefore reset the `unique_ptr` during one of those outer expressions, leaving the main task dereferencing the freed `Section` before it ever reaches the new locked recheck. `peek()` also does not reserve the mutex, so moving it earlier would not make unlocked `section` reads safe.

Please leave only non-`section` eligibility checks (and the optional `!RenderLock::peek()` fast gate) outside, acquire `RenderLock`, and move **all** `section` state checks into the locked block at both sites. That keeps the healthy-path behavior and the existing no-wait fast gate while actually removing the unsynchronised pointer/object reads this PR is meant to fix.

Verification on the current head: all 138 host tests pass; repository CI and all hardware builds are green. This remaining issue is specifically the concurrency safety boundary and is not exercised by those tests.

hooligan333:
@barbarhan Thank you — you're right, and the framing helped: the locked re-checks closed the TOCTOU window after the acquire but left the same unsynchronised-read class alive one step earlier, inside the guard expressions themselves.

Revised in eb9b0efc:

- Both outer guards now evaluate **only lock-free state**: the `!RenderLock::peek()` no-wait fast gate plus `buildViewportWidth` / `partialRebuildStartFailed` at the deferred-start site, and the heap gate at the pump. No `section` reads outside the lock.
- **Every `section` dereference happens under `RenderLock`** at both sites; the pump's `isPartial() || pageCount < currentPage + BUILD_WINDOW_AHEAD` eligibility condition moved inside the lock with the rest.
- The `cppcheck-suppress knownConditionTrueFalse` annotations became unnecessary (no more duplicated outer/inner conditions) and are gone; `pio check --fail-on-defect low` is clean without them.

Cost note: the lock is now taken on loop ticks that previously early-outed on unlocked `section` state. The acquire is uncontended in those cases (`peek()` gates render-in-flight), so it's a take/release in the microseconds — no change in page-turn feel on hardware.

One adjacent observation, flagged for scope rather than silently widened: the idle-prewarm guard a few lines up in the same `loop()` (pre-existing on develop, untouched by this PR) peeks `section->isBuilding()` / `section->currentPage` unlocked in the same way. Happy to fold the same restructure into this PR or leave it for a follow-up — whichever you and the maintainers prefer.

Verification on the revision: `pio run` for `x4pro` and `default` both green, `pio check --fail-on-defect low` clean. A reading sanity pass on a physical X4 Pro (partial-extension resume, background build, page-turn feel) is queued; I'll note here if anything differs.

hooligan333:
One self-found refinement, pushed as e6ac9230 — surfaced by an adversarial review pass on a downstream integration of this PR's shape, and it traces back to the previous revision here:

**What was wrong:** moving all section checks under the lock (correct, per the review above) left the *outer* gates evaluating only near-constant state — so they pass on virtually every `loop()` pass, and the blocking `RenderLock` constructor ran 1–2× per tick. On develop's original code these guards took **zero** locks in steady-state reading (the unlocked `section` peeks — the bug — also happened to be the early-out). Losing the narrow peek→acquire race then parks the loop task behind an entire page render (up to seconds for grayscale), stalling input polling for its duration; a press+release inside that window is lost.

**The fix:** both sites are deferrable background work that retries next pass anyway, so they now use a non-blocking try-acquire (new `RenderLock::TryAcquire` constructor, +2 lines in `ActivityManager.cpp`) and skip the pass when the mutex is contended. The locking invariant from the previous revision is unchanged — every `section` read still happens under the lock — and the steady-state cost returns to zero contention, matching develop's original behavior on the healthy path.

Verification: `pio run` for `x4pro` and `default` green, `pio check --fail-on-defect low` clean on this revision. The same shape is getting extended on-device testing in the downstream branch; I'll note here if anything differs.

lpla:
@hooligan333 Thanks for explicitly calling out the adjacent idle-prewarm path. I’d prefer folding it into this PR.

The guard around `EpubReaderActivity.cpp:298-301` still dereferences `section` before acquiring `RenderLock`, while the render task can reset or replace it under that lock. That is the same ownership/race invariant this PR is fixing in the deferred-build paths, so leaving it behind would make the protection incomplete.

Please keep the same non-blocking/deferrable behavior when moving those checks under the lock. With that included, the scope would still be focused and internally consistent.

PR #3123, opened 2026-08-19
fix: match capitalized umlaut Ä/Ö/Ü and ẞ words in dictionary lookup

sorig:
## Summary

* **What is the goal of this PR?** German dictionary lookups fail for words whose only difference from the stored form is a capital umlaut or capital ẞ — e.g. sentence-initial *Überprüf*, *Ähnlich*, *Öfter* or all-caps *STRAẞE* report "not found" even though the lowercase forms are in the StarDict `.idx`/`.syn`. `Dictionary::cleanWord()` lowercases ASCII bytes only, and `StringUtils::asciiCaseCmp` compares non-ASCII bytes verbatim, so `Ü` (`C3 9C`) never matches `ü` (`C3 BC`).
* **What changes are included?** A new `germanUppercaseFolded()` helper (Ä/Ö/Ü → ä/ö/ü, `C3 84/96/9C → C3 A4/B6/BC`; capital ẞ `E1 BA 9E` → ß `C3 9F`) and one extra probe in `lookup()`: when the unfolded exact-match and synonym probes miss, the folded variant is probed against `.idx` and `.syn` before the stemmer runs. The fold is deliberately a **fallback probe, not a query rewrite**: capitalized headwords (*Überprüfung*, *Ärzte*, *Öl*) are stored with their uppercase umlaut bytes and must keep matching the unfolded query first — folding unconditionally would trade the sentence-initial fix for a miss on every umlaut-initial noun. Files, sort order, and the comparator are untouched.

## Scope Check

Please confirm:

- [x] The stock firmware does not already handle this well, **and** no other popular CrossPoint fork already does (this is a defect in the existing `.syn` lookup path added in #2696, not new functionality).
- [x] If this PR touches `freeink-sdk/`, `lib/hal/`, the bootloader, OTA, or recovery code, I have coordinated with the relevant maintainer. (It touches none of these.)

## Additional Context

**Reproduction.** With the reader.dict DE-EN StarDict dictionary (543,060 `.syn` entries), tap a sentence-initial umlaut word such as *Überprüf* or *Übrigens* in any German EPUB on current `develop`: "not found". The lowercase forms `überprüf`/`übrigens` resolve fine. All-caps words containing capital ẞ (U+1E9E), e.g. *STRAẞE* in a chapter heading, fail the same way.

**Why fallback-probe rather than folding in `cleanWord()`.** A first iteration folded the query unconditionally; that regressed every capitalized-umlaut headword (*Überprüfung*, *Übung*, *Ärzte*, *Äpfel*, …), because the folded query no longer byte-matched the stored headword and `asciiCaseCmp` cannot fold non-ASCII on the comparison side. Probing unfolded-first preserves the exact behavior of every previously-working lookup; the folded reprobe only adds hits. Cost on the reprobe path is one extra `locate()` (+ one `locateSynonym()` on miss) reusing the already-open session handles, and only for words that actually contain Ä/Ö/Ü/ẞ and missed the primary probes.

**Verification.**

* Offline simulation of the exact lookup path (same folding, same comparison, `.idx` + `.syn` bisect) against the real dictionary files, 39 cases, all passing: 15 previously-missing capitalized forms now resolve (*Überprüf* → überprüfen, *Änderte* → ändern, *STRAẞE* → Straße, *FUẞBALL* → Fußball, …); 22 controls return exactly the same headword as before the patch, including 10 capitalized-umlaut headwords (*Überprüfung* → Überprüfung, *Ärzte* → Arzt, *Öl* → Öl, …) that guard against the query-rewrite regression; 2 out-of-vocabulary words still miss.
* `germanUppercaseFolded()` exercised natively under ASan/UBSan, including the 3-byte→2-byte ẞ shrink, lone Ä/ẞ, empty input, no-op inputs, and truncated UTF-8 sequences at end of string (no out-of-bounds read).
* Built from this branch and tested on an X4: previously-failing words now resolve in the reader. A small test EPUB with the fixed/control/miss word sets is available for QA.

**Memory / flash impact.** Negligible. One extra `std::string` allocation only on the reprobe path (folded variant of an already-cleaned word, typically <32 bytes); flash delta is a few dozen bytes of code. No new file handles — the reprobe reuses the open `LookupSession`.

**Risk / areas to focus on.** The fold is limited to the four German uppercase codepoints and applied only to fallback queries, because `.idx`/`.syn` are sorted under ASCII-only case folding with non-ASCII bytes compared verbatim — folding anything else (or folding inside the comparator) would break the binary search's sort-order assumption. Reviewers may want to double-check the probe-order reasoning in the comments on `germanUppercaseFolded()` and in `lookup()`.

---

### AI Usage

Did you use AI tools to help write this code? **YES** — the fix and its verification harness were written with an AI assistant (Claude); I reviewed the change and tested it on-device.

### DIFF
--- a/src/util/Dictionary.cpp
+++ b/src/util/Dictionary.cpp
@@ -597,38 +597,44 @@ std::string Dictionary::cleanWord(const char* word) {
   return result;
 }
 
-std::string Dictionary::germanUppercaseFolded(const std::string& word) {
+bool Dictionary::germanUppercaseFolded(const char* word, char* out, size_t outSize) {
   // Fold the German uppercase codepoints asciiCaseCmp cannot: Ä/Ö/Ü
   // (C3 84/96/9C -> C3 A4/B6/BC) and capital ẞ (E1 BA 9E -> C3 9F, one byte
-  // shorter). Returns "" when nothing folded so lookup() skips the reprobe.
+  // shorter). False when nothing folded — lookup() skips the reprobe — or when
+  // the result would not fit, which no matchable query can hit: readWordInto()
+  // caps headwords below the same size, so an over-long query has no match.
   // A fallback probe rather than part of cleanWord(): capitalized headwords
   // ("Überprüfung", "Ärzte") are stored with their uppercase umlaut bytes and
   // must keep matching an unfolded query first — folding unconditionally would
   // trade the sentence-initial fix for a miss on every umlaut-initial noun.
-  std::string folded;
+  const auto* b = reinterpret_cast<const unsigned char*>(word);
+  size_t w = 0;
   bool changed = false;
-  folded.reserve(word.size());
-  for (size_t i = 0; i < word.size();) {
-    const auto c = static_cast<unsigned char>(word[i]);
-    const auto c1 = i + 1 < word.size() ? static_cast<unsigned char>(word[i + 1]) : 0;
-    if (c == 0xC3 && (c1 == 0x84 || c1 == 0x96 || c1 == 0x9C)) {  // Ä Ö Ü
-      folded += static_cast<char>(0xC3);
-      folded += static_cast<char>(c1 + 0x20);  // -> ä ö ü
+  for (size_t i = 0; b[i] != '\0';) {
+    // b[i] != 0 makes b[i + 1] readable; the ẞ arm reads b[i + 2] only after
+    // b[i + 1] == 0xBA rules out the terminator.
+    unsigned char first = b[i];
+    unsigned char second = 0;
+    size_t consumed = 1;
+    if (b[i] == 0xC3 && (b[i + 1] == 0x84 || b[i + 1] == 0x96 || b[i + 1] == 0x9C)) {  // Ä Ö Ü
+      first = 0xC3;
+      second = b[i + 1] + 0x20;  // -> ä ö ü
+      consumed = 2;
       changed = true;
-      i += 2;
-    } else if (c == 0xE1 && c1 == 0xBA && i + 2 < word.size() &&
-               static_cast<unsigned char>(word[i + 2]) == 0x9E) {  // ẞ
-      folded += static_cast<char>(0xC3);
-      folded += static_cast<char>(0x9F);  // -> ß
+    } else if (b[i] == 0xE1 && b[i + 1] == 0xBA && b[i + 2] == 0x9E) {  // ẞ
+      first = 0xC3;
+      second = 0x9F;  // -> ß
+      consumed = 3;
       changed = true;
-      i += 3;
-    } else {
-      folded += static_cast<char>(c);
-      i++;
     }
+    const size_t emit = second != 0 ? 2 : 1;
+    if (w + emit >= outSize) return false;  // keep room for the terminator
+    out[w++] = static_cast<char>(first);
+    if (second != 0) out[w++] = static_cast<char>(second);
+    i += consumed;
   }
-  if (!changed) folded.clear();
-  return folded;
+  out[w] = '\0';
+  return changed;
 }
 
 void Dictionary::stemVariants(const std::string& word, std::vector<std::string>& out) {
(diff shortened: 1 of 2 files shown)


sorig:
[umlaut-folding-testcases_2.epub.zip](https://github.com/user-attachments/files/31223635/umlaut-folding-testcases_2.epub.zip)
 Here's an epub with test cases that can be used to verify the fix

src/util/Dictionary.cpp
```diff
@@ -597,6 +597,40 @@ std::string Dictionary::cleanWord(const char* word) {
...
+  // Fold the German uppercase codepoints asciiCaseCmp cannot: Ä/Ö/Ü
+  // (C3 84/96/9C -> C3 A4/B6/BC) and capital ẞ (E1 BA 9E -> C3 9F, one byte
+  // shorter). Returns "" when nothing folded so lookup() skips the reprobe.
+  // A fallback probe rather than part of cleanWord(): capitalized headwords
+  // ("Überprüfung", "Ärzte") are stored with their uppercase umlaut bytes and
+  // must keep matching an unfolded query first — folding unconditionally would
+  // trade the sentence-initial fix for a miss on every umlaut-initial noun.
+  std::string folded;
+  bool changed = false;
+  folded.reserve(word.size());
```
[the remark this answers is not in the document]
sorig: Good catch — fixed in 862964b9. germanUppercaseFolded() now writes into a reusable fixed foldBuf owned by Dictionary (separate from wordBuf, which locate() clobbers mid-scan); an oversized folded query returns false, which is always a legitimate skip since readWordInto() caps headwords at the same length. Note lookup() still allocates in cleanWord()/stemVariants() — pre-existing and untouched here.

src/util/Dictionary.cpp
```diff
@@ -597,6 +597,46 @@ std::string Dictionary::cleanWord(const char* word) {
   return result;
 }
 
+bool Dictionary::germanUppercaseFolded(const char* word, char* out, size_t outSize) {
```
unnamedd, line 600: @sorig personally speaking, I do think it is very dangerous if we start to implement in the code base, this kind of method. It is strictly focused in the german language. We have too many languages in the world, with too many variables that would be in need of a special support like the one you brought here. 
If we decide to cover them all, the implementation here would become a mess.

I really think we need to find a different approach.

@itsthisjustin @Uri-Tauber may I have a second opinion here, please?

src/util/Dictionary.cpp
```diff
@@ -597,6 +597,46 @@ std::string Dictionary::cleanWord(const char* word) {
...
+  const auto* b = reinterpret_cast<const unsigned char*>(word);
+  size_t w = 0;
+  bool changed = false;
+  for (size_t i = 0; b[i] != '\0';) {
+    // b[i] != 0 makes b[i + 1] readable; the ẞ arm reads b[i + 2] only after
+    // b[i + 1] == 0xBA rules out the terminator.
+    unsigned char first = b[i];
+    unsigned char second = 0;
+    size_t consumed = 1;
+    if (b[i] == 0xC3 && (b[i + 1] == 0x84 || b[i + 1] == 0x96 || b[i + 1] == 0x9C)) {  // Ä Ö Ü
```
unnamedd, line 619: Perhaps we should have one source of truth where all the special characters will come from.

unnamedd (changes requested):
Hi @sorig 👋 , 
first of all, thanks for your PR.

I had problems to test your PR due to the lack dictionary you mentioned in the PR description. Perhaps, next time you use an extra resource, you could share with us for the tests during the review, like you did with the book (really appreciated).

I do have a concern regarding your implementation here, and I prefer to have a second opinion before we move on, okay?

src/util/Dictionary.cpp
itsthisjustin: Probably should include language files like we do the hyphenation or localization strings?

Uri-Tauber: @unnamedd Completely agree. This implementation is too specific. Probably would break something down the road.
Not to mention, `src/util/Dictionary.cpp` is very bad location to have this. If we want such a thing, it should live in UFT8 lib or something.

sorig:
Hey @unnamedd please see the separate comment I left for the epub you can test with :)

sorig:
> [umlaut-folding-testcases_2.epub.zip](https://github.com/user-attachments/files/31223635/umlaut-folding-testcases_2.epub.zip)
> 
>  Here's an epub with test cases that can be used to verify the fix
> 
> 

@unnamedd

src/util/Dictionary.cpp
sorig: Thanks for your input! Will try implementing something more general

unnamedd:
Hi @sorig, that epub I saw, and I comment about it in my original comment. What I meant was the dictionary you used and mentioned in the PR's description.

I don't have it, and would be awesome to have the same tools as you for a proper testing here.
Sorry, but I try to be a very boring guy when I am reviewing PRs 😁

arthurgiroux (commented):
A better way to implement this would be to implement generic UTF-8 case folding.
The unicode consortium provides a mapping that could be used: https://www.unicode.org/Public/UCD/latest/ucd/CaseFolding.txt

arthurgiroux (changes requested):

PR #3134, opened 2026-08-20
fix(epub): ignore trailing non-XML data after </html> tag in ChapterHtmlSlimParser

bjesus:
Fixes an issue where EPUB files containing trailing padding bytes or non-XML data after the closing `</html>` tag cause background incremental section builds to fail (`Parse error during incremental build`), resetting the user's reading position back to page 0 / chapter start.

### Problem
Certain EPUB files have padding bytes (such as `\x05\x05...`, `\x02\x02`, or `\x10\x10...`) appended after `</html>`. 
While standard web browsers and HTML parsers ignore bytes after `</html>`, Expat returns `XML_STATUS_ERROR` (`not well-formed (invalid token)`) when reading the final chunk. When this happens during background section compilation (`Section::buildSomeMore()`), the incomplete section cache is abandoned, causing the reader to reload the section and reset pagination to page 0.

### Fix
- Added `htmlEnded_` to `ChapterHtmlSlimParser` to track when `endElement()` processes `</html>`.
- In `ChapterHtmlSlimParser::parseStep()`, if Expat returns `XML_STATUS_ERROR` after `htmlEnded_` is already `true`, the error is treated as non-fatal trailing data (`ParseStatus::Done`) instead of failing with `ParseStatus::Error`.

### Testing
- Tested in the desktop simulator and verified on physical hardware (CrossPoint X4 Pro). EPUBs with trailing padding now parse cleanly to completion without resetting section pagination.

### DIFF
--- a/lib/Epub/Epub/parsers/ChapterHtmlSlimParser.cpp
+++ b/lib/Epub/Epub/parsers/ChapterHtmlSlimParser.cpp
@@ -1542,11 +1542,15 @@ void XMLCALL ChapterHtmlSlimParser::endElement(void* userData, const XML_Char* n
   if (strcmp(name, "body") == 0) {
     self->insideBody = false;
   }
+  if (strcmp(name, "html") == 0) {
+    self->htmlEnded_ = true;
+  }
 }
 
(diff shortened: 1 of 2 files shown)


Uri-Tauber (approved):
Nice one!

PR #2610, opened 2026-07-14
feat: add Amiri font for Arabic script support

rxmmah:
## Summary
* **What is the goal of this PR?**
  Adds Amiri as a bundled font for Arabic-script reading support.

* **What changes are included?**
  Added an `Amiri` entry to `lib/EpdFont/scripts/sd-fonts.yaml`, sourced from the `google/fonts` OFL mirror (the official `aliftype/amiri` release is zip-only and doesn't fit the manifest's per-file URL pattern).

## Additional Context
OFL-licensed.

---
### AI Usage
Did you use AI tools to help write this code? **NO**

Issue #2809, opened 2026-07-30
Add OpenType shaping support for Bengali (Indic scripts)


mdmahmud1:
### Scope Self-Check (required)

- [x] I have read SCOPE.md and ROADMAP.md.
- [x] This is **not** a new theme or theming change (themes are temporarily closed pending the move to SD-loaded themes).
- [x] This is **not** a new external network connector (sync engine, cloud storage, remote file access, OPDS extensions beyond what exists, or any new "talk to a server" feature).
- [x] This is **not** an interactive app (game, calculator, notepad), writing/authoring tool, RSS/news/browser feature, media playback feature, or PDF rendering.
- [x] The stock firmware does **not** already handle this well.
- [x] No other popular CrossPoint fork already handles this well (or, if one does, I explain below why CrossPoint still needs it).

### Problem this solves

Thank you for the recent Font Builder patch.

The patch successfully fixes Bengali glyph extraction. Bengali .cpfont files now install correctly, and individual Bengali characters render on the X4.

The remaining issue is OpenType shaping.

## Current behavior

Unicode Bengali EPUBs display individual characters correctly, but conjuncts (যুক্তাক্ষর) and vowel positioning are not shaped correctly.

Examples:

- প্র

- ব্র

- ক্র

- গ্র

- শ্র

- ক্ষ

- জ্ঞ

- ত্র

Words such as:

- বাংলা

- শিক্ষা

- প্রজ্ঞা

- শ্রদ্ধা

- স্বাধীনতা

do not render correctly because they require OpenType shaping.

## Expected behavior

CrossPoint Reader should perform Unicode/OpenType shaping for Bengali (and other Indic scripts), including:

- GSUB substitutions

- GPOS positioning

- conjunct formation

- vowel reordering and positioning

- ZWJ/ZWNJ handling

Using HarfBuzz or another OpenType shaping engine would solve this.

## What has already been verified

✔ Font Builder now includes Bengali glyphs.

✔ .cpfont files install correctly.

✔ Unicode Bengali EPUBs open successfully.

✔ Individual Bengali letters render correctly.

The remaining missing feature is OpenType shaping.

I'd be happy to provide a Bengali regression test EPUB containing:

- vowels

- consonants

- vowel signs

- common conjuncts

- numerals

- punctuation

- real text passages

to help validate future releases.

Thank you again for your excellent support.

### Why the stock firmware (and other forks) do not already solve this

[...]

PR #3146, opened 2026-08-21
feat: language-specific fonts

Uri-Tauber:
## Summary

With the new script-grouping, we can add more fonts without harming the user UX.
I will start this with three Hebrew fonts + Arabic Amiri from #2610.
Ultimately, I want 3-4 fonts for every language which the built-in reader fonts doesn't cover.
The branch would be merged before release 1.6.0

A few rules for contributing:
1. Direct your PR to `chore/fonts`, not `develop`.
2. **Suggest fonts only for your own native language**. Typography is something non-speakers find much harder to judge.
3. Open source fonts with permissive license. 

A reminder: CrossPoint doesn't support Indic scripts.

---

### AI Usage

### DIFF
--- a/lib/EpdFont/scripts/sd-fonts.yaml
+++ b/lib/EpdFont/scripts/sd-fonts.yaml
@@ -286,3 +286,45 @@ families:
       bold:       {url: "https://raw.githubusercontent.com/jacobxperez/lexica-ultralegible/main/fonts/ttf/LexicaUltralegible-Bold.ttf"}
       italic:     {url: "https://raw.githubusercontent.com/jacobxperez/lexica-ultralegible/main/fonts/ttf/LexicaUltralegible-Italic.ttf"}
       bolditalic: {url: "https://raw.githubusercontent.com/jacobxperez/lexica-ultralegible/main/fonts/ttf/LexicaUltralegible-BoldItalic.ttf"}
+
+  # ── Hebrew ─────────────────────────────────────────────────────────────
+
+  - name: FrankRuhlLibre
+    description: "Hebrew serif (Hebrew, Latin)"
+    scripts: [hebrew, latin]
+    intervals: hebrew,latin-ext,ipa-chars,punctuation,reading,symbols
+    sizes: [12, 14, 16, 18]
+    styles:
+      regular: {url: "https://raw.githubusercontent.com/google/fonts/main/ofl/frankruhllibre/FrankRuhlLibre%5Bwght%5D.ttf", variable: {wght: 400}}
+      bold:    {url: "https://raw.githubusercontent.com/google/fonts/main/ofl/frankruhllibre/FrankRuhlLibre%5Bwght%5D.ttf", variable: {wght: 700}}
+
+  - name: DavidLibre
+    description: "Traditional Hebrew serif (Hebrew, Latin)"
+    scripts: [hebrew, latin]
+    intervals: hebrew,latin-ext,ipa-chars,punctuation,reading,symbols
+    sizes: [12, 14, 16, 18]
+    styles:
+      regular: {url: "https://raw.githubusercontent.com/google/fonts/main/ofl/davidlibre/DavidLibre-Regular.ttf"}
+      bold:    {url: "https://raw.githubusercontent.com/google/fonts/main/ofl/davidlibre/DavidLibre-Bold.ttf"}
+
+  - name: Alef
+    description: "Simple, legible Hebrew sans-serif (Hebrew, Latin)"
+    scripts: [hebrew, latin]
+    intervals: hebrew,latin-ext,ipa-chars,punctuation,reading,symbols
+    sizes: [12, 14, 16, 18]
+    styles:
+      regular: {url: "https://raw.githubusercontent.com/google/fonts/main/ofl/alef/Alef-Regular.ttf"}
+      bold:    {url: "https://raw.githubusercontent.com/google/fonts/main/ofl/alef/Alef-Bold.ttf"}
+
+  # ── Arabic ─────────────────────────────────────────────────────────────
+
+  - name: Amiri
+    description: "Arabic naskh-style serif (Arabic, Latin)"
+    scripts: [arabic, latin]
+    intervals: arabic,latin-ext,ipa-chars
+    sizes: [12, 14, 16, 18]
+    styles:
+      regular:    {url: "https://raw.githubusercontent.com/google/fonts/main/ofl/amiri/Amiri-Regular.ttf"}
+      bold:       {url: "https://raw.githubusercontent.com/google/fonts/main/ofl/amiri/Amiri-Bold.ttf"}
+      italic:     {url: "https://raw.githubusercontent.com/google/fonts/main/ofl/amiri/Amiri-Italic.ttf"}
+      bolditalic: {url: "https://raw.githubusercontent.com/google/fonts/main/ofl/amiri/Amiri-BoldItalic.ttf"}
--- a/lib/EpdFont/scripts/sd-fonts.yaml
+++ b/lib/EpdFont/scripts/sd-fonts.yaml
@@ -328,3 +328,49 @@ families:
       bold:       {url: "https://raw.githubusercontent.com/google/fonts/main/ofl/amiri/Amiri-Bold.ttf"}
       italic:     {url: "https://raw.githubusercontent.com/google/fonts/main/ofl/amiri/Amiri-Italic.ttf"}
       bolditalic: {url: "https://raw.githubusercontent.com/google/fonts/main/ofl/amiri/Amiri-BoldItalic.ttf"}
+
+  # ── Korean ─────────────────────────────────────────────────────────────
+
+  - name: Pretendard
+    description: "Modern Korean sans-serif (Hangul, Latin)"
+    scripts: [hangul, latin]
+    intervals: hangul,latin-ext,punctuation,reading,symbols
+    sizes: [8, 10, 12, 14, 16, 18]
+    styles:
+      regular: {url: "https://raw.githubusercontent.com/orioncactus/pretendard/v1.3.9/packages/pretendard/dist/public/static/Pretendard-Regular.otf"}
+      bold:    {url: "https://raw.githubusercontent.com/orioncactus/pretendard/v1.3.9/packages/pretendard/dist/public/static/Pretendard-Bold.otf"}
+
+  - name: PretendardMedium
+    description: "Pretendard at Medium weight, slightly heavier body text (Hangul, Latin)"
+    scripts: [hangul, latin]
+    intervals: hangul,latin-ext,punctuation,reading,symbols
+    sizes: [8, 10, 12, 14, 16, 18]
+    styles:
+      regular: {url: "https://raw.githubusercontent.com/orioncactus/pretendard/v1.3.9/packages/pretendard/dist/public/static/Pretendard-Medium.otf"}
+      bold:    {url: "https://raw.githubusercontent.com/orioncactus/pretendard/v1.3.9/packages/pretendard/dist/public/static/Pretendard-ExtraBold.otf"}
+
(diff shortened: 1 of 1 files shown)


Uri-Tauber:
@rxmmah @brianhuster @serialx 
Tagging you as Arabic and CJK speakers.

Question: Should we have separate fonts for Chinese and Korean, or one bigger CJK font?
What do you use for your daily reading?

unnamedd:
@Uri-Tauber based on your description, perhaps it would be better if we created a topic in the discussion or used one of the existing ones for fonts to indicate where people could contribute more effectively, as you pointed out in `chore/fonts`.

Uri-Tauber:
@unnamedd Good idea.

Uri-Tauber:
# AI-generated guide on adding new font:

  ## 1. Add the family to the catalog

  Edit `lib/EpdFont/scripts/sd-fonts.yaml`. This file now owns both conversion inputs and Font Manager presentation.

```
  - name: MyFont
    description: "Readable serif with Latin and Hebrew coverage"
    scripts: [latin, hebrew]
    intervals: latin-ext,hebrew,ipa-chars
    sizes: [12, 14, 16, 18]
    styles:
      regular:    {url: "https://example.org/MyFont-Regular.ttf"}
      bold:       {url: "https://example.org/MyFont-Bold.ttf"}
      italic:     {url: "https://example.org/MyFont-Italic.ttf"}
      bolditalic: {url: "https://example.org/MyFont-BoldItalic.ttf"}
```

  The fields have distinct jobs:

  - **name:** becomes the UI family name and filename prefix, such as `MyFont_12.cpfont`.
  - **description:** shown beneath the family name.
  - **scripts:** controls which Font Manager groups contain the family. Every tag must exist in `scriptGroups` at `lib/EpdFont/scripts/sd-fonts.yaml:34`.
  - **intervals:** controls which glyphs are actually rasterized into the font.
  - **sizes:** controls generated `.cpfont` files.
  - **styles:** accepts a downloadable url; variable fonts may also specify variable axis values, as documented at `lib/EpdFont/scripts/sd-fonts.yaml:19`.

  Crucially, **scripts** and **intervals** are independent. Writing `scripts: [hebrew]` does not include Hebrew glyphs; **intervals** must include hebrew too.
  Conversely, adding Hebrew coverage does not automatically place the family under the Hebrew browser group.

  Prefer immutable, versioned source URLs. The current release workflow rebuilds every family, so floating main/master URLs can unexpectedly change existing font assets.

  For CJK UI fallback, include sizes 8, 10, 12 as well as reader sizes; the requirement is documented at `docs/sd-card-fonts.md:67`.

  ## 2. Add new groups or coverage only when necessary

  If the desired browser group already exists, just reference it in scripts.

  For a genuinely new group:
```
  scriptGroups:
    - {tag: devanagari, label: Devanagari}
```

  No firmware change is required for a new manifest-defined label, but firmware supports at most 32 groups because membership is represented by a 32-bit mask at `src/activities/settings/FontDownloadActivity.h:69`.

  If no appropriate glyph interval exists, add one to `INTERVAL_PRESETS` in `lib/EpdFont/scripts/fontconvert_sdcard.py:39` and document it. Adding the browser group alone does not add glyph coverage.

unnamedd (approved):

Uri-Tauber:
@unnamedd Dont merge it yet.
I want people to contribute more fonts before the release.

unnamedd:
> @unnamedd Dont merge it yet. I want people to contribute more fonts before the release.

Never planned! You know the best moment to merge it! ;)

brianhuster:
> Should we have separate fonts for Chinese and Korean, or one bigger CJK font?

Yes, I think there should be seperated fonts for:
- Simplified Chinese 
- Traditional Chinese (Taiwan, Macau)
- Traditional Chinese (Hong Kong)
- Japanese
- Korean

Like what Noto Project does

Because many Chinese characters despite sharing the same Unicode codepoint, have different glyphs depending on language on region. See https://en.wikipedia.org/wiki/Han_unification

chohoyeon-ops:
> > Should we have separate fonts for Chinese and Korean, or one bigger CJK font?
> 
> Yes, I think there should be seperated fonts for:
> 
> * Simplified Chinese
> * Traditional Chinese (Taiwan, Macau)
> * Traditional Chinese (Hong Kong)
> * Japanese
> * Korean
> 
> Like what Noto Project does
> 
> Because many Chinese characters despite sharing the same Unicode codepoint, have different glyphs depending on language on region. See https://en.wikipedia.org/wiki/Han_unification

I agree that separate regional CJK fonts make sense, especially because Han unification means the same Unicode code point can have different glyph forms.

But this also raises a question for CrossPoint: how does the renderer determine which regional glyph should be used for a given CJK character?

For example, if the same Unicode code point exists in Noto Sans CJK KR, JP, SC, TC and HK with different glyph forms, does CrossPoint use the language metadata of the EPUB/document to select the appropriate font, or is the user expected to select the regional font manually?

I think this distinction is particularly important for Korean readers, since a Korean book may contain a significant amount of Hanja while still requiring the Korean glyph forms.

Also, Korean users tend to use a fairly wide variety of fonts for reading, rather than sticking to a single "Korean font".

For example, KoPubWorld would probably be perfectly fine as a default/bundled font with broad CJK coverage. Then users who care about typography can simply provide their own preferred font.

So from a Korean user's perspective, I wouldn't worry too much about finding the one "correct" Korean CJK font. Good coverage + font choice/fallback is probably more useful.

Some Korean e-book users routinely install or swap fonts themselves, so having a reasonable bundled font and allowing custom fonts may be more useful than trying to optimize for one particular typeface.

rxmmah:
for arabic i have tested many fonts only three worked well:
Amiri
Noto Kufi Arabic
Noto Naskh Arabic UI

brianhuster:
> For example, if the same Unicode code point exists in Noto Sans CJK KR, JP, SC, TC and HK with different glyph forms, does CrossPoint use the language metadata of the EPUB/document to select the appropriate font, or is the user expected to select the regional font manually?

I don't think this PR needs to answer that question

> a Korean book may contain a significant amount of Hanja while still requiring the Korean glyph forms.

Crosspoint just need to use the Hanja glyphs found in the Korean font?

chohoyeon-ops:
> > For example, if the same Unicode code point exists in Noto Sans CJK KR, JP, SC, TC and HK with different glyph forms, does CrossPoint use the language metadata of the EPUB/document to select the appropriate font, or is the user expected to select the regional font manually?
> 
> I don't think this PR needs to answer that question
> 
> > a Korean book may contain a significant amount of Hanja while still requiring the Korean glyph forms.
> 
> Crosspoint just need to use the Hanja glyphs found in the Korean font?

I agree that the PR doesn't need to define the full regional glyph selection logic. My question was specifically about the architectural assumption behind bundling KR/JP/SC/TC/HK variants into one font: if CrossPoint doesn't already have a way to select a regional glyph from a unified CJK font, then that would be a separate implementation requirement rather than something the font itself solves..

For Korean, I actually prefer the simpler model: use a Korean font as the primary font, including its Hanja glyphs, and fall back to another font only when necessary. I wasn't suggesting that CrossPoint should dynamically choose between Korean/Japanese/Chinese glyph variants from one unified font.

brianhuster:
> about the architectural assumption behind bundling KR/JP/SC/TC/HK variants into one font

I don't remember anyone saying that. Uri-Tauber just asked a question, that's all

chohoyeon-ops:
> > about the architectural assumption behind bundling KR/JP/SC/TC/HK variants into one font
> 
> I don't remember anyone saying that. Uri-Tauber just asked a question, that's all

You're right. I may have phrased that too strongly. I wasn't assuming that a unified CJK font had been decided. I was only asking what the implications would be if that option were chosen, since the regional glyph selection would then become a separate rendering concern.

Personally, I would prefer separate regional fonts for Korean and Japanese/Chinese, with the Korean font providing its own Hanja glyphs and fallback for characters it doesn't cover.

Uri-Tauber:
@brianhuster @chohoyeon-ops sounds to me like the best course of action is to have seperate fonts, so we can keep this language identifying logic out of CrossPoint.

@rxmmah Thanks for the info, Amiri is already in. It would be a great help if you could open a PR for the other two.

serialx:
Added popular Korean fonts in #3177

gangusa:
Thanks for making language-specific font contributions possible.

I would like to contribute Kannada fonts, but I noticed the statement in this PR that **“CrossPoint doesn't support Indic scripts.”**

I am using an XTEINK X3 for reading Kannada EPUBs, and this is currently the main limitation.

Kannada requires proper Indic text shaping, not just Unicode glyph coverage. For example:

**ಕನ್ನಡ · ಕರ್ನಾಟಕ · ನಮಸ್ಕಾರ · ಶ್ರೀ · ಕ್ರ · ಕ್ಕ · ಸ್ತ್ರ · ಕ್ಷ · ಜ್ಞ**

These require conjunct formation, vowel-sign reordering, combining-mark positioning and proper Indic/OpenType shaping.

Would it be possible to consider **Indic script shaping support** as a future feature or roadmap item?
Fonts 
**We should choose fonts with licenses that allow redistribution, for example:

Noto Sans Kannada
Noto Serif Kannada
Tiro Kannada
Baloo Tamma 2**

Once Indic shaping is supported, I would be happy to contribute 3–4 high-quality Kannada fonts as suggested in this PR.

I can also help test Kannada rendering and provide native Kannada test strings and EPUB samples.

Thank you for the excellent work on CrossPoint!

brianhuster:
@gangusa https://github.com/crosspoint-reader/crosspoint-reader/issues/2809#issuecomment-5152646431

brianhuster:
@Uri-Tauber I tried building the Japanese font [BIZ UDMincho](https://fonts.google.com/specimen/BIZ+UDMincho) using the following command

```bash
~/.platformio/penv/bin/python lib/EpdFont/scripts/fontconvert_sdcard.py \
      --intervals cjk,latin-ext,ipa-chars \
      --sizes 8,10,12,14,16,18 \
      --name "BIZUDMincho" \
      --regular path/to/BIZUDMincho-Regular.ttf \
      --bold path/to/BIZUDMincho-Bold.ttf \
      --output-dir BIZUDMincho
```
However when I upload the files to Crosspoint and select that font, it seems invalid or something. The reader cannot display any Japanese characters, no, in fact it always automatically goes back to Noto Serif. Could you please check anything wrong with the cpfont I built [BIZUDMincho_cpfont.zip](https://github.com/user-attachments/files/31499147/BIZUDMincho.zip). Here is a Japanese epub in case you want to test [銀河鉄道の夜 - 宮沢賢治.epub](https://drive.google.com/file/d/1f6SkEeY19NKX5ie7JhfwMoMbXVOi3Zwc/view?usp=drive_link)

Uri-Tauber:
@brianhuster Sorry on the late reply. I missed your comment somehow.
I will look into this on Sunday.

brianhuster:
@Uri-Tauber I think I know what the cause of the error is. The font BIZ has too many intervals (as it only includes characters needed for Japanese language), more than the threshold 4096. We have several ways to solve this
- Increasing the threshold to 8192
- Use a Noto CJK font as fallback to fill the gaps between intervals

Not sure which is preferred

Uri-Tauber:
Hmm. 
I didnt noticed there was hard limit on number of intervals. I thought the glyphs count is the important cap.
I need to check why we have this limit before I could decide.

BTW, why Korean fonts managed okay?

brianhuster:
Here is the line with such limit https://github.com/crosspoint-reader/crosspoint-reader/blob/7db14a015ca4b1d74d3cbed131a748f3029a4517/lib/EpdFont/SdCardFont.cpp#L609

> why Korean fonts managed okay?

Because Korean fonts included here only contain Hangul (Korean alphabet) and no CJK ideographs (Chinese characters). Though there are more than 11,000 Hangul glyphs, they are only a few intervals

PR #3163, opened 2026-08-22
feat: load non-built-in UI languages from SD language packs

eszter007:
## Summary

* **What is the goal of this PR?** Make it *possible* to ship UI languages as SD packs instead of compiling all 32
  into every build. This PR doesn't change what CrossPoint itself ships. The default is `all`. The option to drop languages enables forks focused on other languages (e.g. CJK) to have smaller builds without overshooting the partition size.
* **What changes are included?** The pack format and loader, the generator support, and an opt-in switch that is off
  by default.

## Motivation
In my fork matcha-reader, which is focused on Japanese, I came close to the partition size as the entire UI is available in Japanese and CJK is built-in. When the SD card plugin code was merged in, the size came to 99% of the partition without much room for future features. After dropping all languages beside EN, ES, DE, FR and JA itself, the partition size was at 87%, which is significant.
This feature allows forks like mine to drop certain languages in favor of a smaller size. Upstream crosspoint could also benefit as less common languages could be also build on-demand, without increasing the package size for everyone.

## The configuraton

`platformio.ini` gains `custom_i18n_builtin_langs`, default `all`:

```ini
custom_i18n_builtin_langs = all          ; today's behaviour — everything compiled in
custom_i18n_builtin_langs = en,es,fr,de  ; example with English, Spanish, French, German compiled in, the other 28 ship as packs
```
Packs complement the list. English is always built-in as it serves as the fallback language.

## The loader

* Non-built-in languages are emitted by `scripts/gen_i18n.py` as `.cplang` packs for `/.crosspoint/lang/`.
* Languages without a pack are still listed in the picker, marked **Needs pack** (`STR_LANGUAGE_PACK_MISSING`) 
  so they are discoverable. Screenshot from a build with only EN included:
  
Bildschirmfoto 2026-08-22 um 11 42 48

* Packs are validated on load: magic, format version, and key count against `StrId::_COUNT`. Offsets index strings by
  key *position*, so a pack from a different build would return the wrong strings; it is refused instead.
* Docs: new §3.9 in `USER_GUIDE.md`, and a Language Packs section in `docs/i18n.md` with the on-disk format table.

## Scope Check

- [x] The stock firmware does not already handle this well — every language is currently compiled in unconditionally,
      with no way to opt out.
- [x] Does **not** touch `freeink-sdk/`, `lib/hal/`, the bootloader, OTA, or recovery code.

## Additional Context

Example:
**Flash / memory impact**, measured on `-e default` against `develop` at `69dd947f`:

| | flash | bytes | vs develop |
|---|---|---|---|
| `develop` | 87.0% | 5,703,985 | — |
| this branch, default `all` | 87.0% | 5,704,359 | **+374 B** |
| this branch, `en,es,fr,de` | 82.8% | 5,426,529 | −277,456 B |

**Verification.** `pio run -e default` builds clean at both settings; `gen_i18n.py` emits all 28 packs when a list is
set, and none at the default. Tested on a X4 on-device.

### AI Usage

Did you use AI tools to help write this code? _**YES**_ (Claude Code, Opus 5)


Uri-Tauber:
@eszter007 Nice one. 
Personally I would prefer a slimmer version: Just make `gen_i18n.py` to depend on a flag somewhere. Plus something similar for the built-in fonts. Keep the infrastructure out of the firmware itself.

Supporting language files from the SD is a path I dont want to walk right now.

PS. You could save further flash by removing `minibidi` library from your fork. It only relevant to Hebrew and Arabic. If you dont intend to support those,its a dead weight.

eszter007:
@Uri-Tauber Thank you, will take a look at this when I am back from my vacation and make a simpler version for upstream then :)

eszter007:
@Uri-Tauber I created a new PR with the slimmer version. Due to some plugin related bad commit message in my fork, I needed to rewrite my git history.

Issue #2512, opened 2026-07-02
Pulling a Koreader Sync sending me to incorrect percentage and position


maxemillianian:
### Affected Version

1.4.1

### Describe the Bug

When I push progress to Koreader Sync, then pull that same progress on the same device (X4), the book jumps to an incorrect position. On both of the books I've tried so far it jumps me to the beginning of the chapter.

When I pull the progress on my Kobo (using Koreader), the pull takes me to the correct percentage.

I tried 1.4.1, 1.4.0, and the nightly from July 1: the issue is consistent throughout the releases.

I tried syncing to a new Koreader Sync account, and I tried both matching by file name and by binary hash: the issue persisted. 

I then tried CrossInk (1.3.4), and the pull works fine. 

### Steps to Reproduce

1. Push progress to Koreader Sync
2. Pull that just-synced progress from Koreader Sync
3. Find oneself at an unexpected and incorrect position in the book

### Expected Behavior

I expected, after clicking 'pull from remote,' for the percentage I arrive at in the book to be the same percentage as is listed on 'pull from remote.'

### Relevant Log Output/Screenshots

```shell

```

MakuroDev:
I’m still seeing this behavior on 1.5.0

itsthisjustin:
@Uri-Tauber @uxjulia I definitely don't have this between crosspoint devices but i see some reports of this happening when sending off crosspoint. I am wondering if the recent change to how we do text layout changed something?

Uri-Tauber:
@thiagokokada Can you take a look?

thiagokokada:
I can try to investigate but I need more details, like the book, what the exactly behavior you're seeing, which devices are being synced, KOReader version, etc.

uxjulia:
@MakuroDev can you provide the book you're experiencing this with and the exact chapter you see this with

agosez:
I  encountered some books with XML comments - `<!--This is a comment-->` - that are treated as a single opening tag instead of a complete opening / closing element by the ProgressMapper and this leads to incorrect xpath resolution

WhoTheHeck:
good catch, @agosez. My linked PR only fixes the generation of precise outgoing KOSync positions. I didnt think of XML comments affecting incoming XPath resolution. I'll look into a dedicated fix.

PR #3174, opened 2026-08-22
fix(KOSync): preserve precise KOSync upload progress positions

WhoTheHeck:
## Summary

  Preserve the exact EPUB text position when CrossPoint uploads reading progress through KOSync.

  * Use the exact text position when creating the KOReader-compatible XPath.
  * Keep paragraph, percentage, and synthetic mappings as fallbacks.
  * Preserve text-node index, character offset, nested ancestry, offset zero, and UTF-8 codepoint handling.
  * Add focused host regression tests.

Refs #2512.

## Scope Check

- [x] This PR is **not** a new external network connector. It corrects position generation in the existing KOSync integration.
- [x] The stock firmware does not already handle this well, and no other reviewed PR fixes this exact path.

Not applicable: this is a focused correction to existing EPUB progress tracking.

## Additional Context

PR #1686 introduced paragraph-based upload mapping. PR #2805 later made the visible-text offset authoritative, but the outgoing conversion still preferred paragraph metadata whenever it was available.

This made the result cache-dependent. With the same firmware, an incomplete cache produced a detailed position such as:

`/body/DocFragment[48]/body/p[39]/text()[1].34`

After the cache completed, the upload could become:

`/body/DocFragment[48]/body/p[38]`

The latter identifies a paragraph but loses the exact position inside it.

This may explain the v1.5 report in #2512. The original v1.4 report predates PR #2805, so this PR references rather than closes that broader issue.

## Related Work

PR #2247 addresses EPUB chapters whose visible text is stored directly inside `<div>` or `<body>` elements. It appears complementary: #2247 expands which XHTML structures can be resolved, while this PR changes which available position source is preferred during upload.

## Verification

- Formatting and static analysis pass.
- The full host test suite passes: 144/144 tests.
- Compared with `upstream/develop`, the `default` firmware build:
  - Adds no RAM usage.
  - Adds 694 bytes of reported flash usage.
- The fix was compiled and verified on an X3.
- Device logs confirmed:
  - The outgoing position retained `/text()[N].offset`.
  - The upload received a 2xx response.
  - No reset, OOM, or panic occurred.
- The rebased patch was unchanged and compiled successfully again.

Kavita was used as the KOReader Sync server. The upload and subsequent server read-back were checked separately because Kavita may normalize the stored XPath; a successful HTTP response alone does not confirm the stored position.

---

### AI Usage


thiagokokada (approved):
Code-wise LGTM, also tested Crosspoint -> KOReader and KOReader -> CrossPoint.

unnamedd:
@WhoTheHeck may you please fix the conflict here?

WhoTheHeck:
@unnamedd done 👍

unnamedd (approved):
Thanks for the help @WhoTheHeck!

PR #3175, opened 2026-08-22
feat: Brings back features from the x4pro beta that everyone wanted

itsthisjustin:
1. Introduces REFRESH_NEVER (value 5) as a new e-ink refresh frequency option. 

2. Adds night mode everywhere 

3. Adds a swipe up from bottom gesture for the reader menu


Uri-Tauber:
Why night mode everywhere is only for frontlight devices?
Seems like a good idea for every device?

prokrypt:
> Why night mode everywhere is only for frontlight devices? Seems like a good idea for every device?

ghosting, probably.

Uri-Tauber:
@prokrypt Ghosting is a problem for frotlight devices as well.

prokrypt:
@Uri-Tauber I haven't had any [bad] ghosting issues on my x4 pro though. I was one of the people wanting a no-refresh option.

itsthisjustin:
> > Why night mode everywhere is only for frontlight devices? Seems like a good idea for every device?
> 
> ghosting, probably.

Yep and it's mostly a bigger pain for users with frontlights. Trying to mitigate a ton of github issues with folks reporting slight ghosting after turning it on everywhere

itsthisjustin:
> Why night mode everywhere is only for frontlight devices?
> Seems like a good idea for every device?

I'm open to you changing this but initially when added julia had balked at the ghosting on her standard x4. I really don't see "bad" ghosting on my end, especially on pro. It ghosts a bit in settings but it's not horrible

Uri-Tauber:
@itsthisjustin I'm a bit confused here. Does having Night Mode everywhere make more ghosting or less? Is it a bigger problem for x4pro?
I dont see any ghosting with the implementation as it is now on `develop`.

I just think it better be consistent. Otherwise, we would get complains from users with two devices, "why its not the same?"

itsthisjustin:
> @itsthisjustin I'm a bit confused here. Does having Night Mode everywhere make more ghosting or less? Is it a bigger problem for x4pro? I dont see any ghosting with the implementation as it is now on `develop`.
> 
> I just think it better be consistent. Otherwise, we would get complains from users with two devices, "why its not the same?"

The implemenation right now is reader only and doesn't cause ghosting because of how the reader works and drawing text in consistent areas on the screen. This adds dark mode in all the UI which if you're monkeying with settings a bunch will cause some ghosting until you refresh. 

Fwiw I did notice that stock firmware has an automatic refresh on a number of refreshes not a number of pages and is UI wide so it will even refresh in the middle of a progress bar updating if it hits that threshold

itsthisjustin:
I'm just gonna put it everywhere. we can rc it, see if people complain, and go from there

itsthisjustin:
Ready for review and merge @uri-tauber

Uri-Tauber (approved):
LGTM. 
At least from the firmware side. I have no idea what you did in the SDK, and Im afraid to ask :smile:

itsthisjustin:
> LGTM. At least from the firmware side. I have no idea what you did in the SDK, and Im afraid to ask 😄

Ha it just bumps some stuff we aren't even using yet, so it's harmless. USB mode stuff, a credentials error on sd-plugins. I like to try to keep sdk up to date within PRs to avoid regressions between them. It also adds all the controls needed for the frontlight panel in that other pr. Prevents forks from pulling older sdk commits too where they might break something for users

PR #3177, opened 2026-08-23
feat: add Korean font families

serialx:
## Summary

Adds Korean coverage to the SD font catalog, following the contribution rules in #3146 (native language, permissive license, targeted at `chore/fonts`). All fonts are SIL OFL 1.1.

| Family | regular / bold | Source | Notes |
|---|---|---|---|
| `Pretendard` | Regular / Bold | [orioncactus/pretendard](https://github.com/orioncactus/pretendard) v1.3.9 | Modern sans, Inter-based Latin |
| `PretendardMedium` | Medium / ExtraBold | same | Heavier body weight |
| `RIDIBatang` | single weight | [RIDI](https://ridicorp.com/ridibatang/), mirror pinned by commit | RIDI's e-book serif |
| `MaruBuri` | Regular / Bold | [Naver](https://hangeul.naver.com/fonts/search?f=maruburi), mirror pinned by commit | Screen-reading serif |
| `MaruBuriSemiBold` | SemiBold / Bold | same | Heavier body weight |

Details:
- Sizes **8, 10, 12** are generated alongside 12–18 so the size-matched CJK UI fallback (`docs/sd-card-fonts.md`) renders Korean book titles and file names in the library/browser. These are the first catalog families to enable that.
- Intervals: `hangul,latin-ext,punctuation,reading,symbols` — no `cjk` preset. Korean books don't need the ideograph block, and including it roughly doubled file size and made the largest sizes fail to load on device.
- Weight-variant families exist because none of these fonts ship italics but all ship several weights; pairing a heavier regular with a heavier bold keeps `<b>` distinguishable.
- Source URLs are pinned to a tag (Pretendard) or commit SHA (mirrors) so the release workflow rebuilds reproducibly.
- No firmware changes: the `hangul` browser group and glyph preset already exist.

Built locally with `build-sd-fonts.py` (12–24 MB per family, six `.cpfont` each) and tested on an Xteink X3: Korean renders correctly in book content and in the 8/10/12 pt UI rows.

---

### AI Usage

Did you use AI tools to help write this code? _**< YES >**_ — Claude Code drafted the catalog entries, verified licenses/coverage and ran the local font builds; font selection, weight pairings and on-device testing were done by me.

### DIFF
--- a/lib/EpdFont/scripts/sd-fonts.yaml
+++ b/lib/EpdFont/scripts/sd-fonts.yaml
@@ -334,7 +334,7 @@ families:
   - name: Pretendard
     description: "Modern Korean sans-serif (Hangul, Latin)"
     scripts: [hangul, latin]
-    intervals: hangul,cjk,latin-ext,punctuation,reading,symbols
+    intervals: hangul,latin-ext,punctuation,reading,symbols
     sizes: [8, 10, 12, 14, 16, 18]
     styles:
       regular: {url: "https://raw.githubusercontent.com/orioncactus/pretendard/v1.3.9/packages/pretendard/dist/public/static/Pretendard-Regular.otf"}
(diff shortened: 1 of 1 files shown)


serialx:
BTW: I wanted spaces in the font name field but Claude Code said it would break.

> - release-fonts.yml uploads the .cpfonts as GitHub release assets, and GitHub rewrites spaces in asset filenames to . — the asset would be served as Pretendard.Medium_12.cpfont.
> - The manifest, though, records the original filepath.name (generate-font-manifest.py:214), and the firmware builds the download URL by raw concatenation with no URL encoding: std::string url = baseUrl_ + file.name; (FontDownloadActivity.cpp:449). So the device would request .../Pretendard Medium_12.cpfont — a URL with a literal space that also doesn't match the renamed asset. Download fails.

Uri-Tauber:
I don't feel strongly about it, but I think it's a shame to choose three weights of the same font. I'd go with something typographically different. Unless it's standard in Korea to use the exact same typeface in every and each book.

serialx:
> I don't feel strongly about it, but I think it's a shame to choose three weights of the same font. I'd go with something typographically different. Unless it's standard in Korea to use the exact same typeface in every and each book.

@Uri-Tauber Ah, I didn't really explain why I have added multiple weights of the same font. It's fairly normal for Korean eink readers to support font weight override. Since you'll be unable to browse Korean web, I'll quote ChatGPT[1]'s reply:

> Korean text is also particularly sensitive to perceived stroke thickness on e-ink displays: thin Hangul strokes can look faint depending on the font, panel resolution, antialiasing, and refresh/rendering. So readers may want something closer to “make all body text slightly heavier,” rather than merely choosing a naturally heavier font.

Even non-eink ebook readers support font weight override support:

IMG_4999

Also If you'd like for me to add multiple types of Korean fonts, I can add a few popular fonts too. But surely for them I'd like to add multiple font weights too. :)

[1]: https://chatgpt.com/share/6a8ae9d5-b138-83ee-ade7-79f2bf7b1be1

serialx:
Or If you'd like for me to add only one per fonts for now, I can remove the bolder weights. Just tell me. I'll do it. 👍

serialx:
I removed Pretendard Bold, and added two popular Korean ebook fonts that are serif. RIDIBatang does not have ANY bold fonts so it doesn't support bold at all. But it's very popular eink font so I've included it.

I've tried to include KoPub Batang/Dotum fonts, which are also popular fonts. But they are not OSI compliant license, so I've excluded them from this PR.

Uri-Tauber (approved):
LGTM

Uri-Tauber:
Thank you.
I do want to have a hard cap on the number of fonts per language, at least for start. 5 is a bit over what I have in mind, but its fine.

PR #3186, opened 2026-08-23
fix(fonts): stop Ubuntu Medium glyphs colliding in the UI

fain182:
## Summary

* **What is the goal of this PR?** Fix Cyrillic glyphs that render fused together in the UI, reported in #3099.
* **What changes are included?** Ubuntu Medium is converted through the FreeType auto-hinter instead of its own bytecode; the two Medium UI headers and the font IDs are regenerated.

At UI ppem Ubuntu Medium's bytecode grid-fits `U+0442` (т) to 11px inside a 10px advance and lands its left bearing at **-1**, so the glyph hangs into the one before it. In Russian "скрытые" the `ы` and `т` merge into a single shape and the word reads as something else.
fix_before_after

The outline is not at fault — its left bearing is positive in both weights — and Ubuntu Bold grid-fits cleanly, so only the Medium face changes. It joins the Hebrew fallback already on the auto-hinter in #3176.

```
U+0442   before: { 11, 11, 161, -1, ... }     after: { 10, 11, 161, +1, ... }
gap after ы:     -0.5px (overlap)                    +2.5px
```

**Stacked on #3176** — it needs the per-face `--autohint-font` flag introduced there, so the diff shows that PR's commit too. Merge after it.

## Scope Check

- [x] The stock firmware does not already handle this well.

## Additional Context

Layout is untouched. The converter stores `linearHoriAdvance`, which is unhinted, so hinting cannot move anything: **zero** of the 1031 glyph advances change, and Latin, Cyrillic and Vietnamese strings measure identical widths before and after. Only bitmaps and bearings move — 320 glyphs at 10pt, 269 at 12pt.

No stem drops to a 1px hairline in Latin, Greek or Cyrillic at either size, so nothing gets thinner on e-ink. `verify-ui-noto-fonts.py` passes (Hebrew 101.9%/95.3%, Arabic 95.3%/102.6%).

Flash 5,706,019 bytes (87.1%), RAM 16.6% — 46 bytes over #3176 alone.

Worth noting for reviewers: Ubuntu Sans 1.006 fixes this same defect upstream (it is hinted with ttfautohint), but even its Bold renders 2px stems at UI ppem, which would give back the weight #3096 was after. That is why this fixes the hinting rather than bumping the font.

---

### AI Usage


fain182:
@bizzon2 your photos from #3099 found a real bug — thank you for taking them.

At the size the file list and settings use, Ubuntu Medium was drawing `т` one pixel to the left of where it belongs, so it ran into the letter before it. In "скрытые" the `ы` and `т` merged into one shape. This PR fixes that; the gap comes back and the word reads correctly again.

Could you try it? Once the checks below go green, open the **Build xteink_x4_x3** check → *Summary* → download the firmware artifact, and flash that.

One thing I want to be honest about: I measured the rest of your report carefully and could not reproduce it. Compared with 1.5.0, the new font uses the same amount of ink or slightly less, and the words come out slightly **wider**, not tighter. What did change is that 1.5.0 drew stems as a ragged mix of 1px and 2px and they are now all even — my guess is that evenness is what reads as "bolder", but that is a guess.

So this fixes the merged letters, not the heavier look. If the text still feels too bold to you after trying this build, say so and we will treat that as its own question — it is worth solving properly rather than reverting a change that helped other scripts.

itsthisjustin (approved):

PR #3188, opened 2026-08-23
fix: progress bar thickness setting ignored in landscape (X4 Pro)

### LINKED ISSUE #3017: Progress bar thickness settings don't work in landscape modes (X4 Pro)
### Affected Version

X4Pro Beta21

### Describe the Bug

In either Landscape CW or Landscape CCW when the progress bar is enabled, the thickness setting doesn't have any effect, it's always thick, perhaps thicker than it's supposed to be. Progress bar displays correctly in Portrait mode.

[progress_bar_landscape_bug.bmp](https://github.com/user-attachments/files/31043345/progress_bar_landscape_bug.bmp)

### Steps to Reproduce

1. Settings -> Display -> Reading Orientation, set to Landscape CW
2. Settings -> Display -> Customise Status Bar, set Progress Bar to "Chapter" and Progress Bar Thickness to Thin.
3. Open a book, observe the progress bar

### Expected Behavior

The progress bar should change thickness according to the setting selected.

```shell

```

puls72s:
## Summary

* **What is the goal of this PR?** Fix #3017 — the reader's progress bar thickness setting appears to have little to no effect in landscape orientations on the X4 Pro, and the bar looks permanently thick there.
* **What changes are included?** `BaseTheme::drawStatusBar`'s progress-bar fill-to-edge extension used `orientedMarginBottom`, which rotates with reading orientation. On the X4 Pro, `viewableInsets.right`/`left` (7px — tuned for scrollbar visibility, per the existing code comment, not measured bezel overlap) differ a lot from `top`/`bottom` (9/3px), so landscape orientations picked up a much larger fill amount than portrait, inflating the bar's rendered height independently of the user's Progress Bar Thickness setting. The fill amount now always uses the board's actual (unrotated) bottom inset, so it's identical in every orientation. Portrait's appearance is unchanged; landscape now matches it.

## Scope Check

- [x] Does not touch `freeink-sdk/`, `lib/hal/`, bootloader, OTA, or recovery code — the fix stays entirely inside `BaseTheme.cpp` and reads existing `BoardConfig` data, doesn't change it.

## Additional Context

While investigating, I measured the X4 Pro's actual physical bezel overlap with a throwaway diagnostic firmware (a full-screen pixel ruler) and photographed it: all four edges show close to 0px real overlap, versus the 9/7/3/7px `viewableInsets` currently assumed for this board. That's a separate, bigger question about the underlying hardware calibration data (which lives in the `freeink-sdk` submodule, a different repo) — filed as [Free-Ink/freeink-sdk#54](https://github.com/Free-Ink/freeink-sdk/issues/54) rather than bundled in here. This PR only fixes the orientation-inconsistency bug in crosspoint-reader's own rendering logic, independent of whatever those numbers turn out to be.

Docs-only... no — code-only change, no memory/flash impact (single `int` swap). Verify by opening a book with Progress Bar set to "Chapter"/Thin vs Thick, comparing Portrait vs Landscape CW/CCW — the bar's height should now match between orientations for the same setting.

### AI Usage

Did you use AI tools to help write this code? **YES** — Claude (Anthropic) traced the rendering path, took and analyzed the hardware measurement, and drafted the fix; reviewed by the submitter before opening this PR.

### DIFF
--- a/src/components/themes/BaseTheme.cpp
+++ b/src/components/themes/BaseTheme.cpp
@@ -1,5 +1,6 @@
 #include "BaseTheme.h"
 
+#include <BoardConfig.h>
 #include <FreeInkUIGfxRenderer.h>
 #include <GfxRenderer.h>
 #include <HalClock.h>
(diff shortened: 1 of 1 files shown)


itsthisjustin:
The scrollbar only effects the right edge so it's possible the bottom inset is wrong in the sdk. If we set bottom inset to 0 does this bug go away?

itsthisjustin:
Closing this as we found out that every x4s display is installed slightly differently. Correct me if I'm understanding the PR wrong

PR #3197, opened 2026-08-24
feat: Add BLE HID input support 

itsthisjustin:
Implements BLE HID keyboard input functionality for CrossPoint. Includes a PlatformIO pre-build script that patches arduino-esp32 core to define _btLibraryInUse as a weak symbol, fixing linker errors when using NimBLE-Arduino host with ESP-IDF BLE controller-only configuration. The patch ensures compatibility when IDF host stacks are disabled to avoid symbol conflicts.


src/activities/reader/EpubReaderMenuActivity.cpp
```diff
@@ -179,6 +201,14 @@ void EpubReaderMenuActivity::buildScreen(UiScreen& screen) {
       menuRowItems[i].value = I18N.get(orientationLabels[pendingOrientation]);
     } else if (action == MenuAction::AUTO_PAGE_TURN) {
       menuRowItems[i].value = pageTurnLabels[selectedPageTurnOption];
+    } else if (action == MenuAction::TOGGLE_BLUETOOTH) {
+      if (SETTINGS.bluetoothEnabled) {
+        menuRowItems[i].value = !BleHid.isRunning() ? I18N.get(StrId::STR_CONNECTING)
+                                                    : (BleHid.isConnected() ? I18N.get(StrId::STR_STATE_ON)
+                                                                            : I18N.get(StrId::STR_CONNECTING));
```
unnamedd: Let's avoid nesting one ternary inside of another. This makes the code confusing.

src/activities/settings/BluetoothSettingsActivity.cpp
```diff
@@ -0,0 +1,310 @@
...
+}
+
+void BluetoothSettingsActivity::activateIndex(const int index) {
+  if (index < 0 || index >= rowCount) return;
+  app.clearTapFlash();
+  const int16_t value = rowValues[index];
+
+  if (view == View::Menu) {
+    switch (static_cast<Action>(value)) {
+      case Action::ToggleBt:
```
unnamedd: The action can be better named

```suggestion
      case Action::ToggleBluetooth:
```

src/activities/settings/BluetoothSettingsActivity.cpp
```diff
@@ -0,0 +1,310 @@
...
+  return false;
+}
+
+void BluetoothSettingsActivity::drawChrome() {
+  UiListActivity::drawChrome();
+  // Sub-header: transient banner when set, else the live connection status.
+  const auto& metrics = UITheme::getInstance().getMetrics();
+  const char* status = !banner.empty()        ? banner.c_str()
+                       : BleHid.isConnected() ? BleHid.connectedName()
+                                              : tr(STR_BT_NOT_CONNECTED);
```
unnamedd: Another nested ternary that can be split into at least 2 variables to keep better readability

src/activities/settings/BluetoothSettingsActivity.cpp
```diff
@@ -0,0 +1,310 @@
...
+  props.count = static_cast<uint16_t>(rowCount);
+  props.action = ACTION_ROW;
+  props.inputMask = fui::InputTouch | fui::InputLongPress;  // physical buttons stay in loop()
+  // Master-switch geometry: larger than the list default so it reads as the
+  // primary control, inset from the row edge, and rounded with the same theme
+  // shape token the buttons and the frontlight panel's sliders inherit.
+  const auto& theme = screen.theme();
+  props.toggleWidth = mappedInput.hasTouch() ? 52 : 44;
+  props.toggleHeight = mappedInput.hasTouch() ? 26 : 22;
+  props.toggleKnobInset = 4;
```
unnamedd: Those magic numbers are dangerous for a future SD Themes, and we shouldn't use them, in case strictly necessary, we need to move them to a file with constants values.

unnamedd (commented):
Forgot to push some pending comments I made here... I didn't finish the review yet. Will back for it soon.

PR #3143, opened 2026-08-20
fix(epub): improve RTL paragraph direction auto-detection and line alignment

bjesus:
Fix some RTL edge cases.

Before:
screenshot_2026-08-20_19-47-03_151594582screenshot_2026-08-20_19-47-16_362853196

After:
screenshot_2026-08-20_19-45-04_061165367 screenshot_2026-08-20_19-44-42_851779830 

### Changes

1. **Paragraph-Level Script Direction Detection:**
   - Added `BidiUtils::detectTextDirection()` helper to compare strong RTL (`R`/`AL`) vs LTR (`L`) character counts across word collections.
   - Replaced 5-word paragraph RTL probe in `ParsedText::layoutAndExtractLines()` with `detectTextDirection()`, correctly identifying RTL paragraphs even if they start with a Latin word.

2. **Consistent Line Alignment & Last-Line Positioning:**
   - Evaluated line alignment, visual word reordering (`computeVisualWordOrder`), and line positioning using paragraph base direction (`blockStyle.isRtl`).
   - Ensured un-stretched last lines in RTL paragraphs align flush against the right margin (`xpos = effectivePageWidth - contentWidth`) instead of floating on the left edge.

3. **Unit Tests:**
   - Added unit tests in `MiniBidiArabicTest.cpp` verifying visual word ordering for Hebrew with embedded Latin words and dominant script direction detection.

### Testing
- Tested in the desktop simulator with mixed Hebrew/Latin test cases and real EPUB books.
- All 140 GoogleTest unit tests pass (`ctest`).

PR #3198, opened 2026-08-24
fix: RTL inline direction inheritance

Uri-Tauber:
# Summary

Mixed Hebrew and English passages now keep the book's intended reading direction, including paragraphs that begin with an English term.

suppress #3143, which was a downstream change, rather than taking care of the root.

---

### AI Usage

### DIFF
--- a/lib/Epub/Epub/Section.cpp
+++ b/lib/Epub/Epub/Section.cpp
@@ -43,7 +43,9 @@ namespace {
 // v40: Ruby groups remain intact when a large text block is soft-flushed.
 // v41: Simple HTML table rows are laid out as positioned columns instead of
 //      flattened paragraphs with synthetic row/cell labels.
-constexpr uint8_t SECTION_FILE_VERSION = 42;
+// v42: Closing a block strips inherited vertical margins and padding.
+// v43: Paragraph base direction excludes direction changes from inline elements.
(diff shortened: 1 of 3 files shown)


itsthisjustin:
@Uri-Tauber you ready for this to be merged?

Uri-Tauber:
@itsthisjustin not yet. I think I can do this in a cleaner way.

Uri-Tauber:

itsthisjustin (approved):

bjesus:
Looks great! The only thing I still see left is this issue, with parenthesis mixed with numbers or LTR text.

screenshot_2026-09-01_17-54-25_420570164

Uri-Tauber:
@bjesus Yeah. I already tried in the past to fix that, and decided it wasnt worth the extra data structures needed to maintain it.
Basickly, parenthesis (along with period, comma, space, etc...) are **weak** BiDi characters (I pretty sure its the right term), which mean they don't have their own direction - LTR or RTL - but they inherit the direction of the text before them. 
Which cause the problem you describe, where the numbers make parenthesis LTR.

bjesus:
I understand now, thank you for explaining! I guess we can indeed leave this aside for now, it's more of a nicety rather than a blocker.

Issue #2612, opened 2026-07-14
Feature request: Expose reading session data through KOReader Sync / BookOrbit integration


IreneGoodnight:
Hi!
First of all, thank you for CrossPoint — it's an amazing firmware and I've been really enjoying using it on my XTEINK X4. I have a feature request regarding the KOReader Sync implementation.

At the moment, CrossPoint successfully synchronizes the current reading position with KOReader Sync servers (including BookOrbit), so the reading percentage is updated correctly. However, only the reading position is synchronized.

As a result:
BookOrbit updates the reading percentage, but does not create Reading Sessions or reading history entries.
Other KOReader devices connected to the same sync server receive only the updated percentage, but do not receive reading statistics such as reading time, session history, etc.

The official BookOrbit plugin for KOReader sends richer reading events, allowing BookOrbit to record:
reading sessions,
reading history,
reading time,
reading statistics,
device activity.

CrossPoint currently appears to send only the standard progress update.
Would it be possible to expose additional reading session information through the synchronization API, or support the extended BookOrbit protocol used by the official KOReader plugin?
Even basic session information (start/end of reading, timestamp, reading duration, etc.) would significantly improve synchronization with BookOrbit and other KOReader-based ecosystems.

My use case:
XTEINK X4 with CrossPoint + BookOrbit as the library server + KOReader on Android connected to the same server

Currently:
+ Reading progress synchronizes correctly.
+ Book matching works correctly.
- Reading sessions are not created.
- Reading time and statistics are not synchronized across devices.

Thank you for considering this feature! <333

jb567:
There are a couple of issues that might make this more tricky.

The X4 clock is not accurate, so this would likely be far more accurate on the X3, if you are trying to capture reading session information.

Currently crosspoint doesn't have stats tracking, and its been [discussed](https://github.com/crosspoint-reader/crosspoint-reader/issues/1600), but I'm not 100% on where it ended up.

I know the forks [cpr-vcodex](https://github.com/franssjz/cpr-vcodex) and [crossink](https://github.com/uxjulia/CrossInk) have some stats tracking.

[...]

PR #3204, opened 2026-08-24
feat: expose reader sessions to SD plugins

samfoy:
## Summary

- add a service-neutral `reader.session` event with fixed-scalar active-reading accounting
- flush sessions before the existing sleep event drain, with an idempotent exit fallback
- let any queued plugin event with `connect: true` request bounded sleep-time Wi-Fi
- treat events-only `device.json` manifests as background plugins and open their README when present
- document and test the exact session contract, including compact `end_time == start_time + duration_seconds`

This targets `feat-sd-plugins` and is intentionally a dependent extension of #3114. It does not add a provider endpoint, transport, cache, sidecar, or provider-specific firmware behavior. The existing plugin event owner carries the complete lifecycle.

A concrete consumer is published separately as the [BookOrbit Reading Sync v1.0.0 plugin](https://github.com/samfoy/crosspoint-bookorbit-plugin/releases/tag/v1.0.0). All BookOrbit URLs, credentials, payloads, and setup UI remain in that SD plugin rather than firmware.

Addresses #2612.

## Contract

`reader.session` exposes these string substitutions:

- `book`
- `document` (existing binary KOSync partial-MD5)
- `start_time`
- `end_time`
- `duration_seconds`
- `start_progress_bp`
- `end_progress_bp`
- `progress_scale` (`10000`)

Only successful ordinary forward turns contribute. Dwells under 2 seconds are ignored, each contribution is capped at 30 minutes, and summaries under 10 seconds are not emitted. The helper uses 40 bytes of bounded scalar state and performs no per-page allocation or I/O.

## Verification

- `./bin/clang-format-fix -g`
- `cmake -S test -B build/test && cmake --build build/test -j2`
- `ctest --test-dir build/test --output-on-failure -j2`: **162/162 passed** on this exact PR commit
- local whole-feature acceptance gate on the equivalent product tree: **162/162 passed**, structural lifecycle checks passed, X4 Pro linked with `ReaderSession` symbols
- `pio run -e gh_release -j1`: **passed** for the shared X3/X4 ESP32-C3 release target
  - RAM: **55,316 / 327,680 bytes (16.9%)**
  - flash: **5,810,665 / 6,553,600 bytes (88.7%)**
  - `esptool image-info`: detected **ESP32-C3**, valid checksum and validation hash
  - C3 ELF: six `ReaderSession` symbols linked
- X4 Pro build: **passed**
  - RAM: **65,524 / 327,680 bytes (20.0%)**, no baseline increase
  - flash: **5,701,286 / 6,553,600 bytes (87.0%)**, +2,524 bytes from the recorded plugin-branch baseline
- formatter, `git diff --check`, clean tracked/index state, and service-coupling check passed


itsthisjustin:
The use case here is syncing to bookorbit like kosync or what? Just trying to understand why we'd do it this way vs just using the crosspoint sync server over kosync

samfoy:
Good question, and the short answer is that KOSync and this event carry different shapes of data, so the sync server can't absorb this without becoming a different protocol.

**KOSync is a cursor; this is an interval.** `KOReaderProgress` is `document` + `progress` + `percentage` + `timestamp` — "where am I right now", last-write-wins. There is no duration, time-spent, or elapsed field anywhere in `lib/KOReaderSync`. What BookOrbit-style reading sync needs is the opposite shape: an append-only log of *sessions*, each with a start, an end, and a progress delta.

You can't derive the second from the first. Reconstructing dwell time server-side from a stream of cursor pushes means the server has to guess what happened between two positions, and it guesses wrong in exactly the common cases — device slept mid-chapter, book left open on the nightstand, reader idled with the page up, progress moved backward on a re-read. The device is the only place that knows a turn actually succeeded and that the dwell was real. That's why the accounting lives in `ReaderSession` and drops dwells under 2s, caps each contribution at 30 minutes, and refuses to emit under 10s.

**On extending the CrossPoint sync server instead:** that's the option I'd expect to be right, and it still needs the firmware-side half that this PR is. Sessions only close at a boundary — the flush here hangs off `prepareForSleep()` ahead of the existing sleep event drain, with the `connect: true` path for bounded sleep-time Wi-Fi. That plumbing is the actual work; the wire format is the easy part. Once a session exists as an event, a sync-server-backed consumer is just another plugin subscribing to it, and BookOrbit stops being special.

**And it keeps SCOPE.md §3 satisfied.** "No new external network connectors" is the line, and the alternative — firmware talking to BookOrbit directly — is squarely on the wrong side of it. This adds no endpoint, transport, cache, sidecar, or provider-specific behavior; all URLs, credentials, payloads, and setup UI live in the SD plugin. It's the same pattern `KOReaderMetadata::extra` already uses, where the firmware forwards sidecar keys verbatim and assigns them no meaning.

So the trade is: one service-neutral event with a documented contract, versus either firmware that knows about a service or a server that infers reading time it can't actually see. Happy to be talked out of it if you'd rather the session shape land as a KOSync extension first — that changes where the contract is documented, not what the firmware has to do.

samfoy:
`a267ea24` is the new head, and the PR reports MERGEABLE with unchanged scope: 22 files, 501 insertions, and 35 deletions.

- The rebase from 20 commits behind `feat-sd-plugins` produced one purely additive conflict in `test/CMakeLists.txt`.
- This branch added `reader_session` and `plugin_manifest_classification`, while upstream added `font_cache_manager` and `css_parser`.
- I kept all four entries because every other line was identical between both sides.
- The patch remains identical at 22 files and +501/-35, with matching change counts for every file.
- All 10 checks pass on the rebased head, including `unit-tests`, `cppcheck`, `clang-format`, and all four firmware builds.

samfoy:
I rebased onto `feat-sd-plugins` and pushed `e4ce7b1b`: the pull request now reports MERGEABLE, 0 commits behind, versus CONFLICTING and 89 behind. The scope remains 22 files, +577/-60.

## Continuous integration

Continuous integration (CI) is red on this pull request and the base branch, with the same 7 failures and 2 passes.

| Checks | This pull request | Base run `34783063427` |
|---|---|---|
| `Build default`, `Build papermono`, `Build sticky`, `Build x4c`, `Build x4pro`, `unit-tests`, `Test Status` | 7 fail | 7 fail |
| `clang-format`, `cppcheck` | 2 pass | 2 pass |

Base run `34783063427` uses `feat-sd-plugins` at `dfe2c5f4`, my merge base, without my commit.

The firmware fails in `lib/hal/HalDisplay.h` and `lib/hal/HalDisplay.cpp`, with errors such as `'GrayscaleCapabilities' in namespace 'freeink' does not name a type`. These files come from the `freeink-sdk` software development kit (SDK) submodule. `.gitmodules` specifies `branch = main`. Both my head and the base use submodule commit `cb9167d541c0f6e9d57cf8eae1f564a939883ecc`, so this change cannot alter that submodule content. My diff touches no hardware abstraction layer (HAL), grayscale, or display file.

The `unit-tests` compile failure is in `test/chapter_html_slim_parser/ChapterHtmlSlimParserTest.cpp`: `no match for 'operator==' (operand types are 'const WordStore::StoredWord' and 'const char [6]')`. I reproduced the identical 2 errors on upstream alone, with my commit absent. My diff touches neither that test nor `WordStore`.

## Conflict resolution

I resolved 8 conflict hunks across 6 files.

| File | Resolution |
|---|---|
| `test/CMakeLists.txt` | Kept upstream's 10 new subdirectories and my 2. The additions do not overlap. |
| `src/activities/reader/ReaderActivity.h` | Reinstated `getProgressPercent()` for 5 call sites, as detailed below. |
| `src/activities/reader/ReaderActivity.cpp` | Kept upstream's memory log line and my `flushReaderSession()`. The flush precedes `ReaderExit` so the session is durable first. |
| `src/main.cpp` | Adopted upstream's early-return structure from "Simplify sleep.enter WiFi connection logic" and kept my widened predicate, as detailed below. |
| `src/activities/plugins/PluginCatalogActivity.cpp` | Resolved 2 hunks: expressed my three-way `deviceKind` classification through upstream's new `addRow` lambda and kept the 1 include still in use. |
| `docs/plugin-events.md` | Resolved 2 hunks: kept both contract statements because the merged code implements both rules. |

The helper returns because this change introduces 5 call sites. Upstream added `getProgressPercent()` in `d3e55c53` with an earlier feature slice, then removed it in `c1e1fec3`, which "removes unnecessary wrapper structs". That removal was correct: no caller remained. I recorded this history above the method.

| Caller | Uses |
|---|---|
| `ReaderActivity.cpp` | Progress string and session render-complete basis points: 2 sites. |
| `EpubReaderActivity` | Its `getProgressBasisPoints()` override and 2 fallback sites: 3 sites. |

The sleep-entry WiFi join uses this pull request's new `pluginevents::wantsConnectAny()` instead of `pluginevents::wantsConnect(pluginevents::Event::SleepEnter)`. This change adds a fifth event, `reader.session`, which queues during the reading turn and drains on the same sleep. When nothing subscribes to `sleep.enter`, that predicate leaves a `reader.session` subscriber undelivered. This is the common case for a progress-sync plugin.

Both document versions captured part of the contract. In `src/util/PluginEvents.cpp`, line 88 reads each handler's `"connect": true` key. Line 93 forces that bit on for `SleepEnter`. I verified that the 2 rules coexist and documented both.

## Local verification and patch preservation

I built the host tests with cmake and ninja, matching `.github/workflows/ci.yml` lines 169 to 175.

| Stage | Result |
|---|---|
| Old head `a267ea24`, before rebase | 186 of 186 tests passed |
| After rebase, excluding the pre-existing failure above | 253 of 253 tests passed |
| My suites: `ReaderSession` and `PluginManifestClassification` | 15 of 15 passed |
| All 24 other test targets | Built with 0 errors |

The extra 67 tests come from upstream's new suites that the rebase pulled in.

I compared patches against their merge bases because a head-to-head diff includes 89 upstream commits. The file set remains identical: 22 files. Of those, 16 are byte-identical. The 6 changed files are exactly the 6 with conflict resolution.

If useful, I can open a separate pull request for the `ChapterHtmlSlimParserTest` comparison failure. The firmware break needs an SDK pin decision from the maintainers, outside this pull request.

src/activities/reader/ReaderActivity.cpp
```diff
@@ -97,6 +109,39 @@ void ReaderActivity::onExit() {
   endOfBookOptionsReady.store(false, std::memory_order_release);
 }
 
+void ReaderActivity::prepareForSleep() { flushReaderSession(); }
+
+void ReaderActivity::flushReaderSession() {
```
dasJ, line 114: I think it would also make sense to also trigger this in `launchKOReaderSync()` or even in `saveProgress()`

PR #3214, opened 2026-08-26
fix(x4pro): hold power-latch pins HIGH through deep sleep

### LINKED ISSUE #2863: X4 Pro - Short Press Power Button to Sleep/Wake
### Affected Version

1.5.0-x4pro

### Describe the Bug

Not sure if this is a bug (entirely new to engaging with custom firmware!)- I previously used Crosspoint on my X3, and had the Short Power Button Click set to 'Sleep'. This made it so that I could easily sleep & wake the device with a quick button press (much nicer than holding the button down).

I received my X4 Pro yesterday and flashed Beta 10, and then set up the same setting there. It does work to sleep the device, but waking the device still requires holding the button down (and the X4 Pro's power button seems a bit more finicky in general).

### Steps to Reproduce

1. Flash Beta 10 on X4 Pro
2. Set 'Short Power Button Click' to 'Sleep'
3. Attempt to wake the device from sleep

### Expected Behavior

It'd be great if the same short press was able to wake the device, but I'm aware this may be down to different hardware rather than anything Crosspoint can impact!

```shell

```

Belphemur:
## Summary
Battery-only X4 Pro units lost their fast deep-sleep wake (slow cold boot on the next power-button press). Root cause: `freeink::PowerManager::deepSleep()` calls `esp_sleep_config_gpio_isolate()`, which strips every pad that has no armed `gpio_hold_en()`. The X4 Pro's master peripheral-rail enable is carried as `power.latch0` = GPIO1 — `holdPowerRails()` drives it HIGH at boot, but nothing armed a sleep hold for it, so the latch floated as soon as the serial/pogo adapter stopped feeding the rail. Adapter-connected devices masked the bug because external power kept the rail up.

The fix holds every configured power-latch pin driven and held HIGH before entering deep sleep in `HalPowerManager::startDeepSleep()`:

- Generic over `BoardConfig::ACTIVE.power.latch0/latch1` — covers the X4 Pro's GPIO1 and the Sticky's PWR_HOLD/PWR_LOCK (GPIO45/46) with no per-board code.
- Skips `XTEINK_C3_GPIO13`: it *is* `power.latch0` on the C3 Xteink boards and is driven LOW in this same function on purpose (battery power-off).
- Peripheral rails (SD / touch / EPD reset) keep their existing OFF-level holds from `powerDownRailsForSleep()`, so sleep current stays low.
- Same release-hold-before-drive discipline used by the GPIO13 block and `holdPowerRails()`.

Fixes #2863

## Testing
- [x] `pio run -e x4pro` — SUCCESS
- [x] clang-format 21 clean (`bin/clang-format-fix -g` reports no changes)
- [ ] Hardware: X4 Pro, battery only → sleep → press power button wakes instantly instead of cold boot


Belphemur:
Sorry for the noise, too much in this PR, I need to test it more anyway

PR #3215, opened 2026-08-26
fix(x4pro): hold power-latch pins HIGH through deep sleep

Belphemur:
## Summary

* **What is the goal of this PR?** Fixes the battery-only deep-sleep wake bug on the Xteink X4 Pro (crosspoint-reader/crosspoint-reader#2863). After a battery-only sleep, the next power-button press cold-booted (~8–15 s to content) instead of fast-waking (~1–2 s). With the serial/pogo adapter connected, external power masked the problem.
* **What changes are included?** One change in `HalPowerManager::startDeepSleep()` (+21 lines, `lib/hal/HalPowerManager.cpp` only): every configured power-latch pin (`BoardConfig::ACTIVE.power.latch0/latch1`) is driven HIGH and armed with `gpio_hold_en()` before entering deep sleep.

Root cause: `freeink::PowerManager::deepSleep()` calls `esp_sleep_config_gpio_isolate()`, which strips every pad without an armed hold. The X4 Pro's master peripheral-rail enable is `power.latch0` = GPIO1 — `holdPowerRails()` drives it HIGH at boot but arms no sleep hold, so once isolated the latch floats LOW when external power is removed. The fix keeps all configured power-latch pins driven and held HIGH through sleep; peripheral rails (SD / touch / EPD reset) keep their existing OFF-level holds from `powerDownRailsForSleep()`, so standby current doesn't regress. The loop deliberately skips GPIO13 on the C3 Xteink boards: it *is* `power.latch0` there and is driven LOW on purpose as the battery power-off. Generic over the profile's latch pins — also covers the Sticky's PWR_HOLD/PWR_LOCK (GPIO45/46).

## Scope Check

Please confirm:

      the relevant maintainer.
      → Touches `lib/hal/` only (`HalPowerManager.cpp`); filed against #2863 where maintainer triage is already happening. No SDK changes required (SDK pointer unchanged at `e4d3cc3`).

Stock firmware note: this restores behavior stock users get implicitly (adapter plugged in or self-latching hardware revisions); it fixes wake for battery-only use, which stock firmware does not handle well.

## Additional Context

* Memory / flash impact: none measurable — ~21 lines of flash, no new RAM, no heap allocation. Runs once per sleep entry.
* Risk area to focus on: the C3 GPIO13 exclusion (battery power-off must still drop the pin LOW); covered by keeping that block ahead of the new loop and skipping the pin in the loop.
* Verification evidence:
  - `pio run -e x4pro` SUCCESS locally on this exact commit.
  - Identical patch fully green on CI twice: upstream run 32916960939 (Build default/sticky/papermono/x4pro, unit tests, cppcheck, clang-format) and Belphemur/crosspoint-reader#3 (all checks pass incl. CodeRabbit).
  - Hardware test procedure for reviewers: flash, sleep on battery, unplug adapter 60 s, short-press power → expect instant wake landing back on the page (`ESP_RST_DEEPSLEEP` + `ESP_SLEEP_WAKEUP_GPIO`), not splash/cold boot (`ESP_RST_POWERON`).

---

### AI Usage

Did you use AI tools to help write this code? _**YES**_ — root-cause analysis and patch drafted with an AI coding agent (Hermes), reviewed and tested by hand before submission.

### DIFF
--- a/lib/hal/HalPowerManager.cpp
+++ b/lib/hal/HalPowerManager.cpp
@@ -89,6 +89,27 @@ void HalPowerManager::startDeepSleep(HalGPIO& gpio) const {
   }
 #endif
 
+  // Hold every configured power-latch pin HIGH through deep sleep. These are
+  // keep-alive enables (the X4 Pro's master peripheral rail on GPIO1, the
+  // Sticky's PWR_HOLD/PWR_LOCK): deepSleep() isolates all pads
+  // (esp_sleep_config_gpio_isolate), so a latch without an armed hold loses its
+  // output driver and floats — on the X4 Pro the latch drops as soon as
+  // external power leaves (serial/pogo adapter unplugged), and the next power-
+  // button press cold-boots instead of fast-waking. holdPowerRails() asserted
+  // the latches at boot but arms no sleep hold; arm it here instead. Skips
+  // XTEINK_C3_GPIO13: it IS power.latch0 on the C3 Xteink boards, where the
+  // block above drives it LOW on purpose (battery power-off).
+  for (const int8_t pin : {BoardConfig::ACTIVE.power.latch0, BoardConfig::ACTIVE.power.latch1}) {
+    if (pin < 0 || static_cast<gpio_num_t>(pin) == XTEINK_C3_GPIO13) continue;
+    const auto g = static_cast<gpio_num_t>(pin);
+    // Release any surviving pad hold first: a held pad silently ignores the
+    // drive below (same trap as the GPIO13 block above).
+    gpio_hold_dis(g);
+    pinMode(pin, OUTPUT);
+    digitalWrite(pin, HIGH);
+    gpio_hold_en(g);
+  }
+
   // Cut the gated peripheral rails (touch/SD/EPD on boards like the Sticky) and
   // hold the enables off through deep sleep — otherwise the GT911 and SD card
   // stay powered all through "off" and drain the battery. No-op on boards with


itsthisjustin (approved):

Mark31415:
I did this change locally before and I'm experiencing 0,7% battery consumption per h in deepsleep. Maybe we need to look into further improvements. Does someone else have the problem with pr? Will pull it in my local branch and start testing now.

Belphemur:
Actually trying this https://github.com/Free-Ink/freeink-sdk/pull/63 on the screen to force it to go into deep sleep, testing this as we speak.

Currently, seems we have a power leak somewhere. It should only consume 3 µA (correct, parked) or 50 µA (tens-of-µA worst case), so nothing noticeable.

@Mark31415 do you have other suggestion, this PR was actually from reading your comment on the issue

Mark31415:
I'm about to test this AI idea:

> approach is to explicitly hold the frontlight control lines GPIO8 and GPIO9 LOW during deep sleep.
>Currently, the following happens:
>The brightness is set to a PWM duty cycle of 0.
>The LEDC channels remain configured as LEDC_SLEEP_MODE_KEEP_ALIVE.
>As a result, normal GPIO sleep isolation is disabled for GPIO8/9.
>There is no explicit gpio_hold_en() for these pins.
>If GPIO1 is now keeping the power supply enabled, the frontlight driver or its inputs could therefore be drawing >leakage current.
>A sensible sequence would be:
>Before entering deep sleep, configure GPIO8 and GPIO9 as outputs, drive them LOW, and enable the GPIO holds.
>On boot, release the holds before initializing LEDC.



-> I guess your fix is much better
> Actually trying this https://github.com/Free-Ink/freeink-sdk/pull/63 on the screen to force it to go into deep sleep, testing this as we speak.

Belphemur:
@Mark31415 — confirmed, that's a real leakage path and your fix shape is correct.

On the X4 Pro the frontlight LEDs (GPIO8 cool / GPIO9 warm) are LEDC channels configured `LEDC_SLEEP_MODE_KEEP_ALIVE`, so the pads stay driven by the LEDC peripheral through deep sleep. With #3215 holding the master rail (GPIO1) HIGH for fast-wake, the frontlight driver IC stays powered and keeps drawing quiescent/leakage current even at brightness 0. "Off" is dark, not depowered.

I've implemented it as you suggested — drive GPIO8/9 LOW + hold them through deep sleep, and release the LEDC KEEP_ALIVE clock — as a `FrontlightManager::park()` in freeink-sdk (the submodule that owns the frontlight driver), wired into the X4 Pro sleep path in the firmware:
- SDK: https://github.com/Free-Ink/freeink-sdk/pull/63 (head now includes the park)
- Firmware: https://github.com/Belphemur/crosspoint-reader/pull/...

`park()` drives both pads LOW and holds them (survives `esp_deep_sleep_start` via `gpio_deep_sleep_hold_en`); `releaseOnWake()` drops the hold at boot so the LEDC channels re-attach cleanly. No-op on boards without a frontlight. Builds green for `x4pro`.

Worth measuring with a battery in series to confirm the µA drop — the LED itself is off so this is driver-IC quiescent, likely tens of µA, but on a 1100 mAh pack it adds up over weeks. Your PR #63 panel-park test is the right way to force the sleep and check.

Mark31415:
Your PR is included in my local build and flashed, device is at 100% and i will check in 12h.

Belphemur:
@Mark31415 you also need this https://github.com/Belphemur/crosspoint-reader/pull/15

That is using the updated lib properly, else the frontlight doesn't get updated

If easier, you can try directly my fork (I have couple more feature ported from Crossink too for the X4Pro and other I built, like dictionary when long pressing a word)
https://github.com/Belphemur/crosspoint-reader

Mark31415:
I was cherrypicking with codex. But your dictionary feature seems to be exactly the next thing i wanted to develop. Will check your fork tomorrow!

Belphemur:
@Mark31415 you'll need to cherry-pick again, I had broken the front light, it wouldn't come back on. This is now fixed.

PR #3216, opened 2026-08-26
feat(controls): configurable Home-key double click (default: toggle frontlight)

Belphemur:
## Summary

* **What is this PR?** Makes the X4 Pro's capacitive **Home button double-click** perform a
user-configurable action, defaulting to **toggling the frontlight**.
  - `src/CrossPointSettings.h`: new persisted setting `homeButtonDoubleClickAction`
    (`Off` / `Frontlight` / `Go Home`; default `Frontlight`). `Off` keeps single clicks
    fully instant (the arbitration window is bypassed entirely).
  - `src/main.cpp`: home-key tap arbitration in the main loop (300 ms double-click window,
    mirroring the existing power-button double-click frontlight toggle) plus a shared
    `toggleFrontlightByShortcut()` helper now used by both paths.
  - `src/util/HomeTapTracker.h` + `test/home_tap_tracker/`: pure header-only state machine
    with host gtest coverage of the window/disarm semantics.
  - `src/MappedInputManager.{h,cpp}`: deferred-home-gesture latch so the delayed single click
    flows through the unchanged activity routing (`wasHomeGesture()`).
  - `src/SettingsList.h`: new Controls row, shown only on boards with a Home key
    (`BoardConfig::hasHomeKey()` filter), consistent with the existing reader-menu-gesture row.
  - i18n: new English string + regenerated `I18nStrings.*`/`I18nKeys.h` via
    `scripts/gen_i18n.py` (other languages fall back to English).

## Scope Check

Please confirm:

      the relevant maintainer.

Note on scope: the popular fork **CrossInk** (`uxjulia/crossink`) ships exactly this feature on
the X4 Pro (configurable Home tap / double-tap / long-press with frontlight as the double-tap
default). This PR ports that proven interaction model into the stock firmware, adapted to this
codebase (single configurable double-click action + Off state). No SDK/HAL code is touched —
the only `lib/hal`-adjacent change is an app-level latch inside `MappedInputManager`, which
already lives in `src/`.

## Additional Context

* Single-click behavior is preserved; it is delivered ~300 ms later while the firmware listens
  for a second click (same trade-off CrossInk makes). Setting the action to `Off` restores
  today's zero-latency clicks because the arbiter is skipped when disabled.
* Home-key holds are unaffected (SDK suppresses the companion tap); long-press-to-sleep,
  wake, and the power-button double-click path keep priority in the loop ordering.
* Memory impact: one persisted `uint8_t`, a few bytes of static state, no heap allocations;
  new flash strings are minimal and shared where possible.
* Verification: `pio run -e x4pro` clean; host test suite green including the new
  `home_tap_tracker` gtests; `scripts/gen_i18n.py` runs without missing-string errors.

---

### AI Usage


Belphemur:
Thanks @coderabbitai and @copilot-pull-request-reviewer — status of the findings against the final head (73785bd4):

1. **Late tap after a stalled frame must not count as a double click** (`HomeTapTracker.h`) — fixed in 293b4678: `update()` now checks expiry *before* honoring `secondTapSeen`, with a regression test (`LateTapAfterStallIsNotDoubleClick`).

2. **Expired single tap leaking through the Home screen** (`main.cpp`) — fixed in 293b4678: when the window expires while `ActivityManager::isOnHomeScreen()` (new accessor), the tap is swallowed instead of queued, since no activity would ever consume `wasHomeGesture()` there.

3. **Concurrent tap on the expiry frame must be preserved** (`main.cpp`, latest) — implemented in 73785bd4 exactly as suggested: after queueing the late single click, a tap present on the same frame re-arms a fresh window instead of being lost.

Additionally, 73785bd4 refined arbitration after checking the SDK: `InputManager` clears its one-shot edges every `update()`, so the arbiter now samples Home edges every frame but only intercepts frames that actually carry Home events — page turns/touch are never delayed by the 300 ms window, holds flow through to the reader's configured long-press action, and the expired single click delivers on the same loop pass whenever possible.

Older comments on files outside this PR's diff belong to the fork-vs-upstream baseline and no longer appear in the changed-files list. CI is green across all four board builds, clang-format, cppcheck, and host unit tests.

PR #3223, opened 2026-08-26
fix: Add fallback USB detection via battery charging state

itsthisjustin:
When no digital USB-detect pin is available, infer external power connection by checking if the battery is charging. This supports boards like Sticky that use analog voltage dividers instead of digital detect lines.


itsthisjustin:
@Uri-Tauber this solves an issue reported by seeed where battery status wasn't reported when plugged in AND it wasn't charging during reading

Uri-Tauber (approved):
The code LGTM.
I didnt checked the SDK change. Is it relevant to the fix?

PR #3185, opened 2026-08-23
fix(KOSync): avoid redundant NTP sync before KOReader sync

WhoTheHeck:
## Summary

* **What is the goal of this PR?**

  Remove the redundant KOSync-local NTP cycle and five-second wait before KOReader sync requests.

* **What changes are included?**

  CrossPoint already synchronizes its clock through the normal Wi-Fi flow. KOSync was starting another NTP cycle and waiting up to five seconds before every sync. This PR removes that extra step.

  KOSync requests from CrossPoint do not include a client timestamp, so the request does not need this extra clock synchronization.

  The crashes reported in #3121 and #3048 happen around this extra NTP lifecycle. KOSync was tested successfully on an X3 after this change, but that device did not reproduce the reported crashes before the change. Confirmation from affected users is therefore still required before claiming either issue is fixed, but removing the redundant wait at minimum accelerates the sync process.

## Scope Check

CrossPoint is intentionally narrow. See [SCOPE.md](../blob/master/SCOPE.md) and [ROADMAP.md](../blob/master/ROADMAP.md).
Please confirm:

- [x] I have read SCOPE.md and ROADMAP.md.
- [x] This PR is **not** a new built-in theme (themes are temporarily closed pending the move to SD-loaded themes).
- [x] This PR is **not** a new external network connector (sync engine, cloud storage, remote file access, etc.).
- [x] This PR is **not** an interactive app, writing tool, RSS/news/browser, media playback, or PDF feature.
- [x] The stock firmware does not already handle this well, **and** no other popular CrossPoint fork already does
      (or, if one does, I explain why CrossPoint still needs it below).
- [x] If this PR touches `freeink-sdk/`, `lib/hal/`, the bootloader, OTA, or recovery code, I have coordinated with
      the relevant maintainer. (Not applicable: this PR does not touch those areas.)

**If this PR was opened against the previous (broader) scope and was already in flight under Phase 0, link the
relevant Discussion or issue so reviewers can see the history.**

Not applicable; this is a new, narrowly scoped fix.

## Additional Context

This is the final PR covering the three KOSync issues and improvement opportunities I identified. It accompanies PRs #3111 and #3174.

Refs #3121

Refs #3048

Normal Wi-Fi-owned RTC/NTP synchronization remains unchanged. This change adds no heap allocation or protocol field. No RAM or flash increase is expected because it only removes code.

Local verification:

- Host CMake/Ninja build and CTest: 138/138 tests passed.
[...]

PR #3111, opened 2026-08-18
fix(KOSync): compare mapped KOReader sync positions

WhoTheHeck:
## Summary

* **What is the goal of this PR?**

  Fix Smart Sync choosing the wrong direction when CrossPoint's local percentage and KOReader's remote percentage disagree with the mapped reading order.

  Fixes #2982.

* **What changes are included?**

  * Compare trusted spine, page, and visible-text positions before falling back to percentage.
  * Preserve the local mapped position through the reader-to-sync activity handoff.
  * Map primary and alternate remote records before selecting the later position. The primary record remains selected for ties or unknown comparisons, and uploads remain bound to the configured primary hash.
  * Use the same comparison result for Smart Sync actions and manual dialog preselection.
  * Add host regression tests for both percentage-inversion directions and fallback behavior.

## Scope Check

CrossPoint is intentionally narrow. See [SCOPE.md](../blob/master/SCOPE.md) and
[ROADMAP.md](../blob/master/ROADMAP.md).

Please confirm:

- [x] I have read SCOPE.md and ROADMAP.md.
- [x] This PR is **not** a new built-in theme (themes are temporarily closed pending the move to SD-loaded themes).
- [x] This PR is **not** a new external network connector. It corrects the existing KOReader Smart Sync decision path.
- [x] This PR is **not** an interactive app, writing tool, RSS/news/browser, media playback, or PDF feature.
- [x] The stock firmware does not already handle this well, **and** no other popular CrossPoint fork already does
      (or, if one does, I explain why CrossPoint still needs it below). This fixes the existing Smart Sync ordering
      defect tracked by #2982.
- [x] This PR does not touch `freeink-sdk/`, `lib/hal/`, the bootloader, OTA, or recovery code; no relevant maintainer
      coordination is required.

## Additional Context

CrossPoint and KOReader calculate overall progress using different pagination and layout information, so their percentages are not always directly comparable. A position can be physically later in the book while reporting a lower percentage.

The comparison now prefers mapped evidence in this order: resolved spine, equal mapped page, visible-text offset, mapped page, and finally finite percentage with the existing tolerance. If no reliable comparison is possible, Smart Sync falls back to manual selection.

The comparator performs no dynamic allocation. The final default build reports unchanged static RAM and an increase of approximately 830 bytes of flash.

### Verification

- Host tests: **149/149 passed**, including 11 focused comparison tests.
- Static analysis, formatting, and the `default` ESP32-C3 and `sticky` ESP32-S3 firmware builds pass.
[...]

PR #3226, opened 2026-08-26
fix(KOSync): resolve precise KOSync pull progress positions

WhoTheHeck:

Resolve precise reading positions received from KOSync more reliably.

- Replace the simple incoming markup scanner with a streaming parser that understands EPUB markup.
- Handle XML comments, processing instructions, CDATA, declarations/DOCTYPE, namespace prefixes, attributes, and self-closing elements correctly.
- Keep `text()[N]` numbering aligned with the EPUB’s actual text nodes.
- Fall back to percentage mapping when an XPath cannot be parsed safely.
- Add focused regression tests for complete-buffer and one-byte streaming.

Refs #2512.

## Context

In #2512, @agosez reported that an XML comment such as:

`<!--This is a comment-->`

could be treated as an opening element by `ProgressMapper`. This corrupted element depth and text-node tracking while resolving an incoming XPath.

The problem was broader than comments alone. EPUB markup can also contain processing instructions, CDATA sections, declarations, namespace-qualified elements, quoted attributes, self-closing elements, and markup delimiters split across streamed chunks.

This PR replaces the simple scanner with a streaming parser that recognizes these different kinds of markup and keeps its parsing state between chunks. It uses fixed-size buffers and does not allocate memory while parsing.

If the parser encounters malformed or unsupported markup, it rejects the XPath result and keeps the existing percentage fallback instead of continuing with incorrect element depth or text-node information.

## Relationship to #3174

This PR complements #3174, but fixes the opposite direction:

- #3174 preserves a precise XPath when CrossPoint uploads progress.
- This PR resolves that XPath more reliably when CrossPoint pulls progress.

Together, they improve both sides of precise KOSync position handling.

This PR references #2512 rather than closes because the issue contains broader reports that still require separate end-to-end verification.

## Scope Check

- [x] This is a focused correction to the existing KOSync integration.
- [x] This does not add a new network connector or user-facing feature.
- [x] This does not touch `freeink-sdk/`, `lib/hal/`, the bootloader, OTA, or recovery code.
- [x] The streaming parser uses fixed-size state and buffers rather than new heap allocations.

## Verification

- Focused progress-mapper tests pass: 12/12.
- Full host test suite passes: 156/156.
- Formatting passes.
- Static analysis passes with no defects.
- The default firmware build passes.
- A normal KOSync sync and pull flow was verified on an X3.

The host tests cover:

- XML comments and processing instructions
- Visible and hidden CDATA
- Declarations and DOCTYPE internal subsets
- Namespace-qualified element names
- Quoted and unquoted attributes
- Self-closing and empty elements
- Text-node boundaries
- UTF-8, entities, and CRLF normalization
- Markup split into one-byte stream fragments
- Safe percentage fallback for malformed input

---

### AI Usage


WhoTheHeck:
> ⚠️ Outside diff range comments (1)
> > lib/KOReaderSync/ProgressMapper.cpp (1)> `205-233`: _🩺 Stability & Availability_ | _🟠 Major_ | _🏗️ Heavy lift_
> > > **Move `ParagraphStreamer` state out of the caller stack.**
> > > `ParagraphStreamer` requires at least 296 bytes for its members on ESP32-C3, before padding and base-class storage (`lib/KOReaderSync/ProgressMapper.cpp:203-330`). `toCrossPoint` keeps `strict` alive while it creates `relaxed` (`lib/KOReaderSync/ProgressMapper.cpp:1239-1247`). `generateXPath` also creates a local instance (`lib/KOReaderSync/ProgressMapper.cpp:1364-1365`). Move these instances to a reusable pool or `std::unique_ptr`; do not use bare `new`.
> > > 

This is a valid broader stack-usage concern, but it largely predates this PR. With the X3 compiler, ParagraphStreamer is 376 bytes on develop and 380 bytes on this branch. The reported toCrossPoint() stack frame changes from 1,136 to 1,152 bytes, while generateXPath() changes from 560 to 576 bytes. Moving these objects to the heap would add new allocation and OOM behavior beyond the markup-correctness scope of this PR.

Uri-Tauber:
@thiagokokada, seems like @WhoTheHeck open four different KoSync fixes. #3111 #3174 #3185 and #3226.
Now, I thought your #2805 made sync completely accurate. 
So, could you check if and how those PRs improve the sync?

thiagokokada:
> @thiagokokada, seems like @WhoTheHeck open four different KoSync fixes. #3111 #3174 #3185 and #3226. Now, I thought your #2805 made sync completely accurate. So, could you check if and how those PRs improve the sync?

From the PR descriptions they all make sense and were things that I left off from #2805:
- #3111: this is SmartSync related, something that I wasn't aware when I wrote #2805, but yes, when I was testing SmartSync later I realize this was needed to be fixed (I didn't though because I do not care too much about it)
- #3174: this was actually something that I was aware it was an issue, to actually work my approach needed to have the index for the chapter we were jumping finished. Gladly to see this being fixed
- #3185: I am not too aware how NTP sync works, but the description makes sense
- #3226: this is the only one that I am not too sure, I explicitly didn't want to use XPath in the original approach to not increase the firmware size too much. But if I remember correctly when I was looking at the original implementation from KOReader this was how it was supposed to be done. So while I believe it is probably the correct way to do this it is also the most invasive change

I will try to review and test at least the first 3 PRs since they're things that I am interested in getting fixed. The last one I am less sure.

WhoTheHeck:
To add to this: I agree that #3226 is of least importance and could be postponed. Maybe overhead doesn't justify the accuracy gains.

WhoTheHeck:
thanks for taking the time to review the 3 of them, @thiagokokada! I also still run all of them in my testbuilds ever since. Anything I can do on my side to help get these merged and enable more reliable sync for everyone, @Uri-Tauber?

PR #3239, opened 2026-08-27
feat: add EEGO A4 board profile + LM3630A frontlight

rimuwu1:
Adds the EEGO A4 (ESP32-S3 N16R8, UC8279C 768x552) board profile and fixes the HAL frontlight so the LM3630A controller actually initializes.

- platformio.ini: adds [env:eego_a4] extending [base] with esp32-s3-devkitc1-n16r8, EEGO A4 PSRAM/revision flags, and the frontlight-defines.
- FirmwareBoardTag: maps FREEINK_DEVICE_EEGO_A4 to board name 'eego_a4'.
- HalFrontlight: LM3630A-present check now runs after manager.begin() (the controller is probed there), so the UI frontlight control shows up. Requires the LM3630A driver in Free-Ink/freeink-sdk (see Free-Ink/freeink-sdk#62).

### DIFF
--- a/lib/hal/HalFrontlight.cpp
+++ b/lib/hal/HalFrontlight.cpp
@@ -5,9 +5,10 @@
 HalFrontlight HalFrontlight::instance;
 
 void HalFrontlight::begin(const uint8_t brightness, const uint8_t warmth, const bool on) {
-  if (!manager.present()) return;
-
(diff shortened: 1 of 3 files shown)


itsthisjustin:
I have a pr here: #3237 fyi that also adds support so i'll look at merging this afterwards

rimuwu1:
alright thanks!

PR #3246, opened 2026-08-28
chore: make custom_sdkconfig opt-in per env

itsthisjustin:
pioarduino triggers its hybrid Arduino-core rebuild on the mere presence of the custom_sdkconfig option, inherited from [base] included, and the hybrid lib-compile pass discards build_unflags (espidf.py replaces BUILD_UNFLAGS with "" after build_flags are already applied). Any env that needs to unflag base options can therefore never compile, and the hybrid-rebuilt S3 core also lacks the arduino_tinyusb component, which breaks USBMSC/USB on the X4 Pro.

Move the sdkconfig block to a [hybrid] section and have envs opt in explicitly with custom_sdkconfig = ${hybrid.custom_sdkconfig}. The resolved configuration is unchanged for every env except x4pro, which now stays on the prebuilt core (with TinyUSB) as it must.

Pairs with #3242

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -70,41 +70,11 @@ board_build.flash_mode = dio
 board_build.flash_size = 16MB
 board_build.partitions = partitions.csv
 
-; MEMFIX-PORT: custom_sdkconfig heap reclamation (~32-37 KB). Rebuilds the
-; Arduino core libs on first build (slower once, cached after; needs the CMake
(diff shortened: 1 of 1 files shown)


itsthisjustin:
Closing in favor of [#usb host](https://github.com/crosspoint-reader/crosspoint-reader/pull/3203)

PR #3247, opened 2026-08-28
fix: apply viewable insets to top and bottom bars

itsthisjustin:
Centralize the board viewable-inset math (bezel / rounded-corner clearance from BoardConfig::ACTIVE.viewableInsets) in a single UITheme::getContentArea, oriented to the current rotation, and derive getScreenSafeArea from it. Headers, status bars, and button hints previously applied insets only on the sides, so panels with top/bottom bezel overlap (recessed or rounded-corner glass) clipped the bars.

Align Home and keyboard touch targets with the inset-adjusted layout so taps land on what is drawn, and fix the WiFi network-list header, which passed the already-inset safe area into drawHeader (which self-insets) and double-inset on bezel panels.

### DIFF
--- a/src/activities/home/HomeActivity.cpp
+++ b/src/activities/home/HomeActivity.cpp
@@ -293,6 +293,28 @@ void HomeActivity::render(RenderLock&&) {
   // self-insets in drawHeader, so only the cover/menu need it here.
   const Rect content = UITheme::getInstance().getContentArea(renderer);
 
+  // Record the tile rect so storeCoverBuffer (called from the theme) knows which
+  // sub-region of the framebuffer to snapshot. ~16 KB in Portrait instead of the
+  // 48 KB full framebuffer the previous bind captured. Assigned before the
+  // restore below because a stored buffer is only valid for the geometry it was
+  // captured at: a theme/orientation change (the control center can move
+  // SETTINGS.orientation while Home sits on the stack and re-renders) shifts the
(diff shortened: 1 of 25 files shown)


Uri-Tauber:
`drawHeader(Rect)` now adds horizontal insets internally, while at least five existing callers already pass `getScreenSafeArea() rectangles` — and that function now includes the same insets. The PR repairs one caller (`WifiSelectionActivity`) but leaves the ownership rule split.

Uri-Tauber:
Let me see if I can fix this myself.

itsthisjustin:
> `drawHeader(Rect)` now adds horizontal insets internally, while at least five existing callers already pass `getScreenSafeArea() rectangles` — and that function now includes the same insets. The PR repairs one caller (`WifiSelectionActivity`) but leaves the ownership rule split.

Ok I did try to fix this only in places where I saw issues to avoid breaking something but if you want to clean it up, you won't get any argument out of me. FYI there's a couple open PRs around insets on x4 pro as it turns out, a lot of these devices apparently had their display installed slightly differently

Uri-Tauber:
@itsthisjustin I believe #570 and #618 are relevant here.

itsthisjustin:
> @itsthisjustin I believe #570 and #618 are relevant here.

Yeah I don't think this will fix it because the displays are all different. However it does allow us to eventually just add a display alignment setting that lets you set these offsets manually once everything reads from them

PR #3250, opened 2026-08-28
feat: add non-allocating urlOrigin() utility

sfoulad:
## What

Adds `urlOrigin()`, a small header-only utility (`lib/UrlOrigin/UrlOrigin.h`)
that extracts the `scheme://host[:port]` prefix from a URL as a
`std::string_view` over the caller's own buffer -- allocating nothing.

## Why

Intended for a redirect loop that needs to compare origins across hops --
e.g. to decide whether to keep forwarding a sensitive header like
`Authorization` when a request follows a redirect to a different host (see
`runGet()`/`runGetWolf()` in `src/network/HttpDownloader.cpp`).

The non-allocating property specifically matters because origin comparison
can run on every redirect hop, including ones immediately before a TLS
handshake that needs a large contiguous heap block for RSA signature
verification -- an extra allocation in that exact window is itself a
fragmentation risk on RAM-constrained targets (this codebase runs on
ESP32-C3 with ~380KB usable RAM and no PSRAM).

## Scope of this PR

This PR only adds the utility and its test coverage -- there is no caller in
this repo yet. Landing it standalone first means the string-parsing logic
gets its own focused review and test coverage before a follow-up PR wires it
into an actual redirect-loop origin check, rather than bundling both changes
together.

## Design notes

- A view into the caller's own buffer; never allocates.
- Handles the edge case where a URL has no path but does have a query string
  or fragment (e.g. `https://host?key=value`) -- stopping at the first `/`
  alone would fold the query string into the "origin" and misclassify a
  same-origin redirect as cross-origin. `find_first_of("/?#", ...)` avoids
  this.
- Not null-terminated (it's a `string_view`) -- documented in the header so
  callers don't pass it across a C API boundary.

## Testing

- New host test suite (`test/url_origin/UrlOriginTest.cpp`, GoogleTest):
  10/10 passing locally, including a differential test against a
  `std::string`-allocating reference implementation across every case to
  guard the heap-free rewrite against silently changing origin-comparison
  semantics.


arthurgiroux (commented):
I see that you mention in the description of the PR that it's submitted as a separate PR to make sure it's reviewed independently but I don't think it's the best approach.
I'd rather have everything related to the same feature in the same PR, split in commits that can be reviewed independently, and have a comment in the PR to mention that the review should be made commit-wise or tag the specific commit.
When looking at just this PR it can't be merged on its own, it's also not clear if it's mandatory or not for the other PR and it's asking the reviewer to merge or rebase branches to test it which is not ideal.
I'll let the main contributor pitch in if they disagree.

PR #3251, opened 2026-08-28
chore: pin third-party GitHub Actions to commit SHA

sfoulad:
## What

Pins every third-party GitHub Action reference in `.github/workflows/*.yml`
to the exact commit SHA its floating major-version tag (`@v4`, `@v6`, `@v7`)
currently resolves to, with the version kept as a trailing comment. This
matches the pattern the repo already uses for `step-security/harden-runner`
(`uses: step-security/harden-runner@ec9f2d5... # v2.13.0`), just extended to
the remaining actions that didn't have it yet.

## Why

A floating tag can be repointed by the action's maintainer -- intentionally
in a bad release, or via a compromised publishing account -- to different
code than what was reviewed when the workflow was last touched, with no
change to this repo triggering re-review. GitHub's own Actions security
hardening guidance recommends pinning third-party actions to a full-length
commit SHA for exactly this reason. Pinning to SHA closes that gap while the
trailing version comment keeps the intended version readable and diffable on
future bumps.

## Scope

Only the `uses:` lines for actions that were still on a floating tag:
`actions/checkout`, `actions/setup-python`, `actions/cache`,
`actions/upload-artifact` (both `@v4` and `@v6` call sites), `astral-sh/setup-uv`,
and `amannn/action-semantic-pull-request`. Each SHA was resolved directly from
GitHub's API against the tag it replaces (`repos/{owner}/{repo}/git/ref/tags/{tag}`),
not guessed or hand-typed.

No behavior change -- every SHA is the exact commit its corresponding tag
currently points to, so this run's outputs are identical to before.

## Testing

- `actionlint`: clean (one pre-existing shellcheck warning on an unrelated,
  untouched line in `release_candidate.yml`).
- All five workflow files parse as valid YAML (`python3 -c "import yaml; ..."`
  per file).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,11 +12,11 @@ jobs:
   clang-format:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v6
+      - uses: actions/checkout@d23441a48e516b6c34aea4fa41551a30e30af803 # v6.1.0
         with:
           submodules: recursive
 
-      - uses: actions/setup-python@v6
+      - uses: actions/setup-python@ece7cb06caefa5fff74198d8649806c4678c61a1 # v6.3.0
         with:
           python-version: '3.13'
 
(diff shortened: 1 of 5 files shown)


arthurgiroux (approved):
All the commit hash are valid and point to the version that is mentioned in the comments.

PR #3255, opened 2026-08-28
fix: reduce getSettingsList() lambda stack usage

sfoulad:
## Summary

`SettingsList.h`'s `getSettingsList()` builds its ~58-entry `SettingInfo` vector via aggregate initialization. Since each `SettingInfo` carries several `std::function`/`std::vector` members, the aggregate-init list forces the compiler to materialize every entry's temporary on the stack simultaneously before copying into the vector -- peak stack usage scales with the whole list, not one entry.

This PR reserves capacity up front and `push_back`s entries one at a time instead, so only one temporary is alive (moved in, then destroyed) at any point.

## Verification

- `-fstack-usage`: the lambda's stack frame drops accordingly.
- A token-sequence comparison confirms the transformation is behaviorally identical to the original aggregate list -- same entries, same order, same `#if`/`#else`/`#endif` guards, same `.withTextSettings()` chains.
- `pio run -e default` and `pio run -e x4pro`: both succeed.
- `pio check -e default --skip-packages` (cppcheck): no defects.

## Test plan

- [x] Build `default` and `x4pro` environments
- [x] cppcheck clean
- [x] clang-format clean
- [ ] CI green

### DIFF
--- a/src/SettingsList.h
+++ b/src/SettingsList.h
@@ -215,236 +215,250 @@ inline std::vector<SettingInfo> getSettingsList(const SdCardFontRegistry* regist
     statusBarClockValues[CrossPointSettings::STATUS_BAR_CLOCK_RIGHT] = StrId::STR_DIR_RIGHT;
     statusBarClockValues[CrossPointSettings::STATUS_BAR_CLOCK_LEFT] = StrId::STR_DIR_LEFT;
 
-    std::vector<SettingInfo> v = {
-        // --- Display ---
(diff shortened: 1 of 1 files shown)


sfoulad:
Status update: CI is green and there are no outstanding review findings. This is a small standalone fix — the lambda table in `getSettingsList()` burns stack that's tight on the C3 default task size (we hit a real boot-loop from it downstream); we carry the same change as a local patch that conflicts whenever the settings list grows upstream. Requesting a quick review; it's mechanical and behavior-preserving.

Issue #3259, opened 2026-08-28
Home menu is cut off in landscape and cannot be reached


winst0niuss:
### Affected Version

`develop` @ 86976ab2 (also reproduces on the current `develop` HEAD)

### Describe the Bug

The home screen lays the menu out below a cover tile whose height is a fixed theme metric that does not depend on the screen height — `homeCoverTileHeight` in [src/components/themes/BaseTheme.h](https://github.com/crosspoint-reader/crosspoint-reader/blob/develop/src/components/themes/BaseTheme.h) (400 for the Classic metrics, 350 for RoundedRaff), consumed as-is in `HomeActivity::render()`:

```cpp
coverRectH = metrics.homeCoverTileHeight;
GUI.drawRecentBookCover(renderer, Rect{0, metrics.homeTopPadding, pageWidth, metrics.homeCoverTileHeight}, ...);
GUI.drawButtonMenu(renderer, Rect{0, metrics.homeTopPadding + metrics.homeCoverTileHeight + metrics.homeMenuTopOffset, ...
```

In portrait the panel is 800 px tall and everything fits. In landscape it is only 480 px tall, but the tile keeps its portrait height, so the menu block starts far too low: the last entries (File Transfer, Settings, and OPDS Browser when servers are configured) are pushed past the bottom edge of the screen.

The home screen has no scrolling, so there is no way to bring those entries back into view.

### Steps to Reproduce

1. Open an EPUB with a cover so the "continue reading" tile is shown on the home screen.
2. Open the quick-settings panel and switch the orientation to Landscape CW (or CCW).
3. The home menu is cut off at the bottom edge — the lower entries are not drawn.
4. There is no way to scroll them into view.

### Expected Behavior

The cover tile is sized against the available screen height (or the menu is made scrollable), so every home menu entry stays reachable in all four orientations.

### Relevant Log Output/Screenshots

Reproduced in the desktop simulator with the X4 Pro profile; not yet checked on hardware, but the layout metrics are device-independent.

winst0niuss:
I'll fix this

winst0niuss:
Fixed by 4f0b4b7 (#3264): the quick-settings orientation tile no longer turns the renderer, so the home screen can no longer end up in landscape.

Note for the future: the cover tile height is still a fixed theme metric, so if another path ever renders home in a landscape frame, that layout will need to be sized against the available screen height.

PR #3264, opened 2026-08-28
fix: keep the home screen in portrait

### LINKED ISSUE #3258: Home screen cover art is corrupted after changing orientation
### Affected Version

`develop` @ 86976ab2 (also reproduces on the current `develop` HEAD)

### Describe the Bug

On the home screen the "continue reading" cover is drawn from SD only once. After the first draw, `HomeActivity` snapshots that framebuffer region and restores the raw pixels on every later render instead of re-drawing the bitmap:

- `HomeActivity::storeCoverBuffer()` / `restoreCoverBuffer()` — [src/activities/home/HomeActivity.cpp#L135-L159](https://github.com/crosspoint-reader/crosspoint-reader/blob/develop/src/activities/home/HomeActivity.cpp#L135-L159)
- the tile rect is recomputed from `getScreenWidth()` on every render — [src/activities/home/HomeActivity.cpp#L300-L303](https://github.com/crosspoint-reader/crosspoint-reader/blob/develop/src/activities/home/HomeActivity.cpp#L300-L303)

Nothing invalidates that snapshot when the screen orientation changes. `coverRendered` and `coverBufferStored` are cleared only in `freeCoverBuffer()`, which runs from `onExit()` and from `loadRecentCovers()` — never on an orientation change.

So after switching Portrait -> Landscape CW, a snapshot captured at the portrait width is restored into a region of the landscape width. The row stride no longer matches, the image shears diagonally across the tile, part of it spills past the left edge of the tile, and the title/author text is drawn on top of the garbage.

The cover stays corrupted until something forces a full reload of the recent-books list.

### Steps to Reproduce

1. Open an EPUB that has a cover, so it becomes the "continue reading" entry on the home screen.
2. Go back to the home screen and let the cover render.
3. Open the quick-settings panel.
4. Tap the orientation button (Portrait -> Landscape CW).
5. The cover tile now shows sheared/duplicated image data instead of the cover.

### Expected Behavior

The cover snapshot is discarded when the orientation (or any screen geometry) changes, and the cover is re-rendered at the new geometry.

Reproduced in the desktop simulator with the X4 Pro profile; not yet checked on hardware, but the caching logic is device-independent.

winst0niuss:
Fixes #3258, #3259.

### The bug

The orientation tile in the quick-settings panel cycles `SETTINGS.orientation`, which is the *reading* orientation — it lives under the Reader category in `SettingsList.h`, and `ReaderActivity::onExit()` restores portrait when a book is closed. Every other screen is laid out for portrait.

The panel, however, turns the renderer for whatever screen sits under it. Cycling the orientation from the home screen therefore leaves the home screen itself in landscape, and it does not fit there: the cover tile is a fixed theme metric that does not depend on the screen height (400 for Classic, 350 for RoundedRaff, 300 for Lyra Extended, 242 for Lyra). In portrait's 800 px everything fits; in landscape's 480 px the menu block starts too low and the lower entries — File Transfer, Settings, and OPDS Browser when servers are configured — are pushed past the bottom edge. The home screen has no scrolling, so there is no way to bring them back.

### The fix

Have the home screen claim portrait when it paints, the same way `ReaderActivity`, `XtcReaderActivity` and `SleepActivity` already claim theirs. The reading orientation keeps working where it belongs — it is still saved, and the next book still opens turned.

It has to happen in `render()` rather than `onEnter()`: `ActivityManager::Pop` repaints the screen under a dismissed panel without calling `onEnter()`, which is the exact path this bug takes.

An earlier attempt made the panel skip the turn unless a reader was underneath. That was dropped: it also took the turn away from screens that handle landscape fine — the file browser lays out correctly in landscape — and it made the tile look inert, since nothing on screen moved when it was tapped.

### Verification

Built for `default` (ESP32-C3) and exercised in the desktop simulator on the X4 Pro profile, theme Lyra Extended (three covers — the tightest case).

- Panel over the home screen, tap the orientation tile: the panel dismisses as before, the setting advances and is saved, and the home screen repaints in portrait, complete.
- Open a book with the setting on Landscape CW: the reader opens turned, unchanged.
- Panel over the file browser, tap the tile: the file browser still turns — no capability removed from screens that handle it.

Not yet verified on hardware.

### Before / after

Home screen after cycling the orientation to Landscape CW, theme Lyra Extended. The "before" shot is taken on top of #3260 so the separate cover-cache bug does not obscure the cropped menu.

| `develop` | this branch |
|---|---|
| home screen cropped in landscape | home screen stays in portrait |

### DIFF
--- a/src/activities/util/FrontlightPanelActivity.cpp
+++ b/src/activities/util/FrontlightPanelActivity.cpp
@@ -12,7 +12,6 @@
 
 #include "CrossPointSettings.h"
 #include "MappedInputManager.h"
-#include "activities/reader/ReaderUtils.h"
 #include "components/UITheme.h"
 #include "components/UIThemeTokens.h"
 #include "components/icons/customListIcons.h"
@@ -161,17 +160,9 @@ void FrontlightPanelActivity::runTile(const int idx) {
     case 2:  // Cycle the reading orientation
       SETTINGS.orientation = static_cast<uint8_t>((SETTINGS.orientation + 1) % 4);
       SETTINGS.saveToFile();
-      // Nothing else would turn the renderer: ActivityManager::Pop restores the
-      // activity underneath without calling onEnter(), so its
-      // applyInitialOrientation() never runs. Apply it here instead.
-      ReaderUtils::applyOrientation(renderer, SETTINGS.orientation);
-      // Close rather than redraw in place: the turned panel lays out against
-      // the new frame while the old panel's pixels sit in the old frame, so
-      // repeated taps stacked stale panels on screen. The screen underneath
-      // repaints in the new orientation and its refresh carries the cleanup a
-      // whole-frame rewrite needs.
-      renderer.promoteNextRefresh(HalDisplay::FULL_REFRESH);
-      close();
+      // Only the setting changes: turning the renderer cropped the portrait-only
+      // screens the panel opens over. The reader reflows on its next loop().
+      requestUpdate();
       break;
     case 3:  // Touch reader controls (for reading with the palm on the glass)
       // Toggles the existing Settings -> Controls option, nothing lower-level:


itsthisjustin:
This is the actual fix for #3258 and I'd want to do this instead. That's a bug in the new panel that isn't supposed to be rotating the entire UI. I never noticed as I only tested in the reader.

winst0niuss:
Agreed, keeping this one as the fix for both  #3258, #3259

winst0niuss:
@itsthisjustin, check please

itsthisjustin:
I'm assuming I know the answer to this but shouldn't it just set the reader setting and save it? Probably breaks automatically changing orientation in reader though. I'm just concerned this fix won't work on every single other screen this new panel is available from

winst0niuss:
You're right. The tile now only writes the setting and no longer turns the renderer, so nothing outside the reader can end up in a layout it cannot handle.

The reader still turns: `EpubReaderActivity::loop()` already reflows when the applied orientation differs from the setting, so it picks the change up once the panel is dismissed. The panel repaints instead of closing, so the tile label shows the new mode right away.

The `HomeActivity` change is dropped. Pushed; updating the title and description next.

itsthisjustin (approved):

PR #3267, opened 2026-08-28
fix: dismiss the quick-settings sheet with a swipe up

### LINKED ISSUE #3266: Quick-settings sheet is hard to dismiss in landscape
### Affected Version

`develop` @ 86976ab2

### Describe the Bug

The quick-settings sheet is dismissed by tapping below it — `FrontlightPanelActivity::loop()` closes on a release with `touchY >= panelBottom`. In portrait that leaves roughly 345 px of free screen under the sheet, so it is easy to hit.

In landscape the sheet keeps its height while the screen is only 480 px tall, so the strip left under it is about 25-30 px — a target far below any comfortable touch size. In practice you have to aim at the few pixels between the sheet's bottom rule and the screen edge, or dismiss the sheet by pressing "Refresh Screen" (which closes it as a side effect of doing something else).

The sheet does draw a grabber — the short bar centred at its bottom edge, the usual affordance for pulling a sheet away — but it is inert. `computePanelBottom()` counts the grabber band as part of the sheet, so `panelBottom` sits *below* the grabber and a tap on it is treated as a tap inside the sheet.

Touch-only boards feel this most: on a board with hardware buttons, Back also dismisses the sheet.

### Steps to Reproduce

1. On a touch board (X4 Pro), set the reading orientation to Landscape CW or CCW.
2. Open a book so the reader is on screen in landscape.
3. Swipe down from the top edge to open the quick-settings sheet.
4. Try to dismiss it by tapping: only the thin strip between the sheet and the bottom edge works, and the grabber does nothing.

### Expected Behavior

Tapping the grabber dismisses the sheet, the way its shape suggests. That keeps a finger-sized target in every orientation without changing the sheet's layout.

Quick-settings sheet over the reader in landscape — note how little room is left below it (the reader's status bar is still visible under the sheet's bottom rule):

quick-settings sheet in landscape leaves almost no room below it

winst0niuss:
Fixes #3266.

### The bug

`FrontlightPanelActivity::loop()` dismisses the sheet on a release with `touchY >= panelBottom`, and `computePanelBottom()` counts the grabber band as part of the sheet. So the only way out is the screen *below* the sheet, and the gesture the grabber advertises, pulling the sheet back up the way it was pulled down, does nothing.

In portrait that is invisible: roughly 345 px of free screen sit under the sheet. In landscape the sheet keeps its height on a 480 px screen and the strip shrinks to about 25 px, well under a comfortable touch target. Dismissing it means aiming between the sheet's bottom rule and the screen edge, or pressing "Refresh Screen" to close it as a side effect. Boards with hardware buttons still have Back; touch-only boards do not.

### The fix

Close the sheet on an up-swipe, which is what every phone answers to and what the grabber advertises. Six added lines in `loop()`, no layout change: the sheet, the tiles and the grabber are drawn exactly as before, and the existing release-below-the-sheet path keeps working.

Two details make the gesture safe to accept anywhere on screen:

* a swipe release is reported off-target (`touchY = -1`), so nothing under the finger is activated on the way out;
* `draggingSlider` is already set on the held frames of a slider drag, so a fast flick on the brightness or warmth pill never reads as a dismissal.

The panel is opened by a top-edge *down*-swipe, so the opening gesture cannot close it in the frame it appears in. On boards without a Home key the bottom-edge up-swipe already closed the panel through `handleHomeGesture()`; the behaviour is now the same on every board.

### Verification

Built for `default` (ESP32-C3) and exercised in the desktop simulator on the X4 Pro profile, in landscape (the tight case):

- swipe up over the sheet, from the grabber and from the sheet body alike, and it dismisses;
- drag the brightness slider and release, the sheet stays open, `draggingSlider` doing its job;
- tap below the sheet, it still dismisses, unchanged from `develop`;
- the same swipe on `develop`, the sheet stays up, which is the bug.

Not yet verified on hardware.

### Screenshot

The sheet over the reader in landscape, the grabber is the short bar at the bottom, and the strip under the sheet is all that used to be tappable:

quick-settings sheet in landscape

### DIFF
--- a/src/activities/util/FrontlightPanelActivity.cpp
+++ b/src/activities/util/FrontlightPanelActivity.cpp
@@ -240,6 +240,12 @@ void FrontlightPanelActivity::loop() {
       if (touch.event.dragPermille >= 0) draggingSlider = true;
       return;
     }
+    // Swipe up dismisses the sheet, the way it was pulled down from the top
+    // edge. draggingSlider keeps a fast slider flick from closing it.
+    if (!draggingSlider && mappedInput.wasSwipe() == MappedInputManager::SwipeDir::Up) {
+      close();
+      return;
+    }
     // panelBottom > 0 guards the frame the sheet opens in: the release that
     // opened it (a status-bar tap) is still in the input snapshot when the panel
     // runs its first loop(), and panelBottom is only known once render() has


unnamedd (changes requested):
Hi @winst0niuss, first of all, thanks for your PR.

I was testing your fix locally and on a real X4 Pro, and I do think we have a half-baked implementation here.

Regarding the real user experience, I do think we need to make a small improvement here.

When we turn the reader to landscape mode and pull the Frontlight Panel down from the top, we see it being presented. However, based on your implementation, even though it is nice to be able to tap the grabber to dismiss the panel, I really do think this goes against the most intuitive user experience found in most mobile operating systems —and in ours as well— where you simply swipe it up to dismiss it.

What I would expect here is to swipe up and dismiss it. This would be my first instinct.

I really think that implementing just a tap, is not enough to have a full and good experience.

Let me know what you think.

winst0niuss:
Hi, @unnamedd!
I agree with you. 
I tried it myself and came to the same conclusion: tap-to-dismiss isn't intuitive. The grabber looks like something you drag, not tap, so nobody would guess the gesture. Swipe-up is what feels natural here.

I'm reworking the PR to dismiss the panel with a swipe-up. I'll ping you when it's ready.

unnamedd:
Awesome! The points I raised here were basically because I misread your verification section when you mentioned “tap.” I was dragging it up, and it wasn’t working. So, my first instinct didn’t work, but when I read it again more carefully, I saw your observation.

winst0niuss:
@unnamedd, ready for review and verify on hardware

unnamedd (approved):
It looks good to me. Thanks for the fix and simplification of the code.

PR #3274, opened 2026-08-28
feat: optional Settings extension provider for downstream integrations

sfoulad:
## Problem

A downstream fork that wants to add its own Settings tab, rows, or actions
today has to patch `SettingsActivity.cpp`'s hardcoded 4-category switch and
extend `SettingAction` for every product-specific action -- there's no
extension point. Several small forks/companion apps building on CrossPoint
have run into this; the current answer is "diverge from `SettingsActivity`
forever," which defeats the point of building on a shared reader.

## What this adds

A single optional hook a downstream integration can register before
`SettingsActivity` is entered:

```cpp
struct SettingsExtensionCategory {
  std::string label;                  // pre-localized by the provider
  std::vector<SettingInfo> settings;  // reuses the existing row model
};
using SettingsExtensionProvider = std::vector<SettingsExtensionCategory> (*)();

void setSettingsExtensionProvider(SettingsExtensionProvider provider);
```

`SettingsActivity` calls the provider once per rebuild (the same cadence it
already uses to rescan fonts/dictionaries) and appends whatever categories
come back after the 4 built-in tabs. **Unset (the default) is a single
null-pointer check** -- zero extra categories, zero extra rows, and the
existing 4-tab behavior is unchanged.

Two smaller additions round out what a contributed row can do, both
mirroring patterns `SettingInfo` already has for `ENUM`/`STRING`:

- `SettingInfo::DynamicToggle` / `DynamicValue` -- `TOGGLE`/`VALUE` rows
  backed by an external getter/setter, the same shape `DynamicEnum`/
  `DynamicString` already use for settings that don't live in
  `CrossPointSettings` (e.g. `KOReaderCredentialStore`).
- `SettingAction::Extension` + `SettingInfo::actionHandler` -- one generic
  enum value that dispatches to a caller-supplied `std::function<void()>`,
  so a downstream integration can add any number of action rows without a
  new `SettingAction` case per feature.
- `SettingInfo::withLabel()` -- a row label override for contributed rows,
  which have no compiled-in `StrId` to point at.

No product names, branding, or downstream-specific logic anywhere in this
diff -- it's a storage slot and a dispatch case, both structurally identical
to patterns already in this file.

## Why both button and touch work automatically

Contributed rows are plain `SettingInfo` values flowing through the exact
same `buildScreen()`/`toggleCurrentSetting()`/`settingValueText()` code
paths as every built-in row, so they get the FUI list rendering, tab bar,
and button/touch input handling for free -- no separate rendering path per
input mode.

## Refactor note

`SettingInfo`/`SettingType`/`SettingAction`/`SettingsExtensionCategory`
moved out of `SettingsActivity.h` into a new `SettingsTypes.h`. Pure
relocation, no field or behavior changes -- it decouples the settings data
model from `SettingsActivity`'s `GfxRenderer`/`MappedInputManager`/
`UiTabListActivity` dependencies, which is what makes the new host test
possible (this file previously had zero test coverage, since the model was
inseparable from the Activity).

## Testing

- `test/settings_extension/SettingsExtensionTest.cpp` (8 cases, host-run
  via the existing gtest harness): provider defaults to unset and round-
  trips through set/clear; a synthetic provider proves 2 categories with
  toggle/value/action rows all come back correctly typed; direct tests of
  `DynamicToggle`'s invert semantics and `DynamicValue`'s wrap-on-overflow
  step, matching `SettingsActivity::toggleCurrentSetting()`'s existing
  inline logic exactly.
- `pio run` on all 4 boards (`default`, `x4pro`, `sticky`, `papermono`):
  all SUCCESS.
- `pio check -e default --skip-packages` (cppcheck) on the changed files:
  no defects found.
- `./bin/clang-format-fix`: clean.

## Scope

No product-specific code. This only adds the extension point; nothing in
this repo registers a provider (the default -- and current -- behavior of
every existing category/row is unaffected).


sfoulad:
Addressed in cd6c11bf:

1. **`stepTab()` used `categoryCount` instead of `tabCount()`** — confirmed real bug, fixed. Button-driven Confirm-on-tab-bar cycling would wrap 0→1→2→3→0 and never reach a provider's contributed tabs, even though the touch tap-on-tab path (`onTabAction`/`selectCategory`) already handled the full range correctly. Good catch.

2. **`(cur + 1) % totalValues` modulo-by-zero in the dynamic-ENUM branch** — confirmed real (UB, crash risk if `totalValues` is 0, e.g. an empty `DynamicEnum` or exactly 256 values truncating through the `uint8_t` cast). This is pre-existing upstream code this PR didn't introduce and none of this PR's own rows can trigger it, but it's in a block this PR already touches, so I added a one-line guard rather than leaving it.

3. **Heap churn in `rebuildSettingsLists()` reassigning `extraCategories` on every rebuild** — leaving as-is, for two reasons: (a) it's zero-cost in the byte-identical/no-provider case (`extraCategories.clear()` on an already-empty vector allocates nothing), which is the case for every existing CrossPoint user; (b) when a provider *is* registered, this matches the exact pattern the same function already uses one line below for `dictionaries` (`std::vector<DictionaryEntry> dictionaries; DictionaryRegistry::discover(dictionaries);` — a fresh SD-directory-listing rebuild on every call) and for `sdFontSystem.refreshIfDirty()`. Introducing a non-owning-view/static-pool storage scheme here would be inconsistent with the surrounding code and disproportionate to the actual cost (one vector realloc vs. an SD directory scan in the same function). Happy to revisit if a maintainer prefers otherwise.

All 4 board builds (default/x4pro/sticky/papermono), the host test suite, cppcheck, and clang-format all re-verified green after these changes.

sfoulad:
Good catch, fixed in 64bc46a4 -- kept the count as \`size_t\` through the comparison and modulo instead of truncating through \`uint8_t\` up front, so an exactly-256-entry \`DynamicEnum\` cycles correctly instead of silently going inert. Cast only happens at the final \`valueSetter\` call.

Re-verified: \`pio run -e default\` SUCCESS.

sfoulad:
Re the outside-diff finding on `SettingsActivity.cpp` lines 501-503 ("Avoid `std::string` allocation in the value-rendering path") — I looked at this one and I'd rather not act on it in this PR. Three reasons, all checkable against `develop`:

**1. It's the pre-existing pattern, not something this PR introduces.** `settingValueText()` returns `std::string` in `develop` today (`SettingsActivity.cpp:427` at b42927b4), and `develop`'s own line 458 is:

```cpp
return std::to_string(SETTINGS.*(setting.valuePtr));
```

That line is byte-identical to line 499 in this PR — unchanged, same function. The new lines 501-503 are the callback-backed mirror of it:

```cpp
if (setting.type == SettingType::VALUE && setting.valueGetter) {
  return std::to_string(setting.valueGetter());
}
```

Deviating here would make the extension path behave differently from the built-in path two lines above it, for no user-visible gain.

**2. In practice this usually isn't an allocation at all.** The result lands in `rowValues_[i]`, a `std::vector<std::string>` element that `buildScreen()` deliberately *assigns into* to reuse its existing capacity (`SettingsActivity.cpp:478`, and the comment above it: "assigning into the existing strings (no vector growth)"). The values are short numerics, so `std::to_string` stays inside libstdc++'s SSO buffer and the assignment reuses the destination's storage. There's no per-repaint heap traffic here to remove.

**3. Removing it genuinely would mean changing `settingValueText()`'s return type** and every call site — upstream code this PR otherwise doesn't touch. That's a reasonable cleanup on its own merits, but it isn't the extension point, and folding it in would push this past the "smallest possible generic hook" shape I was aiming for. Happy to open it separately if you'd like it done.

For the record, the three inline findings from the first pass were all real and are fixed: `stepTab()` passing `categoryCount` instead of `tabCount()` (so button-driven navigation couldn't reach contributed tabs) and the modulo-by-zero in the dynamic-ENUM branch, both in cd6c11bf; the exactly-256-entry enum truncating through `uint8_t`, in 64bc46a4.

sfoulad:
Status update: all four CodeRabbit review rounds are addressed (two real fixes — the `stepTab()` tab-count bug and the 256-entry DynamicEnum truncation — plus one out-of-diff suggestion declined with reasoning in-thread). CI is green across all four build targets, cppcheck, and unit tests. Downstream we currently carry SettingsExtension/SettingsTypes/SettingsActivity as local patches that conflict on every upstream sync; this PR is the hook that lets us drop them. Requesting a maintainer review of the extension-point API shape — happy to adjust naming or narrow the surface if preferred.

sfoulad:
**Follow-up pushed: four defects in this PR's own code, all from the tab band no longer having a fixed size.**

This PR re-consults the provider on every rebuild so its categories stay current (sign-in/sign-out). That makes `tabCount()` a moving target, and four places assumed it wasn't.

1. **Out-of-bounds per-tab nav state (blocker).** `UiTabListActivity::onEnter()` sizes `tabNavs` from `tabCount()` as its first statement, but `SettingsActivity::onEnter()` calls it *before* `rebuildSettingsLists()` has run the provider — and never resizes afterwards. Selecting an extension tab then indexed `tabNavs` past its end: an out-of-bounds heap write with `NDEBUG`, an abort otherwise. `rebuildSettingsLists()` now resizes `tabNavs` to the current band on every rebuild, so the base's `tabNavs.size() == tabCount()` invariant holds for a provider that appears, grows or shrinks at runtime. No change to `UiTabListActivity` — `tabNavs` is already protected for exactly this kind of subclass management.

2. **`ExtensionAction()` left `nameId` uninitialized (blocker).** `SettingInfo::nameId` had no default member initializer and `ExtensionAction()` is the one factory that never assigns it. `SettingsActivity` compares `nameId` against specific `StrId`s for *every* row, so stack garbage matching a built-in row would route an extension row into that built-in's screen. Every member of `SettingInfo` (and `ValueRange`) now carries a default; `SettingInfo` is const-default-constructible and the test asserts that at compile time, so it can't regress.

3. **Confirm hint named the wrong tab.** `stepTab()` was updated to `tabCount()`; the hint in `render()` still wrapped over `categoryCount` and read the built-in `categoryNames[]` table, so on the last two tabs it named a tab Confirm did not go to. It now uses `ButtonNavigator::nextIndex(..., tabCount())` and `tabLabel()` — the same wrap and label source as the band.

4. **Extension rows never refreshed after their action.** The `Extension` case called `rebuildSettingsLists()` immediately after the handler returned, but `startActivityForResult()` only sets `pendingActivity`: the rebuild ran before the child screen had done anything, so a row whose label depends on state the child changes stayed stale, and re-activating it re-ran the flow against already-changed state. (Same hazard the `Language` case works around.) New `runAfterExtensionAction()` chains the rebuild onto whatever result handler the action installed, and runs it immediately when the action opened no screen — the host stays ignorant of what any particular action does. It chains onto `Activity::resultHandler`, which `ActivityManager` clears once consumed, so the member is null exactly when nothing was opened.

Index arithmetic for the band (count, built-in/extension split, active-tab clamp, per-tab state sizing) is now a small aggregate `SettingsTabRange` in `SettingsTypes.h`, so the four consumers can't drift apart again.

`SettingsExtensionTest.cpp` grows from 10 to 23 cases covering all four, and the negatives were checked: removing the `nameId` initializer fails to compile, and restoring the old rebuild order fails two tests. Full host suite 406/406; four board builds green.

Uri-Tauber:
Five years ago, this would have add real value. Today, every one of the forks is heavily dependent on AI agents, which dont really care if they write 5 LoC or 500. This PR itself was written by agent (Opus 5, I guess, from the verbose comments).
I dont think we should add extra complexity here upstream for a benefit no-one is going to notice.

Uri-Tauber:
Closing as not planned.

PR #3280, opened 2026-08-29
feat: optional keyboard layer extension for downstream integrations

sfoulad:
## Summary

Let a downstream build add one extra keyboard layer without editing `KeyboardEntryActivity`.

`KeyboardEntryActivity`'s layers are fixed at compile time — the SDK's builtin letter/symbol tables plus the URL panel this activity defines for itself. A downstream integration that needs one more layer (a second script, say) currently has to edit this activity, which means carrying a permanent fork of a file that keeps moving.

Three optional function-pointer hooks in a new `activities/util/KeyboardExtension.h`, all null by default:

| Hook | Purpose | Unset behaviour |
|---|---|---|
| `KeyboardExtensionLayerProvider` | `const KeyboardLayout* (*)(bool shifted)` — the extra layer, or `nullptr` to decline that state | no extra layer exists |
| `KeyboardExtensionDefaultPredicate` | `bool (*)()` — should the keyboard *open* on it? | opens on letters, as today |
| `KeyboardExtensionLabelProvider` | `const char* (*)()` — the layer's own name for the Mode key | Mode key keeps its table label |

**With no provider installed — every stock CrossPoint build — nothing changes.** Each call site is a single null check, the Mode key keeps its two-stop `abc`/`?123` cycle, and the built-in keyboard behaves exactly as before. The diff to existing files is **+39 lines, 0 deletions**.

With a layer installed, the Mode cycle gains a third stop: `abc → ?123 → extra → abc`. The extra layer owns its whole grid including its bottom action row, exactly like the existing URL snippet panel — so a provider supplies its own Mode/Delete/OK keys and the activity needs no per-layer special casing.

## Design notes

- **Shaped like `SettingsExtension.h`** (#3274): plain function pointers, no virtual dispatch, no heap, no `std::function`. Consistent with the repo's guidance on template/`std::function` bloat.
- **`KeyboardExtension.h` forward-declares `freeink::ui::KeyboardLayout`** rather than including FreeInkUI. The hooks only traffic in pointers, so the header adds no include weight to the firmware — and it compiles on the host, which is what makes the hook directly testable.
- **A provider may decline a state** by returning `nullptr` (e.g. offering an unshifted layer only). `currentLayout()` then falls through to the built-in layers, so the keyboard can never be left with no layout to draw.
- **URL entry deliberately does not consult the hooks.** Its layers are EN-arranged tables for hosts and ports, where another script has nothing useful to contribute.
- **CrossPoint gains no knowledge of any specific script, language, or downstream project.** It stores three pointers and draws whatever string the label hook returns.

## Scope Check

CrossPoint is intentionally narrow, so this deliberately adds no feature and no UI of its own — it is an extension point and nothing else. No new setting, no new string, no new key, no behaviour change for any existing user. If you would rather this did not exist at all, that is a reasonable call and I will carry the divergence downstream instead; I am proposing it because the alternative is a permanently forked copy of a file that is actively evolving.

## Verification

- [x] Host tests: **177/177**, including 9 new `KeyboardExtensionTest` cases — unset-by-default, per-shift-state layouts, a provider declining a state, the three hooks being independent, and clearing back to stock
- [ ] Hardware: not exercised on device, since no stock build can reach the new branch (no provider ships with CrossPoint). The unset path is the one every device runs and it is unchanged.


sfoulad:
Status update: CI is green (all build targets, cppcheck, unit tests) and both CodeRabbit passes are resolved. This is the companion to #3274 — downstream we patch KeyboardEntryActivity directly to add an extra layer, which conflicts on sync; the optional extension hook here removes that divergence with no behavior change when unset. Requesting review of the hook placement in KeyboardEntryActivity — glad to rework the registration mechanism if you'd rather it mirror whatever shape #3274 settles on.

Uri-Tauber:
Closing as not planned.
See my comment in #3274

PR #3285, opened 2026-08-29
fix: mask the Wi-Fi password field

sfoulad:
## Summary

Stop drawing the Wi-Fi password in clear text.

`WifiSelectionActivity::promptPasswordEntry()` constructs the keyboard with `InputType::Text`, so every character of the network password is rendered visibly. `KeyboardEntryActivity` already implements masking for `InputType::Password` — this prompt just wasn't asking for it. Every other password prompt in the tree already passes `InputType::Password`; this one was the outlier.

1. The one-line fix at the call site: `InputType::Text` → `InputType::Password`.
2. The masking transform moves out of `displayTextForCurrentState()` into a pure function in a new `activities/util/PasswordMasking.h`, so the rule is testable on the host. The activity keeps ownership of all surrounding state (which `InputType` is active, whether the user toggled visibility, where the cursor sits) and delegates only the character substitution. **Behaviour is identical** — same loop, same reveal semantics.

## Why the extraction

A test that proves "the Wi-Fi prompt requests masking" would need to construct the activity, which needs `GfxRenderer`/`MappedInputManager` and isn't reachable from the host test suite. Extracting the ~8-line transform is the smallest change that makes the actual security property — *what gets drawn is not the secret* — assertable, without introducing any new abstraction or changing how the activity works.

If you'd rather this PR stay strictly one line, I'm happy to drop the extraction and the tests; say the word.

## Tests

6 cases in `test/password_masking/`:

- everything masked when nothing is revealed
- no original character survives anywhere in the output
- exactly the requested position revealed (first / middle / last)
- a reveal index past the end — what the activity passes in cursor mode — masks everything
- empty input stays empty
- multi-byte UTF-8 masks per byte, matching the caller's byte-wise cursor bookkeeping

## Scope Check

CrossPoint is intentionally narrow. This adds no feature and no UI: it corrects one argument at one call site and makes the existing rule testable. No new setting, no new string, no behaviour change for any non-password field.

## Verification

- [x] Host tests: **174/174**, including the 6 new cases
- [ ] Hardware: not exercised on device — the change is a constructor argument and a pure-function extraction, both covered above


sfoulad:
Addressed — thanks, both were fair.

The extra allocation was real and I introduced it: the in-place loop I replaced made one allocation, my const-ref version made two on every repaint of a password field. `maskPasswordText` now takes the string by value and masks in place, and the caller moves its existing buffer in. Behaviour and all 6 tests unchanged.

itsthisjustin:
We changed this on purpose actually due to UX issues with typing on these devices, especially long passwords. I think for a microreader masking the password is a useless endeavor also considering how infrequently folks enter passwords. Closing for now but feel free to open a discussion if this is something the majority of users want.

PR #3286, opened 2026-08-29
fix: set the system clock after Wi-Fi connect when it holds no valid date

sfoulad:
## Summary

Make HTTPS work after a cold boot on boards that cannot keep the date.

A board with no RTC — or one whose RTC only surfaces hour and minute — loses the calendar date across every reboot. ESP-IDF starts at epoch 0, and until something sets a plausible date, **certificate validation fails on every HTTPS request**: catalog fetches, update checks and font downloads all fail after a cold boot, with nothing on screen explaining why.

Nothing currently sets it. The existing post-connect sync is guarded by `halClock.isAvailable() && !SETTINGS.clockHasBeenSynced`, which:

- skips boards with **no RTC** entirely (`isAvailable()` false), and
- skips every board after its **first** successful sync (`clockHasBeenSynced` latched)

Neither covers the case that matters — a device that has an RTC-less or date-less clock and reboots.

1. Two device-independent helpers on `HalClock`: `isSystemTimeValid()` and `quickSyncSystemTime()`. `HalClock.cpp` already pulls `<WiFi.h>` and `<esp_sntp.h>`, so nothing new is introduced.
2. One call after a successful Wi-Fi connection, before the existing RTC sync.
3. `lib/hal/SystemTimePlausibility.h` — the plausibility rule as a pure predicate, so it's testable on the host without `HalClock.h`'s RTC/SNTP dependencies.

## Cost

The check is a single comparison; the ~5s SNTP wait is paid **only when the clock is actually invalid**. A board that already holds a good date is completely unaffected, and no connect gets slower in the common case — which is what makes this safe to run unconditionally rather than gating it on board type.

`quickSyncSystemTime()` returns false immediately if Wi-Fi isn't connected, so it's harmless if the ordering ever changes.

## Tests

5 cases in `test/system_time_plausibility/`: epoch 0 (the post-boot default) rejected; anything below the threshold rejected; the threshold itself accepted; an ordinary present-day reading accepted; and the constant pinned to the instant its comment documents, so it can't drift silently.

`quickSyncSystemTime()` itself is **not** unit-tested — it blocks on real SNTP and Wi-Fi state, and isolating that would need injection machinery this change doesn't otherwise justify. Flagging that rather than building scaffolding around it.

## Scope Check

CrossPoint is intentionally narrow. No new feature, no UI, no new setting or string — this fixes a failure mode that already exists on shipped boards. If you'd prefer the helpers live somewhere other than `HalClock`, or the call site sit elsewhere in the connect path, happy to move them.

## Verification

- [x] Host tests: **173/173**, including the 5 new cases
- [ ] Hardware: the cold-boot HTTPS failure this fixes was reproduced on a device; the fix itself is not yet exercised on one


sfoulad:
Addressed — thanks, both were fair.

The discarded bool is a fair smell, but I have kept the behaviour: `quickSyncSystemTime()` already logs its own failure, and a network that answered DHCP but not NTP is still a usable connection — failing Wi-Fi setup over it would leave the user worse off than the HTTPS state they would have had anyway. Marked the discard `(void)` with a comment saying why, rather than adding an error path that would block a working connection.

itsthisjustin:
I believe this will be handled in #3114

sfoulad:
Regarding overlap with #3114's `TrustedTime`: I checked its head (89ff1436). It calls `trustedtime::startSync()` after every station join, but that call is deliberately **non-blocking** (`lib/TrustedTime/TrustedTime.cpp` — it only kicks off `esp_sntp_init()` and returns), and its boot-time restore only helps once a "floor" has been persisted to NVS on a *previous* sync — on a first-ever cold boot (or after an NVS erase) the clock stays at epoch 0 until the async poll happens to complete. So the exact failure this PR targets — an HTTPS request (catalog, update check, font download) fired immediately after the first successful connect on an RTC-less board — can still race an unset clock under #3114. This PR closes that window by blocking (up to 5s, and only when `isSystemTimeValid()` is already false) so certificate validation is guaranteed a plausible date before the connect flow returns. TrustedTime's purpose is also different (a monotonic floor for loan-expiry enforcement); the two changes are complementary, and this one is independent of the much larger plugin-system PR landing.

sfoulad:
Correcting my own framing in this PR's description, against `develop` @ `790a0817`.

**The cert-validation failure I described does not occur upstream today.** `platformio.ini:41` sets `-DFREEINK_NET_WOLFSSL=1` in `[base]`, so every env takes the wolfSSL path, and that path calls `http.setInsecure()` (`HttpDownloader.cpp:75`, inside the `#if defined(FREEINK_NET_WOLFSSL)` block opened at line 66) — which reaches `wolfSSL_CTX_set_verify(ctx, WOLFSSL_VERIFY_NONE, nullptr)` in SecureNet. `notBefore`/`notAfter` are never checked, so an epoch-0 clock cannot fail a handshake. The `crt_bundle_attach` branch at `HttpDownloader.cpp:151` sits inside `#if !defined(FREEINK_NET_WOLFSSL)` (line 131) and is dead on every shipped env. My "reproduced on hardware" note came from a downstream build where verification is enabled; that was not a fair claim to make about this repo.

**The clock gap itself is real and, if anything, wider than the description says.** `HalClock::syncFromNTP()` returns at `HalClock.cpp:70` (`if (!_available) return false;`) before reaching `configTzTime()` at line 78, so a board with no RTC never issues an SNTP request at all. `configTzTime` at line 78 is the only clock-setting call in the tree — `git grep settimeofday -- lib/ src/` returns nothing, and nothing restores the system clock from the RTC at boot — so `time(nullptr)` is 0 on essentially every boot on every board.

That makes this a **prerequisite for removing `setInsecure()`**, not a fix for a user-visible failure today. I'm happy for it to be judged on those terms, or closed if that ordering isn't wanted.

On #3114 handling it: #3114 is still open, so nothing on `develop` covers this yet. Its hook is `trustedtime::startSync()`, which is non-blocking and opens with `if (esp_sntp_enabled()) return;` — and since `syncFromNTP()`'s `configTzTime()` leaves SNTP enabled, it no-ops on RTC boards in #3114's own ordering. The two also both drive SNTP (`syncNow()` calls `esp_sntp_stop()` and reconfigures). They merge without textual conflict, both hooks landing in `checkConnectionStatus()`, which is probably worth resolving before either lands.

No further pings from me on this one — happy to close it if you'd prefer the `setInsecure()` removal to come first as a single change.

PR #3296, opened 2026-08-29
feat: add touch navigation for EPUB links

Uri-Tauber:
## Summary

* **What is the goal of this PR?** make links and footnotes tappable on touch devices.


lib/Epub/Epub/Page.h
```diff
@@ -94,6 +97,21 @@ class Page {
     footnotes.push_back(entry);
   }
 
+  bool addLink(const char* href, int16_t x, int16_t y, int16_t width, int16_t height) {
```
arthurgiroux, line 101: It might be good to use strnlen and strncpy in this method.

Uri-Tauber: strncpy will pad the rest of the string with zeros. a bit less efficient.

PR #3298, opened 2026-08-30
feat: verify OTA TLS against pinned GitHub trust anchors

sfoulad:
## Summary

Make OTA actually verify who it downloads firmware from.

Both the release check and the firmware download run over the wolfSSL transport with `setInsecure()` — encrypted, but the peer is never verified. With no image signing in the release pipeline, transport authentication is the only integrity boundary OTA has; today a network attacker (rogue AP, DNS) can serve arbitrary firmware to every device that updates.

The existing wolfSSL transport is kept (its memory profile is what fits the C3 at OTA time). Verification is added only for the two hosts OTA talks to:

| Piece | What it does |
|---|---|
| `src/network/OtaCaCerts.h` (new) | Flash-resident anchors (~2.9 KB): **USERTrust ECC** (api.github.com → Sectigo ECDSA chain) + **ISRG Root X1** (release-assets.githubusercontent.com → Let's Encrypt, RSA-4096 cross-sign). Both chains probed live; fingerprints in the header. |
| `HttpDownloader` | `fetchUrl()` gains an optional `caPem`. Verified mode: `setCACert()` per hop, non-https refused up front, https→http redirect downgrade refused. **No bundle = exactly today's behaviour** for every existing caller. |
| `OtaUpdater` | Passes the anchors on the check and the download. Nothing else changes — user-configured OPDS/KOSync servers can't be pinned and keep their current behaviour. |
| `scripts/patch_wolfssl.py` | `WOLFSSL_ALT_CERT_CHAINS` (served cross-sign intermediates otherwise fail `ASN_NO_SIGNER_E` on today's Sectigo/ISRG chains — reproduced on hardware), `WOLFSSL_SHA384/512` (the `-DWOLFSSL_OPTIONS_H` build flag satisfies options.h's include guard before the file is read, silently skipping the defaults that would enable these), `WOLFSSL_SP_384` (P-384 ECDSA on-chip). |

## Dependency

Hostname verification lives in the SDK: **Free-Ink/freeink-sdk#65** adds the missing `wolfSSL_check_domain_name()` call to `SecureClient` (without it, `setCACert` validates the chain but not the identity) plus fail-closed CA loading. This PR is correct with or without it, but only becomes a complete authentication story once the SDK pin advances past that fix.

## Why not mbedTLS / esp_crt_bundle for OTA?

It was tried downstream: routing the firmware download through the verified esp_http_client path works on paper but fails on real C3 hardware at the release-CDN redirect — mbedTLS's RSA verify needs a large contiguous allocation exactly when the largest free block is ~21–39 KB (eight recorded field failures, all at the download stage with a healthy flash path). The wolfSSL transport with SP math and 2 KB fragments is the configuration that actually fits.

## Scope Check

No new transport abstraction, no behaviour change for any non-OTA caller, no new setting or string. If you'd rather the anchors live elsewhere or the parameter ride a different overload, happy to adjust.

## Verification

- [x] `pio run -e default` builds clean
- [x] Chain fixtures: correct chain + hostname verifies; wrong CA fails; hostname mismatch fails; downgrade refused (host tests kept downstream; portable here on request)
- [ ] Hardware: not exercised in this repo — the failure mode this fixes (and the ALT_CERT_CHAINS failure without it) were reproduced on an X3 downstream


sfoulad:
Status update: CI is fully green (all four device builds, unit tests, cppcheck, clang-format) and CodeRabbit reported no blocking findings. This replaces the `setInsecure()` OTA path with verification against pinned GitHub trust anchors, which we currently maintain downstream as OtaCaCerts + HttpDownloader patches that drift on every sync. Requesting a security-focused review of the anchor set and the fail-closed behavior when the bundle can't load; the companion hostname-verification fix is Free-Ink/freeink-sdk#65.

sfoulad:
**Tightening this PR's own entry point before it lands — pushed.**

Integrating this change downstream surfaced three things worth fixing here rather than in a follow-up, since they all sit on the fail-open side of a security boundary.

**1. An optional trailing `caPem` makes "verified" and "unverified" the same call.** The API change was `const char* caPem = nullptr` on the three `fetchUrl` overloads and `runGetSecure`. wolfSSL ships no default CA bundle, so a caller that *intends* a verified fetch but passes a null or empty bundle — a refactor that drops the argument, a config path that yields an empty string — silently lands on `setInsecure()`. "Verified with no anchors" and "deliberately unverified" become indistinguishable at the call site, and the fail-open reading is the default.

Replaced with a distinct entry point, `fetchUrlVerified(url, onData, caPem, username = "", password = "")`, reverting `caPem` off the three `fetchUrl` overloads and `runGetSecure` so their signatures return to what they were. `OtaUpdater`'s two call sites move over; no other caller changes.

**2. The https-only precondition never ran on non-wolfSSL builds.** It lived inside `runGetWolf`, under `#if defined(FREEINK_NET_WOLFSSL)`. On an `esp_http_client` build (the simulator envs), `runGetSecure` routes to `runGet`, `caPem` is discarded, and no precondition runs at all — an `http://` URL handed to a "verified" fetch is fetched in clear. The precondition now lives outside any `#if` and runs on every build. On the esp_http_client path `caPem` is genuinely unused (`esp_crt_bundle` verifies the chain), so the anchor half of the check is conditional and the transport half is not.

**3. The scheme checks were case-sensitive.** Both the precondition and the redirect-downgrade refusal used `url.rfind("https://", 0) != 0`. URL schemes are case-insensitive per RFC 3986 §3.1 and `esp_http_client` parses them that way, so `HTTPS://…` was being refused in verified mode. Both now use a case-insensitive `urlIsHttps()`.

**Where the helper lives.** `src/network/HttpVerifiedFetch.h`, new in this PR — header-only, pure, no Arduino/ESP-IDF includes. Not `lib/UrlOrigin/UrlOrigin.h`, since that is #3250 and still open; not `HttpDownloader.h`, which pulls in `HalStorage.h` and so cannot be compiled on the host. Keeping the precondition pure is what makes it provable: `test/http_verified_fetch` builds the same 11 cases twice, once per TLS backend, because whether missing anchors are fatal is a compile-time property. Removing the case-folding line fails three of them, so that coverage isn't vacuous.

`OtaCaCerts.h`, the anchors, the `setCACert`/`setInsecure` branch and `scripts/patch_wolfssl.py` are unchanged. One comment on the `setCACert` branch is corrected: `SecureClient` at this repo's current SDK pin does not verify the hostname against the presented certificate, so what this authenticates today is the issuing chain, not the peer it belongs to — the Free-Ink/freeink-sdk#65 dependency the description already discloses. I deliberately did **not** bump the submodule pin here, since #65 is still open and this repo pins released SDK commits.

sfoulad:
CI here is red on an infrastructure failure, not on this change — could a maintainer re-run it? I do not have permission to.

```
Build papermono → Install uv → ##[error]fetch failed
Test Status     → "Fail because needed jobs failed"   (cascade from the above)
```

The `uv` install step failed to fetch during setup. `papermono` builds clean locally at this commit, and every other target in the same run is green — clang-format, cppcheck, unit tests, Build x4pro, Title Check.

No push needed from me; a re-run should clear it. I have deliberately not pushed an empty commit to retrigger, since that adds noise to the history for a network hiccup.

PR #3305, opened 2026-08-30
fix: drop section caches when the CSS rule set behind them is gone

sfoulad:
### Problem

Section files store styles already resolved against the CSS rule set that was loaded when they were laid out, and their header (`Section.cpp:122-158`) carries `spec.embeddedStyle` but no CSS rule-set identity — so nothing detects a mismatch later. #3005 made the `sections/` wipe conditional on the CSS cache having been rewritten, which opens a window the previous unconditional wipe did not have.

First boot after a `CSS_CACHE_VERSION` bump (9→10 shipped in #3005 itself, because `interpretAlignment()` now maps `text-align: start/end` to `Start`/`End` rather than `Left`/`Right`):

1. `inspectCache()` reads version 9 → `CacheStatus::Invalid`
2. `deleteCache()` (`Epub.cpp:440`) destroys the v9 cache
3. `parseCssFiles()` returns `Error` — from an SD hiccup, or far more plausibly from free heap under `MIN_HEAP_FOR_CSS_PARSING`, which skips every stylesheet and yields a `Partial` with zero rules that `Epub.cpp` converts to `Error`
4. `cssCacheChanged` is false, so `sections/` **survives**

Already-cached chapters keep rendering with v9 alignment while newly built chapters get no CSS at all. It self-heals only on a later successful open. This affects every device on upgrade.

### Fix

Move the decision into `CssParser::sectionCacheIsStale()`, a pure predicate over the three states that drive it, and drop `sections/` whenever the rule set they were built from was **deleted** on the way in, as well as whenever it was **replaced**.

What #3005 was protecting still holds: the low-memory retry and the already-partial-cache retry both leave the cache file untouched, so their section caches survive. A `Missing` cache with a failed parse also keeps its sections — nothing changed hands, so wiping would be churn on every open of a book with no parseable CSS.

The build-from-scratch path returns to an unconditional wipe: `book.bin` has just been rebuilt there, so any surviving section file was laid out against a different spine whatever the CSS parse returned.

### Notes

No format version bump. The header gains no field — the bug is in the invalidation decision, not in what the file stores, and adding CSS identity to every section header would cost disk bytes plus a one-shot global cache wipe to fix something the predicate fixes for free.

### Testing

Six new cases covering every combination of `CacheStatus` × `CacheLoadResult` × `ParseResult`, including both regression paths (`Invalid`+`Error`, `Complete`-header-fails-to-hydrate+`Error`) and both no-churn paths. Full host suite green; four board builds green.

### DIFF
--- a/lib/Epub/Epub.cpp
+++ b/lib/Epub/Epub.cpp
@@ -455,11 +455,9 @@ bool Epub::load(const bool buildIfMissing, const bool skipLoadingCss) {
           LOG_ERR("EBP", "Failed to reload cache after CSS rebuild");
           return false;
         }
-        const bool cssCacheChanged =
-            cssParseResult == CssParser::ParseResult::Complete ||
-            (cssParseResult == CssParser::ParseResult::Partial && cacheStatus != CssParser::CacheStatus::Partial);
-        if (cssCacheChanged) {
-          // The CSS cache changed, so section caches must use the same rule set.
+        if (CssParser::sectionCacheIsStale(cacheStatus, cacheLoadResult, cssParseResult)) {
+          // Section caches hold styles already resolved against the previous rule set, and
+          // that rule set is either gone or replaced.
           Storage.removeDir((cachePath + "/sections").c_str());
         }
       }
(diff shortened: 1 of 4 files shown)


Uri-Tauber:
Please keep section-cache invalidation inside Epub::load() instead of adding CssParser::sectionCacheIsStale().

  At `lib/Epub/Epub.cpp:437-463`, record in a local boolean whether the CSS cache was deleted or successfully replaced, then remove sections/ when that boolean is true. `Epub::load()` already performs those operations, so it should not ask CssParser to reconstruct them later from three result enums.

  Remove `sectionCacheIsStale()` from `CssParser.cpp`, `CssParser.h`, and its synthetic state-combination tests. Keep the unconditional section-cache removal after rebuilding book.bin at `Epub.cpp:560-566`.

  This keeps one owner for the cache lifecycle and prevents future CSS recovery changes from requiring a matching update to a separate truth table.

PR #3307, opened 2026-08-30
fix(reader): allocate ReaderToolbarUi without abort()ing on OOM

sfoulad:
### Problem

`ReaderToolbarUi` is roughly 2 KB (32 `std::string`s, 16 `ListItem`s, two 324-byte `StyleSet`-carrying prop structs, `FreeInkApp<24,6>`) and is allocated on demand at exactly the wrong moment: the reader menu opens with an EPUB section resident, which is peak heap pressure.

Under `-fno-exceptions` the throwing `new` inside `std::make_unique` calls `abort()` rather than returning null, so an allocation failure there **resets the device** instead of failing to open a menu. Neither call site (`EpubReaderActivity.cpp:251`, `:1815`) checked the result.

`AGENTS.md` mandates `makeUniqueNoThrow<T>()` for exactly this, and `ReaderActivity.cpp` already uses it three lines away.

### Fix

Route both sites through `ensureToolbarUi()`, which uses `makeUniqueNoThrow` and `LOG_ERR`s on failure. `openOverlay()` restores the previous overlay state and `openReaderMenu()` returns to the clean page, so the menu simply does not open. `renderOverlay()` and `handleOverlayInput()` already tolerate a null `toolbarUi`, so nothing downstream needed changing.

### Note for a separate change

`EpubReaderActivity.cpp` still has 11 bare `std::make_unique` sites for child activities (lines 514, 593, 972, 1243, 1270, 1284, 1335, 1354, 1456, 1500, 3060) — every one aborts on OOM. Converting those is bigger than a search-and-replace: child activities are handed to `startActivityForResult()` / `replaceActivity()`, so `ActivityManager` needs a defined null contract first. Happy to follow up if you'd like that shape decided.

### DIFF
--- a/src/activities/reader/EpubReaderActivity.cpp
+++ b/src/activities/reader/EpubReaderActivity.cpp
@@ -248,7 +248,11 @@ void EpubReaderActivity::openReaderMenu() {
     focusedTool = 0;
     panelHoldJumped = false;
     panelCursorShown = !mappedInput.hasTouch();
-    if (!toolbarUi) toolbarUi = std::make_unique<ReaderToolbarUi>(renderer);
+    if (!ensureToolbarUi()) {
+      // Nothing to draw the menu with; leave the page as it is rather than aborting.
+      overlay = Overlay::None;
+      return;
+    }
     toolbarUi->begin();
     discardOverlayPage();
     requestUpdate();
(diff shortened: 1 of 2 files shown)


lpla (approved):

lpla:
@Uri-Tauber is this fine for the v1.6.0 or should we wait for next release for merging?

Uri-Tauber:
@lpla I would prefer real-hardware evidence of the OOM before merging.
So far @sfoulad opened 20+ PRs, but never replied to any of my questions on their other PRs...

PR #3308, opened 2026-08-30
feat: category landing screen for Settings on touch boards

sfoulad:
> **Stacked on #3274.** This branch carries #3274's six commits rebased onto `develop`, then one commit of its own (`e5030f90`). It needs the `SettingsExtension` provider to exist, so please merge #3274 first — or review this against #3274's head. The reviewable change here is the last commit only.

### Problem

`SettingsActivity` renders its categories as a horizontal text tab band. That works at four categories, but a downstream integration registering an extension provider (#3274) pushes it to six, and the labels truncate to `Disp… Rea… Con… Syst… Apps More`. On a touch-only board this is both unreadable and a poor touch target — the whole band is one row tall.

### Change

On touch boards, Settings opens on a **category landing screen**: one card per category, each with an icon and its full name. Tapping a card opens that category's rows; Back returns to the landing. Non-touch boards are untouched and keep the tab band.

This is a **presentation mode of the existing `SettingsActivity`**, not a second Settings implementation. There is no category table, no row table and no action table in the new code — every one of those is a call into the existing model:

| Thing | Call | Resolves to |
|---|---|---|
| category count | `tabCount()` | 4 built-ins + `extraCategories.size()` |
| card label | `tabLabel(i)` | the same override the tab band uses |
| opening a category | `selectCategory(i)` | points `currentSettings` at the built-in list or `extraCategories[…].settings` |
| rows and values | untouched `buildScreen()` / `settingValueText()` | `getSettingsList()` |
| activating a row | untouched `toggleCurrentSetting()` | the existing `SettingAction` switch |

Adding a setting to `getSettingsList()`, or a category to a provider, surfaces it on the landing screen with no further change.

### Non-touch boards

Gated on runtime `mappedInput.hasTouch()` — not `#if FREEINK_CAP_TOUCH`, because `env:default` builds X3 and X4 into one binary, so this is a per-board decision at runtime. That matches how `SettingsActivity::activateIndex`, `UiListActivity::syncListViewport` and `UiTabListActivity::syncTabListViewport` already branch.

The commit removes exactly two things from the button path — the `buildTabBar(screen)` call and the Confirm-hint expression — and both come back verbatim under the guard. Every other change is additive and guarded the same way, so with `hasTouch()` false each new branch reduces to the original expression and the nine new member functions have no reachable caller.

### Geometry

`SettingsCategoryGridLayout.h` is a pure, renderer-free header. `cellRect()` is handed straight to `fui::button()` as the draw rect and `cellHitPadding()` as its `hitPadding`; `hitTest()` is *composed from* those same two functions rather than re-deriving row/column arithmetic, so there is no second computation to drift out of sync. It only ever sees `screen.body()`, so it reads no panel dimension and no rotation, and sizes from theme `rowHeight`/`minTouchSize` so it follows UI scale.

### Testing

23 new cases: cell geometry, uniformity, no-overlap, reading order, partial last row, out-of-range; column choice per band (portrait 2×3, landscape 3×2, narrow→1 column, short band buying height, max-height cap and centring); hit-testing at centres, inclusive corners, exclusive right/bottom edges, *every pixel* of both gutters, and misses on a partial row's empty slot; empty and degenerate bands; 1–3 category short lists; all four orientations. Full host suite green.

Zero new i18n keys — the screen reuses `STR_CAT_*`, `STR_SETTINGS_TITLE`, `STR_SELECT`, `STR_BACK`. Icons generated with the SDK's `gen_icons.py` from stock Lucide, following the `listIcons`/`readerToolbarIcons` convention; the manifest records the exact regeneration command.

### One thing to call out

`SettingsExtensionCategory` gains an optional `const freeink::Icon* icon` (forward-declared, so `SettingsTypes.h` still pulls in nothing from the assets library) so a provider supplies its own artwork rather than the host guessing. With no provider registered, upstream shows four cards and the layout handles that — tested.

The landing keeps an `rtl` parameter (default `false`, currently a call-site constant) because the layout header is fully RTL-tested. CrossPoint has no RTL layout direction today, so if you'd rather that came out until there is one, say so and I'll drop it.


fain182:
Screenshots?

unnamedd (changes requested):
Hi @sfoulad, thanks for your PR. However, I kindly ask you to post screenshots of the work introduced by your PR. It is important to keep a record and to everyone to engage in it.

Uri-Tauber:
I'm drafting this PR since I'm not in favor of #3274.
This one have a promise on its own if we will add more categories in the future.

see my comment in https://github.com/crosspoint-reader/crosspoint-reader/pull/3274#issuecomment-5468169377

Uri-Tauber:
Tastes differ and all that, but I think we can do better:

Uri-Tauber:
Closing as not planned.

PR #3311, opened 2026-08-30
fix(ota): derive device chip identity from the build target, not a flash read of the running app

sfoulad:
### The failure

`firmware_flash::runningPartitionChipId()` establishes the **device's** chip identity by reading two bytes at offset 12 of the **running app partition's own image header** in SPI flash. That makes a compile-time fact depend on a runtime read of mutable storage — and both install channels compare against it and fail *closed*:

- `validateImageFile()` → `Result::BAD_CHIP` (SD install)
- `OtaUpdater`'s header sniff → aborts the transfer (network OTA)

A device whose read returns a wrong-but-not-`0xFFFF` value therefore refuses **every** image on **both** channels at once. This is a field report, not a hypothesis: an Xteink X4 Pro is currently update-locked, rejecting even an image whose header is byte-identical to the firmware it is running. On a sealed board with no accessible USB there is no third channel, so the device cannot be recovered in software.

Three things make it worse:

- The value is memoised in a function-local `static`, so one bad read is permanent for the boot, identically on both channels.
- The `esp_err_t` is discarded and nothing is logged inside the function. The only trace is a `LOG_ERR` at the call site — unavailable on exactly the device class that gets locked out.
- `BAD_CHIP` and `WRONG_BOARD` are surfaced with the same user-facing string, so a field report cannot distinguish "wrong MCU family" from "wrong board".

Worth noting the diagnostic that settled it: the ESP-IDF bootloader validates `chip_id` against `CONFIG_IDF_FIRMWARE_CHIP_ID` on *every* boot (`bootloader_common_loader.c:120-125`). The affected partition boots — so its stored `chip_id` is correct, and the fault is the runtime read, not the stored data.

### The fix

Take the identity from the build target, which is what it always was:

```cpp
static constexpr uint16_t deviceChipId() { return CONFIG_IDF_FIRMWARE_CHIP_ID; }
```

This is the *same constant the bootloader compares each image against*, so the pre-flight guard now predicts the boot-time verdict instead of independently re-deriving it from data that can be wrong. No flash read, no `esp_err_t`, no cached failure, and nothing a bad write or a misconfigured flash mode can corrupt.

`esp_chip_info()->model` is kept only as an `#else` fallback, through an explicit switch rather than a cast: `esp_chip_model_t` and `esp_chip_id_t` are separate enums, and ESP32 diverges (model `1` vs id `0x0000`). Unmapped models yield the unknown sentinel, never `0x0000` — which would read as a valid ESP32 id. Both tables are `static_assert`ed against the real SDK enums, so a renumbering breaks the build rather than the guard.

### OTA can no longer erase the fallback before deciding

`esp_ota_begin(..., OTA_SIZE_UNKNOWN, ...)` erases the entire target partition, and it ran *before* the chip check — so a wrong-chip image destroyed the fallback slot and only then got rejected. The compatibility decision now happens on the first 24 header bytes, and `esp_ota_begin()` is deferred until after it passes. Streaming is preserved: nothing beyond the header is buffered, and those bytes are flushed into the partition immediately after `esp_ota_begin()` succeeds. A transfer ending before a full header now returns an error instead of erasing.

Honest limit: the **board tag** lives at an arbitrary offset in the image body, so knowing it up front would mean buffering the whole image — not possible in 380 KB. A wrong-*board* image can still partially land in the inactive slot; `esp_ota_abort()` keeps it from becoming the boot target. Only the chip decision is genuinely erase-free. This is documented at the call site.

### Fail open on unknown identity, closed on a known mismatch

A cross-family install is a rare user error and generally recoverable. An unreadable *own* identity is an internal fault, and turning it into a hard refusal on every channel bricks any device without exposed USB — which is exactly what happened. The board-tag check stays strict: it is derived from the candidate image rather than from the device's own flash, and it is the real discriminator among same-family boards. This PR replaces a fragile identity *source*; it does not weaken the *check*.

### Also

- Both channels now call one pure seam, `firmware_identity::` in `src/network/FirmwareImageIdentity.h` — they were two hand-copied instances of the same comparison, which is how they came to differ.
- Rejections log the compared **values**, not just a result name.
- Split `STR_FIRMWARE_WRONG_CHIP` from `STR_FIRMWARE_WRONG_DEVICE`, so the two causes are distinguishable in a field report. (`arabic.yaml` was missing `STR_FIRMWARE_WRONG_DEVICE` entirely and silently falling back to English; added.)

### Testing

14 new cases in `test/firmware_image_identity/`, full suite 182/182. Coverage includes a cross-family mismatch (proving the guard is not weakened), a golden 24-byte header vector pinning offset and endianness, short-read rejection that can never yield a silent `0x0000`, the model→id table, and two that speak directly to this bug: **`UnknownDeviceIdentityFailsOpen`** — the regression guard against a future "hardening" re-creating the lockout — and **`DeviceIdentityIsSlotIndependent`**, contrasting a simulated per-slot flash read against the slot-free build-target source.

`pio run -e x4pro` builds clean.


Uri-Tauber:
Can you elaborate about the failure you encountered? I added this mechanism in #2880, and I tested before and verified correct firmware is installable, from both SD and OTA.

Uri-Tauber:
Converting to draft until I will understand the value.

sfoulad:
Fair question, and the PR description under-sold it — it reads like a tidy-up. It is not; it came out of a device that could not be updated at all.

## The failure

An X4 Pro running from partition `app0` refused to install **any** firmware, over both channels. Not a bad artifact — the device rejecting images against its own identity.

From `/debug_log.txt`:

```
[FW] SD update rejected: BAD_CHIP path=…
```

`BAD_CHIP` comes from the chip-identity check, which runs first and returns immediately (`FirmwareFlasher.cpp:149-154`), before the board tag (`:216`), size (`:228`) and SHA (`:267`) barriers. Worth noting separately: the on-screen text "Firmware is for a different device" covers both `BAD_CHIP` and `WRONG_BOARD`, so the UI cannot distinguish them — only the log can.

## The decisive probe

We built an image from the *running* build with the ESP header byte-identical (`chip_id = 0x0009`), changed one byte 2.2 MB in to invalidate the board tag, and left a deliberately stale SHA trailer. If the chip check were sound, that image should have been rejected as `WRONG_BOARD` or `BAD_SHA`.

It was rejected `BAD_CHIP` — never reaching either barrier.

**The device rejected an image whose header matched its own running partition’s header.** So `runningPartitionChipId()` was returning a value that is neither `0x0009` nor `0xFFFF` — `0xFFFF` would have skipped the check via the `deviceChip != 0xFFFF` guard and let the install through.

## The asymmetry that pins it to the flash read

Running the same code from **app1**, the chip id read correctly and the install succeeded. `git diff` across the two builds for `FirmwareFlasher.cpp` is **empty** — identical code, different partition, different outcome. Only reading `app0` failed.

The flasher writes verbatim via `esp_partition_write` after an explicit `esp_partition_erase_range`, not `esp_ota_write`, so there is no deferred-header step that could have corrupted offset 12. And `app0` boots fine, so the bootloader accepted the very header the updater could not read.

## Why the build target is the right source

Both install paths consult that one function — SD at `FirmwareFlasher.cpp:149`, OTA at `OtaUpdater.cpp:262`. Reading the running app’s flash to discover which chip you are executing on asks the flash to tell you something the compiler already knows with certainty. The build target cannot be misread, cannot be affected by partition state, and cannot fail in a way that bricks the update path.

On the X4 Pro this matters more than elsewhere: that board has no USB recovery route, so SD is the only firmware channel. When the identity read fails, the device works perfectly as a reader and can never be updated again.

Things we disproved along the way, so nobody repeats them: wrong artifact copied to the card (`.pio/build` held exactly one `firmware.bin`, SHA-identical to the card’s), board-tag content and chunk-straddling in the scanner, and any build difference between the two images — the probe shares the running build’s own header and still fails.

Happy to keep it as a draft if you would rather see it narrowed, or split the diagnostic-logging half from the identity change. But the underlying defect is real, reproducible on hardware, and its failure mode is a device that cannot accept firmware.

PR #3314, opened 2026-08-31
chore: revert #3096

Uri-Tauber:
## Summary

Revert #3096, #3151 and #3186.
People didn't liked the new look.
I think it best to revert than to try another patch.

---

### AI Usage


Uri-Tauber:
Looks great on the simulator. I want to try on actual device before marking as ready for review.

nowster:
The UI text now looks very rough again, especially upper and lower case M, upper case T and the H in the text "Hide Battery %". Perhaps you need to increase the font hinting weight? (X4)
screenshot

Uri-Tauber:
@nowster Can you compare with v1.5.0?
I'm pretty sure it should be exactly the same.

nowster:
> @nowster Can you compare with v1.5.0? I'm pretty sure it should be exactly the same.

Yes, it was that bad on 1.5.0 too. #3096 was a marked improvement.

Uri-Tauber:
@nowster A lot of people complained on #3096. 
I think the majority of users prefer it as it is now.

nowster:
> @nowster A lot of people complained on #3096. I think the majority of users prefer it as it is now.

I can't see any complaints on #3096 itself. Are there any issues where it's mentioned?

Uri-Tauber:
See #3099. There are other places.

nowster:
> See #3099. There are other places.

To my eyes none of the images in #3099 look _wrong_, just different. One problem with the menu fonts is that they're not aliased (2-level, not 4-level) and without hinting enabled that causes the stroke widths to vary a lot (like the H, M and T in the screenshot).

But I'm weird. I have very good visual acuity and my preferred reader font is Bitter at 10pt (self-compiled, with SemiBold as the regular font).

Uri-Tauber:
Personally, I dont see the difference in English, but for Hebrew I felt something was off. I couldn't put my finger what it was exactly, but something was wrong.
Other user complained about Arabic text, so we decided it will be simpler to revert the whole thing, rather than a pile of small fixes.

Anyway, I'm trying to do what the community want, and my understanding was more people wanted it reverted.

PR #3317, opened 2026-08-31
fix: release dictionary font caches on exit

### LINKED ISSUE #3019: Second dictionary lookup results in "Not enough memory"
### Affected Version

v1.5 (or next nightly)

### Describe the Bug

Using v1.5 on an X3 with Bookerly font (with IPA), and Wikitionary, the first word lookup works fine.  If I then look up a second word, I get "Not enough memory" error.

### Steps to Reproduce

Open a book (and went to 4th bookmark)
Select "Look Up"
Select a word and "Lookup"
 (the word and definition appear)
Return to the book
Select "Look Up" again
Select another word and "Look Up"
***fails with "Not enough memory"

### Expected Behavior

The second (and subsequent) look ups should show the word.

```shell
Tried this with a different book and could NOT reproduce the error.  Went back to the original book and, even at 4th bookmark, can not reproduce.  (This was quite consistent before...)
```

Uri-Tauber:
## Summary

Should help with #3019. 

---

### AI Usage

### DIFF
--- a/src/activities/reader/DictionaryDefinitionActivity.cpp
+++ b/src/activities/reader/DictionaryDefinitionActivity.cpp
@@ -44,6 +44,13 @@ void DictionaryDefinitionActivity::onEnter() {
   requestUpdate();
 }
 
+void DictionaryDefinitionActivity::onExit() {
+  Activity::onExit();
+  if (auto* fcm = renderer.getFontCacheManager()) {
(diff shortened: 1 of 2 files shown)


Uri-Tauber:
Confirmed to fix the OOM error in #3019.

itsthisjustin (approved):

PR #3340, opened 2026-09-01
fix(x4pro): shut down SD card before deep sleep

jfmyers9:
## Summary

This fixes an X4 Pro sleep-power regression exposed after #3215 made battery-only fast wake reliable.

On the RC03/develop baseline, I observed approximately 2% battery loss per hour while using the `Cover` sleep screen. The mounted filesystem and SDMMC host remained active until deep sleep, even though the SD power gate was
disabled later in the sleep sequence.

After sleep-screen storage access finishes, this change unmounts the filesystem, and stops the SDMMC host before SD power is disabled. A deep-sleep wake mounts the card normally during boot.

I tested an earlier version containing the same SD teardown on an X4 Pro for 24 hours with the `Cover` sleep screen. The battery indicator showed minimal to no loss, compared with approximately 2% per hour on the RC03/develop baseline.

## Scope Check

- [ ] If this PR touches `freeink-sdk/`, `lib/hal/`, the bootloader, OTA, or recovery code, I have coordinated with the relevant maintainer.

Stock/fork context: #3215 documents that stock firmware does not reliably provide battery-only fast wake on affected hardware. This PR fixes the SD power regression in CrossPoint's fast-wake implementation. No popular fork
is known to solve this combination.

## Additional Context

- Applies only to native-SDMMC targets; SPI-backed storage retains its existing behavior.
- `pio run -e x4pro`: passed.
- `pio run -e default`: passed.

### AI Usage

### DIFF
--- a/lib/hal/HalStorage.cpp
+++ b/lib/hal/HalStorage.cpp
@@ -44,6 +44,28 @@ class HalStorage::StorageLock {
   ~StorageLock() { xSemaphoreGiveRecursive(HalStorage::getInstance().storageMutex); }
 };
 
+bool HalStorage::prepareForDeepSleep() {
+#if FREEINK_SD_SDMMC
+  StorageLock lock;
+  if (!SDCard.ready()) return true;
+
+  // Sleep preparation runs after all filesystem users have stopped. Detach the
+  // volume, synchronize the device, then stop the native host before its
+  // external rail is disabled.
+  FsBlockDeviceInterface* blockDevice = SDCard.detachFilesystemForRawAccess();
+  if (!blockDevice) {
+    LOG_ERR("STORAGE", "Failed to detach SD volume before deep sleep");
+    return false;
+  }
+  const bool synced = blockDevice->syncDevice();
+  if (!synced) LOG_ERR("STORAGE", "Failed to synchronize SD device before deep sleep");
+  blockDevice->end();
+  LOG_DBG("STORAGE", "SD volume and host stopped for deep sleep");
+  return synced;
+#endif
+  return true;
+}
+
 #if FREEINK_CAP_USB_MSC && !FREEINK_SD_SDMMC
 #error "USB Drive requires an SDMMC-backed storage profile"
 #endif
--- a/lib/hal/HalStorage.h
+++ b/lib/hal/HalStorage.h
@@ -24,6 +24,10 @@ class HalStorage {
   HalStorage();
   bool begin();
   bool ready() const;
+  // Release the mounted volume and its block device before deep sleep. Call
+  // only after all file users have stopped; existing HalFiles become invalid.
+  // A deep-sleep wake resets the MCU and mounts storage again through begin().
+  bool prepareForDeepSleep();
   // USB Drive exclusively owns the SD card while active. Callers must stop
   // all filesystem work before beginUsbDrive(), then reboot after endUsbDrive().
   bool beginUsbDrive();
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -289,6 +289,7 @@ void enterDeepSleep(bool fromTimeout = false) {
 
   halTiltSensor.deepSleep();
   display.deepSleep();
+  Storage.prepareForDeepSleep();
   LOG_DBG("MAIN", "Entering deep sleep");
 
   powerManager.startDeepSleep(gpio);
@@ -450,6 +451,7 @@ void setup() {
       // the device in a USB-replug boot loop (or sleep right after a flash).
       break;
 #else
+      Storage.prepareForDeepSleep();
       powerManager.startDeepSleep(gpio);
       break;
 #endif


itsthisjustin:
Very odd that I don't experience this drain while also using cover as my setting. This change is minimal and non distructive so should be fine to merge but I'm curious how you specifically verified it was the sd power?

Uri-Tauber:
@jfmyers9 Have you tested this with RC03 or nightly?

RC03 had another power drainage bug, which @itsthisjustin fixed in #3248.

jfmyers9:
@itsthisjustin Mostly trial and error. I started by investigating the frontlight and pulling in https://github.com/Belphemur/crosspoint-x-reader/pull/15 and [similar fixes](https://github.com/jfmyers9/crosspoint-reader/commit/79e26b59886fa432e3c243a8d45838f52b2877f2), but I was still getting terrible performance. I noticed that I didn’t see the drain with `None` or `Dark`, only with `Cover`. Since that path loads the cover from the SD card, I tried fully shutting down the SD filesystem and host before deep sleep.

I haven’t been able to reproduce the drain since (0% battery loss overnight vs ~20% the night before). So I didn’t directly verify that the SD card was drawing the power, but quiescing it is the change that made the drain disappear in my testing.

@Uri-Tauber I started with RC03, but after the poor performance I've been flashing off the `develop` branch. So I was getting this battery drain as of yesterday's `develop`. I'm happy to flash the current develop again just to confirm that I still see the issue.

itsthisjustin:
> @itsthisjustin Mostly trial and error. I started by investigating the frontlight and pulling in [Belphemur#15](https://github.com/Belphemur/crosspoint-x-reader/pull/15) and [similar fixes](https://github.com/jfmyers9/crosspoint-reader/commit/79e26b59886fa432e3c243a8d45838f52b2877f2), but I was still getting terrible performance. I noticed that I didn’t see the drain with `None` or `Dark`, only with `Cover`. Since that path loads the cover from the SD card, I tried fully shutting down the SD filesystem and host before deep sleep.
> 
> I haven’t been able to reproduce the drain since (0% battery loss overnight vs ~20% the night before). So I didn’t directly verify that the SD card was drawing the power, but quiescing it is the change that made the drain disappear in my testing.
> 
> @Uri-Tauber I started with RC03, but after the poor performance I've been flashing off the `develop` branch. So I was getting this battery drain as of yesterday's `develop`. I'm happy to flash the current develop again just to confirm that I still see the issue.

Please do if you can. I'd like to get this merged

jfmyers9:
Yeah, will do. Flashing now (7bcc3a69) and I’ll report back in an hour or two. Over the last 15 hours I’ve only lost 1% battery, including some active use.

Personally, I think this change has value on its own. Explicitly syncing storage, unmounting the filesystem, and shutting down the SD host after the sleep screen finishes seems like reasonable defensive cleanup regardless of what the root cause ends up being.

itsthisjustin:
Confirmed with a user who is using a multi-meter on the device, the sd power rails were in fact not powering down. Going to merge this.

itsthisjustin (approved):

Uri-Tauber:
from code correctness perspective: 
Please move SD shutdown into `SDCardManager`, which already owns the filesystem and SD host.

Currently, `lib/hal/HalStorage.cpp:55-62` repurposes `detachFilesystemForRawAccess()`, but that API is specifically for USB Drive handoff (`freeink-sdk/libs/hardware/SDCardManager/include/SDCardManager.h:84-87`).

Add a `shutdown()` method to `SDCardManager`, then have `HalStorage` call it while holding its lock. Keep the two sleep calls at `src/main.cpp:292` and `src/main.cpp:454`, and leave `detachFilesystemForRawAccess()` exclusive to USB Drive.

The returned `bool` can also be removed because neither caller uses it.

itsthisjustin:
I may actually just integrate this into another PR i'm working on so they don't conflict. I'll give you credit @jfmyers9

itsthisjustin (changes requested):
from code correctness perspective:
Please move SD shutdown into SDCardManager, which already owns the filesystem and SD host.

Currently, lib/hal/HalStorage.cpp:55-62 repurposes detachFilesystemForRawAccess(), but that API is specifically for USB Drive handoff (freeink-sdk/libs/hardware/SDCardManager/include/SDCardManager.h:84-87).

Add a shutdown() method to SDCardManager, then have HalStorage call it while holding its lock. Keep the two sleep calls at src/main.cpp:292 and src/main.cpp:454, and leave detachFilesystemForRawAccess() exclusive to USB Drive.

The returned bool can also be removed because neither caller uses it.

itsthisjustin:
@Uri-Tauber @jfmyers9 can you test this? https://github.com/crosspoint-reader/crosspoint-reader/pull/3341

jfmyers9:
@itsthisjustin I'll validate the other change, but as you would expect I'm now struggling to repro this. The turnaround on this is painful because validation is just "waiting" at the moment.

If you want to pull this out of the other changeset, that's fine with me and I'll get back to you once I have more data. This could be coincidental with something else that is going on.

itsthisjustin:
> @itsthisjustin I'll validate the other change, but as you would expect I'm now struggling to repro this. The turnaround on this is painful because validation is just "waiting" at the moment.
> 
> If you want to pull this out of the other changeset, that's fine with me and I'll get back to you once I have more data. This could be coincidental with something else that is going on.

The good news is I was able to validate this build with a user with an opened up pro and is measuring voltage

jfmyers9:
Very cool. I've gone ahead and flashed the other build, nothing immediately wrong, and it sounds like you have more data than I do at this point. :ship:

jfmyers9:
Merged in #3341

PR #3341, opened 2026-09-01
fix: sd not shutting down on deep sleep & restores single click power on

### LINKED ISSUE #3332: Device does not turn on normally anymore
### Device

Xteink X4

### Affected Version

1.5.0-dev+3099231

### Describe the Bug

After flashing the nightly version 1.5.0-dev+3099231, my Xteink X4 does not turn on normally anymore. The device only turns on if the reset button is pressed first and the power button afterwards.

### Steps to Reproduce

1. Turn off device via power button
2. Press power button to turn on
3. Device does not turn on and only shows the wallpaper

### Expected Behavior

1. Turn off device via power button
2. Press power button to turn on
3. Device turn on

```shell

```

itsthisjustin:
With Short Power Button Press = Sleep, a single click wakes the device again on every board. The held-button verification still samples the power pin early in setup() (a click is released within milliseconds of boot), but the sleep-or-boot decision now waits until SETTINGS is loaded: click-to-wake is a setting, and on the C3 Xteink boards a battery power-off cuts all power, so only SD state survives to the next boot. Regression from #3191, which required the button to be held through boot regardless of the setting.

Also fixes the issue identified by @jfmyers9 where the SD card wasn't shutting down on the x4pro on sleep

Closes #3332

### DIFF
--- a/lib/hal/HalStorage.cpp
+++ b/lib/hal/HalStorage.cpp
@@ -44,6 +44,11 @@ class HalStorage::StorageLock {
   ~StorageLock() { xSemaphoreGiveRecursive(HalStorage::getInstance().storageMutex); }
 };
 
+void HalStorage::prepareForDeepSleep() {
+  StorageLock lock;
+  SDCard.shutdown();
+}
+
 #if FREEINK_CAP_USB_MSC && !FREEINK_SD_SDMMC
 #error "USB Drive requires an SDMMC-backed storage profile"
 #endif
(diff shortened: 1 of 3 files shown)


Uri-Tauber:
I can confirm the bug from #3332 is fixed here.
Cant say anything about the battery life improvement.

Uri-Tauber (approved):
LGTM

itsthisjustin:
Fyi I do appear to get a false crash/boot after hours in deep sleep on the x4pro. Absolutely no idea what causes it but I'm running a build with extra logging to try to replicate it. It's not really a big deal considering it just immediately boots automatically and it seems to only happen after multiple hours of the device being asleep. Weird, but documenting it here. I've sat at 97% battery though all day though which is nice.

Issue #3116, opened 2026-08-19
EPUB reader silently drops paragraphs/headings when publisher puts epub:type="pagebreak" on content elements (<p>, <h2>, …)


gregbarbosa:
## Summary

Whole paragraphs and headings are silently dropped from rendered books when a publisher marks the print-page boundary by putting `epub:type="pagebreak"` (or `role="doc-pagebreak"`) **on the content element itself** instead of on an empty marker span.

## Affected code

`lib/Epub/Epub/parsers/ChapterHtmlSlimParser.cpp`, `startElement()` (~line 909 on current `develop`):

```cpp
// Skip blocks with role="doc-pagebreak" and epub:type="pagebreak"
if (atts != nullptr) {
  for (int i = 0; atts[i]; i += 2) {
    if (strcmp(atts[i], "role") == 0 && strcmp(atts[i + 1], "doc-pagebreak") == 0 ||
        strcmp(atts[i], "epub:type") == 0 && strcmp(atts[i + 1], "pagebreak") == 0) {
      self->skipUntilDepth = self->depth;
      self->depth += 1;
      return;
    }
  }
}
```

The skip is applied to **any** element carrying the attribute, and `skipUntilDepth` discards the element's entire subtree. That is correct for the common idiom — an empty marker:

```html
<span epub:type="pagebreak" id="p556" title="556"/>
```

but some publishers instead put the attribute on the element that *starts* the print page:

```html
<p class="txt">"But you do?"</p>
<p class="txt" epub:type="pagebreak" id="p557" title="557">And Manon had to wonder if she'd somehow … somehow become tangled in whatever plans the queen had woven. …</p>
```

and also on part-title headings:

```html
<h2 class="prect" epub:type="pagebreak">Part One: The Fire-Bringer</h2>
```

In that idiom the element carries real content, and the current code silently deletes it from the book.

## Real-world evidence

A merged trade EPUB (Bloomsbury-style InDesign export; *Empire of Storms* + *Tower of Dawn*):

| markup | occurrences | current behavior |
|---|---|---|
| `<span epub:type="pagebreak"/>` (empty marker) | 932 | correctly skipped (renders nothing) |
| `<p … epub:type="pagebreak">…text…</p>` | **349** | **paragraph dropped** |
| `<h2 … epub:type="pagebreak">…title…</h2>` | **79** | **heading dropped** |

[...]

Issue #1770, opened 2026-04-27
Words included in pagebreak tag not displayed and missing


ninamaria31:
### Affected Version

1.2.0

### Describe the Bug

In my book sometimes the first word of a paragraph is missing (see image). I couldn't find a pattern on when exactly this happens, as in most paragraphs the first word is displayed correctly. So I looked at the underlying .xhtml file via Calibre (e-book edit) and I think I figured out that the reason is probably the pagebreak tag. Only the word included in the tag is missing and not being displayed on the e-reader.

Example: The "He", which is included in the tag is missing when reading on the XTEINK but it is displayed correctly when opening the Epub on my laptop in the previewer (both in Calibre and Apple Books). 

`<p class="calibre6"><span epub:type="pagebreak" id="pg18">He</span> unwrapped a cigar, bit the end off, and spit it into an ashtray. `

<img width="726" height="967" alt="Image" src="https://github.com/user-attachments/assets/b564179a-3481-497b-bbe9-15ba0b3309a6" />

### Steps to Reproduce

Open an Epub which contains page break formatting like in the example code snippet. 
`<p class="calibre6"><span epub:type="pagebreak" id="pg18">He</span> unwrapped a cigar, bit the end off, and spit it into an ashtray. `

### Expected Behavior

A word included in the page break should be displayed on the reader. 

### Relevant Log Output/Screenshots

```shell

```

insunaa:
According to a discussion thread this is intended behaviour and it is apparently the epub that is mistaken here.

I am using the attached patch to change this behaviour for my own X4

[pagebreaks.patch](https://github.com/user-attachments/files/27339053/pagebreaks.patch)

Patch in clear text, tho due to whitespace handling it may not work as intended if not using the file:

```patch
diff --git a/lib/Epub/Epub/parsers/ChapterHtmlSlimParser.cpp b/lib/Epub/Epub/parsers/ChapterHtmlSlimParser.cpp
index eee24c4..f6f4590 100644
--- a/lib/Epub/Epub/parsers/ChapterHtmlSlimParser.cpp
+++ b/lib/Epub/Epub/parsers/ChapterHtmlSlimParser.cpp
@@ -504,6 +504,8 @@ void XMLCALL ChapterHtmlSlimParser::startElement(void* userData, const XML_Char*
   }

   // Skip blocks with role="doc-pagebreak" and epub:type="pagebreak"
+  if (false)
+  {
   if (atts != nullptr) {
[...]

PR #3349, opened 2026-09-02
fix(epub): stop dropping real text carried by pagebreak markers

s0lness:
## Problem

Elements carrying `role="doc-pagebreak"` / `epub:type="pagebreak"` are skipped together with their whole subtree (`ChapterHtmlSlimParser.cpp`, since #58). That is right for the well-formed idiom (an empty marker with the page number in `title`/`aria-label`), but real books misuse the attribute two ways, and in both cases CrossPoint silently deletes book text:

- **Calibre conversions that wrap the continuation of the paragraph inside the marker span**, e.g. `... the sentence stops<br/><span epub:type="pagebreak" aria-label="136" id="page136" title="136">and its continuation, dozens of words of real prose, sits inside the marker.</span>` — 77 dropped sentence fragments found in one such epub.
- **Publishers that put the attribute directly on the `<p>` or heading that starts the print page** — #3116 documents 428 dropped paragraphs and part titles in a single InDesign-exported book; #1770 reported the same class of loss earlier.

Other readers (KOReader, Calibre viewer, Apple Books) render this content.

## Fix

Two-pronged, matching the two idioms:

- **Text-bearing block elements** (`p`, `h1`-`h6`, `blockquote`, `li`) with the attribute now render completely normally, keeping their block style. In that idiom the label lives in the `title` attribute anyway, so nothing new becomes visible.
- **Marker-ish elements** (`span`, `a`, `div`, `hr`, ...) capture their text speculatively (32-byte fixed buffer, no heap) and drop it only when it looks like the page label: empty, byte-equal to the element's `aria-label`/`title`, or at most 8 chars of arabic/roman numerals. Anything else is replayed through the normal text path (buffer overflow or a nested child element proves real content and replays immediately). Replays rewind `visibleTextOffset` to the marker start so reading-resume positions stay correct. Empty markers and bare page numbers render nothing, exactly as before.

`SECTION_FILE_VERSION` 44 -> 45 so cached section layouts rebuild.

## Testing

- `pio run -e gh_release` builds clean on this branch.
- Verified on an Xteink X3 (against the v1.5.0 baseline, same patch): the 77 wrapped fragments render; a section with 15 such markers paginates, caches and quick-resumes normally; books with empty markers show no stray page numbers.

## Known limits

- Replayed span content joins the current block (inline flow); for the block-element idiom that is why block tags bypass capture entirely.
- The roman-numeral heuristic could still hide a marker-wrapped fragment that is a single short word spelled only with `ivxlcdm` letters; never seen in practice.


s0lness:
Added a second commit: when a marker's captured content is replayed as real prose, a `<br>` sitting immediately before the marker is swallowed instead of materializing, so the restored sentence flows as one paragraph. The `<br>` and the marker are emitted together by these converters as one print-page artifact; a marker that resolves as empty/label still materializes the `<br>` as genuine formatting. Implemented as a deferred-`<br>` flag resolved at every following content/structure event, byte-for-byte identical to today for documents without markers. Separate commit on purpose: easy to drop if you prefer the stricter render-the-br behavior.

s0lness:
Third commit, two refinements found while dogfooding: (1) with "Extra paragraph spacing" enabled, a mid-paragraph `<br>` rendered with the half-line block gap and read as a paragraph break — the gap is now withheld at `<br>`-with-text boundaries and given back when the `<br>` block stays empty, so scene breaks are pixel-identical to before; (2) an empty pagebreak marker containing a `<br>` (`<span epub:type="pagebreak"...><br/></span>`, seen in the wild) no longer emits a stray line break — a `<br>` child is swallowed during capture instead of counting as real content. Branch still builds clean with `pio run -e gh_release`.

PR #3353, opened 2026-09-02
fix(webserver): normalize every user-supplied path and escape file names in the files page

s0lness:
## Summary

* **What is the goal of this PR?** Make every user-supplied path and file name that reaches the SD card through the web server go through the same normalization and validation, and stop file names from breaking out of HTML attributes in the files page.
  * `normalizeWebPath()` (already used by rename and move) is now applied to the `path` argument of `/api/files`, `/download`, `/upload`, `/create-folder`, `/delete`, and to the WebSocket upload `START:` path. Before this, those handlers only prefixed `/` and trimmed a trailing slash, so `..` segments went straight to the storage layer.
  * A small `isSafePathComponent()` helper rejects `/`, `\` and `..` in single-segment arguments: the multipart upload `filename`, the WebSocket upload file name, and the `name` of `/create-folder` (which also gets the existing `isProtectedItemName()` check it was skipping). This mirrors what `handleFontUploadData` already does for font uploads.
  * `FilesPage.html`: file and folder names interpolated into `onclick="..."` attributes are now HTML-escaped after the existing JS-quote escape, so a name containing `"`, `<` or `>` (uploadable before this change) cannot break out of the attribute. `escapeHtml()` already existed on the page.

No behavior change for legitimate paths: `normalizeWebPath()` returns the same shape (leading `/`, no trailing `/`, root as `/`) the handlers produced by hand.

## Scope Check

Please confirm:

      the relevant maintainer. (Not applicable: only `src/network/` is touched.)

## Additional Context

**Why:** the web server is unauthenticated on the LAN by design, so input handling is the only guard. Rename and move normalized their paths; the other five handlers and the two upload paths did not, and the divergence was easy to miss because `WebDAVHandler` has its own copy of the same logic. This PR only makes the existing checks consistent; it does not add authentication.

**Verification (device on Wi-Fi with the file transfer screen open, `HOST` = its IP):**

1. Upload with a file name carrying a parent reference:
   `curl -F 'file=@test.epub;filename=../x.epub' "http://HOST/upload?path=/Books"`
   Before: the request is accepted and the storage layer decides where `x.epub` lands. After: the upload is rejected with "Invalid file name".
2. Create a folder named `..`:
   `curl -X POST -d "name=..&path=/Books" http://HOST/create-folder`
   After: `400 Invalid folder name`.
3. Paths with `..` are collapsed before any storage call:
   `curl "http://HOST/api/files?path=/Books/../Books"` lists `/Books` (same result as before, now via `normalizeWebPath`).
4. Files page: upload a file named `a"b.epub`; the row's move/rename/delete buttons still work and the page source shows the name escaped inside the `onclick` attribute.
5. Regression: normal upload, download, rename, move, delete and folder creation from the files page behave as before.

**Memory / flash impact:** none measurable. `gh_release` build on this branch: RAM 17.1% (56140 B), Flash 82.4% (5401083 B), identical to `develop` at `b95965f4`. `bin/clang-format-fix`, `pio check` and the native unit tests (173/173) pass locally.

---

### AI Usage

Did you use AI tools to help write this code? _**YES**_ (Claude Code; the findings come from a code audit, the diff was reviewed and built by hand before opening this PR.)

https://claude.ai/code/session_017HaL1cqtG2c3L475nYDJCc

### DIFF
--- a/lib/FsHelpers/FsHelpers.cpp
+++ b/lib/FsHelpers/FsHelpers.cpp
@@ -2,6 +2,7 @@
 
 #include <algorithm>
 #include <cctype>
+#include <cstdint>
 #include <cstring>
 #include <string_view>
 #include <vector>
@@ -181,6 +182,19 @@ std::string extractFolderPath(const std::string& filePath) {
     return "/";
   }
   return filePath.substr(0, lastSlash);
+}
+
+bool isSafePathComponent(std::string_view name) {
+  if (name.empty()) {
+    return false;
+  }
+  if (name.find('/') != std::string_view::npos || name.find('\\') != std::string_view::npos) {
+    return false;
+  }
+  if (name == "." || name == "..") {
+    return false;
+  }
+  return true;
 }
 
 void sanitizePathComponentForFat32(const char* input, char* output, size_t maxLen) {
--- a/lib/FsHelpers/FsHelpers.cpp
+++ b/lib/FsHelpers/FsHelpers.cpp
@@ -185,16 +185,7 @@ std::string extractFolderPath(const std::string& filePath) {
 }
 
 bool isSafePathComponent(std::string_view name) {
-  if (name.empty()) {
-    return false;
-  }
-  if (name.find('/') != std::string_view::npos || name.find('\\') != std::string_view::npos) {
-    return false;
-  }
-  if (name == "." || name == "..") {
-    return false;
-  }
-  return true;
+  return !name.empty() && name.find_first_of("/\\") == std::string_view::npos && name != "." && name != "..";
 }
 
 void sanitizePathComponentForFat32(const char* input, char* output, size_t maxLen) {
--- a/lib/FsHelpers/FsHelpers.h
+++ b/lib/FsHelpers/FsHelpers.h
@@ -71,6 +71,15 @@ inline bool hasCssExtension(const String& fileName) {
 }
 std::string extractFolderPath(const std::string& filePath);
 
+// Rejects an empty component, one containing '/' or '\', or the exact components
+// "." and "..", so a single filename/folder-name argument can never be used to
+// escape the directory it is placed into. Names like "volume..2.epub" or
+// "notes...txt" that merely contain ".." are accepted.
+bool isSafePathComponent(std::string_view name);
+inline bool isSafePathComponent(const String& name) {
+  return isSafePathComponent(std::string_view{name.c_str(), name.length()});
+}
+
 /**
  * Sanitize a filename/path component for FAT32 in a caller-provided buffer.
  * Replaces invalid path characters, spaces, and control characters with '-'.
--- a/src/network/CrossPointWebServer.cpp
+++ b/src/network/CrossPointWebServer.cpp
@@ -837,10 +837,12 @@ void CrossPointWebServer::handleCreateFolder() const {
     return;
   }
   if (!isSafePathComponent(folderName)) {
+    LOG_DBG("WEB", "Rejected unsafe folder name: %s", folderName.c_str());
     server->send(400, "text/plain", "Invalid folder name");
     return;
   }
   if (isProtectedItemName(folderName)) {
+    LOG_DBG("WEB", "Rejected protected folder name: %s", folderName.c_str());
     server->send(403, "text/plain", "Cannot create protected item");
     return;
   }
--- a/src/network/CrossPointWebServer.cpp
+++ b/src/network/CrossPointWebServer.cpp
@@ -82,21 +82,6 @@ bool isProtectedItemName(const String& name) {
   return false;
 }
 
-// Rejects path separators and parent-directory references so a single filename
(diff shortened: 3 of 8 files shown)


s0lness:
Review follow-up:

- **Applied**: the two new folder-name rejections in `handleCreateFolder` now log before returning (the upload and WebSocket rejections already did).
- **Not changed, on purpose**: the error bodies are not routed through `tr()`. Every existing 4xx/5xx body in this file ("Missing path", "Invalid path", "Cannot access protected items", ...) is a plain English API string; `tr()` is used for the rendered pages. Localizing only the new ones would be inconsistent; happy to do a separate pass on all of them if that is wanted.
- **Not changed, out of scope**: the "protected subtree" point. Access control here is name-based (items starting with "." and the HIDDEN_ITEMS list) and was already that way before this PR; the PR only makes the existing checks apply on every handler. A subtree rule would be a behavior change for the `showHiddenFiles` users and deserves its own discussion.

lpla:
> Review follow-up:
> 
> * **Applied**: the two new folder-name rejections in `handleCreateFolder` now log before returning (the upload and WebSocket rejections already did).
> * **Not changed, on purpose**: the error bodies are not routed through `tr()`. Every existing 4xx/5xx body in this file ("Missing path", "Invalid path", "Cannot access protected items", ...) is a plain English API string; `tr()` is used for the rendered pages. Localizing only the new ones would be inconsistent; happy to do a separate pass on all of them if that is wanted.
> * **Not changed, out of scope**: the "protected subtree" point. Access control here is name-based (items starting with "." and the HIDDEN_ITEMS list) and was already that way before this PR; the PR only makes the existing checks apply on every handler. A subtree rule would be a behavior change for the `showHiddenFiles` users and deserves its own discussion.

@s0lness Thanks, I agree that localizing only these new API responses would be inconsistent, and that changing the protected-subtree policy belongs in a separate discussion.

There is one narrower correctness issue in the new helper, though: `name.indexOf("..") >= 0` rejects legitimate FAT filenames such as `volume..2.epub` or `notes...txt`.

Since `/` and `\` are already rejected, only the exact components `.` and `..` need special handling here. Could you narrow that condition and add a couple of focused cases for valid names containing repeated dots?

s0lness:
Done in 2663990c. The check now lives in `FsHelpers::isSafePathComponent()` (std::string_view + the usual `String` overload) and only rejects an empty component, any `/` or `\`, and the exact `.` / `..` components; `volume..2.epub` and `notes...txt` are accepted. Added `test/fs_helpers` with the repeated-dots cases (verified red under the old `indexOf("..")` semantics, green now), the separator cases, and two `normalisePath` cases documenting the traversal collapse this PR relies on. Native suite: 178/178.

lpla (approved):

lpla:
@Uri-Tauber @itsthisjustin @uxjulia should we merge this before 1.6.0?

Uri-Tauber:
Nitpick: Stop putting filenames and paths directly inside each button’s JavaScript.

Store the file information on the row as ordinary data, then use one shared click handler to decide whether the user clicked Move, Rename, or Delete. This keeps user-controlled text separate from executable code and makes future actions safe by default, without requiring every new button to repeat the correct escaping sequence.

s0lness:
@Uri-Tauber You are right, and sorry for making you point it out. Escaping the values at every call site was treating the symptom; the inline handlers were the actual problem.

Done in a92fb33e: the move/rename/delete buttons on file and folder rows now carry `data-action`, `data-name`, `data-path` and `data-folder`, and a single delegated `click` listener on the static `#file-table` container reads `closest('[data-action]')` and dispatches to the right modal. The names travel as attribute text and come back through `dataset` as plain strings, so there is no JS string context left and the double escaping is gone. The listener is bound once (guarded by a `data-` flag on the container, which is never replaced, only its `innerHTML`), and the modal openers keep their existing signatures.

Left inline: the "Delete Selected" toolbar button, which takes no arguments and reads the checkbox selection itself, and the static modal/confirm buttons.

@lpla thanks for the approval. Happy for this to wait until after 1.6.0 if that is safer for the release; nothing here is urgent.

Uri-Tauber:
Thanks @s0lness.
If its not urgent, let wait after release 1.6.0.

s0lness:
1.6.0 is out. This one is approved and green whenever you want it.

Uri-Tauber:
Please make two small simplifications before merge. 
In FilesPage.html, move the row-action code from lines 1959–1979 into the existing file-table click handler at lines 2992–2999. Delete the second listener and the actionsBound marker because hydrate() has only one caller. 
In FsHelpers.cpp, replace the branches at lines 187–198 with one return expression that checks empty names, separators, ., and ...

Should remove ~20 LoC

s0lness:
Good catch @Uri-Tauber, done in 5c3df539: the row actions now live in the existing file-table click handler (second listener and actionsBound marker gone), and isSafePathComponent is a single return expression. Net -16 lines, tests still 196/196.

PR #3354, opened 2026-09-02
fix: keep reading position and bookmarks when a book is renamed or moved

s0lness:
## Summary

* **What is the goal of this PR?** A book that is renamed or moved keeps its reading position, its bookmarks and its recent-books entry, whichever path moved it (web files page, WebDAV, or the reader's own "move finished book to /Read").
  * New helper `moveBookData(oldPath, newPath)` (`src/util/BookDataMove.{h,cpp}`): clears any stale cache and bookmarks file already sitting at the destination, renames the book's cache directory (which holds `progress.bin`), renames its bookmarks file, and calls `RECENT_BOOKS.updatePath()`. Cache paths come from each format's own `getCachePath()` (EPUB / XTC / TXT), the same dispatch `clearBookCache()` already uses, so no hash logic is duplicated. Every step is non-fatal and logged.
  * `EpubReaderActivity::moveFinishedBookToReadFolder` now uses the helper. It already renamed the cache dir and updated recents, but never moved the bookmarks file, so bookmarks vanished after the automatic move to `/Read`.
  * Web server `handleRename` / `handleMove`: they called `clearBookCache()` before renaming, which deletes the cache dir together with `progress.bin`; the reading position was lost on every rename or move made from the files page. They now rename first and call the helper on success.
  * WebDAV `MOVE` / `COPY` with overwrite: the destination's stale cache and bookmarks are cleared before the destination file is removed (PUT already did this); `MOVE` then migrates the source's data like the other paths. `COPY` only clears the destination (a copy is a new book).

Bookmarks live in `/.crosspoint/bookmarks/<flattened book path>.json` (`BookmarkUtil::getBookmarkPath`), outside the cache dir, which is why the existing cache-dir rename was not enough.

## Scope Check

Please confirm:

      the relevant maintainer. (Not applicable: `src/util`, `src/network`, `src/activities/reader` only.)

## Additional Context

**Reproduction (before this PR):**
1. Open a book, read a few pages, add a bookmark, leave the reader.
2. Files page: rename the book (or move it to another folder).
3. Reopen it: it starts from the beginning and the bookmark list is empty. The old bookmarks JSON is left orphaned on the card.
4. With "Move finished books to Read folder" enabled, finish a book that has bookmarks: the position survives (cache dir was renamed) but the bookmarks are gone.

**Verification (after):**
1. Same steps 1-3: the book resumes on the same page and the bookmark is listed.
2. WebDAV: `MOVE` the book to a new name, overwriting an existing file at the destination; reopen on the device: position and bookmarks follow the book, and nothing from the overwritten file remains.
3. WebDAV `COPY` onto an existing file: the copy starts fresh (no phantom progress from the file it replaced).
4. Regression: delete from the files page and from the on-device browser still removes the cache as before (unchanged code paths).

The on-device file browser has no rename/move, so nothing to route there.

**Memory / flash impact:** RAM unchanged (17.1%, 56140 B). Flash +1.8 KB (5401083 → 5402893 B on `gh_release`, mostly the new helper and its log strings). `bin/clang-format-fix`, `pio check` and the native unit tests (173/173) pass locally.

---

### AI Usage

Did you use AI tools to help write this code? _**YES**_ (Claude Code; the bugs were found by a code audit, the diff was reviewed and built by hand before opening this PR.)

https://claude.ai/code/session_017HaL1cqtG2c3L475nYDJCc

### DIFF
--- a/lib/FsHelpers/FsHelpers.cpp
+++ b/lib/FsHelpers/FsHelpers.cpp
@@ -174,6 +174,10 @@ bool hasTxtExtension(std::string_view fileName) { return checkFileExtension(file
 
 bool hasMarkdownExtension(std::string_view fileName) { return checkFileExtension(fileName, ".md"); }
 
+bool hasPlainTextBookExtension(std::string_view fileName) {
+  return hasTxtExtension(fileName) || hasMarkdownExtension(fileName);
+}
+
 bool hasCssExtension(std::string_view fileName) { return checkFileExtension(fileName, ".css"); }
 
 std::string extractFolderPath(const std::string& filePath) {
--- a/lib/FsHelpers/FsHelpers.h
+++ b/lib/FsHelpers/FsHelpers.h
@@ -64,6 +64,13 @@ inline bool hasTxtExtension(const String& fileName) {
 // Check for .md extension (case-insensitive)
 bool hasMarkdownExtension(std::string_view fileName);
 
+// Check for .txt or .md extension (case-insensitive): the plain-text book formats that
+// share TxtReaderActivity and the same on-disk cache layout (see Txt::getCachePath()).
+bool hasPlainTextBookExtension(std::string_view fileName);
+inline bool hasPlainTextBookExtension(const String& fileName) {
+  return hasPlainTextBookExtension(std::string_view{fileName.c_str(), fileName.length()});
+}
+
 // Check for .css extension (case-insensitive)
 bool hasCssExtension(std::string_view fileName);
 inline bool hasCssExtension(const String& fileName) {
--- a/src/activities/reader/EpubReaderActivity.cpp
+++ b/src/activities/reader/EpubReaderActivity.cpp
@@ -41,6 +41,7 @@
 #include "activities/settings/TextSettingsActivity.h"
 #include "components/UITheme.h"
 #include "fontIds.h"
+#include "util/BookDataMove.h"
 #include "util/BookmarkUtil.h"
 #include "util/ButtonNavigator.h"
 #include "util/ScreenshotUtil.h"
@@ -124,22 +125,14 @@ std::string buildReadFolderDestination(const std::string& srcPath) {
   return dstPath;
 }
 
-void moveFinishedBookToReadFolder(const std::string& srcPath, const std::string& dstPath,
-                                  const std::string& oldCachePath) {
+void moveFinishedBookToReadFolder(const std::string& srcPath, const std::string& dstPath) {
   LOG_INF("ERS", "Moving finished epub: %s -> %s", srcPath.c_str(), dstPath.c_str());
   if (!Storage.rename(srcPath.c_str(), dstPath.c_str())) {
     LOG_ERR("ERS", "Failed to move finished book to '/Read' folder");
     return;
   }
 
-  const std::string newCachePath = "/.crosspoint/epub_" + std::to_string(std::hash<std::string>{}(dstPath));
-  if (!oldCachePath.empty() && Storage.exists(oldCachePath.c_str())) {
-    if (!Storage.rename(oldCachePath.c_str(), newCachePath.c_str())) {
-      LOG_ERR("ERS", "Failed to rename cache dir %s -> %s (non-fatal)", oldCachePath.c_str(), newCachePath.c_str());
-    }
-  }
-
-  RECENT_BOOKS.updatePath(srcPath, dstPath, oldCachePath, newCachePath);
+  moveBookData(srcPath, dstPath);
   if (APP_STATE.openEpubPath == srcPath) {
     APP_STATE.openEpubPath = dstPath;
     APP_STATE.saveToFile();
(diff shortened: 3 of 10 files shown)


s0lness:
Review follow-up, pushed as one commit:

- **Applied**: WebDAV MOVE no longer touches the destination's cache/bookmarks before the rename (`moveBookData` clears them after it succeeds); WebDAV COPY clears the destination's stale data after a successful copy, unconditionally (a leftover cache can exist even when no file is there); `moveBookData` skips the bookmarks removal/rename when the old and new paths flatten to the same bookmarks file; when the new path is not a supported book type it drops the recent-books entry instead of rewriting it with an empty cache path.
- **Not changed, on purpose**: propagating a `moveBookData` failure as a request failure. The file operation the client asked for has already succeeded at that point; reporting 500 would misdescribe a completed rename/move. Data migration is best effort and every failed step is logged. Before this PR that data was deleted unconditionally, so a partial migration is strictly better than the previous behavior.

lpla:
> Review follow-up, pushed as one commit:
> 
> * **Applied**: WebDAV MOVE no longer touches the destination's cache/bookmarks before the rename (`moveBookData` clears them after it succeeds); WebDAV COPY clears the destination's stale data after a successful copy, unconditionally (a leftover cache can exist even when no file is there); `moveBookData` skips the bookmarks removal/rename when the old and new paths flatten to the same bookmarks file; when the new path is not a supported book type it drops the recent-books entry instead of rewriting it with an empty cache path.
> * **Not changed, on purpose**: propagating a `moveBookData` failure as a request failure. The file operation the client asked for has already succeeded at that point; reporting 500 would misdescribe a completed rename/move. Data migration is best effort and every failed step is logged. Before this PR that data was deleted unconditionally, so a partial migration is strictly better than the previous behavior.

@s0lness The best-effort policy makes sense to me: once the requested file move has succeeded, returning an HTTP error because auxiliary cache or bookmark migration failed would misreport the actual filesystem operation.

Given that policy, I think the cleanest follow-up is to make `moveBookData()` explicitly best-effort instead of returning a `bool`. All four callers currently ignore the result, and the result does not cover every possible cleanup failure anyway. The per-step logging already provides the useful diagnostic information.

I also agree that `RECENT_BOOKS` should follow the successfully moved file even if an auxiliary cache/bookmark rename fails; leaving it pointed at the now-missing old path would be worse.

Could you change the function to `void` and document the best-effort contract briefly? That would make the behavior and API agree, without pretending the already-completed file operation can be rolled back transactionally.

s0lness:
Done in 88f95dbf: `moveBookData()` is now `void`, with the best-effort contract documented on the declaration (called after the file move succeeded; cache dir, bookmarks file and recent-books entry migrated independently; failures only logged; the recent-books entry always follows the new path). No caller used the return value, so nothing else changed. Build, cppcheck and unit tests pass locally.

lpla:
One supported-format case still seems to be missing here, the Markdown books.

`ReaderActivity` routes both `.txt` and `.md` through `TxtReaderActivity`, and `RecentBooksStore` also recognizes both extensions. However, `bookCachePathFor()` only checks `hasTxtExtension()`.

As a result, moving or renaming an `.md` book treats the destination as unsupported: its TXT cache is not migrated and its recent-books entry is removed.

Could you include `FsHelpers::hasMarkdownExtension(path)` in the TXT branch and add a focused `.md` → `.md` move/rename case?

s0lness:
Good catch, fixed in 0de847f1. Rather than repeating the `hasTxtExtension || hasMarkdownExtension` pair a fourth time (ReaderActivity, RecentBooksStore and UITheme already spell it out), I added `FsHelpers::hasPlainTextBookExtension()` and used it in `bookCachePathFor()`. `clearBookCache()` in BookCacheUtils had the same gap (a deleted `.md` book never had its cache cleared, and `moveBookData` relies on it for stale destination data), so it uses the predicate too. Test: new `test/fs_helpers` suite covering `book.md`, `BOOK.MD`, `notes.txt` (true) and `book.epub`, `book.bak` (false). Testing `bookCachePathFor()` itself natively would need stubs for HalStorage and the Epub/Xtc/Txt classes, which the test tree does not have.

Note for merging: #3353 also introduces `test/fs_helpers` (same CMake target, different cases). Whichever of the two lands second will need a trivial rebase to merge the two test files; I will do it as soon as the first one is in.

Kile:
This is such a big QOL improvement! Is there any way/best effort to restore progress, bookmarks etc if the file was moved in the non internal view viewer (eg laptop)? Using cached info like metadata and file size on startup or even a dedicated button if you wanna minimise overhead? Probably out of scope for this PR but I prefer moving my files on my laptop and that would really help me out as well.

s0lness:
That's a good point didn't think about this, I'll mull over it a bit.

s0lness:
Not a bad idea, but I think it's too much for this PR. Handling a rename done from the laptop properly means either keying the cache by content instead of path (then two copies of the same book share one progress, and every existing cache needs a migration step so nobody loses their position on update), or a heuristic match on title/author when the cache is missing. Both go well beyond a move/rename fix, and renaming from the laptop doesn't seem to happen that often anyway. Happy to revisit in a separate issue if it comes up more.

Kile:
Fair enough yeah it's rather a bigger thing. Thanks for considering though and maybe keeping it in mind when deciding what to work on in the future :)

PR #3359, opened 2026-09-03
fix: regenerate stale built-in font content-hash IDs

YuunJiee:
## Summary

* **What is the goal of this PR?** Fix `src/fontIds.h`'s content-hash font IDs, which have been out of sync with the actual built-in font files since #3144.
  - Regenerated `src/fontIds.h` via `lib/EpdFont/scripts/build-font-ids.sh` so all 11 font IDs reflect the current content of the built-in font header files they're supposed to hash.
  - Fixed `build-font-ids.sh` itself: it never emitted the `static_assert` sentinel-guard block that's part of the committed file, so even a contributor who *did* remember to re-run it after changing font files would get an incomplete file and have to hand-patch the tail back in. The script now generates the assert block from the same list of font ID names, so its output is the complete, correct file.

## Scope Check

Please confirm:

## Additional Context

**Root cause**: #3144 ("perf: Reduce built-in font flash by 323 KB") rewrote nearly all built-in font `.h` files (34 files across notosans, notoserif, and ubuntu) but didn't regenerate `src/fontIds.h`. Verified concretely: recomputing `NOTOSERIF_12_FONT_ID`'s hash from the font files as they existed *before* #3144 reproduces the value currently committed (`85340443`); recomputing it from the current file contents gives a different value (`778531645`). So the constant in `develop` right now reflects font content that no longer exists. #3345 (Perso-Arabic UI glyphs) touched more of the same font files afterward and also didn't regenerate it, carrying the drift forward.

**Impact**: None functionally today — the IDs are only used as internal map keys, and the same (stale) value is used consistently to both produce and look up each font, so nothing currently breaks. This fixes the drift before it causes a real problem (e.g. any future tooling that assumes "same font content ⇒ same ID").

**Why now**: build-font-ids.sh requires Ruby, which wasn't available in the environment that produced #3144 and #3345 (my own environment initially lacked it too — installed it specifically to fix this).

**Testing**: `pio run -e default` builds successfully; flash usage unchanged (5,445,503 bytes / 83.1%), as expected since this only changes 32-bit ID constants, not the font data tables themselves.

**Flash/memory impact**: None (integer constants only, no data table changes).

---

### AI Usage

https://claude.ai/code/session_01DHb1364JDxjwkGW2p1siUF


Uri-Tauber (approved):
LGTM

PR #3360, opened 2026-09-03
feat: add Traditional Chinese (zh-Hant) UI support

YuunJiee:
## Summary

* **What is the goal of this PR?** Add Traditional Chinese (zh-Hant) as a supported UI language, without adding a full CJK glyph set to every default build's firmware flash (see maintainer feedback in this PR's history — a per-language flash tax on all builds was correctly rejected).
  - New translation file `lib/I18n/translations/chinese-traditional.yaml` (460 keys).
  - **Built-in UI fonts gain only 4 glyphs** — the characters in the language's own name, "繁體中文" (繁/體/中/文) — so it's legible and selectable in the language picker itself. **Total flash cost: ~1.1KB** across the 5 built-in UI font files (measured, `pio run -e default`), not the ~120KB a full per-language glyph set would cost. `gen_cjk_ui_intervals.py` now scans every translation file's `_language_name` (restricted to Han/Kana ranges) instead of one language's full string content, so this stays cheap if another CJK language is ever added.
  - **Full CJK menu rendering comes from an SD-card font instead of firmware flash.** Added `NotoSansCJKtc`/`NotoSerifCJKtc` to `lib/EpdFont/scripts/sd-fonts.yaml` as downloadable SD reading-font families — a declarative addition, same mechanism as the existing Noto Sans/Serif entries, no build-pipeline code changes.
  - `SdCardFontSystem` gains `ensureUiLanguageFallback()`: when the UI language is zh-Hant, it wires up a CJK SD font as the fallback for built-in UI text, independent of whatever reading font the user has chosen (previously, this fallback mechanism only activated when the *reading* font happened to have CJK coverage, e.g. for book titles in list views).
  - If the UI language needs CJK and neither the reading font nor an installed SD font provides it, the language selection reverts and shows a dismissible notice instead of leaving menus half-legible.
  - Regenerated `src/fontIds.h` for the 3 affected font IDs, on top of #3359.

## Scope Check

Please confirm:

## Additional Context

**Flash impact** (measured via `pio run -e default`, vs. `develop`):
- Font glyphs: ~1.1KB across the 5 built-in UI font files (vs. ~120KB for a full per-language CJK glyph set — see this PR's earlier revision).
- zh-Hant string table (`chinese-traditional.yaml`, 460 keys): ~8.57KB flash (measured by diffing `gen_i18n.py`'s reported string-table size with/without the file present). For reference, Persian's string table (already shipped, `develop`) is ~15.6KB — larger than zh-Hant's, since CJK needs fewer characters to say the same thing, offsetting UTF-8's 3-bytes-per-character cost.
- **Total: ~9.7KB net firmware impact**, and it doesn't touch `platformio.ini`/CI/the release process — one firmware, one release channel.

Build flash usage: 5,458,275 / 6,553,600 bytes (83.3%), consistent with `develop`.

**Testing** (desktop simulator, manual screenshots):
- Language picker: "繁體中文" renders correctly among all other languages.
- zh-Hant + a CJK reading font selected: full menu renders correctly (existing reading-font-driven fallback path).
- zh-Hant + no CJK reading font: full menu still renders correctly via the new independent fallback.
- Switching reading fonts / toggling UI language back and forth: no lost fallback registration, no crash (fixed a font-ID collision and two renderer-state-scoping bugs along the way).
- No CJK SD font installed: selecting zh-Hant shows a warning and reverts to the previous language instead of leaving the UI unreadable.
- `pio run -e default`: builds successfully.

  

---

### AI Usage

https://claude.ai/code/session_01YUYVdkAQgB49Yiy25yyccW


Uri-Tauber:
Quoting myself (https://github.com/crosspoint-reader/crosspoint-reader/discussions/1757#discussioncomment-17997311)
> My point of view is we cant add CJK fonts to the firmware - they simply take too much space. As a direct outcome, we won't accept CJK UI translations.
> But apart from that, we're open to CJK-support PRs. I think it would be awesome if bilanguagenal users could have upstream crosspoint as the way to go.

YuunJiee:
Thanks for the context — that's a fair constraint, and it makes sense not to want a per-CJK-language flash tax baked into every default build.

I've reworked this PR to respect that: the built-in UI fonts now only gain **4 additional glyphs** total ("繁體中文", the language's own name) — **~1.1KB across all 5 UI font files**, vs. the ~120KB a full CJK UI glyph set would have cost. That's just enough for the language to be legible and selectable in the language picker itself; every other translated string still renders via `tr()`/`I18N` exactly like any other language.

The actual CJK menu rendering comes from the SD-card font system instead of firmware flash: I've added `NotoSansCJKtc` / `NotoSerifCJKtc` as downloadable SD reading-font families (declarative addition to `sd-fonts.yaml`, no new code — the same mechanism as the existing Noto Sans/Serif entries), and extended `SdCardFontSystem` so a CJK UI language automatically uses whichever CJK SD font is installed as a fallback for menu text — independent of whatever reading font the user has chosen. If no CJK SD font is installed, selecting the language reverts back with a notice instead of leaving the UI half-legible.

Net firmware impact: ~1.1KB, not ~120KB.

Pushed the updated approach and rewrote the PR description to match. Let me know if you'd want anything about the fallback behavior (e.g. which family it picks by default) done differently.

Uri-Tauber:
Hmm. 
How much space the UI strings take?

Uri-Tauber:
We need to think about this. We already decided against any CJK UI fonts/strings.
But I'm willing to rethink this.

Lets draft this for now, and we'll see later.

YuunJiee:
Good question — I hadn't measured that separately. Just did:

- Font glyphs (previous number): ~1.1KB
- zh-Hant string table (`chinese-traditional.yaml`, 460 keys): ~8.57KB flash (measured by removing the file and diffing `gen_i18n.py`'s reported string-table size)
- **Total: ~9.7KB**

For reference, I measured Persian's string table the same way (already shipped, on `develop`): **~15.6KB** — larger than zh-Hant's. The raw YAML source size follows the same pattern: `chinese-traditional.yaml` (17.6KB) is smaller than `arabic.yaml` (22.6KB), `persian.yaml` (24.6KB), and `russian.yaml` (24.0KB) — Chinese needs fewer characters to say the same thing, which offsets UTF-8's 3-bytes-per-CJK-character cost.

Happy to leave this as a draft while you think it over — no rush on my end.

brianhuster:
> My point of view is we cant add CJK fonts to the firmware - they simply take too much space. As a direct outcome, we won't accept CJK UI translation

As I understand, the "built-in Chinese font" here only has 4 glyphs for the only 4 characters in the language name "繁 體 中 文" which means traditional Chinese. It takes much less space than any other built-in fonts, even Latin font. When users select 繁體中文 for the first time, they will be prompted to install a traditional Chinese font, only after they have done that that they can use Chinese UI. I think it is a smart design that Crosspoint needs to consider anyway, because it is not just CJK scripts that have a lot of glyphs, some complex Indic scripts like Devanagari (used for Hindi, Nepalese, etc) has more than 1000 ligatures, which likely can't be fit in a built-in font

YuunJiee:
Thanks for the summary, that's exactly the design. And great point about Devanagari/complex Indic scripts — this pattern (built-in font covers only the language's own name for the picker, everything else via SD-installed fallback) generalizes well beyond CJK. Happy to help think through that if it's useful down the line.

PR #3366, opened 2026-09-03
feat: Library view

Uri-Tauber:
# Summary

This PR introduces a new Library screen, letting users to easily find and sort their books.

The feature was created by @oreglio in #2878 and (very slowly) reviewed by me through a series of PRs. The goal of this PR is to allow the community to test the feature and provide feedback.

I changed few things afterwards:
- Replaced the jump-to-letter grid with a collapsible list. The previous grid only supported the English alphabet, which made it useless for non-Latin languages.
- Adjusted the controls to better align with the overall look and feel of the firmware.
- Other minor tweaks.

As this is a substantial feature, there are still to be refined. 
I need thorough testing and feedback from the team.

Simulator video
Screencast_20260904_143047


arthurgiroux (commented):
It seems that the utf8 implementation is growing more and more with each PR with composing, decomposing, case folding etc.
It might be good to look at what would be the footprint of integrating and existing library that already does all this work, for example https://github.com/JuliaStrings/utf8proc.

digitaltembo:
This looks like a nice feature! I've got a UX suggestion - skip the Title Z-A tab, just have Added, Title, and Author, but display a little arrow indicating sort direction next to the active tab, and if you select the tab again, toggle the sort direction
Screenshot 2026-09-03 at 10 13 43 PM

And then lots of scope creep, but as a future extension I would love to be able to have dynamic tabs based on tags in the ebook metadata, so having <dc:subject>tab:favorite</dc:subject> in one book adds a new 'Favorite' tab and <dc:subject>tab:genre:Sci-Fi</dc:subject> and <dc:subject>tab:genre:Fantasy</dc:subject> in two books adds a new 'Genre' tab grouped by either Sci-Fi or Fantasy

itsthisjustin:
> This looks like a nice feature! I've got a UX suggestion - skip the Title Z-A tab, just have Added, Title, and Author, but display a little arrow indicating sort direction next to the active tab, and if you select the tab again, toggle the sort direction
> Screenshot 2026-09-03 at 10 13 43 PM
> 
> And then lots of scope creep, but as a future extension I would love to be able to have dynamic tabs based on tags in the ebook metadata, so having <dc:subject>tab:favorite</dc:subject> in one book adds a new 'Favorite' tab and <dc:subject>tab:genre:Sci-Fi</dc:subject> and <dc:subject>tab:genre:Fantasy</dc:subject> in two books adds a new 'Genre' tab grouped by either Sci-Fi or Fantasy

In a similar vein I'd suggest the first tab be all files and be sorted by date added and have an arrow as well for flipping that order so all 3 tabs are simply different sorts ascending and descending. Imo for this feature to exist it would need to simply replace the browse files feature so we don't have 3 ways to view files

Uri-Tauber:
> It seems that the utf8 implementation is growing more and more with each PR with composing, decomposing, case folding etc.
> It might be good to look at what would be the footprint of integrating and existing library that already does all this work, for example [JuliaStrings/utf8proc](https://github.com/JuliaStrings/utf8proc).

Thats a good idea. I will look into this next week. Feel free to open PoC PR if you want.

Uri-Tauber:
> And then lots of scope creep, but as a future extension I would love to be able to have dynamic tabs based on tags in the ebook metadata, so having dc:subjecttab:favorite</dc:subject> in one book adds a new 'Favorite' tab and dc:subjecttab:genre:Sci-Fi</dc:subject> and dc:subjecttab:genre:Fantasy</dc:subject> in two books adds a new 'Genre' tab grouped by either Sci-Fi or Fantasy

I like the idea, but I don't think we want to scan tags across the whole library just to decide which tabs to show - that's a fair bit of added complexity and could have unwanted side-effects. 
My idea is to have a settings menu where you pick your preferred sorting scheme.
e.g. one person sorts by date/name/author, another by recently-opened/language/series. We could also fold the "recent books" screen into this, so there's just one place to manage how your library is organized.

@itsthisjustin @digitaltembo @oreglio What do you think?

Uri-Tauber:
> In a similar vein I'd suggest the first tab be all files and be sorted by date added and have an arrow as well for flipping that order so all 3 tabs are simply different sorts ascending and descending. Imo for this feature to exist it would need to simply replace the browse files feature so we don't have 3 ways to view files

Done. 

image

I updated the video in the PR description.

oreglio:
>My idea is to have a settings menu where you pick your preferred sorting scheme.

@Uri-Tauber I find that to be a good idea, it's elegant and easy

fazo96:
Gave a quick shot at this using the build from the CI on my x4 pro, it works great. It took only a few seconds to index 198 books, and then it opens instantly when using it again. Search is just instant, and the keyboard is surprisingly usable. Wow!

Also this build, compared to the beta I flashed a few days ago, seems to be more successful in rendering covers and the dictionary does not give the oom error anymore.

hamaluik:
I've updated my PR for adding series to the browsing: #3059 (targeting this feature branch), I ended up just porting the feature since so much of the upstream had changed. I haven't dug into why, but whatever the upstream changes were they make building the library quite a bit slower than oreglio's original PRs (~30 seconds on my device & library vs. the original that IIRC was < 10s).

Uri-Tauber:
@hamaluik Thanks for reporting. I will look into this tomorrow. 
Just to confirm, with the same library, the latest commit in #2878 take ~10 secs, and the latest commit here take ~30 secs? How many books you have in your lib?

hamaluik:
163 books. I don't have the original numbers but I can flash them if we need to measure for sure. I don't have time to dig into at the moment (I can later this weekend), but claude thinks it's because `stageRecord()` in the library builder parse's every epub metadata before reconciliation so its effectively rescanning metadata that it (should) know hasn't changed.

Uri-Tauber:
If you could give me exact commits hashs I can test, it would be best.
30 secs for 163 books is indeed way too much.

lib/LibraryIndex/LibraryText.h
```diff
@@ -0,0 +1,120 @@
...
+// Longest author key written into an index record. Sized so a key fits the
+// record's fixed field; measured to collide 0 times over a 69-book library.
+inline constexpr size_t AUTHOR_KEY_MAX_BYTES = 12;
+
+// Join an indexed folder with its basename using exactly one separator. The
+// root folder is stored as "/"; treating it like a normal directory would
+// reconstruct "//book.epub", which hashes differently and may not open.
+std::string joinLibraryPath(std::string_view folder, std::string_view name);
+
+// Casefold and strip diacritics for matching and sorting.
```
digitaltembo, line 44: I would expect stripping diacritics to not be universally valid for sorting, do you see locale-based alphabetic sorting as a possibility in the future? I see why it would be preferable from a performance standpoint and for searching, so I kind of think both the original and the folded strings/prefixes make sense to store

lib/Epub/Epub/parsers/ContentOpfParser.cpp
```diff
@@ -339,21 +384,22 @@ void XMLCALL ContentOpfParser::startElement(void* userData, const XML_Char* name
...
-      self->author.append(", ");  // Add separator for multiple authors
-    }
-    self->author.append(s, len);
+    appendMetadataText(self->author, s, len, self->metadataSpacePending, &self->authorSeparatorPending);
     return;
   }
 
   if (self->state == IN_BOOK_LANGUAGE) {
-    self->language.append(s, len);
+    appendMetadataText(self->language, s, len, self->metadataSpacePending);
```
digitaltembo, line 402: What kind of restrictions are there on the metadata that we want to cache? Saving author-sort, title, sort, subjects, genre, maybe pagecount from the metadata seems like it would be helpful. Possibly too heavy to live in the heap, but maybe in other index files?
I'm also just looking at standard-ebooks metadata rules to try and get a sense of what might be supplied by a "good" ebook, I assume it is wildly inconsistent when it comes to books that people actually have.

Uri-Tauber:
I did a bunch of optimizations, which should improve indexing time. CC @hamaluik
The biggest one is rebuilding index try to use the existing files rather than starting from zero, so after the first cold indexing, adding new books should take no time.

I also improved somewhat the first cold index. On my x4 with 259 epubs, I measured ~25 secs on 51da68f and 20.9 secs on 5077fb0.

Another thing is now we can support up to 4096 books with in-place bounded sorting. In theory, at least. I still need to stress-test it.

I also worked on my plan 
> My idea is to have a settings menu where you pick your preferred sorting scheme.

But it ended being much bigger rework than I thought, and the indexing time was significantly slower, so I hold this direction for now.
You can check it in https://github.com/Uri-Tauber/crosspoint-reader-fork/tree/feat/library-view.

Overall, I'm pretty happy with how the new library is right now. 
What I need is some help with stress testing, and UI/UX review from @itsthisjustin or @uxjulia.

hamaluik:
Welp I didn't get a chance to look over the weekend, but I just came today to see your fixes.. I ran them, my library rebuilds in 3.6 seconds now! :tada:

Rebased my series addition to this feature as well, it is still working against the upstream changes.

itsthisjustin:
> > In a similar vein I'd suggest the first tab be all files and be sorted by date added and have an arrow as well for flipping that order so all 3 tabs are simply different sorts ascending and descending. Imo for this feature to exist it would need to simply replace the browse files feature so we don't have 3 ways to view files
> 
> Done. 
> 
> image
> 
> I updated the video in the PR description.

If this screenshot is still current, the tab bar and the header line aren't supposed to have that gap. I'd like to get this merged as it'll help us get a cover view I bet with shared apis but I feel it should replace browse files via a setting as adding a new menu item especially as with opds/plugins active we are out of space for menu items vertically

Uri-Tauber:
@itsthisjustin Good catch. I didnt noticed that!

image

> but I feel it should replace browse files via a setting as adding a new menu item especially as with opds/plugins active we are out of space for menu items vertically

On Lyra we have enough space for six entries. Lyra Ext and Roundedraff are a problem.
I think **file browser** and **library** have two completely different use cases. 
What we can do it to fold **Recent** into **Library**.

Does SD plugins add new menu item?

itsthisjustin:
> @itsthisjustin Good catch. I didnt noticed that!
> 
> image
> 
> > but I feel it should replace browse files via a setting as adding a new menu item especially as with opds/plugins active we are out of space for menu items vertically
> 
> On Lyra we have enough space for six entries. Lyra Ext and Roundedraff are a problem.
> I think **file browser** and **library** have two completely different use cases. 
> What we can do it to fold **Recent** into **Library**.
> 
> Does SD plugins add new menu item?

No and that's why I noticed this. Cause I roll opds into plugins IF plugins are active

Uri-Tauber:
Screencast_20260910_082534

I moved Recent books into Library to address @itsthisjustin point. 

What I still need to do before merging: testing with different devices. Stress-testing with huge libraries. Testing with corrupted epubs - making sure it fail gracefully. Looking into CodeRabbit comments - checking if it found anything valuable. 

Help would be appreciated!

itsthisjustin:
> Screencast_20260910_082534
> 
> I moved Recent books into Library to address @itsthisjustin point. 
> 
> What I still need to do before merging: testing with different devices. Stress-testing with huge libraries. Testing with corrupted epubs - making sure it fail gracefully. Looking into CodeRabbit comments - checking if it found anything valuable. 
> 
> Help would be appreciated! 

Now I know why you wanted the plugins icon changed haha. We can do that. This looks good

Uri-Tauber:
~found a weird bug when adding new books. hold with merging.~ FIXED

muhas:
Maybe toggle sorting directions by long-pressing the button, and by short-pressing it, just like in the settings, move to the next element

The first transition to the list of books is now possible only with the side buttons, and in their list the buttons under the screen already work - the difference in button behavior in the settings and library is very confusing 

The inconsistency in controls is annoying

itsthisjustin:
> Maybe toggle sorting directions by long-pressing the button, and by short-pressing it, just like in the settings, move to the next element
> 
> The first transition to the list of books is now possible only with the side buttons, and in their list the buttons under the screen already work - the difference in button behavior in the settings and library is very confusing 
> 
> The inconsistency in controls is annoying

Agree with this

Uri-Tauber:
The inconsistency annoy me too, but long press open search on non-touch device. 
I thought to change the Settings navigation, but that probably annoy people. 
Open to suggestions.

itsthisjustin:
> The inconsistency annoy me too, but long press open search on non-touch device. 
> I thought to change the Settings navigation, but that probably annoy people. 
> Open to suggestions.

For OPDS search is just done by going up when you're already at the top

itsthisjustin:
Been using this today. It's weird to me that I can't press and hold to delete files from the other views besides recent. It was the first thing I wanted to do after finishing a book was remove it and library view didn't let me. Can I add that vs press and hold giving me an alphabet?

itsthisjustin:
Also I have a ton of files showing up like this which isn't useful. Also how do I refresh the library? Added two books, they dont show up at all.

oreglio:
> Been using this today. It's weird to me that I can't press and hold to delete files from the other views besides recent. It was the first thing I wanted to do after finishing a book was remove it and library view didn't let me. Can I add that vs press and hold giving me an alphabet?

The first time I pushed the library there was a limited set of features, I added many more after but I did not wanted to make the scope of the original PR larger, I thought I'd wait for the feature to pass and add the other features later. This was part of the features along with some shortcuts, favorites view, multiline etc. It has to be added along in the current version, I don't want to interfere with Uri's work but I can do it if necessary.

> Also I have a ton of files showing up like this which isn't useful. Also how do I refresh the library? Added two books, they dont show up at all.

I tried different version of the firmware and I never saw that pop up. Is that things generated by the feature ?

itsthisjustin:
> > Been using this today. It's weird to me that I can't press and hold to delete files from the other views besides recent. It was the first thing I wanted to do after finishing a book was remove it and library view didn't let me. Can I add that vs press and hold giving me an alphabet?
> 
> The first time I pushed the library there was a limited set of features, I added many more after but I did not wanted to make the scope of the original PR larger, I thought I'd wait for the feature to pass and add the other features later. This was part of the features along with some shortcuts, favorites view, multiline etc. It has to be added along in the current version, I don't want to interfere with Uri's work but I can do it if necessary.
> 
> > Also I have a ton of files showing up like this which isn't useful. Also how do I refresh the library? Added two books, they dont show up at all.
> 
> I tried different version of the firmware and I never saw that pop up. Is that things generated by the feature ?

Thats just from this branch. Im wondering what the blank name titles are. Not sure if it's hidden files or what.

itsthisjustin:
And I think its ok to keep improving here. Removing files would be needed at a minimum at least. It's weird ux to be missing

Uri-Tauber:
> Thats just from this branch. Im wondering what the blank name titles are. Not sure if it's hidden files or what.

Non-epub fall back to filename. The entries with just `.txt` are weird. Possibly a bug with parsing filenames. I will try to reproduce.

muhas:
> Simplify library navigation to unified ring model

Wow! Thanks 

This navigation is really clearer!

Long press left/right on header menu do not switch to the next menu item as in the settings

itsthisjustin:
> > Simplify library navigation to unified ring model
> 
> Wow! Thanks 
> 
> This navigation is really clearer!
> 
> Long press left/right on header menu do not switch to the next menu item as in the settings

Honestly had no idea settings did that haha

digitaltembo:
Maybe could ignore this idea but: what if "recent" was just the top of "added"? That is, combine the two lists into effectively a single list by modified time, where opening to read counts as a modification.

itsthisjustin:
> Maybe could ignore this idea but: what if "recent" was just the top of "added"? That is, combine the two lists into effectively a single list by modified time, where opening to read counts as a modification. 

This would allow the recents list to be much longer which people have asked for but then removes the ability to remove them from recents. I'm not against it

itsthisjustin:
> Maybe could ignore this idea but: what if "recent" was just the top of "added"? That is, combine the two lists into effectively a single list by modified time, where opening to read counts as a modification. 

Would recent be on top or intermixed? I.e. if I read a book but then added 3 books to my sd card, this would push the actual recently read books down the list

itsthisjustin:
> It seems that the utf8 implementation is growing more and more with each PR with composing, decomposing, case folding etc. It might be good to look at what would be the footprint of integrating and existing library that already does all this work, for example [JuliaStrings/utf8proc](https://github.com/JuliaStrings/utf8proc).

Could you open a PR for this?

itsthisjustin (approved):

itsthisjustin:
Gonna merge this after testing on 4 different devices and fixing a bunch of issues with tabs and lists along the way. We can RC it and any further improvements to the library can go in a fresh PR

Uri-Tauber:
> > It seems that the utf8 implementation is growing more and more with each PR with composing, decomposing, case folding etc. It might be good to look at what would be the footprint of integrating and existing library that already does all this work, for example [JuliaStrings/utf8proc](https://github.com/JuliaStrings/utf8proc?rgh-link-date=2026-09-14T05%3A05%3A32Z).
> 
> Could you open a PR for this?

The problem is `Utf8` currently mixes four responsibilities: decoding/encoding, buffer-boundary helpers, Unicode classification tailored to rendering, and partial NFC/decomposition.
So a cleanup is definitely a good idea, but we need to be super careful not to break anything.

EDIT: utf8proc is huge. Adding it without serious trancuation would add the firmware ~300kb.

muhas:
I tried commit ad949bd and his work fine, but nightly build develop tree show library as unsorted and could not rebuild the index

I delete index file (I thought maybe it was him), but nothing has changed
The new index is half the size of the ad949bd build (≈750Kb, previously, about one and a half megabytes on the same books)

Uri-Tauber:
@muhas try to delete the index, reboot and immediately go to settings->system->rebuild library index.

How many books do you have? What languages are the titles?

muhas:
> @muhas try to delete the index, reboot and immediately go to settings->system->rebuild library index.
> 
> How many books do you have? What languages are the titles?

Russian language titles ≈7800 books  (4096 were previously indexed in ad949bd when add navigation using the select key in the header, ≈4min create index)
After deleting index and rebooting, the rebuild from the settings works, but the library unsorted 4100 books

I have now left 50 books on the sdcard, everything worked fine

Uri-Tauber:
Yeah. 7800 files will do this. 4096 is the hard cap for now.

PR #1870, opened 2026-05-08
chore: wire native simulator integration

sabraman:
## Summary

Reworks the native emulator PR into a thin CrossPoint integration for the shared `crosspoint-simulator` library.

The reusable SDL, HAL, storage, Wi-Fi, WebServer, WebSockets, NetworkClient, OTA, and firmware-update simulator pieces now live in `uxjulia/crosspoint-simulator`. This PR keeps only the CrossPoint-specific glue needed to compile and run the firmware against that simulator.

## What changed

- Adds `[env:simulator]` to PlatformIO.
- Consumes `crosspoint-simulator` from `https://github.com/uxjulia/crosspoint-simulator.git`.
- Compiles CrossPoint real `CrossPointWebServer.cpp` and `WebDAVHandler.cpp` against simulator WebServer/WebSockets/NetworkClient shims.
- Excludes destructive firmware/OTA implementation files from the native simulator build; simulator provides non-destructive stubs.
- Shows non-80 simulator ports in the Web Server QR/URL UI.
- Handles `OtaUpdater::NO_UPDATE` as an explicit no-update state so the simulator OTA stub does not render as a failure.

## Dependency

- `uxjulia/crosspoint-simulator#8` is merged.
- This PR now points at the upstream simulator repository instead of the temporary fork branch.

## Validation

- `git diff --check` passes.
- `~/.platformio/penv/bin/pio run -e simulator` passes against `simulator@1.0.0+sha.2532823`.
- `~/.platformio/penv/bin/pio run` passes.

## Notes

This force-push intentionally removed the previous in-tree emulator implementation from this PR. The goal is to keep reusable simulator infrastructure outside this firmware repo and leave only firmware-specific integration here.

PR #3379, opened 2026-09-05
test: add X3 and X4 simulator smoke checks

marcusquinn:
## Summary

Add a small X3/X4 simulator smoke check for PR review, using the official external simulator rather than another emulator implementation.

- A separate `platformio.simulator.ini` pins the simulator revision and supplies Linux/macOS host settings without changing device/release environments. Runtime verification below is on Linux only.
- `.github/workflows/simulator.yml` runs on disposable Ubuntu runners, with read-only permissions, no secrets and no persisted checkout credentials. Manual runs can test an upstream SHA or PR ref with the same harness.
- `scripts/simulator_smoke.py` generates an original EPUB in a fresh virtual SD card, checks Home, opens via persisted resume state, turns next/previous/next, exits the reader normally and restarts to reopen the saved page.
- Failures include wrong dimensions, blank/missing screens, incorrect page transitions, changed round-trip content, wrong persisted progress, crashes and timeouts. Logs, screenshots, the fixture, results and exact revisions are uploaded.

This complements #1870 rather than replacing its developer-facing integration. It leaves the SDL/HAL implementation in `crosspoint-simulator`, does not add an allocator/OOM simulator, and does not change branch protection or replace existing firmware/unit checks.

## Where to view the screenshots

1. Open the [successful upstream X3/X4 simulator run](https://github.com/crosspoint-reader/crosspoint-reader/actions/runs/33939859185).
2. Scroll to **Artifacts** and download `simulator-x3-33939859185-1` or `simulator-x4-33939859185-1` (GitHub sign-in is required).
3. Unzip the download and open the PNGs in `smoke/`:
   - `home.png` — Home screen.
   - `page-1.png` and `page-2.png` — rendered book and page-forward result.
   - `page-1-back.png` and `page-2-again.png` — next/previous round trip.
   - `page-2-reopened.png` — saved page after restarting.

`smoke/result.json` contains the assertions' results; `home.log`, `reading.log` and `reopen.log` contain the execution logs. These are **simulator captures, not device photos**. GitHub provides artifact ZIP downloads rather than inline image previews; retention is **14 days from each run**.

## Verification

Successful hosted run: https://github.com/marcusquinn/crosspoint-reader/actions/runs/33937839330

Firmware/harness head: `59bda6bc38fa074d2b2b3b2b03dd877c7fc5464d` (production source based on `9d2f234e`). Simulator: `c55f168bc0e677fdb32312c8be4b5874469465e6`.

| Profile | Display | Page transitions (zero-based) | Persisted/reopened page | Result |
| --- | --- | --- | --- | --- |
| X3 | 528×792 | 0 → 1 → 0 → 1 | 1 of 26 | Pass |
| X4 | 480×800 | 0 → 1 → 0 → 1 | 1 of 28 | Pass |

Both used the same generated EPUB SHA-256: `78fbeec2c40051d75a17ea8a6335e50482abeadd55aa8266d6d68e7cb7a4276c`. Artifacts contain the screenshots and `result.json` per device (14-day retention).

Additional checks: `actionlint`, Python syntax compilation, generated XML validation and `git diff --check` passed. A local negative probe returning exit zero without a screenshot was rejected as intended. The initial hosted run also caught a test lifecycle mistake: quitting inside the reader correctly leaves its crash-loop guard set; the runner now presses Back before restarting instead of bypassing that guard.

## Scope Check

- [x] Not a new built-in theme.
- [x] Not a new external network connector.
- [x] This addresses development/PR verification, not an existing stock-firmware reading feature; it reuses the official simulator.
- [x] No changes to `freeink-sdk/`, `lib/hal/`, bootloader, OTA or recovery implementations.

## Additional Context / Runtime Testing

Simulator runtime verified on hosted Linux; **not tested on physical hardware**. No firmware RAM/flash impact. This does not establish ESP32 memory limits, real page-turn speed, e-ink ghosting/waveforms, SD timing, power behaviour or firmware flashing safety. Default simulator image shims are unchanged; the fixture tests text.

The test uses fixed-time input/capture schedules and fails if expected states are missed; it does not retry failures into passes. Screenshot equality is within a device run, excluding the status bar, not a cross-platform golden-image gate. Reopening uses real saved progress; initial opening is seeded through the normal resume path, not file-browser selection. The separate macOS build path is available, but screenshot assertions target 1x Linux rather than Retina.

### AI Usage

**YES.** AI-assisted implementation and source review. The verification above is from actual commands and hosted simulator runs; no hardware results are claimed.

### DIFF
--- a/.github/workflows/simulator.yml
+++ b/.github/workflows/simulator.yml
@@ -1,0 +1,88 @@
+name: Simulator smoke
+
+on:
+  pull_request:
+  push:
+    branches: [develop, 'ci/simulator-*']
+  workflow_dispatch:
+    inputs:
+      upstream_ref:
+        description: 'Optional upstream commit SHA or refs/pull/NUMBER/head to test with this harness'
+        required: false
+        type: string
+
+permissions:
+  contents: read
+
+concurrency:
+  group: simulator-${{ github.workflow }}-${{ github.ref }}-${{ inputs.upstream_ref }}
+  cancel-in-progress: true
+
+jobs:
+  smoke:
+    name: Simulator ${{ matrix.device }}
+    runs-on: ubuntu-24.04
+    timeout-minutes: 20
+    strategy:
+      fail-fast: false
+      matrix:
+        device: [x3, x4]
+    steps:
+      - uses: actions/checkout@v6
+        with:
+          path: harness
+          persist-credentials: false
+
+      - uses: actions/checkout@v6
+        with:
+          repository: ${{ inputs.upstream_ref && 'crosspoint-reader/crosspoint-reader' || github.repository }}
+          ref: ${{ inputs.upstream_ref || github.sha }}
+          path: firmware
+          submodules: recursive
+          persist-credentials: false
+
+      - uses: actions/setup-python@v6
+        with:
+          python-version: '3.13'
+
+      - name: Install native dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y libsdl2-dev libssl-dev xvfb xauth
+          python -m pip install platformio==6.1.19 PyYAML==6.0.3 Pillow==12.3.0
+
+      - name: Record provenance and install build-only configuration
+        run: |
+          mkdir -p artifacts
+          git -C harness rev-parse HEAD > artifacts/harness-sha.txt
+          git -C firmware rev-parse HEAD > artifacts/firmware-sha.txt
+          git -C firmware submodule status --recursive > artifacts/submodules.txt
+          cp harness/platformio.simulator.ini firmware/platformio.simulator.ini
+          cp harness/scripts/simulator_platform.py firmware/scripts/simulator_platform.py
+          cp harness/platformio.simulator.ini artifacts/
+
+      - name: Build simulator
+        working-directory: firmware
+        env:
+          DEVICE: ${{ matrix.device }}
+        run: |
+          set -o pipefail
+          pio run -c platformio.simulator.ini -e "simulator_$DEVICE" -j 2 2>&1 | tee ../artifacts/build.log
+
+      - name: Exercise reader and saved progress
+        env:
+          DEVICE: ${{ matrix.device }}
+          LIBGL_ALWAYS_SOFTWARE: '1'
+        run: |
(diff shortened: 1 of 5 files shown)


Uri-Tauber:
For the life of me, I cant find a real change in which this check will fail.

marcusquinn:
Fair question—I hadn't supplied a failing control. I've now run the **unchanged test and generated EPUB from this PR** against the original firmware and two deliberately injected faults in the real reader code:

| Firmware behaviour | Build X3 / X4 | Test X3 / X4 | What caught it |
| --- | --- | --- | --- |
| [Original PR head](https://github.com/marcusquinn/crosspoint-reader/actions/runs/34504516699) | Pass / Pass | Pass / Pass | — |
| [Forward skips a page (`++` → `+=2`)](https://github.com/marcusquinn/crosspoint-reader/actions/runs/34504520013) | Pass / Pass | **Fail / Fail** | Next then previous no longer reproduces the original page |
| [Progress saving reports success but omits the disk write](https://github.com/marcusquinn/crosspoint-reader/actions/runs/34504522539) | Pass / Pass | **Fail / Fail** | `Expected one persisted EPUB progress file` |

For the second fault, rendering and the success-log sequence still pass; the check fails on the actual missing file, not a missing log message.

Here is the X4 result after the same next/previous sequence—working firmware on the left, page-skip fault on the right:

Same NEXT/PREVIOUS sequence: working firmware returns to page 1; the injected fault returns to page 2

[Exact patches, SHAs, all six results and limitations](https://github.com/marcusquinn/crosspoint-reader/blob/944788ddad9cd1ca2f9d452416c18525f6db6986/docs/pr3379-evidence/README.md) are preserved outside the expiring Actions artifacts. No flashing or local setup is needed to inspect this.

These are **injected fault controls, not historical bugs**; the demonstration branches are not for merging or flashing. They establish narrow page-navigation/persistence coverage, not comprehensive rendering correctness or hardware behaviour. Consistently wrong typography could still pass. This PR's firmware/test head is unchanged.

marcusquinn:
My aim to share is; the more CI/CD can do preflight checks for you and others, the more time is preserved for the real hardware testing of issues that can only be verified that way. Just an idea, as I'm grateful for your work, and this is the area I have most interest in that might help you to have inspiration from to help the cause.

PR #3386, opened 2026-09-05
feat: add Waveshare ESP32-S3-ePaper-3.97 board support

pazyuk:
Adds the `ws397` envs for the **Waveshare ESP32-S3-ePaper-3.97** — a 3.97" 800x480 SSD1677 reader on an ESP32-S3-WROOM-1-N16R8, with an AXP2101 PMIC, 4-bit SDMMC, PCF85063 RTC and a QMI8658 IMU.

Brought up and tested on real hardware: it boots, mounts the card, renders and pages through an EPUB, sleeps and wakes.

> **Draft — depends on Free-Ink/freeink-sdk#79.** The board profile and the AXP2101 support live in the SDK, and the submodule here points at `3fe3936d62e6` on that PR's branch. CI cannot fetch that commit until the SDK PR merges; happy to re-point the submodule at the merged SHA and undraft on request.

## Firmware side of the port

- `platformio.ini`: `ws397` and `ws397-gh_release` envs. Follows the `sticky` shape (S3, `firmware_tuned`) plus `USE_BLOCK_DEVICE_INTERFACE=1` for the SDMMC card. Names its own `CROSSPOINT_VERSION`, as `x4pro`/`papermono` do, since `scripts/git_branch.py` only injects one for `default`/`sticky`.
- `FirmwareBoardTag.cpp`: board name `ws397`.
- `main.cpp`: a USB-power cold boot stays awake, as on the EEGO A4 — USB detect here is a PMIC register and charge termination reads as "no USB", so a post-flash reset would otherwise be misclassified as a USB cold boot and sleep immediately.

## Keyboard fix (second commit, useful beyond this board)

The on-screen keyboard is a 2D grid: Up/Down change row, Left/Right move along it. This board has a three-way rocker plus BOOT and wires no horizontal key, so only the first column of each row was reachable and text entry was impossible.

When the profile declares no Left/Right pin, Up/Down now walk the grid in reading order, rolling into the neighbouring row at either end. It keys off `BoardConfig::ACTIVE.input.left/right`, so boards that do wire a horizontal key are byte-identical.

## Why PSRAM stays off

The module is an N16R8, so the question is fair. The 48KB framebuffer fits in DRAM; nothing on this board's path allocates from SPIRAM (`FREEINK_FB_PSRAM` is for M5Paper/PaperMono, and the SSD1677 gray path uses plain `malloc`); the build sits at 20.3% RAM. Meanwhile `BOARD_HAS_PSRAM` would raise the idle CPU floor from 10 MHz to 80 MHz (`HalPowerManager::LOW_POWER_FREQ`) — a real cost on a battery reader whose dominant state is a static page. Leaving it off also keeps the env eligible for `firmware_tuned`.

## Flashing note

The board powers down a few seconds after reset unless firmware is running and holding the PMIC, so a plain `pio run -t upload` can lose the port mid-write. Flashing with the chip parked in the ROM downloader (`esptool --after no-reset`) works reliably. Documented in the SDK support doc.


tisanina:
Hi @pazyuk and @crosspoint-reader maintainers,

I completed physical validation of the Waveshare ESP32-S3-ePaper-3.97 port on a retail ESP32-S3-WROOM-1-N16R8 board.

### Validation summary

- **Display**: PASS — correct NO_FLIP orientation, sharp text, clean 4-level grayscale, no observed ghosting.
- **SDMMC**: PASS — card detection, mounting, EPUB parsing, cache creation, and reading.
- **AXP2101 / display rail / battery telemetry**: PASS.
- **RTC and IMU**: PASS — PCF85063 at 0x51, QMI8658 at 0x6B.
- **Rocker input and keyboard fallback**: PASS.
- **Deep sleep, BOOT wake, and custom sleep screens**: PASS.

I also tested the current PR changes against FreeInk at `ffcbb1c`, which includes the merged Waveshare SDK support.

### Bug: reader state is lost after sleep/wake

The Waveshare profile maps BOOT (GPIO0) as both Back and Power, and the same pin is the deep-sleep wake source.

This creates two problems when reading:
1. Holding BOOT to sleep triggers the Back long-press behaviour before the power sleep timer fires, returning to Home first.
2. Holding BOOT to wake leaves the key pressed during boot; the startup Back-button safety check then routes the device to Home instead of restoring the open reader session.

### Proposed fix

I prepared and hardware-validated a small, board-generic fix which detects when Back and Power share the same assigned GPIO.
- For shared Back/Power keys, Back is handled only after a short release; a long hold remains available for sleep.
- The boot-time held-Back fallback is skipped only for boards with that shared-key configuration.
- The separate Back/Power path remains unchanged.

**Hardware regression results**:
- Long BOOT press from a book: sleeps without returning to Home.
- Short BOOT press to wake: resumes the exact reading page.
- Short BOOT press while reading: returns to Home.
- Long BOOT press from Home: sleeps and wakes back to Home.

The change is formatted and available for review/cherry-pick here:
https://github.com/tisanina/crosspoint-reader/commit/680ccb79

If helpful, I can also provide the exact rebase/submodule update used for validation.

PR #3222, opened 2026-08-26
perf: optimize style-aware font prewarm

Uri-Tauber:
# Summary

This ports two small optimizations from #3083 by @serialx:

1. **Avoid doing the same font work multiple times.**
   Previously, when a page mixed regular, bold, italic, and bold-italic text, each style could prepare characters that actually belonged to the other styles too. The new code keeps them separate, so each font style only prepares the characters it really needs.

2. **Avoid temporary memory allocations while scanning a page.**
   The font scanner previously created and resized temporary strings while collecting characters. It now uses a small fixed-size workspace instead. This costs about 2 KB of permanently reserved RAM, but avoids repeatedly allocating and freeing memory while pages are rendered, which should make memory usage more predictable and reduce fragmentation over time.

## Results

On the mixed-style test book, normal page prewarming dropped from about **80–81 ms to 61–63 ms**, roughly a **20–25% reduction**. Total page rendering on those pages dropped from about **719 ms to 697 ms**.

The more expensive pages also improved noticeably: font preparation dropped from roughly **157–161 ms to 91–92 ms**, with total render time falling from about **950 ms to 810 ms**. The e-ink refresh itself is unchanged, so it still makes up most of the page-turn time.
Memory behavior also looks much healthier during the test. Before the change, the largest available continuous block of memory fell as low as about **78 KB**. After the change it remained around **106 KB** through the same workload. 
The trade-off is that the fixed workspace permanently reserves roughly **2 KB**, so this is mainly about reducing fragmentation rather than increasing headline free RAM.

For ordinary books that mostly use one font style, this should not make a noticeable difference. The benefit is mainly for pages that mix several font styles, plus more stable memory usage during longer reading sessions.

Most of the LoC are the new test suite. The net code is ~170LoC

---

### AI Usage

While CrossPoint doesn't have restrictions on AI tools in contributing, please be transparent about their usage as it
helps set the right context for reviewers.

Did you use AI tools to help write this code? _**< PARTIALLY >**_

PR #3397, opened 2026-09-06
chore: update pioarduino to 55.03.311

serialx:
## Summary

While preparing automated on-device testing for future font-rendering improvements, I tested the existing serial screenshot command on an Xteink X3 and found that transfers were being truncated.

The investigation pointed to an upstream HWCDC transmit bug. [espressif/arduino-esp32#12606](https://github.com/espressif/arduino-esp32/pull/12606) addresses data loss and transmit hangs during sustained writes. See the original report, [#12494](https://github.com/espressif/arduino-esp32/issues/12494).

This PR updates pioarduino from **55.03.37 to 55.03.311**, bringing Arduino-ESP32 **3.3.7 → 3.3.11** and ESP-IDF **5.5.2 → 5.5.5**, including that fix. I chose 3.3.11 to include the intervening maintenance updates as well, and validated its build compatibility and X3 heap impact. [Release notes](https://github.com/pioarduino/platform-espressif32/releases/tag/55.03.311).

On X3, the final configuration reduces total heap capacity by **720 bytes** compared with develop; the largest free block at Home is unchanged.

The build integration also needs updating so cached framework libraries, generated SDK configuration, and application linker flags remain consistent after the upgrade:

- Update framework caching to account for board memory settings. Clear stale generated SDK configuration on rebuilds so old values do not override the requested configuration, and restore prebuilt libraries when switching from a custom-framework profile.
- Align application linker flags with the rebuilt C3 SDK to work around a limitation in pioarduino 55.03.311's `custom_sdkconfig` support. It rebuilds SDK libraries and configuration headers but retains the packaged libc linker flags for the Arduino application. ESP-IDF 5.5.5 enables RAM-resident optimized libc by default, consuming SRAM that would otherwise be available to the C3 heap. Restoring the ROM libc configuration used with 3.3.7 recovers **1,024 bytes of heap on X3** compared with the new default, and requires adjusting both the SDK setting and the application linker flags. The hook supplies that missing linker adjustment.
- Preserve existing C3 configuration values where supported by the newer SDK, keep C3-specific settings separate from S3 profiles, and satisfy the newer Modbus configuration constraint.

## Cppcheck compatibility

pioarduino also updates the bundled cppcheck package from **2.11.0 to 2.20.1**. This project's CI fails on any unsuppressed low-, medium-, or high-severity finding, including style and performance diagnostics. The newer checker flags existing reader code, so the platform upgrade requires a code cleanup to pass those checks.

The cleanup is isolated in a **separate commit**. It makes stateless helpers static, tightens const interfaces, returns stored metadata by const reference, replaces search loops with standard algorithms, removes an unused local, and explicitly initializes the dictionary tag buffer. These changes preserve reader behavior and add no heap allocations.

The analyzer uses exhaustive analysis to avoid branch-limit notices. Diagnostics owned by the separately maintained FreeInk SDK are excluded, and cppcheck's checker-status summary is suppressed because it is informational output about the analyzer itself. CI retains its existing failure thresholds and installs `libpcre3`, which the packaged Linux cppcheck executable requires.

## Memory impact

Serial measurements on the same X3 at Home after a fresh boot, compared with pristine `develop` at `54337e6d`. The Arduino 3.3.11 values were reconfirmed on cleanup commit `13def4c3`:

| Metric (bytes) | Arduino 3.3.7 | Arduino 3.3.11, this configuration | Delta |
|---|---:|---:|---:|
| Total heap capacity | 268,560 | 267,840 | −720 |
| Free heap at Home | 101,488 | 100,724 | −764 |
| Minimum free since boot | 98,580 | 97,836 | −744 |
| Largest free block | 94,196 | 94,196 | 0 |

On C3, instruction RAM and data RAM share SRAM. Keeping libc in ROM avoids the additional RAM-resident implementations introduced by the newer SDK configuration. The 720-byte heap-capacity decrease consists of 512 bytes of additional IRAM reservation and 208 bytes of static data and alignment. Linker symbols confirm that the affected memory and string functions resolve to ROM.

Heap measurements were taken at Home; rendering performance was not benchmarked.

## Validation

- On the cleanup commit, cppcheck reports **zero findings locally and in Linux CI**. All four CI firmware builds (`default`, `sticky`, `x4pro`, and `papermono`), unit tests, and formatting checks pass. Local C3 and S3 builds also pass; the S3 SDK build emits const-qualifier warnings in ESP-IDF and its managed components.
- Before the cppcheck cleanup, builds passed for `default`, `gh_release`, `sticky`, and `x4pro`. The C3 framework rebuild and application link completed in one invocation. Switching from `sticky` to `x4pro` correctly restored prebuilt framework libraries. The Sticky SDK build emitted USB const-qualifier warnings.
- Flashed and booted the C3 platform-upgrade build on X3 before the cppcheck cleanup. It passed **200 consecutive screenshot captures** using the unchanged handler's single `logSerial.write(buf, bufferSize)` call. Every raw serial payload was exactly 52,272 bytes with its expected end marker. On the unchanged Home screen, all 200 received payloads had the same SHA-256 hash. Six serial heap samples before and after the batch were identical to the 3.3.11 column above.

- Flashed and booted cleanup commit `13def4c3` on the same X3. Six serial heap samples over 60 seconds matched all four Arduino 3.3.11 values above exactly. A Home-screen screenshot also transferred successfully with the expected 52,272-byte payload and end marker.

- I also performed basic manual smoke testing on my X3 with `13def4c3`; the functionality I tested appeared to work normally.

Hardware validation was limited to X3.

## Scope Check

- Stock/fork feature comparison: not applicable to this dependency update for CrossPoint's existing serial interface.
- No changes to `freeink-sdk/`, `lib/hal/`, bootloader, OTA, or recovery source. The SDK/toolchain upgrade warrants the usual device smoke testing before merge.

### AI Usage

**YES** — AI tools assisted with investigation, build-hook changes, cppcheck cleanup, and validation.

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -16,7 +16,19 @@ check_tool = cppcheck
 ; missingInclude (project headers) is suppressed alongside missingIncludeSystem: on a
 ; fresh CI checkout cppcheck has no resolved include paths, so it reports every
 ; project header as missing (~400 information-level lines) and fails the job.
-check_flags = --enable=all --suppress=missingIncludeSystem --suppress=missingInclude --suppress=unusedFunction --suppress=unmatchedSuppression --suppress=*:*/.pio/* --inline-suppr
+check_flags =
+  --enable=all
+  --check-level=exhaustive
+  --suppress=missingIncludeSystem
+  --suppress=missingInclude
+  --suppress=unusedFunction
+  --suppress=unmatchedSuppression
+  --suppress=*:*/.pio/*
+  ; The SDK is maintained separately; keep this check scoped to reader-owned code.
+  --suppress=*:*/freeink-sdk/*
+  ; Analyzer status output is not a source-code defect.
+  --suppress=checkersReport
+  --inline-suppr
 check_skip_packages = yes
 
 board_upload.flash_size = 16MB
--- a/platformio.ini
+++ b/platformio.ini
@@ -7,7 +7,7 @@ extra_configs = platformio.local.ini
 version = 1.6.0
 
 [base]
-platform = https://github.com/pioarduino/platform-espressif32/releases/download/55.03.37/platform-espressif32.zip
+platform = https://github.com/pioarduino/platform-espressif32/releases/download/55.03.311/platform-espressif32.zip
 board = esp32-c3-devkitm-1
 framework = arduino
 monitor_speed = 115200
@@ -83,6 +83,7 @@ extra_scripts =
   pre:scripts/git_branch.py
   pre:scripts/patch_jpegdec.py
   post:scripts/register_unit_tests_target.py
+  post:scripts/patch_arduino_rom_libc.py
 
 ; Libraries
 lib_deps =
@@ -146,6 +147,8 @@ custom_sdkconfig =
   ; ESP_I2S.h, which the isolated core rebuild can't resolve. Unused here anyway,
   ; so drop it or the sticky env fails to build.
   CONFIG_ARDUINO_SELECTIVE_ESP_SR=n
+  ; esp-modbus requires its API blocking limit to cover the response timeout.
+  CONFIG_FMB_MASTER_MAX_API_BLOCKING_TIME_MS=10000
 
 ; Drop unused cloud components from the core rebuild. esp_insights/rainmaker
 ; require embedded server certs the lib builder can't generate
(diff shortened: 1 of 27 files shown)


serialx:
@Uri-Tauber I finally got time to continue work on font optimizations and this bug came up. This took the rest of the day. 🤦 

Since this touches many parts of the code (because of cppcheck updates), can you take a look?

platformio.ini
```diff
@@ -16,7 +16,19 @@ check_tool = cppcheck
 ; missingInclude (project headers) is suppressed alongside missingIncludeSystem: on a
 ; fresh CI checkout cppcheck has no resolved include paths, so it reports every
 ; project header as missing (~400 information-level lines) and fails the job.
-check_flags = --enable=all --suppress=missingIncludeSystem --suppress=missingInclude --suppress=unusedFunction --suppress=unmatchedSuppression --suppress=*:*/.pio/* --inline-suppr
+check_flags =
+  --enable=all
+  --check-level=exhaustive
```
Uri-Tauber, line 21: This line is new.

serialx: The newer cppcheck emitted 68 branch-limit notices under its default analysis level. CI treats those informational notices as low-severity findings, so they caused failure.

I added `--check-level=exhaustive` to remove that analysis limit rather than suppress the notices. The tradeoff is longer analysis:

  | Before | Exhaustive
-- | -- | --
Linux CI | 39 s | 66 s

Uri-Tauber:
@serialx If you continue with font optimizations, notice I cherry-picked one of the commits from your #3083 in #3222.
Since #3144 gained us ~320kb, I'm not too troubled about space right now. Making text rendering faster would be a bigger win.

Uri-Tauber (approved):
lgtm

PR #3412, opened 2026-09-07
fix: Fixes KOSync memory checks and reduces memory pressure

### LINKED ISSUE #3301: Image placement instead of images after 2-3 images shown.
### Device

Xteink X4 Pro

### Affected Version

1.6.0 rc

### Describe the Bug

After 2 or 3 images in a first books I’ve opened, system does not show other images - only placeholder. Also stop showing images in other books.

### Steps to Reproduce

1. Go to book
2. Wait till image has to be displayed

### Expected Behavior

Display images

```shell

```

itsthisjustin:
Reader caches and the menu snapshot are released before position mapping, which now borrows the framebuffer for ZIP decompression. The TLS preflight check accounts for PSRAM available to the default allocator.

Image failures are suppressed only for the current page render, allowing transient failures to retry on the next render. ZIP errors now identify the failed allocation and EPUB entry, and memory logs include PSRAM usage.

Also may fix #3301


unnamedd (approved):
Looks good to me!

PR #3418, opened 2026-09-07
fix(reader): synchronize end-of-book menu selection

Daviex:
## Summary

The input task updates the end-of-book selection while the render task reads it for list painting. The existing `isLoaded` release/acquire pair publishes the initial list but does not synchronize later writes to the plain `int selector`.

Use an atomic selector with relaxed loads/stores. Input handling takes one snapshot after touch routing; rendering reads the selection atomically. The immutable list retains its existing publication mechanism. Only the two reader-menu files change.

## Scope Check

- [x] This is not a new built-in theme.
- [x] This is not a new external network connector.
- [x] This is not an interactive app, writing tool, RSS/news/browser, media playback, or PDF feature.
- [x] This improves the existing reader. It extracts a generic correction developed in my fork; upstream currently has the unsynchronized selector.
- HAL / SDK / bootloader / OTA / recovery coordination: not applicable.

## Validation and additional context

- Built `pio run -e default -e x4pro` from `develop` at `7db14a01` plus only this patch; both passed with the pinned SDK.
- Complete `bin/clang-format-fix` wrapper and `git diff --check` passed.
- Reviewed the single-writer flow, touch selection, Confirm and wrapping navigation. Relaxed ordering suffices because the atomic does not publish other data.
- Static RAM unchanged: 56,360 bytes (C3) / 100,040 bytes (X4 Pro). Application flash delta: +72 / +56 bytes respectively.
- No hardware stress test or ThreadSanitizer reproduction was run. Device verification: navigate/tap recommendations during repeated repaints, then verify Confirm, Home and wrapping navigation.

### AI Usage

Did you use AI tools to help write this code? **YES** — implementation, review and build verification were AI-assisted.

### DIFF
--- a/src/activities/reader/EndOfBookOptions.cpp
+++ b/src/activities/reader/EndOfBookOptions.cpp
@@ -36,7 +36,7 @@ void EndOfBookOptions::loadOnce(const std::string& currentBookPath) {
   }
   folder = FsHelpers::extractFolderPath(currentBookPath);
   names = NextBookFinder::findNextBooks(currentBookPath, MAX_SUGGESTIONS);
-  selector = 0;
+  selector.store(0, std::memory_order_relaxed);
   if (!names.empty()) {
     // One-time app setup on the render task, before the first render/route.
     resetUi();
@@ -84,7 +84,7 @@ std::string EndOfBookOptions::fullPath(const size_t index) const {
 void EndOfBookOptions::onRowEvent(const fui::ActionEvent& event, void* user) {
   auto* self = static_cast<EndOfBookOptions*>(user);
   if (event.value < 0 || event.value > static_cast<int16_t>(self->names.size())) return;
-  self->selector = event.value;
+  self->selector.store(event.value, std::memory_order_relaxed);
   // The tapped row leaves this screen (open book or home); a lingering flash
   // would gray an unrelated element on the next render.
   self->app.clearTapFlash();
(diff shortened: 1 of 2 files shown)


Uri-Tauber:
Hey @Daviex.
Do you have actual hardware you tested on?

Daviex:
Hello @Uri-Tauber,

all the PRs I created today were extracted from a customized version of CFW in which I had added some extra functionality.

Today I used GPT-6 Astra to help identify and extract the parts of that code that could also be useful for the base CFW, independently of my specific customization.

That said, these changes come from the version I actually use on my XTeink X4 Pro every day, and I haven’t encountered any issues with them so far.

This PR is also related to #3421. I had several books with very long filenames, and the UI was cutting off the final part of the name. In my case, that part was important because I needed it to distinguish between multiple EPUB files split from the same book.

If you need more info I'm here for it.

Best regards, Davide.

Uri-Tauber:
Can you point me to your custom firmware repo?

Daviex:
Actually it is private for now, sent your access as collaborator so you can access to it (I think you received an email for it), there are actually some releases but the latest version is not uploaded, I will upload it ASAP.

Releases has only the firmware for X4Pro, as it is the only device I have

Uri-Tauber:
got it. thx.

Uri-Tauber (approved):

PR #3422, opened 2026-09-07
perf(kosync): release the counting parser before the XPath pass

Daviex:
## Summary

`findXPathForProgress()` first counts chapter text, then parses it again to locate the requested position. The first `ParagraphTextCounter` currently remains in scope until the function returns, so its Expat allocations overlap the second parser.

Limit the counter to a nested scope and retain only `totalVisibleChars`. The second pass starts after the counter and its parser have been destroyed. Production changes are confined to that lifetime boundary in one function; no parsing algorithm or ownership API changes.

## Scope Check

- [x] I have read SCOPE.md and ROADMAP.md.
- [x] This is not a new built-in theme.
- [x] This is not a new external network connector.
- [x] This is not an interactive app, writing tool, RSS/news/browser, media playback, or PDF feature.
- [x] This improves an existing CrossPoint path and extracts generic work developed in my fork; no account, cloud integration or firmware-branding changes are included.

- HAL / SDK / bootloader / OTA / recovery coordination: not applicable.

## Validation and additional context

- 10 host tests compile the production resolver with vendored Expat and a simulated EPUB stream. All pass; the same suite against unmodified `7db14a01` fails three parser-lifetime checks while seven result/error-path checks pass.
- Real Expat create/free and allocation hooks measure peak live parser count **2 → 1**. The baseline keeps 16 allocations from the counting parser live when the resolver is created; the patch leaves **zero**. This is a lifecycle measurement, not a claimed byte saving.
- Golden anchors, empty chapters, malformed XML, stream errors and parser-allocation failure are covered.
- `pio run -e default -e x4pro` passes from the base plus only this patch. Complete formatting wrapper and whitespace checks pass.
- Static RAM unchanged; measured application flash delta +2 bytes (C3), +24 (X4 Pro). Transient heap overlap is removed without introducing an allocation.
- No hardware heap telemetry was captured. Device follow-up: compare KOSync progress output and minimum free heap during percentage-based chapter resolution.

### AI Usage

Did you use AI tools to help write this code? **YES** — implementation, tests, review and builds were AI-assisted.

PR #3424, opened 2026-09-07
fix(kosync): preserve precise visible-text anchors on sync

Daviex:
## Summary

The EPUB reader already supplies a visible-text offset, but outgoing KOSync progress currently chooses a paragraph or page approximation. A page beginning inside nested inline text can therefore lose its exact position when synchronized.

Prefer the visible offset and stream the chapter to resolve an actual text-node XPath, retaining paragraph/page fallbacks when no exact anchor is available. Preserve explicit zero offsets such as `text()[2].0`. Fix the matching inbound text-node count so leading/consecutive child elements do not invent empty nodes; exported anchors now round-trip through those layouts. Element-start anchors retain their existing meaning.

For example, in `<p><em>alpha</em> beta</p>`, offset 5 resolves to `p[1]/text()[1].0` and returns to offset 5.

## Scope Check

- [x] This is not a new built-in theme.
- [x] This is not a new external network connector.
- [x] This is not an interactive app, writing tool, RSS/news/browser, media playback, or PDF feature.
- [x] This improves an existing CrossPoint path and extracts generic work developed in my fork; no account, cloud integration or firmware-branding changes are included.

- HAL / SDK / bootloader / OTA / recovery coordination: not applicable.

## Validation and additional context

- 22 dedicated position tests and the complete host suite pass: **197/197**. The initial API-compatible regression set has 11 failures and three fallback passes on unmodified `7db14a01`. Two further regressions reproduced the inbound node-boundary problem before its correction.
- Coverage includes nested/leading/consecutive inline children, body text, hidden elements, images, Unicode, named/numeric entities, CRLF/whitespace, zero offsets, end boundaries, chunks of 1–17 bytes, malformed input and fallback. Eight markup layouts round-trip across every offset and all those chunk sizes.
- Tests compile the actual mapper, resolver and vendored Expat. EPUB/Section storage is simulated; a physical-device exchange with another KOSync client was not performed.
- Both `pio run -e default -e x4pro` builds pass from `7db14a01` plus only this patch. Complete formatting wrapper and whitespace checks pass; independent review covered the final node-count correction.
- Static RAM unchanged; application flash delta +2,820 bytes (C3) / +2,452 (X4 Pro). The new resolver is allocated once with a checked allocation, streams 1024-byte chunks and stops after finding the target. Parser/string/sibling bookkeeping still uses heap allocations; this is not a claim of complete OOM hardening.
- Exact resolution falls back if body comments, processing instructions or CDATA appear before the target, or ancestry exceeds the inbound mapper's 16-step limit. This avoids exporting anchors the existing inbound scanner cannot reliably read.
- The separate parser-lifetime optimization in #3422 is not included here; both branches start from the same upstream base.
- Suggested device verification: push from a page starting inside inline text, resume on another KOSync client, then synchronize back and confirm the same paragraph/visible offset.

### AI Usage

Did you use AI tools to help write this code? **YES** — implementation, tests, review and builds were AI-assisted.


Daviex:
@coderabbitai full review

WhoTheHeck:
Nice to see more KOSync improvements! I think #3424 overlaps with the main goal of my earlier #3174: CrossPoint already has an exact visible-text position, but the KOSync upload can fall back to an imprecise paragraph or page XPath.

#3424 adds useful broader tests and also fixes an inbound text-node edge case. #3174 is the smaller implementation and reuses the existing resolver to keep it small. A possible smaller combined path would therefore be to keep the #3174 approach, bring over the relevant text-node fix and regression tests from #3424, and avoid adding a second resolver.

#3424 fixes one incoming text-node case. #3226 is the broader incoming-parser fix: it also handles comments, CDATA, namespaces and malformed markup. Since it is a larger change, it may be better to split it into smaller parts if preferred.

My suggested order for the open KOSync improvements is:

1) #3111 independently, for Smart Sync direction selection.
2) Decide on a combined #3174/#3424 approach for precise outgoing positions.
3) Decide whether to take #3226 as the broader incoming-parser follow-up, or extract the smaller inbound text-node fix from #3424.
4) Add #3422 once the resolver approach is settled.
5) Handle #3425 and #3426 separately as document-ID changes, with #3425 landing first.

Based on maintainer advice, we can rebase, split, or combine the existing work accordingly.

Daviex:
Thanks @WhoTheHeck for laying out the overlap and the proposed order. I'm happy to coordinate and combine the useful parts of these contributions.

Using the existing resolver through the #3174 approach sounds worth pursuing, provided we preserve the exact-position behavior and regression coverage added here. I'm happy for the relevant direct-text-node fix and tests from #3424 to be carried over, and to adapt or split this PR once the maintainers indicate their preferred direction.

For incoming positions, it would be useful to have their guidance on whether to take the broader #3226 work or start with the smaller text-node correction. Keeping #3422 until the resolver approach is settled also makes sense. I agree that #3425 and #3426 should remain separate document-ID changes, with #3425 landing first.

Maintainers: would you prefer the outgoing-position work to be consolidated around #3174 with the relevant fixes/tests from this PR, or a different split? I'll wait for that guidance before restructuring these PRs.

Daviex:
@coderabbitai full review

The PR has been rebased onto the current upstream `develop` at `92ef5707`. Please review the complete current changeset, including the rebase compatibility updates.

Daviex:
@coderabbitai full review

The CI compatibility fix and test expectation updates are now pushed in a0f3cb9. Please review the complete current changeset.

Daviex:
The test-scope audit found a regression in the previous rebase adjustment (`ba0f3cb9`): seven existing upstream resolver tests were changed to accept an empty fallback where upstream already returns a precise text-node anchor for comments, processing instructions and CDATA. Those expectation changes were not justified as compatibility fixes.

Re-running the original upstream resolver tests against this PR produces eight failures out of 15. Seven are the lost behaviors above; the eighth is the deliberate extension to body/div text and is a different case. For example, `before<!--comment-->after` at visible offset 6 loses `/body/DocFragment[1]/body/p[1]/text()[2].0` and returns an empty string. The checked-in suite passes 37/37 only with the altered expectations.

This needs an implementation correction, not fewer regression tests. Since #3174 has landed, the smaller direction discussed above is to retain the upstream resolver and its original contracts, then keep the relevant inbound text-node correction and targeted tests from this PR. No resolver rewrite is included in this test-only audit. The current green CI should not be taken as evidence that these upstream behaviors are preserved.

Daviex:
Updated in `027f4e0a` on upstream `8c84ef32`. Following the maintainer's request, exact and percentage resolution now share the existing upstream `XPathProgressResolver`; the second resolver is removed. Direct body/div text, explicit zero/end offsets and inbound XML node boundaries remain covered. The seven original XML expectations are restored, including invisible-text exclusion, and comments/PI/CDATA now round-trip exactly instead of forcing the previous XML fallback.

All 38 focused host tests pass (22 position + 16 XPath); formatting/whitespace checks pass. Fixture logging/constant-data notes were also addressed. #3422's separate parser-lifetime optimization is not imported. The description now reflects the implementation on current upstream. No local firmware build or new device exchange was run.

@coderabbitai full review

Daviex:
Addressed the list-counting mismatch in `e0e40bd2`. `ParagraphTextCounter` now counts text when inside either `p` or `li`, matching the percentage resolver and counting paragraphs nested in list items only once.

Added one focused regression with plain list text, a nested paragraph and a final paragraph. It checks the 50% anchor and that 100% reaches the final paragraph. All 39 focused cases pass (17 XPath + 22 position); the full formatting wrapper is idempotent after two verified worktree passes. No local firmware build was run.

@coderabbitai full review

PR #3430, opened 2026-09-07
feat(keyboard): add an Arabic layout

rxmmah:
Appends the SDK's new ArabicAr layout to the picker table, so Arabic joins the set a user can enable and cycle through with the script key. It takes the next free bit rather than displacing anything, keeping saved masks readable, and the row labels itself from the existing Arabic UI language name.

The layout has no Latin letters, so it stays out of LATIN_BITS and the picker keeps refusing to switch off the last Latin one -- a Wi-Fi password still has to be typeable.

Bumps freeink-sdk for the layout tables.

### AI Usage

### DIFF
--- a/src/activities/util/KeyboardEntryActivity.h
+++ b/src/activities/util/KeyboardEntryActivity.h
@@ -46,7 +46,8 @@ class KeyboardEntryActivity : public Activity {
   // app-defined tables in the .cpp.
   freeink::ui::KeyboardLayoutId layoutId = freeink::ui::KeyboardLayoutId::QwertyEn;
   // Asks the SDK for a layout variant with the language key. Only the Latin
-  // layouts honour it; the Cyrillic and Hebrew tables carry the key either way.
+  // layouts honour it; the Cyrillic, Hebrew and Arabic tables carry the key
+  // either way.
   // Resolved once in onEnter(): the set cannot change while a keyboard is on
   // screen, and currentLayout() runs on every loop pass.
   bool showLangKey = false;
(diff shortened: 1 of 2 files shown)


rxmmah:
@itsthisjustin  @Uri-Tauber could you please take a look at this?

Uri-Tauber:
How well it perform on your hardware?
I had the impression Arabic is more complex since you need to shape letters in real time, so a new letter can change the previous glyphs.

rxmmah:
> How well it perform on your hardware? I had the impression Arabic is more complex since you need to shape letters in real time, so a new letter can change the previous glyphs.

you are right. i have tested it on the simulator. and everything worked.
image

Uri-Tauber:
Cool.

Why did you added new row here? wouldn't it make more sense to have it in the SDK?

rxmmah:
The Arabic key layout is already in the SDK. This row registers it in CrossPoint’s layout picker and associates it with the Arabic UI language, following the existing layouts. I appended it because the table order determines the saved settings bits, so existing selections stay intact. Did you mean moving the list of available layouts into the SDK too?

Uri-Tauber:
Not sure I understood you. I meant `fui::KeyboardKey ARABIC_PUNCTUATION_ROW`. What does it add on top of the existing Arabic layout?

rxmmah:
> Not sure I understood you. I meant `fui::KeyboardKey ARABIC_PUNCTUATION_ROW`. What does it add on top of the existing Arabic layout?

You’re right, the punctuation belongs in the SDK. I’ve removed the custom row locally so this PR uses the SDK’s existing symbols layout.

Uri-Tauber:
Can you fix the submodule pointer, and do a quick sanity check on real hardware?

rxmmah:
> Can you fix the submodule pointer, and do a quick sanity check on real hardware?

No  the SDK pointer to 48f5949, matching current develop.
i flashed it on my x4c and it's all right.

Uri-Tauber (approved):

PR #3439, opened 2026-09-08
fix: stabilize X3 EPUB anti-aliasing

### LINKED ISSUE #3400: Speckle artifacts / fading on text ( AA: On )
### Device

Xteink X3

### Affected Version

v1.6 , v1.5

### Describe the Bug

On X3 ( newer display )
When reading with anti-aliasing turn on, after 2-3 page turns, text became lighter with speckle artifacts (presumably from greys font edge on previous page? due to AA) 
Texts became black again after full page refresh. 

Image

Image 

### Steps to Reproduce

1. Turn on Anti-aliasing
2. go next page
3. go back to original page. 
4. keep repeating and see more artifacts, text became lighter overall.

### Expected Behavior

Maintain black text. 
Other firmware such as (crossink / yacp ) do not have this issue. 
If i look closely on screen, seems like other firmware render the text again after each page-turn, but without full refresh) ?

```shell

```

uxjulia:
## Summary

  * Ports the fix for X3 EPUB text anti-aliasing that can fade or produce speckles by using the controller’s required grayscale base and preconditioning refresh sequence from Crossink. Fixes #3400

  * Add a `supportsAsyncGrayscaleBase()` capability so the reader only overlaps grayscale composition with a normal async B/W refresh on displays where that is valid.
  * Route normal X3 anti-aliased pages through the controller-specific grayscale base refresh.
  * Run the scheduled X3 cleanup refresh before grayscale preconditioning, preserving a stable B/W controller baseline before grayscale planes are applied.
  * Keep existing async overlap behavior for compatible non-X3 displays and retain existing image-page handling.

## Additional Context

* This changes refresh sequencing only; it adds no persistent storage, cache-format, or long-lived heap allocations.
* X3 cannot safely use the generic async-refresh path as a grayscale base, while compatible displays retain that overlap path.
* Affected user has verified a build with this fix and reports the issue was resolved.

### DIFF
--- a/lib/GfxRenderer/GfxRenderer.cpp
+++ b/lib/GfxRenderer/GfxRenderer.cpp
@@ -1701,6 +1701,8 @@ void GfxRenderer::waitRefreshComplete() const { display.waitRefreshComplete(); }
 
 bool GfxRenderer::supportsAsyncRefresh() const { return !fadingFix && display.supportsAsyncRefresh(); }
 
+bool GfxRenderer::supportsAsyncGrayscaleBase() const { return !fadingFix && display.supportsAsyncGrayscaleBase(); }
+
 size_t GfxRenderer::readFramebufferRegion(int x, int y, int w, int h, uint8_t* dst, size_t dstCapacity) const {
   if (dst == nullptr || w <= 0 || h <= 0) return 0;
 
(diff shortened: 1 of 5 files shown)


itsthisjustin:
I don't mind merging this but shouldn't this be in the SDK at a driver level? Especially since x3 has multiple displays?

serialx:
I tested this and it's way better now!

uxjulia:
> I don't mind merging this but shouldn't this be in the SDK at a driver level? Especially since x3 has multiple displays?

Opened the relevant freeink-sdk PR (https://github.com/Free-Ink/freeink-sdk/pull/89). I'll update the code for this once it's merged.

itsthisjustin (approved):

itsthisjustin:
@uxjulia ready to merge?

uxjulia:
> @uxjulia ready to merge?

Yep!

PR #3465, opened 2026-09-09
fix: display grayscale BMPs in the viewer

### LINKED ISSUE #3464: BMP viewer skips grayscale passes, leaving gray pixels black (X3, v1.6.0)
### Device

Xteink X3

### Affected Version

1.6.0, confirmed by the device's `/api/status`. The same rendering omission is also present on `develop` at `2bad1d09fd7db7dc6f21738cdb9986c51591f73d`.

### Describe the Bug

Opening a grayscale BMP through **Browse Files** leaves its gray pixels black. Displaying the **same file** as the custom sleep screen initially shows the same dark image, then approximately 1–2 seconds later a further refresh makes the gray tones appear. The ordinary image viewer never performs that second stage.

This is particularly obvious in a checkerboard comparison containing alternating regions of a pre-dithered black/white photo and a grayscale version: the black/white regions display normally, while grayscale regions become nearly solid black in the viewer. Both regions render as intended once the sleep-screen refresh finishes.

The observed file is an uncompressed 528×792, 4-bit indexed BMP using only native grayscale palette entries. No scaling or high-color quantization is needed. Sleep Screen is **Custom** and Sleep Screen Cover Filter is **None**.

### Steps to Reproduce

1. Copy a native-palette grayscale BMP to the SD card. A minimal four-band test image can be generated with the dependency-free Python script below.
2. Open it in **Browse Files** and wait for rendering to finish. The gray bands remain black, rather than receiving their grayscale update.
3. Set the same BMP as the custom sleep screen, with the cover filter set to **None**.
4. Put the device to sleep and wait for the complete refresh. The gray tones appear after the initial black base image.

### Expected Behavior

The image viewer should complete grayscale rendering and show the same tones as the final sleep-screen image. Black/white images should retain their existing rendering behavior.

### Source Analysis

Links below are pinned to the 1.6.0 release commit, `54337e6d73fc628f4ba523ddc89a743ca8c6e4c5`.

* [`BmpViewerActivity::onEnter()`](https://github.com/crosspoint-reader/crosspoint-reader/blob/54337e6d73fc628f4ba523ddc89a743ca8c6e4c5/src/activities/util/BmpViewerActivity.cpp#L157-L166) calls `drawBitmap()`, draws button hints, and unconditionally finishes with `displayBuffer(FAST_REFRESH)`. There is no grayscale branch or subsequent gray-plane rendering.
* [`GfxRenderer::drawBitmap()`](https://github.com/crosspoint-reader/crosspoint-reader/blob/54337e6d73fc628f4ba523ddc89a743ca8c6e4c5/lib/GfxRenderer/GfxRenderer.cpp#L1434-L1442) deliberately paints **all decoded values below 3 black** in BW mode. That includes both intermediate gray levels. This is the base image that remains visible in the viewer.
* [`SleepActivity::renderBitmapSleepScreen()`](https://github.com/crosspoint-reader/crosspoint-reader/blob/54337e6d73fc628f4ba523ddc89a743ca8c6e4c5/src/activities/boot_sleep/SleepActivity.cpp#L625-L660) instead calls `displayGrayscaleBase(HALF_REFRESH)`, rewinds and renders the LSB and MSB planes, copies both gray buffers, calls `displayGrayBuffer()`, and restores BW mode. This matches the extra refresh observed on the hardware.

The viewer appears to need the equivalent grayscale-aware sequence, with its button-hint regions preserved and render mode restored afterward. The sleep implementation specifically documents why its grayscale base uses HALF refresh, so merely changing FAST to FULL would not add the missing gray planes.

Related: #2879 discusses a sleep-screen grayscale failure in an older release. In this report the sleep-screen path succeeds; the viewer stops before the grayscale passes.

### Minimal Test Image

This creates four vertical solid bands with palette values 0, 85, 170, and 255. It isolates the missing render passes from photographic content and dithering. The file generator and resulting BMP were checked locally; hardware observations above are from the photographic comparison.

```python
from pathlib import Path
import struct

width, height = 528, 792
levels = (0, 85, 170, 255)
palette = b''.join(bytes((g, g, g, 0)) for g in levels)
row = b''.join(bytes(((i << 4) | i,)) * (width // 8) for i in range(4))
stride = ((width * 4 + 31) // 32) * 4
pixels = (row + bytes(stride - len(row))) * height
offset = 14 + 40 + len(palette)
header = struct.pack('<2sIHHI', b'BM', offset + len(pixels), 0, 0, offset)
dib = struct.pack('<IiiHHIIiiII', 40, width, height, 1, 4, 0,
                  len(pixels), 2835, 2835, 4, 4)
Path('gray-levels.bmp').write_bytes(header + dib + palette + pixels)
```

No serial logs captured. The missing calls were confirmed by source inspection of 1.6.0 and current `develop`; no patched firmware has been flashed or tested.

bos:
## Summary

Opening a native grayscale BMP in the image viewer leaves its gray pixels black, even though the same file displays correctly as a sleep image. The viewer renders the black/white base but never submits the grayscale planes.

This follows the sleep-screen sequence for BMPs with more than one bit per pixel: display the grayscale base, rewind and render each gray plane, then perform the grayscale refresh. One-bit BMPs keep their existing fast refresh. Each theme clears its button-hint footprint from the gray planes so the image cannot tint the monochrome controls.

After the gray refresh, the viewer re-renders the BW framebuffer and restores the controller baseline for popups and subsequent navigation. Failed rewinds fall back to a BW refresh, with the renderer returned to BW mode.

Fixes #3464.

## Scope Check

- [x] This PR is not a new interactive app, writing tool, RSS/news/browser, media playback, or PDF feature.
- [ ] Stock firmware / popular-fork comparison: not surveyed. This fixes a bug within CrossPoint's existing viewer, using rendering behavior already present in its sleep screen.
- SDK / HAL / bootloader / OTA / recovery coordination: not applicable; none of these are changed.

## Validation

- `pio run -e default`: passed (the shared X3/X4 ESP32-C3 firmware).
- `pio check -e default` scoped to all four changed C++ files, failing on low/medium/high defects: passed.
- `./bin/clang-format-fix -g` with clang-format 21: passed.
- GitHub CI: all four firmware builds (`default`, `sticky`, `x4pro`, and `papermono`), unit tests, clang-format, and cppcheck passed.
- Hardware: installed CI firmware `1.6.0-dev-detached-d42869e` on an Xteink X3. I confirmed that this fixes the original grayscale BMP image-viewer bug.

Broader hardware coverage of BMP bit depths, scaling, orientations, themes, navigation, and error paths has not yet been verified.

## Additional Context

The viewer reuses the existing framebuffer and bitmap row-buffer path. It adds two gray renders and one BW reconstruction for multibit BMPs, without an additional full-screen snapshot allocation. Runtime heap usage has not been measured.

### AI Usage

YES

### DIFF
--- a/src/activities/util/BmpViewerActivity.cpp
+++ b/src/activities/util/BmpViewerActivity.cpp
@@ -153,15 +153,48 @@ void BmpViewerActivity::onEnter() {
       GUI.fillPopupProgress(renderer, popupRect, 50);
 
       renderer.clearScreen();
-      // Assuming drawBitmap defaults to 0,0 crop if omitted, or pass explicitly: drawBitmap(bitmap, x, y, pageWidth,
-      // pageHeight, 0, 0)
       renderer.drawBitmap(bitmap, x, y, pageWidth, pageHeight, 0, 0);
 
       // Draw UI hints on the base layer
       GUI.drawButtonHints(renderer, labels.btn1, labels.btn2, labels.btn3, labels.btn4);
-      // Single pass for non-grayscale images
+      if (bitmap.hasGreyscale()) {
+        renderer.displayGrayscaleBase(HalDisplay::HALF_REFRESH);
 
-      renderer.displayBuffer(HalDisplay::FAST_REFRESH);
+        bool planesReady = true;
+        for (const auto mode : {GfxRenderer::GRAYSCALE_LSB, GfxRenderer::GRAYSCALE_MSB}) {
+          if (bitmap.rewindToData() != BmpReaderError::Ok) {
+            LOG_ERR("BMP", "Failed to rewind bitmap for grayscale rendering");
+            planesReady = false;
+            break;
+          }
+          renderer.clearScreen(0x00);
+          renderer.setRenderMode(mode);
+          renderer.drawBitmap(bitmap, x, y, pageWidth, pageHeight, 0, 0);
+          GUI.drawButtonHints(renderer, labels.btn1, labels.btn2, labels.btn3, labels.btn4);
+          if (mode == GfxRenderer::GRAYSCALE_LSB) {
+            renderer.copyGrayscaleLsbBuffers();
+          } else {
+            renderer.copyGrayscaleMsbBuffers();
+          }
+        }
+        if (planesReady) renderer.displayGrayBuffer();
+
+        // Rebuild the BW framebuffer for popups and subsequent differential updates.
+        renderer.setRenderMode(GfxRenderer::BW);
+        renderer.clearScreen();
+        if (bitmap.rewindToData() == BmpReaderError::Ok) {
+          renderer.drawBitmap(bitmap, x, y, pageWidth, pageHeight, 0, 0);
+        } else {
+          LOG_ERR("BMP", "Failed to rewind bitmap to restore the BW framebuffer");
+          renderer.drawCenteredText(UI_10_FONT_ID, pageHeight / 2, tr(STR_FILE_OPEN_FAILED));
+          planesReady = false;
+        }
+        GUI.drawButtonHints(renderer, labels.btn1, labels.btn2, labels.btn3, labels.btn4);
+        renderer.cleanupGrayscaleWithFrameBuffer();
+        if (!planesReady) renderer.displayBuffer(HalDisplay::HALF_REFRESH);
+      } else {
+        renderer.displayBuffer(HalDisplay::FAST_REFRESH);
+      }
 
     } else {
       // Handle file parsing error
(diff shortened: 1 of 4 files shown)


Uri-Tauber:
Hey, @bos. 
Can you take before/after pictures with your phone?

Another thing, I think BW rendering was chosen deliberately to reduce rendering time. Can you measure how much this change add?

bos:
before-after

Here's a before/after comparison. As you can see the difference is quite noticeable.

Adding greyscale support to the viewer makes it take about 3 seconds to settle on my X3. Performance on truly monochrome images is unaffected.

PR #3478, opened 2026-09-09
chore: Consolidates grayscale capability checks and enables absolute grayscale for supported screens

itsthisjustin:
Consolidates grayscale capability checks and enables absolute grayscale for supported image screens, sleep overlays, and UC8279 image pages. SSD panels use the new image thresholds; other panels retain adjusted legacy tuning.

Updates FreeInk SDK with improved UC gray separation and the SSD refresh fix that separates painting from power- off.

Based on work by @theZiz and @zgredex

### DIFF
--- a/src/activities/boot_sleep/SleepActivity.cpp
+++ b/src/activities/boot_sleep/SleepActivity.cpp
@@ -555,7 +555,9 @@ void SleepActivity::renderCustomSleepScreen() const {
   // This takes priority over the /sleep folder.
   HalFile file;
   if (Storage.openFileForRead("SLP", "/sleep.bmp", file)) {
-    Bitmap bitmap(file, true);
+    Bitmap bitmap(file, true,
+                  renderer.grayscaleCapabilities(HalDisplay::GrayscaleMode::Absolute).supported() &&
+                      SETTINGS.sleepScreenCoverFilter == CrossPointSettings::SLEEP_SCREEN_COVER_FILTER::NO_FILTER);
     if (bitmap.parseHeaders() == BmpReaderError::Ok) {
       LOG_DBG("SLP", "Loading: /sleep.bmp");
       renderBitmapSleepScreen(bitmap);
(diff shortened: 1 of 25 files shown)


src/activities/boot_sleep/SleepActivity.cpp
```diff
@@ -626,14 +640,21 @@ void SleepActivity::renderBitmapSleepScreen(const Bitmap& bitmap, const bool pre
       bitmap.hasGreyscale() && (preserveBackground || SETTINGS.sleepScreenCoverFilter ==
                                                           CrossPointSettings::SLEEP_SCREEN_COVER_FILTER::NO_FILTER);
 
-  renderer.drawBitmap(bitmap, x, y, pageWidth, pageHeight, cropX, cropY);
+  if (!renderer.drawBitmap(bitmap, x, y, pageWidth, pageHeight, cropX, cropY)) {
```
theZiz, line 643: Nit: Would it make sense to move the `const bool absolute = ...` from line 636 higher and check for `!absolute` here so that we don't render the BW bases although we overwrite them later anyway?

Or do you want a fallback if the next render steps fail for some reason still having at least the BW cover?

itsthisjustin: Yep want a fallback for memory purposes or weird image issues

theZiz: Tbh, I just noticed in code and in the logging (because you can see the bitmap is read 3 times). The render time itself is fine for me, especially as we are speaking of turning the device of.

uxjulia (approved):
Tested on device. Covers have ever so slightly more contrast, doesn't look any worse. No perceived performance issues.

itsthisjustin:
> Tested on device. Covers have ever so slightly more contrast, doesn't look any worse. No perceived performance issues.

How rude! It's a pretty big difference on high quality images. See attached. 
(1) Discord  @Ziz - Vivaldi 2026-09-10 at 8 54 18 AM

uxjulia:
> > Tested on device. Covers have ever so slightly more contrast, doesn't look any worse. No perceived performance issues.
> 
> How rude! It's a pretty big difference on high quality images. See attached. (1) Discord @Ziz - Vivaldi 2026-09-10 at 8 54 18 AM

lol my images never really looked like the one on the left though

itsthisjustin:
> > > Tested on device. Covers have ever so slightly more contrast, doesn't look any worse. No perceived performance issues.
> > 
> > 
> > How rude! It's a pretty big difference on high quality images. See attached. (1) Discord @Ziz - Vivaldi 2026-09-10 at 8 54 18 AM
> 
> lol my images never really looked like the one on the left though

Yeah depends on the image. The 7 1/2th Deaths of Evelyn Hardcastle has a black cover with gray text and was almost entirely unrecognizable as a sleep cover before this change

bos:
Ships crossing in the night! I had fixed much of this in #3465. Thank you for sorting this out, it's a big improvement!

bos:
before-after

Importantly, this fixes a nasty waveform bug that caused gresycale images to display with noticeable vertical stripes.

PR #3506, opened 2026-09-12
feat: Increased keyboard height and spacing

itsthisjustin:
Much easier to tap on

I could use a quick test on the x3 @uxjulia


uxjulia (approved):
Tested on X3. Looks kind of silly on non-touch devices lol but works as expected. I'd prefer to gate this to touch devices, but I can understand keeping it simple by having the same design across the board.

itsthisjustin:
> Tested on X3. Looks kind of silly on non-touch devices lol but works as expected. I'd prefer to gate this to touch devices, but I can understand keeping it simple by having the same design across the board.

Easy enough to do though. Can just have cap touch equals a different height

PR #3510, opened 2026-09-12
feat: add serial device control for hardware testing

serialx:
## Summary

⚠️ NOTE: This is functioning right now, but I think it still need more testing. I plan to do more testing and perform manual code review myself, then make it ready for review. Maintainers/contributors are welcome to test it as-is and give me feedback.

Make UI tests repeatable on real hardware while collecting the project’s standard serial logs.

The debugging monitor gains a localhost server that shares its serial connection with `python3 scripts/device_control.py`. It works headless or alongside the existing monitor graph.

The CLI supports:

- Logical button presses with activity and page-change expectations.
- UI readiness and reader-position queries.
- Serial screenshots, rotated and inverted in Python using framebuffer metadata.
- Command files, exclusive control leases, log streaming, and serial release/reconnect for flashing.

Firmware replies identify the boot, request, and input-completion state. Batches validate argument syntax before execution, stop on failure, and never automatically replay uncertain input.

Example, with a book open:

```sh
# Terminal 1
python3 scripts/debugging_monitor.py --serve --headless

# Terminal 2
python3 scripts/device_control.py state
python3 scripts/device_control.py press page-forward --repeat 3 --expect-page-change
python3 scripts/device_control.py screenshot page.pbm
```

## Scope Check

- [x] I have read SCOPE.md and ROADMAP.md.
- [x] This PR is **not** a new built-in theme (themes are temporarily closed pending the move to SD-loaded themes).
- [x] This PR is **not** a new external network connector (sync engine, cloud storage, remote file access, etc.).
- [x] This PR is **not** an interactive app, writing tool, RSS/news/browser, media playback, or PDF feature.
- [ ] The stock firmware does not already handle this well, **and** no other popular CrossPoint fork already does (or, if one does, I explain why CrossPoint still needs it below).
- [ ] If this PR touches `freeink-sdk/`, `lib/hal/`, the bootloader, OTA, or recovery code, I have coordinated with the relevant maintainer.

The stock/fork comparison and HAL maintainer coordination remain outstanding.

## Validation

- All 23 Python tests and the native input-state tests passed.
- Base and candidate builds passed for `default`, `gh_release`, and `slim`.
- On an unlocked X3, verified readiness and capture across 40 activity targets, with selected navigation paths.
- Final device smoke tests covered repeated page turns, all four orientations, screenshot formats, batch failures, lease contention, Ctrl-C cleanup, and serial reconnection.

[...]

PR #3521, opened 2026-09-13
fix: reduce font-cache heap fragmentation

serialx:
## Summary

This is the second PR in my heap-fragmentation investigation, following #3518. It changes when SD-card font caches keep and release memory, reducing failed bitmap allocations and long page renders observed on an X3.

### Why allocations fail when memory is still free

A glyph is the drawn shape of a character. Before drawing a page, the reader copies the needed glyph images, or **bitmaps**, from the SD card into a RAM cache. Loading them ahead of drawing is called **prewarming**. The reader also keeps smaller arrays describing the glyphs and their locations, plus tables of character widths used to lay out text.

The bitmap buffer needs one continuous block of memory. Total free heap counts all available memory, including separate gaps between allocations. For example, two separate 12 KB gaps provide 24 KB in total, but cannot satisfy one 20 KB request. This is heap fragmentation. Freeing an allocation between two gaps can let the allocator combine them into a larger block.

Related: [#2884](https://github.com/crosspoint-reader/crosspoint-reader/pull/2884) by @itsthisjustin addresses the same large-bitmap allocation problem by dividing the bitmap cache into 4 KB chunks. This PR changes which glyphs remain cached and when temporary or rebuildable data is released, while retaining the current single-buffer storage. These changes may complement chunked storage; the measurements below compare this PR against develop and do not evaluate #2884.

The current font-cache behavior makes that problem more likely: complete page requests can retain earlier pages' glyphs, increasing the bitmap size, while smaller arrays remain allocated during bitmap replacement. It can also release a newly prefetched page when checking whether its buffer is oversized, making the next draw load it again.

### What changes

1. **Keep the glyph set appropriate to the caller.** A complete page scan already identifies everything that page needs. When its cache must be rebuilt, it now requests that set without adding earlier pages' glyphs. Existing buffers are reused when they fit, and a page whose glyphs are already loaded still gets a cache hit. Individual UI labels continue to accumulate glyphs, so drawing one label does not evict another label's characters.
2. **Use a prefetched page before considering it for trimming.** The cache still tracks repeated use of only a small part of its bitmap buffer. It now checks whether to release that oversized buffer when a later request needs different glyphs, after checking for a cache hit. Closing the prewarm operation can no longer trigger that trim before the page is drawn.
3. **Make room for the large bitmap before rebuilding smaller arrays.** Temporary character-to-glyph mappings are released after reading glyph descriptions. If the bitmap must grow, the temporary SD read-order array and old bitmap are also released first. If the largest free block is still too small, rebuildable glyph data is released and the bitmap is allocated before reconstructing that data. This gives adjacent free regions a chance to join together.
4. **Reclaim cached widths as a final recovery step.** If that bitmap allocation fails, the cache releases its character-width tables and retries once. Layout can rebuild those tables later. If recovery still fails, the existing fallback that attempts to cache fewer glyphs remains available.

The two existing temporary arrays now use checked `unique_ptr` ownership, so early returns free them automatically. Their sizes depend on the glyph count and can exceed the embedded stack budget. No permanent buffers or fields are added. Recovery can require an extra metadata read and later width-table rebuilding; those costs remain part of the tested behavior.

## Scope Check

- [ ] The stock firmware does not already handle this well, **and** no other popular CrossPoint fork already does (or, if one does, I explain why CrossPoint still needs it below).
- [ ] If this PR touches `freeink-sdk/`, `lib/hal/`, the bootloader, OTA, or recovery code, I have coordinated with the relevant maintainer.

This is a reading-performance and memory fix aligned with Phase 1. Stock/fork comparison has not been established. SDK/HAL/bootloader/OTA/recovery coordination is not applicable.

### Independent X3 comparison

Both `default` builds (LOG_LEVEL=2) use develop `c33a8b0e`, including #3518, plus identical #3510 automation support. Only `SdCardFont.cpp/.h` and `FontCacheManager.cpp/.h` differ between test builds. #3510 is excluded from this PR. The separate change to size the temporary font-metrics buffer to its input, and the other local parser changes, are also excluded.

The tests use the same X3 and SD card, Korean Demian, and Pretendard 14 pt: portrait, justified text, normal line spacing, extra paragraph spacing, margin 5, antialiasing and embedded styles enabled.

The reader has two relevant caches: the **chapter layout on SD**, which stores where text goes on each page, and the **font bitmap cache in RAM**, which this PR changes. We tested both of these situations:

- **Building the chapter layout:** clear the book cache independently for each build, reboot into chapter index 4/page 3, then perform three rounds of 20 forward and 20 backward turns. These 120 turns exercise incremental chapter building and subsequent revisits; neither run recorded idle prewarming.
- **Reading an existing chapter layout:** preserve the same completed 36-page layout across firmware changes, reboot each build into page 3, then perform 20 forward and 20 backward turns. This holds page content fixed and allows the reader to prewarm the next page while idle.

Each input follows at least one second of idle time and a readiness check. Render timings include display updates, but exclude waiting for an ongoing idle prewarm to finish. Idle prewarming is reported separately. Periodic full display refreshes, which clean up e-ink ghosting, are excluded from the ordinary-turn timings below; their positions matched across builds.

### Results

Times are **median / 95th percentile**, in milliseconds. The percentile uses nearest rank; with 19 observations, it is the largest observation. **Change = patch − baseline**, with timing changes also shown as median / 95th percentile. Lower times and failure counts are better; higher free-memory values mean more memory is available.

| Ordinary rendering or prewarming | Baseline | Patch | Change |
|---|---:|---:|---:|
| Next page while layout builds (56 turns) | 1,746 / 2,395 | 1,722 / 2,298 | −24 / −97 |
| Slowest next page while layout builds | 13,193 | 2,498 | −10,695 |
| Next page with existing layout (19 turns) | 1,339 / 1,378 | 1,344 / 1,386 | +5 / +8 |
| Previous page with existing layout (19 turns) | 1,908 / 1,980 | 1,720 / 1,810 | −188 / −170 |
| Idle prewarm with existing layout (39 observations) | 449 / 4,271 | 323 / 3,158 | −126 / −1,113 |

Memory figures come from existing ten-second logs. **Largest free block** means the largest single allocation the heap can currently satisfy; **total free heap** also includes smaller, separated regions. The minimum sampled value can miss shorter dips between samples.

| Memory and allocation results | Baseline | Patch | Change |
|---|---:|---:|---:|
| Failed bitmap attempts while layout builds | 19 | 1, recovered | −18 attempts |
| Failed bitmap attempts with existing layout | 14 | 0 | −14 attempts |
| Minimum sampled largest block while layout builds | 7,668 B | 21,492 B | +13,824 B |
| Minimum sampled largest block with existing layout | 27,636 B | 36,852 B | +9,216 B |
| Median total free heap while layout builds | 42,904 B | 47,246 B | +4,342 B |
| Median total free heap with existing layout | 61,248 B | 76,966 B | +15,718 B |

The remaining failed attempt recovered after releasing width tables: the largest free block grew from 22,516 to 36,852 B, allowing the full bitmap request to succeed. It did not need the fallback that retries with fewer glyphs, and produced no bitmap error log. It is still counted as a failed allocation attempt above.

Both builds completed 160 Demian turns without a crash or missing render timing. Captured start/end reading pixels match across builds in both workflows. This checks those captures, excluding the status footer; it does not verify every page or physical grayscale output.

The clearest measured benefits are fewer allocation failures, avoiding the observed 13-second render, and faster backward rendering. Typical next-page rendering with an existing layout is essentially unchanged (median +5 ms). Improvement depends on when memory is measured: while the layout was building, median largest block was slightly lower (26,100 → 25,588 B), as was the lowest free heap recorded since boot (14,476 → 13,936 B, including setup). Two patched idle prewarms still took about 3.2 seconds. These results are one sequential comparison on one device, not fixed memory savings or a general speedup percentage. Static DRAM is unchanged; the test firmware grows 768 B.

An earlier control used separately built chapter layouts, but baseline page 3 began one line later and its layout build logged width-cache allocation failures. That control is excluded from the existing-layout results above, which use the same preserved layout for both builds.

### Opening-note failure reproduction

In a separate fresh-cache trial, Korean *The Selfish Gene* aborted on baseline after four temporary font-metrics buffer allocation errors when entering its opening note (옮긴이의 말, chapter index 5). With this patch, all five pages completed with zero allocation errors; the chapter built in 1,064 ms. Baseline never completed that parse, so there is no valid parse-time speedup ratio. The temporary metrics-buffer size is unchanged in this PR.

### Validation

- All 244 host tests pass. Twelve focused font-cache tests also pass with AddressSanitizer and UndefinedBehaviorSanitizer enabled.
- Regression tests check that complete pages replace old glyph sets, UI labels still accumulate, prefetched data survives until drawing, and bitmap bytes remain correct after buffer growth or metadata rebuilding. A forced bitmap allocation failure verifies recovery through width-table release and subsequent width-table rebuilding.
- Default firmware builds with and without #3510 pass.
- Targeted cppcheck reports no high/medium findings and eight style findings.
- Formatting and diff checks pass.

---

### AI Usage


itsthisjustin (approved):

itsthisjustin:
I should close my pr after this closes right?

serialx:
> I should close my pr after this closes right?

I think your PR and this is complementary to each other. This PR do not touch the allocation sizes of the buffer.

PR #3528, opened 2026-09-13
feat: add word and character spacing controls

serialx:
## Summary

While looking through stale PRs, I came across #645 by @jonasdiemer—the oldest unmerged PR I found. Word spacing was a feature I already wanted to add, so I used that work as the starting point for a fresh PR adapted to the current codebase. I also added character spacing as a separate control.

The motivation for character spacing comes from Korean reading preferences reflected in community projects such as [ridibatang-9](https://github.com/mcnorton/ridibatang-9), which offers fonts with tighter Hangul spacing while leaving word spaces unchanged. I wanted to make spacing adjustable directly in the reader, without requiring a modified font.

Both controls are available under **Text Settings → Layout** and can be used with any language.

- **Word spacing:** 50–200% in 25% steps; default 100%.
- **Character spacing:** −2 to +2 px in 1 px steps; default 0 px. Applies between adjacent glyphs; whitespace keeps its own advance.
- Both settings update the preview, persist across restarts, and invalidate cached page layouts when changed.

### Implementation

`GfxRenderer` holds the two values as scoped pen state (`TextSpacingScope`) that its space, kerning, advance and draw calls honour. `ParsedText::layoutAndExtractLines` opens the scope during layout and `TextBlock::render` opens it while drawing, so no renderer, `Page` or `Section` signature changed and UI text outside a scope keeps the font's native spacing. Both values are stored per `TextBlock` in the section cache and validated in the section header (cache version **47**). `sizeof(BlockStyle)` is unchanged; the fields occupy existing padding. Firmware diff: 19 files, +192 / −36 lines against `develop`.

## Scope Check

- [ ] The stock firmware does not already handle this well, and no other popular CrossPoint fork already does (or, if one does, I explain why CrossPoint still needs it below).

This change concerns core typography and legibility. It does not modify the SDK, HAL, bootloader, OTA, or recovery code. The stock/fork comparison above still needs confirmation.

### Screenshots

All captures are from an Xteink X3 in portrait. All other settings are unchanged within each comparison pair. Click an image to inspect the full-resolution capture.

**Text Settings options**

| Word spacing: 50–200% | Character spacing: −2 to +2 px |
|:---:|:---:|
| Word spacing options, with 100% selected | Character spacing options, with 0 px selected |

**English — Aesop’s Fables**

**Character spacing:** word spacing stays at **100%** in both images.

| Before: **0 px** (default) | After: **−2 px** |
|:---:|:---:|
| Aesop’s Fables: 0 px character spacing, 100% word spacing | Aesop’s Fables: minus 2 px character spacing, 100% word spacing |

**Word spacing:** character spacing stays at **0 px** in both images.

| Before: **100%** (default) | After: **50%** |
|:---:|:---:|
| Aesop’s Fables: 100% word spacing, 0 px character spacing | Aesop’s Fables: 50% word spacing, 0 px character spacing |

<details>
<summary>Korean comparisons — 데미안 (click to expand)</summary>

The opening of 데미안 in Pretendard 16 pt, with left alignment to keep justification from redistributing the spacing difference.

**Character spacing:** word spacing stays at **100%** in both images.

| Before: **0 px** (default) | After: **−2 px** |
|:---:|:---:|
| 데미안: 0 px character spacing, 100% word spacing, left aligned | 데미안: minus 2 px character spacing, 100% word spacing, left aligned |

**Word spacing:** character spacing stays at **0 px** in both images.

| Before: **100%** (default) | After: **50%** |
|:---:|:---:|
| 데미안: 100% word spacing, 0 px character spacing, left aligned | 데미안: 50% word spacing, 0 px character spacing, left aligned |

</details>

### Validation

- Default and x4pro firmware builds pass for the branch; all **335 host tests** pass.
- Host tests cover Korean syllable tracking with scaled word spaces, the wrap threshold under word spacing, the cache round trip of the two fields, and a parser-level test that drives `startElement`/`characterData` with −1 px / 150% and checks the resulting word positions.
- Xteink X3 device checks: settings persist across reboots, both popups work, the preview follows both settings, and changing either setting invalidates the section cache. Framebuffer diffs of the same page show that −2 px and +2 px change the book layout, while 0 px / 100% renders byte-identical to `develop`.

### Performance and memory

**Whole-feature comparison against the merge base:** `develop` `32f2b2aa` → PR `e85ef86e` (this branch's single commit). Both builds carry the same serial-control test commit so the device can be driven over USB; both use the same SDK and toolchain. The settings-list allocation fix that was previously part of this branch is now its own PR and is not included here.

Xteink X3, 데미안, Pretendard 16 pt, portrait, justified, Extra Wide line spacing, anti-aliased text, default spacing (word 100%, character 0 px). Builds were flashed alternately, three boots each. Every boot deleted the book cache from the reader menu, reopened cold at the same position, did six warmup turns and six measured turns over the identical page sequence (spine 3 pages 2–6, spine 4 pages 1–7), then restored the start position: **18 measured turns per build**.

| Measurement | `develop` `32f2b2aa` | PR `e85ef86e` | Difference |
|---|---:|---:|---:|
| Median page drawing (`bw_render`) | 75.0 ms | 75.0 ms | 0.0 ms |
| Median total rendering, including display refresh | 1,419.5 ms | 1,419.0 ms | −0.5 ms |
| Median cold layout preparation (cache miss to first page render) | 4,504 ms | 4,485 ms | −19 ms |
| Median free heap after measured pages | 49,748 B | 49,340 B | −408 B |
| Largest free block after measured pages | 17,396 B | 17,396 B | 0 B |
| Lowest heap watermark (`Min Free`) | 28,060 B | 27,592 B | −468 B |
| Static RAM | 56,664 B | 56,664 B | 0 B |
| Flash | 5,554,531 B | 5,555,937 B | +1,406 B |

Timings are within run-to-run noise. The feature costs about 0.4–0.5 KB of heap at steady state; 264 B of that is the two new entries in the static settings list (`sizeof(SettingInfo)` is 132 B on the C3), the rest is the two settings keys in the JSON document. Spacing adds two bytes per TextBlock and two per section header to the SD cache.

These are firmware timings from the existing `Page render` and `MEM` logs (heap sampled every 10 s), not button-to-visible latency, and remain short, single-book measurements.

### AI Usage

Did you use AI tools to help write this code? **YES** — implementation, tests, device-test automation, and PR drafting used AI assistance.


itsthisjustin:
What's the drawing speed compared to develop instead of that other PR though? does it increase?

serialx:
> What's the drawing speed compared to develop instead of that other PR though? does it increase?

Oops! My agent updated the PR with the wrong metrics after code review fixes. The metrics are updated and show comparison with develop!

itsthisjustin (approved):

itsthisjustin:
@Uri-Tauber I'm good with this if you are.

serialx:
I’ve pushed an optimiezed update: addressing layout overhead and heap usage.

- Calculate boundary spacing once, reusing the existing line-break buffer.
- Fit spacing fields into existing padding, keeping `TextBlock` and `BlockStyle` the same size as baseline.
- Reserve the settings list’s final capacity upfront. On X3, it now allocates **59 slots instead of growing to 116**. That growth reallocated the vector and retained unused capacity, changing subsequent allocation placement and contributing to the heap-fragmentation regression. Avoiding it improves the minimum sampled largest free block in this workload.

I repeated the comparison against the same develop common ancestor, `2b2ba72f`. These fresh measurements supersede the earlier table.

The workload remains 데미안 on Xteink X3, Pretendard 16 pt, portrait, justified, Extra Wide, **100% word spacing / 0 px character spacing**. Three fresh boots per build, identical startup files and page sequence, six warmup turns followed by six measured turns: **18 measured turns per build**.

| Measurement | Baseline `2b2ba72f` | PR `458e9d42` | Difference (PR − baseline) |
|---|---:|---:|---:|
| Median page drawing | 353.0 ms | 347.5 ms | −5.5 ms |
| Median total rendering, including display refresh | 1,598.5 ms | 1,591.5 ms | −7.0 ms |
| Median initial layout preparation | 1,590.0 ms | 1,450.8 ms | −139.2 ms |
| Median free heap after measured pages | 59,020 B | 48,868 B\* | −10,152 B |
| Largest free block after measured pages | 42,996 B | 22,516 B\* | −20,480 B |
| Lowest sampled free heap during measured turns | 29,448 B | 36,324 B | +6,876 B |
| Smallest sampled largest free block | 20,468 B | 22,516 B | +2,048 B |
| Static RAM, without measurement probes | 56,656 B | 56,656 B | 0 B |
| Flash, without measurement probes | 5,494,431 B | 5,500,087 B | +5,656 B |

\* **After-page heap:** The settings allocation fix also leaves more total free heap during rendering. On the traced page, baseline reaches font-cache cleanup with 37,388 B free, below the existing 40 KiB release threshold; the updated build has 44,280 B free and retains the font buffers. This retention decision uses total free heap, so the larger contiguous block is a separate improvement. **Minimum rendering headroom improves, but retaining font buffers reduces after-page headroom.** Allocation traces confirmed this behavior.

Initial preparation is now 139.2 ms below baseline, and page-rendering medians show no slowdown. Total rendering includes font prewarming, display refresh and cleanup; these are firmware timings, not button-to-visible latency. Static RAM is unchanged; this preallocation fix contributes 780 B of the total flash increase. Cache version remains **47**.

Builds passed, all 15 candidate page-body captures matched baseline, and no crashes or rendering allocation failures occurred. These remain short, single-book measurements at default spacing; footer pixels were excluded.

Uri-Tauber:
Wow. Is it supposed to be so big? #645 was only 50 LoC...
I see big part of the LoC are host tests, but the firmware side looks too complicated, touches too many places.

serialx:
> Wow. Is it supposed to be so big? #645 was only 50 LoC... I see big part of the LoC are host tests, but the firmware side looks too complicated, touches too many places.

I totally understand. The character spacing needed many plumbing. Word spacing settings have too much rows and needed special treatment in horizontal UI.

Heap optimizations made additional changes too.

I'll see what I can do.

itsthisjustin:
@serialx sorry about the missing commit in the SDK. Also merged a PR from Julia that will just add some settings you need to solve a conflict with.

serialx:
Follow-up to my note above: rewrote the branch and force-pushed `7ca5fa34` (rebased on `develop` `32f2b2aa`). Same two controls, much less plumbing.

| Scope | Before (`458e9d42`) | Now (`7ca5fa34`) |
|---|---:|---:|
| Firmware (lib + src) | 25 files, +433 / −164 | 19 files, +208 / −58 |
| Host tests | 12 files, +672 / −8 | 2 files, +143 / −4 |
| Total | 38 files, +1117 / −173 | 22 files, +364 / −63 |

- The renderer now holds the two values as scoped pen state (`TextSpacingScope`) that its measure/draw calls honour, so no `Page`/`Section`/renderer signatures change. The two bytes are stored per `TextBlock` in the cache and validated in the section header.
- Word spacing is a numeric setting (50–200%), which drops the label strings and fits in landscape, so the OptionPopup scrolling change is out; I'll open it as a separate PR since it also helps the 8-option margin popup.
- Dropped the line-break compaction, ruby rework and ligature handling; the settings-list allocation fix is now a small separate commit.

Tests are down to four, including one that drives the real parser. Re-verified on the X3 with framebuffer diffs (−2 px / +2 px change the layout, defaults are byte-identical to `develop`) and re-measured; the description has the updated table. CI is green.

Also I've updated the device test measurements in the original PR body. There is basically no render perf regression and there is more free heap because I've fixed one heap fragmentation bug during this development.

Uri-Tauber:
Can you split the SettingsList.h optimization from 7ca5fa34e into a separate PR?

serialx:
Done: the SettingsList.h change is now #3569, and this PR is back to the single feature commit. The description's measurement table is updated to the feature-only build against `develop`; the free-heap gain moved to #3569 with it.

itsthisjustin:
@Uri-Tauber massive clean up here.  You good with it?

Uri-Tauber:
I still want to review the code. 
Since its not super urgent bugfix, I think it can wait two-three days more. 
@serialx Are you okay with that?

serialx:
Absolutely! Take your time

Uri-Tauber:
@serialx, Sorry for the delay. I review the code and I have two architectural problems I want you to fix before merging (my original review got quite messy, so this is a Codex rewrite.)

## 1. Avoid mutable reader-specific typography state in `GfxRenderer`

The main architectural concern here is that this PR makes `GfxRenderer` carry mutable reader-specific typography state.

`GfxRenderer` is otherwise a shared graphics service used across the UI: menus, settings, status bars, dialogs, and book rendering. Character spacing is specific to reader text layout/rendering, so making it ambient renderer state creates behavior that depends on which scope happens to be active rather than on the arguments to the rendering call itself.

The RAII guard in `GfxRenderer.h:315-335` avoids changing the existing page/render signatures, which I understand is attractive. However, it also causes `getTextWidth()` to effectively have two different implementations depending on `characterSpacing_`:

* normal path: `GfxRenderer.cpp:627`
* tracking path: `GfxRenderer.cpp:2045`

Those paths are not equivalent. In particular, the tracking-aware path halves SUP/SUB advances while the normal path does not. That difference then leaks into `TextBlock.cpp`, where we need the special-case check in `TextBlock.cpp:264-267` to keep decorations from getting the wrong width.

That is a warning sign: `TextBlock` is compensating for hidden renderer state rather than simply asking for the width it actually needs.

A smaller immediate fix would be to use the tracking-aware measurement function directly for decoration width in `TextBlock.cpp:262` and `TextBlock.cpp:274`. Decorations are measuring words that were already rendered with tracking, so that function already represents the width we want, including SUP/SUB handling. That would remove both the `blockStyle.characterSpacing == 0` workaround and the split behavior inside the ordinary `getTextWidth()` path.

Longer term, I think the cleaner shape is to keep `GfxRenderer` stateless with respect to reader typography and pass tracking explicitly to the relevant text measurement/rendering calls, e.g. an optional `int8_t tracking = 0`. `TextBlock` already owns the `BlockStyle`, so it has the information at the correct layer.

That would let us remove:

* the mutable typography fields in `GfxRenderer`
* the RAII typography scope
* the duplicated test stub
* the conditional redirect inside the normal text-width function
* the compensating SUP/SUB special case in `TextBlock`

The main benefit is not just cleanliness: future callers measuring superscript/subscript text should not get different geometry depending on whether some unrelated typography scope happens to be active.

---

## 2. `wordSpacingPercent` should not be persisted into `BlockStyle` or renderer state

I think `wordSpacingPercent` currently has a much larger lifecycle than it needs.

Word spacing matters during paragraph layout, when `ParsedText` decides where words go and where lines wrap. Once a `TextBlock` has been produced, every word already has its final X position. Page rendering does not draw space glyphs between those words, so `wordSpacingPercent` is no longer needed.

Because it was grouped together with character spacing, the PR currently:

* stores it in `BlockStyle`
* serializes/deserializes it for cached blocks
* restores it into `GfxRenderer`
* carries it through page rendering

but nothing in the page-rendering path actually reads it.

So this is effectively dead state and dead cache I/O, and it also gives future readers of `TextBlock.cpp` the false impression that word spacing still affects rendering there.

I suggest keeping `wordSpacingPercent` only where it has semantic value:

* `ReaderRenderSpec`
* the section/cache header, so changing the setting still invalidates cached layout
* `ParsedText`, where the space advance is scaled during line breaking/layout

Then remove it from:

* `BlockStyle`
* `TextBlock` serialization/deserialization
* the renderer typography state
* the page-rendering setup

This also keeps a useful distinction between the two settings:

* **character spacing** affects actual glyph positions and therefore both layout and drawing
* **word spacing** only affects layout before the words receive their final coordinates

---

## Also noted

`TextSettingsActivity.cpp:705-716` builds a `std::vector<std::string>` and performs heap-backed string formatting every time the word-spacing popup opens.

That is inconsistent with the rest of this UI code, where we generally try to avoid unnecessary heap churn on interaction paths. This could probably use a fixed static option array, similar to the existing pattern around `TextSettingsActivity.cpp:48`, or a small integer formatter without dynamic allocation.

serialx:
Thanks for the review, Uri. I’m still learning about typesetting and rendering, so your feedback helps a lot. I’ve addressed your concerns with Codex’s help:

- Removed the mutable typography state and RAII scope from `GfxRenderer`. Tracking is passed explicitly through measurement and drawing calls. The measurement helpers remain free functions in their original locations.
- Kept word spacing in paragraph layout and section-cache validation, removing it from `BlockStyle`, per-block serialization, and rendering setup.
- Replaced the word-spacing popup’s temporary vector and formatting loop with fixed options following the existing UI pattern.

`getTextWidth()` retains its bounding-box behavior. Decorations now use advance measurement directly, removing the conditional SUP/SUB workaround.

One additional implementation choice: decoration measurement uses an explicit `Rendered` mode and the same base direction as drawing. This bypasses the SD-font layout shortcut so kerning and ligatures are included. Switching decorations from bounding-box widths to advances can slightly change underline lengths even at zero tracking.

The final cache format remains v47. The firmware build and all 13 existing parser tests pass.

Before the final helper cleanup, I also ran paired X3 comparisons against the merge base using 데미안. Default-spacing page content matched, and spacing controls, persistence, and all four orientations passed. Render timings were comparable.

Uri-Tauber:
Wow, that was quick. 
I'm very busy right now. I should have the time to do one final pass tomorrow night.

Uri-Tauber (approved):

Uri-Tauber:
Awesome work

PR #3534, opened 2026-09-13
fix: dropped presses while a list repaints

Techneaux:
## Summary

* **What is the goal of this PR?** Fix a 1.6.0 regression: in list screens a button press only counts once the previous cursor move is on screen, so paging through the book browser, Recently Read, the in-reader menu or a chapter list means waiting out every e-ink refresh. [Reported](https://www.reddit.com/r/CrossPointReader/s/BhBk7nFqL6) by an X4 Original user; reproduced on X3. No existing issue found.
  * `UiListActivity::moveSelectionTo()` sets `ListNav::followOnBuild` instead of taking a blocking `RenderLock` and calling `follow()` itself.
  * `UiTabListActivity::moveRingTo()` moves to the same rail, dropping its duplicate of `syncTabListViewport`'s ring maths and its unguarded main-task write to `top`.
  * Two comment updates.

### Root cause

From the FreeInkUI list migration (#2957). `moveSelectionTo()` takes a blocking `RenderLock` on the main task, `renderTaskLoop` holds `renderingMutex` for the whole `render()` including `displayBuffer()`, and `main.cpp` samples the buttons exactly once per loop pass. So once the first press starts a repaint, the second parks the main loop for the rest of the refresh. `InputManager::update()` derives edges from level changes and clears them every call with no queue, so a press that starts *and* ends inside that window is never observed at all. Each accepted step also costs its own serialized render.

### The fix

The field the lock was protecting is not `selected` (only the main task writes that) but `top`, which the render task owns through `ListNav::onListRendered()`, and which `follow()` writes. `ListNav::syncToProps()` already resolves the viewport under `followOnBuild`, on the render task. So the main task sets the flag and lets the build do the follow: nothing blocks the input path, and `top` becomes single-writer.

Removing the lock without deferring the follow would also stop the stall, but would leave `top` written from both tasks, so this keeps #2957's viewport model rather than working around it.

### Behaviour

| Scenario | Today | With this PR |
|---|---|---|
| Tap Down 6× during one list refresh | 2–4 register, each its own refresh | all 6, 1–2 refreshes |
| Down, Down, Confirm within a second | opens the row one below | opens the row two below |
| Hold Down while a refresh is in flight | one page per refresh | pages every ~500 ms |
| Button polling while a list repaints | stopped for the whole e-ink write | continuous |

Hold-to-page now draws fewer intermediate pages, since steps coalesce into one render. The landing position is unchanged.

## Scope Check

## Additional Context

* **This is an existing pattern, not a new one.** `TextSettingsActivity.cpp:155-158` and `:483-486` already do `n.selected = …; n.followOnBuild = true; requestUpdate();` unlocked on the main task, as does `ReaderToolbarUi.cpp:284`. `OpdsBookBrowserActivity` and `WifiSelectionActivity` mutate their own `ListNav` unlocked too.
* **Side benefit:** `follow()` now runs *after* `visibleRows` has been measured for the current band, instead of using the previous build's estimate, which is the order `syncToProps` was written for.
* **Coalescing is guaranteed:** `renderTaskLoop` does `ulTaskNotifyTake(pdTRUE, …)` *before* `render()`, so a notify issued during a render always schedules exactly one more; N notifies collapse to one.
* **Checked so it needn't be re-derived in review:** all 16 `UiListActivity`/`UiTabListActivity` subclasses were swept for main-task reads of `top` / `drawnRows` / `pageRows()`. `EpubReaderChapterSelectionActivity.cpp:154` (`refreshTocWindow(nav.top)`) runs on the **render** task from `buildScreen`, immediately after `syncListViewport()`, so it still reads the post-follow `top` of the frame being drawn. That ordering is now load-bearing. The `pageRows()` reads in the continuous-nav lambdas were already outside the lock (arguments evaluate before the call). Worst case is one frame where the viewport did not follow, corrected by the already-queued render; no pointer or lifetime hazard, since only two scalars are written.
* **Deliberately not changed:** swipe-scroll (`UiListActivity.cpp:89-102`) keeps its blocking lock: `top` is genuinely two-writer for a relative scroll and there is no deferral rail for that (touch boards only). `FileBrowserActivity`'s SD `loadFiles()` under a lock on Confirm/Back is a one-shot action, not a mashed path. `SettingsActivity.cpp:152` (`activeNav().top = 0`) is left alone on purpose: `selectCategory()` runs *before* its callers write `selected`, so arming there could consume the flag against the previous tab's ring position. Arming must always be the last write.
* **Honest caveat:** `selected` is not single-writer on tab lists: `syncTabListViewport:105` clamps it from the render task. That is pre-existing and this PR neither creates nor worsens it.
* **Memory / flash:** `gh_release` flash **−70 B**, static RAM unchanged (56,640 B). No heap allocation, no new members.
* **Checks run locally:** `pio run -e default -e gh_release` both succeed, `./bin/clang-format-fix -g` clean, `pio check` (cppcheck) reports no defects on the touched files.

### Hardware testing

Tested on an Xteink X3 (ESP32-C3, UC8253) running current `develop` plus this commit, `gh_release` build flashed to the app slot.

* **The regression case passes:** mashing Down six times in a long file-browser list now lands six rows down in one or two refreshes.
* **Regression guards all behave:** wrap-around from the last row; Confirm acting on the row actually navigated to after two fast steps; Recents short-Confirm vs long-press-Confirm; hold-to-page and back; the Settings tab ring (`UiTabListActivity`); chapter selection on a long TOC (the `refreshTocWindow` ordering check); and File Browser enter/Back navigation.

Not tested on a touch board (X4 Pro / Paper Mono). The swipe-scroll path is unchanged, but a second pair of eyes there would be welcome.

---

### AI Usage

Did you use AI tools to help write this code? _**YES**_. Root-cause analysis, implementation and PR text with Claude Code; reviewed and hardware-tested by the author.
https://claude.ai/code/session_01NDaDZA34Z4Zvn7oqmAvtvG


Techneaux:
For context: the same user report also covered the EPUB reader's page turns, which is a separate mechanism (a latched `pendingManualTurn` rather than a blocking lock). Filed with root-cause analysis and options at #3536 — no PR for it yet, since the approach is worth agreeing on first.

itsthisjustin (approved):

clb92:
Nice work, this is annoying me a bit in 1.6.0

PR #3547, opened 2026-09-14
fix: Let OPDS catalog screens auto-sleep

Uri-Tauber:
## Summary

We already defines distinct loading, browsing, downloading, error, and search states, but unconditionally prevents sleep in OpdsBookBrowserActivity.
Now prevents sleep only while connecting, loading, searching, or downloading; idle browsing and error screens may sleep.

---

### AI Usage


itsthisjustin:
Guess I don't see the point in this but I also don't see the downside. If I'm connected to wifi and browsing something idk why I'd ever want it to sleep

Uri-Tauber:
@itsthisjustin I saw someone complained they left the device on browsing and forgot to turn off, just to discover a drained battery tomorrow.

prokrypt:
> Guess I don't see the point in this but I also don't see the downside. If I'm connected to wifi and browsing something idk why I'd ever want it to sleep

You've never fallen asleep in the middle of trying to get to page 131 out of 157 in a long list of books? 😂

itsthisjustin:
> @itsthisjustin I saw someone complained they left the device on browsing and forgot to turn off, just to discover a drained battery tomorrow.

Any risk in just sleeping without doing proper wifi teardown?

Uri-Tauber:
> Any risk in just sleeping without doing proper wifi teardown?

`enterDeepSleep()` disconnects Wi‑Fi, and turns the radio off before deep sleep.

itsthisjustin (approved):

PR #3549, opened 2026-09-14
feat: Rename bookmarks

Uri-Tauber:
## Summary

Add option to rename bookmark, instead of using a sentence from the bookmarked page.

When Holding *confirm* a popup menu in opened instead of *delete*. The menu have three entries: open, rename, delete.

Backwards compatible with existing bookmarks.

---

### AI Usage


Uri-Tauber:
Currently `delete` and `rename` use left/right buttons. I should think on better UX. 
Also, haven't tested on touch device yet.

Uri-Tauber:
Simulator recording

Screencast_20260914_215048

itsthisjustin (approved):

PR #3556, opened 2026-09-14
fix: warp lists in library

Uri-Tauber:
## Summary

Pressing _up_ when standing on the first title in library should jump to the last title - not go up to tabs bar.
Currently, there isnt a way to get to the end of the list without scrolling all the items, since _up_ in tabs bar opens _search_.
Granted, its not terrible since the user can reverse the list order, but still.

Riding along another small UX change: move Text Settings in legacy reader menu to end of the list - making it more accessible than before - just one press up.


itsthisjustin:
You can just press and hold to page cant you? Seems harder to toggle tabs with this no?

Uri-Tauber:
@itsthisjustin Think you have 300 books, and you looking for one down the list. Its 10 down presses vs one up press. 
Also, you can still get back to tab bar with _back_.
Overall, it make the UI more consistent with the rest of the firmware.

itsthisjustin (approved):

itsthisjustin:
Lol just caught the typo in the commit title. Oh well

PR #3563, opened 2026-09-15
feat: Add AboutActivity for device information display

itsthisjustin:
Implements a read-only activity that displays hardware information (device profile, display controller, touch, frontlight, RTC, IMU), firmware version, and live diagnostics like free heap memory.


lib/I18n/translations/english.yaml
```diff
@@ -162,6 +162,18 @@ STR_KOREADER_SYNC: "KOReader Sync"
...
+STR_ABOUT_FIRMWARE: "Firmware"
+STR_ABOUT_CHIP: "Chip"
+STR_ABOUT_FLASH: "Flash"
+STR_ABOUT_DISPLAY: "Display Controller"
+STR_ABOUT_RESOLUTION: "Resolution"
+STR_ABOUT_FRONTLIGHT: "Frontlight"
+STR_ABOUT_RTC: "Clock (RTC)"
+STR_ABOUT_IMU: "Tilt Sensor (IMU)"
+STR_ABOUT_FREE_HEAP: "Free Memory"
+STR_ABOUT_MAC: "MAC Address"
```
Uri-Tauber: I prefer to have those as hardcoded strings. Not something to be translated.

itsthisjustin: Ok

itsthisjustin: This is done now

src/activities/settings/AboutActivity.cpp
```diff
@@ -0,0 +1,150 @@
...
+  ITEM_MAC,
+};
+
+// Deliberately hardcoded English, exempt from the tr() rule: support reads
+// these screenshots across every device language, so the labels must be
+// identical on every unit.
+const char* const menuNames[AboutActivity::ITEM_COUNT] = {
+    "Device",      "Firmware",    "Chip",       "Flash", "Display Controller",
+    "Resolution",  "Touch",       "Frontlight", "RTC",   "Tilt Sensor (IMU)",
+    "Free Memory", "MAC Address",
```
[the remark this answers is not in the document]
itsthisjustin: We did this on purpose to make remote debugging with users to avoid translating strings on our end when telling them to read a value off the screen

Uri-Tauber (approved):
LGTM
It would make bug reports much easier to reproduce.

Uri-Tauber:
Two nitpicks:
1. Free memory seems to be useless - always the same number (which make sense).
2. I think **Language** should be the last item under **System** - to help someone who accidently chose the wrong language to find it quickly.

itsthisjustin:
> Two nitpicks:
> 1. Free memory seems to be useless - always the same number (which make sense).
> 2. I think **Language** should be the last item under **System** - to help someone who accidently chose the wrong language to find it quickly. 

Feel free to edit. Not wed to either of these. I mostly just need this to get display variant info from users

BooBSD:
What about USB mass storage, WiFi version and speed info?

itsthisjustin:
> What about USB mass storage, WiFi version and speed info?

Meaning showing the usb mass storage as available or not as a feature? Couldn't you just go into file transfer to see that?

Uri-Tauber:
@BooBSD You can find this info by looking the chip online. 
The main goal of this activity is helping identify something you cant know from the device model alone - screen variant. x3 and x4pro ship with different screen controllers, which can make debugging bug reports really tedious.
All the other info is pretty redundant - nothing you cant find with Google in half a minute.

PR #3569, opened 2026-09-15
perf: keep settings list at its initial allocation

serialx:
## Summary

Split out of #3528 at @Uri-Tauber's request.

`getSettingsList()` builds its static base list from an initializer list, which allocates exactly the entry count, and then inserts the tilt page-turn entry (X3) and the double-click frontlight entry (X4 Pro) into that full vector. Each insert reallocates the vector at double capacity, and because the list is a function-local static, the unused half stays allocated for the process lifetime.

On the X3 that is 58 entries doubled to 116 slots, and `sizeof(SettingInfo)` is 132 B on the C3, so about 7.5 KB of heap is held by empty slots. This PR declares both entries in the initializer list and erases them when the hardware is absent, so the list keeps its initial allocation. The visible order of the settings is unchanged.

## Scope Check

- [x] The stock firmware does not already handle this well, and no other popular CrossPoint fork already does.

### Measurement

Xteink X3, `develop` `79a28ab3` vs this branch `c606eb59`, both with the same serial-control test commit so the device can be driven over USB. 데미안, Pretendard 16 pt, portrait, justified, Extra Wide, anti-aliased text, default spacing. Builds flashed alternately, three boots each; every boot deleted the book cache, reopened cold at the same position, did six warmup and six measured page turns over the identical sequence: **18 measured turns per build**.

| Measurement | `develop` `79a28ab3` | PR `c606eb59` | Difference |
|---|---:|---:|---:|
| Median page drawing (`bw_render`) | 75.5 ms | 75.0 ms | −0.5 ms |
| Median total rendering, including display refresh | 1,419.0 ms | 1,415.5 ms | −3.5 ms |
| Median cold layout preparation (cache miss to first page render) | 4,502 ms | 4,466 ms | −36 ms |
| Median free heap after measured pages | 49,736 B | 53,640 B | +3,904 B |
| Largest free block after measured pages | 36,852 B | 21,492 B\* | −15,360 B\* |
| Lowest heap watermark (`Min Free`) | 28,060 B | 24,080 B\* | −3,980 B\* |
| Static RAM | 56,664 B | 56,664 B | 0 B |
| Flash | 5,557,441 B | 5,558,319 B | +878 B |

\* **Side effect of the extra headroom, not fragmentation caused by this change:** `SdCardFont` keeps its font mini-data resident only when free heap is at least 40 KiB at cleanup. On this workload `develop` sits just below that line and frees the data after each page (the first heap sample after a page shows about 70 KB free), while this build sits above it and retains the data (about 41 KB free). The retained arena is what lowers the largest free block and the low-water mark here; in exchange the next page's fonts are already loaded, which is where the 36 ms faster cold layout comes from. Whether the 40 KiB threshold is the right one is a separate question from this PR.

Default firmware build passes; 335 host tests pass.

### AI Usage


serialx:
BTW: The papermono CI build error seem to be Github CDN error, not related to this PR.

Uri-Tauber (approved):
LGTM

PR #3573, opened 2026-09-15
fix: speed up long status titles

### LINKED ISSUE #2751: Slow page turn with long chapter names
### Affected Version

1.5

### Summary

Page-turn latency scales with the length of the current chapter's `navLabel`
text (from `toc.ncx`, or the equivalent link text in an EPUB3 `nav.xhtml`),
when the status bar's chapter-title element is enabled. On affected books,
this adds multiple seconds *per page turn*, independent of the Embedded
Style setting.

### Reproduction

Two EPUBs from the same series/publisher, same reading settings, same
device (Xteink X4):

| | `toc.ncx` navLabels | Page-turn time |
|---|---|---|
| Book A (fine) | Short titles, e.g. `"Chapter 12"` (avg ~10 chars, max ~22 chars) | < 1s |
| Book B (slow) | Full first sentence of each chapter (avg ~450 chars, max ~1,470 chars) | ~2.75s |

Both books:
- Have 0 embedded fonts, comparable chapter file sizes, comparable CSS rule
  counts, comparable inline-markup density (spans/tags per 1000 characters).
- Show the *same* page-turn time whether **Embedded Style is ON or OFF** --
  ruling out CSS/style resolution as the cause, since that's the one setting
  that's supposed to bypass the book's own styling.
- Are structurally well-formed (valid XML, standard `container.xml` /
  `content.opf` / `toc.ncx`, no duplicate ids, no unusual zip compression).

The only material difference found between the two books, after comparing
zip structure, XML validity, CSS, font-family declarations, markup density,
and DOM structure, was `navLabel` text length. Manually truncating the
`navLabel` `<text>` content in Book B's `toc.ncx` to ~5 words per chapter
(e.g. `"Like most forms of corruption\u2026"` instead of the full sentence)
resolved the slowdown; hiding the chapter-title status bar element entirely
also immediately resolved it, without any file changes.

### Likely mechanism

The status bar's chapter-title element appears to re-resolve and/or
re-measure/truncate the current chapter's label text on every page render,
rather than once per chapter. If the source text is long (hundreds to
thousands of characters, as with EPUBs whose spine/nav generator used the
first sentence of each chapter as its label instead of a short title), this
becomes a meaningful per-page cost on the ESP32-C3's limited CPU -- even
though only a short, possibly-ellipsized substring is ever actually
displayed.

### Suggested fix

Cache the *display* label (already truncated/ellipsized to fit the status
bar width) per chapter index, rather than recomputing it from the full
`navLabel`/nav-doc text on every page:

```
if current_chapter_index != cached_label_chapter_index:
    display_label = compute_and_truncate(navLabel_text[current_chapter_index])
    cached_label_chapter_index = current_chapter_index
    cached_display_label = display_label
draw_status_bar(cached_display_label)
```

This would need only to invalidate/recompute on chapter boundary crossings
(which are already detected for other purposes, e.g. pre-indexing the next
chapter per the 1.2.0 release notes), not on every page turn within a
chapter. This fixes the problem for *all* existing EPUBs -- including ones
with long navLabels that users haven't manually cleaned up -- without
requiring any change to the book file itself, and without needing users to
discover "hide the status bar" as a workaround.

As a secondary, smaller improvement: if the label source string exceeds some
threshold (e.g. a couple hundred characters), it's reasonable to assume the
generator used first-sentence-as-title rather than an intentional short
title, and truncate to the first N words once at book-index time rather
than per status-bar-draw.

### Environment

- Device: Xteink X4
- Firmware: CrossPoint
- EPUB generator: Calibre 9.5.0 (both books), one from a Calibre-normalized
  source, one from an InDesign-derived `epub` with per-chapter navLabels set
  to the chapter's opening sentence.

### Steps to Reproduce

Open an epub with long chapter names (ie: auto generated chapter names), try to turn the page after opening the book.

### Expected Behavior

Regular page turn speed

```shell

```

Uri-Tauber:
## Summary

Should fix #2751.

---

### AI Usage


serialx:
Tested `feae5c15` on an **Xteink X3** against merge base `3555ff55`. The fix resolves the slowdown in a fixture with a **1,470-character chapter title**.

Three interleaved boots per build, six warmup turns and six measured turns per boot: **18 measured turns per side**, with identical settings and serial-control instrumentation.

| Measurement | Baseline | PR | Difference |
|---|---:|---:|---:|
| Median B&W drawing | 10,131 ms | 125.5 ms | −98.8% |
| Median total rendering, including refresh | 21,553.5 ms | 1,470.5 ms | −93.2% |
| Cold cache → first completed render | 18,301 ms | 3,958 ms | −14,343 ms |
| Post-turn free heap | 90,004 B | 90,004 B | 0 |
| Largest free block | 81,908 B | 81,908 B | 0 |
| Lowest heap watermark | 70,612 B | 70,964 B | +352 B |
| Static RAM | 56,688 B | 56,688 B | 0 |
| Flash | 5,563,411 B | 5,563,423 B | +12 B |

The improvement comes from avoiding repeated full-title measurements during both prewarming and drawing. No persistent heap regression was observed; the small watermark difference should not be interpreted as lasting memory savings.

All page sequences matched. Short titles, mixed UTF-8 titles, and all four orientations passed. Across 14 screenshot comparisons, page bodies were identical; complete framebuffers matched when battery and progress counters matched. The host comparison also passed 113,858 truncation cases without mismatches.

Coverage is X3 with synthetic fixtures; I haven’t tested the original reporter’s EPUB or other hardware.

serialx (approved):
LGTM.

Better alternative would need getTextBounds like method but calculates max character count with bounded width in EpdFont.

I think because this "slow" path only triggers with long chapter names, it seems OK.

Possible future optimization target, maybe?

Issue #3564, opened 2026-09-15
WiFi sessions permanently fragment the heap on touch boards — the silentRestart() defrag never runs


suhsoohong:
### Device

Xteink X4 Pro

### Affected Version

`1.6.5rc-6-g6c83eddb` (code unchanged since 1.6.0; measurements below taken on `1.6.0rc-67`)

---

## Describe the Bug

`silentRestart()` is the firmware's remedy for WiFi-induced heap fragmentation, and
the comment above `deepSleepInProgress` says so:

> WiFi activities call `silentRestart()` in `onExit()` **to clear heap fragmentation on the way out**

Every WiFi activity ends by calling it: Clock Sync, OTA, Calibre, OPDS, font
download, KOReader sync, the web server.

On touch boards it never reaches `ESP.restart()` (`src/main.cpp`, ~L153–L172):

```cpp
static bool finishWifiSessionWithoutRestart() {
  if (!BoardConfig::hasTouch()) return false;
  // A software reset does not cycle externally powered touch/frontlight rails.
  // Shut down the network stack in place so those peripherals retain state.
  if (esp_sntp_enabled()) esp_sntp_stop();
  WiFi.mode(WIFI_OFF);
  delay(100);
  LOG_DBG("MAIN", "WiFi stopped without restart on touch device");
  return true;
}

void silentRestart() {
  if (deepSleepInProgress) return;
#if FREEINK_CAP_TOUCH
  if (finishWifiSessionWithoutRestart()) return;   // <-- returns here on X4 Pro
#endif
  ...
  ESP.restart();
}
```

The X4 Pro qualifies on both tests: `FREEINK_CAP_TOUCH` includes
`FREEINK_DEVICE_X4PRO`, and `hasTouch()` is true at runtime (GT911 at 0x5D).

**The early return is deliberate and the reasoning is sound** — a software reset
does not cycle externally powered rails, and on this board the GT911 sits on an
active-low rail alongside the frontlight, so rebooting to defrag risks leaving
touch or the frontlight in a bad state. This isn't an oversight. What's missing
is any handling of the consequence: the fragmentation the reboot exists to clear
is simply never cleared on these boards.

## Steps to Reproduce

1. Cold-boot the device. From the home screen, read free heap and max contiguous
   allocation (`heap_caps_get_largest_free_block`).
2. Settings → Clock Sync. Let NTP complete. Read again.
3. Exit back to the home screen. Read again, several times over ~30s.

[...]

PR #3575, opened 2026-09-16
fix: preserve frontlight state across silent restarts

uxjulia:
## Summary

  * Ports the frontlight state fix from Crossink to keep the frontlight unchanged through silent restarts

  * Capture the live `Frontlight.isOn()` state in RTC memory alongside the existing silent-restart target.
  * Restore that captured state after a silent restart instead of deriving it from “Restore Light on Wake.”
  * Apply the same restart setup for Home, reader, and storage-handoff restart paths; normal boots and wake behavior remain unchanged.

## Additional Context

* The change adds one RTC-resident 32-bit payload and no heap allocation.


Uri-Tauber (approved):
LGTM

itsthisjustin:
> LGTM
> 

Don't merge just yet. I am going to actually bring the soft reboot back that I disabled on the pro in this same PR

itsthisjustin:
Closes #3564

itsthisjustin (approved):

PR #3578, opened 2026-09-16
chore: Update license

Uri-Tauber:
see title


LICENSE
```diff
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (c) 2025 Dave Allie
+Copyright (c) 2025 CrossPoint organization
```
itsthisjustin: CrossPoint Reader instead?

PR #3589, opened 2026-09-16
feat: add ability to clip/highlight text for EPUBs

uxjulia:
## Summary

**Note**: Extended on-device testing is highly recommended using various EPUBs with different languages and glyphs. Only preliminary on-device testing has been done. This PR is meant to act as a baseline to be further tested and built upon as needed.

  * Ports Crossink's implementation which allows EPUB clipping and highlight support so readers can save passages, see them highlighted, revisit or delete them, and export them in Kindle-compatible format via a `My Clippings.txt`

  * Adds **Create Clipping** and **View Clippings** actions, plus configurable long-press Confirm and short power-button shortcuts.
  * Supports button-based and touch selection, including bounded selection across up to three preloaded pages.
  * Stores per-book clipping data under `/.crosspoint/clippings`, with versioned writes, migration when books move to `/Read`, a 256-clipping limit, and layout-safe highlight anchors.
  * Adds gray in-reader highlights, bounded UTF-8-safe list previews, full-text detail view, and confirmed deletion.
  * Appends saved excerpts to `/My Clippings.txt` with Kindle-style book, location, chapter, and timestamp information.
  * Adds translations, user-guide documentation, and 13 host tests for preview truncation, UTF-8 handling, whitespace normalization, malformed reads, record boundaries, and buffer reuse.

## Additional Context

* New highlights are shown only when the current page count and reader layout match the saved anchor. Legacy records remain readable in the clipping list but are not painted without a valid layout signature.
* Full clipping text is stored on the SD card and capped at 4 KiB; list previews are capped at 256 bytes and reuse an eight-row window. Kindle export is capped at 2 KiB per excerpt.
* If exporting to `My Clippings.txt` fails, the in-book clipping is rolled back where possible and a failure message is shown.

---

### AI Usage

Did you use AI tools to help write this code? _**< YES>**_


uxjulia:
Note: The 256 limit is a C3 constraint, S3 devices can probably be increased to something like 1024. I have a user that forked Crossink that said 1024 has been working fine for them. Something to consider.

serialx:
Normal-reading QA on **X3 / `demian.epub`**: PR `ee1237db` versus its `develop` merge base `19a63e0d`, with identical instrumentation. Three cold-start runs and **72 measured renders per build**; no clipping actions.

| Measurement | Baseline | PR | Delta |
|---|---:|---:|---:|
| Median page drawing | 74 ms | 74 ms | 0 ms |
| Median rendering, including refresh | 1,414.5 ms | 1,417.5 ms | +3 ms |
| Median cold layout | 4,472 ms | 4,503 ms | +31 ms |
| Median free heap after reading | 52,276 B | 51,704 B | −572 B |
| Median largest free block | 18,420 B | 18,420 B | 0 B |
| Lowest free-heap watermark | 17,212 B | 16,896 B | −316 B |
| Static RAM | 56,688 B | 56,808 B | +120 B |
| Flash | 5,563,643 B | 5,597,225 B | +33,582 B |

No growing heap deficit during measured forward reading. Paired page bodies matched. The small persistent cost is consistent with clipping storage reserving records and metadata.

Original reading position restored. Clipping-specific QA remains pending.

serialx:
Nice feature! I did some agent based automated QA based on my PR #3510. I've verified a few issues by hand. And added one original human issue (3. Navigation gap: saved clippings cannot return to the passage.)

Hope this helps!

---

Tested on **Xteink X3**, PR `ee1237db`, with USB-control instrumentation. **Korean version of Demian; used Pretendard 16 pt.** Synthetic text tests used Pretendard 8 pt, margin 5, justified alignment, Extra Wide line spacing, extra paragraph spacing On and hyphenation Off. Test inputs and setup are included below.

**Human verified**

**1. Saved highlights disappear during ordinary reading.** This requires a long chapter whose pagination is still partial: clear its book cache, reopen, save a word on the first page, read forward far enough to extend pagination, then return. In the Korean Demian test, the exact sequence was spine 4/page 1 (zero-based spine), 12 pages forward and 12 back. The highlight disappeared, while its excerpt remained in View Clippings. Cache rebuilds also reproduced this. `src/ClippingStore.h:94` requires the saved page count to match the current count, which grows during incremental pagination.

**2. Clipping-list focus disagrees with the opened item.** Save two distinct clips, open View Clippings, press Down once, wait for the display to settle, then Confirm. The first excerpt is visibly focused, but the second opens. Delete uses the same underlying selection. `src/activities/reader/EpubReaderClippingListActivity.cpp:177` passes `true` as the third `syncListViewport` argument; that argument is an integer selection offset, not `hasSubtitle`.

**3. Navigation gap: saved clippings cannot return to the passage.** Open View Clippings and select an entry: it opens the excerpt detail, with no action to jump to the page containing the clipping. I expected a saved clipping to provide a way back to its location in the book. This is a missing navigation feature, separate from the incorrect list focus above.

**5. Short Power → Create Clipping stores the wrong enum value.** Select it in Settings → Controls, then inspect `shortPwrBtn` in `/.crosspoint/settings.json`: it is **5**, although the reader's Create Clipping action is **6**. Both the device UI and API display Create Clipping. The non-touch options at `src/SettingsList.h:358` omit Confirm without remapping indices. Manual device testing also confirmed that this shortcut does not start clipping.

**Agent tested only — not human verified**

**4. File Transfer failed after growing from 255 to 256 clips.** Start with a disposable book containing 255 saved records; a helper below creates that setup from one clipping. Reboot so the 255-entry index is freshly loaded, open the book, save one more clip, then enter File Transfer without another reboot. Connect a host to its hotspot and try HTTP access. Reader Free/MaxAlloc fell from **97,440/81,908 B to 76,184/32,756 B**. File Transfer reached **7,048/2,036 B**; seven HTTP connection attempts timed out while serial remained responsive. Rebooting before File Transfer recovered access. The index growth adds approximately 20,400 B, retained by `src/ClippingStore.cpp:114–121` after leaving the book. The precise HTTP failure cause remains unconfirmed. A separate settings-download/serial stall with 256 records also required a reset.

**6. Byte truncation splits UTF-8 and shortens the saved range.** Put a paragraph of numbered tokens in a test EPUB, generated by `" ".join(f"abcdefghijklmnoéabécdefg{i:03d}" for i in range(230))`. At the tested settings, selecting tokens 000–185 across three pages covers 5,579 bytes. Storage cut this to **4,096 bytes** and export to **2,000**, both inside an `é`, while the highlight still covered the full selection. Separately, use 30 `é` characters as the EPUB navigation chapter title and save a clip: its stored title is cut inside a character at 47 bytes.

**7. Excerpt reconstruction changes source text.** Add `<p>well-<br/>known</p>` to a test EPUB chapter and select both displayed words: it saves as `wellknown`. In the Korean Demian test, selecting across visual line wraps also inserted spaces inside words, e.g. `정겨운` → `정겨 운`. Both paths are in `src/activities/reader/ClipSelectionActivity.cpp:271–286`.

**8. Detail view removes literal tag-shaped text.** Add `<p>Literal &lt;tag&gt; and 3 &lt; 5 &amp; 8 &gt; 2 must stay.</p>` to a test EPUB chapter. Clip the displayed sentence, then open its detail: `<tag>` disappears there, although storage/export preserve it. The plain-text detail path calls `htmlToPlainText()` at `src/activities/reader/DictionaryDefinitionActivity.cpp:41`.

**9. The 17th range saves but is invisible.** Use a fully paginated chapter to avoid the changing-page-count issue. On one text page, create 16 disjoint one-word clippings and verify their highlights are visible. Save a 17th word on that same page: its record exists, but its highlight is absent. Delete the first record and the new range becomes visible. The automated test seeded the initial 16 records, then performed the save/delete through the UI. Rendering collects only 16 matching ranges (`src/activities/reader/EpubReaderActivity.cpp:1916–1931`).

**10. Selection button hints cover text.** At 8 pt/margin 5, enter selection in inverted portrait or either landscape orientation. Hints overlap the first paragraph or line edges; returning to the reader exposes the obscured text.

Three additional cold PR boots with **16 saved clips / 72 measured turns** used retained merge-base (`19a63e0d`) measurements with matching book, settings and instrumentation. Per boot: clear the book cache; start at spine 3/page 1; make six warmup turns, then 24 measured turns; wait 13 seconds after warmup and every six measured turns for heap samples; then turn back 30 pages. All forward spine/page sequences matched:

| Measurement | Baseline | PR | Delta |
|---|---:|---:|---:|
| Median page drawing | 74 ms | 75 ms | +1 ms |
| Median total rendering | 1,414.5 ms | 1,419.5 ms | +5 ms |
| Median cold layout | 4,472 ms | 4,511 ms | +39 ms |
| Settled free heap | 52,276 B | 50,608 B | −1,668 B |
| Largest free block, median | 18,420 B | 34,804 B | +16,384 B |
| Lowest Min Free | 17,212 B | 15,608 B | −1,604 B |
| Static RAM | 56,688 B | 56,808 B | +120 B |
| Flash | 5,563,643 B | 5,597,225 B | +33,582 B |

**Measurement limits:** the highlight-persistence defect suppresses saved ranges, so these numbers do not establish continuous active-highlight performance. PR largest-block results were 18,420/34,804/34,804 B; the larger median is not evidence of a general fragmentation improvement.

Twenty create/cancel cycles and twenty save/view/delete cycles showed no growing settled-heap deficit. The 257th save was safely rejected without changing store/export bytes; deleting one allowed another save. Original device data and settings were restored after QA.


<details>
<summary>255-record setup without creating 255 clippings manually</summary>

Use a disposable EPUB with exactly one saved clipping. Close the book and download its file from `/.crosspoint/clippings/epub_<book-hash>.bin`; the file header contains the book path so you can identify the correct store. Keep the original file and `/My Clippings.txt` backed up. Save this script as `seed-255.py`, then run `python3 seed-255.py /path/to/downloaded-store.bin`:

```python
import struct
import sys
from pathlib import Path

source = Path(sys.argv[1])
data = source.read_bytes()
assert struct.unpack_from("<BH", data) == (3, 1), "Need a v3 file with exactly one clipping"
offset = 3
for _ in range(3):  # length-prefixed book title, author, path
    size = struct.unpack_from("<I", data, offset)[0]
    offset += 4 + size
record = data[offset:]
# v3 record: 24-byte anchor, 48-byte chapter title, uint16 text size, text.
assert len(record) >= 74
assert len(record) == 74 + struct.unpack_from("<H", record, 72)[0]
header = bytearray(data[:offset])
struct.pack_into("<H", header, 1, 255)
output = source.with_name("capacity-255.bin")
with output.open("xb") as handle:  # refuse to overwrite an existing output
    handle.write(header + record * 255)
print(output)

```

The script writes a new file containing 255 duplicate records; it does not change the downloaded original. With the reader closed, replace only that disposable book's on-device store with the output, preserving the original `epub_<book-hash>.bin` filename. Reboot, open the book, and save one new clipping through the UI. This exercises the 255→256 index growth; distinct highlighted passages are not required for this case. Restore the backed-up store/export after testing.

The helper's generated format was checked locally against the PR's v3 record format. The on-device QA used a separately prepared 255-record seed; this helper has not had an additional device run.

</details>

uxjulia:
> Hey @uxjulia. Just noticed this now. Thanks for bringing this upstream!
> 
> One thing I like to know is why you choose to keep the highlight location as absolute page numbers, e.g. "page 5, word 12 to page 5, word 20." This is quite fragile in my eyes. For some re-flow changes the reader simply remove the highlight. For others (the new word spacing, for example), nothing will be done - resulting in the wrong words get highlighted.
> 
> Why not use offset in xhtml file, like bookmarks and kosync do?

Agreed, offsets would be more robust and are probably the better long-term approach. I implemented this feature in CrossInk way before even thinking about the kosync-like offset approach, so that's the only reason this port is using page and word location. If you want to go the offset route, we can update this PR, it'd just require a bit more regression testing since that's not the approach Crossink uses so there would be more unknowns.

Uri-Tauber:
> Agreed, offsets would be more robust and are probably the better long-term approach. I implemented this feature in CrossInk way before even thinking about the kosync-like offset approach, so that's the only reason this port is using page and word location. If you want to go the offset route, we can update this PR, it'd just require a bit more regression testing since that's not the approach Crossink uses so there would be more unknowns.

Awesome. I would like that.

PR #3594, opened 2026-09-17
fix(ci): pin pioarduino core inside the platform penv

serialx:
## Summary

Every PR has failed `Build default` and `Build sticky` since 2026-09-15 14:11 UTC, regardless of its contents. The other three build envs pass. This change pins the PlatformIO core used by the platform's nested build to the same version as the outer core, and rotates the cache key so runners drop the poisoned penv.

## Symptom

```
*** [.pio/build/default/firmware.elf] ModuleNotFoundError : No module named 'SCons.Tool.FortranCommon'
```

## Cause

pioarduino released core 6.2.0 on 2026-09-15 (PyPI upload 14:12 UTC; our last green run was 13:50 UTC). That core requires SCons 4.11.1. Our pinned platform, 55.03.311, still declares SCons 4.8.1 in its `platform.json` and reinstalls it whenever it sees another version (`platform.py`, `_handle_existing_tool`). Under core 6.2.0 the two reinstall over each other on every run, and the SCons process that is already running loses its own package directory. SCons imports its link helper lazily at link time, so the import fails there. The module does exist in 4.11.1; the error is a symptom of the wipe, not a missing file. Upstream tracks this as pioarduino/platform-espressif32#529, and the maintainer states 55.03.311 is not compatible with core 6.2.0.

CI already pins the outer core to 6.1.19, which is why `x4pro`, `x4c` and `papermono` pass. `default` and `sticky` extend `firmware_tuned`, whose `custom_sdkconfig` triggers the hybrid build: the platform compiles the IDF libraries, then spawns a second `pio run` from `~/.platformio/penv`. The platform bootstraps that penv with an open-ended `pioarduino>=6.1.19`, which now resolves to 6.2.0. The failing log shows the moment it happens: `Pio has been successfully upgraded to 6.2.0!`, two tool-scons installs, then the link failure 18 seconds later. The cache carries that penv between runs, so every PR hits it.

## Change

- New step in the build job, after the cache restore: run `pio pkg install` so the platform creates the penv, then force `pioarduino==6.1.19` into it. The platform's bootstrap leaves a satisfying version alone on later runs.
- Cache key `pio-` becomes `pio-v2-` in both jobs so the cached penv with 6.2.0 is discarded.

## Alternatives considered

- Upgrading the outer core to 6.2.0 does not help. A local build with 6.2.0 on both sides reproduces the same error in 23 seconds, because the platform's 4.8.1 pin still fights the core.
- Waiting for a platform release that ships core 6.2.0 and SCons 4.11.1 is the eventual fix. Their develop branch has it, but with an open bug under verbose builds (pioarduino/platform-espressif32#531). When that release lands, this step and the pin can be removed together with the platform bump.

## Verification

- Reproduced the failure locally with core 6.2.0 (`pio run -e default` fails at link in 23 seconds with the FortranCommon error).
- Emulated CI end to end with a fresh `PLATFORMIO_CORE_DIR`: the platform bootstrapped a new penv with pioarduino 6.2.0, the pin step replaced it with 6.1.19, and `pio run -e default` then compiled the IDF libraries, spawned the nested Arduino build and linked successfully.
- CI on this PR is the final check: `Build default` and `Build sticky` must go green.

## Note for developers

Local machines are affected the same way. Anyone whose `~/.platformio/penv` picked up 6.2.0 can apply the same pin:

```
uv pip install --python ~/.platformio/penv/bin/python "pioarduino==6.1.19"
```


muhas:
To get rid of the error, for local build I changed deps in ~/.platformio/platforms/espressif32/platform.json
```
    "tool-scons": {                                                                                                                                                                                                  
      "type": "tool",                                                                                                                                                                                                
      "optional": true,                                                                                                                                                                                              
      "owner": "pioarduino",                                                                                                                                                                                         
      "package-version": "4.41101.0",                                                                                                                                                                                
      "version": "https://github.com/pioarduino/registry/releases/download/0.0.1/scons-4.10.1.zip"                                                                                                                   
    }  
```

Uri-Tauber:
Quoting Justin:

> claude says we should move to a lockfile after i argued that simply pinning that package is going to result in a waterfall of having to pin every package

serialx:
A lockfile does not solve the problem. The tool the platform integrates, pio-lock, records library versions, git SHAs and the toolchain packages under ~/.platformio/packages. The penv is a Python virtualenv the platform creates from its own hard-coded dependency list, with pioarduino>=6.1.19 open-ended. No PlatformIO lockfile controls that list.

itsthisjustin:
Reading the rest of your description at least says this isn't unique to us. Let's merge and check the fallout

itsthisjustin (approved):

Uri-Tauber:
Fine by me.
Thx @serialx!

Uri-Tauber (approved):

PR #3603, opened 2026-09-17
fix: Re-Enable toolbar reader menu on button-only devices

### LINKED ISSUE #3236: Toolbar reader menu UI/UX issues on x4
### Affected Version

develop branch with #3182

### Describe the Bug

The new toolbar reader menu from @jetaudio has great potential, but some unexpected (to me) UI/UX issues:

1. In reader mode, I have side buttons mapped to Up=next page and Down=previous page. Long-pressing the buttons moves in the analogous direction by chapter (i.e., long-press up moves to *next* chapter). However, in the menu, pressing up moves immediately to the *previous* chapter.
2. In the Contents and More submenus, the UI shows what appear to be menu page counts. However, these do not update as expected. For example, in the More submenu, it always shows page "1/2". Even scrolling to the bottom of the contents doesn't reach the final page count.
3. Button tooltips appear in the submenus but not in the main menu. This shifts the other UI elements around.
4. Button presses in the menu don't seem to work reliably. Not sure if this some kind of debounce issue around the screen update?

### Steps to Reproduce

1. Turn on "Toolbar" reader menu in settings
2. Read epub
3. Open menu

### Expected Behavior

Intuitive on-screen display tied to button presses as expected. (Note, I have no touch-screen device.)

```shell

```

itsthisjustin:
The toolbar reader menu is now available on button boards, not just touch devices. Button navigation works through the focused-tool pill and panel cursor. Also implements async overlay refresh to keep UI responsive during screen updates - the event loop continues polling input while the waveform runs, so chrome responds to taps and buttons immediately instead of blocking. Removes button hints from panel overlays.

Closes #3236 #3561


Uri-Tauber:
I found it a bit unintuitive the volume buttons move between chapters with short press when the menu is open.
Apart from that, Awesome work.

Uri-Tauber (approved):

itsthisjustin:
> I found it a bit unintuitive the volume buttons move between chapters with short press when the menu is open.
> Apart from that, Awesome work.

It's how you'd interact with the chapter menu above the toolbar without having to go up and back down which is kind of a pointless interaction that adds a click for no reason

PR #3605, opened 2026-09-17
fix(epub): avoid full section builds during random/backward navigation

0137:
## Summary

This PR fixes EPUB navigation failures on memory-constrained devices such as the Xteink X4.
Some EPUB content could be read normally page-by-page, but navigating to the same area with Go to % or by turning backward across a spine boundary could fail with: "Indexing failed - invalid book"

The issue was caused by these navigation paths forcing the target section to be paginated to completion in the foreground instead of using the existing incremental builder.

## Scope Check

      the relevant maintainer.

## Changes

- Remove the one-shot full section build used by Go to %.
- Replace the UINT16_MAX fake page target used for backward spine navigation with a dedicated pendingLastPageJump state.
- Let both operations complete through the existing incremental/background pagination path.
- Reuse the existing heap checks used by background pagination.
- Fix the percentage spine-boundary comparison so an exact boundary maps to the next spine.

## Testing

Tested on an Xteink X4 with an EPUB that reproduced the issue.
Before this change:
- several percentage jumps failed with Indexing failed - invalid book;
- backward navigation across some spine boundaries also failed;
- the same content could still be reached through normal reading or other navigation paths.
After this change, the previously failing percentage jumps and backward navigation succeed.

---

### AI Usage


serialx:
**Device test on an X3 (ESP32-C3, no PSRAM): not ready for merge**

Baseline: merge base 32f2b2aa. Candidate: 6efc18fd. Both builds had the serial-control PR (#3510) cherry-picked so the device could be driven from a script; that code is identical on both sides and cancels out of the numbers. Workload: demian.epub, book cache deleted before every boot, 3 interleaved boots per build.

**Summary:** the navigation fixes behave correctly, but moving these builds into the background exposes a CPU-frequency race that hung the board once, and makes the jumps about 3× slower.

### What works

- Go to % and backward spine crossing land on the same spine/page as baseline on every boot, and the landing page body is byte-identical (only the footer band differs).
- Back from End of book is now correct. On baseline the reader keeps a stale `pendingPageJump` afterwards and reports busy forever (3/3 boots); the candidate clears it (3/3).
- Ordinary page turns, heap after turns, and static RAM are unchanged. Flash is 120 B smaller.

### Problem 1: hard hang after a deferred build (1 of 7)

After one Go-to-% build finished, the firmware stopped completely: no page render, no serial replies, and the 10 s heap sampler stopped. USB still enumerated. Only a flash reset recovered it. The last log lines were:

```
[123722] [DBG] [PWR] Restoring normal CPU frequency
[123721] [DBG] [PWR] Restoring normal CPU frequency
```

Two tasks printed that in the same millisecond. The same pair appeared in 3 of 7 candidate builds and never on baseline. `HalPowerManager::setPowerSaving()` has no mutex (lib/hal/HalPowerManager.cpp:33), so when the deferred build completes, the main loop and the render task's `HalPowerManager::Lock` both call `setCpuFrequencyMhz()` concurrently. The race is pre-existing, but this PR creates the trigger: a multi-second build that drops into low-power mode, then wakes both tasks at the same moment.

### Problem 2: jumps are 2.6 to 3× slower

The deferred build doesn't reset the idle timer and isn't included in `skipLoopDelay()` (EpubReaderActivity.cpp:1136), so after 3 s the main loop drops the C3 to 10 MHz with 50 ms sleeps. Layout throughput falls from ~11 pages/s to ~1.4 pages/s, and the user watches the Indexing popup for 23 to 28 s instead of 9 s.

| Navigation, median of 3 (press to render) | Baseline | Candidate |
|---|---:|---:|
| Go to % into uncached spine 8 (53 pages) | 8.8 s | 23.0 s |
| Page-back into uncached spine 5 (62 pages) | 9.3 s | 28.1 s |
| Go to 100% (spine 12, 1 page) | 4.7 s | 2.4 s |
| Back from End of book | 1.5 s | 1.6 s |

### Standard rows

| S1: 6 measured turns from spine 3 page 1 | Baseline | Candidate | Diff |
|---|---:|---:|---:|
| Median page drawing (bw_render) | 75.0 ms | 75.0 ms | 0 |
| Median total rendering | 1,420 ms | 1,417.5 ms | -2.5 |
| Median cold layout preparation | 4,506 ms | 4,484 ms | -22 |
| Free heap after measured pages | 49,796 B | 49,736 B | -60 |
| Largest free block after measured pages | 28,660 B | 28,660 B | 0 |
| Static RAM / flash | 56,664 / 5,554,531 B | 56,664 / 5,554,411 B | 0 / -120 |

Free heap and largest block after the jumps were identical between builds. Heap stayed above 68 KB free during the builds on both sides, so I couldn't reproduce the original "Indexing failed" OOM with this book on an X3; the exact-boundary `<` fix was only checked by reading the code. Happy to re-measure once the PR is updated.

0137:
@serialx
Thanks for the detailed testing. I added pendingPercentJump and pendingLastPageJump to the condition in skipLoopDelay() so that these deferred navigation builds stay at normal CPU frequency and avoid the idle loop delay.

My original tests used a EPUB with relatively short spine items. Looking back at the logs, the device did enter low-power mode during pagination, but the remaining work finished soon afterward, so I missed the slowdown.

I also tested Go to % and backward spine navigation using Demian. With this change, navigation no longer appears slower than the baseline timings you reported, and there are no low-power transitions during pagination.

A brief low-power transition still occurs after pagination completes, before rendering starts. I have not reproduced the hang in my tests, but I cannot yet confirm that it is resolved.

PR #3608, opened 2026-09-18
feat: refresh library

Uri-Tauber:
## Summary

* 
Currently, the only way to manually rebuild library index (when adding new books, or moving things around) was in Settings->system, which make a little sense. This PR move it into library screen itself.

long press on items in Recent previously only deleted books. Now, it open popup menu with four entries: open, delete, remove from recent and refresh library.

Touch devices also get new button next to search.
Small SDK change made this possible.

### x4
screenshot-48303.bmp

### Sticky
screenshot-23789.bmp

---

### AI Usage


Uri-Tauber:
Auto-rebuild every time the user open the library was considered and rejected. It can take 3-4 secs for medium library (~200 books), which the user will perceive as annoying delay.

itsthisjustin:
Can we say Refresh Library?

Uri-Tauber:
> Can we say Refresh Library?

Sure.

Uri-Tauber:
I changed **Rebuild library index** -> **Refresh library**

Commit 0d72658 adds a "dirty flag" so the library rebuilds its book automatically when we know for sure new book was added.

- When you download a book from the OPDS store (`OpdsBookBrowserActivity`) or upload one via the web server (`CrossPointWebServer`), the code creates a tiny marker file called .crosspoint/library.dirty.
- When you open the Library screen, it checks for that marker. If it's there, the catalog is rebuilt and the marker is deleted.
- If nothing changed, no marker exists, so opening the library stays fast (no rebuild).

The manual refresh is still needed if the user add new books directly to SD or via usb.

NOTE: I cant test the dirty flag right now. Help will be appreciated.

unnamedd:
@Uri-Tauber I am testing in the devices right now, and there are only 2 things that called my attention:
1. Perhaps, would be better if also add the refresh button in the topbar for the non-touch devices. As we do have the magnifying class to search, instead of adding it to the left, we would do:
```
Library                  [  search  ]    [  refresh  ]
```
This way, when we navigate up, the first focus would be on refresh, otherwise we don't go to search.
But this is just an idea, you can ignore.

2. Overall it looks good and it is working fine for me in the real devices, however, I don't think we should list files from the hidden folders in the Library. I saw you filtered files (before I was seeing more files), but I guess we can ignore anything form folders starting with `.`.  What do you think?

itsthisjustin:
> @Uri-Tauber I am testing in the devices right now, and there are only 2 things that called my attention:
> 1. Perhaps, would be better if also add the refresh button in the topbar for the non-touch devices. As we do have the magnifying class to search, instead of adding it to the left, we would do:
> ```
> Library                  [  search  ]    [  refresh  ]
> ```
> This way, when we navigate up, the first focus would be on refresh, otherwise we don't go to search.
> But this is just an idea, you can ignore.
> 
> 2. Overall it looks good and it is working fine for me in the real devices, however, I don't think we should list files from the hidden folders in the Library. I saw you filtered files (before I was seeing more files), but I guess we can ignore anything form folders starting with `.`.  What do you think? 

This was honestly what I had in mind

Uri-Tauber:
@unnamedd thanks for testing. 

On point 1, I already considered that, basicly to have two tab bars and the top one will include actions, but I decided against that. From UX perspective, I think **search** is much more frequent action than refreshing the library, and such design will mean you have to press three (maybe two...) times to get there instead of one.

The only thing I dont like about the my current design is you can get the popup menu only when on **Recent**. On other lists, long press collapse the list. 

Regarding point 2, I thought we already filtering those, but I could be wrong. Should be an easy fix. 

I'm not sure on which point @itsthisjustin agreed.

itsthisjustin:
> @unnamedd thanks for testing. 
> 
> On point 1, I already considered that, basicly to have two tab bars and the top one will include actions, but I decided against that. From UX perspective, I think **search** is much more frequent action than refreshing the library, and such design will mean you have to press three (maybe two...) times to get there instead of one.
> 
> The only thing I dont like about the my current design is you can get the popup menu only when on **Recent**. On other lists, long press collapse the list. 
> 
> Regarding point 2, I thought we already filtering those, but I could be wrong. Should be an easy fix. 
> 
> I'm not sure on which point @itsthisjustin agreed.

I think recent is fine as a location considering its the first place you'd look. The issue for me is its entirely hidden. We will have so many inbound support tickets saying my new files aren't showing up if we dont automate it and maybe 50% less if we show a refresh icon 

Yes we should also hide hidden files

Uri-Tauber:
> I think recent is fine as a location considering its the first place you'd look. The issue for me is its entirely hidden. We will have so many inbound support tickets saying my new files aren't showing up if we don't automate it and maybe 50% less if we show a refresh icon

Well, it should be pretty intuitive for anyone who ever used smartphone. Long press open the options menu.

I do agree in principle, that's the reason I did add refresh icon on touch devices... We're pretty limited with buttons-only device. (And its fun to work along those limits, but still).

unnamedd:
@itsthisjustin one thing that I have tested and it worked fine was regarding the `.dirty`. When I uploaded books via USB and downloaded from OPDS server and opened the Library, the library was refreshed. 

With that, I guess we can kinda assume the refresh button at the top might be just redundant. 

Regarding the non-touch devices, my biggest concern is to have it too hidden, it is not an easy thing to discover in a first glance, that's why I suggested to move it to the top, like in the touch devices.

unnamedd:
> On point 1, I already considered that, basicly to have two tab bars and the top one will include actions, but I decided against that. From UX perspective, I think **search** is much more frequent action than refreshing the library, and such design will mean you have to press three (maybe two...) times to get there instead of one.

@Uri-Tauber 100% with you here, on the other hand, it is the one who is giving me gray hair haha, because I used to navigate up when I want to go to the last item in the list, and now, with the search there, when I go up, I end up directly in the search. Perhaps, if we just select the item and do not open the keyboard, it would be even better.

Uri-Tauber:
> @Uri-Tauber 100% with you here, on the other hand, it is the one who is giving me gray hair haha, because I used to navigate up when I want to go to the last item in the list, and now, with the search there, when I go up, I end up directly in the search. Perhaps, if we just select the item and do not open the keyboard, it would be even better.

If you on the list and press up you jump to the last item. Only if you are on the tabs bar and press up you open search. I imagine people will open search by accident once or twice and then understand how it working. The search icon above the tabs is good indication of "Press up to search".

I do see your point, though. Feel free to edit the PR to see if you could come with better UX.

unnamedd:
> I do see your point, though. Feel free to edit the PR to see if you could come with better UX.

@Uri-Tauber I honestly do think we can make this kind improvement later if we understand it worth an improvement, or if it starts to bother too many people. The awesome is the enemy of the good, it is good what we achieved so far, and I am okay with that.

I really just had those 2 points I raised before, and (again) we can improve later.

Uri-Tauber:
I checked @unnamedd second point, and the scanner already rejects dot-prefixed entries before directory recursion at `lib/LibraryIndex/LibraryBuilder.cpp:472`.

itsthisjustin (approved):

rxmmah:
Why don’t you guys use a short press of the power button to refresh? :)

PR #3609, opened 2026-09-18
fix: keep JLREQ line-edge characters off the head and foot of a line

osakanataro:
## Summary

* **What is the goal of this PR?** Stop Japanese text from wrapping with ー, ・, 々, ‐, … or ％ at the head of a line, or with ¥ / № at the foot. The CJK line-break rules only knew brackets and sentence punctuation; every other class that Japanese typesetting keeps off a line edge was breakable.
* **What changes are included?** The line-start and line-end prohibitions in `ParsedText` are extended to the character classes of [JLREQ](https://www.w3.org/TR/jlreq/) (W3C, *Requirements for Japanese Text Layout*, §3.1.7 and §3.1.8, members from Appendix A), a same-mark run of leaders and dashes (…… ――) is kept together, and the predicates move out of `ParsedText.cpp`'s anonymous namespace into `CjkLineBreak.h` so they can be unit-tested — the same shape as `TokenBoundary.h`. A gtest suite pins each class. `SECTION_FILE_VERSION` 46 → 47, since cached layouts change.

### What was missing

`isNoBreakBeforeCjkPunctuation()` / `isNoBreakAfterCjkPunctuation()` came in with #2288 and cover brackets, quotation marks and sentence punctuation (、。！？ and their ASCII/fullwidth forms). JLREQ §3.1.7 states the rule as: *"In principle, no line should begin with closing brackets (cl-02), hyphens (cl-03), dividing punctuation marks (cl-04), middle dots (cl-05), full stops (cl-06), commas (cl-07), iteration marks (cl-09), a prolonged sound mark (cl-10), small kana (cl-11) …"* — and the missing classes turn up in ordinary prose far more often than a bracket lands on a line edge:

| JLREQ class | characters added | example that used to wrap wrong |
|---|---|---|
| cl-10 prolonged sound mark | ー ｰ | サーバ**ー** → ー alone at the head of a line |
| cl-05 middle dots | ・ ･ | ジョン**・**スミス |
| cl-09 iteration marks | 々 〻 ゝ ゞ ヽ ヾ | 人**々** |
| cl-11 small kana | ぁぃぅぇぉっゃゅょゎゕゖ ァィゥェォッャュョヮヵヶ ㇰ–ㇿ | ニャ**ー** → ャ alone at the head of a line |
| cl-04 beyond ! ? | ‼ ⁇ ⁈ ⁉ | |
| cl-03 hyphens | ‐ – 〜 ゠ ～ | 一〜二 |
| cl-08 inseparable | — ― ‥ … (and same-mark runs) | ……で始まる行 |
| cl-13 postfixed abbreviations | % ％ ℃ ′ ″ ° ‰ ¢ ￠ ℓ ㏋ and the squared unit symbols | ５０**％** |
| cl-12 prefixed abbreviations (line end) | ¥ ￥ $ ＄ £ ￡ # ＃ € № | **¥**500 split before the number |
| cl-01 / cl-02 gaps | 〝 〟 ｟ ｠ ｡ ､ | |

Small kana are part of the rule, not an addition: §3.1.7's Note records that "not a small number of books" relax 々, ー and small kana, but the principle keeps them off the line head, and that is what this PR follows.

### How the mechanism works

Nothing about the line breaker changes. CJK text is tokenized one character at a time, and `hasCjkBreakOpportunityBetween()` decides for each adjacent pair whether the tokenizer may split there. A pair with no opportunity stays in one token, so a prohibited character simply travels with its neighbour. That is transitive, which is why `『「漢」』。` never splits anywhere without any pull-back loop. Latin runs are untouched: the predicate returns false before any table lookup unless one side is in a CJK block.

## Scope Check

Please confirm:

- [x] The stock firmware does not already handle this well, **and** no other popular CrossPoint fork already does. crosspoint-jp (zrn-ns) added the same classes on 2026-09-13 in its own `Kinsoku.h` with a pull-back pass; this PR reaches the same result through the existing glue-based tokenizer, so the change to the layout code is a table, not a new pass.

## Additional Context

**Tests.** `test/cjk_line_break/CjkLineBreakTest.cpp` (9 cases): every added character is asserted against a kanji and a hiragana neighbour, prohibitions are one-sided (」漢 and ー漢 still break), nested brackets chain, leader runs stay together but release what follows, small kana cannot start a line, and the Appendix A members beyond the ASCII forms (‼ ¢ ＄) are covered. Run with the usual `cmake -S test -B build/test && cmake --build build/test --target CjkLineBreakTest`.

**Device.** Verified on an Xteink X3 (ESP32-C3) with a firmware built from this branch, using a test EPUB whose first chapter forces each class onto a line edge (three pages, 18 pt CJK SD font): ー, ・, ‥, …, 〜, 」, 。 and small kana (ニャーニ**ャ**ー, which used to split before the ャ) no longer start a line; 「 （ 『 【 〔 〈 《 no longer end one; …… and ―― stay together. A vertical-writing book from the same fork showed no change, as expected (vertical layout uses its own column breaker). One observation unrelated to this PR: text inside an inline `<code>` element is attached to its neighbours across the element boundary, which can leave a badly spread justified line before it; that behaviour is the same on `develop`.

**Memory / flash impact.** No new allocation and no stored state; the predicates are `switch` statements, so the tables live in flash.

**Risk.** Only pairs where at least one side is CJK are affected. A pair that used to break and no longer does produces a slightly longer glued token; for the worst realistic case (a long run of ……) that token is still shorter than one line. `SECTION_FILE_VERSION` is bumped so existing section caches are rebuilt on first open.

---

### AI Usage

Did you use AI tools to help write this code? _**YES**_ — investigation, implementation, tests, and this description.

https://claude.ai/code/session_01GSMiDKH4CDUUtu78uJCe4F


Uri-Tauber:
Thats looks awfully specific to **your** use-case.
Can you think of a way to do it more generally?

osakanataro:
Fair point — the tables are named after JLREQ, and that makes it read as a Japanese-only rule. The characters themselves are not Japanese-specific, though: every class the PR adds is one that Unicode's line breaking algorithm ([UAX #14, *Unicode Line Breaking Algorithm*](https://www.unicode.org/reports/tr14/), with the per-character classes in [`LineBreak.txt`](https://www.unicode.org/Public/UCD/latest/ucd/LineBreak.txt)) already defines for all scripts, and the pair rules are the standard ones from that report:

| PR table (JLREQ name) | UAX #14 class | rule |
|---|---|---|
| middle dots, iteration marks, ‼⁇⁈⁉, 〜゠ | NS (nonstarter) | LB16: no break before |
| prolonged sound mark ー, small kana | CJ (conditional Japanese starter) | treated as NS under `line-break: strict` |
| postfixed % ％ ℃ ′ ″ ° ‰ ¢ | PO | LB23/LB25: no break before |
| prefixed ¥ $ £ € № | PR | LB23/LB25: no break after |
| … ‥ | IN (inseparable) | LB22: no break before; IN × IN |
| — | B2 | B2 × B2 |
| 〝〟｟｠｡､ | OP / CL | LB13/LB14 (same as the brackets #2288 already handles) |

So the general form of this change is: drop the hand-written character lists (mine and the ones from #2288), replace them with a compact table of UAX #14 classes for the code points that can sit next to CJK text, generated from `LineBreak.txt` by a script kept in the repo, and let `hasCjkBreakOpportunityBetween()` apply the class pair rules (no break before CL/CP/EX/NS/PO/IN, no break after OP/PR, B2×B2, IN×IN). The mechanism, the tests and the section-version bump stay as they are; only the source of truth changes from a hand list to the Unicode property. Chinese and Korean punctuation (，．、。！？「」etc.) then get the same treatment for free, and a future character needs no code change — just a regenerated table.

Two things I'd like your view on before I rework it:

1. **CJ.** UAX #14 says CJ acts as NS under strict line breaking and as ID (breakable) under normal/loose. Japanese publishers' EPUBs almost always want strict for ー and small kana; I'd propose strict as the default, since that is what every Japanese reader app does, but it could also follow the CSS `line-break` property if you'd rather keep it neutral.
2. **Scope of the table.** Applying UAX #14 only to pairs where one side is CJK (as now) keeps Latin breaking untouched. Going further — UAX #14 for all text — is a much larger change to the layout code and I'd keep it out of this PR.

If that's the shape you had in mind, I'll rework the PR that way; if you meant something else by "more general", let me know and I'll adjust. The five characters that JLREQ lists but Unicode does not classify as PO/PR (ℓ ㏋ ㎡ # ―) would simply drop out, which I'm fine with.

PR #3611, opened 2026-09-18
fix: improves word lookup by improving stripping logic and lowercasing word

alycklama:
## Summary

- **What is the goal of this PR?** Fixes dictionary word lookups in `DictionaryWordSelectActivity` that were silently failing whenever the selected word had leading/trailing punctuation or wasn't an exact-case match for its headword — most visibly for non-Latin scripts like Greek (`"Γεια"`, `«Γεια»`, `ουρλιάζει"...»`), but the fix is script-agnostic.

  - `Dictionary::cleanWord()` (`src/util/Dictionary.cpp`) is rewritten to decode the word as full UTF-8 codepoints instead of peeking at raw bytes. The previous byte-level check treated _any_ byte ≥ 0x80 as "part of a word," so multi-byte punctuation whose lead byte wasn't part of the one hand-picked 3-byte range (e.g. guillemets `«»`, ellipses, CJK/fullwidth punctuation) was never stripped from word edges.
  - Word-edge classification is now Unicode-category-driven rather than a hand-picked punctuation blacklist: a codepoint counts as "part of the word" iff it's a Letter, Number, or Mark
    (`L*`/`N*`/`M*`). Everything else — quotes, ellipses, brackets, symbols, in any script — is trimmed from both ends. This removes the maintenance burden of enumerating punctuation ranges and closes the class of bug entirely, not just the specific characters reported.
  - Lookup keys are now case-folded before the dictionary search, so e.g. `"Γεια"` and `"γεια"` resolve to the same headword. Folding is table-driven from Unicode case-mapping data rather than hardcoded per-script logic, so it isn't limited to Greek — it works for any script with a simple 1:1 case pairing (Latin, Greek, Cyrillic, Armenian, Cherokee, Deseret, etc.).
  - Two new generated lookup tables added under `lib/Utf8/`:
    - `Utf8WordCharTable.h` — 794-range table of Letter/Number/Mark codepoints (~6.2KB flash), backing the new `utf8IsWordChar()`.
    - `Utf8CaseFoldTable.h` — 181-range table of simple lowercase case-fold mappings (~2.8KB flash), backing the new `utf8ToLowerSimple()`.

      Both are generated from Python's `unicodedata` (documented in the file headers) and binary-searched, following the same pattern as the existing `Utf8ComposeTable.h`/`utf8ComposeNfc()`
      in this file.

## Scope Check

Please confirm:

- [ ] ~~If this PR touches `freeink-sdk/`, `lib/hal/`, the bootloader, OTA, or recovery code, I have coordinated with the relevant maintainer.~~

## Additional Context

- This is a bug fix to the existing dictionary lookup feature, not new functionality — no scope expansion.
- **Memory / flash impact**: two new `static constexpr` tables live in flash (.rodata), included only from `Utf8.cpp` (single translation unit, matching the existing `Utf8ComposeTable.h` pattern) so there's no per-includer duplication: `Utf8WordCharTable.h` (794 × 8-byte ranges ≈ 6.2KB) and `Utf8CaseFoldTable.h` (181 × 16-byte ranges ≈ 2.8KB) — roughly 9KB flash total. No new RAM/heap usage — `cleanWord()` still builds one `std::string` on the stack per lookup, same as before.
- `Dictionary::cleanWord()` is on the dictionary lookup path only (triggered on Confirm/tap in `DictionaryWordSelectActivity`), not the render/scroll hot path, so the switch from a byte scan to codepoint decoding plus two binary searches per boundary codepoint has no perceptible performance impact.
- Areas to focus review on: the generated table headers are large and not hand-reviewable line-by-line — the file header comments explain provenance (Python `unicodedata`, Unicode 15.0.0) and the binary-search lookup functions in `Utf8.cpp` are the part worth reviewing closely.
- Tested extensively on my X4 and it works fine without any issues.

---

### AI Usage


lib/Utf8/Utf8CaseFoldTable.h
```diff
@@ -0,0 +1,100 @@
+// Auto-generated simple (single-codepoint) Unicode case-folding table,
+// derived from Python 3.12.3's unicodedata (Unicode 15.0.0): for every
```
brianhuster, line 2: Shouldn't the Python script be commited too?

alycklama: That's a good point. I'll add the script to generate the fold table.

alycklama:
@coderabbit review

PR #3618, opened 2026-09-18
feat: choose which UI languages are compiled in

eszter007:
## Summary

* **What is the goal of this PR?** Let a build choose which UI languages are compiled in, so a fork that only needs a few of them can drop the rest. This is the slimmer version of #3163, following @Uri-Tauber's review there: the whole thing is a `gen_i18n.py` flag, without language packs loading from the SD card.
  * `scripts/gen_i18n.py` reads `custom_i18n_builtin_langs` from `platformio.ini` (and accepts `--builtin-langs` on the command line). It emits string tables only for the listed languages, plus English.
  * `platformio.ini` declares the option in `[base]` with the default `all`.
  * `LanguageSelectActivity.h`: the picker sizes its list from the generated `SORTED_LANGUAGE_INDICES` instead of the whole `Language` enum. This is the only firmware change.
  * `docs/i18n.md`: a short "Building With Fewer Languages" section.

```ini
custom_i18n_builtin_langs = all      ; default: every language, output unchanged
custom_i18n_builtin_langs = de,fr    ; English, German and French only
```

### What gets left out, and what doesn't

Only the per-language **string blobs and offset tables** are dropped, and those are the bulk of the cost. Every language keeps its `Language` enum value, code and display name, because code elsewhere names languages directly:

* `keyboard_layouts::ALL` pairs each layout with a `Language`, uses it to label the row, and treats its row order as the persisted bitmask. A French-less build still offers the AZERTY layout, and it still needs to be called "Français".
* `V1_LANGUAGES` (the `language.bin` migration table) names 22 languages.

Keeping the enum complete means both compile unchanged. The alternative was `#if` guards around every such reference.

### Behaviour

* **Picker:** lists only the built-in languages.
* **Saved language not in the build** (e.g. a device flashed from a full build to a reduced one): `getLanguageStrings()` already has a `default:` branch returning English, and the generator no longer emits a `case` for left-out languages, so they fall into it. The device shows English and does not crash. The saved code is left alone, so flashing a build that includes the language again brings it back. `CHARACTER_SETS` gives such a language the English character set, to match the strings it now renders.
* **English** is always included, since every other language's offset table points into it.
* **An unknown code is an error**, listing the valid codes. A typo cannot silently ship a build without a language.

## Scope Check

Please confirm:

      the relevant maintainer.

Flash footprint is listed as a current focus area in SCOPE.md. This PR touches no `freeink-sdk/`, `lib/hal/`, bootloader, OTA or recovery code.

## Additional Context

**Nothing changes by default.** With `custom_i18n_builtin_langs = all`, the three generated files (`I18nKeys.h`, `I18nStrings.h`, `I18nStrings.cpp`) are **byte-identical** to `develop`'s, diffed directly.

**Flash**, `pio run -e default` on `develop` at `4b17a7bb`:

| `custom_i18n_builtin_langs` | flash used | bytes |
|---|---|---|
| `all` (default) | 85.0% | 5,568,565 |
| `de,fr` | 79.8% | 5,229,163 (**−339,402 B**) |

**Tested with a `de,fr` build:**
* the language picker lists exactly English, Deutsch and Français
* with the saved language set to Russian (not built in), the device boots into English with no errors, and no picker row is marked selected
* selecting Deutsch switches the UI to German
* the keyboard-layout screen still lists all ten layouts, in order, with correct names, including the seven whose languages are not built in

image-1789768391352

`pio run` passes at both settings.

Supersedes #3163.

---

### AI Usage


Uri-Tauber (approved):
LGTM

Issue #2531, opened 2026-07-03
Allow users to select fonts for UI


brianhuster:
It seems like the UI font is hardcoded and cannot be changed by users, which mean many book titles are not displayed in the menu if they contain characters not supported by the UI font

So Crosspoint should allow users to choose a font for the UI

noomly:
Out of curiosity, why was this issue closed? Will it be possible to change the UI's font in 1.5?

brianhuster:
Crosspoint will use reader font in case there are characters not in built-in font. That sound enough for me, so I close it

noomly:
It would still be useful to be able to change the whole set to a custom font though. I could open a new one but since I would use the same description maybe we could leave that one open?

brianhuster:
In this project, feature requests are supposed to go to "Discussion" anyway. It was my mistake for opening it in "Issues"

PR #3624, opened 2026-09-19
feat: add dictionary font selection

rxmmah:
## Summary

  - Adds dictionary font family selection independently of the book font.
  - Adds Settings → Reader → Dictionary Font Family with Same as Reader, Noto Serif, Noto Sans, and SD Font.
  - SD Font opens a scrollable picker of installed font families. The selection persists across restarts.
  - Applies the selected font to plain-text and HTML definitions, using the nearest supported size to the reader’s point size.

  ## Scope Check

  Upstream currently uses the book font for dictionary definitions. This change allows a separate family while preserving that behavior by default.

  No SDK, HAL, bootloader, OTA, or recovery files are changed. The maintainer-coordination item is not applicable.

  ## Additional Context

  - SD dictionary fonts temporarily replace the loaded book font; the book font is restored when the definition closes. Missing or unreadable dictionary fonts fall back to the reader font.

  - Picker arrays are allocated once per visit and freed when it closes. No second SD font family is kept resident. Exact RAM and flash deltas have not been measured.

### AI Usage


rxmmah:
@Uri-Tauber hey, will be this helpful?

Uri-Tauber:
@rxmmah Hmmm. I dont know. 
All the SD fonts already include IPA coverage, if I'm not mistaken.

brianhuster:
Do people often want to dictionary font to be the same as UI font, reader font, or different from both? If people want dictionary font to be same as UI font, we can add an UI font setting instead, because I recall people requested that https://github.com/crosspoint-reader/crosspoint-reader/issues/2531#issuecomment-5072979034

rxmmah:
> Do people often want to dictionary font to be the same as UI font, reader font, or different from both? If people want dictionary font to be same as UI font, we can add an UI font setting instead, because I recall people requested that [#2531 (comment)](https://github.com/crosspoint-reader/crosspoint-reader/issues/2531#issuecomment-5072979034)

UI font will cover the languages issue. like if you look up english words in japanese.

